PR #1, opened 2017-10-27
Moved to JanusAdapter, removed easyrtc and server, updated build system.

robertlong:
[naf-janus-adapter](https://github.com/robertlong/naf-janus-adapter) is finally working with networked-aframe!

This PR moves us to using the Janus adapter which makes all of the easyrtc stuff obsolete. I removed the signaling server and moved us back to webpack-dev-server. There's a proper build step with babel-minify integrated. Everything should be set up property for a future deploy script for static hosting.

Sorry it's sort of a noisy PR. I have Prettier set to format on save. Maybe we should enforce Prettier rules on a pre-commit hook?


public/index.html
```diff
@@ -39,72 +36,45 @@
         ></a-entity>
       </script>
 
-      <script id="hand-template" type="text/html">
+            <script id="hand-template" type="text/html">
```
netpro2k: Whats with this indenting?

package.json
```diff
@@ -4,28 +4,26 @@
...
   "dependencies": {
     "aframe-input-mapping-component": "https://github.com/fernandojsg/aframe-input-mapping-component",
     "aframe-teleport-controls": "https://github.com/netpro2k/aframe-teleport-controls#feature/teleport-origin",
-    "easyrtc": "^1.1.0",
-    "express": "^4.15.5",
-    "networked-aframe": "https://github.com/netpro2k/networked-aframe#feature/positional-audio",
-    "pleasejs": "^0.4.2",
-    "socket.io": "^2.0.3"
+    "naf-janus-adapter": "^0.1.1",
+    "networked-aframe": "https://github.com/netpro2k/networked-aframe#feature/register-adapter",
```
netpro2k: Make sure this branch includes latest changes on master of main fork of networked-aframe as that includes positional audio changes.

public/index.html
```diff
@@ -1,33 +1,30 @@
...
+                     room: 2;
+                     serverURL: wss://quander.me:8989;
                      audio: true;
                      debug: true;
-                     connectOnLoad: false"
-    mute-mic="eventSrc: a-scene; toggleEvents: action_mute">
-    <a-assets>
-      <img id="grid" src="assets/grid.png" crossorigin="anonymous">
-      <img id="sky" src="https://cdn.aframe.io/360-image-gallery-boilerplate/img/sechelt.jpg" crossorigin="anonymous" />
+                     connectOnLoad: false" mute-mic="eventSrc: a-scene; toggleEvents: action_mute">
```
netpro2k: Minor, but prefer to keep attributes broken up like they were before.

public/index.html
```diff
@@ -39,72 +36,45 @@
...
+                networked="template:#hand-template;showLocalTemplate:false;"></a-entity>
+        </a-entity>
 
-    <a-entity class="head" gltf-model="#rock-island" position="0 0 0">
-    <a-sky color="#DDFFD9"></a-sky>
-  </a-scene>
-  <script>
-   document.querySelector('a-scene').addEventListener('loaded', onSceneLoad)
-  </script>
+        <a-entity class="head" gltf-model="#rock-island" position="0 0 0">
```
netpro2k: Ah actually looks like an unclosed tag here, wonder if that's why indenting got all screwey

public/index.html
```diff
@@ -1,33 +1,30 @@
...
+    <script src="https://aframe.io/releases/0.7.0/aframe.min.js"></script>
+    <script src="./app.bundle.js"></script>
 </head>
 
 <body>
-  <a-scene
-    networked-scene="adapter: easyrtc;
+    <a-scene networked-scene="adapter: janus;
+                     room: 2;
+                     serverURL: wss://quander.me:8989;
```
netpro2k: Is all of us hitting this server going to cause any issues with any testing you guys are doing?

robertlong: I put this demo on another room id. So we'll be separated from all that.

brianpeiris:
We should use https://github.com/observing/pre-commit if we want to enforce prettier. It would get messy otherwise.

public/index.html
brianpeiris: More than minor, IMO. I find long lines of attributes pretty unreadable, especially for A-Frame.

hthetiot:
note networked-aframe use easyrtc not janus...

PR #2, opened 2017-10-30
Added personal-space-bubble-component.

robertlong:
The personal space bubble component controls the visibility of other objects when they get too close to your own personal space bubble components. Currently the component is added to all avatar heads and hands. There is also a configurable radius property that determines how close other users can get to you before they appear invisible.

The components register themselves with a system that does a distance check every tick.

### DIFF
--- a/src/systems/personal-space-bubble.js
+++ b/src/systems/personal-space-bubble.js
@@ -1,61 +1,81 @@
-var posA = new AFRAME.THREE.Vector3();
-var posB = new AFRAME.THREE.Vector3();
-
-function distance(entityA, entityB) {
-  entityA.object3D.getWorldPosition(posA);
-  entityB.object3D.getWorldPosition(posB);
-  return posA.distanceTo(posB);
-}
+var invaderPos = new AFRAME.THREE.Vector3();
+var bubblePos = new AFRAME.THREE.Vector3();
 
 AFRAME.registerSystem("personal-space-bubble", {
   init() {
-    this.myEntities = [];
-    this.entities = [];
+    this.invaders = [];
+    this.bubbles = [];
   },
 
-  registerEntity(el) {
-    var networkedEl = NAF.utils.getNetworkedEntity(el);
-    var owner = NAF.utils.getNetworkOwner(networkedEl);
+  registerBubble(el) {
+    this.bubbles.push(el);
+  },
 
-    if (owner !== NAF.clientId) {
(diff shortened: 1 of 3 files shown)


src/systems/personal-space-bubble.js
```diff
@@ -0,0 +1,73 @@
+var posA = new AFRAME.THREE.Vector3();
+var posB = new AFRAME.THREE.Vector3();
+
+function distance(entityA, entityB) {
+  entityA.object3D.getWorldPosition(posA);
+  entityB.object3D.getWorldPosition(posB);
+  return posA.distanceTo(posB);
```
brianpeiris: Slightly more efficient to use `distanceToSquared` with a squared radius.

src/systems/personal-space-bubble.js
```diff
@@ -0,0 +1,73 @@
...
+      var otherEntity = this.entities[j];
+
+      var visible = true;
+
+      for (var i = 0; i < this.myEntities.length; i++) {
+        var myEntity = this.myEntities[i];
+
+        var d = distance(myEntity, otherEntity);
+
+        if (d < myEntity.components["personal-space-bubble"].data.radius) {
```
brianpeiris: Maybe cache `radius` in `myEntities` instead of accessing it in this nested loop.

src/systems/personal-space-bubble.js
```diff
@@ -0,0 +1,73 @@
...
+          visible = false;
+          break;
+        }
+      }
+
+      otherEntity.object3D.visible = visible;
```
netpro2k: This should probably be using `setAttribute` though not sure its worth the performance hit. Might lead to some confusion if we try to network this or deal with anything else that might be checking the `visible` component.

src/systems/personal-space-bubble.js
```diff
@@ -0,0 +1,93 @@
...
+      var radius = bubble.components["personal-space-bubble"].data.radius;
+      var radiusSquared = radius * radius;
+
+      // Hide the invader if inside the bubble
+      for (var j = 0; j < this.invaders.length; j++) {
+        var invader = this.invaders[j];
+
+        invaderPos.setFromMatrixPosition(invader.object3D.matrixWorld);
+
+        var distanceSquared = bubblePos.distanceTo(invaderPos);
```
brianpeiris: Should be `distanceToSquared`.

PR #4, opened 2017-10-30
Added hand-controls-child-visibility component and network schema.

### LINKED ISSUE #3: Hand models should be invisible unless controllers are connected.
Currently your own hand models as well as every other avatar's hand models are visible despite them not having tracked controllers. These models should be invisible until the tracked controllers are active.

robertlong:
Fixes #3 and makes hand visibility networked.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -2,7 +2,7 @@
 
 <head>
     <title>A-Frame Social VR Demo</title>
-    <script src="https://aframe.io/releases/0.7.0/aframe.min.js"></script>
+    <script src="https://aframe.io/releases/0.7.0/aframe.js"></script>
     <script src="./app.bundle.js"></script>
 </head>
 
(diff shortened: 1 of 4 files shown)


public/index.html
```diff
@@ -58,7 +58,7 @@
 
             <a-entity id="nametag" networked="template:#nametag-template;showLocalTemplate:false;"></a-entity>
 
-            <a-entity id="left-hand" hand-controls="left" axis-dpad="centerZone: 1" teleport-controls="cameraRig: #player-rig; teleportOrigin: #head; button: action_teleport_"
+            <a-entity id="left-hand" hand-controls="left" visible="false" hand-controls-child-visibility axis-dpad="centerZone: 1" teleport-controls="cameraRig: #player-rig; teleportOrigin: #head; button: action_teleport_"
```
netpro2k: Might want to remove the `visible="false"` here since the `hand-controls-child-visibility` component sets this up in init. As it is now I might expect if I changed this to true that I would see them by default.

src/components/hand-controls-child-visibility.js
```diff
@@ -0,0 +1,21 @@
+AFRAME.registerComponent("hand-controls-child-visibility", {
```
netpro2k: Since this is now controlling the visibility of the element itself (not the children specifically) it should probably be called `hand-controls-visibility`

src/index.js
```diff
@@ -24,6 +25,11 @@ NAF.schemas.add({
   ]
 });
 
+NAF.schemas.add({
+  template: "#hand-template",
+  components: ["position", "rotation", "visible"]
```
netpro2k: This is what I am liking about NAF so far, the networking has no idea about how this visible component is getting updated, and the `hand-controls-child-visibility` component has no idea its dealing with networking.

public/index.html
robertlong: Good point.

src/index.js
robertlong: Yeah, it's holding up really well so far!

PR #9, opened 2017-11-03
virtual-gamepad-controls component for mobile 2d locomotion

robertlong:
Adds two on-screen thumbsticks for movement and horizontal rotation when on mobile.

The query string `?gamepad=true` can be added to force enable the gamepad on desktop.

Moves the enter vr button to the top right of the screen so that it doesn't interfere with the right thumbstick. I could use some feedback on how best to do this. My current CSS doesn't look great on all screen sizes.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -13,12 +13,14 @@
 </head>
 
 <body>
-    <a-scene networked-scene="adapter: janus;
-                         room: 3;
-                         serverURL: wss://quander.me:8989;
-                         audio: true;
-                         debug: true;
-                         connectOnLoad: false;" mute-mic="eventSrc: a-scene; toggleEvents: action_mute">
+    <a-scene
+        networked-scene="adapter: janus;
+                            room: 2;
+                            serverURL: wss://quander.me:8989;
+                            audio: true;
+                            debug: true;
+                            connectOnLoad: false;"
+        mute-mic="eventSrc: a-scene; toggleEvents: action_mute">
 
         <a-assets>
             <img id="grid" src="assets/grid.png" crossorigin="anonymous" />
--- a/public/index.html
+++ b/public/index.html
@@ -4,6 +4,12 @@
     <title>Mozilla Mixed Reality Social Client</title>
     <script src="https://aframe.io/releases/0.7.0/aframe.js"></script>
     <script src="./app.bundle.js"></script>
+    <style>
+        .a-enter-vr {
+            top: 90px;
+            bottom: auto;
+        }
+    </style>
 </head>
 
 <body>
--- a/public/index.html
+++ b/public/index.html
@@ -7,14 +7,12 @@
 </head>
 
 <body>
-    <a-scene
-        networked-scene="adapter: janus;
-                         room: 2;
+    <a-scene networked-scene="adapter: janus;
+                         room: 3;
                          serverURL: wss://quander.me:8989;
                          audio: true;
                          debug: true;
-                         connectOnLoad: false;"
-        mute-mic="eventSrc: a-scene; toggleEvents: action_mute">
+                         connectOnLoad: false;" mute-mic="eventSrc: a-scene; toggleEvents: action_mute">
 
         <a-assets>
             <img id="grid" src="assets/grid.png" crossorigin="anonymous" />
@@ -54,7 +52,7 @@
             </script>
         </a-assets>
 
-        <a-entity id="player-rig" networked wasd-controls snap-rotation="pivotSrc: #head">
+        <a-entity id="player-rig" networked virtual-gamepad-controls wasd-controls snap-rotation="pivotSrc: #head">
             <a-sphere scale="0.1 0.1 0.1"></a-sphere>
             <a-entity id="head" camera="userHeight: 1.6" personal-space-bubble look-controls networked="template:#head-template;showLocalTemplate:false;"></a-entity>
 
--- a/src/components/virtual-gamepad-controls.js
+++ b/src/components/virtual-gamepad-controls.js
@@ -24,13 +24,15 @@ AFRAME.registerComponent("virtual-gamepad-controls", {
     const leftStick = nipplejs.create({
       zone: leftTouchZone,
       mode: "static",
-      color: "white"
+      color: "white",
+      position: { left: "50px", bottom: "50px" }
     });
 
     const rightStick = nipplejs.create({
       zone: rightTouchZone,
       mode: "static",
-      color: "white"
+      color: "white",
+      position: { right: "50px", bottom: "50px" }
     });
 
     this.onJoystickChanged = this.onJoystickChanged.bind(this);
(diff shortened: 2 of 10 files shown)


public/index.html
```diff
@@ -4,17 +4,21 @@
     <title>Mozilla Mixed Reality Social Client</title>
     <script src="https://aframe.io/releases/0.7.0/aframe.js"></script>
     <script src="./app.bundle.js"></script>
+    <style>
+        .a-enter-vr {
+            top: 90px;
+            bottom: auto;
+        }
```
brianpeiris: This is fine for now. We'll probably rework it when we have our own HUD/UI

public/index.html
```diff
@@ -4,17 +4,21 @@
...
-        networked-scene="adapter: janus;
-                         room: 2;
+    <a-scene networked-scene="adapter: janus;
+                         room: 3;
                          serverURL: wss://quander.me:8989;
                          audio: true;
                          debug: true;
-                         connectOnLoad: false;"
-        mute-mic="eventSrc: a-scene; toggleEvents: action_mute">
+                         connectOnLoad: false;" mute-mic="eventSrc: a-scene; toggleEvents: action_mute">
```
brianpeiris: Still using prettier on the HTML? 
Also, why change the room number?

src/components/virtual-gamepad-controls.js
```diff
@@ -0,0 +1,188 @@
...
+    leftTouchZone.style.top = 0;
+    leftTouchZone.style.bottom = 0;
+    leftTouchZone.style.right = "50%";
+
+    var rightTouchZone = document.createElement("div");
+    rightTouchZone.style.position = "absolute";
+    rightTouchZone.style.left = "50%";
+    rightTouchZone.style.top = 0;
+    rightTouchZone.style.bottom = 0;
+    rightTouchZone.style.right = 0;
```
brianpeiris: I would put this in a CSS file and assign a className here instead.

brianpeiris: Actually, maybe this is fine since it's nice that the component is self-contained.

src/components/virtual-gamepad-controls.js
```diff
@@ -0,0 +1,188 @@
...
+    } else {
+      if (event.type === "move") {
+        // Set pitch and yaw angles on right stick move
+        var angle = data.angle.radian;
+        var force = data.force < 1 ? data.force : 1;
+        this.yaw = Math.cos(angle) * -force;
+        // this.pitch = Math.sin(angle) * force;
+      } else {
+        this.yaw = 0;
+        // this.pitch = 0;
```
brianpeiris: Delete commented code

src/components/virtual-gamepad-controls.js
```diff
@@ -0,0 +1,188 @@
+import nipplejs from "nipplejs";
+
+var THREE = AFRAME.THREE;
+var DEGREES = Math.PI / 180;
+var HALF_PI = Math.PI / 2;
```
brianpeiris: These constants are unused.

src/components/virtual-gamepad-controls.js
```diff
@@ -0,0 +1,188 @@
...
+    // Apply joystick yaw rotation
+    obj.applyMatrix(this.yawMatrix);
+    // Reapply camera (head) rotation
+    obj.applyMatrix(this.camRotationMatrix);
+    // Reapply playspace (player rig) rotation
+    obj.applyMatrix(this.rotationMatrix);
+    // Reapply playspace (player rig) translation
+    obj.applyMatrix(this.trans);
+
+    // @TODO this is really ugly, can't just set the position/rotation directly or they wont network
```
brianpeiris: How is it ugly?

brianpeiris:
lgtm!

src/components/virtual-gamepad-controls.js
```diff
@@ -0,0 +1,188 @@
...
+
+    this.cameraEl = document.querySelector("[camera]");
+
+    this.onEnterVr = this.onEnterVr.bind(this);
+    this.onExitVr = this.onExitVr.bind(this);
+    this.el.sceneEl.addEventListener("enter-vr", this.onEnterVr);
+    this.el.sceneEl.addEventListener("exit-vr", this.onExitVr);
+  },
+
+  onJoystickChanged(event, data) {
```
netpro2k: Id use someting other than `data` here as I at first confused it with `this.data`, maybe `details`?

src/components/virtual-gamepad-controls.js
netpro2k: Because you would want to just be able to call `this.el.setAttribute("rotation", someQuaternion)` and not have to create a new object with components all in degrees only to have aframe convert them back to radians and stick them back in a quaternion.

public/index.html
brianpeiris: +1 to either using a sane HTML formatter (not prettier) or leaving formatting as you found it.

src/components/virtual-gamepad-controls.js
brianpeiris: Sure, but that's a-frame. I don't see the value of pointing out in a TODO if we can't actually do anything about it.

public/index.html
robertlong: Yeah I should have used that instead. Also, I’m not sure what’s up with prettier. This file should be ignored. Maybe it’s VSCode’s default html format? I have everything set to format on save for prettier but I should probably make sure that ignores html.  I’ll look into it when I get back.

src/components/virtual-gamepad-controls.js
robertlong: Yeah. These can be deleted. I found that out after looking at the snap rotation component.

src/components/virtual-gamepad-controls.js
```diff
@@ -0,0 +1,188 @@
...
+
+AFRAME.registerComponent("virtual-gamepad-controls", {
+  schema: {
+    movementSpeed: { default: 2 },
+    lookSpeed: { default: 60 }
+  },
+
+  init() {
+    // Setup gamepad elements
+    var leftTouchZone = document.createElement("div");
```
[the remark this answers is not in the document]
robertlong: I need to understand how Babel transforms const, but I think that may be a good lint rule.

src/components/virtual-gamepad-controls.js
robertlong: We can use a webpack loader for this.

src/components/virtual-gamepad-controls.js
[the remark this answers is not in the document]
robertlong: Looks fine: https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&code_lz=MYewdgzgLgBAHjAvDKAnArgUwNxA&debug=false&circleciRepo=&evaluate=false&lineWrap=true&presets=es2015%2Creact%2Cstage-2&targets=&version=6.26.0

I'll add it.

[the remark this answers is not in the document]
brianpeiris: I wonder if we need babel, or at least if we can turn off specific transforms. Most browsers support `const` natively now.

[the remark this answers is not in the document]
brianpeiris: Yeah, we can basically target es2016 now. babel will leave `const` and `let` untransformed.

PR #10, opened 2017-11-08
Feature/character controller

johnshaughnessy:
Basic character controller working with vive, wasd, and touchscreen inputs.
Fixes https://github.com/mozilla/socialmr/issues/58 (should I link this here?)


public/index.html
```diff
@@ -59,15 +58,11 @@
                     scale="6 6 6"></a-entity>
             </script>
         </a-assets>
+        <a-entity id="player-rig" networked character-controller="pivot: #head">
+            <a-sphere scale="0.1 0.1 0.1"></a-sphere>
+            <a-entity id="head" camera="userHeight: 1.6" personal-space-bubble networked="template:#head-template;showLocalTemplate:false;"<script src="<a-entity id="nametag" networked="template:#nametag-template;showLocalTemplate:false;"></a-entity>
```
brianpeiris: What's with this `"<script src="<a-entity id="nametag"`? Looks like a bad merge.

src/components/character-controller.js
```diff
@@ -0,0 +1,300 @@
...
+    );
+  },
+
+  onStopMoving: function(event) {
+    this.accelerationInput.set(0, 0, 0);
+  },
+
+  onTranslateX: function(event) {
+    // The type check here is because the last axismove event that is captured here sends
+    // the el as the event.detail. This should probably be caught earlier.
```
brianpeiris: Is this still true? I can't tell where the event.detail is set to el.

src/components/character-controller.js
```diff
@@ -0,0 +1,300 @@
...
+  onMoveLeft: function(event) {
+    this.accelerationInput.x = -0.8;
+  },
+
+  onDontMoveLeft: function(event) {
+    this.accelerationInput.x = 0;
+  },
+
+  onMoveRight: function(event) {
+    this.accelerationInput.x = 0.8;
```
brianpeiris: Probably put `0.8` in a constant and reuse it.

src/components/character-controller.js
```diff
@@ -0,0 +1,300 @@
...
+var MAX_DELTA = 0.2;
+var TAU = Math.PI * 2;
+
+// Does not have any type of collisions yet.
+AFRAME.registerComponent("character-controller", {
+  schema: {
+    groundAcc: { default: 10 },
+    easing: { default: 8 },
+    pivot: { type: "selector" },
+    snapRotationRadian: { default: TAU / 8 }
```
brianpeiris: Maybe just use `45 * THREE.DEG2RAD` here and get rid of `TAU`.

src/components/character-controller.js
```diff
@@ -0,0 +1,300 @@
...
+    var rotationInvMatrix = new THREE.Matrix4();
+    var pivotRotationMatrix = new THREE.Matrix4();
+    var pivotRotationInvMatrix = new THREE.Matrix4();
+    var position = new THREE.Vector3();
+    var currentPosition = new THREE.Vector3();
+    var movementVector = new THREE.Vector3();
+
+    return function(t, dt) {
+      const deltaSeconds = dt / 1000;
+      const rotationSpeed = -3;
```
robertlong: Probably best to make `rotationSpeed` a constant as well. Maybe put all of these in the schema with default values so they are configurable on the component.

src/components/virtual-gamepad-controls.js
```diff
@@ -45,112 +42,41 @@ AFRAME.registerComponent("virtual-gamepad-controls", {
...
-    this.cameraEl = document.querySelector("[camera]");
-
     this.onEnterVr = this.onEnterVr.bind(this);
     this.onExitVr = this.onExitVr.bind(this);
     this.el.sceneEl.addEventListener("enter-vr", this.onEnterVr);
     this.el.sceneEl.addEventListener("exit-vr", this.onExitVr);
   },
 
-  onJoystickChanged(event, joystick) {
+  onJoystickChanged(event, data) {
```
robertlong: I'd rename data back to joystick. `data` is easy to mistake with `this.data`.

src/components/character-controller.js
johnshaughnessy: Sure, will use THREE.Math.DEG2RAD instead.

src/components/character-controller.js
johnshaughnessy: Not true anymore. Thanks for catching.

src/components/character-controller.js
johnshaughnessy: 👍

PR #11, opened 2017-11-14
Add a room list

netpro2k:
Since Janus has no concept of public/private rooms nor do rooms exist when people are not in them just hardcoded a list of 5 "public" rooms and created a quick lobby page to display them. Art still needs to be updated, likely with an image of the environment we end up using. Also made up a fake "cap" for now just to see what it would look like UX wise, probably will remove.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -26,37 +26,41 @@ import Config from "./config";
 registerNetworkScheams();
 registerInputMappings();
 
-window.onSceneLoad = function() {
-  const qs = queryString.parse(location.search);
-  const scene = document.querySelector("a-scene");
+window.App = {
+  onSceneLoad() {
+    const qs = queryString.parse(location.search);
+    const scene = document.querySelector("a-scene");
 
-  scene.setAttribute("networked-scene", {
-    room:
-      qs.room && !isNaN(parseInt(qs.room))
-        ? parseInt(qs.room)
-        : Config.default_room,
-    serverURL: Config.janus_server_url
-  });
+    scene.setAttribute("networked-scene", {
+      room:
+        qs.room && !isNaN(parseInt(qs.room))
+          ? parseInt(qs.room)
+          : Config.default_room,
+      serverURL: Config.janus_server_url
+    });
 
-  if (!qs.stats || !/off|false|0/.test(qs.stats)) {
-    scene.setAttribute("stats", true);
+    if (!qs.stats || !/off|false|0/.test(qs.stats)) {
+      scene.setAttribute("stats", true);
+    }
+
+    if (AFRAME.utils.device.isMobile() || qs.gamepad) {
+      const playerRig = document.querySelector("#player-rig");
+      playerRig.setAttribute("virtual-gamepad-controls", {});
+    }
+
+    let username = qs.name;
+    if (!username) {
+      username = promptForName(username); // promptForName is blocking
+    }
+    const myNametag = document.querySelector("#player-rig .nametag");
+    myNametag.setAttribute("text", "value", username);
+
+    document.body.addEventListener("connected", App.onConnect);
+
+    scene.components["networked-scene"].connect();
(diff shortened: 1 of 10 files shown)


src/index.js
```diff
@@ -51,3 +56,7 @@ window.onSceneLoad = function() {
 
   scene.components["networked-scene"].connect();
 };
+
+window.onConnect = function() {
+  document.getElementById("loader").style.display = "none";
+};
```
brianpeiris: Maybe create a global `App` namespace instead of continuing to pollute `window`?
```
window.App = {
    onSceneLoad: ...
    onConnect: ...
};
```

src/lobby.js
```diff
@@ -0,0 +1,126 @@
...
+  onWebsocketOpen() {
+    this.session
+      .create()
+      .then(() => {
+        this.handle = new mj.JanusPluginHandle(this.session);
+        return this.handle.attach("janus.plugin.sfu").then(this.updateRooms);
+      })
+      .then(() => {
+        this.pollRoomsIntervalHandle = setInterval(this.updateRooms, 10000);
+      });
```
brianpeiris: Maybe fine for now but we should probably just adopt `async`/`await` syntax instead of `.then(...)`. It makes for much nicer code.

src/lobby.js
```diff
@@ -0,0 +1,126 @@
...
+    return Promise.all(
+      Config.public_rooms.map(room_id => {
+        return this.handle
+          .sendMessage({
+            kind: "listusers",
+            room_id
+          })
+          .then(signal => ({
+            id: room_id,
+            limit: 12,
```
brianpeiris: Is this limit just for show right now?

src/lobby.js
```diff
@@ -0,0 +1,126 @@
...
+    return (
+      <div className="overlay">
+        <RoomList rooms={this.state.rooms} />
+      </div>
+    );
+  }
+}
+
+const goToRoom = roomId => {
+  window.open(`room.html?room=${roomId}`, "_blank");
```
brianpeiris: Hmm. I wouldn't expect it to open a new window. Why not just navigate the current tab?

src/lobby.js
```diff
@@ -0,0 +1,126 @@
...
+    publicRooms +
+    Math.floor(Math.random() * (Number.MAX_SAFE_INTEGER - publicRooms));
+  goToRoom(roomId);
+};
+
+const RoomListItem = ({ room }) => {
+  return (
+    <li
+      className="room-item mdl-list__item mdl-list__item--two-line"
+      onClick={() => goToRoom(room.id)}
```
brianpeiris: This should just be a link, not a clickable `li`.

brianpeiris: Should also probably use a material design button to indicate clickability. Maybe a round button with an arrow or a "join" button.

src/lobby.js
netpro2k: Yeah agree, I didn't feel like figuring out how to set up the tooling for that. Robert already did it for the janus adapter so I figured I would just let him handle it :)

src/lobby.js
netpro2k: Yep, just imagining how that might work. May just remove it.

src/lobby.js
netpro2k: Yeah I was having issues getting this to work with mdl so I just did it this way for the moment, can spend a bit longer getting it to work cuz I agree I would like to be able to do things like middle click the link.

src/lobby.js
netpro2k: Just what I thought to do. I don't feel strongly about it so I will change it.

src/lobby.js
robertlong: Babel is already set up to transform async/await with `babel-preset-env`. We can change it whenever.

PR #12, opened 2017-11-16
New Avatar Rig

robertlong:
+ New Avatar Model
+ body-controller component positions/rotates body model based on head translation/rotation
+ hand-controls2 component removes default hand model and animations (animations to be implemented by a separate component that responds to the gesture events)
- Remove rig-selector component
- Remove avatar-customization component
- Temporarily remove now deployment because it causes `yarn install` to hang.


src/components/body-controller.js
```diff
@@ -0,0 +1,48 @@
...
+    this.bodyPositionVector = new THREE.Vector3();
+  },
+
+  tick(time, dt) {
+    const object3D = this.el.object3D;
+    const cameraObject3D = this.data.camera.object3D;
+
+    // Set Rotation
+    const cameraQuat = cameraObject3D.quaternion;
+    this.targetAngle.set(0, cameraQuat.y, 0, cameraQuat.w);
```
johnshaughnessy: I have not seen someone set the fields of quaternion manually like this to get a desired rotation. How does this work?

PR #13, opened 2017-11-16
New janus SFU API

brianpeiris:


src/input-mappings.js
```diff
@@ -1,37 +1,39 @@
...
-        e_press: "action_snap_rotate_right",
-        w_down: "action_move_forward",
-        w_up: "action_dont_move_forward",
-        a_down: "action_move_left",
-        a_up: "action_dont_move_left",
-        s_down: "action_move_backward",
-        s_up: "action_dont_move_backward",
-        d_down: "action_move_right",
-        d_up: "action_dont_move_right"
+    mappings: {
```
brianpeiris: We were pointing to `master` input-mapping-component in package.json. This is the new API. I also fixed the package to a SHA above.

src/lobby.js
```diff
@@ -46,20 +46,12 @@ class Lobby extends React.Component {
...
-            limit: 12,
-            users: signal.plugindata.data.response.user_ids
-          }));
-      })
-    );
+    return this.handle
+      .sendMessage({ kind: "listusers" })
+      .then(signal => {
+        const usersByRoom = signal.plugindata.data.response.users;
+        return Config.public_rooms.map(id => ({id, limit: 12, users: usersByRoom[id] || []}));
```
brianpeiris: We could actually list all the occupied rooms with the new Janus API, but I assumed we want to keep the lobby list as it is (with rooms 1-5 pre-defined). This effectively makes all other rooms pseudo-private, or at least unlisted from the UI.

package.json
```diff
@@ -9,14 +9,12 @@
...
-    "prettier"
-  ],
   "dependencies": {
     "aframe": "0.7.0",
-    "aframe-input-mapping-component": "https://github.com/fernandojsg/aframe-input-mapping-component",
+    "aframe-input-mapping-component": "https://github.com/fernandojsg/aframe-input-mapping-component#6ebc38f",
     "aframe-teleport-controls": "https://github.com/netpro2k/aframe-teleport-controls#feature/teleport-origin",
-    "naf-janus-adapter": "^0.1.4",
+    "minijanus": "^0.1.6",
+    "naf-janus-adapter": "https://github.com/mozilla/naf-janus-adapter#new-janus-api",
```
brianpeiris: This should point to a new npm package version when we publish it (mozilla/naf-janus-adapter#6)

src/lobby.js
robertlong: Yeah, I think that's good enough for now.

package.json
robertlong: Updated

src/lobby.js
netpro2k: Yeah, the pseudo privacy is why I did it this way. Doesn't seem worth doing this in Janus as we probably want to handle all room listing and such in reticulum eventually, so this is good enough for now.

PR #15, opened 2017-11-21
Fixes for Daydream

robertlong:
+ Fixes a bug that @johnshaughnessy and I found with input mapping events and the `virtual-gamepad-controls` component.
+ Adds initializing the `daydream-controls` component to `hand-controls2` component.
+ Uses a custom version of AFrame that sets the correct camera height when entering VR on mobile devices (See https://github.com/aframevr/aframe/pull/3111).
+ Adds an `offline=true` query string option so you can get past the loading screen when offline.
+ Fixes a typo with the avatar body mesh.

### DIFF
--- a/public/room.html
+++ b/public/room.html
@@ -2,6 +2,7 @@
 
 <head>
     <title>Mozilla Mixed Reality Social Client</title>
+    <script src="https://webrtc.github.io/adapter/adapter-latest.js"></script>
     <script src="./app.bundle.js"></script>
     <style>
         .a-enter-vr {
--- a/public/room.html
+++ b/public/room.html
@@ -2,7 +2,7 @@
 
 <head>
     <title>Mozilla Mixed Reality Social Client</title>
-    <script src="https://webrtc.github.io/adapter/adapter-latest.js"></script>
+    <script src="https://webrtc.github.io/adapter/adapter-6.0.2.js"></script>
     <script src="./app.bundle.js"></script>
     <style>
         .a-enter-vr {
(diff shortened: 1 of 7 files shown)


public/room.html
```diff
@@ -2,6 +2,7 @@
 
 <head>
     <title>Mozilla Mixed Reality Social Client</title>
+    <script src="https://webrtc.github.io/adapter/adapter-latest.js"></script>
```
brianpeiris: Should we really point to `latest`?

src/components/character-controller.js
```diff
@@ -116,15 +116,27 @@ AFRAME.registerComponent("character-controller", {
   },
 
   onTranslateX: function(event) {
-    this.accelerationInput.setX(event.detail);
+    if (typeof event.detail !== "object") {
+      this.accelerationInput.setX(event.detail);
+    } else {
+      this.accelerationInput.setX(0);
+    }
```
brianpeiris: Is this a bug in input-mapping that we should submit a PR for? Maybe put a TODO in here.

src/components/hand-controls2.js
```diff
@@ -106,6 +106,7 @@ AFRAME.registerComponent("hand-controls2", {
       el.setAttribute("vive-controls", controlConfiguration);
```
brianpeiris: (why is this component called `hand-controls2`? Surely it deserves a better name)

netpro2k: I think its a direct extension of the aframe built in component `hand-controls` . It's likely something we might want to land upstream but we don't know exactly what we want yet so its not worth maintaining a seperate fork of aframe over. Thus `hand-controls2` was born.

PR #22, opened 2017-11-28
Feature/auth for allhands

InfiniteLee:
if jwt is present in cookies, grab the user's name from it to pre-populate the name prompt field


src/utils.js
```diff
@@ -166,10 +166,24 @@ export function generateName() {
...
+  var value = "; " + document.cookie;
+  var parts = value.split("; " + name + "=");
+  if (parts.length == 2) return parts.pop().split(";").shift();
+}
+
+export function parseJwt (token) {
+  var base64Url = token.split('.')[1];
+  var base64 = base64Url.replace('-', '+').replace('_', '/');
+  return JSON.parse(window.atob(base64));
+}
```
brianpeiris: Oh. Is the token not signed and encrypted?

brianpeiris: If not, maybe the thing creating the cookie can just make it a json string to begin with so we don't have to do this little dance here.

netpro2k: JWT are signed, but not encrypted

PR #27, opened 2017-11-28
Update input mappings to use latest AFIM, with activators and behaviours

johnshaughnessy:


public/room.html
```diff
@@ -120,25 +122,41 @@
...
                 networked="template: #nametag-template; showLocalTemplate: false;"
             ></a-entity>
 
+                <!-- oculus-touch-controls-extended="
+                     dpad: true;
+                     dpad_deadzone: 0.8;
+                     dpad_livezone: 0.3;
+                     dpad_directions: 8;
+                     dpad_turbo: true;
+                     dpad_haptic_intensity: none" -->
```
johnshaughnessy: If this is active on #left-hand or #right-hand while you try to use vive controllers, then it will subscribe to "axismove" thinking it is the oculus touch controller's joystick. So I commented this component out.

johnshaughnessy: I think what I actually want to do is add this component at runtime iff an oculus touch controller is found (and likewise for the vive-controls-extended component), but I'm not yet sure how to do this.

public/room.html
```diff
@@ -120,25 +122,41 @@
...
+                    dpad_haptic_intensity: none;
+                    dpad_deadzone: 0.6;
+                    dpad_pressed_haptic_intensity: low"
+                oculus-touch-controls-extended="
+                    dpad: true;
+                    dpad_deadzone: 0.8;
+                    dpad_livezone: 0.3;
+                    dpad_directions: 8;
+                    dpad_turbo: true;
+                    dpad_haptic_intensity: none"
```
brianpeiris: Maybe most, if not all, of these settings should be the default, so there's less noise in the html?

public/room.html
```diff
@@ -112,6 +112,8 @@
             id="player-rig"
             networked
             character-controller="pivot: #head"
+            wasd-dpad
+            dpad-as-axes="inputName:dpad; name:keyboard_dpad_axes; emitter: #left-hand"
```
brianpeiris: Most of these settings are already defaults.

src/components/split-axis-events.js
```diff
@@ -1,30 +1,376 @@
...
+    const name = this.data.name;
+    const inputName = this.data.inputName;
+    const emitter = document.querySelector(this.data.emitter);
+    return function(event) {
+      emitter.emit(name, { axis: [axes[0], axes[1]] });
+    };
+  }
+});
+
+AFRAME.registerComponent("wasd-dpad", {
```
brianpeiris: Generally, I think we should have a limit of one component per source file, IMO.

src/components/split-axis-events.js
```diff
@@ -1,30 +1,376 @@
...
+  tick: function(t, dt) {
+    const { north, east, south, west } = this.data;
+    var direction = "";
+    direction += this.keys[north] ? "north" : "";
+    direction += this.keys[south] ? "south" : "";
+    direction += this.keys[east] ? "east" : "";
+    direction += this.keys[west] ? "west" : "";
+    if (direction !== "") {
+      this.el.emit(`dpad_${direction}`);
+    }
```
brianpeiris: I think it might be okay to `emit` on every tick, though arguably you can throttle this a considerably. Key repetition events are typically on the order of 400ms. 
Even still, I would try to avoid all this string manipulation and only re-construct strings when the keys actually change. Worth profiling first though.

src/components/character-controller.js
```diff
@@ -49,25 +36,7 @@ AFRAME.registerComponent("character-controller", {
...
-    eventSrc.addEventListener(
-      "action_dont_move_backward",
-      this.onDontMoveBackward
-    );
-    eventSrc.addEventListener("action_move_left", this.onMoveLeft);
-    eventSrc.addEventListener("action_dont_move_left", this.onDontMoveLeft);
-    eventSrc.addEventListener("action_move_right", this.onMoveRight);
-    eventSrc.addEventListener("action_dont_move_right", this.onDontMoveRight);
-
+    eventSrc.addEventListener("move", this.onMove);
```
brianpeiris: IMO it's better for handler functions to have descriptive names instead of simply repeating the event name. 
e.g. This could be `eventSrc.addEventListener("move", this.setAccelerationInput);` instead.

src/components/split-axis-events.js
```diff
@@ -1,30 +1,376 @@
...
+      strength = 0.2;
+      duration = 12;
+    }
+    if (intensity === "high") {
+      strength = 1;
+      duration = 12;
+    }
+    if (intensity === "none") {
+      return;
+    }
```
brianpeiris: A `switch` statement would make sense here.

src/components/split-axis-events.js
```diff
@@ -1,30 +1,376 @@
...
+      axis: [x, y]
+    });
+
+    var deadzone = this.data.dpad_deadzone;
+    var turbo = this.data.dpad_turbo;
+    var pressedTurbo = this.data.dpad_pressed_turbo;
+    var livezone = this.data.dpad_livezone;
+    var directions = this.data.dpad_directions;
+    var hapticIntensity = this.data.dpad_haptic_intensity;
+    var pressedHapticIntensity = this.data.dpad_pressed_haptic_intensity;
```
brianpeiris: Use `let` instead of `var`. Though it looks like you can use `const` for most of these.

src/components/split-axis-events.js
```diff
@@ -1,30 +1,376 @@
...
+
+    if (!this.trackpadPressed && !this.dpadCanFire) {
+      return;
+    }
+    if (this.trackpadPressed && !this.dpadPressedCanFire) {
+      return;
+    }
+    var dpadEvent = `${hand}_trackpad_dpad${this.trackpadPressed
+      ? "_pressed"
+      : ""}_${direction}`;
```
brianpeiris: Extracting the ternary expression into a separate variable would be more readable.

public/room.html
johnshaughnessy: They're both enabled for now.

src/utils.js
```diff
@@ -173,3 +173,47 @@ export function promptForName() {
   } while (!(username && username.length));
   return username;
 }
+
+export function angleToDirection(angle) {
```
netpro2k: We should probably just ditch this in favor of the `angleTo4Direction`, it makes for cleaner event names anyway (avoids the "downdown").

Also I know I wrote the original function but I can't help but feel there is a much cleaner way to do this hehe.

src/components/oculus-touch-controls-extended.js
```diff
@@ -0,0 +1,62 @@
...
+    const x = event.detail.axis[0];
+    const y = event.detail.axis[1];
+    const deadzone = this.data.dpad_deadzone;
+    const turbo = this.data.dpad_turbo;
+    const livezone = this.data.dpad_livezone;
+    const directions = this.data.dpad_directions;
+    const haptic_intensity = this.data.dpad_haptic_intensity;
+    const hand = this.data.hand;
+
+    event.target.emit(`${hand}_axismove`, {
```
netpro2k: The handedness can go away as this can now be handled in afim, unclear why we have to flip the y axis here.

src/input-mappings.js
```diff
@@ -1,47 +1,94 @@
...
-          w_up: "action_dont_move_forward",
-          a_down: "action_move_left",
-          a_up: "action_dont_move_left",
-          s_down: "action_move_backward",
-          s_up: "action_dont_move_backward",
-          d_down: "action_move_right",
-          d_up: "action_dont_move_right"
-        }
+        xbuttondown: "action_mute",
+        gripdown: "middle_ring_pinky_down",
```
netpro2k: I would either get rid of these grip actions (since they are not used) or add them to the inGameActions map above

src/input-mappings.js
```diff
@@ -1,47 +1,94 @@
...
+      },
+      keyboard: {
+        m_press: "action_mute",
+        q_press: "snap_rotate_left",
+        e_press: "snap_rotate_right",
+        v_press: "action_share_screen",
+
+        // We can't create a keyboard behaviour with AFIM yet,
+        // so these will get captured by wasd-to-analog2d
+        w_down: "w_down",
```
netpro2k: So these are certainly ugly, but thats good as it points out a clear hole in AFIM that needs to get addressed.

src/components/haptic-feedback.js
```diff
@@ -0,0 +1,57 @@
...
+  init: function() {
+    this.pulse = this.pulse.bind(this);
+    this.getActuator = this.getActuator.bind(this);
+    this.getActuator().then(actuator => {
+      this.actuator = actuator;
+    });
+  },
+
+  getActuator() {
+    return new Promise((resolve, reject) => {
```
netpro2k: Still really not a fan of this but if we cant just listen to `controllerconnected` for some reason I guess this will have to do...

src/activators/reversey.js
```diff
@@ -0,0 +1,20 @@
+function ReverseY(el, button, onActivate) {
```
netpro2k: Interesting use of an activator.. Not quite changing how the action activates but I guess it makes sense. It's definitely good to experiment with this so we can flesh out what kinds of features we want in AFIM

src/activators/shortpress.js
```diff
@@ -0,0 +1,39 @@
+function ShortPress(el, button, onActivate) {
```
netpro2k: Some might even call this a click 😆

src/behaviours/oculus-touch-joystick-dpad4.js
```diff
@@ -0,0 +1,31 @@
...
+  this.hapticIntensity = "low";
+  this.emitDPad4 = this.emitDPad4.bind(this);
+  el.addEventListener("axismove", this.emitDPad4);
+}
+
+oculus_touch_joystick_dpad4.prototype = {
+  emitDPad4: function(event) {
+    const x = event.detail.axis[0];
+    const y = event.detail.axis[1];
+    const inCenter =
```
netpro2k: This should be checking `x*x + y*y < centerRadiusSquared`, its currently a center square instead of a circle. centerRadiusSquared should be computed once in the constructor

src/behaviours/vive-trackpad-dpad4.js
```diff
@@ -0,0 +1,57 @@
...
+  press: function(_) {
+    this.pressed = true;
+  },
+  unpress: function(_) {
+    this.pressed = false;
+  },
+  emitDPad4: function(event) {
+    const x = event.detail.axis[0];
+    const y = event.detail.axis[1];
+    const inCenter =
```
netpro2k: same as above

src/components/character-controller.js
```diff
@@ -298,9 +168,13 @@ AFRAME.registerComponent("character-controller", {
       velocity.z = 0;
     }
 
-    const dvx = data.groundAcc * dt * this.accelerationInput.x * this.boost;
-    const dvz = data.groundAcc * dt * -this.accelerationInput.z * this.boost;
+    const dvx = data.groundAcc * dt * this.accelerationInput.x;
```
netpro2k: Did you intend to get rid of the boost multiplier here?

src/components/wasd-to-analog2d.js
```diff
@@ -0,0 +1,100 @@
+AFRAME.registerComponent("wasd-to-analog2d", {
```
netpro2k: This is pretty ugly but its just pointing out a hole in AFIM, so this is good for now.

src/input-mappings.js
```diff
@@ -1,47 +1,94 @@
...
+    default: {
+      "oculus-touch-controls": {
+        joystick: "oculus_touch_joystick_dpad4"
+      },
+      "vive-controls": {
+        trackpad: "vive_trackpad_dpad4"
+      }
+    }
+  },
+  mappings: {
```
netpro2k: Most of the mappigns turn out pretty nice with the behaviors and activators, afim is coming along

src/room.js
```diff
@@ -19,19 +25,28 @@ import "./components/virtual-gamepad-controls";
...
-import registerNetworkScheams from "./network-schemas";
-import registerInputMappings from "./input-mappings";
 import { promptForName, getCookie, parseJwt } from "./utils";
-
-registerNetworkScheams();
-registerInputMappings();
+import registerNetworkSchemas from "./network-schemas";
+import { inGameActions, config } from "./input-mappings";
+
+AFRAME.registerInputBehaviour("vive_trackpad_dpad4", vive_trackpad_dpad4);
```
netpro2k: There is now enough stuff here that I'd maybe move all this into input-mappings.js and export a `registerInputMappigns` function that sets up all this stuff, similar to what we do for network schemas.

templates/room.hbs
```diff
@@ -130,25 +131,24 @@
...
                 id="nametag"
                 networked="template: #nametag-template; showLocalTemplate: false;"
             ></a-entity>
 
             <a-entity
                 id="left-hand"
-                split-axis-events
                 hand-controls2="left"
-                axis-dpad="centerZone: 1"
+                tracked-controls
```
netpro2k: tracked-controls gets added by hand-controls (via the platform specific controller components), I dont think you need this here.

src/activators/shortpress.js
johnshaughnessy: LOL good point...

PR #38, opened 2017-12-04
Add HandlebarsTemplatePlugin for rewriting asset urls.

robertlong:
I wrote HandlebarsTemplatePlugin, because [html-webpack-plugin](https://github.com/jantimon/html-webpack-plugin) appends the javascript bundle script tag to the html body and doesn't let us rewrite the url. I also couldn't find an existing solution that would let us do that.

The plugin is pretty basic. It compiles all the handlebars templates in the specified templates path and recompiles them on change when webpack is in watch mode. Helpers can be registered in the plugin options. The compiled templates are saved to the output directory.

I added the asset handlebars helper which is documented in the readme.

I also updated the build instructions and added `npm start` which is an alias for `npm run dev`.

### DIFF
--- a/webpack.common.js
+++ b/webpack.common.js
@@ -1,4 +1,6 @@
 const path = require("path");
+const HandlebarsTemplatePlugin = require("./templates/HandlebarsTemplatePlugin");
+const Handlebars = require("handlebars");
 
 module.exports = {
   entry: {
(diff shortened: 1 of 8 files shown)


webpack.common.js
```diff
@@ -22,5 +24,28 @@ module.exports = {
...
+      templatesPath: path.resolve(__dirname, "templates"),
+      helpers: {
+        /**
+         * Register a handlebars helper that prepends the base asset path.
+         * Useful for things like placing assets on a CDN and cache busting.
+         * Example:
+         * input: <img src="{{asset "asset.png"}}"/>
+         * output: <img src="https://cdn.mysite.com/asset.png?c="/>
+         */
+        asset: assetPath => {
```
netpro2k: Just `asset` feels like a weird name for this. Maybe `base_asset_path` to patch the envar?

PR #39, opened 2017-12-06
Webpack Refactor

robertlong:
This PR adds another generated html file that points to separate assets on the smoke test server.

It also adds a Chrome origin trial meta tag for enabling WebVR and Gamepad extensions on Chrome 62+.

Cdn url cache busting now uses a CRC hash so that builds only require clients to download updated files.

Moved vendor assets and webpack manifest to separate webpack chunks for long term caching.

Switched Javascript minification from babel minify to uglify with parallel workers to speed up build times.

Define `process.env.NODE_ENV = "production"` in the webpack bundle for React.

Bundled CSS is extracted and minified in production.


webpack.prod.js
```diff
@@ -1,8 +1,55 @@
...
+        },
+        {
+          fileName: "room.html",
+          templatePath: path.resolve(__dirname, "templates", "room.hbs"),
+          data: {
+            baseAssetsPath: process.env.BASE_ASSETS_PATH,
+            config
+          }
+        },
+        {
```
gfodor: note this should be `smoke-`, and I would probably suggest doing something more generic that will take a list of page names (`["index", "room"]`) and generically generate templates for `x.html` and `x-smoke.html`

config.js
```diff
@@ -0,0 +1,16 @@
+module.exports = {
+  // This origin trial token is used to enable WebVR and Gamepad Extensions on Chrome 62+
+  // You can find more information about getting your own origin trial token here: https://github.com/GoogleChrome/OriginTrials/blob/gh-pages/developer-guide.md
+  originTrialToken:
+    "AvIMoF4hyRZQVfSfksoqP+7qzwa4FSBzHRHvUyzC8rMATJVRbcOiLewBxbXtJVyV3N62gsZv7PoSNtDqqtjzYAcAAABkeyJvcmlnaW4iOiJodHRwczovL3JldGljdWx1bS5pbzo0NDMiLCJmZWF0dXJlIjoiV2ViVlIxLjFNNjIiLCJleHBpcnkiOjE1MTYxNDYyMDQsImlzU3ViZG9tYWluIjp0cnVlfQ==",
+  originTrialExpires: "2018-01-16",
+
+  // These variables are availible on the window.GLOBAL object
+  // The exported object must be a valid argument to JSON.stringify()
+  global: {
```
netpro2k: Why do this global thing instead of just having the webpack config import config.js from its previous location?

robertlong: Good point. I was thinking that all the per build config should be in one place, but now that I'm looking at it I think a bigger reason is `process.env.JANUS_SERVER` isn't available in the browser context so I'd assume we'd want to keep it out of the `/src` directory.

I could move everything but the janus server back to the original config file though. Would that be better?

config.js
```diff
@@ -0,0 +1,16 @@
+module.exports = {
+  // This origin trial token is used to enable WebVR and Gamepad Extensions on Chrome 62+
+  // You can find more information about getting your own origin trial token here: https://github.com/GoogleChrome/OriginTrials/blob/gh-pages/developer-guide.md
+  originTrialToken:
+    "AvIMoF4hyRZQVfSfksoqP+7qzwa4FSBzHRHvUyzC8rMATJVRbcOiLewBxbXtJVyV3N62gsZv7PoSNtDqqtjzYAcAAABkeyJvcmlnaW4iOiJodHRwczovL3JldGljdWx1bS5pbzo0NDMiLCJmZWF0dXJlIjoiV2ViVlIxLjFNNjIiLCJleHBpcnkiOjE1MTYxNDYyMDQsImlzU3ViZG9tYWluIjp0cnVlfQ==",
+  originTrialExpires: "2018-01-16",
+
+  // These variables are availible on the window.GLOBAL object
```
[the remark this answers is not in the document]
robertlong: Yup, thanks

.gitignore
```diff
@@ -57,8 +57,8 @@ typings/
 # dotenv environment variables file
 .env
 
-# webpack bundle
-public/*.bundle.js*
-public/*.html
+# Ignore everything in the root of the public folder
```
netpro2k: 🐾 🔫 Um, surely there is a better way to do this where we wont be shooting ourselves in the foot later? Why not have assets also be copped from a src directory or have all the output go to `dist` and copy over anything in public?

PR #41, opened 2017-12-06
update spawn positions

InfiniteLee:
users will now randomly spawn around the table facing towards the cemter (may vary depending on headset configuration)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -27,7 +27,7 @@ import "./systems/personal-space-bubble";
 
 import registerNetworkScheams from "./network-schemas";
 import registerInputMappings from "./input-mappings";
-import { promptForName, getCookie, parseJwt } from "./utils";
+import { promptForName, getCookie, parseJwt, getSpawnPositionInCircle, getRotationToTarget } from "./utils";
 import Config from "./config";
 
 registerNetworkScheams();
@@ -93,6 +93,15 @@ window.App = {
 
     const myNametag = document.querySelector("#player-rig .nametag");
     myNametag.setAttribute("text", "value", username);
+
+    //change spawn position & rotation
+    const myHead = document.querySelector("#player-rig #head");
+    let rotation = Math.random() * 2 * Math.PI;
+    let position = myHead.getAttribute("position");
+    let newPosition = getSpawnPositionInCircle(0, 0, 4, rotation);
+    newPosition.y = position.y;
+    myHead.setAttribute("position", newPosition);
+    myHead.setAttribute("rotation", getRotationToTarget(newPosition, {x:0, y:0, z:0}));
 
     scene.addEventListener("action_share_screen", shareScreen);
 
--- a/src/index.js
+++ b/src/index.js
@@ -23,11 +23,12 @@ import "./components/split-axis-events";
 import "./components/networked-video-player";
 import "./components/offset-relative-to";
 import "./components/cached-gltf-model";
+import "./components/spawn-controller";
 import "./systems/personal-space-bubble";
 
 import registerNetworkScheams from "./network-schemas";
 import registerInputMappings from "./input-mappings";
-import { promptForName, getCookie, parseJwt, getSpawnPositionInCircle, getRotationToTarget } from "./utils";
+import { promptForName, getCookie, parseJwt } from "./utils";
 import Config from "./config";
 
 registerNetworkScheams();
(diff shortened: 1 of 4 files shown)


src/index.js
```diff
@@ -94,6 +94,15 @@ window.App = {
...
     myNametag.setAttribute("text", "value", username);
 
+    //change spawn position & rotation
+    const myHead = document.querySelector("#player-rig #head");
+    let rotation = Math.random() * 2 * Math.PI;
+    let position = myHead.getAttribute("position");
+    let newPosition = getSpawnPositionInCircle(0, 0, 4, rotation);
+    newPosition.y = position.y;
+    myHead.setAttribute("position", newPosition);
+    myHead.setAttribute("rotation", getRotationToTarget(newPosition, {x:0, y:0, z:0}));
```
brianpeiris: This logic should really be in a component on the `head` element.

src/utils.js
```diff
@@ -187,3 +187,24 @@ export function parseJwt (token) {
...
+  eye = new THREE.Vector3(eye.x, eye.y, eye.z);
+  target = new THREE.Vector3(target.x, target.y, target.z);
+  m4.lookAt(eye, target, new THREE.Vector3(0,1,0));
+  let euler = new THREE.Euler();
+  euler.setFromRotationMatrix(m4, "YXZ");
+  return {x: 0, y:(toDegrees(euler.y) + 360) % 360, z:0};
+}
+
+export function toDegrees(angle) {
+  return angle * (180 / Math.PI);
```
netpro2k: Threejs has a constant for this `THREE.Math.RAD2DEG` so I would ditch this functin and just do the multiplication inline like we do here https://github.com/mozilla/mr-social-client/blob/b08841cdaaa800b0ff96b1f5ccd3b453314f886d/src/components/character-controller.js#L257-L261

src/utils.js
```diff
@@ -187,3 +187,24 @@ export function parseJwt (token) {
...
+}
```
brianpeiris: THREE already has a utility for this

src/index.js
```diff
@@ -94,6 +94,15 @@ window.App = {
     const myNametag = document.querySelector("#player-rig .nametag");
     myNametag.setAttribute("text", "value", username);
 
+    //change spawn position & rotation
+    const myHead = document.querySelector("#player-rig #head");
```
netpro2k: This stuff should definitely live inside a component that we just put on the player rig, see https://github.com/networked-aframe/networked-aframe/blob/master/server/static/js/spawn-in-circle.component.js for an example that I think pretty much does this same thing. We may want to just use that but going to comment on the rest of the PR anyway just for learning purposes.

src/utils.js
InfiniteLee: will replace

src/utils.js
```diff
@@ -187,3 +187,24 @@ export function parseJwt (token) {
...
+
+export function getSpawnPositionInCircle(originX, originZ, radius, rotation) {
+  let x = originX + radius * Math.cos(rotation);
+  let z = originZ + radius * Math.sin(rotation);
+
+  return {x: x, z: z};
+}
+
+export function getRotationToTarget(eye, target) {
+  let m4 = new THREE.Matrix4();
```
netpro2k: Since this function only gets called once it doesn't matter here, but just FYI if this was getting called more than once you would want to re-use the threejs primitives like: https://github.com/mozilla/mr-social-client/blob/b08841cdaaa800b0ff96b1f5ccd3b453314f886d/src/components/character-controller.js#L199-L211

src/index.js
netpro2k: Object3D has a lookAt funciton, I would just use that https://threejs.org/docs/#api/core/Object3D, though you will still have to call setAttribute to update the aframe side of things like we do here Threejs has a constant for this `THREE.Math.RAD2DEG` so I would ditch this functin and just do the multiplication inline like we do here https://github.com/mozilla/mr-social-client/blob/b08841cdaaa800b0ff96b1f5ccd3b453314f886d/src/components/character-controller.js#L257-L261 (its a bit silly because you end up doing RAD -> DEG -> RAD, but that's just a result of aframe deciding to use degrees instead of radians.

netpro2k: it should probably be on the player rig not the head, but we probably want it to end up placing the head above the spawn point we pick (like we do for teleporting)

templates/room.hbs
```diff
@@ -119,6 +119,7 @@
                 personal-space-bubble
                 look-controls
                 networked="template: #head-template; showLocalTemplate: false;" 
+                spawn-controller="radius: 4"
```
netpro2k: Hmm this should not be on the head, it should be on the rig, that does mean we will have to deal with moving the rig with respect to the head, like we do for teleporting.

netpro2k: Hmm this should not be on the head, it should be on the rig, that does mean we will have to deal with moving the rig with respect to the head, like we do for teleporting. Otherwise the rig is going to remain at the spawn point and also this position is going to be wiped out by the HMD

PR #49, opened 2018-01-18
Add mic feedback to watch and top-of-page indicators.

johnshaughnessy:
Note that the top-of-page mic state indicator is now always present.


src/room.js
```diff
@@ -24,7 +24,7 @@ import "./components/mute-mic";
 import "./components/audio-feedback";
 import "./components/nametag-transform";
 import "./components/bone-mute-state-indicator";
```
netpro2k: Either the audio feedback stuff should be in a different component or this component should get renamed to `bone-mic-state-indicator` to match what you did with the `2d-mute-state-indicator`

johnshaughnessy:
The HUD will replace this

PR #52, opened 2018-02-13
Basic AR support

brianpeiris:


package.json
```diff
@@ -10,14 +10,14 @@
     "prettier": "prettier --write src/**/*.js"
   },
   "dependencies": {
-    "aframe": "https://github.com/robertlong/aframe#fix-increasing-offset",
```
brianpeiris: aframe.js is now a vendor asset since aframe-xr expects it to be a global. We can switch it to the official aframe CDN when the next version is released.

brianpeiris: I forget exactly how it failed. In any case, I was okay with making it a global script, since we wanted to take advantage of the A-Frame CDN, for caching.

webpack.dev.js
```diff
@@ -2,12 +2,24 @@ const path = require("path");
...
 const WriteFilePlugin = require("write-file-webpack-plugin");
+const fs = require("fs");
 
 module.exports = merge(common, {
   devtool: "inline-source-map",
   devServer: {
     contentBase: path.resolve(__dirname, "public"),
-    disableHostCheck: true
+    disableHostCheck: true,
+    before: function (app) {
```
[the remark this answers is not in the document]
brianpeiris: Usually the HEAD requests fail silently, but it was super annoying with the WebARonARCore browser because each failed request would raise a dialog.

PR #53, opened 2018-02-14
Don't block connection on microphone, screen sharing permissions

mquander:
This takes advantage of new renegotiation support in naf-janus-adapter and in Janus to accomplish this. Basically, now we can call `setLocalMediaStream` at any time, so we just connect without any regard to whether we have a local media stream yet and then set the media stream if and when we obtain it.

(We still prompt for microphone permissions at startup, since we think more or less everyone wants to use their microphone. But we don't wait to connect until they're granted, and we don't prompt for screen sharing permissions until a user presses `v`.)


mquander:
Incidentally, this change exposes a kind of performance bug, which is that connecting doesn't happen concurrently with other unrelated scene loading tasks (e.g. loading GLTF models.)

PR #57, opened 2018-03-02
Webpack Refactor and React Setup

robertlong:
Fixes a number of issues we were having with our current Webpack setup and introduces scaffolding for our new onboarding flow.

Fixes:
- Generates a HTTPS cert if one does not already exist when running `yarn start` You should only need to accept the untrusted https cert dialog once per computer, every time you wipe your `/certs` directory, or every 365 days. Fixes #54 
- All environment configuration is stored in a `.env` file now. This is generated on `yarn install` if it doesn't already exist. This should enable us to more easily use environment variables for feature flags and environment specific configuration.
- Upgraded to Webpack 4. This comes with some speed improvements, automatically chunking bundles, the ability to include WASM modules, and [much more](https://medium.com/webpack/webpack-4-released-today-6cdb994702d4)
- Better asset cache busting. Our assets should use proper content hashes that are appended at build time.
- Replaced my rudimentary Handlebars templating plugin with `html-loader`,`html-webpack-plugin`, and a 14 line custom LodashTemplatePlugin. You no longer have to use the `{{asset "url"}}` helper. `` and `<a-asset-item src="" />` tags will have their src attributes required by webpack, piped through `file-loader` and the url will be rewritten with the proper content hash.
- Consolidated our multiple webpack configs into one config. Removed `webpack-merge`. This should make our webpack config considerably easier to understand.
- Added our git repository to package.json. This just seemed like a good idea.
- Added `react-router` and made a *basic* onboarding flow in `onboarding.js`. This page can be reached at `/onboarding.html`

Note:
If you are working on things in `room.hbs` or `lobby.hbs` you may run into some merge conflicts. Those files have been moved to `room.html` and `lobby.html`.


package.json
```diff
@@ -25,35 +28,31 @@
...
+    "html-webpack-plugin": "webpack-contrib/html-webpack-plugin",
+    "lodash": "^4.17.5",
     "prettier": "^1.7.0",
-    "style-loader": "^0.19.0",
-    "uglifyjs-webpack-plugin": "^1.1.2",
-    "webpack": "^3.6.0",
-    "webpack-dev-server": "^2.9.3",
-    "webpack-merge": "^4.1.0",
-    "write-file-webpack-plugin": "^4.2.0"
+    "rimraf": "^2.6.2",
```
gfodor: sigh

robertlong: Haha, what's crazy is how large this package is just to get proper support on Windows: https://github.com/isaacs/rimraf/blob/master/rimraf.js

brianpeiris:
Didn't get to review this in time, but it looks good to me anyway. Nice work!

PR #59, opened 2018-03-05
Mode switching HUD

netpro2k:
The main goal of this branch was to explore the idea of a modal HUD, namely automated ways of invoking the hud, presenting different UI in VR vs mouse/touch mode, and switching between different input modalities when the HUD was being displayed.

The UI itself (`2d-hud` and `in-world-hud`) are largely throw away, pending actual design process, so not a ton of time was spent refining them or making them particularly reusable. It's likely that when we want to do this "for real" we should think about a way to share code between the 2d and in world UI (maybe using react to render aframe components).

On touch devices or when using keyboard and mouse, the UI is presented as simple 2d DOM elements that can be clicked. When in VR, the in world HUD should automatically come into view whenever the user looks down. While the HUD is locked in place, the teleport controls are disabled and instead a laser pointer control is enabled to interact with the UI elements. The specific heuristics around when the HUD is shown, when it locks into place, and when input modes are actually changed definitely has room for tweaking, but the basic building blocks are here.

The mode switching itself also doesn't yet account for what happens if multiple things want to be managing what mode the app is in. The `hud-detector` assumes full control for deciding what mode the app is in based on where the user is looking. This will probably need to change in the future.


src/components/2d-hud.js
```diff
@@ -0,0 +1,73 @@
...
+    this.micEl.style["background-color"] = `rgb(${red * 255},240,240)`;
+  },
+
+  onStateToggled(e) {
+    if (e.detail !== "muted") return;
+    this.updateMuteState();
+  },
+
+  updateMuteState() {
+    const muted = this.el.sceneEl.is("muted");
```
brianpeiris: Hmm, would feel better if the "muted" state was on the player, no?

src/components/in-world-hud.js
```diff
@@ -0,0 +1,155 @@
...
+
+  play() {
+    this.mic.addEventListener("raycaster-intersected", this.onMicHover);
+    this.mic.addEventListener("raycaster-intersected-cleared", this.onMicHoverExit);
+
+    this.nametag.addEventListener("raycaster-intersected", this.onNametagHovered);
+    this.nametag.addEventListener("raycaster-intersected-cleared", this.onNametagUnhovered);
+
+    this.avatar.addEventListener("raycaster-intersected", this.onAvatarHovered);
+    this.avatar.addEventListener("raycaster-intersected-cleared", this.onAvatarUnhovered);
```
brianpeiris: Seems like we should create a button element/component that encapsulates haptic and hover states.

src/components/in-world-hud.js
```diff
@@ -0,0 +1,155 @@
...
+    this.onUsernameChanged = this.onUsernameChanged.bind(this);
+    scene.addEventListener("username-changed", this.onUsernameChanged);
+
+    this.onNametagHovered = () => {
+      this.nametag.setAttribute("color", "#00eeee");
+      this.data.haptic.emit("haptic_pulse", { intensity: "low" });
+    };
+    this.onNametagUnhovered = () => {
+      this.nametag.setAttribute("color", "white");
+    };
```
brianpeiris: A "down" state would be nice on these buttons.

src/room.html
```diff
@@ -100,15 +113,33 @@
...
             animated-robot-hands
         >
+
+            <a-entity
+                id="player-hud"
+                hud-controller="head: #player-camera;"
+                vr-mode-toggle-visibility
+                vr-mode-toggle-playing__hud-controller
+            >
+                <a-entity in-world-hud="haptic:#player-right-controller;raycaster:#player-right-controller;" rotation="-39.01000608674482 0 0">
```
brianpeiris: `-39.01000608674482` do we really need that precision?

src/systems/app-mode.js
```diff
@@ -0,0 +1,227 @@
+/* global AFRAME, console, setTimeout, clearTimeout */
+
+const AppModes = Object.freeze({ DEFAULT: "default", HUD: "hud" });
+
+/**
+ * Simple system for keeping track of a modal app state
+ */
+AFRAME.registerSystem("app-mode", {
+  init() {
+    console.log("init app mode system");
```
brianpeiris: remove console logs

src/components/2d-hud.js
```diff
@@ -0,0 +1,73 @@
...
+    this.micEl.classList.toggle(styles.muted, muted);
+  }
+});
```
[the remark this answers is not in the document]
netpro2k: Yep that's probably the case, we also would probably want to see if we can drive the in-world-hud that way also somehow.

src/components/in-world-hud.js
netpro2k: Yep, I avoided suggesting that because I am considering the entire 2d-hud and in-world-hud to be throw away.

src/room.html
netpro2k: :) I think this probably came from the inspector, will just round it off heh

src/components/2d-hud.js
netpro2k: It was on the scene just so that anything can emit `action_mute` events, but not sure that's still important. That said this state is probably going to live in react land soon which will be more similar to it being on the scene than on the player, so I may just keep it as is.

src/systems/app-mode.js
```diff
@@ -0,0 +1,227 @@
...
+AFRAME.registerSystem("app-mode", {
+  init() {
+    console.log("init app mode system");
+    this.setMode(AppModes.DEFAULT);
+  },
+
+  setMode(newMode) {
+    if (Object.values(AppModes).includes(newMode) && newMode !== this.mode) {
+      this.mode = newMode;
+      this.el.emit("appmode-change", { mode: this.mode });
```
brianpeiris: Maybe make the event name `"app-mode-change"` for consistency.

src/systems/app-mode.js
```diff
@@ -0,0 +1,227 @@
...
+    }
+  }
+});
+
+/**
+ * Toggle the isPlaying state of a component based on app mode
+ */
+AFRAME.registerComponent("app-mode-toggle-playing", {
```
brianpeiris: Can we move the components into `src/components/app-mode.js`?

src/systems/app-mode.js
```diff
@@ -0,0 +1,227 @@
...
+const TWOPI = Math.PI * 2;
+function deltaAngle(a, b) {
+  const p = Math.abs(b - a) % TWOPI;
+  return p > Math.PI ? TWOPI - p : p;
+}
+
+/**
+ * Positions the HUD and toggles app mode based on where the user is looking
+ */
+AFRAME.registerComponent("hud-controller", {
```
brianpeiris: This component probably deserves its own file.

src/systems/app-mode.js
```diff
@@ -0,0 +1,227 @@
...
+ */
+AFRAME.registerComponent("vr-mode-toggle-playing", {
+  multiple: true,
+  schema: {
+    invert: { type: "boolean", default: false }
+  },
+
+  init() {
+    this.updateComponentState = this.updateComponentState.bind(this);
+    this.updateComponentState();
```
brianpeiris: This should be in `play()`, since A-frame will play all the components when the scene has finished init'ing.

brianpeiris:
I know this is just a first pass at the HUD, but I want to leave my UX feedback while it's fresh:

- In-vr hud is at an odd height when sitting down. I have to bend my wrist pretty severely to reach the mute button
- Mute button doesn't have a "down" state, so its hard to tell if I actually pressed it. I have to move my cursor ray off the button to confirm.
- Hard-head-locked UI feels bad. Especially when it's near my periphery, the in-vr hud looks almost like a display glitch. A floating head follow would be nice.
- Would be nice if the in-vr hud reduced opacity down to 0 when it dips down below you. Would help with the above point as well.
- Lots of convergence conflict going on with the blue raycaster line. Conflicts with the wall/floor in the background, conflicts with part of the hud itself, also conflicts with the avatar hand.
- Stopping WASD/thumbstick movement feels pretty intrusive and unexpected, but I think I'd have to get a better feel for it during normal use.

src/components/in-world-hud.js
johnshaughnessy: ✔️

src/room.html
johnshaughnessy: ✔️

src/systems/app-mode.js
johnshaughnessy: ✔️

src/systems/app-mode.js
johnshaughnessy: ✔️

src/systems/app-mode.js
johnshaughnessy: ✔️

gfodor:
Now that the React UI has landed, the 2D parts of this probably need to be retrofitted to work with it. 

Things that probably should be done before the PR is merged:
- Convert CSS to BEM and refactor into SCSS (may be a no-op) and put in `assets/stylesheets`

After this PR is merged, a more significant PR should be written that:
- Reads the user's name and mic state from a unified dataflow (the name from the store, the mic level from the micLevel state on the root react view)
- The 2d hud should be rewritten to be a react component that is part of the UiRoot tree

gfodor:
Added a new issue here: https://github.com/mozilla/mr-social-client/issues/104

PR #62, opened 2018-03-08
Inflate components

robertlong:
Inflate components in `<a-gltf-entity>` from the glTF `node.extras` property.

Use [fbx-gltf-pipeline](https://github.com/MozillaReality/fbx-gltf-pipeline) to take a FBX file, convert it to .gltf, add component data from a JSON file, and then convert to .glb.

The THREE.GLTFLoader will then place the `node.extras` data on `object3D.userData` and it will be passed to a deserializer as `componentData`.

We then inflate components that are registered through:
```js
AFRAME.AGLTFEntity.registerComponent(componentKey, componentName, [deserializer]);
```

If no deserializer is specified the default deserializer is used:
```js
function defaultDeserializer(el, componentName, componentData) {
  if (AFRAME.components[componentName].multiple && Array.isArray(componentData)) {
    for (let i = 0; i < componentData.length; i++) {
      el.setAttribute(componentName + "__" + i, componentData[i]);
    }
  } else {
    el.setAttribute(componentName, componentData);
  }
}
```


robertlong:
I think we need a better example for this than our robot avatar with it's head spinning around. Anyone have any suggestions?

robertlong:
Maybe we can do the eye blinking animation and audio feedback using this method? That'd be a good start and make it so that avatars can have separate behavior.

robertlong:
Added the audio feedback to the avatar model and removed the `animation-mixer` component from the head for now. When the fixes for hand animation land, I should be able to fix eye blinking and add that component to the avatar model as well.

src/aentity-components.js
```diff
@@ -0,0 +1,3 @@
+export default function registerComponents() {
```
netpro2k: Title of this file should include "gtlf" somewhere, maybe `gltf-component-mappings.js` or `a-gltf-entity-components.js`.

src/assets/avatars/Bot_With_Components.json
```diff
@@ -0,0 +1,7 @@
+{
```
netpro2k: Feels a bit odd to me that these don't live with the asset... Though it is true that keeping things up to date between branches will be problematic, you end up copying the resulting glb anyway, so this file feels like it should live with the source asset.

src/elements/a-gltf-entity.js
```diff
@@ -83,6 +102,20 @@ const inflateEntities = function(classPrefix, parentEl, node) {
 
   el.setObject3D(node.type.toLowerCase(), node);
 
+  const entityComponents = node.userData;
```
netpro2k: We should put them in a sub-key rather than just putting it in the root of userData, since other things may also stick stuff in userData.

src/elements/a-gltf-entity.js
```diff
@@ -83,6 +102,20 @@ const inflateEntities = function(classPrefix, parentEl, node) {
...
+
+  if (entityComponents) {
+    for (const prop in entityComponents) {
+      if (entityComponents.hasOwnProperty(prop)) {
+        const { deserializer, componentName } = AFRAME.AGLTFEntity.components[prop];
+
+        if (deserializer) {
+          deserializer(el, componentName, entityComponents[prop]);
```
netpro2k: Not a very strong preference, but this usage makes me prefer the term "inflator' just because I expect a "deserializer" function to return a value.

src/aentity-components.js
netpro2k: Also this file should technically require `a-gltf-entity.js`

src/assets/avatars/Bot_With_Components.json
robertlong: > Though it is true that keeping things up to date between branches will be problematic

I still think you want to version the component data with the component. I don't think we want to tie that versioning to the assets repo.

PR #63, opened 2018-03-09
Progressively load assets on mobile/desktop

robertlong:
Introduces `AFRAME.quality` which can be set to `low` or `high`.

Introduces `<a-progressive-asset src="fallback.glb" low-src="low.glb" high-src="high.glb" >`.

`low-src` and `high-src` are loaded based on the quality setting. `src` is used if neither are set. `src` is also needed because `<a-assets>` [waits on assets](https://github.com/aframevr/aframe/blob/master/src/core/a-assets.js#L84) with the `src` attribute.

I've updated GLTFLoader to the most recent version from the THREE.js dev branch. This includes support for `KHR_materials_unlit`.

I've also updated all the assets to use the unlit extension.

The avatar also has a high quality model that uses the `THREE.StandardMaterial`.


robertlong:
I'd like to be able to uncomment those lines having to do with sRGB encoding before we merge this. I think that we need to update our assets to get that to work.

Otherwise I think this is good to go.

src/elements/a-gltf-entity.js
```diff
@@ -121,7 +121,18 @@ AFRAME.registerElement("a-gltf-entity", {
         // If the src attribute is a selector, get the url from the asset item.
         if (src.charAt(0) === "#") {
           const assetEl = document.getElementById(src.substring(1));
-          src = assetEl.getAttribute("src");
+
+          const fallbackSrc = assetEl.getAttribute("src");
+          const highSrc = assetEl.getAttribute("high-src");
+          const lowSrc = assetEl.getAttribute("low-src");
+
+          if (highSrc && AFRAME.quality === "high") {
```
brianpeiris: Are we planning to submit `a-gltf-entity` and `a-progressive-asset` to A-Frame core? It feels a bit presumptive to claim `AFRAME.quality` if we aren't. I'd prefer we use our own global instead, like `APP.quality` or even `SOCIAL.quality`.

src/elements/a-progressive-asset.js
```diff
@@ -0,0 +1,57 @@
...
+
+        let src = fallbackSrc;
+
+        if (AFRAME.quality === "high") {
+          src = highSrc;
+        } else if (AFRAME.quality === "low") {
+          src = lowSrc;
+        }
+
+        this.fileLoader.setResponseType(this.getAttribute("response-type") || inferResponseType(src));
```
brianpeiris: `inferResponseType` is undefined

src/elements/a-progressive-asset.js
```diff
@@ -0,0 +1,57 @@
...
+          src,
+          function handleOnLoad(response) {
+            self.data = response;
+            /*
+            Workaround for a Chrome bug. If another XHR is sent to the same url before the
+            previous one closes, the second request never finishes.
+            setTimeout finishes the first request and lets the logic triggered by load open
+            subsequent requests.
+            setTimeout can be removed once the fix for the bug below ships:
+            https://bugs.chromium.org/p/chromium/issues/detail?id=633696&q=component%3ABlink%3ENetwork%3EXHR%20&colspec=ID%20Pri%20M%20Stars%20ReleaseBlock%20Component%20Status%20Owner%20Summary%20OS%20Modified
```
brianpeiris: Hmm, this bug is marked "Fixed" in Chrome 53+

src/elements/a-gltf-entity.js
netpro2k: In the component inflation branch we introduce the `AFRAME.AGLTFEntity` namespace, seems like we should put this in there so that we can release `a-progressive-asset` and `a-gltf-entity` as its own  "enchained GLTF loading" package.

src/elements/a-progressive-asset.js
```diff
@@ -0,0 +1,57 @@
+AFRAME.registerElement("a-progressive-asset", {
+  prototype: Object.create(AFRAME.ANode.prototype, {
+    createdCallback: {
+      value() {
+        this.data = null;
+        this.isAssetItem = true;
+        this.fileLoader = document.querySelector("a-assets").fileLoader;
```
netpro2k: Hmm, I don't think there is anything special about a-asset's fileloader so we should maybe just create our own. If not, we should probably get it from this element's parent node (since its required to be in an a-assets block), as there could be multiple scenes in the document.

src/elements/a-progressive-asset.js
```diff
@@ -0,0 +1,57 @@
...
+          src = highSrc;
+        } else if (AFRAME.quality === "low") {
+          src = lowSrc;
+        }
+
+        this.fileLoader.setResponseType(this.getAttribute("response-type") || inferResponseType(src));
+        this.fileLoader.load(
+          src,
+          function handleOnLoad(response) {
+            self.data = response;
```
netpro2k: I realize this isn't a new issue, but since we now are controlling both the loader and the asset, we should probably skip calling `GLTFLoader.prototype.load` and instead just pass this data directly to `GLTFLoader.prototype.parse`. In practice it doesn't actually matter since its cached, but seems clearer what is actually going on.

src/room.html
```diff
@@ -149,6 +156,9 @@
             </a-gltf-entity>
         </a-entity>
 
+        <!-- Lights -->
+        <a-entity light="type: directional; color: #F9FFCE; intensity: 0.6" position="0.002 5.231 -15.3"></a-entity>
```
netpro2k: We probably want a way to only include this in "high" quality mode

src/elements/a-progressive-asset.js
robertlong: I extracted this code from: https://github.com/aframevr/aframe/blob/master/src/core/a-assets.js#L94

I've added the attribution. I think we should probably keep this element as close to the original source as possible and I'll update it as `a-asset-item` changes.

src/elements/a-progressive-asset.js
robertlong: Gonna go with the later and reuse the FileLoader from the parent.

src/elements/a-progressive-asset.js
robertlong: Hmm, do we want a-progressive-asset to be specific to gltf assets? We could load lower resolution textures on lower-end platforms.

src/room.html
robertlong: Hmm. How do we want to do that? More templates?

PR #64, opened 2018-03-12
Fix networked video component

brianpeiris:
Also tweaks the webpack config a bit to fix a cert issue and external access.
Also fixes the "culling" issue with an updated a-frame. (See https://github.com/aframevr/aframe/pull/3448)

### DIFF
--- a/src/components/networked-video-player.js
+++ b/src/components/networked-video-player.js
@@ -32,10 +32,12 @@ AFRAME.registerComponent("networked-video-player", {
 
     const v = document.createElement("video");
     v.id = `nvp-video-${ownerId}`;
+    // muted and autoplay so that more restrictive browsers (e.g. Safari on iOS) will actually play the video.
+    v.muted = true;
+    v.autoplay = true;
     v.classList.add(styles.video);
     v.srcObject = new MediaStream(stream.getVideoTracks()); // We only want the video track so make a new MediaStream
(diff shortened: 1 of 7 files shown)


src/components/networked-video-player.js
```diff
@@ -21,15 +19,13 @@ AFRAME.registerComponent("networked-video-player", {
...
     if (!networkedEl) {
-      throw new Error(
-        "Video player must be added on a node, or a child of a node, with the `networked` component."
-      );
+      throw new Error("Video player must be added on a node, or a child of a node, with the `networked` component.");
     }
 
     const ownerId = networkedEl.components.networked.data.owner;
-    const stream = await NAF.connection.adapter.getMediaStream(ownerId);
+    const stream = await NAF.connection.adapter.getMediaStream(ownerId, "video");
```
netpro2k: Hmm, I don't think there is a second argument to `getMediaStream`. https://github.com/mozilla/naf-janus-adapter/blob/master/src/index.js#L351

brianpeiris: Yup, added it for naf-janus-adapter here: https://github.com/mozilla/naf-janus-adapter/pull/29

Not sure if it's worth suggesting it for all networked-aframe adapters.

robertlong:
Updated naf-janus-adapter. LGTM 👍

PR #66, opened 2018-03-14
Feature/interactables

InfiniteLee:
Overview:
Adding in `aframe-physics-system`, `aframe-super-hands-component`, and `aframe-physics-extras` packages (currently pointed at forks for the latter two). This branch combines those with `networked-aframe` to allow users to interact with networked physics objects, or "interactables". 

New components (I tried to name them to infer what their dependencies were, e.g. 'super' implies use with `super-hands` and 'networked' for use with `networked-aframe`):

- `super-networked-interactable` - defines an object as an interactable with the ability for transferring of ownership and updating of the physics model (e.g. switching between dynamic and static).
- `super-cursor` - a `super-hand` component/3d mouse cursor, useful for physics interactions in desktop mode. Currently, pointer lock is enabled in the `look-controls` component to make this easier to use.
- `super-spawner` - a spawner for "interactable" objects manipulated by `super-hands`. Grabbing the spawner creates a clone the specified template.
- `networked-counter` - a generic tool for tracking the ownership of networked entities and setting limits to how many can be owned and how long they live for if not actively being manipulated.
- `event-repeater` - This is still a bit of a WIP. This was added to support the fact that the `super-hands` components can't be added directly to the avatar-rig anymore (the position for the collider is wrong, as the avatar-rig hands are now offset from the origin of the controllers). The `super-hands` components are now attached to the avatar template so that their own offset can be applied, but the events from the controllers still only come from the rig hands (the `hand-controls2` component, I believe). Unfortunately, `super-hands` only listens for those events on the entity it is attached to, which is where this component comes in. (For some reason, this seems to break the `stretchable` component, still looking into why).

Note: mobile support still needs work, but it is largely tied up with how input is handled for mobile in general.


src/room.js
```diff
@@ -156,13 +164,30 @@ async function onSceneLoad() {
...
+    const entity = document.createElement("a-entity");
+    entity.setAttribute("body", "mass: 1;");
+    entity.setAttribute("grabbable", "");
+    entity.setAttribute("stretchable", "");
+    if (e.target && e.target != sceneEl) {
+      entity.setAttribute("position", e.target.object3D.getWorldPosition());
+    } else {
+      entity.setAttribute("position", "0 3 0");
+    }
+    entity.setAttribute("networked", "template: #physics-cube;");
```
robertlong: Is string serialization needed here?

src/components/super-spawner.js
```diff
@@ -0,0 +1,40 @@
+AFRAME.registerComponent("super-spawner", {
+  schema: {
+    template: { default: "" },
+    spawn_position: { type: "vec3" }
```
robertlong: Multi-word AFrame component properties should use camel case instead of snake case.

src/components/super-networked-interactable.js
```diff
@@ -0,0 +1,53 @@
...
+    this.el.removeEventListener("grab-start", this.grabStartListener);
+    this.el.removeEventListener("ownership-lost", this.ownershipLostListener);
+  },
+
+  _onGrabStart: function(e) {
+    this.hand = e.detail.hand;
+    if (this.networkedEl && !NAF.utils.isMine(this.networkedEl)) {
+      if (NAF.utils.takeOwnership(this.networkedEl)) {
+        console.log("1");
+        this.el.setAttribute("body", `mass: ${this.data.mass};`);
```
robertlong: Do we need string serialization here?

src/components/super-cursor.js
```diff
@@ -0,0 +1,85 @@
...
+        this.currentDistance = intersections[0].distance;
+        this.currentDistanceMod = 0;
+      } else {
+        this.currentDistance = this.data.maxDistance;
+      }
+    }
+
+    if (this.isGrabbing || !isIntersecting) {
+      const head = this.el.object3D;
+      const origin = head.getWorldPosition();
```
robertlong: This allocates a new vector every frame. Can you cache these vectors? https://threejs.org/docs/index.html#api/core/Object3D.getWorldPosition

src/components/super-cursor.js
```diff
@@ -0,0 +1,85 @@
...
+      const direction = head.getWorldDirection();
+      const distance = Math.min(
+        Math.max(this.data.minDistance, this.currentDistance - this.currentDistanceMod),
+        this.data.maxDistance
+      );
+      this.currentDistanceMod = this.currentDistance - distance;
+      direction.multiplyScalar(-distance);
+      const point = new THREE.Vector3();
+      point.addVectors(origin, direction);
+      this.data.cursor.setAttribute("position", { x: point.x, y: point.y, z: point.z });
```
robertlong: Let's set the object3D position directly. Ex: `this.data.cursor.object3D.position.set()`.

src/components/super-cursor.js
```diff
@@ -0,0 +1,85 @@
...
+      const head = this.el.object3D;
+      const origin = head.getWorldPosition();
+      const direction = head.getWorldDirection();
+      const distance = Math.min(
+        Math.max(this.data.minDistance, this.currentDistance - this.currentDistanceMod),
+        this.data.maxDistance
+      );
+      this.currentDistanceMod = this.currentDistance - distance;
+      direction.multiplyScalar(-distance);
+      const point = new THREE.Vector3();
```
robertlong: Also allocates a vector every frame.

src/components/networked-counter.js
```diff
@@ -0,0 +1,114 @@
...
+    this._removeTimeout(id);
+    this._addTimeout(id);
+    this.queue[id].ts = Date.now();
+  },
+
+  _destroyOldest: function() {
+    if (this.count > this.data.max) {
+      let oldest = null,
+        ts = Number.MAX_VALUE;
+      Object.keys(this.queue).forEach(function(id) {
```
robertlong: Object.keys allocates memory. Use this method instead:
```
for (const prop in obj) {
  if (obj.hasOwnProperty(prop)) {
    console.log(`obj.${prop} = ${obj[prop]}`);
  } 
}
```

src/components/super-cursor.js
```diff
@@ -0,0 +1,85 @@
...
+
+  init: function() {
+    this.isGrabbing = false;
+    this.wasIntersecting = false;
+    this.currentDistance = this.data.maxDistance;
+    this.currentDistanceMod = 0;
+    this.enabled = true;
+    this.isGrabbing = false;
+
+    document.addEventListener("mousedown", e => {
```
robertlong: We need to clean up these event listeners in `.destroy()`

src/components/networked-counter.js
```diff
@@ -0,0 +1,114 @@
...
+    max: { default: 3 },
+    ttl: { default: 120 },
+    grab_event: { type: "string", default: "grab-start" },
+    release_event: { type: "string", default: "grab-end" }
+  },
+
+  init: function() {
+    this.count = 0;
+    this.queue = {};
+    this.timeouts = {};
```
robertlong: We should call clearTimeout on all of these timeouts in `.destroy()`

src/components/super-spawner.js
```diff
@@ -0,0 +1,40 @@
...
+  },
+
+  init: function() {
+    this.el.addEventListener("grab-start", e => {
```
robertlong: Let's clean up this event listener in `.destroy()`

src/components/super-spawner.js
```diff
@@ -0,0 +1,40 @@
...
+      entity.addEventListener("componentinitialized", e => {
+        if (e.detail.name == "grabbable") {
+          resolve();
+        }
+      });
+    });
+
+    const bodyloaded = new Promise((resolve, reject) => {
+      entity.addEventListener("body-loaded", e => {
+        resolve();
```
robertlong: Let's clean up these event listeners when resolving.

src/components/event-repeater.js
```diff
@@ -0,0 +1,15 @@
+AFRAME.registerComponent("event-repeater", {
+  schema: {
+    target: { type: "selector" },
+    events: { type: "array" }
+  },
+
+  init: function() {
+    const events = this.data.events;
+    for (let i = 0; i < events.length; i++) {
+      this.data.target.addEventListener(events[i], e => {
```
robertlong: Lets clean up these event listeners in `.destroy()`

src/room.js
```diff
@@ -156,13 +164,30 @@ async function onSceneLoad() {
 
     await shareMedia(true, sharingScreen);
   }
+
+  scene.addEventListener("action_spawn_cube", (e) => {
```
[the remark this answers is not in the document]
InfiniteLee: oops this is old and isnt used anymore. Got lost in the merge a while back.

src/components/event-repeater.js
```diff
@@ -0,0 +1,15 @@
+AFRAME.registerComponent("event-repeater", {
```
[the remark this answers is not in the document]
InfiniteLee: It does both: it listens for an event on another entity, then re-emits the event on the attached entity.

PR #67, opened 2018-03-15
Avatar Switching

netpro2k:
Add support for swapping source on `a-gltf-entity` and then use this for avatar switching.


src/components/player-info.js
```diff
@@ -0,0 +1,32 @@
...
+  },
+  pause() {
+    this.el.removeEventListener("model-loaded", this.applyProperties);
+  },
+  update(oldProps) {
+    this.applyProperties();
+  },
+  applyProperties() {
+    const nametagEl = this.el.querySelector(".nametag");
+    console.log("updating properties", this.data, nametagEl);
```
brianpeiris: remove console log

src/elements/a-gltf-entity.js
```diff
@@ -130,116 +135,165 @@ const inflateEntities = function(classPrefix, parentEl, node) {
...
+              selector: templateEl.getAttribute("data-selector"),
+              templateRoot: document.importNode(templateEl.content.firstElementChild, true)
+            })
+          );
+          setTimeout(resolve, 0);
+        });
+      }
+    },
+
+    setSrc: {
```
brianpeiris: This should probably be renamed, it does a lot more than simply set the src.

src/elements/a-gltf-entity.js
```diff
@@ -130,116 +135,165 @@ const inflateEntities = function(classPrefix, parentEl, node) {
...
+      value() {
+        return new Promise((resolve, reject) => {
+          this.templates = [];
+          this.querySelectorAll(":scope > template").forEach(templateEl =>
+            this.templates.push({
+              selector: templateEl.getAttribute("data-selector"),
+              templateRoot: document.importNode(templateEl.content.firstElementChild, true)
+            })
+          );
+          setTimeout(resolve, 0);
```
brianpeiris: There was a useful comment above this line in the previous version, I would retain it.

src/elements/a-gltf-entity.js
```diff
@@ -130,116 +135,165 @@ const inflateEntities = function(classPrefix, parentEl, node) {
...
-            finalizeLoad();
-          }
-        };
+          const model = await cachedLoadGLTF(src);
 
-        const inflate = (model, callback) => {
-          inflateEntities("", this, model);
-          this.querySelectorAll(":scope > template").forEach(attachTemplate);
+          // If we started loading something else already
+          // TODO: there should be a way to cancel loading instead
```
brianpeiris: This seems important. `GLTFLoader` ought to return an XHR that you can `.abort()`, just like `FileLoader` does.

src/elements/a-gltf-entity.js
```diff
@@ -130,116 +135,165 @@ const inflateEntities = function(classPrefix, parentEl, node) {
...
           }
 
-          this.model = gltfModel.scene || gltfModel.scenes[0];
-          this.model.animations = gltfModel.animations;
+          if (src === this.lastSrc) return;
+          this.lastSrc = src;
 
-          this.setObject3D("mesh", this.model);
-          this.emit("model-loaded", { format: "gltf", model: this.model });
+          if (!src) return;
```
brianpeiris: Seems like it's worth throwing/logging an error if we hit this case.

netpro2k: This is actually a valid case, it's how you would unload a model. That said it should actually be unloading the model, and I guess a warning couldnt hurt.

src/components/ik-controller.js
```diff
@@ -23,13 +23,16 @@ AFRAME.registerComponent("ik-root", {
       this.rightController = this.el.querySelector(this.data.rightController);
       updated = true;
```
brianpeiris: `updated` is unused.

src/elements/a-gltf-entity.js
```diff
@@ -96,9 +103,8 @@ const inflateEntities = function(classPrefix, parentEl, node) {
     z: node.scale.z
   });
 
-  node.position.set(0, 0, 0);
-  node.rotation.set(0, 0, 0);
-  node.scale.set(1, 1, 1);
+  node.matrixAutoUpdate = false;
+  node.matrix.identity();
```
brianpeiris: Probably worth adding a comment explaining why we're doing this.

PR #71, opened 2018-03-15
Feature/initial react ui

gfodor:
Adds unstyled initial entry flows for all platforms


src/storage/store.js
```diff
@@ -0,0 +1,77 @@
...
+    return this[STORE_STATE_CACHE_KEY];
+  }
+
+  subscribe(subscriber) {
+    this.subscribers.add(subscriber);
+  }
+
+  update(newState) {
+    if (newState.id) {
+      console.error("Store id is immutable.");
```
brianpeiris: Why not throw an error here?

src/storage/store.js
```diff
@@ -0,0 +1,77 @@
...
+      return;
+    }
+
+    const finalState = { ...this.state, ...newState };
+    const isValid = validator.validate(finalState, SCHEMA).valid;
+
+    if (!isValid) {
+      console.warn(`Write of ${JSON.stringify(finalState)} to store failed schema validation.`)
```
brianpeiris: Same here, throw an error.

src/storage/store.js
```diff
@@ -0,0 +1,77 @@
...
+    if (!this.hasOwnProperty(STORE_STATE_CACHE_KEY)) {
+      this[STORE_STATE_CACHE_KEY] = JSON.parse(localStorage.getItem(LOCAL_STORE_KEY));
+    }
+
+    return this[STORE_STATE_CACHE_KEY];
+  }
+
+  subscribe(subscriber) {
+    this.subscribers.add(subscriber);
+  }
```
brianpeiris: Couldn't we extend from EventTarget instead?

src/utils/vr-caps-detect.js
```diff
@@ -0,0 +1,81 @@
...
+  return navigator.userAgent.match(/\WAndroid\W/);
+}
+
+function isMaybeDaydreamCompatibleDevice() {
+  return navigator.userAgent.match(/\WAndroid\W/);
+}
+
+// Blacklist of VR device name regex matchers that we do not want to consider as valid VR devices
+// that can be entered into as a "generic" entry flow.
+const GENERIC_ENTRY_TYPE_DEVICE_BLACKLIST = [/cardboard/i];
```
brianpeiris: Why blacklist Cardboard? We may be cutting off a huge swath of people who currently enjoy WebVR content via Cardboard because they can't afford a legit VR device. I feel pretty strongly that we shouldn't limit access on those grounds.

src/storage/store.js
gfodor: yup didn't have that dependency at the time i wrote this

src/utils/vr-caps-detect.js
gfodor: the thinking was less about cutting off cardboard users and more about ensuring that people don't accidentally go into cardboard mode if they have a gearvr or daydream. i'll think about how to avoid that without inadvertently cutting off cardboard users.

brianpeiris: Ah, okay. In theory you won't have to worry about the Daydream case. As long as we keep our origin trial up to date, we can detect browser exposing the WebVR APIs. With GearVR, we'll have to rely on UA parsing unfortunately

gfodor: i'm torn on this one. for people who are in a WebVR capable browser (Chrome or Samsung internet today) that own legit VR hardware, we'll be presenting them in many cases with a choice in the menu that is fundamentally going to head them in the wrong direction, confuse them, and possibly damage the app's perception.

Concrete example: I own a GearVR and have the page open in Chrome. I do not have Daydream installed. I will be presented with two choices: "Open in Cardboard" (since the device exists) and "Open on GearVR" (since we always show this because we can't detect it) -- the cardboard choice will have a VR headset icon next to it so many of these users will erroneously choose that option. They won't know that they made a wrong choice and the experience of it just going into lens distortion mode and not using their headset properly will look like, at best, a bug, at worst, normal and alienate them from thinking the app is solid. 

i'm leaning towards leaving things as-is -- one of my additional assumptions is that for ~100% of the people who engage cardboard mode their device is not going to hit every frame anyway so this will just make them sick, and they should be using magic window mode. do we know if that's a bad assumption?

src/react-components/ui-root.js
```diff
@@ -0,0 +1,422 @@
...
+  <button {...props}>
+    Enter on GearVR
+  </button>
+);
+
+const DaydreamEntryButton = (props) => (
+  <button {...props}>
+    Enter on Daydream
+  </button>
+);
```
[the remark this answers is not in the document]
gfodor: yup I was thinking about how to work this in. I think "Using Google Chrome" makes a lot of sense since it jumps you to another browser on-screen and it's something users will understand. I don't think that we need to convey that they are going into Oculus Browser though, because conceptually they are probably unlikly to even know what Oculus Browser is, and since the next step after clicking is to put on their headset, they are unlikely to react to this as an "unexpected redirection to another browser" as they'd reasonably be expected to in the Chrome case.

gfodor:
Oh and the reason adding "Using Oculus Browser" has a downside if we can
avoid it is because many people may not recognize that product and think
they may not have it installed, even though it comes with Home. So they may
decide to go try to find it in the Oculus store, etc, before trying to get
in (and they won't find it obviously, so may end up even more confused)

src/utils/vr-caps-detect.js
brianpeiris: > ~100% of the people who engage cardboard mode their device is not going to hit every frame 

I think this is a bad assumption. There are many high-end phones out there that happen to not support Daydream or GearVR.

Furthermore, people who will opt to use Cardboard mode are likely to know what experience they're getting into. If they use a Cardboard regularly, they either have a good enough phone or they are very tolerant to VR sickness.

I would rather provide the Cardboard option at the end of the list when we can't be certain of Daydream/GearVR support. We can even design to indicate that it's the less desirable option if you prefer (e.g. no VR icon, greyed out button). 

I'd be fine with removing the Cardboard option if we prove otherwise after a couple of months of metrics, but I strongly disagree with cutting off that audience to begin with.

gfodor: Ok I think de-emphasizing it somehow might be a good way to do it. We already have an "Enter on another device" prompt in the design (for standalones, etc) so perhaps it can eventually be behind that screen. For now I can put the cardboard option in the place of that text until we build out the whole "duck.link" flow.

brianpeiris:
We can close this now. I'll add the changes to Store in my avatar selector PR.

PR #72, opened 2018-03-16
screen sharing checkbox and toggling

brianpeiris:


src/react-components/ui-root.js
```diff
@@ -357,13 +357,21 @@ class UIRoot extends Component {
...
     ? (
       <div>
         <TwoDEntryButton onClick={this.enter2D}/>
-        { this.props.availableVREntryTypes.generic !== VR_DEVICE_AVAILABILITY.no && <GenericEntryButton onClick={this.enterVR}/> }
-        { this.props.availableVREntryTypes.gearvr !== VR_DEVICE_AVAILABILITY.no && <GearVREntryButton onClick={this.enterGearVR}/> }
-        { this.props.availableVREntryTypes.daydream !== VR_DEVICE_AVAILABILITY.no && <DaydreamEntryButton onClick={this.enterDaydream}/> }
+        { entryTypes.generic !== VR_DEVICE_AVAILABILITY.no && <GenericEntryButton onClick={this.enterVR}/> }
+        { entryTypes.gearvr !== VR_DEVICE_AVAILABILITY.no && <GearVREntryButton onClick={this.enterGearVR}/> }
+        { entryTypes.daydream !== VR_DEVICE_AVAILABILITY.no && <DaydreamEntryButton onClick={this.enterDaydream}/> }
+        <label>
```
gfodor: eventually we need to only show this on desktops, or maybe there's a better way to detect screen sharing ability

brianpeiris: Yeah, I wasn't sure if we want to have a "share your camera" fallback as well. We can do that on desktop as well as mobile.

src/react-components/ui-root.js
```diff
@@ -407,6 +422,14 @@ class UIRoot extends Component {
               subtitle={this.state.availableVREntryTypes.daydream == VR_DEVICE_AVAILABILITY.maybe ? daydreamMaybeSubtitle : "" }/> }
         { this.state.availableVREntryTypes.cardboard !== VR_DEVICE_AVAILABILITY.no &&
           (<div className="entry-panel__secondary" onClick={this.enterVR}><FormattedMessage id="entry.cardboard"/></div>) }
+        { !mobiledetect.mobile() && /firefox/i.test(navigator.userAgent) && (
```
gfodor: this firefox test surprised me -- perhaps extract it into a variable to explain why its needed?

gfodor:
Probably a separate PR, but I'd propose that to safely "disable" this until P2P is in, we'll want to reject the spawning/acceptance of video media streams from other peers unless a query string var is set.

brianpeiris:
:+1: added query params to enable and accept screen sharing

PR #75, opened 2018-03-17
Better 2D cursor

InfiniteLee:
no more cursor lock; super-cursor follows around cursor screen position; look-controls are paused when an interactable is grabbed; colors changed; cursor no longer appears to be in "hovered" state when moused over the floor or table; system cursor is hidden.


src/components/super-cursor.js
```diff
@@ -62,49 +86,70 @@ AFRAME.registerComponent("super-cursor", {
...
+      this.data.cursor.setAttribute("material", { color: this.data.cursorColorHovered });
+    } else if (!this.isGrabbing && !this.isInteractable && this.wasIntersecting) {
       this.wasIntersecting = false;
-      this.data.cursor.setAttribute("material", {color: "#00EFFF"});
+      this.data.cursor.setAttribute("material", { color: this.data.cursorColorUnhovered });
     }
   },
 
   _handleMouseDown: function(e) {
+    this.mouseDown = true;
```
netpro2k: This flag appears to be unused.

src/components/super-cursor.js
```diff
@@ -48,10 +62,20 @@ AFRAME.registerComponent("super-cursor", {
...
+    this.origin = raycaster.ray.origin;
+    this.direction = raycaster.ray.direction;
+    this.el.setAttribute("raycaster", { origin: this.origin, direction: this.direction });
+
+    let className = null;
+
     if (!this.isGrabbing) {
       const intersections = this.el.components.raycaster.intersections;
       if (intersections.length > 0 && intersections[0].distance <= this.data.maxDistance) {
         isIntersecting = true;
```
netpro2k: Instead of storing `isIntersecting`, `className`, `point` and `currentDistance` I would probably just store the interesection itself.

src/components/super-cursor.js
```diff
@@ -62,49 +86,70 @@ AFRAME.registerComponent("super-cursor", {
...
 
     if (this.isGrabbing || !isIntersecting) {
-      const head = this.el.object3D;
-      head.getWorldPosition(this.origin);
-      head.getWorldDirection(this.direction);
       const distance = Math.min(
         Math.max(this.data.minDistance, this.currentDistance - this.currentDistanceMod),
         this.data.maxDistance
       );
       this.currentDistanceMod = this.currentDistance - distance;
```
netpro2k: bit confused what this is doing here?

InfiniteLee: this is the logic that allows you to scroll the mouse wheel to move the object closer or farther from you.

PR #77, opened 2018-03-21
Fix issues with forced entry and images

gfodor:
Fixes:
- Sets w/h on entry button and level images so no re-layout before images load
- Fixes GearVR icon
- Fixes race condition to handle forced vr entry mode


brianpeiris:
LGTM

PR #106, opened 2018-03-26
Avatar Switching UI

### LINKED ISSUE #86: Avatar selection
Need to be able to pick an avatar in the UI.

brianpeiris:
- Added new avatar assets and included them in `room.html`
- Add an avatar selector to the profile entry panel
- The avatar selector is an embedded a-frame scene. It lives in inside an iframe, hosted by `avatar-selector.html`.
- Communication between the avatar selector and the panel are handled via a location hash on the iframe and the postMessage API.
- A carousel of avatars is generated from a new avatar manifest defined in `src/assets/avatars/avatars.json`

Closes #86


src/assets/stylesheets/avatar-selector.scss
```diff
@@ -0,0 +1,47 @@
...
+#selector-root {
+  height: 100%;
+}
+.avatar-selector {
+  overflow: hidden;
+  height: 100%;
+  display: flex;
+  justify-content: center;
+
+  &__prev-button:focus, &__next-button:focus {
```
gfodor: +1 gotta refactor my scss at some point

src/assets/stylesheets/avatar-selector.scss
```diff
@@ -0,0 +1,47 @@
...
+    outline: none;
+  }
+  &__prev-button:active, &__next-button:active {
+    color: grey;
+  }
+  &__prev-button, &__next-button {
+    position: absolute;
+    top: 50%;
+    margin-top: -4em;
+    appearance: none;
```
gfodor: TIL

gfodor:
Perhaps we should refer to this a "picker" or "chooser" for end-user? Or just drop calling it anything specific? (Ex "Loading avatar selector..." becomes "Loading..." or "Loading Avatars..." etc)

src/elements/a-gltf-entity.js
```diff
@@ -284,7 +287,6 @@ AFRAME.registerElement("a-gltf-entity", {
         if (attr === "src") {
           this.setSrc(newVal);
         }
-        AFRAME.AEntity.prototype.attributeChangedCallback.call(this, attr, oldVal, newVal);
```
gfodor: intentional? Not well versed in this code but this seems like a impactful change.

src/react-components/avatar-selector.js
```diff
@@ -0,0 +1,119 @@
...
+import { injectIntl, FormattedMessage } from 'react-intl';
+
+class AvatarSelector extends Component {
+  static propTypes = {
+    avatars: PropTypes.array,
+    avatar: PropTypes.string,
+    onChange: PropTypes.func,
+  }
+
+  getAvatarIndex(direction=0) {
```
gfodor: should this be a `= () => {`? looks like this function assumes consistent `this`

src/react-components/avatar-selector.js
```diff
@@ -0,0 +1,119 @@
...
+    const currAvatarIndex = this.props.avatars.findIndex(avatar => avatar.id === this.props.avatar);
+    const numAvatars = this.props.avatars.length;
+    return ((currAvatarIndex + direction) % numAvatars + numAvatars) % numAvatars;
+  }
+
+  nextAvatar = () => {
+    const newAvatarIndex = this.getAvatarIndex(1);
```
gfodor: I'd probably extract two oneliner methods `nextAvatarIndex` and `previousAvatarIndex`

src/react-components/avatar-selector.js
```diff
@@ -0,0 +1,119 @@
...
+    this.animation.handleMixinUpdate();
+    this.animation.start();
+  }
+
+  render () {
+    const avatarAssets = this.props.avatars.map(avatar => (
+      <a-progressive-asset
+        id={avatar.id}
+        key={avatar.id}
+        response-type="arraybuffer"
```
gfodor: robert is refactoring things so the semantics here are we have the "normal/default/pbr" asset, and the "unlit" asset, instead of low vs high

src/react-components/avatar-selector.js
```diff
@@ -0,0 +1,119 @@
...
+
+    return (
+      <div className="avatar-selector">
+      <span className="avatar-selector__loading">
+        <FormattedMessage id="profile.avatar-selector.loading"/>
+      </span>
+      <a-scene vr-mode-ui="enabled: false" ref={sce => this.scene = sce}>
+        <a-assets>
+          {avatarAssets}
+          <a-asset-item
```
gfodor: we probably should add a issue to eventually create a environment for this and also point this to a gltf bundle url instead of a local asset glb

src/react-components/avatar-selector.js
```diff
@@ -0,0 +1,119 @@
...
+          light="type: ambient; color: #FFF"
+        ></a-entity>
+        <a-gltf-entity
+          id="meeting-space"
+          src="#meeting-space1-mesh"
+          position="0 0 0"
+        ></a-gltf-entity>
+      </a-scene>
+      <button className="avatar-selector__prev-button" onClick={this.prevAvatar}>
+        <i className="avatar-selector__button-icon material-icons">keyboard_arrow_left</i>
```
gfodor: on my branch i'm ripping out the dependency on material design lite -- we can keep it if you think it's important but if this is the only coupling left probably worth just extracting these icons. (I'm not going to use it for the home landing page)

src/storage/store.js
```diff
@@ -17,6 +17,7 @@ export const SCHEMA = {
       additionalProperties: false,
       properties: {
         display_name: { type: "string", pattern: "^[A-Za-z0-9-]{3,32}$" },
+        avatar: { type: "string" },
```
gfodor: I would call this "avatar_url" -- or we may want to consider future proofing this by storing some other higher-level selection choice. (Probably URL is ok though once we get gltf bundles set up)

src/react-components/profile-entry-panel.js
```diff
@@ -13,49 +13,76 @@ class ProfileEntryPanel extends Component {
...
+    if (e.source !== this.avatarSelector.contentWindow) { return; }
+    this.setState({avatar: e.data.avatar});
+  }
+
   componentDidMount() {
     // stop propagation so that avatar doesn't move when wasd'ing during text input.
-    this.nameInput.addEventListener('keydown', e => e.stopPropagation());
-    this.nameInput.addEventListener('keypress', e => e.stopPropagation());
-    this.nameInput.addEventListener('keyup', e => e.stopPropagation());
+    this.nameInput.addEventListener('keydown', this.stopPropagation);
```
gfodor: On the fence if we need these. I believe since now a-frame has no input mappings until you join the room, it should work OK. however if we want to allow in-game identity changes then this should stay a-frame will be receiving input at that point.

src/react-components/profile-entry-panel.js
```diff
@@ -13,49 +13,76 @@ class ProfileEntryPanel extends Component {
...
+    };
     this.props.store.addEventListener("statechanged", this.storeUpdated);
   }
 
   storeUpdated = () => {
-    this.setState({name: this.props.store.state.profile.display_name});
+    this.setState({
+      display_name: this.props.store.state.profile.display_name,
+      avatar: this.props.store.state.profile.avatar,
+    });
```
gfodor: minor refactoring suggestion
```
{ avatar, display_name } = this.props.store.state;
this.setState({ avatar, display_name });
```

src/react-components/avatar-selector.js
brianpeiris: Sure, I guess it improves consistency.

src/react-components/avatar-selector.js
brianpeiris: @robertlong Should I avoid using low/high here?

src/react-components/avatar-selector.js
brianpeiris: Have you actually been extracting the icons out of MDL? How about using Font Awesome?

src/storage/store.js
brianpeiris: It's actually the id as defined in the `avatars.json` file I added, so I'll go with avatar_id. 
BTW, is there a reason you went with snake_case instead of camelCase for the schema properties?

src/react-components/profile-entry-panel.js
brianpeiris: Yeah, I'll try removing these.

src/react-components/avatar-selector.js
brianpeiris: Created #110

src/react-components/avatar-selector.js
gfodor: I've been drawing them myself! :) (except for the little person icon, which I grabbed from google's material design site iirc)

src/storage/store.js
gfodor: ah cool, makes sense. wrt snake case i just tend to write persistent data structure schemas that way (avro schema, db tables, etc) not sure if this is un-local-storagey :)

brianpeiris: It's un-javascripty and inconsistent with all our other naming, but I'm not that strongly against it.

src/react-components/avatar-selector.js
brianpeiris: Hmm, okay. I'd rather just use Font Awesome if you want to get rid of MDL. But I guess we're going to throw all of this design away and use bespoke assets before release?

src/elements/a-gltf-entity.js
```diff
@@ -212,10 +215,10 @@ AFRAME.registerElement("a-gltf-entity", {
       value() {
         return new Promise((resolve, reject) => {
           this.templates = [];
-          this.querySelectorAll(":scope > template").forEach(templateEl =>
+          this.querySelectorAll(":scope > template").forEach(templateEl => 
             this.templates.push({
               selector: templateEl.getAttribute("data-selector"),
-              templateRoot: document.importNode(templateEl.content.firstElementChild, true)
+              templateRoot: document.importNode(templateEl.firstElementChild || templateEl.content.firstElementChild, true)
```
netpro2k: Hmm is this some browser compatability fix?

src/elements/a-gltf-entity.js
```diff
@@ -229,6 +232,7 @@ AFRAME.registerElement("a-gltf-entity", {
           // If the src attribute is a selector, get the url from the asset item.
           if (src && src.charAt(0) === "#") {
             const assetEl = document.getElementById(src.substring(1));
+            if (!assetEl) { return; }
```
netpro2k: This should at least log an error. It might also make sense to have it emit `model-error` but not sure about that one.

src/elements/a-gltf-entity.js
netpro2k: Hmm, this is calling the AEntity version of attributeChangedCallback though, doesnt this still need to happen so that other components on the `a-gltf-entity` get their values updated?

src/react-components/avatar-selector.js
```diff
@@ -0,0 +1,119 @@
...
+
+  prevAvatar = () => {
+    const newAvatarIndex = this.getAvatarIndex(-1);
+    this.props.onChange(this.props.avatars[newAvatarIndex].id);
+  }
+
+  componentDidUpdate() {
+    // HACK - a-animation ought to restart the animation when the `to` attribute changes, but it doesn't
+    // so we need to force it here.
+    const currRot = this.animation.parentNode.getAttribute('rotation');
```
netpro2k: You probably want to filter for specific component updates here otherwise if we add more properties we could end up restarting the animation unintentionally.

src/react-components/avatar-selector.js
netpro2k: I too am a bit confused by this. If we plan to replace them we definitely should just be pulling somehting off the shelf. And even if we don't not sure icon design is the best use of any of our time lol

src/react-components/profile-entry-panel.js
netpro2k: I think it definitely makes sense to allow in game identity chagnes. That said we should probably switch to a different action set when we have the UI open instead of just blocking keyboard events.

src/assets/stylesheets/avatar-selector.scss
netpro2k: When we do we might also want to switch to using `:local` styles and importing in js instead of using css imports. This has the advantage of scoping all the css so we dont have to use really long names or worry about unintended cascading. Also, it ties the css to just the code using it, so it makes dead code removal easier because you don't have to guess what rules might be applying to what.

src/room.html
```diff
@@ -24,13 +24,69 @@
...
             <a-progressive-asset
-                id="bot-skinned-mesh"
+                id="botdefault"
                 response-type="arraybuffer"
                 src="./assets/avatars/BotDefault_Avatar_Unlit.glb"
                 high-src="./assets/avatars/BotDefault_Avatar.glb"
                 low-src="./assets/avatars/BotDefault_Avatar_Unlit.glb"
             ></a-progressive-asset>
-            <a-asset-item id="bot-dom-mesh" response-type="arraybuffer" src="./assets/avatars/BotDom_Avatar.glb"></a-asset-item>
+            <a-progressive-asset
```
netpro2k: Hmm.. This preloads ALL the avatars meaning the scene wont load until we download quite a large chunk of data, much of which we may never use. This is maybe fine for now but we should be just refering to avatars by URL and then we can just download them as needed. We will then probably have to figure out a way to show download progress for other users avatars.

src/room.js
```diff
@@ -133,7 +133,7 @@ async function exitScene() {
 function applyProfile(playerRig) {
   playerRig.setAttribute("player-info", {
     displayName: store.state.profile.display_name,
-    avatar: store.state.profile.avatar || "#bot-skinned-mesh"
+    avatar: '#' + (store.state.profile.avatar || "botdefault")
```
netpro2k: I think we should include the "#" in the avatar, this means its a "bundled" avatar but I should also be able to just point it at a URL (even if we dont have a UI for that yet)

src/storage/store.js
netpro2k: I vote we store the thing we end up passing to `src` aka either a url or a selector of a "bundled" `a-progressive-asset`, though I prefer we eventually ditch the later for everything but some sort of preloading avatar.

src/react-components/avatar-selector.js
gfodor: sure we can use Font Awesome. i didn't see a mic/speaker (or more recently, a webvr) icon that suited my needs in the public google design icon set so I just drew them. for these angle brackets we can def just use font awesome. in either case i think we want to pack all the necessary assets into the client repo and not link externally.

my main point was dropping the material design lite dependency (which I think is a good idea since we're not using it) will break these icons, so we should extract them, etc. i don't particularly care where they come from or if we make them ourselves etc.

src/storage/store.js
gfodor: I think if we can avoid a URL for now that is probably an good idea since there's a good chance URLs are gonna change for avatars in the near term. (For example, as we figure out the whole deploy pipeline, domain, GLTF bundle URLs, etc.)

I am not opposed to our long term plan being that your current avatar(s) are lists of URLs though.

gfodor: also, wrt "custom URLs for avatars" I believe we can have it both ways. We can have an `avatar_url` field, which if present, is assumed to be the custom override. (And perhaps eventually this can be the field we use for our 1st party avatars.)

The main thing I think we want to be careful about is burning in URLs to the 1st party avatars, because if we decide to change them we'll end up having to blow away peoples' avatar choices and/or maintain local storage migration scripts :/

src/room.html
gfodor: this is just history repeating itself :)

seriously though this is def a temporary solution. we should make sure the download time isn't too crazy, and perhaps we should not block loading the app on the avatars.

my mental model is that the environments "lead the way" wrt forcing us to build the tools and systems we will eventually need for avatars. For example the GLTF bundle work was forced into place by the environment needs, and the asset pipeline and image based customization stuff will probably fall out of the construction kit work. The avatars will probably force us to extend this work (for example, introducing LOD properly) but most of the story around how it all fits together will be in place once we have working dynamic custom environments for people.

src/storage/store.js
brianpeiris: Yeah, I imagine a lot is going to change with the asset bundles, so I'm not going to attempt anything fancy here. It's not difficult to change anyway.

src/react-components/profile-entry-panel.js
brianpeiris: Ah, forgot about in-game identity changing. Er, I agree with switching action sets, but I'm going to leave that for later, this PR is big enough.

src/room.html
brianpeiris: Yeah, I put in the avatars and the avatars.json manifest knowing it would be thrown away, which is fine, they were not the focus of this PR.

src/room.js
brianpeiris: I knew this was temporary. It's easy to change, so I'm going to leave it to later. I don't think I've precluded URL-based avatars in any significant way.

src/react-components/avatar-selector.js
brianpeiris: I'm going to leave this as it is until we decide on a final asset bundle format.

PR #109, opened 2018-03-27
Use babel for eslint parsing and prettier

gfodor:
Because we are now using 2 babel ES extensions eslint/prettier no longer works. This PR updates their configs to use babel as their parser.


brianpeiris:
Added these changes to #128

PR #112, opened 2018-03-27
Feature/new reticulum layout

gfodor:
This goes along with https://github.com/mozilla/reticulum/pull/9

This sets up the hub creation flow and integration with the new Reticulum API for creating and fetching hub info. It also lands the `gltf-bundle` component which enables the client to load GLTF bundles. Webpack has been updated so that you can drop the entire build output into the root of the reticulum static serving folder and it will "just work."

Note the creation flow works but is still unstyled.


src/hub.js
```diff
@@ -245,17 +252,45 @@ function mountUI(scene) {
...
 }
 
-const onReady = () => {
+const onReady = async () => {
   const scene = document.querySelector("a-scene");
   document.querySelector("a-scene canvas").classList.add("blurred");
   window.APP.scene = scene;
-  mountUI(scene);
+
+  const uiRoot = mountUI(scene);
```
gfodor: my linter says this is a bad idea, I will address this in my next PR for the UI

src/react-components/home-root.js
```diff
@@ -0,0 +1,62 @@
...
+    const environmentLoads = ENVIRONMENT_URLS.map(
+      src =>
+        new Promise(async resolve => {
+          const res = await fetch(src);
+          const data = await res.json();
+          data.bundle_url = src;
+          environments.push(data);
+          resolve();
+        })
+    );
```
brianpeiris: This is equivalent, but maybe too clever:

```javascript
const environmentLoads = ENVIRONMENT_URLS.map(
  src => (async () => {
    const res = await fetch(src);
    const data = await res.json();
    data.bundle_url = src;
    environments.push(data);
  })()
);
```

src/react-components/home-root.js
```diff
@@ -0,0 +1,62 @@
...
+
+    Promise.all(environmentLoads).then(() => this.setState({ environments }));
+  };
+
+  render() {
+    return (
+      <IntlProvider locale={lang} messages={messages}>
+        <div>{this.state.environments.length > 0 && <HubCreatePanel environments={this.state.environments} />}</div>
```
brianpeiris: `<div>` is redundant I think

src/react-components/hub-create-panel.js
```diff
@@ -0,0 +1,85 @@
...
+    environments: PropTypes.array
+  };
+
+  state = {
+    name: generateHubName(),
+    environmentBundleUrl: ""
+  };
+
+  componentDidMount() {
+    this.setState({ environmentBundleUrl: this.props.environments[0].bundle_url });
```
brianpeiris: This can be in the `constructor`, no?

src/react-components/hub-create-panel.js
```diff
@@ -0,0 +1,85 @@
+import React, { Component } from "react";
+import PropTypes from "prop-types";
+import { injectIntl, FormattedMessage } from "react-intl";
+import { SCHEMA } from "../storage/store";
```
brianpeiris: `SCHEMA` is unused.

webpack.config.js
```diff
@@ -186,23 +186,27 @@ const config = {
...
-      chunks: ["lobby"]
+      chunks: ["index"]
     }),
     new HTMLWebpackPlugin({
-      filename: "room.html",
-      template: path.join(__dirname, "src", "room.html"),
-      chunks: ["room"],
+      filename: "hub.html",
+      template: path.join(__dirname, "src", "hub.html"),
+      chunks: ["hub"],
```
brianpeiris: Hmm. "hub" != "room"?

src/hub.js
```diff
@@ -245,17 +252,45 @@ function mountUI(scene) {
...
+    return;
+  }
+
+  const hubId = document.location.pathname.substring(1).split("/")[0];
+  console.log(`Hub ID: ${hubId}`);
+  const res = await fetch(`/api/v1/hubs/${hubId}`);
+  const data = await res.json();
+  const hub = data.hubs[0];
+  const defaultSpaceChannel = hub.channels.find(c =>
+    c.attributes.find(a => a.length === 1 && a[0] === "default-space")
```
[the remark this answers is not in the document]
gfodor: yeah I can probably dial back the format here, its probably overengineering. basically i'm going on "past experience" with having arbitrary named key/value tag pairs for the space. so this is [["default-space"]] which is effectively a attribute with no value (otherwise the array would have two elements. we could change that to a map, or drop it altogether.

i think hubs[0] is fine because the API should eventually support fetching multiple hubs, so there's a wrapper `hubs` key that has an array attached. not sure if there is a more idiomatic way to do that these days but this is typically how I deal with future proofing API response going from 1 -> N

src/react-components/home-root.js
```diff
@@ -0,0 +1,62 @@
...
+import HubCreatePanel from "./hub-create-panel.js";
+
+const navigatorLang = (navigator.languages && navigator.languages[0]) || navigator.language || navigator.userLanguage;
+
+const lang = navigatorLang.toLowerCase().split(/[_-]+/)[0];
+
+import localeData from "../assets/translations.data.json";
+addLocaleData([...en]);
+
+const messages = localeData[lang] || localeData.en;
```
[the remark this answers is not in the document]
gfodor: yeah basically once per root element, which I guess could increase with time

src/react-components/home-root.js
```diff
@@ -0,0 +1,62 @@
...
+  static propTypes = {
+    intl: PropTypes.object
+  };
+
+  state = {
+    environments: []
+  };
+
+  componentDidMount() {
+    this.loadEnvironments();
```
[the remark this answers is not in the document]
gfodor: yup

src/react-components/home-root.js
gfodor: yeah this is going to fill out with the full design

webpack.config.js
gfodor: yeah this page will be rendered 1-1 (now and forever I think) to the request to render a hub, regardless of if there is 1 virtual space to a hub or not

src/hub.js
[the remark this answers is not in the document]
brianpeiris: Fair enough, `hubs[0]` makes sense.

src/hub.js
```diff
@@ -153,12 +155,16 @@ async function enterScene(mediaStream, enterInVR) {
 
   AFRAME.registerInputActions(inGameActions, "default");
 
+  document.querySelector("#player-camera").setAttribute("look-controls", "pointerLockEnabled: true;");
+
+  const qs = queryString.parse(location.search);
```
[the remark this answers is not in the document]
gfodor: haha, turns out we parse it 4 times!

src/index.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+    <meta charset="utf-8">
+    <title>moz://a duck</title>
```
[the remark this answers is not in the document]
gfodor: will do in styling pass

PR #113, opened 2018-03-27
aframe-extras navmesh

InfiniteLee:
added aframe-extras navmesh/pathfinding for our character-controller; replaced nav-mesh with new, much smaller one generated with blender navigation mesh tool; limiting teleport to navmesh; moved duck to be closer to edge of table;


InfiniteLee:
Tried using having a detached part of the navmesh for the table and the shelf in the back of the room, but appears to be unsupported in the pathfinding library.

src/components/character-controller.js
```diff
@@ -120,9 +123,23 @@ AFRAME.registerComponent("character-controller", {
...
+        let start = root.position;
+        end.copy(this.velocity).multiplyScalar(dt / 1000).add(start);
+        const nav = this.el.sceneEl.systems.nav;
+        this.navGroup = this.navGroup || nav.getGroup(start);
+        this.navNode = this.navNode || nav.getNode(start, this.navGroup);
+        this.navNode = nav.clampStep(start, end, this.navGroup, this.navNode, clampedEnd);
+        root.position.copy(clampedEnd);
+        root.position.x += this.velocity.x * dt / 1000;
+        root.position.y += this.velocity.y * dt / 1000;
+        root.z += this.velocity.z * dt / 1000;
```
brianpeiris: What's this?

src/room.html
```diff
@@ -168,7 +168,7 @@
                 class="left-controller"
                 hand-controls2="left"
                 tracked-controls
-                teleport-controls="cameraRig: #player-rig; teleportOrigin: #player-camera; button: action_teleport_"
+                teleport-controls="cameraRig: #player-rig; teleportOrigin: #player-camera; button: action_teleport_; collisionEntities: #floor-nav"
```
brianpeiris: nit: wrap the properties now that the component values are rather long.

src/components/character-controller.js
```diff
@@ -120,9 +123,23 @@ AFRAME.registerComponent("character-controller", {
...
       });
 
-      this.el.setAttribute("position", root.position);
-
       this.pendingSnapRotationMatrix.identity(); // Revert to identity
+
+      //copied from aframe-extras movement-controls
+      if (this.velocity.lengthSq() > EPS) {
+        let start = root.position;
+        end.copy(this.velocity).multiplyScalar(dt / 1000).add(start);
```
netpro2k: Velocity is already applied above, `root.position` is already the target position we are trying to move to.

src/components/character-controller.js
netpro2k: This whole second applicaiton of velocity looks to be a copy/paste typo https://github.com/donmccurdy/aframe-extras/blob/master/src/controls/movement-controls.js#L124

PR #114, opened 2018-03-28
Network the animation state of the hands.

johnshaughnessy:


src/components/hand-poses.js
```diff
@@ -0,0 +1,124 @@
...
+  init() {
+    this.animatePose = this.animatePose.bind(this);
+    this.animatePoses = this.animatePoses.bind(this);
+    this.firstUpdate = this.firstUpdate.bind(this);
+    if (this.el.querySelector(this.data.mixer)) {
+      this.mixer = this.el.querySelector(this.data.mixer).components["animation-mixer"];
+    }
+
+    let onLoad;
+    onLoad = () => {
```
brianpeiris: Why not `const onLoad =`?

src/components/hand-poses.js
```diff
@@ -0,0 +1,124 @@
...
+      if (this.el.querySelector(this.data.mixer)) {
+        this.mixer = this.el.querySelector(this.data.mixer).components["animation-mixer"];
```
brianpeiris: Avoid calling querySelector twice.

brianpeiris: Same on line 24

src/components/hand-poses.js
```diff
@@ -0,0 +1,124 @@
...
+      }
+      if (!this.mixer) {
+        // Can't find the mixer until spawned into the scene.
+        // TODO: Figure out the right event to listen to to get this done.
+        window.setTimeout(onLoad, 100);
+        return;
+      }
```
brianpeiris: Maybe the events are more sane after #125

src/components/hand-poses.js
johnshaughnessy: no reason. changed to `const onLoad =`

src/components/hand-poses.js
johnshaughnessy: 👍

src/components/hand-poses.js
netpro2k: If you just put this on the element with the mixer all of this (and the selector) shuould go away. The mixer is on RootScene so a new mixer, and new hand-poses will be created each time an avatar is loaded.

src/components/hand-poses.js
```diff
@@ -0,0 +1,125 @@
...
+    const suffix = hand == "left" ? "_L" : "_R";
+    const from = (this[`${hand}ClipFrom`] = this.mixer.mixer.clipAction(prev + suffix, this.clipActionObject));
+    const to = (this[`${hand}ClipTo`] = this.mixer.mixer.clipAction(curr + suffix, this.clipActionObject));
+
+    from.fadeOut(duration);
+    to.fadeIn(duration);
+    to.play();
+    from.play();
+
+    this.mixer.update(0.001);
```
netpro2k: Whats this about? Calling update manually on a component doesnt seem right, and besides that the update function takes an object with the old data, not a float. Is this supsoed to be calling `this.mixer.mixer.update`? If so, still seems odd that we would manually have to tick the mixer.

src/components/hand-poses.js
```diff
@@ -0,0 +1,125 @@
...
+      const object3DMap = this.mixer.el.object3DMap;
+      const rootObj = object3DMap.mesh || object3DMap.scene;
+      this.clipActionObject = rootObj.parent;
+      this.firstUpdate();
+    } else {
+      this.animatePoses(oldData);
+    }
+  },
+
+  firstUpdate() {
```
netpro2k: Why can't this be done in init?

src/components/hand-poses.js
```diff
@@ -0,0 +1,125 @@
...
+    this.leftClipFrom = this.leftClipTo = this.mixer.mixer.clipAction(POSES.open + "_L", this.clipActionObject);
```
netpro2k: The "from" clips don't appear to ever be used, and the 'to" clips seems like they could be just local variables.

netpro2k: Nevermind, found the dynamic access to them.. but that feels not ideal

src/components/hand-poses.js
```diff
@@ -0,0 +1,125 @@
...
+  thumbDown: "thumbDown",
+  indexDown: "indexDown",
+  mrpDown: "mrpDown",
+  thumbUp: "thumbsUp",
+  point: "point",
+  fist: "allGrip",
+  pinch: "pinch"
+};
+
+AFRAME.registerComponent("hand-poses", {
```
netpro2k: Haven't fully worked through it but it seems like this component could be much cleaner if you handled each hand's pose in its's own instance of the component, since their states are completely independant anyway. It would remove a lot of the duplicate code and wierd dynamic property access going on.

src/components/hand-poses.js
```diff
@@ -0,0 +1,125 @@
...
+    this.data.left.addEventListener("hand-pose", this.setHandPose);
+    this.data.right.addEventListener("hand-pose", this.setHandPose);
+  },
+
+  pause: function() {
+    this.data.left.removeEventListener("hand-pose", this.setHandPose);
+    this.data.right.removeEventListener("hand-pose", this.setHandPose);
+  },
+
+  setHandPose: function(evt) {
```
netpro2k: if you do the above suggestion where you make multiple instances of `hand-poses` this would become:
```javascript
const hand = evt.target === this.data.left ? "left" : "right";
this.el.setAttribute(`hand-poses__${hand}`, { pose });
```

src/hub.html
```diff
@@ -105,7 +105,7 @@
             </template>
 
             <template id="remote-avatar-template">
-                <a-entity ik-root player-info>
+                <a-entity ik-root player-info hand-poses="mixer: .RootScene">
```
netpro2k: I think `hand-poses` should jsut be on RootScene so it doesnt have to try and deal with any async stuff to get the mixer

src/hub.html
```diff
@@ -207,6 +207,8 @@
             spawn-controller="radius: 4;"
             wasd-to-analog2d
             character-controller="pivot: #player-camera"
+            hand-poses="mixer: .RootScene"
+            hand-poses-controller
```
netpro2k: I'd move both of these onto RootScene so they get recreated any time an avatar is switched.

PR #125, opened 2018-03-30
Convert a-gltf-entity to a gltf-model-plus component

netpro2k:
The original intent of `a-gltf-entity` was to allow using components referring to nested attributes of gltf models. With the additions of templating and setting src dynamically this usecase no longer makes much sense and mostly just adds complexity. This PR gets rid of the custom element in favor of a component `gltf-model-plus` with thes same functionality (minus the delaying of `load` on the entity its on).

This PR also includes a quick unrelated fix for the mic icon I noticed while testing in Chrome.


src/room.html
```diff
@@ -272,34 +272,34 @@
...
 
-        <a-gltf-entity
+        <a-entity
             id="outdoor-facade"
-            src="#outdoor-facade-mesh"
+            gltf-model-plus="src: #outdoor-facade-mesh"
             position="0 0 0"
             xr="ar: false"
-        ></a-gltf-entity>
+        ></a-ntity>
```
brianpeiris: typo

src/components/gltf-model-plus.js
```diff
@@ -187,133 +187,95 @@ function cachedLoadGLTF(src, onProgress) {
...
-            }
-          }
+  init() {
+    this.loadTemplates();
+  },
 
-          if (src === this.lastSrc) return;
-          this.lastSrc = src;
+  update() {
+    this.applySrc(this.data.src);
```
robertlong: Add a `.catch` clause here so that any errors will be handled correctly.

src/react-components/2d-hud.css
```diff
@@ -37,6 +37,8 @@
   display: flex;
   width: 48px;
   height: 48px;
+  -webkit-mask: url(../assets/hud/unmuted.png);
```
robertlong: Why is this in here? Does your editor autoprefix or is this a bug fix?

netpro2k: bugfix, I noted it in the PR description.

src/components/gltf-model-plus.js
netpro2k: applySrc checks for errors, so it will always resolve.

Issue #116, opened 2018-03-28
Entry flow UX changes based on feedback


gfodor:
UX pass should add:

- [x] Blur on background canvas may be prohibitively expensive on mobile, can disable + decrease opacity instead. (#159)
- [x] Show "identity picker" as first screen until the user enters a custom name (#175)
- [x] Fix up mic/speaker UX so there is no confusion about them being buttons. (Either remove circle, or make them buttons to play tone/change mic.) (#175)
- [x] Also use new "blue" #2F80ED (#175)
- [x] For some reason the "mic permission" button is way too big, it is supposed to be the same size as the other buttons. (#175)
- [x] microphone level should have increased sensitifity (probably something like min(level * 2, 1.0)) (#175)
- ---
- [x] Consider more affordances for text buttons at bottom (#177)
- [x] Add a minimum age, ToS/PP checkbox before allowing entry (#177)
- [x] Fonts not loading on mobile for hub.html (#177)
- [x] session ended screen mis-styled now (#177)
- [x] Also fix behavior where we are holding a reference to the react component root for the UI (#177)
- ---
- [x] Add a "share link" UX (#232)
- [x] Add the name of the room to the UI somewhere (#232)
- [x] Enter key should create hub (#232)
- [x] Preload environment images (#232)
- [x] Header should not expand vertically on landing page (#210)

gfodor:
Also use new "blue" #2F80ED

gfodor:
Also fix behavior where we are holding a reference to the react component root for the UI

gfodor:
Enter key should create hud
Preload environment images

gfodor:
Header should not expand vertically on landing page
Fonts not loading on mobile for hub.html

gfodor:
Blur on background canvas may be prohibitively expensive on mobile, can disable + decrease opacity instead.

gfodor:
For some reason the "mic permission" button is way too big, it is supposed to be the same size as the other buttons.

gfodor:
microphone level should have increased sensitifity (probably something like `min(level * 2, 1.0)`)

brianpeiris:
Gonna start working on some of these today.

robertlong:
@gfodor @brianpeiris that background blur is giving me about 3fps on FF on my mac 😬 Can we split that into its own issue and get that fixed sooner than the rest of these styling issues?

brianpeiris:
Sure, I can do that.

PR #129, opened 2018-03-30
Big delint

brianpeiris:
Based on #128 and compared to an automatically prettified `other/prettier` branch.


src/hub.js
```diff
@@ -235,6 +237,7 @@ function mountUI(scene) {
     />,
     document.getElementById("ui-root")
   );
+  /* eslint-enable react/no-render-return-value */
```
brianpeiris: See #116

PR #130, opened 2018-03-30
Set the renderOrder of the items that have to appear on top

johnshaughnessy:
This PR sets the render order of the items in the in-world-hud to 1 so that they appear on top of the hud's background without blending with it. (The hud's background's render order is 0.)
It also sets the raycaster's renderOrder to 2 so that it appears on top of all of it. In order for the raycaster to appear on top, it needs to also be drawn during the transparent objects' render pass, which happens after all the opaque objects. Hence we set the raycaster's line's opacity to 0.99, forcing it into the transparent pass.


brianpeiris:
👍 Latest changes look good to me. 

I would also suggest setting the `far` property on the raycaster, so that the line doesn't extend too far and conflict with so much of the environment `raycaster="objects:.hud; showLine: true; far: 2;"`

PR #132, opened 2018-03-30
Feature/creation styling

gfodor:
Styling for homepage, refactor a bunch of SASS (WIP)


src/react-components/avatar-selector.js
```diff
@@ -1,124 +1,113 @@
-import React, { Component } from 'react';
```
gfodor: these changes are 99% just prettier + linter autofixes

brianpeiris: 😬 thanks

src/react-components/hub-create-panel.js
```diff
@@ -48,37 +60,104 @@ class HubCreatePanel extends Component {
...
+
+                if (this.state.expanded) {
+                  this.setState({ name: generateHubName() });
+                } else {
+                  this.setState({ expanded: true });
+                }
+              }}
+            >
+              {this.state.expanded ? (
+                <img className="create-panel__form__rotate_button" src="../assets/images/dice_icon.svg" />
```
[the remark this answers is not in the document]
gfodor: /shrug will fix

src/assets/stylesheets/hub-create.scss
```diff
@@ -0,0 +1,221 @@
...
+	text-align: center;
+      }
+
+      @media (max-width: 520px) {
+	padding-left: 15px;
+	font-size: 1.2em;
+      }
+      
+      &--expanded {
+	padding-top: 365px;
```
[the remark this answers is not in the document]
gfodor: ah nice one, this was burning me and this makes much more sense. can remove the `z-index`'es probably now

src/assets/stylesheets/hub-create.scss
```diff
@@ -0,0 +1,221 @@
...
+      height: 95px;
+      left: 4px;
+      display: flex;
+      align-items: center;
+      justify-content: center;
+      cursor: pointer;
+      bottom: 0;
+
+      @media (max-width: 520px) {
+	display: none;
```
brianpeiris: Hmm, this hides the button on my S8. Is that intentional?

gfodor: yes unfortunately it ends up truncating the text field too much. this means when you create a room on your phone you can't pick the scene. (yes this is bad.) what i probably should do is expand the thing automatically via a viewport check at runtime. what do you think?

gfodor: actually I might be able to refactor everything to be CSS-based for expansion so maybe can use media queries.

brianpeiris: The randomize button is still hidden after you expand it. Maybe we could move the text input on to its own row and have the randomize and submit button below it.
Not needed for this PR though.

gfodor: I put in the hack for now so it expands on small screens. Def should fix later though.

PR #134, opened 2018-03-30
Better solution to account for html prefix for avatar selector

brianpeiris:


netpro2k:
The fact that we need this seems very wrong. Why would we generate smoke versions of each html file instead of just building a smoketest version of the app and serving it up somehwere else?

brianpeiris:
@netpro2k hopefully @gfodor answered your questions in slack.

gfodor:
posted about this in slack but it's generally a bit of a catch-22 -- the goal of the smoke test is to minimize as much as possible the delta between both the code and environment before rolling out the change to production. so basically the ideal scenario is that "we roll out the same code to one machine, test there, and then roll the same code out as-is to the remainder of the machines."

I think probably the best configuration we can get to given the nature of our project is:
- We migrate away from using reticulum as the asset host for everything except the `html` files that are served up by reticulum directly
- The other assets (css, js, etc) are digested as they are now but are deployed by CI to S3
- The origin of the asset cloudfront end point is changed to the S3 bucket

The advantage of this approach is that master's built assets are always pushed to the asset CDN at commit-time, which is safe because they are digested and won't conflict with any other references to the old versions.

If we do that, and we also set things up so we have a digested version of `avatar-picker.html`, I believe we can then hit the same exact URLs on the smoke test box since there won't be a need for a smoke-box bound asset origin.

gfodor:
(Also this PR lgtm this is the cleanest mechanism I think we can do for now to minimize changes in runtime behavior between the smoke box and the non-smoke boxes until we have a unified asset origin for both machine types that will allow us to push the assets out pre-emptively in a safe way.)

PR #140, opened 2018-04-04
Space Bubble fixes and improvments

netpro2k:
Space bubbles are currently broken. This PR fixes them but also attempts to make them better by:
- allowing individual `personal-space-invaders` to specify their own radius, giving better control over when personal space should be triggered
- using transparency for certain personal space invasions (instead of fully hiding the objects)

This also cleans up some of the visibility code that was scattered throughout.

Todo:
[] Need to deal with the fact that both `ik-controller` and `personal-space-bubble` want to control hand visiblity.

### DIFF
--- a/src/systems/personal-space-bubble.js
+++ b/src/systems/personal-space-bubble.js
@@ -49,7 +49,7 @@ AFRAME.registerSystem("personal-space-bubble", {
 
     for (let i = 0; i < this.invaders.length; i++) {
       this.invaders[i].el.object3D.updateMatrixWorld(true);
-      this.invaders[i].setVisibility(true);
+      this.invaders[i].setInvading(false);
     }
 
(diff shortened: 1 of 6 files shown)


src/systems/personal-space-bubble.js
```diff
@@ -40,55 +44,120 @@ AFRAME.registerSystem("personal-space-bubble", {
...
+  schema: {
+    meshSelector: { type: "string" }
+  },
+  init() {
+    this.targetMesh = this.el.querySelector(this.data.meshSelector).object3DMap.skinnedmesh;
+    console.log("target", this.targetMesh);
+  }
+});
+
+function findInvderMesh(entity) {
```
haydenjameslee: Small typo: `findInvderMesh` -> `findInvaderMesh`

PR #141, opened 2018-04-04
Bug/allow denial of mic

gfodor:
This PR:

- Fixes the flow so if you deny mic permissions, it will continue and just not show the mic level.
- Fixes bugs and adds local storage state so that your mic choice will stick across sessions.

### DIFF
--- a/src/react-components/ui-root.js
+++ b/src/react-components/ui-root.js
@@ -288,17 +288,14 @@ class UIRoot extends Component {
 
   setMediaStreamToDefault = async () => {
     let hasAudio = false;
-    console.log(this.props.store.state);
     const { lastUsedMicDeviceId } = this.props.store.state;
 
     // Try to fetch last used mic, and if we don't get it then fall back to default.
     if (lastUsedMicDeviceId) {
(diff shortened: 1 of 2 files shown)


src/react-components/ui-root.js
```diff
@@ -399,17 +433,20 @@ class UIRoot extends Component {
...
-    );
+    return this.micLabelForMediaStream(this.state.mediaStream);
+  };
+
+  micDeviceIdForMicLabel = label => {
+    return this.state.micDevices.filter(d => d.label === label).map(d => d.deviceId)[0];
   };
 
   selectedMicDeviceId = () => {
-    return this.state.micDevices.filter(d => d.label === this.selectedMicLabel).map(d => d.deviceId)[0];
```
gfodor: yes this was an ugly embarassing bug (no `()` on `selectedMicLabel`)

brianpeiris: I spent a bit of time trying Flow on this repo, but it was a hassle to make it understand AFrame and Three and it just spat out hundreds of errors, so it would take some effort to iron all of that out.

PR #146, opened 2018-04-06
Add support for MOZ_alt_materials

robertlong:
MOZ_alt_materials and the updated `gltf-model-plus` component replaces `<a-progressive-asset>`. Unlit materials will be used on mobile and PBR materials are used everywhere else.

In the future we will want to move away from determining what technique to use based on your device type and instead move to something based on the device performance.


src/components/gltf-model-plus.js
```diff
@@ -214,6 +218,15 @@ AFRAME.registerComponent("gltf-model-plus", {
     );
   },
 
+  getPreferredTechnique() {
```
netpro2k: Why not just actually set the default schema value?
```
preferredTechnique: { default: AFRAME.utils.device.isMobile() ? "KHR_materials_unlit": "pbrMetallicRoughness" }
```

src/vendor/GLTFLoader.js
```diff
@@ -14,6 +15,7 @@ THREE.GLTFLoader = ( function () {
 
 		this.manager = ( manager !== undefined ) ? manager : THREE.DefaultLoadingManager;
 		this.dracoLoader = null;
+		this.preferredTechnique = null;
```
netpro2k: Any reason not to make this a constructor parameter? (Or maybe add an options map to the constructor instead)

src/vendor/GLTFLoader.js
```diff
@@ -874,6 +883,33 @@ THREE.GLTFLoader = ( function () {
...
+		this.name = EXTENSIONS.MOZ_ALT_MATERIALS;
+		this.preferredTechnique = preferredTechnique;
+
+	}
+
+	MOZAltMaterialsExtension.prototype.getAltMaterial = function ( materialDef, materials ) {
+
+		if ( materialDef.extensions && materialDef.extensions[ EXTENSIONS.MOZ_ALT_MATERIALS ]) {
+
+			var mamExtension = materialDef.extensions[ EXTENSIONS.MOZ_ALT_MATERIALS ];
```
netpro2k: Minor but id simplify this to:
```
const mamExtension = materialDef.extensions && materialDef.extensions[ EXTENSIONS.MOZ_ALT_MATERIALS ];
if(mamExtension) {
```

src/vendor/GLTFLoader.js
robertlong: I'm just following what the Draco loader is doing. I'll submit a PR to Three.js and see what they think.

PR #147, opened 2018-04-06
Use external environments

netpro2k:
- add support for relative urls in bundles
- remvoes the bundled environments from the client, and the tempaltes from webpack build step
- adds support for incomplete metadata in the environment selector
- adds the urls to all the existing environments to the list of selectable environment
- fixes an unrelated but annoying proptype warning in UiRoot


src/components/gltf-bundle.js
```diff
@@ -5,6 +5,7 @@ AFRAME.registerComponent("gltf-bundle", {
 
   init: async function() {
     this._addGltfEntitiesForBundleJson = this._addGltfEntitiesForBundleJson.bind(this);
+    this.baseURL = new URL(THREE.LoaderUtils.extractUrlBase(this.data.src), window.location.href);
```
gfodor: is `window.location` the right thing to use to cover `iframe`s?

src/hub.js
```diff
@@ -261,7 +261,12 @@ const onReady = async () => {
   if (qs.room) {
     // If ?room is set, this is `yarn start`, so just use a default environment and query string room.
     uiRoot.setState({ janusRoomId: qs.room && !isNaN(parseInt(qs.room)) ? parseInt(qs.room) : 1 });
-    initialEnvironmentEl.setAttribute("gltf-bundle", "src: /assets/environments/cliff_meeting_space/bundle.json");
+    initialEnvironmentEl.setAttribute("gltf-bundle", {
+      src: "https://asset-bundles-dev.reticulum.io/rooms/meetingroom/MeetingRoom.bundle.json"
```
gfodor: Comments intentional?

gfodor: this should point to `-prod` now

src/react-components/avatar-selector.js
```diff
@@ -4,7 +4,9 @@ import { injectIntl, FormattedMessage } from "react-intl";
 import FontAwesomeIcon from "@fortawesome/react-fontawesome";
 import faAngleLeft from "@fortawesome/fontawesome-free-solid/faAngleLeft";
 import faAngleRight from "@fortawesome/fontawesome-free-solid/faAngleRight";
-import meetingSpace from "../assets/environments/cliff_meeting_space/MeetingSpace1_mesh.glb";
+
+// TODO: we should make a bundle for avatar picker with it's own geometry, for now just use the indoor part of the meting room
+const meetingSpace = "https://asset-bundles-dev.reticulum.io/rooms/meetingroom/MeetingSpace1_mesh-d48250ebc6.gltf";
```
gfodor: this should point to `-prod` now

src/react-components/home-root.js
```diff
@@ -16,7 +16,10 @@ addLocaleData([...en]);
 const messages = localeData[lang] || localeData.en;
 
 const ENVIRONMENT_URLS = [
-  `${document.location.protocol}//${document.location.host}/assets/environments/cliff_meeting_space/bundle.json`
+  "https://asset-bundles-dev.reticulum.io/rooms/meetingroom/MeetingRoom.bundle.json",
+  "https://asset-bundles-dev.reticulum.io/rooms/theater/TheaterMeshes.bundle.json",
+  "https://asset-bundles-dev.reticulum.io/rooms/atrium/AtriumMeshes.bundle.json",
+  "https://asset-bundles-dev.reticulum.io/rooms/courtyard/CourtyardMeshes.bundle.json"
```
gfodor: ditto, `-prod`

src/components/gltf-bundle.js
robertlong: I think its just to support relative paths to gltf bundles.

netpro2k: I think they are equivelent, in an iframe you would have to do `window.parent.location` to get the outer context's location (if you had permission to do so).

PR #148, opened 2018-04-06
Reuse NAF's media stream source to avoid distorted audio in Safari

### LINKED ISSUE #99: Safari WebRTC audio not working
Probably should get this fixed before MVP launch due to the surface area of the platform, etc.

brianpeiris:
Using this modification to networked-aframe: https://github.com/MozillaReality/networked-aframe/commit/0fecbec73d6508e7afcb99b19df3f3b21947bc77

Closes #99


brianpeiris:
Corresponding networked-aframe PR: https://github.com/networked-aframe/networked-aframe/pull/122

src/components/audio-feedback.js
```diff
@@ -1,20 +1,13 @@
...
-    if (!stream) {
-      return;
-    }
-
-    const ctx = THREE.AudioContext.getContext();
-    const source = ctx.createMediaStreamSource(stream);
-    this.analyser = ctx.createAnalyser();
-    this.levels = new Uint8Array(this.analyser.frequencyBinCount);
-    source.connect(this.analyser);
+    this.el.addEventListener("sound-source-set", event => {
```
netpro2k: I'd prefer a model wher eyou got the `networked-audio-source` component and asked it for a source, but an evented model works too. Probably just down to personal preference .

package.json
```diff
@@ -33,7 +33,7 @@
     "mobile-detect": "^1.4.1",
     "moving-average": "^1.0.0",
     "naf-janus-adapter": "https://github.com/mozilla/naf-janus-adapter#feature/disconnect",
-    "networked-aframe": "https://github.com/mozillareality/networked-aframe#mr-social-client/master",
+    "networked-aframe": "github:mozillareality/networked-aframe#feature/sound-source-set",
```
netpro2k: We have a few PRs in flight again that depend on unlanded NAF changes, we should use the `mr-social-client/master` integration branch again

brianpeiris: :+1: will do

src/components/audio-feedback.js
brianpeiris: I wanted to avoided reaching into `el.components`. Feels like bad practice.

netpro2k: I don't think its really bad practice to have components depend on other components. Though I do thnk it would be not ideal to reach directly into that other components data. I assumed the stream might be exposed by a function that returned a promise to a stream.

brianpeiris: Hmm, yeah, fair point. I'll bring it up on the NAF PR.

PR #151, opened 2018-04-06
Enter in VR properly on GearVR

gfodor:
This PR properly handles the `vr_entry_type=gearvr` case once we've entered Oculus Browser. (Previously it would just continually try to go to the `ovrweb` URL)

### DIFF
--- a/src/react-components/ui-root.js
+++ b/src/react-components/ui-root.js
@@ -250,7 +250,7 @@ class UIRoot extends Component {
   };
 
   enterGearVR = async () => {
-    if (this.state.availableVREntryTypes.daydream !== VR_DEVICE_AVAILABILITY.yes) {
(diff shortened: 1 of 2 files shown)


src/react-components/ui-root.js
```diff
@@ -246,18 +250,22 @@ class UIRoot extends Component {
   };
 
   enterGearVR = async () => {
-    this.exit();
+    if (this.state.availableVREntryTypes.daydream !== VR_DEVICE_AVAILABILITY.yes) {
```
brianpeiris: Wouldn't it be more clear to test for `isOculusBrowser`?

gfodor: whoa, major bug there. nice catch. it should check for `gearvr` i think

gfodor: the `gearvr` check for `yes` basically implies it is safe to enter VR as is. right now `yes` is only true within Oculus Browser, as per the other change int his app.

PR #153, opened 2018-04-06
Close mic resources on entry

gfodor:


robertlong:
Looks good to me. Although, UIRoot should probably be split into multiple components in the future.

PR #157, opened 2018-04-06
Add closing environment html tag

robertlong:
Custom html elements are required to have closing tags.


gfodor:
on the fence about this one, but I think it's OK

just kidding

PR #159, opened 2018-04-07
Bug/macos blur perf

brianpeiris:
Figured out a way to keep the blur but reduce perf. Basically I let a-frame render until its loaded the environment, then I pause the render loop. This seems to fix the perf issue with the blur effect on macos Firefox. (For #116)

Also fixed a bunch of small issues:
- Fix regression with look-controls so that it's not active when the react UI is up.
- Removed unused `onConnect` method
- Fix htmlPrefix error message
- Fix GLTFLoader timer warnings

### DIFF
--- a/src/hub.js
+++ b/src/hub.js
@@ -1,5 +1,6 @@
 import "./assets/stylesheets/hub.scss";
 import queryString from "query-string";
+import { debounce } from "lodash";
 
 import { patchWebGLRenderingContext } from "./utils/webgl";
 patchWebGLRenderingContext();
@@ -126,6 +127,7 @@ async function enterScene(mediaStream, enterInVR, janusRoomId) {
   const scene = document.querySelector("a-scene");
   const playerRig = document.querySelector("#player-rig");
   document.querySelector("a-scene canvas").classList.remove("blurred");
+  scene.render();
 
   if (enterInVR) {
     scene.enterVR();
@@ -179,11 +181,7 @@ async function enterScene(mediaStream, enterInVR, janusRoomId) {
     screenEntity.setAttribute("visible", sharingScreen);
   });
 
-  if (qsTruthy("offline")) {
-    onConnect();
-  } else {
-    document.body.addEventListener("connected", onConnect);
-
+  if (!qsTruthy("offline")) {
     scene.components["networked-scene"].connect();
 
     if (mediaStream) {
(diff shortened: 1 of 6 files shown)


src/vendor/GLTFLoader.js
```diff
@@ -80,7 +80,7 @@ THREE.GLTFLoader = ( function () {
 
 		},
 
-		parse: function ( data, path, onLoad, onError ) {
+		parse: function ( url, data, path, onLoad, onError ) {
```
netpro2k: I kept wanting to make this change but couldnt figure out a clean way to do it without altering the signature of parse but still get the url. I guess modifying it is fine but id add it as the last param to keep compatability. Would prefer to try and come up with a solution we can submit upstream though...

src/hub.js
```diff
@@ -255,7 +250,10 @@ const onReady = async () => {
   const environmentRoot = document.querySelector("#environment-root");
 
   const initialEnvironmentEl = document.createElement("a-entity");
-  initialEnvironmentEl.addEventListener("bundleloaded", () => uiRoot.setState({ initialEnvironmentLoaded: true }));
+  initialEnvironmentEl.addEventListener("bundleloaded", () => {
+    uiRoot.setState({ initialEnvironmentLoaded: true });
+    setTimeout(() => scene.renderer.animate(null), 100);
```
netpro2k: Why is this timeout needed? Can it be 0?

gfodor:
Nice one!

robertlong:
@brianpeiris I'm not seeing the look-controls attached to the camera:

```
document.querySelector("[look-controls]") === null
```

src/vendor/GLTFLoader.js
brianpeiris: Added the url as an instance param instead, to avoid changing the signature.

Most of the other changes are about MOZ_ALT_MATERIALS vs upstream https://gist.github.com/brianpeiris/59f9707f87c679e94e3f5411664b57d2

src/hub.js
brianpeiris: Yeah, you're right, changed it to 0 timeout and added a comment. Just need to wait one tick to actually allow the environment to render.

brianpeiris:
@robertlong Good catch. Fixed.

PR #160, opened 2018-04-07
lint and delint html

brianpeiris:
Add [htmlhint](https://github.com/yaniswang/HTMLHint)


scripts/lint-html.js
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env node
```
netpro2k: Surprised you had to write this script. Can't we use something like https://github.com/htmllint/htmllint-cli ?

netpro2k:
We should probably try and get this merged soon since its going to likely keep creating conflicts. Also would be nice to stop making silly html errors heh

scripts/lint-html.js
brianpeiris: The only reason I had to write this script is to strip the `<% %>` template tags from our HTML (in a cross-platform manner).

PR #161, opened 2018-04-09
Feature/basic phoenix channel

gfodor:
This PR:

- Adds back in phoenix channels, gets hub info on join
- Sends a "entry event" to reticulum to record stats

Goes with: https://github.com/mozilla/reticulum/pull/15

### DIFF
--- a/src/hub.js
+++ b/src/hub.js
@@ -1,5 +1,7 @@
 import "./assets/stylesheets/hub.scss";
+import uuid from "uuid/v4";
 import queryString from "query-string";
+import { Socket } from "phoenix";
 
 import { patchWebGLRenderingContext } from "./utils/webgl";
 patchWebGLRenderingContext();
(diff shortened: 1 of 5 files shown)


src/hub.js
```diff
@@ -181,6 +190,12 @@ async function enterScene(mediaStream, enterInVR, janusRoomId) {
   });
 
   if (!qsTruthy("offline")) {
+    document.body.addEventListener("connected", () => {
+      hubChannel.sendEntryEvent().then(() => {
+        store.update({ lastEnteredAt: moment().format() });
```
brianpeiris: `moment().format()` is in the user's current timezone. We should use UTC, no? `moment().toJSON()` does that.

brianpeiris: More importantly, `moment().format()` does not contain timezone information.

brianpeiris: Oh, nevermind. We're using moment-timezone, which _does_ return timezone information 😑

gfodor: Cool, you're right though we should store it in UTC

PR #174, opened 2018-04-10
HUD Rework

netpro2k:
- Reposition the HUD overhead
- Remove the username 
- Prep for freeze/uncreeze toggle and new button layout
- Remove depth re-ordering to see if we can live without it in its new position

Note: this does not (yet) modify the behavior of the hud mode-switching enabling laser pointer and disabling movement/teleporting. This will happen after the 3D cursor has been reworked. 
Also note that both the center button (an avatar right now) and the secondary mute button are both visual placeholders.


johnshaughnessy:
I think it's easy to miss the name selection option when entering the room, so I suspect people will join with a random name and then not know why people are calling them by that name. I think putting the name in the UI might help with this.

src/systems/app-mode.js
```diff
@@ -101,10 +101,10 @@ function deltaAngle(a, b) {
 AFRAME.registerComponent("hud-controller", {
   schema: {
     head: { type: "selector" },
-    offset: { default: 1 }, // distance from hud below head,
-    lookCutoff: { default: -25 }, // angle at which the hud should be "on",
+    offset: { default: 0.7 }, // distance from hud above head,
+    lookCutoff: { default: 20 }, // angle at which the hud should be "on",
     animRange: { default: 30 }, // degrees over which to animate the hud into view
-    yawCutoff: { default: 100 } // yaw degrees at wich the hud should reoirent even if the user is looking down
+    yawCutoff: { default: 50 } // yaw degrees at wich the hud should reoirent even if the user is looking up
```
johnshaughnessy: at which*
reorient*

johnshaughnessy:
image
There is some Z-fighting in the HUD.

netpro2k:
Looked into mxiing the mic feedback stuff and its a bit messy so its best being its own PR. This wasnt a regression in this PR anyway, so I think this seems ok.

src/react-components/2d-hud.js
```diff
@@ -4,18 +4,21 @@ import cx from "classnames";
 
 import styles from "../assets/stylesheets/2d-hud.css";
 
-const TwoDHUD = ({ name, muted, onToggleMute }) => (
+const TwoDHUD = ({ muted, onToggleMute }) => (
   <div className={styles.container}>
-    <div className={cx("ui-interactive", styles.bg)}>
+    <div className={cx("ui-interactive", styles.panel, styles.left)}>
       <div className={cx(styles.mic, { [styles.muted]: muted })} onClick={onToggleMute} />
```
brianpeiris: IMO anything with an `onClick` on it should be a `<button>`. It's semantic and you get interaction states for free.

Issue #167, opened 2018-04-09
Update teleporter/ratchet turning input schemes


gfodor:
D-Pad for ratchet turning, center press for teleporter on all platforms. 

On 6-dof, consider making teleporter input buttons the same on both hands, and making locomotion controls less discoverable and modal to enable.

netpro2k:
I'd vote for not adding a mode here but instead adding handedness selection to the profile screen. Should be fairly easy to toggle which hand is considered primary.

PR #178, opened 2018-04-11
Update input schemes

### LINKED ISSUE #163: Add input mapping for arrow keys in addition to WASD
During our last standup a user was not familiar with the WASD controls and expected the arrow keys to move them around. We should probably add the arrow keys as an additional method of moving the character controller.

netpro2k:
This updates the input schemes (#167) on all platforms to be more consistant and remove duplicated funcitonality on buttons. 

Big changes are upport for ratchet turning to gearvr and daydream, using trigger for grab on vive, remove of duplicate teleport butotns on oculus, and removal of mute buttons on oculus and vive (these are now handled via the HUD). Daydream and gearvr still have no way to pick up objects, but this will come with the cursor changes. 

Also fixes #163 by adding support for arrow keys.

### DIFF
--- a/src/input-mappings.js
+++ b/src/input-mappings.js
@@ -21,29 +21,35 @@ const config = {
   behaviours: {
     default: {
       "oculus-touch-controls": {
-        joystick: "oculus_touch_joystick_dpad4"
+        joystick: "joystick_dpad4"
       },
       "vive-controls": {
-        trackpad: "vive_trackpad_dpad4"
+        trackpad: "trackpad_dpad4"
+      },
+      "daydream-controls": {
+        trackpad: "trackpad_dpad4"
+      },
+      "gearvr-controls": {
+        trackpad: "trackpad_dpad4"
       }
     }
   },
   mappings: {
     default: {
       "vive-controls": {
-        menudown: ["action_mute", "thumb_down"],
-        menuup: "thumb_up",
         "trackpad.pressedmove": { left: "move" },
         trackpad_dpad4_pressed_west_down: { right: "snap_rotate_left" },
         trackpad_dpad4_pressed_east_down: { right: "snap_rotate_right" },
         trackpad_dpad4_pressed_center_down: { right: "action_teleport_down" },
         trackpadup: { right: "action_teleport_up" },
-        gripdown: ["action_grab", "middle_ring_pinky_down", "index_down"],
-        gripup: ["action_release", "middle_ring_pinky_up", "index_up"],
+        menudown: "thumb_down",
+        menuup: "thumb_up",
+        gripdown: ["middle_ring_pinky_down", "index_down"],
+        gripup: ["middle_ring_pinky_up", "index_up"],
         trackpadtouchstart: "thumb_down",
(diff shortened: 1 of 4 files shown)


src/input-mappings.js
```diff
@@ -100,14 +107,14 @@ const config = {
...
         d_up: "d_up",
-        W_down: "w_down",
-        W_up: "w_up",
-        A_down: "a_down",
-        A_up: "a_up",
-        S_down: "s_down",
-        S_up: "s_up",
-        D_down: "d_down",
-        D_up: "d_up"
+        arrowup_down: "w_down",
```
InfiniteLee: Can we add the actual arrow keys here while we're at it?

netpro2k: those are the actual arrow keys. This line is binding the up arrow press to the "w_down" action (which gets converted to movement by the `wasd-to-analog2d` component. It's a bit roundabout and should get better with changes to aframe-input-mappings

src/input-mappings.js
```diff
@@ -21,29 +21,35 @@ const config = {
...
         "trackpad.pressedmove": { left: "move" },
         trackpad_dpad4_pressed_west_down: { right: "snap_rotate_left" },
         trackpad_dpad4_pressed_east_down: { right: "snap_rotate_right" },
         trackpad_dpad4_pressed_center_down: { right: "action_teleport_down" },
         trackpadup: { right: "action_teleport_up" },
-        gripdown: ["action_grab", "middle_ring_pinky_down", "index_down"],
-        gripup: ["action_release", "middle_ring_pinky_up", "index_up"],
+        menudown: "thumb_down",
+        menuup: "thumb_up",
+        gripdown: ["middle_ring_pinky_down", "index_down"],
```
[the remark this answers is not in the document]
netpro2k: Yeah I was on the fence about it. Figured we might want to use it for something else later, but maybe not. The only thing I am a bit unsure about is if we allow it for grabbing but then not for other actions (like clicking with the cursor) will that be confusing.. Maybe not.

src/input-mappings.js
```diff
@@ -67,20 +72,22 @@ const config = {
...
+        trackpad_dpad4_pressed_west_down: "snap_rotate_left",
+        trackpad_dpad4_pressed_east_down: "snap_rotate_right",
+        trackpad_dpad4_pressed_center_down: "action_teleport_down",
         trackpadup: "action_teleport_up"
       },
       "gearvr-controls": {
-        trackpaddown: "action_teleport_down",
+        trackpad_dpad4_pressed_west_down: "snap_rotate_left",
+        trackpad_dpad4_pressed_east_down: "snap_rotate_right",
+        trackpad_dpad4_pressed_center_down: "action_teleport_down",
```
brianpeiris: Snap rotate works well on gearvr, but it feels odd that up and down are dead. We should map them to move forward/back or at least map them to teleport.

netpro2k: Ok, I think I will map them to teleport for all the trackpad platforms. We can implement a sort of blink-forward if it makes sense later.

PR #179, opened 2018-04-11
Fix lerping in networked-aframe

robertlong:
Merge https://github.com/networked-aframe/networked-aframe/pull/124 into our fork of networked-aframe.

Tests still need to be fixed and we probably want to be able to switch off lerping on a per component basis in the network schema. Right now snap rotation has the previously known bug where rotating once will take a remote entity a long time to reach its target rotation.

I think we can merge without the fixed tests. I'll work with @haydenjameslee to try to figure out what's causing unrelated tests to break and add new tests in.

Fixes https://github.com/mozilla/mr-social-client/issues/74


brianpeiris:
While you're at it, `player-rig` still uses `attachLocalTemplate`.

robertlong:
@brianpeiris what version of `networked-aframe` should I be pointing at now? Are you going to merge your branch into `mr-social-client/master`?

PR #182, opened 2018-04-11
fix/workaround for teleport-controls collisionEntities on mobile

InfiniteLee:
Add `nav-mesh-helper` component to force calling `queryCollisionEntities()` on `teleport-controls` once an environment has changed.

fixes mozilla/socialmr#78


src/components/nav-mesh-helper.js
```diff
@@ -0,0 +1,16 @@
+AFRAME.registerComponent("nav-mesh-helper", {
+  schema: {
+    teleportControls: { type: "selectorAll", default: "[teleport-controls]" }
+  },
+
+  init: function() {
+    const teleportControls = this.data.teleportControls;
+    this.el.addEventListener("model-loaded", () => {
```
brianpeiris: I think this gets called multiple times if the environment has multiple layers. Not sure if you want to debounce queryCollisionEntities, if it's expensive.

netpro2k: Looks like we have a `bundleloaded` event fired by `gltf-bundle` after all the `model-loaded` events have fired

PR #183, opened 2018-04-11
Append all child elements with while loop.

### LINKED ISSUE #171: Fix `gltf-model-plus` to support multiple child entities in templates
As discovered in https://github.com/mozilla/mr-social-client/pull/158, `gltf-model-plus` only seems to inflate the first child entity in a template.

johnshaughnessy:
Fixes #171


netpro2k:
Ah yeah, appending the child is removing it from children, good catch.

PR #190, opened 2018-04-12
Bug/avatar rotation direction

johnshaughnessy:


src/react-components/avatar-selector.js
```diff
@@ -20,8 +20,8 @@ class AvatarSelector extends Component {
     const numAvatars = this.props.avatars.length;
     return ((currAvatarIndex + direction) % numAvatars + numAvatars) % numAvatars;
   };
-  nextAvatarIndex = () => this.getAvatarIndex(1);
-  previousAvatarIndex = () => this.getAvatarIndex(-1);
+  nextAvatarIndex = () => this.getAvatarIndex(-1);
+  previousAvatarIndex = () => this.getAvatarIndex(1);
```
johnshaughnessy: I admit that it's awkward to say that the "nextAvatarIndex" is in the negative direction. Fundamentally the awkwardness comes from the fact that avatars are laid out in the scene counter-clockwise and when viewing them, "right" is "next" but also in the "negative" direction.

PR #197, opened 2018-04-13
Stats panel component that collapses to an FPS counter

### LINKED ISSUE #191: Make stats UI start minimized into a FPS counter
The stats UI is in the way and doesn't need to be visible in production. Showing the FPS at all times is probably useful in this preview period. Users will be able to quantify low performance instead of just reporting that things are slow on their device.

Clicking on the FPS counter will expand the UI.

robertlong:
Modifies the original `stats` component to start collapsed as a FPS counter and expand to the full stats panel when its property is set to true or the FPS counter is clicked. The stats panel collapses back to the FPS counter when the "X" is clicked.

Fixes #191 

## Collapsed Desktop:
desktop-collapsed

## Expanded Desktop
desktop-expanded

## Collapsed Mobile
mobile-collapsed

## Expanded Mobile
mobile-expanded

### DIFF
--- a/src/components/stats-plus.js
+++ b/src/components/stats-plus.js
@@ -1,0 +1,130 @@
+import "./stats-plus.css";
+// Adapted from https://github.com/aframevr/aframe/blob/master/src/components/scene/stats.js
+
+function createStats(scene) {
+  const threeStats = new window.threeStats(scene.renderer);
+  const aframeStats = new window.aframeStats(scene);
+  const plugins = scene.isMobile ? [] : [threeStats, aframeStats];
+  return new window.rStats({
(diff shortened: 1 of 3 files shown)


src/components/stats-plus.js
```diff
@@ -0,0 +1,130 @@
+import "./stats-plus.css";
```
brianpeiris: This "-plus" thing has turned into an anti-pattern IMO. We really should be contributing back to our dependencies, instead of just copy-pasting them. At the very least, we should put these changes in our fork of aframe instead of bleeding it into the client.

robertlong: Sounds good, I'll get some feedback and contribute it back or split it into its own component.

Issue #91, opened 2018-03-25
Disable microphone on mobile when navigating away from app


gfodor:
Currently on mobile, the microphone is left enabled when you navigate away from the app. This is good behavior in apps where you are having a conference call and multitasking, however I don't think its the same for our app.

We should disable the microphone or disconnect the user from Networked AFrame when navigating away from the application on mobile. We can probably use window.onblur on mobile only.

This is maybe desirable on mobile, but definitely not on desktop when screen sharing.

netpro2k:
Biggest concern for me here would be false positives to due with the VR entry flow, i'm sure the browser loses focus for variable ammounts of time as part of that.

robertlong:
What about a 10 second grace period before disconnecting?

We set a timeout for 10 seconds on any blur event on mobile. If we regain focus within 10 seconds we cancel the timeout.

Maybe 10 seconds is too short? Sometimes I lose my daydream controller and have to go find it. In that case you'll see the exited screen which will tell you to refresh your browser. That seems like good enough UX for now. It's the most privacy sensitive option but does degrade the UX if you can't enter VR within the grace period.

PR #198, opened 2018-04-13
Exit scene on blur with grace period

robertlong:
Tears down the aframe scene and stops microphone input when navigating away from the tab on a mobile device. Allows for a 10 second grace period to enter VR etc.

Exit screen styling depends on changes in: https://github.com/mozilla/mr-social-client/pull/177

Fixes: #91


src/hub.js
```diff
@@ -113,6 +114,13 @@ concurrentLoadDetector.start();
 store.update({ profile: { ...generateDefaultProfile(), ...(store.state.profile || {}) } });
 
 async function exitScene() {
+  if (NAF.connection.adapter && NAF.connection.adapter.localMediaStream) {
+    const tracks = NAF.connection.adapter.localMediaStream.getTracks();
```
netpro2k: very minor but since this isnt perf sensitive code at all id use forEach
`NAF.connection.adapter.localMediaStream.getTracks().forEach(t => t.stop())`

PR #208, opened 2018-04-17
Add mechanism to exit scene and show message for closed or full rooms

gfodor:
Goes with https://github.com/mozilla/reticulum/pull/18

This PR:

- Pushes a bunch of functions in `hub.js` into local bindings in `onReady`
- Adds logic to show message + exit when a hub is full or closed


src/hub.js
```diff
@@ -255,19 +148,135 @@ function mountUI(scene, props = {}) {
...
+
+    if (!qsTruthy("offline")) {
+      document.body.addEventListener("connected", () => {
+        hubChannel.sendEntryEvent().then(() => {
+          store.update({ lastEnteredAt: moment().toJSON() });
+        });
+      });
+
+      scene.components["networked-scene"].connect().catch(connectError => {
+        // hacky until we get return codes
```
gfodor: actual change is here, other code is largely just moved around

netpro2k: Looks like the actual text is "Room is full." https://github.com/mozilla/janus-plugin-sfu/blob/master/src/lib.rs#L440

src/react-components/ui-root.js
```diff
@@ -499,33 +500,35 @@ class UIRoot extends Component {
...
       return (
         <IntlProvider locale={lang} messages={messages}>
-          <div className="loading-panel">
-            <div className="loader-wrap">
-              <div className="loader">
-                <div className="loader-center" />
-              </div>
-            </div>
+          <div className="exited-panel">
             <div className="loading-panel__title">
```
netpro2k: Are these supposed to be `exited-panel` or are we just re-using the loading-panel styles? I think for proper BEM you should alias them in the css, but this is super minor.

src/hub.js
mquander: brb, putting in an error that says "network connection too full, can't send packets"

brianpeiris: Worth noting this is dependent on https://github.com/mozilla/naf-janus-adapter/pull/31

brianpeiris: Also dependent on https://github.com/mozilla/mr-social-client/pull/207

src/react-components/ui-root.js
gfodor: I wrote a separate bug for this, will deal with it at some point

PR #212, opened 2018-04-18
Freeze Mode and HUD restyle

netpro2k:
- Update visual style of both 2d and in world HUD with new icons
- Add freeze mode toggle (replaces avatar icon)
- Add space bubble toggle (replaces right mute button)
- Adjust hud-controller heuristics a bit to keep HUD locked in place once you have looked at it until you look further down
- HUD remains in the fully down position until the user toggles frozen state at least once to aid in discoverability

TODO:
- [x] wire up space bubble toggle
- [x] test hover events with new cursor stuff (currently hover and click wont work without it)


src/react-components/2d-hud.js
```diff
@@ -2,25 +2,28 @@ import React from "react";
...
-      <div className={styles.avatar} />
+      <div className={cx(styles.iconButton, styles.mute, { [styles.active]: muted })} onClick={onToggleMute} />
     </div>
+    <div
+      className={cx("ui-interactive", styles.iconButton, styles.large, styles.freeze, { [styles.active]: frozen })}
+      onClick={onToggleFreeze}
+    />
     <div className={cx("ui-interactive", styles.panel, styles.right)}>
-      <div className={cx(styles.mic, { [styles.muted]: muted })} onClick={onToggleMute} />
+      <div className={cx(styles.iconButton, styles.bubble, { [styles.active]: muted })} onClick={onToggleMute} />
```
johnshaughnessy: Should toggle space bubble, not mute state.

src/storage/store.js
```diff
@@ -19,6 +19,7 @@ export const SCHEMA = {
       properties: {
         has_agreed_to_terms: { type: "boolean" },
         has_changed_name: { type: "boolean" },
+        has_found_freeze: { type: "boolean" },
```
gfodor: small thing, but I'd probably move these boolean flags outside of `profile` -- the `profile` object was intended to be the user's profile like name and avatar. the terms/name ones we might be forced to leave for backwards compatibility.

src/App.js
```diff
@@ -1,7 +1,10 @@
+import Store from "./storage/store";
+
 export class App {
   constructor() {
     this.scene = null;
     this.quality = "low";
+    this.store = new Store();
```
gfodor: should we try to make this a singleton that is passed in? we currently create one at the root of `hub.js`

PR #214, opened 2018-04-18
Add block buttons on avatars that show during freeze mode

johnshaughnessy:
Adds block UI to the freeze mode.


package.json
```diff
@@ -38,7 +39,7 @@
     "moment": "^2.22.0",
     "moment-timezone": "^0.5.14",
     "moving-average": "^1.0.0",
-    "naf-janus-adapter": "^0.5.2",
+    "naf-janus-adapter": "https://github.com/mozilla/naf-janus-adapter#feature/blocking",
```
netpro2k: Before landing this PR we should actually just get `naf-janus-adapter` deployed to npm with the new changes and use a version number

johnshaughnessy: See https://github.com/mozilla/naf-janus-adapter/pull/35

src/components/block-button.js
```diff
@@ -0,0 +1,31 @@
+window.blockUser = function(id) {
```
netpro2k: I assume these were just global for debugging? Should make them local.

src/components/block-button.js
```diff
@@ -0,0 +1,31 @@
...
+  NAF.connection.adapter.publisher.handle.sendMessage({ kind: "block", whom: id });
```
netpro2k: Hmm I think it's fine that we implement blocking as a proprietary thing in our adapter, but we shouldnt be reaching into internal datastructures of the adapter or be aware of janus protocol to do it, we should just expose a `block(clientId)` function on the adapter that does this internally.

src/components/block-button.js
```diff
@@ -0,0 +1,31 @@
...
+  NAF.connection.entities.removeEntitiesOfClient(id);
```
netpro2k: The adapter's block funciton above should probably just emit an even so that it gets handled by the handler you already have doing this.

src/components/block-button.js
```diff
@@ -0,0 +1,31 @@
...
+};
+
+window.unblockUser = function(id) {
+  NAF.connection.adapter.publisher.handle.sendMessage({ kind: "unblock", whom: id });
```
netpro2k: same as above, should be moved to an `unblock(clientId)` function in the adapter.

src/components/block-button.js
```diff
@@ -0,0 +1,31 @@
...
+  window.setTimeout(() => {
+    // Wait enough time for the unblock to be processed
+    NAF.connection.entities.completeSync(id);
+  }, 1000);
```
netpro2k: Hmm this seems odd. Shouldn't janus send a response back for this? (I assume sendMesage would resolve with a promise). Also same as above shouldnt this be handled by the event listeners you have set up on `document.body`?

src/components/block-button.js
```diff
@@ -0,0 +1,31 @@
...
+};
+
+AFRAME.registerComponent("block-button", {
+  init() {
+    this.el.addEventListener("click", this.onClick);
+    const owner = (function getOwner(el) {
```
netpro2k: Use the NAF method to find the networked element for an entity like in https://github.com/networked-aframe/networked-aframe/blob/master/src/components/networked-audio-source.js#L22

src/components/block-button.js
```diff
@@ -0,0 +1,31 @@
...
+      if (el.firstUpdateData) {
+        return el.firstUpdateData.owner;
+      }
+      if (el.parentEl) {
+        return getOwner(el.parentEl);
```
netpro2k: This code goes away by using the NAF utility anway, but i'd be careful using recursion for walking up the tree like this cuz you could end up blowing up the stack. I know in es6 some sort of tail call optimizaiton was added but not too sure how/if it works with babel.

src/components/block-button.js
```diff
@@ -0,0 +1,31 @@
...
+  NAF.connection.adapter.publisher.handle.sendMessage({ kind: "unblock", whom: id });
+  window.setTimeout(() => {
+    // Wait enough time for the unblock to be processed
+    NAF.connection.entities.completeSync(id);
+  }, 1000);
+};
+
+AFRAME.registerComponent("block-button", {
+  init() {
+    this.el.addEventListener("click", this.onClick);
```
netpro2k: You need to do this aftaer assigning `this.onClick`, hoisting only applies to varaible declarations.

src/components/text-button.js
```diff
@@ -0,0 +1,55 @@
...
+      this.emitHapticPulse();
+    };
+    this.onHoverOut = () => {
+      this.hovering = false;
+      this.updateButtonState();
+    };
+    this.onClick = () => {
+      this.emitHapticPulse();
+    };
+    this.textEl = this.el.querySelector(this.data.textEl);
```
netpro2k: Man we really should push to add a relative selector type to aframe, we end up doing this way too often.

src/components/visible-while-frozen.js
```diff
@@ -0,0 +1,19 @@
+AFRAME.registerComponent("visible-while-frozen", {
+  init() {
+    this.onStateChange = evt => {
+      if (!evt.detail === "frozen") return;
```
netpro2k: I hate how common a pattern this is becoming for us, but it's all for the same reason and should be pretty easy to find and replace once we do something better so...

src/hub.html
```diff
@@ -90,7 +91,35 @@
...
                         <template data-selector=".Chest">
-                            <a-entity personal-space-invader="radius: 0.2; useMaterial: true;" bone-visibility></a-entity>
+                            <a-entity>
+                              <a-entity personal-space-invader="radius: 0.2; useMaterial: true;" bone-visibility> </a-entity>
+                              <a-entity billboard>
+                                  <a-entity
+                                      block-button
+                                      visible-while-frozen
+                                      text-button="haptic:#player-right-controller;
+                                                   textEl:.text;
```
netpro2k: You could use `[text]` selector here instead to avoid the class, not a big deal either way

src/hub.js
```diff
@@ -202,6 +206,14 @@ async function enterScene(mediaStream, enterInVR, janusRoomId) {
     screenEntity.setAttribute("visible", sharingScreen);
   });
 
+  document.body.addEventListener("blocked", ev => {
```
netpro2k: As dicussed, would be nice to make the adapter an event emitter and wire up event listeners on that instead, but this is probably fine for now.

src/hub.html
```diff
@@ -29,6 +29,7 @@
         >
 
         <a-assets>
+            <img id="tooltip"  src="./assets/hud/tooltip.png" >
```
netpro2k: Would be nice to indicate in the filename somehow that this is a 9slice. Android uses `.9.png`, so i'd do that unless you have something better.

package.json
johnshaughnessy: 👍 Sounds good. I'll use yarn link in the meantime.  I changed it here so no one else had to set up the link to test it, but I'll change it back.

src/components/block-button.js
johnshaughnessy: Oh, yes - `sendMessage` returns a promise. Will use that instead of this timeout.
No it shouldn't be handled by the event listener because when client A unblocks client B, only client B receives the "unblocked" message that is being handled by the event listeners.

src/components/block-button.js
johnshaughnessy: 👍

src/components/block-button.js
johnshaughnessy: 👍

src/components/block-button.js
johnshaughnessy: Oh, I misunderstood your earlier suggestion. Yes, the event listeners can handle it for both.

src/components/block-button.js
johnshaughnessy: 👍

src/components/block-button.js
johnshaughnessy: 👍

src/components/block-button.js
johnshaughnessy: noted. I don't want to blow up the stack

src/hub.html
johnshaughnessy: 👍

src/hub.html
johnshaughnessy: oh yea, I forgot about that

src/hub.js
johnshaughnessy: 👍

PR #217, opened 2018-04-18
Update to use aframe CDN (DO NOT MERGE)

gfodor:
Switches to aframe cdn 0.8.2

TODO: test a bunch


brianpeiris:
We now point to a SHA on A-Frame `master` with #367. Would be prudent to switch to the official A-Frame CDN on their next release if we can.

PR #219, opened 2018-04-19
Set up dev server

gfodor:
This PR rigs things up so `yarn start` will use `dev.reticulum.io` for its API and Phoenix Channels domain.

Note: for it to work you will need to hit `https://localhost:8080`, if you need to test on another origin you will need to set up your own reticulum and pass `phx_host` and `phx_port` to `hub.html`.

This also turns off the annoying "open up a new browser window when I run start" thing

### DIFF
--- a/src/react-components/hub-create-panel.js
+++ b/src/react-components/hub-create-panel.js
@@ -44,7 +44,7 @@ class HubCreatePanel extends Component {
       createUrl = `https://dev.reticulum.io${createUrl}`;
     }
 
-    const res = await fetch("https://dev.reticulum.io/api/v1/hubs", {
+    const res = await fetch(createUrl, {
       body: JSON.stringify(payload),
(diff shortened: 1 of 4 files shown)


src/react-components/hub-create-panel.js
```diff
@@ -38,14 +38,25 @@ class HubCreatePanel extends Component {
       hub: { name: this.state.name, default_environment_gltf_bundle_url: environment.bundle_url }
     };
 
-    const res = await fetch("/api/v1/hubs", {
+    let createUrl = "/api/v1/hubs";
+
+    if (process.env.NODE_ENV === "development") {
+      createUrl = `https://dev.reticulum.io${createUrl}`;
```
brianpeiris: Maybe use an ENV var for dev.ret. Same above as well.

PR #231, opened 2018-04-21
check the nav-mesh on teleport

InfiniteLee:
Sometimes won't teleport you if you're trying to teleport far (e.g. very high elevation change), but at least won't teleport you into the air.


netpro2k:
Hmm, do we know why the issue this is attempting to work around happens? `teleport-controls` is supposed to position you vertically at the point it collided with, if its not that's a bug. Would be good to know why/when it happens.

PR #207, opened 2018-04-16
Reconnection

robertlong:
Reconnect to Janus when the websocket unexpectedly closes.

Includes the following PRs:
minijanus: https://github.com/mozilla/minijanus.js/pull/6
naf-janus-adapter: https://github.com/mozilla/naf-janus-adapter/pull/33
networked-aframe: https://github.com/networked-aframe/networked-aframe/pull/125

fixes #82

PR #232, opened 2018-04-21
UX Changes - Part 3

### LINKED ISSUE #116: Entry flow UX changes based on feedback

brianpeiris:
Closes #116. Depends on #208, #207 and https://github.com/mozilla/reticulum/pull/21

- Add footer and menu for hub name, participant stats, invite UI, issue report UI
- Fix home page submit behavior
- Preload environment images


gfodor:
#208 merged

src/react-components/footer.js
```diff
@@ -0,0 +1,89 @@
...
+import faTimes from "@fortawesome/fontawesome-free-solid/faTimes";
+
+import styles from "../assets/stylesheets/footer.scss";
+
+const mobiledetect = new MobileDetect(navigator.userAgent);
+
+export default class Footer extends Component {
+  static propTypes = {
+    hubName: PropTypes.string,
+    participantCount: PropTypes.number,
```
gfodor: maybe `occupantCount` for consistency with janus adapter?

src/react-components/footer.js
```diff
@@ -0,0 +1,89 @@
...
+    onClickReport: PropTypes.func
+  };
+  state = {
+    menuVisible: false
+  };
+  render() {
+    const menuVisible = this.state.menuVisible;
+    return (
+      <div className={styles.container}>
+        {mobiledetect.mobile() ? (
```
gfodor: I think we should change this to use media queries based upon screen width, sorry I wasn't more clear in the mockup

src/react-components/home-root.js
```diff
@@ -242,23 +134,23 @@ class HomeRoot extends Component {
                     className="footer-content__links__link"
                     rel="noopener noreferrer"
                     href="#"
-                    onClick={this.showDialog("slack")}
+                    onClick={this.showDialog(dialogTypes.slack)}
```
gfodor: 👍

src/react-components/hub-create-panel.js
```diff
@@ -64,21 +83,28 @@ class HubCreatePanel extends Component {
     return new RegExp(HUB_NAME_PATTERN).test(this.state.name) && new RegExp(hubAlphaPattern).test(this.state.name);
   };
 
-  setToEnvironmentOffset = offset => {
+  _preloadImage = async src => {
```
gfodor: probably worth renaming to `srcset` to avoid confusion

src/react-components/hub-create-panel.js
```diff
@@ -64,21 +83,28 @@ class HubCreatePanel extends Component {
...
+
+  setToEnvironmentOffset = async offset => {
     const numEnvs = this.props.environments.length;
 
-    this.setState(state => ({
-      environmentIndex: ((state.environmentIndex + offset) % this.props.environments.length + numEnvs) % numEnvs
-    }));
+    const environmentIndex = ((this.state.environmentIndex + offset) % numEnvs + numEnvs) % numEnvs;
+    const environmentThumbnail = this._getEnvironmentThumbnail(environmentIndex);
+    await this._preloadImage(environmentThumbnail.srcset);
```
gfodor: If I'm reading this right, this blocks any visible changes to the UI when the next/prev button is hit until the image loads? (All our thumbnails are the same now so can't readily test.) I'm not sure that's better than showing a placeholder and letting the image load in, since users may think something is broken.

gfodor: If we're worried about it, i'd suggest adding an async prefetcher to load the rest of the images optimistically once you flip `ready: true` -- that way most likely they'll be already loaded but if not the user will still be able to cycle the environments without being blocked if they're not fetched yet. (of course you can get even smarter by loading just the neighboring images, but seems overkill until we have a lot of choices.)

brianpeiris: Alright, changed it to optimistically load all thumbnails now. It delays displaying the hub creation form entriely. We'll have to get a sense for it with actual images in production.

PR #236, opened 2018-04-22
Clean up schema of local storage, remove ToS check

gfodor:
Changes the local storage schema to:

- Roll over to new storage key
- Have nested categories at the root (activity/settings/profile)
- Use camelCase
- Remove the unnecessary/unused user id from storage for privacy reasons

Also removes unnecessary ToS checkbox now that legal has approved it.


gfodor:
Pushed via https://github.com/mozilla/hubs/pull/241

PR #237, opened 2018-04-22
Move hand-pose state out of an inflated gltf-model-plus entity.

johnshaughnessy:
This PR moves the hand pose state onto a component on the entity with the `gltf-model-plus` component, thus side-stepping the race condition caused by trying to network state on components that are created by `gltf-model-plus`, possibly _after_ the first network update.

### DIFF
--- a/src/components/hand-poses.js
+++ b/src/components/hand-poses.js
@@ -26,7 +26,7 @@ AFRAME.registerComponent("hand-pose", {
     this.from.play();
 
     const getNetworkedAvatar = el => {
-      let networkedAvatar = el.components["networked-avatar"];
+      const networkedAvatar = el.components["networked-avatar"];
       if (networkedAvatar) {
         return networkedAvatar;
       }
--- a/src/components/hand-poses.js
+++ b/src/components/hand-poses.js
@@ -26,7 +26,7 @@ AFRAME.registerComponent("hand-pose-state", {
     this.el.removeEventListener("model-loaded", this.setSelfAsStore);
   },
   setSelfAsStore(e) {
-    let poseEl = e.target.querySelector(`[hand-pose__${this.id}]`);
+    const poseEl = e.target.querySelector(`[hand-pose__${this.id}]`);
     poseEl.components[`hand-pose__${this.id}`].store = this;
   }
 });
--- a/src/components/hand-poses.js
+++ b/src/components/hand-poses.js
@@ -11,26 +11,6 @@ const POSES = {
 
 const NETWORK_POSES = ["allOpen", "thumbDown", "indexDown", "mrpDown", "thumbsUp", "point", "allGrip", "pinch"];
 
-AFRAME.registerComponent("hand-pose-state", {
-  multiple: true,
-  schema: {
-    pose: { default: 0 }
-  },
-  init() {
-    this.setSelfAsStore = this.setSelfAsStore.bind(this);
-  },
-  play() {
-    this.el.addEventListener("model-loaded", this.setSelfAsStore);
-  },
-  pause() {
-    this.el.removeEventListener("model-loaded", this.setSelfAsStore);
-  },
-  setSelfAsStore(e) {
-    const poseEl = e.target.querySelector(`[hand-pose__${this.id}]`);
-    poseEl.components[`hand-pose__${this.id}`].store = this;
-  }
-});
-
 AFRAME.registerComponent("hand-pose", {
   multiple: true,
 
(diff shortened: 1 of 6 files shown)


src/components/hand-poses.js
```diff
@@ -55,7 +75,8 @@ AFRAME.registerComponent("hand-pose", {
 AFRAME.registerComponent("hand-pose-controller", {
   multiple: true,
   schema: {
-    eventSrc: { type: "selector" }
+    eventSrc: { type: "selector" },
+    store: { type: "selector" }
```
gfodor: I'd probably call this something else, we already use the term `store` in the UI code and elsewhere for the local storage state

src/components/hand-poses.js
```diff
@@ -28,9 +46,11 @@ AFRAME.registerComponent("hand-pose", {
     this.from.play();
   },
 
-  update(oldData) {
-    if (oldData.pose != this.data.pose) {
-      this.animatePose(NETWORK_POSES[oldData.pose || 0], NETWORK_POSES[this.data.pose]);
+  tick() {
+    if (!this.store) return;
+    if (this.store.data.pose != this.pose) {
```
gfodor: `!==` ?

gfodor: Also I'd probably just do:
`if (!this.store || !this.store.data || this.store.data.pose !== this.pose) return;`

src/components/hand-poses.js
```diff
@@ -11,13 +11,31 @@ const POSES = {
...
+  },
+  play() {
+    this.el.addEventListener("model-loaded", this.setSelfAsStore);
+  },
+  pause() {
+    this.el.removeEventListener("model-loaded", this.setSelfAsStore);
+  },
+  setSelfAsStore(e) {
+    let poseEl = e.target.querySelector(`[hand-pose__${this.id}]`);
+    poseEl.components[`hand-pose__${this.id}`].store = this;
```
gfodor: I'm confused by this, can't the hand pose just look up this element itself via its selector?

gfodor:
Def want others feedback on this, but I was advocating for a *single* network-dedicated behavior on the root entity. So a "networked-avatar" behavior, with l/r poses as its schema.

src/components/hand-poses.js
```diff
@@ -11,13 +11,31 @@ const POSES = {
...
+  },
+  play() {
+    this.el.addEventListener("model-loaded", this.setSelfAsStore);
+  },
+  pause() {
+    this.el.removeEventListener("model-loaded", this.setSelfAsStore);
+  },
+  setSelfAsStore(e) {
+    const poseEl = e.target.querySelector(`[hand-pose__${this.id}]`);
+    poseEl.components[`hand-pose__${this.id}`].store = this;
```
gfodor: it seems that there should be a way to avoid this component having to set this back reference -- ideally it shouldn't care about who is asking it for its state. can't the pose behavior be given a selector to use to look this up at startup time and then just stash the reference to it?

johnshaughnessy: This was tricky so I wasn't sure what's right here. I couldn't think of how to use querySelector on hand-pose to set the back reference, since I don't know a unique selector each networked-instantiated avatar.
One way is to check for the `networked-avatar` behavior on `this.el`, then on its parentEl, then continue up the parent hierarchy until I find it. Do you think this preferable over making the `networked-avatar` (here it's `hand-pose-state` but that will change in my next commit) set the back reference?

src/components/networked-avatar.js
```diff
@@ -0,0 +1,6 @@
+AFRAME.registerComponent("networked-avatar", {
```
johnshaughnessy: This can probably be the same thing as `player-info`. I made it its own thing because `player-info` runs a routine to set the namatag text and the src of the gltf-model-plus every time its attributes change, and these hand poses change all the time. When they're combined, we'll only do those things if those were the attributes that actually changed.

src/components/hand-poses.js
johnshaughnessy: 👍

src/components/hand-poses.js
johnshaughnessy: 👍

src/components/hand-poses.js
johnshaughnessy: I crawled up the parent hierarchy to set the back reference on `hand-pose` because I don't know of a selector to use that will work in the case of network-instantiated players.

gfodor: some quick googling yielded this (I didn't know about it either) -- perhaps the contract can be you pass a selector in and then call `closest()` on the node once it is initialized to find the source of the data?

https://developer.mozilla.org/en-US/docs/Web/API/Element/closest

src/components/networked-avatar.js
gfodor: Also could be useful to split these up for reliable vs unreliable channel delivery. I don't know enough about NAF to know if that's the right separation

gfodor:
lgtm! would like others' input though because I am very out of my element with this stuff. also travis checks are failling

PR #238, opened 2018-04-22
Feature/startup interstitial

gfodor:
This PR:

- Adds an easy to skip interstitial on first launch with blank local storage to prompt the user to sign up for our release notes list.

- Adds the ToU and PP to the two main entry flows.


src/react-components/info-dialog.js
```diff
@@ -140,14 +156,20 @@ class InfoDialog extends Component {
...
-                <label className="mailing-list-form__privacy">
+                <label
+                  className={classNames({
+                    "mailing-list-form__privacy": true,
+                    "mailing-list-form__privacy--hidden":
+                      this.state.mailingListEmail.length === 0 && this.props.allowEmailSkip
+                  })}
+                >
                   <input
                     className="mailing-list-form__privacy_checkbox"
```
brianpeiris: Maybe use the `%checkbox` styling, for consistency with the rest of the UI? Might have to scale it down to match the text though.

src/react-components/home-root.js
```diff
@@ -154,6 +154,22 @@ class HomeRoot extends Component {
                   >
                     <FormattedMessage id="home.report_issue" />
                   </a>
+                  <a
+                    className="footer-content__links__link"
+                    target="_blank"
+                    rel="noopener noreferrer"
+                    href="https://github.com/mozilla/hubs/blob/master/TERMS.md"
```
brianpeiris: Are we actually just going to link to github for the notices? We ought to be able to generate some styled HTML from the markdown at least.

brianpeiris:
The screen looks very dark and void-like with the interstitial up. Maybe a white/grey border around the dialog would help?

image

src/react-components/home-root.js
gfodor: Yeah this is what was suggested by legal -- not 100% sure why but seems good to have a ground truth for now

gfodor:
Abandoned in favor of more user explicit choice to share email.

PR #240, opened 2018-04-22
Feature/get updates ux

gfodor:
This PR:

- Cleans up avatar picker animations + loading
- Adds a Help dialog
- Adds the Share and Help buttons to the initial entry dialog
- Adds the Help and Get Updates dialog to the in-game footer menu
- Bumps onblur timeout to end session to 60 seconds
- Removes outlines from buttons on mobile


src/react-components/footer.js
```diff
@@ -57,6 +62,22 @@ export default class Footer extends Component {
...
+                </span>
+              </button>
+              <button className={styles.menuButton} onClick={this.props.onClickUpdates}>
+                <i className={styles.menuButtonIcon}>
+                  <FontAwesomeIcon icon={faNewspaper} />
+                </i>
+                <span className={styles.menuButtonText}>
+                  <strong>Sign Up</strong> for Updates
+                </span>
+              </button>
```
brianpeiris: This makes the menu too tall in landscape on an S5. The close button gets hidden above the top of the screen.

src/react-components/profile-info-header.js
```diff
@@ -1,19 +1,35 @@
...
 export const ProfileInfoHeader = props => (
   <div className="profile-info-header">
-    <img src="../assets/images/account.svg" onClick={props.onClick} className="profile-info-header__icon" />
-    <div className="profile-info-header__profile_display_name" onClick={props.onClick} title={props.name}>
-      {props.name}
+    <div className="profile-info-header__profile_display_name">
+      <img src="../assets/images/account.svg" onClick={props.onClickName} className="profile-info-header__icon" />
+      <div onClick={props.onClickName} title={props.name}>
+        {props.name}
+      </div>
```
brianpeiris: Long names and the account image are right-aligned and cut off now.
image

src/react-components/info-dialog.js
```diff
@@ -183,11 +185,45 @@ class InfoDialog extends Component {
...
+              The <b>Mic Toggle</b> mutes your mic.
+            </p>
+            <p>
+              The <b>Pause/Resume Toggle</b> pauses all other avatars. You can then block them from having further
+              interactions with you.
+            </p>
+            <p>
+              The <b>Bubble Toggle</b> hides avatars that enter your personal space.
+            </p>
+          </div>
```
brianpeiris: Help content is too tall for iPhone 5/SE and Galaxy S5

src/react-components/footer.js
```diff
@@ -29,7 +34,7 @@ export default class Footer extends Component {
           </div>
           <button className={styles.menuButton} onClick={() => this.setState({ menuVisible: !menuVisible })}>
             <i className={styles.menuButtonIcon}>
-              <FontAwesomeIcon icon={menuVisible ? faTimes : faEllipsisH} />
+              <FontAwesomeIcon icon={menuVisible ? (window.innerWidth < 768 ? faTimes : faArrowDown) : faEllipsisH} />
```
brianpeiris: This could be done with a media query if you had two icons that you hide/show.

src/react-components/footer.js
```diff
@@ -57,6 +62,22 @@ export default class Footer extends Component {
                   <strong>Invite</strong> Others
                 </span>
               </button>
+              <button className={styles.menuButton} onClick={this.props.onClickHelp}>
+                <i className={styles.menuButtonIcon}>
+                  <FontAwesomeIcon icon={faQuestion} />
+                </i>
+                <span className={styles.menuButtonText}>
+                  <strong>Learn</strong> How to Play
```
brianpeiris: "Getting started", or "Learn the basics"

src/react-components/info-dialog.js
```diff
@@ -183,11 +185,45 @@ class InfoDialog extends Component {
           </span>
         );
         break;
+      case InfoDialog.dialogTypes.help:
+        dialogTitle = "How to Play";
```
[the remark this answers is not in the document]
gfodor: Yeah you're right, I think 'Getting Started' and 'Learn the Basics' works better.

src/react-components/info-dialog.js
```diff
@@ -183,11 +185,45 @@ class InfoDialog extends Component {
...
+                srcSet="../assets/images/help-hud@2x.png 2x"
+              />
+            </p>
+            <p>
+              The <b>Mic Toggle</b> mutes your mic.
+            </p>
+            <p>
+              The <b>Pause/Resume Toggle</b> pauses all other avatars. You can then block them from having further
+              interactions with you.
+            </p>
```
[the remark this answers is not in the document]
brianpeiris: Hmm, okay. Thanks for explaining it. I guess we'll have to see if we do end up going in that direction. I've gotta admit, "freeze" is a cool feature 😀

src/react-components/info-dialog.js
gfodor: Also worth mentioning that we can/should get a whole lot smarter about help text. For example, we should have specific copy for each device. So the next pass can probably address this problem since a lot of the copy on an S5 is irrelevant and can be removed since it's not VR capable, etc.

PR #246, opened 2018-04-23
Feature/touch events for cursor

johnshaughnessy:


src/components/cursor-controller.js
```diff
@@ -99,6 +109,11 @@ AFRAME.registerComponent("cursor-controller", {
   },
 
   pause: function() {
+    if (!this.inVR && this.isMobile && !this.hasPointingDevice) {
```
InfiniteLee: I think its safe to just always remove these listeners and not do this check. Doing this check might open us up to lingering eventListeners in the case that you happen to enterVR at some point later, then this pause is called.

src/components/cursor-controller.js
```diff
@@ -241,6 +256,75 @@ AFRAME.registerComponent("cursor-controller", {
...
+        this.data.cursor.object3D.position.copy(intersection.point);
+        this.currentDistance = intersections[0].distance;
+        this.currentDistanceMod = 0;
+      } else {
+        this.currentDistance = this.data.maxDistance;
+      }
+    }
+
+    const lookControls = this.data.camera.components["look-controls"];
+    if (lookControls) lookControls.pause();
```
InfiniteLee: these two lines are being called in 4 different places now so probably should be rolled into a helper function, but I can do that when I merge this in.

Issue #253, opened 2018-04-23
support non-English languages (l10n localisation, i18n internationalisation)


cvan:
**[`l20n`](http://l20n.org)** is a solid project (supported by the L10n team and community here at Mozilla) I've used with great success. Firefox, addons.mozilla.org, and various other projects use different L10n workflows (mostly because of historical reasons), but I'd recommend `l20n` for this project.

gfodor:
i can do a first pass on the parts of the app that are not currently localized because of the limitations of the existing plugin we use. if it works out we can use l20n going forward and try to do a burndown of the old stuff.

gfodor:
untagging this from polish MVP, but will make a pass on this for the next bulky text-oriented UI work

misslivirose:
We need to tag strings and an owner to help guide this along

piercechueng:
We have deployed Hubs Cloud on AWS out of USA and found that need non-english UI. So please consider this feature to be online soon. Thanks a lot.

misslivirose:
Going to resolve this in favor of #2373 since active discussion is happening on the newer issue

PR #271, opened 2018-04-24
Add data channel support detection

gfodor:
Shows message when Data Channels are not supported.


src/hub.js
```diff
@@ -308,8 +308,24 @@ const onReady = async () => {
     }
   };
 
+  const getPlatformUnsupportedReason = () => {
+    if (typeof RTCDataChannelEvent === "undefined") {
```
brianpeiris: Probably more correct to check `RTCPeerConnection.prototype.createDataChannel`

src/react-components/ui-root.js
```diff
@@ -540,6 +532,33 @@ class UIRoot extends Component {
...
+        // TODO i18n, due to links and markup
+        subtitle = (
+          <div>
+            Your browser does not support{" "}
+            <a href="https://developer.mozilla.org/en-US/docs/Web/API/RTCDataChannel" rel="noreferrer noopener">
+              WebRTC Data Channels
+            </a>, which is required to use Hubs.
+            <p />
+            You can <a href="https://firefox.com">Download Firefox</a>, a browser that protects your privacy and works
+            great with Hubs.
```
brianpeiris: Remove firefox promotion

src/react-components/ui-root.js
```diff
@@ -540,6 +532,33 @@ class UIRoot extends Component {
             <a href="https://github.com/mozilla/hubs">Github</a>.
           </div>
         );
+      } else if (this.props.platformUnsupportedReason === "no_data_channels") {
+        // TODO i18n, due to links and markup
+        subtitle = (
+          <div>
+            Your browser does not support{" "}
+            <a href="https://developer.mozilla.org/en-US/docs/Web/API/RTCDataChannel" rel="noreferrer noopener">
```
brianpeiris: Point to https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/createDataChannel#Browser_compatibility

src/react-components/ui-root.js
```diff
@@ -540,6 +532,33 @@ class UIRoot extends Component {
...
+              WebRTC Data Channels
+            </a>, which is required to use Hubs.
+            <p />
```
cvan: could this be two `<p>`s?

src/react-components/ui-root.js
cvan: I'd link to https://www.mozilla.org/firefox/

cvan: agreed that this seems odd to promote Firefox when Data Channels aren't supported, since Chrome, Edge, Safari, and Opera [all support](https://caniuse.com/#feat=rtcpeerconnection) WebRTC Data Channels.

src/react-components/ui-root.js
```diff
@@ -540,6 +532,33 @@ class UIRoot extends Component {
             <a href="https://github.com/mozilla/hubs">Github</a>.
           </div>
         );
+      } else if (this.props.platformUnsupportedReason === "no_data_channels") {
+        // TODO i18n, due to links and markup
```
cvan: filed issue #253 btw for L10n 📖

src/react-components/ui-root.js
```diff
@@ -540,6 +532,33 @@ class UIRoot extends Component {
             <a href="https://github.com/mozilla/hubs">Github</a>.
```
cvan: nit: `Github` -> `GitHub`

src/hub.js
cvan: sounds correct to me, based on this thread: https://wpdev.uservoice.com/forums/257854-microsoft-edge-developer/suggestions/17929639-rtc-data-channels

gfodor: MDN says this API is supported in Edge https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/createDataChannel#Browser_compatibility

brianpeiris: MDN is lying. I submitted a PR to fix that. https://github.com/mdn/browser-compat-data/pull/1940
You're right, the webrtc-adapter we use is polyfilling createDataChannel on Edge when it shouldn't. I guess we can check for the event class, since it works.

PR #282, opened 2018-04-25
Feature/duck polish

InfiniteLee:
A polish pass on stuff around interactables.

- Added sounds to duck spawner (`quack` component)
- `super-spawner` can be configured to limit how often an interactable can be spawned- the spawner will become hidden for the duration
- ducks have fun functionality when scaled (`duck` components)
- remote avatar heads now have a static-body
- this PR: https://github.com/mozilla/hubs/pull/257
- lowered gravity
- fixed issue with `super-hands` component not being on the controller entities. (`controls-shape-offset`)
- added velocity-based haptics while grabbing interactables
- updated `haptic-feedback` component to allow specifying specific values.


src/components/controls-shape-offset.js
```diff
@@ -0,0 +1,45 @@
...
+    this._handleControllerConnected = this._handleControllerConnected.bind(this);
+    this.el.addEventListener("controllerconnected", this._handleControllerConnected);
+  },
+
+  remove: function() {
+    this.el.removeEventListener("controllerconnected", this._handleControllerConnected);
+  },
+
+  tick: function() {
+    if (!this.shapeAdded && this.controller) {
```
InfiniteLee: this runs in tick because it needs to run after the `body` component's init has finished (but we also need to have the controllerconnected listener on init or otherwise you'll miss the event if your controllers are already connected when you start the app).

src/components/duck.js
```diff
@@ -0,0 +1,53 @@
+AFRAME.registerComponent("duck", {
+  schema: {
+    initialForce: { default: 0 },
+    maxForce: { default: 6.5 },
+    maxScale: { default: 5 }
+  },
+
+  init: function() {
+    this.system = this.el.sceneEl.systems.physics;
```
netpro2k: `this.system` is used by aframe for the system for the current component, we should use something different, I suggest `this.physicsSystem`

src/components/duck.js
```diff
@@ -0,0 +1,53 @@
...
+    if (this.el.body) {
+      const currentScale = this.el.object3D.scale.x;
+      const ratio = Math.min(1, (currentScale - this.initialScale) / (this.maxScale - this.initialScale));
+      const force = ratio * this.data.maxForce;
+      if (force > 0) {
+        const angle = Math.random() * Math.PI * 2;
+        const x = Math.cos(angle);
+        const z = Math.sin(angle);
+        this.force.set(x, force, z);
+        this.position.set(x * 0.01, 0, z * 0.01);
```
netpro2k: This line really confused me at first glance until I remembered that this is not the objects posotion but the position to apply the force from. Consuder renaming.

Also whats the `0.01`?

src/components/duck.js
```diff
@@ -0,0 +1,53 @@
...
+    maxScale: { default: 5 }
+  },
+
+  init: function() {
+    this.system = this.el.sceneEl.systems.physics;
+    this.hasBody = false;
+    this.position = new THREE.Vector3();
+    this.force = new THREE.Vector3(0, this.data.initialForce, 0);
+    this.initialScale = this.el.object3D.scale.x;
+    this.maxScale = this.data.maxScale * this.initialScale;
```
netpro2k: I dont get why w want to multiply this here?

src/components/duck.js
```diff
@@ -0,0 +1,53 @@
...
+
+  pause: function() {
+    this.system.removeComponent(this);
+    this.el.removeEventListener("grab-start", this._handleGrabStart);
+  },
+
+  beforeStep: function() {
+    if (this.el.body) {
+      const currentScale = this.el.object3D.scale.x;
+      const ratio = Math.min(1, (currentScale - this.initialScale) / (this.maxScale - this.initialScale));
```
netpro2k: Couldnt this just be `THREE.MATH.clamp(currentScale/this.data.maxScale, 0, 1)`

src/components/duck.js
```diff
@@ -0,0 +1,53 @@
...
+        const z = Math.sin(angle);
+        this.force.set(x, force, z);
+        this.position.set(x * 0.01, 0, z * 0.01);
+        this.el.body.applyForce(this.force, this.position);
+      }
+    }
+  },
+
+  _handleGrabStart: function() {
+    const rand = Math.floor(Math.random() * Math.floor(100));
```
netpro2k: Whats with the `Math.floor(100)`?

src/components/duck.js
```diff
@@ -0,0 +1,53 @@
...
+        const x = Math.cos(angle);
+        const z = Math.sin(angle);
+        this.force.set(x, force, z);
+        this.position.set(x * 0.01, 0, z * 0.01);
+        this.el.body.applyForce(this.force, this.position);
+      }
+    }
+  },
+
+  _handleGrabStart: function() {
```
netpro2k: This looks like it should just use the `quack` component.

src/components/haptic-feedback.js
```diff
@@ -37,21 +37,27 @@ AFRAME.registerComponent("haptic-feedback", {
...
   },
 
-  pulse: function(event) {
+  handlePulse: function(event) {
     const { intensity } = event.detail;
-    if (!strengthForIntensity[intensity]) {
+
+    if (strengthForIntensity[intensity]) {
+      this.pulse(strengthForIntensity[intensity]);
+    } else if (Number(intensity) === intensity) {
```
netpro2k: What necesitated adding this check?

src/components/super-spawner.js
```diff
@@ -51,6 +62,16 @@ AFRAME.registerComponent("super-spawner", {
     const pos = this.data.useCustomSpawnPosition ? this.data.spawnPosition : this.el.getAttribute("position");
     entity.setAttribute("position", pos);
     this.el.sceneEl.appendChild(entity);
+
+    if (this.data.spawnCooldown > 0) {
+      this.el.setAttribute("visible", false);
+      this.el.setAttribute("scale", { x: 0.0001, y: 0.0001, z: 0.0001 });
```
netpro2k: Why do you need to modify the scale?

src/hub.html
```diff
@@ -184,6 +195,10 @@
             super-spawner="template: #interactable-template;" 
             position="2.9 1.2 0" 
             body="mass: 0; type: static; shape: box;"
+            collision-filter="collisionForces: false;"
```
netpro2k: RIP duck cannon, you will be missed

netpro2k:
Also make sure you pull from master to fix the travis-ci error

src/components/duck.js
InfiniteLee: ah oops, I just forgot to remove all the `quack` code from `duck` (originally all the functionality was just on `duck`)

src/components/haptic-feedback.js
InfiniteLee: to verify its a number

src/components/super-spawner.js
InfiniteLee: the collision shapes are still there and "clickable", so this is to make it "physically" disappear

src/components/duck.js
InfiniteLee: The maxScale needs to be relative to the scale of the object set in the html.

src/components/duck.js
InfiniteLee: to limit how far from the center of the duck the force is applied (otherwise its applied from way outside of the duck causing it bounce around excessively)

src/components/duck.js
InfiniteLee: clamp can be used but initialScale still needs to be subtracted from both values before dividing or the math will be wrong

src/components/super-spawner.js
netpro2k: should probably just remove the interactable class

PR #286, opened 2018-04-25
Feature/help icon and attribution

gfodor:
- Repositions attribution on homepage to bottom right in hero
- Adds a persistent help icon to top right in 2d mode


gfodor:
Dom approved via Slack

PR #287, opened 2018-04-25
Avatar selector perf

### LINKED ISSUE #199: Avatar selector performance
The avatar selector spends a lot of time blocking the main thread while parsing each of the avatars. We ought to be able to re-use the bot mesh instead of re-parsing for each bot. Or perhaps use web workers to do the parsing.

brianpeiris:
- Only load current and adjacent avatars on load.
- Load subsequent avatars lazily
- Move avatar selector loader animation outside iframe

Closes #199


johnshaughnessy:
Very nice! Loads much faster than before.

PR #290, opened 2018-04-25
Bug/ui tweaks

brianpeiris:
- "Next" button is too low on mic confirmation page
- "Enter VR" button is too high
- Increase info dialog padding to 30px
- De-emphasize profile header buttons
- Mouse drag sometimes drags selected ui items
- Move fullscreen call to just before enter scene
- Avoid flashing mic-denied icon
- Avoid continuous re-render on mic level

### DIFF
--- a/src/assets/stylesheets/2d-hud.scss
+++ b/src/assets/stylesheets/2d-hud.scss
@@ -6,6 +6,7 @@
   align-items: center;
   height: 80px;
   width: 100%;
+  user-select: none;
 }
 
 :local(.panel) {
--- a/src/assets/stylesheets/audio.scss
+++ b/src/assets/stylesheets/audio.scss
@@ -1,10 +1,17 @@
 .audio-setup-panel {
   text-align: center;
   overflow-y: auto;
+  display: flex; 
+  flex-direction: column;
+  height: 100%;
   
+  &__enter-button-container {
+    flex: 1;
+    display: flex;
+    justify-content: center;
+  }
   &__enter-button {
     @extend %bottom-button;
-    margin: 32px auto;
   }
 
   &__title {
@@ -95,6 +102,16 @@
   flex: 10 1 auto;
   justify-content: flex-start;
   align-items: center;
+  overflow-y: auto;
+
+  &__grant-container {
+    flex: 1;
+    display: flex;
+    flex-direction: column;
+    justify-content: center;
+    align-items: center;
+    min-height: max-content;
+  }
 
   &__title {
     @extend %top-title;
@@ -105,13 +122,14 @@
   }
 
   &__button-container {
-    flex: 10;
+    flex: 5;
     display: flex;
     justify-content: center;
     align-items: center;
     cursor: pointer;
     width: 111px;
     height: 111px;
+    margin: 1em;
   }
 
   &__button {
@@ -120,10 +138,16 @@
     cursor: pointer;
   }
 
+  &__next-container {
+    display: flex;
+  }
+
   &__next {
     @extend %bottom-button;
-    padding-top: 0;
-    padding-bottom: 0;
     flex: 1 1;
+  }
+
+  .invisible {
+    visibility: hidden;
   }
 }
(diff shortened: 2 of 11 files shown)


src/react-components/ui-root.js
```diff
@@ -635,34 +635,41 @@ class UIRoot extends Component {
       ) : null;
 
     const micPanel =
-      this.state.entryStep === ENTRY_STEPS.mic_grant || this.state.entryStep == ENTRY_STEPS.mic_granted ? (
+      this.state.entryStep === ENTRY_STEPS.mic_grant || this.state.entryStep === ENTRY_STEPS.mic_granted ? (
         <div className="mic-grant-panel">
-          <div className="mic-grant-panel__title">
```
brianpeiris: Mostly indentation differences below this line. I'd recommend the white space insensitive diff: https://github.com/mozilla/hubs/pull/290/files?w=1

src/assets/stylesheets/2d-hud.scss
```diff
@@ -6,6 +6,7 @@
   align-items: center;
   height: 80px;
   width: 100%;
+  user-select: none;
```
gfodor: i think i've concluded that there is an infinite number of CSS attributes I do not know about

src/react-components/profile-entry-panel.js
```diff
@@ -87,6 +87,7 @@ class ProfileEntryPanel extends Component {
                 onFocus={e => e.target.select()}
                 onChange={e => this.setState({ displayName: e.target.value })}
                 required
+                spellCheck="false"
```
gfodor: looks like someone needs to install the ornithology pack to their system dictionary

src/react-components/ui-root.js
```diff
@@ -675,100 +682,104 @@ class UIRoot extends Component {
...
-                srcSet="../assets/images/speaker_level@2x.png 2x"
-                className="audio-setup-panel__levels__icon-part"
-              />
+                {this.state.audioTrack ? (
+                  <img
+                    src="../assets/images/mic_level.png"
+                    srcSet="../assets/images/mic_level@2x.png 2x"
+                    className="audio-setup-panel__levels__icon-part"
+                  />
+                ) : (
```
gfodor: It looks like something changed in this PR that causes me to see the "mic denied" image for a few frames before the audio track becomes non-null.

src/assets/stylesheets/profile.scss
```diff
@@ -101,6 +102,8 @@
       &__icon {
 	@extend %fa-icon-button-icon;
 	background: transparent;
```
gfodor: I was suggesting we drop the border altogether and leave the icon there fully-white -- the borders on these tiny icons are just very visually noisy. (I'm starting to think we should just drop them in the footer menu too, but I think they don't look too bad there.) The downside of removing the borders is just that the border has been the "affordance" (if you can call it that) that it's a button. But it kind of breaks down on these very small buttons.

gfodor: Actually after playing around with it, I think just changing this to have `color: $light-grey` looks pretty good

src/assets/stylesheets/audio.scss
```diff
@@ -1,10 +1,17 @@
...
   text-align: center;
   overflow-y: auto;
+  display: flex; 
+  flex-direction: column;
+  height: 100%;
   
+  &__enter-button-container {
+    flex: 1;
+    display: flex;
+    justify-content: center;
```
gfodor: I think all the grant and enter containers should have a bottom margin of 24px, right now the buttons are too flush with the bottom of the dialog imho

src/assets/stylesheets/profile.scss
gfodor: basically if you change `color: $light-grey` these buttons have a new previously unused visual style -- but I think its fine. they are effectively "secondary action buttons" -- I could imagine us re-using that elsewhere. (For example, we may want to restyle the "hamburger menu" button this way)

Issue #272, opened 2018-04-24
Character controller causes unnecessary network updates in some cases


brianpeiris:
- When you hit forward on your keyboard briefly, we send out about 20 network updates, most of which are minute changes that occur after the avatar has visibly stopped moving
- When turning with mouse or keyboard, we sometimes send network updates continuously where the avatar's y position is bouncing slightly
- In Firefox, we network hands even when you're not in VR mode, the hands often move slightly even when they're sitting stationary on a table.
See the video for a demo:
https://gfycat.com/MealyUniqueBarnswallow

johnshaughnessy:
Two PR's to address these issues. The first two are fixed by https://github.com/mozilla/hubs/pull/291
The last one is partially fixed by https://github.com/MozillaReality/networked-aframe/pull/6

We still network the hands even when you're not in VR mode, but this will at least stop them from sending updates when they move slightly while sitting on a table.

PR #291, opened 2018-04-25
Remove unnecessary network updates

johnshaughnessy:
Fixes https://github.com/mozilla/hubs/issues/272
To be merged after https://github.com/MozillaReality/networked-aframe/pull/11


src/components/character-controller.js
```diff
@@ -114,6 +114,9 @@ AFRAME.registerComponent("character-controller", {
       pivotRotationMatrix.makeRotationAxis(rotationAxis, pivot.rotation.y);
       pivotRotationInvMatrix.makeRotationAxis(rotationAxis, -pivot.rotation.y);
       this.updateVelocity(deltaSeconds);
+      if (this.velocity.x === 0 && this.velocity.y === 0 && this.velocity.z === 0) {
```
InfiniteLee: I would use the EPS constant (https://en.wikipedia.org/wiki/Machine_epsilon)
Actually since it's already being used down below, I would just cache if this.velocity.lengthSq() and use it in both places
```
const magnitude = this.velocity.lengthSq();
if (magnitude < EPS)

PR #295, opened 2018-04-25
Show temporary warning to gearvr users

gfodor:
Shows a warning to GearVR users that audio is not currently working


gfodor:
Fix is going out for Oculus Browser

PR #300, opened 2018-04-26
Fix prev/next buttons and background video on iOS

robertlong:


robertlong:
Added a .mp4 background video for iOS support and fixed the environment picker prev/next buttons

robertlong:
Also the whitespace in that file is messed up.

PR #303, opened 2018-04-26
improve Slack links in modal dialog windows

cvan:


src/react-components/info-dialog.js
```diff
@@ -174,12 +175,12 @@ class InfoDialog extends Component {
...
             <a href="https://github.com/mozilla/hubs/issues" target="_blank" rel="noopener noreferrer">
-              Github Issue
+              GitHub Issue
             </a>{" "}
             or e-mail us for support at <a href="mailto:hubs@mozilla.com">hubs@mozilla.com</a>.
             <p />
-            You can also find us in #social on the{" "}
-            <a href="http://webvr-slack.herokuapp.com/" target="_blank" rel="noopener noreferrer">
+            You can also find us in <a href="https://webvr.slack.com/messages/social">#social</a> on the{" "}
+            <a href="https://webvr-slack.herokuapp.com/" target="_blank" rel="noopener noreferrer">
```
gfodor: I think these new links should be `target="_blank"` -- we've made all of the links in open in new windows, since Hubs feels more like an application than a webpage, so user expectation is likely that they will not lose their tab/session/etc by clicking these links to 3rd party sites.

cvan: gotcha, that now makes more sense.

in that case, can we check for `navigator.standalone`? especially on desktop and mobile Safari, it can be quite the hassle since Safari jumps to the newly opened tab (from `_blank` or `window.open`); Chrome on mobile opens tabs in the background luckily.

to reduce the chance for error, what do you think of a lightweight React component for these external links that adds `rel="noopener noreferrer"` for a link.

and instead of `target="_blank"`, perhaps have a **click** event listener on `document.body`; check if the origin is external, and then open `window.open`.

DRY and reduces possible dev error. thoughts?

cvan: also would simplify the markup needed in the locale definitions

src/react-components/info-dialog.js
```diff
@@ -111,7 +111,8 @@ class InfoDialog extends Component {
             <a href="https://webvr-slack.herokuapp.com/" target="_blank" rel="noopener noreferrer">
               WebVR Slack
             </a>{" "}
-            in the #social channel.<br />VR meetups every Friday at noon PST!
+            in the <a href="https://webvr.slack.com/messages/social" target="_blank" rel="noopener noreferrer">#social</a> channel.<br />
+            VR meetups every Friday at noon PDT!
```
cvan: @gfodor: notice `PST` -> `PDT` (do you want to just say `Pacific Time`?)

cvan:
@gfodor ready for a re-review - thanks 👍

src/react-components/info-dialog.js
gfodor: ah yeah that seems right to me!

src/react-components/info-dialog.js
gfodor: sorry I missed this, I didn't know about `navigator.standalone` -- i think this seems like something to explore in general, probably worth a separate PR at some point that just changes this everywhere if we think it's a good idea

cvan:
@gfodor I just fixed some linting issues and fixed the paragraphs. when you get a min, can I get one more quick look?

src/assets/stylesheets/info-dialog.scss
```diff
@@ -79,6 +79,10 @@
   }
 }
 
+.dialog__box__contents__body p:first-child {
```
gfodor: we haven't been too crazy about this, but I think technically this breaks BEM -- probably should be a modifier class. your call though if you'd like to refactor!

gfodor:
Any reason you are working in a fork? Not sure if there is an easy way I'm not aware of, but makes it a bit more difficult to test since I need to clone your repo.

gfodor:
lgtm!

src/assets/stylesheets/info-dialog.scss
cvan: yes, good eye - it does. will fix.

cvan:
> Any reason you are working in a fork? Not sure if there is an easy way I'm not aware of, but makes it a bit more difficult to test since I need to clone your repo.

just habit from other projects. I'll start doing branches off the upstream `origin`. it certainly does make it easier.

src/assets/stylesheets/info-dialog.scss
cvan: the more I think about this, I think pseudo selectors could/should be the exception to the BEM pattern used throughout. let me know, and I can add a `margin-top-less` or something to each first `<p>`.

Issue #310, opened 2018-04-26
pause background video when tab/window loses focus


cvan:
instead of relying on the `autoplay` on the `<video>` element, it'd be nice to just play the video on page load (i.e., when the `HomeRoot` component is mounted), with logic to handle pausing/resuming in a `visibilitychange` event listener.

PR #312, opened 2018-04-26
handle playback of background video when page loses focus (fixes issue #310)

cvan:


gfodor:
LGTM! Lets sync on slack before merging so I can talk you through our deployment process. We perform continuous delivery of changes to master -- HEAD of master should always be what is live on the site.

cvan:
@gfodor: sounds good! will sync with you on Monday on Slack. (did some play testing yesterday with a high-school student I mentor and @caseyyee, so I am filing issues as I remember them so I don’t forget. we were all really impressed by the stability and comms, and spent a good time in it - all three of us.)

src/react-components/home-root.js
```diff
@@ -35,9 +35,39 @@ class HomeRoot extends Component {
...
+    this.loadHomeVideo();
   }
 
+  loadHomeVideo = () => {
+    const videoEl = document.querySelector("#background-video");
+    function initVideo() {
+      videoEl.playbackRate = 0.75;
+      videoEl.play();
+      let inIframe = false;
+      try {
```
netpro2k: Whats the deal with this?

src/react-components/home-root.js
```diff
@@ -35,9 +35,39 @@ class HomeRoot extends Component {
...
+      }
+      function toggleVideo() {
+        // Play the video if the window/tab is visible (or is within an `<iframe>`).
+        if (document.hasFocus() || inIframe) {
+          videoEl.play();
+        } else {
+          videoEl.pause();
+        }
+      }
+      document.addEventListener("visibilitychange", toggleVideo);
```
netpro2k: Probably fine but just out of curiosity why do we need both of these? What cases do focus and blur catch that visiblitychange does not (and vis versa)?

src/react-components/home-root.js
```diff
@@ -35,9 +35,39 @@ class HomeRoot extends Component {
...
+      window.addEventListener("focus", toggleVideo);
+      window.addEventListener("blur", toggleVideo);
+    }
+    if (videoEl.readyState >= videoEl.HAVE_FUTURE_DATA) {
```
cvan: I just have to take a second to lol at these state properties of [`HTMLMediaElement`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/readyState#Syntax)

src/react-components/home-root.js
cvan: also, from the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API#Use_cases):

> Note: While `blur` and `focus` will tell you if the user switches windows, it doesn't necessarily mean it's hidden. Pages only become hidden when the user switches tabs or minimizes the browser window containing the tab.

Issue #316, opened 2018-04-26
redirect static assets' URLs to hashed URLs


cvan:
it'd be nice if https://assets-prod.reticulum.io/assets/images/logo.svg (or https://hubs.mozilla.com/assets/images/logo.svg) redirected to the latest version of the asset (whose filename has the most recent content hash) -> 
https://assets-prod.reticulum.io/assets/images/logo-8f1dfd234d060073524bc29912a24f75.svg

this is useful for things like referencing URLs in `<meta>` tags (for Twitter/Facebook/Pinterest/etc.) and paths in `icons` + `screenshots` in a Web-App Manifest (adding in issue #274).

let me know if using Webpack plugins to rewrite the HTML + JSON is preferred. and I can close this issue, assuing there's not another use case for this.

gfodor:
We've avoided having non-digested URLs on cloudfront domains (to improve caching/remove the need for manual cache invalidation) so for fixed URLs as an explicit copy step and serve it directly from the ALB (which won't be edge cached.) This is what we do with `favicon.png`.

It's preferred however that we edge cache as much as we can, so using the digested filename + updating webpack to pack it in properly is probably the best way to do it.

gfodor:
Oh, and in terms of having non-digested URLs redirect to the digested variants, I don't think we want to do that because if we implemented that now, it would further couple reticulum with the client bundle (since reticulum would have to have redirection logic that incorporated a map of digest shas.) The mid-term plan is to remove the client from the reticulum habitat package and have a s3-backed origin, with server side templates for the root html files. That is gated on us moving the a-scene into react.

cvan:
thanks for the helpful notes. if I understand this correctly, I still am not sure how paths to assets will be defined in the HTML templates, SAAS stylesheets, web-app manifest, and JS files.

would this involve rewriting the URLs in reticulum? there is also a popular S3 plugin for Webpack: https://github.com/webpack-contrib/s3-plugin-webpack

PR #317, opened 2018-04-26
[WIP] add Web-App Manifest (fixes #274)

### LINKED ISSUE #274: create a W3C Web-App Manifest
In order for Chrome + Firefox for Android to detect a web page as a Progressive Web App (PWA), it needs a Web-App Manifest, a Service Worker (it can be empty to start), and served over HTTPS.

Here's an example of a typical manifest: https://w3c.github.io/manifest/#example-manifests

I can open a PR for this; I just wanted to see if there are vector favicons to use first. Also, if you have screenshots, you can send them to me or attach them to this issue.

cvan:
(slightly dependent on issue #316)


src/hub.html
```diff
@@ -294,9 +318,9 @@
                 hand-controls2="right"
                 tracked-controls
                 teleport-controls="
-                    cameraRig: #player-rig;
```
cvan: oops, my editor

src/hub.html
```diff
@@ -17,6 +19,28 @@
     <% } else { %>
         <script src="https://cdn.rawgit.com/brianpeiris/aframe/845825ae694449524c185c44a314d361eead4680/dist/aframe-master.js"></script>
     <% } %>
```
gfodor: We actually inject these tags server-side, since the contents are dynamic (right now, includes the name of the room) see: https://github.com/mozilla/reticulum/pull/23

We can update the reticulum code for any missing tags you've included here. Eventually some of these (like the thumbnail) will be informed by the current state of the room, such as the environment choice and number of CCU.

webpack.config.js
```diff
@@ -206,6 +206,18 @@ const config = {
         to: "favicon.ico"
       }
     ]),
+    new CopyWebpackPlugin([
```
gfodor: We'll need to whitelist these in reticulum here so they will be served statically from origin: https://github.com/mozilla/reticulum/blob/master/lib/ret_web/endpoint.ex#L16

If we want these CDN'ed, we'll need to digest them and presumably add some absolute URLs somewhere to point the browser to them (I'm too unfamililar with the spec to know how to do this offhand but I'm guessing it's possible?)

src/index.html
```diff
@@ -4,9 +4,31 @@
 <head>
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1">
-    <link rel="shortcut icon" type="image/png" href="/favicon.ico"/>
```
gfodor: Totally forgot to add these! Good thinking. (These don't require dynamic injection like the one listed above, obviously.)

src/assets/manifest.webmanifest
```diff
@@ -0,0 +1,25 @@
+{
+  "lang": "en",
+  "dir": "ltr",
+  "name": "Mozilla Hubs",
+  "description": "A new way to get together online. Laugh, play, get stuff done, or just hang out. A WebVR experiment by Mozilla Mixed Reality.",
+  "short_name": "Hubs",
+  "icons": [
+    {
+      "src": "https://assets-prod.reticulum.io/assets/images/icon.png",
```
gfodor: Ideally this would be digested, so it will get busted out of CDN if we change it.

src/hub.html
gfodor: Also worth noting that we plan on ripping out almost everything from this file and having it be rendered client side with react, at which point this will be just a stub and reticulum will have a proper templated server side view for the whole page.

src/assets/manifest.webmanifest
gfodor: (Since this is a slowly changing asset, and it might be quite complex to convince webpack to inject it properly here, I'm not opposed to us punting and just manually busting the CDN if it changes. Worth doing if its easy though.)

src/hub.html
cvan: oh thanks! I was wondering where; this makes more sense!

it's useful to have the HTML document's skeleton (and ideally critical-path assets, like inlined CSS and minimal JS to make the page interactive).

what about server-side rendering with React? just curious about the desire to move more to the client.

webpack.config.js
cvan: the URLs for any `src` value in the manifest (e.g., for `icons` and `screenshots`) can be local or external. local URLs are relative to the manifest URL (just FYI).

I'll follow up with this. I think I'll try [this small `webpack-pwa-manifest` Webpack plugin](https://github.com/arthurbergmz/webpack-pwa-manifest#webpack-pwa-manifest), and see how that works out. thanks for the helpful comments!

src/assets/manifest.webmanifest
cvan: most definitely. I'm going to try [`webpack-pwa-manifest`](https://github.com/arthurbergmz/webpack-pwa-manifest#webpack-pwa-manifest), so I don't reinvent the wheel with this.

src/hub.html
gfodor: I'm not sure what's involved to do that with Phoenix, my guess is that it is probably prohibitive since this is running on the BEAM VM. I agree that would be a good approach if we were running on a JS VM though.

gfodor: Oh, and the desire to move it all into the client is basically just to decouple our reticulum build/deploy process from the hubs project's build process. Once this is done there's a straight shot to being able to a) remove the coupling, so builds and deploys are faster and b) it cleans up the self-hosting story, since reticulum right now is coupled with the "burned in" asset URLs in the client code at package build time, so we can't just tell people to run our Habitat package to get things working.

This is being tracked here: https://github.com/mozilla/hubs/issues/347

PR #319, opened 2018-04-26
Fix i18n for non-english languages + UI tweaks

brianpeiris:
- Set language to "en" if no translations for user's language
- Avoid scrollbars on mailing list dialog
- Add webvr.rocks link to bottom of entry choices dialog on desktop


robertlong:
LGTM 👍

PR #323, opened 2018-04-26
Exit cleanly on blur

brianpeiris:
Fixes https://sentry.prod.mozaws.net/operations/hubs-prod/issues/4272342/

### DIFF
--- a/src/components/stats-plus.js
+++ b/src/components/stats-plus.js
@@ -123,8 +123,8 @@ AFRAME.registerComponent("stats-plus", {
     this.el.setAttribute(this.name, false);
   },
   remove() {
-    this.el.sceneEl.removeListener("enter-vr", this.hide);
-    this.el.sceneEl.removeListener("exit-vr", this.show);
+    this.el.sceneEl.removeEventListener("enter-vr", this.hide);
+    this.el.sceneEl.removeEventListener("exit-vr", this.show);
 
     if (this.statsEl) {
       this.statsEl.parentNode.removeChild(this.statsEl);
--- a/src/hub.js
+++ b/src/hub.js
@@ -195,10 +195,16 @@ const onReady = async () => {
     if (NAF.connection.adapter && NAF.connection.adapter.localMediaStream) {
       NAF.connection.adapter.localMediaStream.getTracks().forEach(t => t.stop());
     }
-    hubChannel.disconnect();
+    if (hubChannel) {
+      hubChannel.disconnect();
+    }
     const scene = document.querySelector("a-scene");
-    scene.renderer.animate(null); // Stop animation loop, TODO A-Frame should do this
-    document.body.removeChild(scene);
+    if (scene) {
+      if (scene.renderer) {
+        scene.renderer.animate(null); // Stop animation loop, TODO A-Frame should do this
+      }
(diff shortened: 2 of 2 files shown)


src/components/stats-plus.js
```diff
@@ -123,8 +123,8 @@ AFRAME.registerComponent("stats-plus", {
     this.el.setAttribute(this.name, false);
   },
   remove() {
-    this.el.sceneEl.removeListener("enter-vr", this.hide);
-    this.el.sceneEl.removeListener("exit-vr", this.show);
+    this.el.sceneEl.removeEventListener("enter-vr", this.hide);
```
cvan: is it ever possible with a race condition for `sceneEl` to not be existent but this component method gets called in a delay?

src/hub.js
```diff
@@ -195,10 +195,16 @@ const onReady = async () => {
     if (NAF.connection.adapter && NAF.connection.adapter.localMediaStream) {
       NAF.connection.adapter.localMediaStream.getTracks().forEach(t => t.stop());
```
cvan: does this cover clean up of `mediaStream`?

src/hub.js
```diff
@@ -195,10 +195,16 @@ const onReady = async () => {
...
+    if (hubChannel) {
+      hubChannel.disconnect();
+    }
     const scene = document.querySelector("a-scene");
-    scene.renderer.animate(null); // Stop animation loop, TODO A-Frame should do this
-    document.body.removeChild(scene);
+    if (scene) {
+      if (scene.renderer) {
+        scene.renderer.animate(null); // Stop animation loop, TODO A-Frame should do this
+      }
```
cvan: is there cleanup to be done for `registerInputActions`?

src/components/stats-plus.js
brianpeiris: Not as far as I know. At least not in usual operation. A-Frame lifecycle should be synchronous and it should be the only thing that calls remove() directly.

src/hub.js
brianpeiris: mediaStreams should be cleaned up by `networked-aframe` when the a-scene is removed from the DOM

src/hub.js
brianpeiris: In theory we could define an empty input mapping, but I'm not sure it would make much of a difference in terms resource usage. The input mapping system will stop ticking when the scene stops ticking anyway, which stops with the renderer loop.

src/hub.js
```diff
@@ -195,10 +195,16 @@ const onReady = async () => {
...
       NAF.connection.adapter.localMediaStream.getTracks().forEach(t => t.stop());
     }
-    hubChannel.disconnect();
+    if (hubChannel) {
+      hubChannel.disconnect();
+    }
     const scene = document.querySelector("a-scene");
-    scene.renderer.animate(null); // Stop animation loop, TODO A-Frame should do this
-    document.body.removeChild(scene);
+    if (scene) {
```
netpro2k: What s the case this and the next check are guarding against? Why would the scene not exist?

cvan: oh, that's why? I hadn't considered that. makes (the most) sense. (should a comment be added?)

PR #343, opened 2018-05-01
cardboard support for cursor-controller

InfiniteLee:
Properly show cursor when in cardboard. Note: cardboard input is broken until we upgrade to aframe 0.8.2.

### DIFF
--- a/src/components/cursor-controller.js
+++ b/src/components/cursor-controller.js
@@ -325,7 +325,6 @@ AFRAME.registerComponent("cursor-controller", {
           return;
         }
       }
-
     }
     this._setLookControlsEnabled(true);
     this.data.cursor.emit("cursor-release", {});
--- a/src/components/cursor-controller.js
+++ b/src/components/cursor-controller.js
@@ -93,6 +93,8 @@ AFRAME.registerComponent("cursor-controller", {
     this.data.playerRig.addEventListener(this.data.primaryUp, this._handlePrimaryUp);
     this.data.playerRig.addEventListener(this.data.grabEvent, this._handlePrimaryDown);
     this.data.playerRig.addEventListener(this.data.releaseEvent, this._handlePrimaryUp);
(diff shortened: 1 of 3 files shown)


src/components/cursor-controller.js
```diff
@@ -370,10 +376,8 @@ AFRAME.registerComponent("cursor-controller", {
   },
 
   _handleEnterVR: function() {
-    if (AFRAME.utils.device.checkHeadsetConnected()) {
```
brianpeiris: Why was it necessary to remove this check? In theory it should return true because the polyfill provides a "Cardboard" headset.

InfiniteLee: I'm not sure if thats true, at least it wasn't in the older version of the polyfill. But regardless listening to just enter/exit VR should be sufficient.

src/components/cursor-controller.js
```diff
@@ -93,6 +93,8 @@ AFRAME.registerComponent("cursor-controller", {
     this.data.playerRig.addEventListener(this.data.primaryUp, this._handlePrimaryUp);
     this.data.playerRig.addEventListener(this.data.grabEvent, this._handlePrimaryDown);
     this.data.playerRig.addEventListener(this.data.releaseEvent, this._handlePrimaryUp);
+    this.data.playerRig.addEventListener("gamepadbuttondown", this._handlePrimaryDown);
```
cvan: (apologies in advance, as I know this is a WIP PR.)

is this a correct branch? this appears to be for #364 perhaps?

btw, `gamepadbuttondown` are events that are proprietary to and work only in Firefox. (see w3c/gamepad#15 for context.)

InfiniteLee: `gamepadbuttondown` is fired by https://github.com/donmccurdy/aframe-extras/blob/master/src/controls/gamepad-controls.js

cvan: gd lol, I should've checked before spamming you with unnecessary info. thanks!

PR #349, opened 2018-05-02
fix/navmesh_teleport

InfiniteLee:
Reset the navGroup and navNode on teleport. This fixes the issue with not being able to teleport long distances (e.g. top of castle walls to bottom) because the navmesh code won't try (and fail) to find a path all the way from your original position.

### DIFF
--- a/src/components/character-controller.js
+++ b/src/components/character-controller.js
@@ -75,7 +75,9 @@ AFRAME.registerComponent("character-controller", {
   },
 
   handleTeleport: function(event) {
-    this.setPositionOnNavMesh(event.detail.oldPosition, this.el.object3D);
+    this.navGroup = null;
+    this.navNode = null;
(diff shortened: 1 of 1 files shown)


src/components/character-controller.js
```diff
@@ -75,7 +75,9 @@ AFRAME.registerComponent("character-controller", {
   },
 
   handleTeleport: function(event) {
-    this.setPositionOnNavMesh(event.detail.oldPosition, this.el.object3D);
+    this.navGroup = null;
+    this.navNode = null;
```
netpro2k: Pretty sure the fix here is just that you are telling it to "move" on the navmesh, from the new position, to the new position (basically just recenter yourself) vs before you were telling it to go from the old position to the new position, along the navmesh, and it would try and find a path. I don't think you need to clear the group or node. Also I think `setPositionOnNavMesh` would be more clear if it took a from/to position instead of just a from position and infering the "to" position based on where the object passed in.

PR #356, opened 2018-05-02
Only import lodash.merge instead of full library

robertlong:
I was digging around our webpack bundle in dev tools and noticed we accidentally imported the whole lodash library. Now we just import the merge function.

### DIFF
--- a/src/storage/store.js
+++ b/src/storage/store.js
@@ -1,5 +1,5 @@
 import { Validator } from "jsonschema";
-import { merge } from "lodash";
+import merge from "lodash/merge";
 
 const LOCAL_STORE_KEY = "___hubs_store";
 const STORE_STATE_CACHE_KEY = Symbol();


src/storage/store.js
```diff
@@ -1,5 +1,5 @@
 import { Validator } from "jsonschema";
-import { merge } from "lodash";
+import merge from "lodash/merge";
```
cvan: is `this.state` flat / only one level deep?

if so, [the `merge` call below](https://github.com/mozilla/hubs/pull/356/files#diff-0a39633babd056d1b5060747099dd5b6R72) could be a `Object.assign` call.

if it's possible for `this.state` to be deep, then, yeah, stick with `lodash/merge` :+1:

robertlong: It looks like it's two levels deep right now and may change so yeah we probably want to use the lodash merge.

cvan: sounds good, go for it - tested, works

PR #358, opened 2018-05-02
Fix regex checking for cardboard support

### LINKED ISSUE #352: Enable cardboard mode on iOS
For those who are brave enough for it, we should allow entry into stereoscopic mode on iOS (with the usual treatment of having it as a secondary option.)

InfiniteLee:
fix the regex so that ios cardboard is properly detected. Fixes #352 

Despite the name of the branch, iOS is explicitly disabled until we fix rendering on carboard iOS. Same for Android/Firefox.


gfodor:
What's the case here this misses? As it is right now I think `\W*` on both ends basically serves no purpose since it will match anything and could be removed. I'm guessing there is an example where it's a substring?

InfiniteLee:
@gfodor I believe it has to do with the fact that "Cardboard" is the first word that shows up in the navigator.getVRDisplays() name ("Cardboard VRDisplay") on iOS. So `\W` alone wont match anything.
```
"Cardboard VRDisplay".match(/\Wcardboard\W/i)
> null
"Cardboard VRDisplay".match(/\W*cardboard\W*/i)
> Array [ "Cardboard " ]
```
vs on Android: 
```
"Google, Inc. Cardboard v1".match(/\Wcardboard\W/i)
> Array [ "Cardboard " ]
```

InfiniteLee:
I'll try that. 

@brianpeiris and I also just discovered this same bug is preventing us from launching cardboard on firefox android.

InfiniteLee:
`\b` wont match if `cardboard` appears in the middle of another string e.g. "GoogleCardboard". I don't know if that will be important or not.

gfodor:
i guess its not the end of the world if we just match the substring. sgtm.

gfodor:
i'd just drop the `\w*`'s now

cvan:
feel free to just do `string.toLowerCase().includes("cardboard")`

src/utils/vr-caps-detect.js
```diff
@@ -70,12 +70,12 @@ export async function getAvailableVREntryTypes() {
       ? VR_DEVICE_AVAILABILITY.yes
       : VR_DEVICE_AVAILABILITY.no;
 
-    cardboard = displays.find(d => d.capabilities.canPresent && d.displayName.match(/\Wcardboard\W/i))
+    cardboard = displays.find(d => d.capabilities.canPresent && d.displayName.match(/\W*cardboard\W*/i))
```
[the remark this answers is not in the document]
gfodor: I'm guessing we need to re-write this all imminently once WebXR lands in origin trial in chrome, so probably not worth worrying about for now

src/utils/vr-caps-detect.js
```diff
@@ -43,6 +43,8 @@ export async function getAvailableVREntryTypes() {
   const isSamsungBrowser = browser.name === "chrome" && navigator.userAgent.match(/SamsungBrowser/);
   const isOculusBrowser = navigator.userAgent.match(/Oculus/);
   const isDaydreamCapableBrowser = !!(isWebVRCapableBrowser && browser.name === "chrome" && !isSamsungBrowser);
+  const isIPhone = navigator.userAgent.match(/iPhone/);
```
brianpeiris: Would we want to guard against other iDevices? See this regex @robertlong constructed previously: https://github.com/mozilla/hubs/blob/5a8c9635aa1414e5dcabf65d9709b880ce1a6cba/src/utils/ios-audio-context-fix.js#L5

cvan: :+1:  I don't own an iPad, but I'd assume the bug exists there as well?

gfodor: we already have `https://www.npmjs.com/package/mobile-detect` in our packages, perhaps that's a good way to check this? or `https://www.npmjs.com/package/device-detect`

PR #361, opened 2018-05-03
Use hub ID directly as NAF/Janus room ID

mquander:
**Warning: we should think before deploying this code.**

This obsoletes the `janusRoomId` that comes back from the hub Phoenix channel -- we can remove that from Reticulum in the future.

When we deploy this, people who join "the same" hub post-deploy as people who joined it pre-deploy will find themselves into a separate room. The pre-deploy folks will have to refresh to get into the new version of the hub. So maybe we should do it when not many people are using the service. I'm not sure what other reasonable thing we can do to mitigate the disruption -- in the future, we should probably be able to pop up a "Client updated -- please refresh" notice on people's browsers.

### DIFF
--- a/src/hub.js
+++ b/src/hub.js
@@ -201,7 +201,7 @@ const onReady = async () => {
     document.body.removeChild(scene);
   };
 
-  const enterScene = async (mediaStream, enterInVR, janusRoomId) => {
+  const enterScene = async (mediaStream, enterInVR, hubId) => {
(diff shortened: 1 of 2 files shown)


gfodor:
note @brianpeiris this will result in merge conflicts with https://github.com/mozilla/hubs/pull/370

brianpeiris:
👍 Merged these changes into #370

PR #367, opened 2018-05-04
Oculus Go support (WIP)

robertlong:
Todo
- [ ] Finish upgrading AFrame (#359)
- [x] Skip enter VR screen
- [x] Taking off the headset doesn't fire "blur" event.  Ensure that we exit the scene and mute mic like on other platforms.
- [x] Force AA on
- [ ] Improve overall performance to hit 60 FPS (possibly another series of PRs)

### DIFF
--- a/src/avatar-selector.html
+++ b/src/avatar-selector.html
@@ -6,9 +6,9 @@
   <title>avatar selector</title>
   <link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons">
   <% if(NODE_ENV === "production") { %>
-    <script src="https://cdn.rawgit.com/brianpeiris/aframe/845825ae694449524c185c44a314d361eead4680/dist/aframe-master.min.js"></script>
+    <script src="https://cdn.rawgit.com/aframevr/aframe/dba7369/dist/aframe-master.min.js" integrity="sha384-cKjxF5vmHB6Jq8O24NlFfpIzDNKSapSmpIKyg6KzGAsliOfbCAJHxE/EZHzw9xU0" crossorigin="anonymous"></script>
   <% } else { %>
-    <script src="https://cdn.rawgit.com/brianpeiris/aframe/845825ae694449524c185c44a314d361eead4680/dist/aframe-master.js"></script>
+    <script src="https://cdn.rawgit.com/aframevr/aframe/dba7369/dist/aframe-master.js" integrity="sha384-lDPOpFIESh74q9UjPSiOy6dIBuvtUzKzBzV9UTvIDW+9eiL3PjlMRBJuTWq8se7Y" crossorigin="anonymous"></script>
   <% } %>
 </head>
 
--- a/src/avatar-selector.html
+++ b/src/avatar-selector.html
@@ -6,9 +6,9 @@
   <title>avatar selector</title>
   <link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons">
   <% if(NODE_ENV === "production") { %>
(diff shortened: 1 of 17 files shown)


robertlong:
Oculus Browser is not emitting a blur, vrdisplaydeactivate, vrdisplayblur, visibilitychange, or vrdisplaypresentchange event when taking off the headset or when it goes into standby. It does emit a second vrdisplayfocus event when taking the headset off, but I feel uncomfortable using that event 😕 

I reached out to the devs at Oculus and hopefully we'll have a fix in soon. I'd rather not merge this without that change as your mic stays on indefinitely with zero indication that it's still recording.

I guess I can hack it by polling the vr device state if we want to get it out earlier?

robertlong:
VR device state didn't work either. I added a timeout in the tick method of the `exit-on-blur` system. This is a hack, but it works quite well for now. I'm assuming nothing else pauses that system/requestAnimationFrame for more than 15 seconds on Oculus Browser.

src/avatar-selector.html
```diff
@@ -6,9 +6,9 @@
   <title>avatar selector</title>
   <link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons">
   <% if(NODE_ENV === "production") { %>
```
cvan: FYI: @gfodor has PR #217, which was originally to bump A-Frame to `0.8.2`: https://github.com/mozilla/hubs/pull/217/files

as I understand it, it's probably best to point to `3e7a4b3`. likely, PR #217 can be closed?

src/hub.html
```diff
@@ -13,9 +13,9 @@
     <link href="https://fonts.googleapis.com/css?family=Zilla+Slab:300,300i,400,400i,700" rel="stylesheet">
 
     <% if(NODE_ENV === "production") { %>
```
cvan: FYI: @gfodor has PR #217, which was originally to bump A-Frame to `0.8.2`: https://github.com/mozilla/hubs/pull/217/files

as I understand it, it's probably best to point to aframevr/aframe@3e7a4b3. likely, PR #217 can be closed?

src/systems/exit-on-blur.js
```diff
@@ -2,18 +2,26 @@ AFRAME.registerSystem("exit-on-blur", {
...
+    this.isOculusBrowser = navigator.userAgent.match(/Oculus/);
 
     window.addEventListener("blur", this.onBlur);
     window.addEventListener("focus", this.onFocus);
 
     this.exitTimeout = null;
   },
 
+  tick() {
+    if (this.isOculusBrowser) {
```
netpro2k: I would add a note about what this hack is for so that we know when it can be removed (when the WebVR events get fixed)

src/hub.js
```diff
@@ -9,7 +9,7 @@ patchWebGLRenderingContext();
 
 import "aframe-xr";
 import "./vendor/GLTFLoader";
-import "networked-aframe";
+import "networked-aframe/src/index";
```
netpro2k: whats the deal with this change?

src/systems/exit-on-blur.js
```diff
@@ -2,18 +2,26 @@ AFRAME.registerSystem("exit-on-blur", {
...
+      this.exitTimeout = setTimeout(this.onTimeout, 15 * 1000);
+    }
+  },
+
   onBlur() {
     if (this.el.isMobile) {
-      this.exitTimeout = setTimeout(() => {
-        this.el.dispatchEvent(new CustomEvent("exit"));
-      }, 30 * 1000);
+      this.exitTimeout = setTimeout(this.onTimeout, 30 * 1000);
```
netpro2k: Should probably clearTimeout here so you dont end up accidentaly firing it multiple times.

robertlong: Yeah, good point.

src/hub.js
robertlong: networked-aframe's package.json exposes the bundled version of the library. This makes debugging harder because no sourcemaps are generated. This is a temporary fix until we change networked-aframe's package.json

robertlong: I've created a PR: https://github.com/networked-aframe/networked-aframe/pull/127

PR #368, opened 2018-05-04
Some small UX fixes for the landing page and dialogs

### LINKED ISSUE #305: contrast too bright on video background
the tagline and subtitle copy is a bit difficult to read with the `<video>` playing in the background. if the `opacity` was bumped down a bit, I think it'd be easier on the eyes.

screen shot 2018-04-25 at 8 45 10 pm

### LINKED ISSUE #313: modal windows should close upon pressing the `Esc` key or clicking outside the modal
1. load https://hubs.mozilla.com/
2. scroll to the bottom of the page to find the footer
3. click one of the first three links
4. try to press `Esc` or click outside the modal (i.e., usual patterns for dismissing) to dismiss the modal
5. notice you have to click/tap the × icon in the top-left corner of the modal

### LINKED ISSUE #315: Add Privacy Notice + ToS to bottom of Help dialog
1. load https://hubs.mozilla.com/
2. notice the _Mozilla_, _Mozilla Mixed Reality_, and _View Source_ links in the header
    screen shot 2018-04-25 at 10 09 27 pm
3. scroll down to the footer to notice the following: _Terms of Use_, _Privacy Notice_, and the _made with 🦆by Mozilla_
    screen shot 2018-04-25 at 10 09 29 pm

<hr>

1. load https://hubs.mozilla.com/0h6bsjd7ofz/snoopy-soft-festivity
2. notice there is no Mozilla branding (nor _Terms of Use_, _Privacy Notice_, and _View Source_ links; I could be missing them somewhere though)
    screen shot 2018-04-25 at 10 10 44 pm
    screen shot 2018-04-25 at 10 10 39 pm
    screen shot 2018-04-25 at 10 10 37 pm
3. is that intentional? (if so, apologies for filing this issue!)

notice the interstitial avatar-selection screen does though show the _Terms of Use_ and _Privacy Notice_:
screen shot 2018-04-25 at 10 01 43 pm

netpro2k:
- Close dialogs on escape (fixes #313)
- Darken hero video (fixes #305)
- Surface ToS and Privacy Policy in help (Fixes #315)

### DIFF
--- a/src/assets/stylesheets/info-dialog.scss
+++ b/src/assets/stylesheets/info-dialog.scss
@@ -3,7 +3,7 @@
   height: 100%;
   top: 0;
   left: 0;
-  position: absolute;
+  position: fixed;
   pointer-events: none;
   color: white;
   z-index: 2;
--- a/src/assets/stylesheets/info-dialog.scss
+++ b/src/assets/stylesheets/info-dialog.scss
@@ -54,6 +54,17 @@
 	a { color: white }
       }
 
+      &__links {
+        display: flex;
+        justify-content: center;
+        margin: 16px 0;
+        a {
+          margin: 0px 12px;
+          color: $light-text;
+        }
+      }
+
+
       &__close {
 	position: absolute;
 	left: 12px;
--- a/src/assets/stylesheets/info-dialog.scss
+++ b/src/assets/stylesheets/info-dialog.scss
@@ -4,7 +4,6 @@
   top: 0;
   left: 0;
   position: fixed;
-  pointer-events: none;
   color: white;
   z-index: 2;
 }
@@ -59,7 +58,7 @@
         justify-content: center;
         margin: 16px 0;
         a {
-          margin: 0px 12px;
+          margin: 0 12px;
           color: $light-text;
         }
       }
--- a/src/react-components/info-dialog.js
+++ b/src/react-components/info-dialog.js
@@ -27,19 +27,26 @@ class InfoDialog extends Component {
 
     const loc = document.location;
     this.shareLink = `${loc.protocol}//${loc.host}${loc.pathname}`;
-    this.onKeyPress = this.onKeyPress.bind(this);
+    this.onKeyDown = this.onKeyDown.bind(this);
+    this.onContainerClicked = this.onContainerClicked.bind(this);
   }
 
   componentDidMount() {
-    window.addEventListener("keypress", this.onKeyPress);
+    window.addEventListener("keydown", this.onKeyDown);
   }
 
   componentWillUnmount() {
-    window.removeEventListener("keypress", this.onKeyPress);
+    window.removeEventListener("keydown", this.onKeyDown);
   }
 
-  onKeyPress(e) {
+  onKeyDown(e) {
     if (e.key === "Escape") {
+      this.props.onCloseDialog();
+    }
+  }
+
+  onContainerClicked(e) {
+    if (e.currentTarget === e.target) {
       this.props.onCloseDialog();
     }
   }
@@ -226,7 +233,7 @@ class InfoDialog extends Component {
             <p>
               The <b>Bubble Toggle</b> hides avatars that enter your personal space.
             </p>
-            <div className="dialog__box__contents__links">
+            <p className="dialog__box__contents__links">
               <a target="_blank" rel="noopener noreferrer" href="https://github.com/mozilla/hubs/blob/master/TERMS.md">
                 <FormattedMessage id="profile.terms_of_use" />
               </a>
@@ -237,7 +244,7 @@ class InfoDialog extends Component {
               >
                 <FormattedMessage id="profile.privacy_notice" />
               </a>
-            </div>
+            </p>
           </div>
         );
         break;
@@ -250,7 +257,7 @@ class InfoDialog extends Component {
 
     return (
       <div className="dialog-overlay">
-        <div className={dialogClasses}>
+        <div className={dialogClasses} onClick={this.onContainerClicked}>
           <div className="dialog__box">
             <div className="dialog__box__contents">
               <button className="dialog__box__contents__close" onClick={this.props.onCloseDialog}>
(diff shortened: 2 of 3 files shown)


src/react-components/info-dialog.js
```diff
@@ -27,6 +27,21 @@ class InfoDialog extends Component {
...
+  componentDidMount() {
+    window.addEventListener("keypress", this.onKeyPress);
+  }
+
+  componentWillUnmount() {
+    window.removeEventListener("keypress", this.onKeyPress);
+  }
+
+  onKeyPress(e) {
+    if (e.key === "Escape") {
```
cvan: can you check `document.activeElement.matches('body, a')`? (or you can check `e.target` too. there are a couple ways to test this. the `a` is because the original footer link still has focus when the `click` event has fired to open the modal but the user has not interacted with the page/modal. usually, `document.activeElement` would be `document.body`.)

in the _Get Updates_ modal, when a user has focused the `<input>` field for the email address, pressing the `Escape` key should not close the dialog.

image

src/react-components/info-dialog.js
```diff
@@ -27,6 +27,21 @@ class InfoDialog extends Component {
 
     const loc = document.location;
     this.shareLink = `${loc.protocol}//${loc.host}${loc.pathname}`;
+    this.onKeyPress = this.onKeyPress.bind(this);
+  }
+
+  componentDidMount() {
+    window.addEventListener("keypress", this.onKeyPress);
```
cvan: can you change this to `keydown`? it doesn't work in Chrome (tested on macOS).

the `keypress` event does fire correctly in both Firefox and Safari. in Chrome, it turns out that the `keypress` event is emitted for character keys, but evidently not for such keys as `Escape`, `ArrowDown`, etc.

src/assets/stylesheets/info-dialog.scss
```diff
@@ -3,7 +3,7 @@
   height: 100%;
   top: 0;
   left: 0;
-  position: absolute;
+  position: fixed;
```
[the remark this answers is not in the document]
cvan: see my comment here: https://github.com/mozilla/hubs/pull/368#issuecomment-387236650

src/react-components/info-dialog.js
```diff
@@ -27,6 +27,21 @@ class InfoDialog extends Component {
...
+  }
+
+  componentWillUnmount() {
+    window.removeEventListener("keypress", this.onKeyPress);
+  }
+
+  onKeyPress(e) {
+    if (e.key === "Escape") {
+      this.props.onCloseDialog();
```
cvan: also, per issue #313 – `… or clicking outside the modal` – can you also handle that case? I see that a common convention used for both desktop and touch/mobile sites/apps.

src/react-components/info-dialog.js
netpro2k: Hmm, I am not sure about that one. I would still expect the modal to close even if I was in the textbox if I hit excape. Checking for the "a" doesnt seem right because I could click somewhere inside the dialog and then no longer have an "a" focused.

src/react-components/info-dialog.js
```diff
@@ -211,6 +226,18 @@ class InfoDialog extends Component {
             <p>
               The <b>Bubble Toggle</b> hides avatars that enter your personal space.
             </p>
+            <div className="dialog__box__contents__links">
```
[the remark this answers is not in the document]
netpro2k: "everythign is a div" is a hard habbit to break hehe.

src/react-components/info-dialog.js
cvan: for accessibility reasons, we try to not hijack keys when form fields are focused. in past projects, we do something [like this](https://github.com/mozilla/marketplace-core-modules/blob/59bb600/core/utils.js#L63-L66).

if it's not too much trouble, can you add a tiny helper function to handle this?

```js
var formTags = /input|keygen|meter|option|output|progress|select|textarea/i;
function fieldFocused (e) {
  return formTags.test(e.target.nodeName);
}
```

and then use as such:

```js
if (!fieldFocused(e.target) && e.key === "Escape") {
```

src/assets/stylesheets/info-dialog.scss
```diff
@@ -3,8 +3,7 @@
   height: 100%;
   top: 0;
   left: 0;
-  position: absolute;
-  pointer-events: none;
+  position: fixed;
```
cvan: I notice the `<body>` scrolling issue (from [my GIF](https://github.com/mozilla/hubs/pull/368#issuecomment-387236650)) still remains - want me to file a follow-up issue for that?

src/react-components/info-dialog.js
```diff
@@ -27,6 +27,28 @@ class InfoDialog extends Component {
...
+  }
+
+  onKeyDown(e) {
+    if (e.key === "Escape") {
+      this.props.onCloseDialog();
+    }
+  }
+
+  onContainerClicked(e) {
+    if (e.currentTarget === e.target) {
```
cvan: I noticed that the Help dialog from a Hub space (e.g., on [https://localhost:8080/hub.html?hub_id=0mnu4u5z3q2](https://localhost:8080/hub.html?hub_id=0mnu4u5z3q2)) does not get closed when clicking outside. can you change this to `if (e.target.closest(".dialog-overlay")) {`? (and likely need to move the `pointer-events: none` style.)

src/react-components/info-dialog.js
```diff
@@ -27,6 +27,28 @@ class InfoDialog extends Component {
 
     const loc = document.location;
     this.shareLink = `${loc.protocol}//${loc.host}${loc.pathname}`;
+    this.onKeyDown = this.onKeyDown.bind(this);
+    this.onContainerClicked = this.onContainerClicked.bind(this);
+  }
+
+  componentDidMount() {
+    window.addEventListener("keydown", this.onKeyDown);
```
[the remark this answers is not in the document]
cvan: (this will reset the focus to `document.body`, which is what we'd expect from clicking a modal IMO)

src/assets/stylesheets/info-dialog.scss
netpro2k: Yep. I mentioned in slack, the rest of these fixes are clearly minimal impact but I could imagine that one having unintended side-effects since it involves putting `pointer-events: none` on the `body`

PR #370, opened 2018-05-08
Motion capture bots

brianpeiris:
- Add a `?bot` mode to hubs that runs a bot client using recorded avatar motions and recorded audio.
- Use [`aframe-motion-capture-components`](https://github.com/dmarcos/aframe-motion-capture-components/) to replay HMD and controller positions instead of running [bots based on NAF updates](https://github.com/mozilla/janus-plugin-sfu/blob/master/client/squawkers-cli.js). ([This tool](http://swimminglessonsformodernlife.com/aframe-motion-capture-components/examples/record.html) can be used to produce recordings)
- Add scripts for running bots locally. Also used by accompanying ops PR https://github.com/mozilla/hubs-ops/pull/16

## Todo:
- [x] Point to bumped naf-janus-adapter after publishing https://github.com/mozilla/naf-janus-adapter/pull/43
- [x] Re-record bot json so that bot doesn't appear outside smoke test room.

### DIFF
--- a/scripts/bot/run-bot.sh
+++ b/scripts/bot/run-bot.sh
@@ -16,6 +16,10 @@ yarn
 echo 'Running Hubs'
 yarn serve --ssl --port 8080 ../../public &
 echo 'Running bots'
-node run-bot.js
+if [ "$1" != "" ]; then
+  node run-bot.js --room $1
+else
+  node run-bot.js
+fi
 
 trap 'kill $(jobs -pr)' SIGINT SIGTERM EXIT


src/utils/autoplay-audio-context-fix.js
```diff
@@ -1,5 +1,5 @@
 /**
- * Mobile Safari will start Audio contexts in a "suspended" state.
+ * Mobile Safari and Chrome will start Audio contexts in a "suspended" state.
```
cvan: sad times we live in!

package.json
```diff
@@ -23,6 +23,7 @@
     "aframe-billboard-component": "^1.0.0",
     "aframe-extras": "https://github.com/MozillaReality/aframe-extras#feature/precompute-nav-mesh",
     "aframe-input-mapping-component": "https://github.com/mozillareality/aframe-input-mapping-component#hubs/master",
+    "aframe-motion-capture-components": "https://github.com/mozillareality/aframe-motion-capture-components#1ca616fa67b627e447b23b35a09da175d8387668",
```
brianpeiris: Points to a branch with support for A-Frame 0.8.0. See https://github.com/dmarcos/aframe-motion-capture-components/pull/40

scripts/bot/run-bot.js
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env node
+const doc = `
+Usage:
+    ./run-bot.js [options]
+
+Options:
+    -h --host=<host>  Hubs host [default: localhost:8080]
+    -r --room=<room>  Room id [default: 234234].
```
gfodor: note these are now able to be strings

src/components/avatar-replay.js
```diff
@@ -0,0 +1,36 @@
...
+    this.el.addEventListener("model-loaded", () => {
+      const cameraReplayer = camera.components["motion-capture-replayer"];
+      cameraReplayer.startReplaying(botRecording.camera);
+      const leftControllerReplayer = leftController.components["motion-capture-replayer"];
+      leftControllerReplayer.startReplaying(botRecording.left);
+      const rightControllerReplayer = rightController.components["motion-capture-replayer"];
+      rightControllerReplayer.startReplaying(botRecording.right);
+    });
+  },
+  _setupController: function(controller) {
```
gfodor: perhaps these blacklisted components should be moved into a const list at the top of the file?

gfodor: also sorry but I have to ask :) what is up with the `hand-controls2`?

gfodor: (part of the reason i ask is if we change it, we could just have a convention where we strip `-controls` components. might be too magical though)

src/components/avatar-replay.js
```diff
@@ -0,0 +1,36 @@
+import botRecording from "../assets/avatars/bot-recording.json";
```
gfodor: might be worth renaming `bot-recording.json` and `bot-recording.mp3` to describe their contents, in case we want to have more than one kind of bot

src/hub.js
```diff
@@ -300,6 +303,21 @@ const onReady = async () => {
...
+      if (qsTruthy("bot")) {
+        playerRig.setAttribute("avatar-replay", {
+          camera: "#player-camera",
+          leftController: "#player-left-controller",
+          rightController: "#player-right-controller"
+        });
+        const audio = document.getElementById("bot-recording");
+        mediaStream.addTrack(audio.captureStream().getAudioTracks()[0]);
+        // wait for runner script to interact with the page so that we can play audio.
+        await new Promise(resolve => {
```
gfodor: i thought our linter would yell at the `{ }`'s here /shrug

src/hub.js
```diff
@@ -352,14 +370,31 @@ const onReady = async () => {
   const initialEnvironmentEl = document.createElement("a-entity");
   initialEnvironmentEl.addEventListener("bundleloaded", () => {
     remountUI({ initialEnvironmentLoaded: true });
-    // Wait a tick plus some margin so that the environments actually render.
-    setTimeout(() => scene.renderer.animate(null), 100);
+    // We never want to stop the render loop when were running in "bot" mode.
+    if (!qsTruthy("bot")) {
```
gfodor: i think this should be pulled up to a `const isBotMode` since we do it in a few places

src/hub.js
```diff
@@ -352,14 +370,31 @@ const onReady = async () => {
...
+      // Stop rendering while the UI is up. We restart the render loop in enterScene.
+      // Wait a tick plus some margin so that the environments actually render.
+      setTimeout(() => scene.renderer.animate(null), 100);
+    }
   });
   environmentRoot.appendChild(initialEnvironmentEl);
 
+  const setRoom = (janusRoomId, hubName) => {
+    if (qsTruthy("bot")) {
```
gfodor: I'd either flip the conditions here to put the non-bot case first, or leave it as is and just return early (perhaps extracting the bot scene entry mechanics into a new function) and drop the `else`. (the reason being I like to have exceptional conditions (like bot mode) come after the main code path or be early exit'ed via a guard clause, but just style thing probably)

src/hub.js
```diff
@@ -352,14 +370,31 @@ const onReady = async () => {
...
+      }
+    } else {
+      remountUI({ janusRoomId, hubName });
+    }
+  };
+
   if (qs.room) {
     // If ?room is set, this is `yarn start`, so just use a default environment and query string room.
-    remountUI({ janusRoomId: qs.room && !isNaN(parseInt(qs.room)) ? parseInt(qs.room) : 1 });
+    setRoom(qs.room && !isNaN(parseInt(qs.room)) ? parseInt(qs.room) : 1);
```
gfodor: I think this inadvertently will set `hubName` on the UI props to `undefined`, since it's not passed in, right?

gfodor: (Probably nbd but just thought I'd mention it)

gfodor:
nice work!

src/components/avatar-replay.js
brianpeiris: Sounds to magical to me. I don't like hand-controls2 either, but we called it that when we had to make modifications to A-Frame's hand-controls but weren't sure if it was worth contributing back.

`const` list sounds like a good idea though

src/hub.js
brianpeiris: Makes sense. I'll flip the conditional.

src/hub.js
brianpeiris: Yeah, this happens when you use `hub.html?room=...` anyway.

src/hub.js
gfodor: oh, right.

scripts/bot/run-bot.js
```diff
@@ -0,0 +1,52 @@
...
+    -h --help         Show this screen.
+`;
+
+const docopt = require("docopt").docopt;
+const options = docopt(doc);
+
+const puppeteer = require("puppeteer");
+const querystring = require("query-string");
```
cvan: not sure which version of Node is required for development and production (though the `.travis.yml` file specifies Node version 9). in Node version 8+, you can use `const { URL } = require("url");`

Issue #266, opened 2018-04-24
Keep track of cursor event identifiers instead of doing distance check between touches for cursor.


johnshaughnessy:
We should know when a touch end event is the one to drop a duck based on its identifier, not its distance to the preview valid touchpoint:
https://github.com/mozilla/hubs/commit/da4b2c02e225507d74a0773122ed6f823c989885#diff-bafcf688d672a56b722fa967f5b16acbR316

Issue #252, opened 2018-04-23
Remove the use of `setTimeout` from the `touchStart` handler in `cursor-controller`.


johnshaughnessy:
Remove the use of `setTimeout` from the `touchStart` handler in `cursor-controller`.

PR #372, opened 2018-05-08
Bug fixes and cleanup in cursor and hud rendering.

johnshaughnessy:
Clean up touch events in cursor controller.

Fixes https://github.com/mozilla/hubs/issues/266 and https://github.com/mozilla/hubs/issues/252.

Fixes a bug where if you `primary-action-down` while hovering on UI you set `grabStarting` to `true`, then if you `primary-action-up` while not hovering on UI, `grabStarting` remains `true` and thus subsequent attempts to pick up ducks is blocked.

Fixes a bug where the cursor was not drawing on top of the HUD properly, and a related bug where the HUD wasn't rendered in the correct order.


src/components/cursor-controller.js
```diff
@@ -255,77 +249,71 @@ AFRAME.registerComponent("cursor-controller", {
   },
 
   _handleTouchStart: function(e) {
-    if (!this.isMobile || this.hasPointingDevice) return;
-
-    const touch = e.touches[0];
-    if (touch.clientY / window.innerHeight >= 0.8) return true;
-    this.mousePos.set(touch.clientX / window.innerWidth * 2 - 1, -(touch.clientY / window.innerHeight) * 2 + 1);
-    this._updateRaycasterIntersections();
+    if (!this.isMobile || this.hasPointingDevice || this.activeTouch) return;
```
InfiniteLee: can you define this.activeTouch up in init() with the other variables?

src/components/cursor-controller.js
```diff
@@ -255,77 +249,71 @@ AFRAME.registerComponent("cursor-controller", {
...
+    for (let i = e.touches.length - 1; i >= 0; i--) {
+      const touch = e.touches[i];
+      if (touch.clientY / window.innerHeight < 0.8) {
+        this.activeTouch = touch;
+        break;
       }
     }
+    if (!this.activeTouch) return;
 
-    this._setLookControlsEnabled(false);
```
InfiniteLee: Is this a design decision to stop doing this, or was it just not doing anything before?

src/components/cursor-controller.js
```diff
@@ -41,30 +41,25 @@ AFRAME.registerComponent("cursor-controller", {
...
-    this._handleCursorLoaded = this._handleCursorLoaded.bind(this);
-    this._handleControllerConnected = this._handleControllerConnected.bind(this);
-    this._handleControllerDisconnected = this._handleControllerDisconnected.bind(this);
-
-    this._handleTouchStart = this._handleTouchStart.bind(this);
-    this._updateRaycasterIntersections = this._updateRaycasterIntersections.bind(this);
-    this._handleTouchMove = this._handleTouchMove.bind(this);
-    this._handleTouchEnd = this._handleTouchEnd.bind(this);
-
-    this.el.sceneEl.renderer.sortObjects = true;
```
InfiniteLee: Is this also not needed anymore for drawing the cursor on top?

src/components/cursor-controller.js
```diff
@@ -41,30 +41,33 @@ AFRAME.registerComponent("cursor-controller", {
...
-    this._handleModelLoaded = this._handleModelLoaded.bind(this);
-    this._handleCursorLoaded = this._handleCursorLoaded.bind(this);
-    this._handleControllerConnected = this._handleControllerConnected.bind(this);
-    this._handleControllerDisconnected = this._handleControllerDisconnected.bind(this);
-
-    this._handleTouchStart = this._handleTouchStart.bind(this);
-    this._updateRaycasterIntersections = this._updateRaycasterIntersections.bind(this);
-    this._handleTouchMove = this._handleTouchMove.bind(this);
-    this._handleTouchEnd = this._handleTouchEnd.bind(this);
+    const functionNames = [
```
cvan: note that this makes the minified code (i.e., name mangling) larger because the strings won't be reduced to single-letter vars

src/components/cursor-controller.js
```diff
@@ -255,77 +258,71 @@ AFRAME.registerComponent("cursor-controller", {
...
-
-  _updateRaycasterIntersections: function() {
-    const raycaster = this.el.components.raycaster.raycaster;
+    // Update the ray and cursor positions
+    const raycasterComp = this.el.components.raycaster;
+    const raycaster = raycasterComp.raycaster;
     const camera = this.data.camera.components.camera.camera;
+    const cursor = this.data.cursor;
+    this.mousePos.set(
+      this.activeTouch.clientX / window.innerWidth * 2 - 1,
```
cvan: just curious: is `-1` always the correct offset?

src/components/cursor-controller.js
```diff
@@ -255,77 +258,71 @@ AFRAME.registerComponent("cursor-controller", {
...
     raycaster.setFromCamera(this.mousePos, camera);
-    this.origin = raycaster.ray.origin;
-    this.direction = raycaster.ray.direction;
-    this.el.setAttribute("raycaster", { origin: this.origin, direction: this.direction });
-    this.el.components.raycaster.checkIntersections();
+    raycasterComp.data.origin = raycaster.ray.origin;
+    raycasterComp.data.direction = raycaster.ray.direction;
+    raycasterComp.checkIntersections();
+    const intersections = raycasterComp.intersections;
+    if (intersections.length == 0 || intersections[0].distance >= this.data.maxDistance) {
```
cvan: nit: when `yarn prettier` is run, the `==` is changed to `===`

src/components/cursor-controller.js
```diff
@@ -255,77 +250,71 @@ AFRAME.registerComponent("cursor-controller", {
...
-    for (let i = 0; i < e.changedTouches.length; i++) {
-      const touch = e.changedTouches[i];
-      const thisTouchDidNotDriveMousePos =
-        Math.abs(touch.clientX - this.lastTouch.clientX) > 0.1 &&
-        Math.abs(touch.clientY - this.lastTouch.clientY) > 0.1;
-      if (thisTouchDidNotDriveMousePos) {
-        return;
+    if (!this.isMobile || this.hasPointingDevice || !this.activeTouch) return;
+
+    if (e.touches) {
```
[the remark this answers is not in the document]
cvan: it appears [a `TouchEvent` will always have a `touches` property](https://m14i.wordpress.com/2009/10/25/javascript-touch-and-gesture-events-iphone-and-android/):

image

<hr>

unrelated but funny observation: `touches` can never be `> 5`. I tried two-hand smashing my iPhone.

src/components/cursor-controller.js
```diff
@@ -255,77 +250,71 @@ AFRAME.registerComponent("cursor-controller", {
...
+      for (let i = 0; i < e.touches.length; i++) {
```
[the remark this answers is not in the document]
cvan: could also do this:

```js
if (e.touches.some(touch => touch.identifier === this.activeTouch.identifier)) { return; }
```

src/components/cursor-controller.js
johnshaughnessy: I can, but is writing `this.activeTouch = null` better than having `this.activeTouch` be `undefined` when I only care if it's truthy?

src/components/cursor-controller.js
johnshaughnessy: Design decision since stopping look controls when the gyro is enabled feels bad

src/components/cursor-controller.js
johnshaughnessy: It was, actually. Added back in with my more recent changes

src/components/cursor-controller.js
```diff
@@ -41,30 +41,25 @@ AFRAME.registerComponent("cursor-controller", {
...
+      "_handleWheel",
+      "_handleEnterVR",
+      "_handleExitVR",
+      "_handlePrimaryDown",
+      "_handlePrimaryUp",
+      "_handleModelLoaded",
+      "_handleControllerConnected",
+      "_handleControllerDisconnected"
+    ];
+    functionNames.forEach(name => {
```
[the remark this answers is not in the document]
johnshaughnessy: 👍

src/components/cursor-controller.js
```diff
@@ -41,30 +41,25 @@ AFRAME.registerComponent("cursor-controller", {
...
-    this._handleTouchEnd = this._handleTouchEnd.bind(this);
-
-    this.el.sceneEl.renderer.sortObjects = true;
-    this.data.cursor.addEventListener("loaded", this.cursorLoadedListener);
-  },
-
-  remove: function() {
-    this.data.cursor.removeEventListener("loaded", this._cursorLoadedListener);
+    const functionNames = [
+      "_handleTouchStart",
```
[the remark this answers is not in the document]
johnshaughnessy: 👍

src/components/cursor-controller.js
```diff
@@ -255,77 +250,71 @@ AFRAME.registerComponent("cursor-controller", {
...
+    this.mousePos.set(
+      this.activeTouch.clientX / window.innerWidth * 2 - 1,
+      -(this.activeTouch.clientY / window.innerHeight) * 2 + 1
+    );
     raycaster.setFromCamera(this.mousePos, camera);
-    this.origin = raycaster.ray.origin;
-    this.direction = raycaster.ray.direction;
-    this.el.setAttribute("raycaster", { origin: this.origin, direction: this.direction });
-    this.el.components.raycaster.checkIntersections();
+    raycasterComp.data.origin = raycaster.ray.origin;
```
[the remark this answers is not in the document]
johnshaughnessy: No good reason, I'll change it back.

src/components/cursor-controller.js
johnshaughnessy: I'm not 100% sure if `clientX` and `innerWidth` are the correct values here, but if they are then the `-1` is the correct offset.
I'm counting on `clientX / innerWidth` to return a value between [0,1]. Then I multiply by 2 to spread it over [0,2]. Then I subtract 1 to put it into the normalized device coordinates that threejs expects ([-1, 1]).

johnshaughnessy: Perhaps `touch.screenX` or `touch.pageX` or `window.outerWidth` is more correct, I'm not sure.

src/components/cursor-controller.js
johnshaughnessy: 👍

src/components/cursor-controller.js
johnshaughnessy: changed to functions

src/components/cursor-controller.js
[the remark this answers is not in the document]
johnshaughnessy: 👍

src/components/cursor-controller.js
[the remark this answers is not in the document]
johnshaughnessy: Tried to do that, I don't know how

src/components/cursor-controller.js
```diff
@@ -41,24 +41,25 @@ AFRAME.registerComponent("cursor-controller", {
...
-      this._handleEnterVR,
-      this._handleExitVR,
-      this._handlePrimaryDown,
-      this._handlePrimaryUp,
-      this._handleModelLoaded,
-      this._handleCursorLoaded,
-      this._handleControllerConnected,
-      this._handleControllerDisconnected
-    ].forEach(fn => {
-      fn = fn.bind(this);
```
johnshaughnessy: (This does not bind the function as desired.)

src/components/cursor-controller.js
```diff
@@ -312,7 +312,7 @@ AFRAME.registerComponent("cursor-controller", {
       !this.isMobile ||
       this.hasPointingDevice ||
       !this.activeTouch ||
-      e.touches.some(touch => touch.identifier === this.activeTouch.identifier)
+      this.some(e.touches, touch => touch.identifier === this.activeTouch.identifier)
```
cvan: is the `this.some` helper because `e.touches` is a `TouchList` and not an `Array`? could you remove the `some` method and call directly `Array.prototype.some.call(e.touches, touch => touch.identifier === this.activeTouch.identifier)`?

src/components/cursor-controller.js
InfiniteLee: It's mainly because thats how all the other variables are defined in this file and seems reasonable to maintain that pattern.

src/components/cursor-controller.js
johnshaughnessy: yes

src/components/cursor-controller.js
```diff
@@ -38,33 +38,36 @@ AFRAME.registerComponent("cursor-controller", {
...
-
-    this._handleTouchStart = this._handleTouchStart.bind(this);
-    this._updateRaycasterIntersections = this._updateRaycasterIntersections.bind(this);
-    this._handleTouchMove = this._handleTouchMove.bind(this);
-    this._handleTouchEnd = this._handleTouchEnd.bind(this);
-
-    this.el.sceneEl.renderer.sortObjects = true;
-    this.data.cursor.addEventListener("loaded", this.cursorLoadedListener);
+    const functionNames = [
+      "_handleTouchStart",
```
cvan: - https://github.com/mozilla/hubs/pull/372#discussion_r186886185
- https://github.com/mozilla/hubs/pull/372#discussion_r186895932

although it was previously 15 lines, this is now 20 lines. using an array of strings increases the chances for linting mistakes, per @netpro2k's comment. and to my comment, it increases the minified bundle size because the strings won't be renamed to single-character function names.

src/components/cursor-controller.js
```diff
@@ -413,7 +404,7 @@ AFRAME.registerComponent("cursor-controller", {
   },
 
   _handleCursorLoaded: function() {
-    this.data.cursor.object3DMap.mesh.renderOrder = 1;
+    this.data.cursor.object3DMap.mesh.renderOrder = window.RENDER_ORDER.CURSOR;
```
cvan: this is now `3`. that's expected, yes?

src/hub.js
```diff
@@ -111,6 +111,12 @@ import { generateDefaultProfile, generateRandomName } from "./utils/identity.js"
 import { getAvailableVREntryTypes, VR_DEVICE_AVAILABILITY } from "./utils/vr-caps-detect.js";
 import ConcurrentLoadDetector from "./utils/concurrent-load-detector.js";
 
+window.RENDER_ORDER = {
```
cvan: can this be scoped under the [`window.APP` global](https://github.com/mozilla/hubs/blob/1b3c30b1/src/hub.js#L78)?

src/hub.js
```diff
@@ -209,6 +215,7 @@ const onReady = async () => {
 
   const enterScene = async (mediaStream, enterInVR, janusRoomId) => {
     const scene = document.querySelector("a-scene");
+    scene.renderer.sortObjects = true;
```
cvan: does this need to be reverted in `exitScene`?

src/hub.js
```diff
@@ -209,6 +215,7 @@ const onReady = async () => {
...
     const playerRig = document.querySelector("#player-rig");
     document.querySelector("a-scene canvas").classList.remove("blurred");
```
cvan: nit (know this wasn't changed in this commit): this can be `scene.querySelector("canvas")`

src/components/cursor-controller.js
johnshaughnessy: Changed it back to the way it was before

src/components/cursor-controller.js
johnshaughnessy: Yes

src/hub.js
johnshaughnessy: I don't think so. I just need to make sure `sortObjects` is enabled -- I stuck it here because the scene must exist and have a renderer by the time we're entering it.

src/hub.js
johnshaughnessy: yes

src/components/cursor-controller.js
```diff
@@ -259,79 +257,67 @@ AFRAME.registerComponent("cursor-controller", {
...
+    // Update the ray and cursor positions
+    const raycasterComp = this.el.components.raycaster;
+    const raycaster = raycasterComp.raycaster;
     const camera = this.data.camera.components.camera.camera;
+    const cursor = this.data.cursor;
+    this.mousePos.set(
+      this.activeTouch.clientX / window.innerWidth * 2 - 1,
+      -(this.activeTouch.clientY / window.innerHeight) * 2 + 1
+    );
     raycaster.setFromCamera(this.mousePos, camera);
```
netpro2k: Probably better as its own PR but we are really jumping through some hoops here to use an aframe raycaster when we really really want it to just act like a ThreeJS raycaster.

src/components/in-world-hud.js
```diff
@@ -7,6 +7,12 @@ AFRAME.registerComponent("in-world-hud", {
     this.mic = this.el.querySelector(".mic");
     this.freeze = this.el.querySelector(".freeze");
     this.bubble = this.el.querySelector(".bubble");
+    this.background = this.el.querySelector(".bg");
+    const renderOrder = window.APP.RENDER_ORDER;
+    this.mic.object3DMap.mesh.renderOrder = renderOrder.HUD;
```
netpro2k: Nit, but at first glance I was confused why these all were teh same, I would maybe call it HUD_CONTNET or HUD_ICONS

src/hub.js
netpro2k: I know we were already doing this, but is there a cost to turning this on and does it increase only with the cost of non-zero renderOrder objects or just total number of objects in the scene?

johnshaughnessy: It scales with the total number of objects in the render list. See 
https://github.com/mrdoob/three.js/blob/master/src/renderers/WebGLRenderer.js#L1123
and
https://github.com/mrdoob/three.js/blob/36444d5660685cf153b99e49c126b5043e4edefd/src/renderers/webgl/WebGLRenderLists.js#L5

src/components/cursor-controller.js
johnshaughnessy: Agreed

src/hub.js
```diff
@@ -86,6 +86,11 @@ if (qs.quality) {
 } else {
   window.APP.quality = isMobile ? "low" : "high";
 }
+window.APP.RENDER_ORDER = {
```
cvan: thanks for this! :+1:

johnshaughnessy:
@cvan Issues #266 and #252 are code cleanliness and have no change in behavior.
The issue with losing the ability to pickup a duck can be reproduced by pressing the primary action button down while hovering over the UI, then releasing the primary action button while not hovering over the UI. It is fixed here: https://github.com/mozilla/hubs/pull/372/commits/1c45644ad28f966b9e1a47df391e0b85d9c7519a
The bug where the HUD wasn't rendering properly can be seen by looking up at it with a VR headset on, and noticing the background, icons, and cursor flip their z ordering based on viewing angle, leading to artifacts where the cursor disappears or the freeze button cutting a square out of the background.

PR #380, opened 2018-05-09
Update audio-context-fix to work in Chrome 66

robertlong:
Removed the user agent check so we do this in all browsers and handle the mouse up event in addition to the touchend event.


src/utils/audio-context-fix.js
```diff
@@ -0,0 +1,22 @@
+/**
+ * Mobile Safari and Chrome will start Audio contexts in a "suspended" state.
```
cvan: maybe rearrange the wording to just `Chrome and Safari`?

src/utils/audio-context-fix.js
```diff
@@ -0,0 +1,22 @@
...
+ * A user interaction (touch/mouse event) is needed in order to resume the AudioContext.
+ */
+
+document.addEventListener("DOMContentLoaded", () => {
+  const ctx = THREE.AudioContext.getContext();
```
cvan: I see this (other) error in Chrome: related?

> `three.js:38150 Uncaught (in promise) DOMException: Cannot decode detached ArrayBuffer`

image

PR #381, opened 2018-05-10
cardboard-controls

InfiniteLee:
added carboard-controls component and hooked it up to cursor-controller. This should properly fix #343

### DIFF
--- a/src/components/cardboard-controls.js
+++ b/src/components/cardboard-controls.js
@@ -1,0 +1,34 @@
+const CARDBOARD_BUTTON = "Cardboard Button";
+
+module.exports = AFRAME.registerComponent("cardboard-controls", {
+  init: function() {
+    this.buttons = {};
+    this.gamepad = null;
+  },
+
+  tick: function() {
+    this.gamepad = this.gamepad || this._getGamepad();
+    if (this.gamepad) {
+      for (var i = 0; i < this.gamepad.buttons.length; i++) {
+        if (this.gamepad.buttons[i].pressed && !this.buttons[i]) {
+          this.el.emit("cardboardbuttondown", {});
+        } else if (!this.gamepad.buttons[i].pressed && this.buttons[i]) {
+          this.el.emit("cardboardbuttonup", {});
+        }
+        this.buttons[i] = this.gamepad.buttons[i].pressed;
+      }
+    } else if (Object.keys(this.buttons)) {
+      this.buttons = {};
+    }
+  },
+
+  _getGamepad: function() {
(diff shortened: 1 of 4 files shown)


src/components/cardboard-controls.js
```diff
@@ -0,0 +1,34 @@
...
+    if (this.gamepad) {
+      for (let i = 0; i < this.gamepad.buttons.length; i++) {
+        if (this.gamepad.buttons[i].pressed && !this.buttons[i]) {
+          this.el.emit("cardboardbuttondown", {});
+        } else if (!this.gamepad.buttons[i].pressed && this.buttons[i]) {
+          this.el.emit("cardboardbuttonup", {});
+        }
+        this.buttons[i] = this.gamepad.buttons[i].pressed;
+      }
+    } else if (Object.keys(this.buttons)) {
```
cvan: this will always be truthy. do you want to check its `length`?

src/components/cardboard-controls.js
```diff
@@ -0,0 +1,34 @@
...
+      this.buttons = {};
+    }
+  },
+
+  _getGamepad: function() {
+    const gamepads = navigator.getGamepads && navigator.getGamepads();
```
cvan: I think you should just bail and return `null` if `navigator.getGamepads` is not truthy

src/components/cardboard-controls.js
```diff
@@ -0,0 +1,34 @@
+const CARDBOARD_BUTTON = "Cardboard Button";
```
gfodor: I was confused about what this was for until I got to the part in the code (I assumed it was basically for a UI string) perhaps rename to `CARDBOARD_BUTTON_GAMEPAD_ID`?

src/components/cursor-controller.js
```diff
@@ -93,8 +93,8 @@ AFRAME.registerComponent("cursor-controller", {
     this.data.playerRig.addEventListener(this.data.primaryUp, this._handlePrimaryUp);
     this.data.playerRig.addEventListener(this.data.grabEvent, this._handlePrimaryDown);
     this.data.playerRig.addEventListener(this.data.releaseEvent, this._handlePrimaryUp);
-    this.data.playerRig.addEventListener("gamepadbuttondown", this._handlePrimaryDown);
```
gfodor: I'm guessing these were unused?

InfiniteLee: this was from my previous attempt when trying to use `gamepad-controls`. I left them in because I knew i was going to reuse them for this PR, but wound up deciding on a different event name

PR #384, opened 2018-05-10
Set HTML prefix for smoke outputs

gfodor:
Fixes an issue causing the HTML prefix to always be set to `smoke-`, which causes the app to always use the smoke box for the avatar selector.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -259,7 +259,7 @@ module.exports = () => {
         } else if (plugin instanceof LodashTemplatePlugin) {
           return new LodashTemplatePlugin({
             ...plugin.options,
-            imports: { ...plugin.options.imports, HTML_PREFIX: "smoke-" }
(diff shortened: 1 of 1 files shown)


webpack.config.js
```diff
@@ -256,6 +256,11 @@ module.exports = () => {
               filename: SMOKE_PREFIX + plugin.options.filename
             })
           );
+        } else if (plugin instanceof LodashTemplatePlugin) {
+          return new LodashTemplatePlugin({
+            ...plugin.options,
+            imports: { ...plugin.options.imports, HTML_PREFIX: "smoke-" }
```
cvan: should this use [the `SMOKE_PREFIX` constant](https://github.com/mozilla/hubs/blob/763680/webpack.config.js#L13)?

gfodor: yes, nice catch!

mquander:
This has been obsoleted with recent client build changes.

PR #386, opened 2018-05-11
Allow specifying full url in bot script

brianpeiris:
So that we can use it on smoke. Also disables telemetry for the bots.

### DIFF
--- a/scripts/bot/run-bot.js
+++ b/scripts/bot/run-bot.js
@@ -4,10 +4,10 @@ Usage:
     ./run-bot.js [options]
 
 Options:
-    -u --url=<url>    Url [default: null].
-    -h --host=<host>  Hubs host if url is not specified [default: localhost:8080]
-    -r --room=<room>  Room id [default: null].
-    -h --help         Show this screen.
+    -u --url=<url>    URL
+    -o --host=<host>  Hubs host if URL is not specified [default: localhost:8080]
+    -r --room=<room>  Room id
+    -h --help         Show this screen
 `;
 
 const docopt = require("docopt").docopt;
--- a/scripts/bot/run-bot.js
+++ b/scripts/bot/run-bot.js
@@ -4,8 +4,9 @@ Usage:
     ./run-bot.js [options]
 
 Options:
-    -h --host=<host>  Hubs host [default: localhost:8080]
-    -r --room=<room>  Room id [default: 234234].
+    -u --url=<url>    Url [default: null].
+    -h --host=<host>  Hubs host if url is not specified [default: localhost:8080]
+    -r --room=<room>  Room id [default: null].
     -h --help         Show this screen.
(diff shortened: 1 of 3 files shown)


scripts/bot/run-bot.js
```diff
@@ -4,8 +4,9 @@ Usage:
     ./run-bot.js [options]
 
 Options:
-    -h --host=<host>  Hubs host [default: localhost:8080]
-    -r --room=<room>  Room id [default: 234234].
+    -u --url=<url>    Url [default: null].
+    -h --host=<host>  Hubs host if url is not specified [default: localhost:8080]
```
cvan: `.` period for consistency

scripts/bot/run-bot.js
```diff
@@ -4,8 +4,9 @@ Usage:
     ./run-bot.js [options]
 
 Options:
-    -h --host=<host>  Hubs host [default: localhost:8080]
-    -r --room=<room>  Room id [default: 234234].
+    -u --url=<url>    Url [default: null].
```
cvan: nit: `Url` -> `URL`

scripts/bot/run-bot.js
```diff
@@ -22,12 +23,19 @@ const querystring = require("query-string");
   page.on("error", err => console.error("ERROR: ", err));
   page.on("pageerror", err => console.error("PAGE ERROR: ", err));
 
+  const baseUrl = options["--url"] || `https://${options["--host"]}/hub.html`;
```
cvan: do you want to throw an error/warning here if `options["--host"]` is missing?

scripts/bot/run-bot.js
johnshaughnessy: ``default: null]` will make `options["--url"]` return the string "null", so that 
```const baseUrl = options["--url"] || `https://${options["--host"]}/hub.html`;```
doesn't do the right thing.

File: scripts/bot/run-bot.js [hunk 4]
```diff
@@ -22,12 +23,19 @@ const querystring = require("query-string");
...
+
   const params = {
-    room: options["--room"],
-    bot: true
+    bot: true,
+    allow_multi: true
   };
-  console.log(params);
-  const url = `https://${options["--host"]}/hub.html?${querystring.stringify(params)}`;
+  const roomOption = options["--room"];
```
johnshaughnessy: Same thing here - `options["--room"]` is the string `"null"`

scripts/bot/run-bot.js
brianpeiris: Host defaults to `"localhost:8080"`.

scripts/bot/run-bot.js
brianpeiris: Oh, good catch. Thanks!

PR #389, opened 2018-05-11
Performance Optimizations

robertlong:
- Add parameter to spawn multiple bots in bot script.
- Migrate away from using events in the audio-feedback components. This resulted in a large performance boost. AFrame's `.emit()` is expensive and allocates memory.
- In `character-controller` we use `object3D.applyMatrix()` multiple times in a single tick. This [method calls](https://github.com/mrdoob/three.js/blob/dev/src/core/Object3D.js#L114) `matrix4.decompose(position, quaternion, scale)` and we only need to update pos/rot/scale once at the end of the matrix multiplications.
- `getWorldPosition()`, `getWorldQuaternion()` and `getWorldDirection()` all call `updateMatrixWorld()`. Instead of calling multiple of these methods back to back we call `updateMatrixWorld()` and then use `matrixWorld` to get the position/quaternion/direction.
- **I need a sanity check on this**, but I don't think we need to call `updateMatrixWorld()` in `personal-space-bubble`.


src/components/audio-feedback.js
```diff
@@ -20,60 +20,38 @@ AFRAME.registerComponent("networked-audio-analyser", {
...
-  init: function() {
-    this.onAudioFrequencyChange = this.onAudioFrequencyChange.bind(this);
-  },
+  tick() {
+    if (!this.mat) return;
 
-  play() {
-    (this.data.analyserSrc || this.el).addEventListener("audioFrequencyChange", this.onAudioFrequencyChange);
-  },
+    const audioAnalyser = this.el.components["networked-audio-analyser"];
```
netpro2k: Probably not a big deal either way, but maybe it makes more sense to store a refference to this in `init()` or `play()`?

src/components/audio-feedback.js
```diff
@@ -20,60 +20,38 @@ AFRAME.registerComponent("networked-audio-analyser", {
...
 
-  pause() {
-    (this.data.analyserSrc || this.el).removeEventListener("audioFrequencyChange", this.onAudioFrequencyChange);
-  },
+    if (!audioAnalyser) return;
```
netpro2k: I would combine the preconditions

src/components/character-controller.js
```diff
@@ -118,22 +118,22 @@ AFRAME.registerComponent("character-controller", {
...
+      root.matrix.premultiply(this.pendingSnapRotationMatrix);
       // Reapply pivot (camera/head) rotation
-      root.applyMatrix(pivotRotationMatrix);
+      root.matrix.premultiply(pivotRotationMatrix);
       // Reapply playspace (player rig) rotation
-      root.applyMatrix(rotationMatrix);
-      // Reapply playspace (player rig) translation
+      root.matrix.premultiply(rotationMatrix);
+      // Reapply playspace (player rig) translation and update pos/rot/scale
       root.applyMatrix(trans);
```
netpro2k: I would just do this a a premultiply and then explicitly call `decompose` unless there is some performacne reason not to. The fact that you had to add the comment points at the side effect being confusing and unexpected.

src/components/cursor-controller.js
```diff
@@ -151,12 +151,18 @@ AFRAME.registerComponent("cursor-controller", {
       this.direction.copy(raycaster.ray.direction);
     } else if ((this.inVR || this.isMobile) && !this.hasPointingDevice) {
       //gaze cursor mode
-      camera.getWorldPosition(this.origin);
-      camera.getWorldDirection(this.direction);
+      camera.updateMatrixWorld(true);
+      this.origin.setFromMatrixPosition(camera.matrixWorld);
+      this.controllerQuaternion.setFromRotationMatrix(camera.matrixWorld);
+      this.direction
```
netpro2k: Probably fine to do in its own PR since irs maybe too divergent from this perf fix, but we can maybe try and get rid of `this.direction` since it seems like duplicate information, but I havent dug deep enough to know if thats true.

src/systems/personal-space-bubble.js
```diff
@@ -68,13 +68,7 @@ AFRAME.registerSystem("personal-space-bubble", {
   tick() {
     if (!this.data.enabled) return;
 
-    // Update matrix positions once for each space bubble and space invader
```
netpro2k: Hmm, so.. I do think this ends up being superfulous in our current app, because the character controller happens to do this already AND the rigs are actually already in world space. We could probably accidentally introduce a bug down the road though if we repaterented rigs and/or updated their position/rotation without updating the matrix. Since it sounded like this was a big performacne bottleneck its probably worth it, but maybe we should drop a comment here for the one of us in the future that is investigating a bug in space bubble positioning :laughing:

johnshaughnessy: I agree these calls to updateMatrixWorld seems unnecessary. You added them 7 months ago and if you don't remember having a specific reason for it, and the space bubble still works, I think it's fine to remove.

netpro2k: I explained above that it's not that we don't remember why it isnt needed, it just happens to not be needed. It is certianly more "correct" to do the update, but it happens to be supurflous given the current configuration of things in our app. Luckily some work is being done in ThreeJS to make superflous updates like this free, so it will be correct to do the "correct" thing again.

src/components/audio-feedback.js
robertlong: I think this is fine.

src/systems/personal-space-bubble.js
mquander: I added the comment you wished for.

Issue #251, opened 2018-04-23
Cursor lock when dragging camera


robertlong:
When you currently drag to look we do not lock the cursor. This causes your cursor to move outside the window and the user can become confused as to where their cursor went.

gfodor:
Did we decide on the exact behavior we want?

cvan:
(let me know if you would like me to file a separate issue as a feature request.) has the [Pointer Lock API](https://developer.mozilla.org/en-US/docs/Web/API/Pointer_Lock_API) been considered? it is popular for FPS games on the Web.

johnshaughnessy:
#391 Doesn't solve this by locking the cursor on drag, but it at least allows people to toggle cursor lock on right click.

johnshaughnessy:
Marking this as closed because we've explored the solutions and chosen one that works OK. Cursor lock when dragging camera is better than the solution we chose, except that a notification about cursor lock would appear every single time you click, and thus we only lock on right click.

Issue #84, opened 2018-03-25
Implement better 2d phone controls


gfodor:
the current "dual axis" controls on mobile work but are not ideal. we should implement the control scheme where single finger drag orbits camera and dual finger pinch-to-zoom moves forwards and backwards.

robertlong:
The pinch to zoom control scheme seems like it would require some sort of tutorial. It's not immediately obvious how to move forward and backward with that control scheme.

gfodor:
yeah I agree a simple popup overlay until the user activates the feature should suffice.

gfodor:
Specifically here is the request for this ticket:

- Touching and dragging on the screen above the stick area rotates the camera

- The gyro, if available, rotates the camera but at small ratio of real world to virtual world radians. Gross head movement should rely upon user touch.

- Pinch and un-pinch to move forward/backward

- Ensure interactables work

- Sticks should still work if you thumb down on the bottom of the screen as usual

- Stretch goal: simple tooltip before user first uses the pinch action to instruct them about it

Use cases to ensure work properly:

- I'm using my phone flat on my desk, but can still navigate the space easily
- I'm using my phone held at an angle in my hand, but can still navigate easily
- I'm using my phone held straight up, and want to be able to "look at" people

johnshaughnessy:
I was looking considering taking on hammer.js as a dependency, but the basic examples are not working in chrome. https://github.com/hammerjs/hammer.js/issues/1130

PR #391, opened 2018-05-12
Improve 2D touch controls, refactor cursor-controller, remove look-controls

johnshaughnessy:
For https://github.com/mozilla/hubs/issues/84
### Requirements
- [x] Touching and dragging on the screen above the stick area rotates the camera
- [x] The gyro, if available, rotates the camera but at small ratio of real world to virtual world radians. 
- [x] Gross head movement should rely upon user touch, the gryo is used for head body expression and minor camera adjustments (for example, looking back and forth between two people)
- [x] Pinch and un-pinch to move forward/backward
- [x] Ensure interactables work
- [x] Sticks should still work if you thumb down on the bottom of the screen as usual
- [ ] Stretch goal: simple tooltip before user first uses the pinch action to instruct them about it

### Use cases to ensure work properly
- [x] I'm using my phone flat on my desk, but can still navigate the space easily
- [x]   I'm using my phone held at an angle in my hand, but can still navigate easily
- [x]  I'm using my phone held straight up, and want to be able to "look at" people

Also for #251 
This doesn't lock the cursor while dragging the camera because doing so would cause a pop up notification to show up every time. But, this does allow a mouse user to right click to toggle cursor lock.

This branch also has several changes to the way we handle cursor and camera interactions on different platforms. See https://github.com/mozilla/hubs/pull/402. The biggest difference is almost all platform-specific logic has been removed from the `cursor-controller`.


johnshaughnessy:
###  Further thoughts on this control scheme
- One problem with the current control scheme is that if you attempt to pick up a duck by touching and dragging with one finger, but you miss or the duck is too far away, then your touch-and-drag moves the camera. This feels bad. Possible solutions:
- - Change the way we interact with ducks: "click" (rapid touch) to pickup, click again to release.
- - Change the camera controls to require two fingers.

- Once I became familiar with pinching to move, I wanted continue moving in the direction I was going after the initial pinch-points were established, so that I did not have to stop to raise my fingers, move them, and press them down again. I would like to implement this alternative. 

- The joysticks, interactables, and the rest of the phone surface all compete for screen real estate, and makes the controls feel cumbersome to me. One possible solution is to provide a menu option for users to enable/disable the joysticks or the multi-touch on the screen. Another solution is to create a hybrid control scheme, as described below:

### Joystick / multitouch hybrid 
A control scheme that will give novice users the same ease-of-use benefits but will allow us to remove the joysticks from the bottom of the screen without losing functionality can be implemented as follows:
- Move forward and backwards by moving my thumbs outward and inward from my initial touch points.
- Move left and right by moving my thumbs like a steering wheel from my initial touch points.
- Rotate left, right, up, and down by moving my thumbs in the same direction.
- Touch and drag a duck to fling it.
- Click a duck to pick it up and keep it in your hand.
- (While the duck is in your hand) Scale the duck by pinching.
- (While the duck is in your hand) Move the duck closer and further by touching (off the duck) and dragging one finger up and down.
- (While the duck is in your hand) Touch and drag the duck to fling it.

src/components/cursor-controller.js
```diff
@@ -26,6 +25,9 @@ AFRAME.registerComponent("cursor-controller", {
   init: function() {
     this.inVR = false;
     this.isMobile = AFRAME.utils.device.isMobile();
+    if (this.isMobile) {
+      this._setCursorVisibility(false);
```
InfiniteLee: Won't this mean you won't see your cursor on mobile vr devices until your first click on something?

johnshaughnessy:
This branch introduces some undesirable weirdness:
- Internal members of `look-controls` are accessed to override the default behavior.
- Additional code paths are added to `cursor-controller` to handle touch events.
These are addressed in another PR which attempts to simplify input handling across devices by removing much of the logic of `cursor-controller` and replacing `look-controls` with the much simpler `camera-controller`. See https://github.com/mozilla/hubs/pull/402

src/components/cursor-controller.js
```diff
@@ -2,451 +2,190 @@ const TARGET_TYPE_NONE = 1;
...
-    this._handleWheel = this._handleWheel.bind(this);
-    this._handleEnterVR = this._handleEnterVR.bind(this);
-    this._handleExitVR = this._handleExitVR.bind(this);
-    this._handlePrimaryDown = this._handlePrimaryDown.bind(this);
-    this._handlePrimaryUp = this._handlePrimaryUp.bind(this);
-    this._handleModelLoaded = this._handleModelLoaded.bind(this);
-    this._handleCursorLoaded = this._handleCursorLoaded.bind(this);
-    this._handleControllerConnected = this._handleControllerConnected.bind(this);
-    this._handleControllerDisconnected = this._handleControllerDisconnected.bind(this);
+    this.forceCursorUpdate = this.forceCursorUpdate.bind(this);
```
netpro2k: I usually only bother binding things if I need to use the bound function for some reason in the same class. If another class needs the bound version it should be responsible for binding it to an instance. And if nothing needs the bound version (which seems to be the case here) then nothing binds it.

src/components/cursor-controller.js
```diff
@@ -2,451 +2,190 @@ const TARGET_TYPE_NONE = 1;
...
-    this.data.playerRig.addEventListener("cardboardbuttondown", this._handlePrimaryDown);
-    this.data.playerRig.addEventListener("cardboardbuttonup", this._handlePrimaryUp);
-
-    this.data.playerRig.addEventListener("model-loaded", this._handleModelLoaded);
-
-    this.el.sceneEl.addEventListener("controllerconnected", this._handleControllerConnected);
-    this.el.sceneEl.addEventListener("controllerdisconnected", this._handleControllerDisconnected);
+  update: function() {
+    if (this.data.rayObject) {
+      this.rayObject = this.data.rayObject.object3D;
```
netpro2k: This is odd, why the extra refference? If you are just doing it to save typing, I would make a local variable in tick.

src/components/cursor-controller.js
```diff
@@ -2,451 +2,191 @@ const TARGET_TYPE_NONE = 1;
...
-  pause: function() {
-    document.removeEventListener("touchstart", this._handleTouchStart);
-    document.removeEventListener("touchmove", this._handleTouchMove);
-    document.removeEventListener("touchend", this._handleTouchEnd);
-    document.removeEventListener("touchcancel", this._handleTouchEnd);
-    document.removeEventListener("mousedown", this._handleMouseDown);
-    document.removeEventListener("mousemove", this._handleMouseMove);
-    document.removeEventListener("mouseup", this._handleMouseUp);
-    document.removeEventListener("wheel", this._handleWheel);
+  tick: (() => {
```
InfiniteLee: iirc, this will cause issues if we ever have more than one cursor-controller (e.g. one for each hand).

src/components/cursor-controller.js
```diff
@@ -2,451 +2,190 @@ const TARGET_TYPE_NONE = 1;
...
+        this.el.setAttribute("line", { start: this.origin.clone(), end: this.data.cursor.object3D.position.clone() });
       }
-      this.wasPhysicalHandGrabbing = isPhysicalHandGrabbing;
-      if (isPhysicalHandGrabbing) return;
-    }
+    };
+  })(),
 
-    //set raycaster origin/direction
+  setRaycasterWithMousePos() {
```
InfiniteLee: This (and the following methods) should be setRaycasterWithMousePos: function() {`, no?

src/components/cursor-controller.js
johnshaughnessy: It's true that the quaternion `rayObjectRotation` will be shared between cursor instances, but since the only way we use it is to temporarily store the ray object's rotation quaternion and then immediately apply it to `this.direction`, it's OK that it's shared.

src/components/cursor-controller.js
johnshaughnessy: 👍

src/components/cursor-controller.js
johnshaughnessy: 👍

src/components/cursor-controller.js
johnshaughnessy: 👍

src/utils/mouse-events-handler.js
```diff
@@ -0,0 +1,115 @@
...
+
+export default class MouseEventsHandler {
+  constructor(cursor, cameraController) {
+    this.cursor = cursor;
+    this.cameraController = cameraController;
+    this.isLeftButtonDown = false;
+    this.isLeftButtonHandledByCursor = false;
+    this.isPointerLocked = false;
+
+    this.addEventListeners = this.addEventListeners.bind(this);
```
InfiniteLee: don't need to bind

src/utils/touch-events-handler.js
```diff
@@ -0,0 +1,165 @@
...
+
+  tearDown() {
+    document.removeEventListener("touchstart", this.handleTouchStart);
+    document.removeEventListener("touchmove", this.handleTouchMove);
+    document.removeEventListener("touchend", this.handleTouchEnd);
+    document.removeEventListener("touchcancel", this.handleTouchEnd);
+  }
+
+  handleTouchStart(e) {
+    this.cursor.setCursorVisibility(false);
```
InfiniteLee: Do this in input-configurator

src/hub.js
```diff
@@ -225,6 +230,7 @@ const onReady = async () => {
 
   const enterScene = async (mediaStream, enterInVR, hubId) => {
     const scene = document.querySelector("a-scene");
+    scene.style.cursor = "none";
```
InfiniteLee: update hub.scss

src/behaviours/trackpad-scrolling.js
```diff
@@ -0,0 +1,58 @@
...
+    this.el.addEventListener(this.move, this.onMove);
+    this.el.addEventListener(this.end, this.onEnd);
+  },
+  removeEventListeners: function() {
+    this.el.removeEventListener(this.start, this.onStart);
+    this.el.removeEventListener(this.move, this.onMove);
+    this.el.removeEventListener(this.end, this.onEnd);
+  },
+  onStart: function(e) {
+    console.log(e);
```
InfiniteLee: remove

src/hub.js
johnshaughnessy: Currently we hide the cursor when the `render-target-loaded` event fires because `look-controls` adds the `a-grab-cursor` class to the `a-canvas` in the scene: https://github.com/aframevr/aframe/blob/master/src/components/look-controls.js#L344
Since we don't want to hide the cursor until we get past the initial menus, should I change this line to add a css class to the scene el and then having a selector for that class with a rule to disable the cursor in the scss file?
e.g. in hub.js `scene.classList.add(".no-cursor")` and in hub.scss `.no-cursor { cursor: none; }`

src/components/input-configurator.js
```diff
@@ -0,0 +1,172 @@
...
+    this.hasPointingDevice = false;
+    this.cursor = this.data.cursorController.components["cursor-controller"];
+    this.gazeTeleporter = this.data.gazeTeleporter.components["teleport-controls"];
+    this.cameraController = this.data.camera.components["pitch-yaw-rotator"];
+    this.playerRig = this.data.playerRig;
+    this.handedness = "right";
+
+    this.onEnterVR = this.onEnterVR.bind(this);
+    this.onExitVR = this.onExitVR.bind(this);
+    this.tearDown = this.tearDown.bind(this);
```
InfiniteLee: no need to bind this either

PR #396, opened 2018-05-21
Add visual indicator to bubble toggling and invading and tweak params

### LINKED ISSUE #328: improve mechanics of Personal-Space Bubble
IMO it's not visually obvious (especially when you're alone/not near others) what the Personal-Space Bubble does / how it works.

([here](https://github.com/mozilla/hubs/blob/master/src/systems/personal-space-bubble.js) is the relevant code.)

netpro2k:
Shows an expanding bubble on toggling the space bubble on/off and flashes the bubble (with a cooldown) when personal space is invaded. This should hopefully help users understand what the bubble toggle does and how the bubble works in general.

Also moved the space bubble to the chest instead of head, bumped the radius to 1m and decreased the transparency of invading avatars further.

Fixes #328

### DIFF
--- a/src/hub.html
+++ b/src/hub.html
@@ -256,7 +256,6 @@
                 class="camera"
                 camera
                 position="0 1.6 0"
-                personal-space-bubble="radius: 0.4"
             >
                 <a-entity
                     id="gaze-teleport"
@@ -329,6 +328,7 @@
 
                 <template data-selector=".Neck">
                     <a-entity>
+                        <a-entity personal-space-bubble="radius: 1" position="0 -0.3 0"></a-entity>
                         <a-entity class="nametag" visible="false" text ></a-entity>
                     </a-entity>
                 </template>
--- a/src/systems/personal-space-bubble.js
+++ b/src/systems/personal-space-bubble.js
@@ -1,3 +1,4 @@
+/* global TWEEN */
 const invaderPos = new AFRAME.THREE.Vector3();
 const bubblePos = new AFRAME.THREE.Vector3();
 
@@ -46,9 +47,12 @@ AFRAME.registerSystem("personal-space-bubble", {
     }
   },
 
-  update() {
+  update(oldData) {
     for (let i = 0; i < this.bubbles.length; i++) {
       this.bubbles[i].updateDebug();
+      if (oldData.enabled !== this.data.enabled) {
+        this.bubbles[i].showStateChange(this.data.enabled);
+      }
     }
 
     for (let i = 0; i < this.invaders.length; i++) {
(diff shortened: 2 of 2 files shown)


src/systems/personal-space-bubble.js
```diff
@@ -177,15 +184,78 @@ AFRAME.registerComponent("personal-space-invader", {
...
   },
 
+  setBubbleVisibility(visible) {
+    this.bubble.visible = visible;
+    this.bubble.material.opacity = visible ? 0.5 : 0.0;
+  },
+
+  showStateChange(enabled) {
+    // Bubble will be hidden again at the end of on/off tweens
+    this.setBubbleVisibility(true);
```
brianpeiris: I think you'd want to stop the `fadeOutTween` before this line, since it feels inconsistent if you toggle the bubble in quick succession otherwise.

src/systems/personal-space-bubble.js
```diff
@@ -177,15 +184,78 @@ AFRAME.registerComponent("personal-space-invader", {
...
+      .easing(TWEEN.Easing.Quadratic.In)
+      .onComplete(() => (this.bubble.visible = false));
+
+    this.fadeOutTween = new TWEEN.Tween(this.bubble.material)
+      .to({ opacity: 0 }, 250)
+      .delay(300)
+      .onComplete(() => (this.bubble.visible = false));
+
+    this.bubbleOnTween = new TWEEN.Tween(this.bubble.scale)
+      .to({ x: 1, y: 1, z: 1 }, 600)
```
johnshaughnessy: I think these transitions happen a bit too quickly, especially the ones that happen when I toggle the button. I suggest bumping bubbleOff/On to take longer than 1000.

src/systems/personal-space-bubble.js
```diff
@@ -177,15 +184,78 @@ AFRAME.registerComponent("personal-space-invader", {
...
+      .easing(TWEEN.Easing.Elastic.InOut)
```
johnshaughnessy: You might like Quadratic.Out so that it quickly gets out of your face and then slowly settles into the scale it ends up at.

src/hub.html
```diff
@@ -329,6 +328,7 @@
 
                 <template data-selector=".Neck">
                     <a-entity>
+                        <a-entity personal-space-bubble="radius: 1" position="0 -0.3 0"></a-entity>
```
johnshaughnessy: Since I'm looking up when I toggle the personal space bubble, I only see (roughly) the second half of the bubble's animation because it starts below my camera and grows outward. I wonder if there's a way (with a cylinder, or something jim makes, or a different center point) to make this look better assuming the user will be looking up when they are discovering what it is.

PR #400, opened 2018-05-22
Fix for logging not being turned on

gfodor:
I'm not sure how or why I thought this was working before -- but it seemed to not be.

### DIFF
--- a/src/hub.html
+++ b/src/hub.html
@@ -19,8 +19,8 @@
     <% } %>
 
     <script>
-        // HACK this has to run after A-Frame but before our bundle, since A-Frame blows away the local storage setting
-        if (/[&?]debug=true/.test(document.location.search) && typeof(localStorage) !== 'undefined') {
+        // HACK: this has to run after A-Frame but before our bundle, since A-Frame blows away the local storage setting
+        if (/[&?]debug=true/.test(window.location.search) && typeof localStorage !== "undefined") {
             localStorage.debug = "naf-janus-adapter:*";
         }
     </script>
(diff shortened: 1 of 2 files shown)


src/hub.html
```diff
@@ -17,6 +17,13 @@
     <% } else { %>
         <script src="https://cdn.rawgit.com/aframevr/aframe/3e7a4b3/dist/aframe-master.js" integrity="sha384-EaMOuyBOi9ERV/lVDwQgz/yFWBDWPsIju5Co6oCZZHXuvbLBO81yPWn80q0BbBn3" crossorigin="anonymous"></script>
     <% } %>
+
+    <script>
+        // HACK this has to run after A-Frame but before our bundle, since A-Frame blows away the local storage setting
+        if (/[&?]debug=true/.test(document.location.search) && typeof(localStorage) !== 'undefined') {
```
cvan: `typeof(localStorage)` -> `typeof localStorage`

cvan: generally better to just use [`window.location`](https://www.w3.org/TR/html50/browsers.html#the-location-interface), as opposed to `document.location`

src/hub.html
```diff
@@ -17,6 +17,13 @@
     <% } else { %>
         <script src="https://cdn.rawgit.com/aframevr/aframe/3e7a4b3/dist/aframe-master.js" integrity="sha384-EaMOuyBOi9ERV/lVDwQgz/yFWBDWPsIju5Co6oCZZHXuvbLBO81yPWn80q0BbBn3" crossorigin="anonymous"></script>
     <% } %>
+
+    <script>
+        // HACK this has to run after A-Frame but before our bundle, since A-Frame blows away the local storage setting
```
cvan: `HACK` -> `HACK:`

src/hub.html
cvan: ' -> " (for prettier linting style)

src/hub.html
cvan: can you [file an A-Frame issue](https://github.com/aframevr/aframe/issues/new) to handle this?

src/hub.html
gfodor: interesting, why is that?

PR #401, opened 2018-05-23
explicitly switch between dynamic and static body types

InfiniteLee:
This should be a performance boost when there are many non-owned interactables in the scene because there are calculations that are run only on bodies set to dynamic in CANNON. See instances of `if(this.type !== Body.DYNAMIC)` in https://github.com/schteppe/cannon.js/blob/master/src/objects/Body.js

`aframe-physics-system` PR to allow this here: https://github.com/donmccurdy/aframe-physics-system/pull/97

### DIFF
--- a/src/components/super-networked-interactable.js
+++ b/src/components/super-networked-interactable.js
@@ -13,7 +13,7 @@ AFRAME.registerComponent("super-networked-interactable", {
     NAF.utils.getNetworkedEntity(this.el).then(networkedEl => {
       this.networkedEl = networkedEl;
       if (!NAF.utils.isMine(networkedEl)) {
-        this.el.setAttribute("body", { type: "dynamic", mass: 0 });
+        this.el.setAttribute("body", { type: "static", mass: 0 });
       } else {
         this.counter.register(networkedEl);
       }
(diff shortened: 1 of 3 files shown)


src/components/super-networked-interactable.js
```diff
@@ -46,7 +46,7 @@ AFRAME.registerComponent("super-networked-interactable", {
     this.hand = e.detail.hand;
     if (this.networkedEl && !NAF.utils.isMine(this.networkedEl)) {
       if (NAF.utils.takeOwnership(this.networkedEl)) {
-        this.el.setAttribute("body", { mass: this.data.mass });
+        this.el.setAttribute("body", { type: "dynamic", mass: this.data.mass });
```
netpro2k: Since we are actually toggling the type now we should be able to get rid of `mass` from this component's schema and just never touch the body's mass.

netpro2k: Since we are actually toggling the type now we should be able to get rid of `mass` from this component's schema and just never touch the body's mass.

InfiniteLee: toggling the type doesn't change the need for changing mass. Changing the mass is still required to get the behavior of switching between dynamic and static. Changing the type simply avoids unnecessary methods from being called on non-owned objects.

InfiniteLee:
I just tried removing the mass property from `super-networked-interactable` and unfortunately I don't think we can get rid of that yet. The problem is you do need to know the original mass of the entity (in order to set it back to the correct value when you grab it). The remote client has no way of knowing what the original mass of the object is except through the `super-networked-interactable` property, however.

PR #407, opened 2018-05-26
Add now entry type

gfodor:
This PR makes it so you can add `_now` to the end of the `vr_entry_type` query string parameter to skip the audio setup (and potentially other steps if we add them later.) Very useful for quick debug cycle.

### DIFF
--- a/src/react-components/ui-root.js
+++ b/src/react-components/ui-root.js
@@ -159,11 +159,11 @@ class UIRoot extends Component {
   handleForcedVREntryType = () => {
     if (!this.props.forcedVREntryType) return;
 
-    if (this.props.forcedVREntryType === "daydream") {
+    if (this.props.forcedVREntryType.startsWith("daydream")) {
       this.enterDaydream();
-    } else if (this.props.forcedVREntryType === "vr") {
+    } else if (this.props.forcedVREntryType.startsWith("vr")) {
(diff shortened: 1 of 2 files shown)


src/react-components/ui-root.js
```diff
@@ -411,19 +411,23 @@ class UIRoot extends Component {
...
     this.setState({ showProfileEntry: false });
   };
 
-  beginAudioSetup = () => {
-    this.setState({ entryStep: ENTRY_STEPS.audio_setup });
+  beginOrSkipAudioSetup = () => {
+    if (!this.props.forcedVREntryType || !this.props.forcedVREntryType.endsWith("_now")) {
+      this.setState({ entryStep: ENTRY_STEPS.audio_setup });
+    } else {
+      setTimeout(this.onAudioReadyButton, 3000); // Need to wait otherwise input doesn't work :/
```
brianpeiris: I think you were running into this setTimeout?

https://github.com/mozilla/hubs/blob/master/src/hub.js#L402

If you bypass that for `qs.vr_entry_type.endsWith("_now")`, you don't need this delay.

gfodor: I think i'll probably leave this PR as-is, since it seems better to keep the `_now` check in one place

PR #409, opened 2018-05-26
Add hubs.local as valid domain for webpack dev server

gfodor:
In order to support running the full stack on a VM, this PR (along with https://github.com/mozilla/reticulum/pull/39) sets things up so you can set `hubs.test` in your local `hosts` file to any IP you'd like to use for testing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,12 @@ To bundle javascript and generate the html templates, run:
 yarn build
 ```
 
+## hubs.local Host Entry
+
+When running the full stack for Hubs (which includes [Reticulum](https://github.com/mozilla/reticulum))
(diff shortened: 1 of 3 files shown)


brianpeiris:
Well, I guess it does improve the workflow in a way. It just means that you have to deal with your OS's hosts file.

gfodor:
@brianpeiris should have mentioned, the benefit to using a proper domain like this is that it allows you to respect the CSP headers (once the relevant reticulum patch goes out) -- so you will be able to use the full normal flow as long as you are able to get `hubs.local` to resolve to the intended LAN IP. in the VM case this is necessary -- it would be nice to have a way to add a hosts entry to an android device but I am guessing that isn't possible without rooting

PR #418, opened 2018-06-05
Bump A-Frame to recent master

mquander:
This bumps A-Frame from start-of-May 0.8.2 to start-of-June 0.8.2. Motivation is to get the three.js r92 -> r93 upgrade, which comes with a relevant performance fix.

I tested simple operation in Ubuntu Firefox and Chrome with voice, movement, and duck in 2D. It would be great if anyone wanted to test in VR and with other controller hardware.


johnshaughnessy:
I didn't find any issues in VR with Oculus Rift and Touch.

mquander:
@johnshaughnessy Updated to latest A-Frame as of today, you should check that your fix is in there if you like.

PR #422, opened 2018-06-06
Media sharing tools

netpro2k:
Basic support for image/video/gltf interactables so we can start experimenting with them. Only really practical on desktop since there is no built in funcitonality for sourcing media, but instead media is inserted by url (by either pasting into the page, or using the button in the 2D HUD).

The media tools are disabled by default and must be flagged on with the `mediaTools` query paramter. Users do not need the paramter to see media, just to insert it.

This also includes the concept of "sticky" objects and "sticky zones". Sticky objects (all interactables for now) will stay flaoting in the air when released with minimal movement. They also will snap into "sticky zones" when they collide with them. Sticky zones can be put into environments.

All media requests (except those coming from our servers) are proxied through our media proxy, farspark. This is done primarily to ensure CORS headers are set correctly, but also has some added features for resolving media from some urls like imgur, giphy, and deviantart. It also has (and will continue to gain) the ability to parse certain kinds of files like PDFs into images, though the client doesnt yet make use of these.

GIFs were particularly hair to deal with. SInce there is no native support in WebGL for rendering GIF frames, the GIF frames must be manually parsed in javascript and then played back frame by frame on a texture. The parsing is done in a worker so should have fairly minimal impact on perf, and also serves as a nice simple test of using workers in our codebase. The parsing code is mostly pulled from `aframe-gif-texture` and is a bit messy, but does the job for now. We will likely want to investigate some performance enhancments like batching multiple GIF frames into single textures and using UV offsets to animate them, rather than swapping textures for each GIF frame.

Lastly this contians some fixes to `networked-counter` that were exposed while testing this, since its much easier to add and remvoe counted objects.

Some known followups that are out of scope for this PR but we want to do in the enar future:
- Get the duck spawners using this same code. Spawners should simply point at a gltf model they want to sapwn and then come in just like any other model using the media tools does.
- Now that there are more things to manipulate, the need for better manipulation tools, especially on non 6DOF devices is quite evident and needs some thinking
- Measure and improve performance. Only some basic profiling was done, but there are likely things that can get better, particularly with videos and gifs.
- Add some sort of loading indicator, epsecially during the initial reticulum/farspark lookup step, since this has no feedback at all
- Find a better solution for the hack added in `offset-relative-to`. This likely will be a networked-aframe patch.
- Actually start experiemnting with sticky-zones in some environments.

TODO:
- [x] Split `image-plus` up into multiple components for resizing, sleep behavior, and reorientation
- [x] Figure out how to deal with CSP issues
- [x] Error handling
- [x] Remvoe hardcoded sticky zones
- [x] Remove physics debugging
- [x] Whitelsit sticky zoens for use in gltf mdoels
- [x] Query param to flag these features on
- [x] Use the same delete button positioning as 3d objects
- [x] Consolidate content-type/url guessing into one spot
- [x] Make sure CSP gets updated to include farspark
- [x] Make sure media resolution calls use the right reticulum host
- [x] Figure out why videos continue playing audio after they are removed
- [x] Settle on a good value for sticky-object


src/hub.html
```diff
@@ -207,6 +224,7 @@
 
         <!-- Interactables -->
         <a-entity id="counter" networked-counter="max: 3; ttl: 120"></a-entity>
+        <a-entity id="media-counter" networked-counter="max: 3;"></a-entity>
```
johnshaughnessy: max: 300

src/components/image-plus.js
```diff
@@ -0,0 +1,74 @@
...
+    const geo = this.el.geometry;
+    let width, height;
+    if (geo && geo.width) {
+      if (geo.height && ratio > 1) {
+        width = geo.width / ratio;
+      } else {
+        height = geo.height * ratio;
+      }
+    } else if (geo && geo.height) {
+      width = geo.width / ratio;
```
johnshaughnessy: Unclear to me what the assumptions are for geo.height and geo.width. Seems like if this line runs, geo.width will be 0 or undefined

netpro2k: Ah yeah, this code is from https://github.com/chenzlabs/aframe-src-fit-component/blob/master/index.js never really went back and looked it over. The intent is that if the geometry has a fixed width and height, those are used as the bounds, otherwise if just one of them is defined, that is used as the bounds, otherwise 1 is used as the bounds.

src/components/sticky-object.js
```diff
@@ -0,0 +1,133 @@
+/* global THREE, CANNON, AFRAME */
+AFRAME.registerComponent("remove-object-button", {
```
johnshaughnessy: probably deserves its own file

src/react-components/info-dialog.js
```diff
@@ -184,6 +197,31 @@ class InfoDialog extends Component {
           </div>
         );
         break;
+      case InfoDialog.dialogTypes.add_media:
+        dialogTitle = "Add Media";
+        dialogBody = (
+          <div>
+            <div>Tip: You can paste media urls directly into hubs with ctrl+v</div>
```
johnshaughnessy: When we add this dialog to mobile we may change the tip to not say ctrl+v

src/hub.html
```diff
@@ -33,7 +33,7 @@
     <a-scene
         renderer="antialias: true"
         networked-scene="adapter: janus; audio: true; debug: true; connectOnLoad: false;"
-        physics="gravity: -6;"
+        physics="gravity: -6; debug: true;"
```
InfiniteLee: debug

src/hub.html
```diff
@@ -183,13 +160,75 @@
                     class="interactable"
                     super-networked-interactable="counter: #counter; mass: 1;"
                     body="type: dynamic; shape: none; mass: 1;"
+                    auto-scale-cannon-physics-body
                     grabbable
-                    stretchable="useWorldPosition: true;"
+                    stretchable="useWorldPosition: true; usePhysics: never"
```
InfiniteLee: why `usePhysics: never`?

src/components/image-plus.js
```diff
@@ -0,0 +1,251 @@
...
...2/GWSQQQYZZJBBBhlkkEEGGWSQQQYZZJBBBhlkkEEGGWSQb63j676waO/qqB7sfTwkr87LpUWLmqI7evQEyCBP470W/Xty2/xS8GMdQAYZZJBBBhlkkEEGGWSQQQYZZJBBBhlkkEEGGWSQQQb5f2xg4KMwZ05DLmW5jiPf/xSGZrpjJ2/s+YjpyJGRLJBXpF3KEMj1WsPGyZC8cnVmW7I9y3PuT+pwQAYZZJBBBhlkkEEGGWSQQQYZZJBBBhlkkEEGGWSQQQYZZJAj+nXbhhCWJ1Gdf3tdsSGfuxaSq9fj+vIayCCDDDLIIIMMMsgggwwyyCCDDDLIIIMMMsgggwwyyCCDDDLI2SD/sm0iTJbORXXliW3hr+aGqC6sf6z+IL+Wtimy538MySOfxtU00J0+lsbI7gL5tiD/HC6Xzkd1tdQRvXpffKmp2JuGnj2drrR7YtuRFHxABhlkkEEGGWSQQQYZZJBBBhlkkEEGGWSQQQYZZJBBriXIk6Wz4UrpcFR/PvVh+D3FHNOF9cvzg/zAqpDcs7L63ffi3mTZnuaoFn98f05+FqedzBDIWatgjl29K3szcoPc0JDXWSZv1cBC2Jw4QwRkkEEGGWSQQQYZZJBBBhlkkEEGGWSQQQYZZJBBBhnkaVX5ObuyNyOmyr+zK68aiC0L5Llz522dQlftlsY/j/Gnp57Jf1Yun3kO5BqCnA39uejVu1IGyNeHh882zvSymULuyHDNfSCDDDLIIIMMMsgggwwyyCCDDDLIIIMMMsgggwwyyCCDfGufbdoX3r1384z23oLXJ0Mp2RHbTRgxne4YHh6+E+RZALmtbUNem2qydCkp8IAMMsgggwwyyCCDDDLIIIMMMsgggwwyyCCDDDLIIIMM8vQh9/TsCq2t66IqlVZMJDcP/a52XbUAbmxs/IX0nrxR7crl8WMg5ww5Y31FXjlTcOUaOC4CZJBBBhlkkEEGGWSQQQYZZJBBBhlkkEEGGWSQQQYZ5FkDeWRkvH10dHys2qUPur3IkFNEu/K4bzne42JDNrNmQDYgg2xABtmAbAzIBmSQDcggG5BBNiAbkGsR8pq0xsju8Iz/mbsz3Ld6a009Qq63k+trZf5IZv69HbUSyCCDDDLIIIMMMsgGZJBBBhlkkEEGGWSQQQYZZJBBBhnkQkH+Gygybr0mifUnAAAAAElFTkSuQmCC";
```
InfiniteLee: why is this base64 hard-coded into this class?

src/components/networked-counter.js
```diff
@@ -6,130 +6,90 @@
...
-    const grabEventListener = this._onGrabbed.bind(this, id);
-    const releaseEventListener = this._onReleased.bind(this, id);
+    const grabEventListener = this._onGrabbed.bind(this, el);
+    const releaseEventListener = this._onReleased.bind(this, el);
 
-    this.queue[id] = {
-      ts: now,
-      el: networkedEl,
+    this.registeredEls.set(el, {
+      ts: Date.now(),
```
InfiniteLee: this is called `timeout` in remove() but `ts` here

src/components/image-plus.js
netpro2k: Don't want to actually request an image since its an error image adn never want that to fail. I did mean to use webpack to do the inlining though. Will fix.

src/hub.html
netpro2k: Yep, have that in the todo items to remove, its left on to help test during review

src/hub.html
johnshaughnessy: The `stretchable` component's adjustment of the cannon body's shapes' scale is redundant with `auto-scale-cannon-physics-body`, but cannot be used at the same time because of the way it calculates the desired scale (It does something like current (cannon shape) scale * delta, instead of current (entity) scale * shape component's half-extents or radius).
Stretchable thus changes the scales of the objects in cannon land but those changes don't get networked. I wrote `auto-scale-cannon-physics-body` to solve this problem by telling super-hands to only adjust the entity's scale and then change the cannon shapes' scale independently (and on both ends of the network).

src/components/auto-scale-cannon-physics-body.js
```diff
@@ -0,0 +1,27 @@
...
+    const scale = this.el.object3D.scale;
+    // Note: This only checks if the LOCAL scale of the object3D changes.
+    if (!almostEquals(0.001, scale, this.prevScale)) {
+      this.prevScale.copy(scale);
+      this.nextUpdateTime = t + 100;
+    }
+
+    if (this.nextUpdateTime > 0 && t > this.nextUpdateTime) {
+      this.nextUpdateTime = -1;
+      this.body.updateCannonScale();
```
johnshaughnessy: https://github.com/MozillaReality/aframe-physics-system/pull/1/files#diff-05cac4c808e28a4aa74a21cbd940a996R139

src/components/remove-networked-object-button.js
```diff
@@ -0,0 +1,21 @@
+AFRAME.registerComponent("remove-networked-object-button", {
+  init() {
+    this.onClick = () => {
+      // const texture = this.targetEl.getObject3D("mesh").material.map;
```
InfiniteLee: remove unused code

src/hub.html
InfiniteLee: 👍

src/components/sticky-object.js
```diff
@@ -0,0 +1,123 @@
...
+  },
+
+  play() {
+    this.el.addEventListener("grab-start", this._onGrab);
+    this.el.addEventListener("grab-end", this._onRelease);
+    this.el.addEventListener("body-loaded", this._onBodyLoaded);
+  },
+
+  pause() {
+    this.el.removeEventListener("grab-start", this._onGrab);
```
InfiniteLee: should `remove` be called in this case?

src/vendor/GLTFLoader.js
```diff
@@ -9,6 +9,8 @@
  * @author Don McCurdy / https://www.donmccurdy.com
```
InfiniteLee: may want to amend this, given the changes

src/components/image-plus.js
```diff
@@ -0,0 +1,252 @@
...
+
+      if (textureCache.has(url)) {
+        const cacheItem = textureCache.get(url);
+        texture = cacheItem.texture;
+        cacheItem.count++;
+      } else {
+        if (url === "error") {
+          texture = errorTexture;
+        } else if (contentType === "image/gif") {
+          console.log("load gif", contentType);
```
InfiniteLee: are these console.logs necessary for production?

src/components/image-plus.js
```diff
@@ -0,0 +1,252 @@
...
+    });
+  },
+
+  loadVideo(url) {
+    return new Promise((resolve, reject) => {
+      const videoEl = createVideoEl(url);
+
+      const texture = new THREE.VideoTexture(videoEl);
+      texture.minFilter = THREE.LinearFilter;
+      videoEl.addEventListener("loadedmetadata", () => resolve(texture));
```
InfiniteLee:
```
videoEl.addEventListener("loadedmetadata", () => resolve(texture), {once: true});
```

src/components/sticky-object.js
```diff
@@ -0,0 +1,123 @@
...
+    this.worldPosition = new THREE.Vector3();
+    this.el.object3D.getWorldQuaternion(this.worldQuaternion);
+    this.el.object3D.getWorldPosition(this.worldPosition);
+
+    const dir = new THREE.Vector3(0, 0, 5).applyQuaternion(this.el.object3D.quaternion);
+    this.bootImpulsePosition = new CANNON.Vec3(0, 0, 0);
+    this.bootImpulse = new CANNON.Vec3();
+    this.bootImpulse.copy(dir);
+
+    this.el.addEventListener("collisions", e => {
```
InfiniteLee: This listener should be cleaned up in `remove` (move the callback into its own bound function?)

src/components/auto-scale-cannon-physics-body.js
```diff
@@ -0,0 +1,27 @@
...
+  init() {
+    this.body = this.el.components["body"];
+    this.prevScale = this.el.object3D.scale.clone();
+    this.nextUpdateTime = -1;
+  },
+
+  tick(t) {
+    const scale = this.el.object3D.scale;
+    // Note: This only checks if the LOCAL scale of the object3D changes.
+    if (!almostEquals(0.001, scale, this.prevScale)) {
```
InfiniteLee: `0.001` should maybe be defined in the schema

src/components/auto-scale-cannon-physics-body.js
```diff
@@ -0,0 +1,27 @@
...
+      this.prevScale.copy(scale);
+      this.nextUpdateTime = t + 100;
```
InfiniteLee: `100` should also probably be defined in the schema

src/workers/gifparsing.worker.js
```diff
@@ -0,0 +1,72 @@
+/**
```
InfiniteLee: I know this was mostly copied from that gtk2k repo, but this file really could use some documentation

src/components/image-plus.js
johnshaughnessy: no I think not

src/components/image-plus.js
johnshaughnessy: woah, that's useful. I didn't know about that `once` flag

src/workers/gifparsing.worker.js
johnshaughnessy: Are you telling me that
`    pos += 13 + +!!(gif[10] & 0x80) * Math.pow(2, (gif[10] & 0x07) + 1) * 3;`
isn't _self documenting_?

jk. I didn't read this code. I'd have to refer to http://www.onicos.com/staff/iz/formats/gif.html or something similar to understand it. @netpro2k ?

johnshaughnessy: This also seems like a good resource (but I haven't read it). http://commandlinefanatic.com/cgi-bin/showarticle.cgi?article=art011

src/components/sticky-object.js
johnshaughnessy: Hmm... how come? I thought something that's stuck would remain stuck through the pause.

src/components/auto-box-collider.js
```diff
@@ -0,0 +1,46 @@
+AFRAME.registerComponent("auto-box-collider", {
+  schema: {
+    resize: { default: false },
+    resizeLength: { default: 0.5 }
+  },
+
+  init() {
+    this.onLoaded = this.onLoaded.bind(this);
+    this.el.addEventListener("model-loaded", this.onLoaded);
```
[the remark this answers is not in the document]
johnshaughnessy: Good point. It might get removed before `model-loaded` fires. I moved it to `remove`. That's sufficient because `onLoaded` calls `removeAttribute`.

PR #426, opened 2018-06-15
Performance tweaks

mquander:
This fixes up the raycaster and character controller to be faster, shaving a few percent off total frame time.


johnshaughnessy:
Nice job!

Issue #181, opened 2018-04-11
Hubs client does not load in WebKit Webviews (Chrome/Slack on iOS)


brianpeiris:
Could also be a problem with Android Slack? We'll have to detect and bust out of the webview into Safari.

robertlong:
How do you get the client to open up in this webview? Links always open up in Firefox for me.

robertlong:
Hmm, I still can't reproduce this on Android or iOS. They both open in a browser for me.

robertlong:
Ok, there's a setting in Slack to set what browser to use when opening links. Mine was set to Safari on iOS and Firefox on Android.

I don't think there is any way to bust out of the webview on iOS.

On Android this works:
```
<html>
  <head></head>
  <script>
    document.location = "intent://google.com#Intent;scheme=http;action=android.intent.action.VIEW;end;";
  </script>
  <body>
  </body>
</html>
```

Should we do this hack on Android or just try to show a message to open the page in Safari/Chrome/Firefox if we detect that they are in a webview?

robertlong:
Removing this from the MVP as there isn't a clean way to do this.

cvan:
I encounter this same issue using Chrome for iOS too. Investigating a solution here.

brianpeiris:
I think we settled on showing a message at least, unless there is a proper solution.

cvan:
> I think we settled on showing a message at least, unless there is a proper solution.

haven't note Remote Debugged it yet, what have you determined the errors/issues are? any Console/Sentry errors?

cvan:
wow, so I just spent hours trying to figure out how to remotely debug on iOS with anything other than Safari. that includes any WebKit webviews (i.e., those used in Chrome, Slack, etc.).

I've tried the following with zero success:

- [`google/ios-webkit-debug-proxy`](https://github.com/google/ios-webkit-debug-proxy)
- [building an iOS WebView app in XCode](https://stackoverflow.com/a/25817718) (simulator got stuck here, and it wasn't exposed from Safari's `Develop > iPhone Simulator` menu)
    
![image](https://user-images.githubusercontent.com/203725/39460367-c2324d2a-4cb7-11e8-8467-a7b95e12c162.png)
- [`RemoteDebug/remotedebug-ios-webkit-adapter`](https://github.com/RemoteDebug/remotedebug-ios-webkit-adapter)
- [`artygus/webkit-webinspector`](https://github.com/artygus/webkit-webinspector)
[...]

PR #432, opened 2018-06-19
Check if browser is an iOS uiWebView

InfiniteLee:
If it is, don't allow user to enter in 2D. Instead, prompt the user to copy the URL and open it in Safari. "fix" for #181

### DIFF
--- a/src/react-components/info-dialog.js
+++ b/src/react-components/info-dialog.js
@@ -13,7 +13,7 @@ class InfoDialog extends Component {
     slack: Symbol("slack"),
     email_submitted: Symbol("email_submitted"),
     invite: Symbol("invite"),
-    safari: Symbol(""),
+    safari: Symbol("safari"),
     updates: Symbol("updates"),
     report: Symbol("report"),
     help: Symbol("help"),
@@ -163,9 +163,7 @@ class InfoDialog extends Component {
         dialogTitle = "Open in Safari";
         dialogBody = (
           <div>
-            <div>
-              Hubs is not supported in your current browser on iOS. Copy and paste this link directly in Safari.
-            </div>
+            <div>Hubs does not support your current browser on iOS. Copy and paste this link directly in Safari.</div>
             <div className="invite-form">
               <input
                 type="text"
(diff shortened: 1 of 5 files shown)


src/react-components/info-dialog.js
```diff
@@ -13,6 +13,7 @@ class InfoDialog extends Component {
     slack: Symbol("slack"),
     email_submitted: Symbol("email_submitted"),
     invite: Symbol("invite"),
+    safari: Symbol(""),
```
brianpeiris: Might as well use `Symbol("safari")` here. It makes for nicer debugging at least.

src/react-components/info-dialog.js
```diff
@@ -155,6 +156,35 @@ class InfoDialog extends Component {
...
+                onFocus={e => e.target.select()}
+                value={document.location}
+                className="invite-form__link_field"
+              />
+              <div className="invite-form__buttons">
+                {navigator.share && (
+                  <button className="invite-form__action-button" onClick={this.shareLinkClicked}>
+                    <span>Share</span>
+                  </button>
+                )}
```
brianpeiris: Does this actually show in Chrome? I would remove it.

src/react-components/info-dialog.js
```diff
@@ -155,6 +156,35 @@ class InfoDialog extends Component {
           </div>
         );
         break;
+      case InfoDialog.dialogTypes.safari:
+        dialogTitle = "Open in Safari";
+        dialogBody = (
+          <div>
+            <div>
+              Hubs is not supported in your current browser on iOS. Copy and paste this link directly in Safari.
```
brianpeiris: Nitpick: I think I'd prefer: "Hubs does not support your current browser on iOS."

src/react-components/info-dialog.js
```diff
@@ -155,6 +156,35 @@ class InfoDialog extends Component {
...
+                value={document.location}
+                className="invite-form__link_field"
+              />
+              <div className="invite-form__buttons">
+                {navigator.share && (
+                  <button className="invite-form__action-button" onClick={this.shareLinkClicked}>
+                    <span>Share</span>
+                  </button>
+                )}
+                <button className="invite-form__action-button" onClick={this.copyLinkClicked}>
```
brianpeiris: `copyLinkClicked` will not include query parameters

PR #445, opened 2018-07-02
Add fly

johnshaughnessy:


johnshaughnessy:
There is a problem where if I ratchet turn a few times, then fly around, my forward direction is wrong. This code seems wrong. Besides that, I've been trying this out on my own, and find this to be a very frustrating way to move around. I keep needing to move along my local y axis (or the global y axis), and rotating my view up or down and then moving forward/backwards is a very awkward/annoying way to accomplish this.

PR #446, opened 2018-07-04
Create only necessary A-Frame entities for GLTF models

mquander:
WIP - poorly tested and I want to fix up the IK thing so we can pull this trick on the head too.

If we know ahead of time which entities will have components on them or a template attached to them, then we can identify scene graph subtrees which can act as leaves of the entity tree and refrain from inflating extra entities underneath them. Roughly cuts in half the number of entities per avatar thanks to all the hand bones that don't need entities.


mquander:
Incidentally, I expect this version to also be a lot faster because of how it handles DOM manipulation (it attaches children to parents in bulk instead of one-at-a-time, and it builds the entity tree from the bottom up, so the whole thing is detached from the DOM until the very end when it is all glommed on at once.) I don't know whether that moves the needle on anything, but it's pleasant.

PR #449, opened 2018-07-05
Add sound effects to the teleport

johnshaughnessy:
Do you _love_ it?


tony056:
Cool! Hope we can add more sound effects to different interactions.

johnshaughnessy:
Dom does *not* _love_ it

Issue #378, opened 2018-05-09
Explore ways to introduce a fixed frame of reference during smooth locomotion


gfodor:
In order to reduce sim sickness, we should consider putting some fixed geometry around the user when moving during smooth motion. Some ideas would be to just show an enlarged avatar body around you, or a transparent sphere with some light dust/blemishes. I'm guessing there might be good examples in other apps we can use, and it would probably help a lot for those of us that get sick when using the feature.

PR #450, opened 2018-07-06
Enhancement: Tunnel Vision

tony056:
- Target Issue: [#378](https://github.com/mozilla/hubs/issues/378)
- Changes:
  - Included essential shaders/postprocessing files found in the example of THREE.js to `utils/postprocessing` & `utils/shaders`.
  - Created `systems/tunnel-effect.js`, an Aframe system that applies vignette postprocessing while the character is moving.


gfodor:
Adding @johnshaughnessy

src/hub.html
```diff
@@ -319,7 +318,8 @@
                     drawIncrementally: true;
                     incrementalDrawMs: 600;
                     hitOpacity: 0.3;
-                    missOpacity: 0.2;"
+                    missOpacity: 0.2;
```
johnshaughnessy: Don't need these changes to teleport-controls attributes in this PR

src/utils/postprocessing/EffectComposer.js
```diff
@@ -0,0 +1,167 @@
+THREE.EffectComposer = function(renderer, renderTarget) {
+  this.renderer = renderer;
+  this.delta = 0;
+  window.addEventListener("vrdisplaypresentchange", this.resize.bind(this));
```
johnshaughnessy: Need to disable this effect when not in VR.

src/systems/tunnel-effect.js
```diff
@@ -0,0 +1,140 @@
...
+
+    if (!this._isPostProcessingReady()) {
+      return;
+    }
+
+    this.characterVelocity = this.characterComponent.velocity;
+    if (this.characterVelocity.distanceTo(STATIC) < CLAMP_VELOCITY) {
+      // the character stops, so we use the aframe default render func
+      this.scene.renderer.render = this.originalRenderFunc;
+      this.isMoving = false;
```
johnshaughnessy: The effect seems to "pop" at the end, when you stop moving. Can we fade away quickly and continuously? It only seems to be a problem when I _stop_ moving. When I _start_ moving, it seems to fade in nicely.

tony056:
- enabled it after entering in VR
- removed the teleport parameters.
- left the switch between postprocessing and regular aframe renderer as the same.

Actually, both starting and stopping are 'pop'. The starting part looks good cuz you don't expect when it will show up and the radius becomes smaller and smaller after it starts. 

Currently, I don't come up with any good approach to fade out after the avatar stops since the vignette shader renders two circles together and no matter how big the radius is, you still can see the gap between two circles.

tony056:
- Done:
  - updated fade out method
  - no new fun in `bindRenderFunc`
  - turned off tunnel vision after exitVR

- Didn't do:
  - create a new global var in character-controller and track it in the tunnel vision.
  - new approach to import the EffectComposer/Pass files.

src/systems/tunnel-effect.js
```diff
@@ -0,0 +1,164 @@
...
+    if (characterSpeed < CLAMP_SPEED) {
+      // the character stops, so we use the aframe default render func
+      const r = this.vignettePass.uniforms["radius"].value;
+      const softness = this.vignettePass.uniforms["softness"].value;
+      if (
+        this.isMoving &&
+        Math.abs(r - TARGET_RADIUS) > CLAMP_RADIUS &&
+        Math.abs(softness - TARGET_SOFTNESS) > CLAMP_SOFTNESS
+      ) {
+        if (this.deltaR !== 0 && this.deltaS !== 0) {
```
johnshaughnessy: I think this line will only be true once, when `this.deltaR` and `this.deltaS` are `undefined.`  Is that desired?

src/systems/tunnel-effect.js
```diff
@@ -0,0 +1,164 @@
...
+  _initPostProcessing: function(event) {
+    if (event.detail.name === this.data.targetComponent) {
+      this.characterEl.removeEventListener("componentinitialized", this._initPostProcessing);
+      this.characterComponent = this.characterEl.components[this.data.targetComponent];
+      this._initComposer();
+    }
+  },
+
+  _enterVR: function() {
+    this.isVR = true;
```
johnshaughnessy: Unfortunately if you press the "f" key without a headset plugged in, the "entervr" event still fires. You go into full screen, but you're not actually in VR. This is not the biggest deal because if you press F11 you can go into fullscreen mode without going into VR, but at some point we may want to fix this.

src/systems/tunnel-effect.js
```diff
@@ -0,0 +1,164 @@
+import "../utils/postprocessing/EffectComposer";
+import "../utils/postprocessing/RenderPass";
+import "../utils/postprocessing/ShaderPass";
+import "../utils/postprocessing/MaskPass";
+import "../utils/shaders/CopyShader";
+import "../utils/shaders/VignetteShader";
+import qsTruthy from "../utils/qs_truthy";
+
+const isDisabled = qsTruthy("disableTunnel");
```
johnshaughnessy: On by default, then disabled with a query parameter will get you the most feedback. Add a note in README.md under the ##Query Params header.

src/systems/tunnel-effect.js
```diff
@@ -74,7 +74,7 @@ AFRAME.registerSystem("tunneleffect", {
         Math.abs(r - TARGET_RADIUS) > CLAMP_RADIUS &&
         Math.abs(softness - TARGET_SOFTNESS) > CLAMP_SOFTNESS
       ) {
-        if (this.deltaR !== 0 && this.deltaS !== 0) {
+        if (!this.deltaR && !this.deltaS) {
           this.deltaR = TARGET_RADIUS - r;
           this.deltaS = softness - TARGET_SOFTNESS;
```
johnshaughnessy: I am still confused about why these are set once with some particular value for `r` and `softness`. Are `deltaR` and `deltaS` supposed to be constants? Do they need to be on `this`? Given their names, I expected them to change whenever `r` or `softness` changed, but perhaps they're better understood to be fixed increments by which we will change `r` or `softness` every frame. In that case, should these constants be defined above with `CLAMP_SPEED`, etc?

tony056: the `deltaR` is defined by the value of `TARGET_RADIUS` and `minRadius`. Since we define the `minRadius` value in schema, I don't think it should be defined as a constant like `CLAMP_SPEED`.

tony056:
solved #378

PR #458, opened 2018-07-17
Remove moment as a dependency

mquander:
This was 400KB of functionality we didn't really need.

The only functionality difference here should be that the `isNewDayWindow` and `isNewMonthWindow` flags are based on fixed intervals of 3600 * 24 and 3600 * 24 * 30 seconds respectively, instead of being one month or day of calendar time. Frankly, this seems better to me for most kinds of analytics anyone would want to do. @gfodor should take a look and decide whether he finds that desirable or undesirable -- we could do it either way if we wanted.

### DIFF
--- a/src/hub.js
+++ b/src/hub.js
@@ -1,5 +1,4 @@
 import "./assets/stylesheets/hub.scss";
-import moment from "moment-timezone";
 import queryString from "query-string";
 
 import { patchWebGLRenderingContext } from "./utils/webgl";
@@ -334,7 +333,7 @@ const onReady = async () => {
       document.body.addEventListener("connected", () => {
         if (!isBotMode) {
           hubChannel.sendEntryEvent().then(() => {
-            store.update({ activity: { lastEnteredAt: moment().toJSON() } });
+            store.update({ activity: { lastEnteredAt: new Date().toISOString() } });
           });
         }
         remountUI({ occupantCount: NAF.connection.adapter.publisher.initialOccupants.length + 1 });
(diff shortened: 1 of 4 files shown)


src/hub.js
```diff
@@ -334,7 +333,7 @@ const onReady = async () => {
       document.body.addEventListener("connected", () => {
         if (!isBotMode) {
           hubChannel.sendEntryEvent().then(() => {
-            store.update({ activity: { lastEnteredAt: moment().toJSON() } });
```
gfodor: unfortunately we can't do this, because this is a persistent schema in local storage. we'll need to either create a new field or write something to migrate the old data. (new clients will throw when the json is passed to the Date constructor I presume)

src/utils/hub-channel.js
```diff
@@ -52,18 +61,15 @@ export default class HubChannel {
...
-    const lastEnteredPst = moment(lastEntered).tz("America/Los_Angeles");
-    const nowPst = moment().tz("America/Los_Angeles");
-    const dayWindowAgo = moment().subtract(1, "day");
-    const monthWindowAgo = moment().subtract(1, "month");
-
-    entryTimingFlags.isNewDaily =
-      lastEnteredPst.dayOfYear() !== nowPst.dayOfYear() || lastEnteredPst.year() !== nowPst.year();
-    entryTimingFlags.isNewMonthly =
-      lastEnteredPst.month() !== nowPst.month() || lastEnteredPst.year() !== nowPst.year();
-    entryTimingFlags.isNewDayWindow = lastEntered.isBefore(dayWindowAgo);
```
gfodor: this seems OK I think -- it changes what we're measuring but we haven't used these fields yet anyhow

src/hub.js
gfodor: (my preference is we write a migration script given that this will break our stats otherwise)

gfodor: ah I misunderstood. i assumed it was returning some kind of JSON blob

PR #462, opened 2018-07-17
Print out Hubs version in console

mquander:
This will make it easier to know that the code you're testing on smoke or production is the code you expect to be testing.


gfodor:
LGTM

Issue #451, opened 2018-07-06
Additional content sharing features


gfodor:
Grab bag of fixes + features for content sharing

Frontend Bugs:
- [x] Google Poly does not load due to failed HEAD request
- [x] Animated GIFs hitting CEF security issue
- [x] Positional audio on videos
- [ ] Mobile perf dropping off at 6 ppl
- [x] Maybe kill videos if they get thrown away/far away
- [x] Some objects were not getting destroyed
- [x] When user pastes a URL in the modal for a UGC site we use the API for, add a "Powered by X" attribution on screen in the modal.
- [x] Fix offsets
- [x] Loading indicator while resolving media
- [x] Rework spawners so they can load any asset
- [x] Support inflation of spawner spawned objects
- [ ] Bizarre shadow map bug when viewing gifs or videos on Oculus Go
- [x] Switching between hands while holding an interactable stops rotation from happening
- [ ] Error images for things that were originally 3d models show delete buttons on all 4 sides insted of just front and back

Backend Bugs:
- [x] Reticulum media api sometimes returns farspark urls for error strings (see comments below)
- [x] Fix 302 redirects
- [ ] Apply for prod giphy API key
- [x] Media API should return "assumed content type" for poly
- [x] Media API should return raw origin URL
- [x] Proxy OPTIONS requests
- [ ] Google drive video fail to return HEAD request so we need to return `expected_content_type` for them

Feature requests:
- [x] Video sync, video pause
- [x] Video volume control
- [x] PDF/slides paging (ok nobody asked for this yet but me :))
- [x] Rotate object at the end of my cursor
- [ ] Snap onto wall surface
- [x] Snap rotate object

netpro2k:
This url:

[...]

PR #463, opened 2018-07-18
Media tools followup fixes

netpro2k:
This PR addresses a bunch of the fixes outlined in https://github.com/mozilla/hubs/issues/451, namely:

- Fixes for loading google poly urls
- Fixes for loading GIFs
- Use positional audio for video/audio objects
- Destroy objects when they "fall off the world"
- Fix bounding boxes for certain models
- Add loading indicators while models/images are loading
- Unify the code used for media tools and spawners (spawners now can specify the url for the media they wish to spawn)
- Support for backend media API changes
- Attribution for Giphy urls in the add media dialog
- Rotate objects at the end of the cursor to face the user

It would probably have been wise to break this up, but most of these were small fixes. The large majority of the work centered around unifying spawners, adding loading indicator, and fixing bounding boxes. Many underlying bugs and race conditions were discovered in aframe-physics-system and super-hands in the process, many of which exist in production today, but just didn't manifest in very noticeable ways. 

There is some additional code cleanup that is probably warranted (particularly position-at-box-shape), though this is in a sufficiently better state than what is in production now that it feels silly to continue holding it back for those.


src/components/position-at-box-shape-border.js
```diff
@@ -1,3 +1,5 @@
+import { getBox } from "../utils/auto-box-collider.js";
```
netpro2k: This file should really be cleaned up, but its localized and I would rather get this PR out and fix this in a followup one.

src/hub.html
```diff
@@ -33,7 +33,7 @@
     <a-scene
         renderer="antialias: true"
         networked-scene="adapter: janus; audio: true; debug: true; connectOnLoad: false;"
-        physics="gravity: -6;"
+        physics="gravity: -6; debug: true;"
```
netpro2k: Gota remember to turn this back off before pushing.

src/hub.html
```diff
@@ -241,7 +203,6 @@
         </a-assets>
 
         <!-- Interactables -->
-        <a-entity id="counter" networked-counter="max: 3; ttl: 120"></a-entity>
```
netpro2k: We no longer have a TTL, but I think with the delete button + `destroy-at-extreme-distances` this is ok.

package.json
```diff
@@ -41,7 +41,7 @@
     "jsonschema": "^1.2.2",
     "moving-average": "^1.0.0",
     "naf-janus-adapter": "^0.10.1",
-    "networked-aframe": "https://github.com/mozillareality/networked-aframe#mr-social-client/master",
+    "networked-aframe": "https://github.com/mozillareality/networked-aframe#bug/aframe-initialization-race-condition",
```
InfiniteLee: can we point back to networked-aframe/master or pull in the latest to the dev fork and keep pointing at that? It got updated with our PR's last week.

src/components/cursor-controller.js
```diff
@@ -97,6 +98,11 @@ AFRAME.registerComponent("cursor-controller", {
       if (this.data.drawLine) {
         this.el.setAttribute("line", { start: this.origin.clone(), end: this.data.cursor.object3D.position.clone() });
       }
+
+      // The curser will always be oriented towards the player about its Y axis, so bjects held by the cursor will rotate towards the player.
```
InfiniteLee: typos:
// The ~~curser~~ cursor will always be oriented towards the player about its Y axis, so ~~bjects~~ objects held by the cursor will rotate towards the player.

src/components/gltf-model-plus.js
```diff
@@ -202,6 +203,7 @@ function cachedLoadGLTF(src, preferredTechnique, onProgress) {
 AFRAME.registerComponent("gltf-model-plus", {
   schema: {
     src: { type: "string" },
+    basePath: { type: "string", default: undefined },
```
InfiniteLee: If basePath is not specified, nothing checks for that case and ultimately will wind up with an error being thrown in `THREE.GLTFLoader` in `cachedLoadGLTF()`. There should probably be a check to catch that case prior to undefined being passed into `THREE.GLTFLoader`.

src/components/gltf-model-plus.js
```diff
@@ -181,11 +181,12 @@ function nextTick() {
   });
 }
 
-function cachedLoadGLTF(src, preferredTechnique, onProgress) {
+function cachedLoadGLTF(src, basePath, preferredTechnique, onProgress) {
   // Load the gltf model from the cache if it exists.
   if (!GLTFCache[src]) {
     GLTFCache[src] = new Promise((resolve, reject) => {
       const gltfLoader = new THREE.GLTFLoader();
+      gltfLoader.path = basePath;
```
InfiniteLee: see comment on `basePath` below

src/components/image-plus.js
```diff
@@ -180,6 +144,7 @@ AFRAME.registerComponent("image-plus", {
       texture.minFilter = THREE.LinearFilter;
       videoEl.addEventListener("loadedmetadata", () => resolve(texture), { once: true });
       videoEl.onerror = reject;
+      texture.audioSource = this.el.sceneEl.audioListener.context.createMediaElementSource(videoEl);
```
InfiniteLee: We shouldn't attach arbitrary objects to existing three.js constructs. If you need to associate the audioSource with a texture you should create a map in this component.

src/components/image-plus.js
```diff
@@ -227,22 +192,49 @@ AFRAME.registerComponent("image-plus", {
...
+      (texture.image.videoWidth || texture.image.width || 1.0);
+    const width = Math.min(1.0, 1.0 / ratio);
+    const height = Math.min(1.0, ratio);
+
+    const geometry = new THREE.PlaneGeometry(width, height, 1, 1);
+    this.mesh = new THREE.Mesh(geometry, material);
+    this.el.setObject3D("mesh", this.mesh);
+    this.el.setAttribute("shape", {
+      shape: "box",
+      halfExtents: { x: width / 2, y: height / 2, z: 0.05 }
```
InfiniteLee: `0.05` should probably be defined in component schema

src/components/image-plus.js
```diff
@@ -227,22 +192,49 @@ AFRAME.registerComponent("image-plus", {
...
+    });
+    if (this.el.components.body && this.el.components.body.body) {
+      this.el.components.body.syncToPhysics(); // not sure if necessary?
```
InfiniteLee: This doesn't seem like it should be necessary, but verify?

src/components/super-spawner.js
```diff
@@ -1,112 +1,127 @@
...
+  },
 
-    const pos = this.data.useCustomSpawnPosition ? this.data.spawnPosition : this.el.getAttribute("position");
-    entity.setAttribute("position", pos);
-    const rot = this.data.useCustomSpawnRotation ? this.data.spawnRotation : this.el.getAttribute("rotation");
-    entity.setAttribute("rotation", rot);
-    this.el.sceneEl.appendChild(entity);
+  onGrabEnd(e) {
+    this.heldEntities.delete(e.detail.hand);
+    // This tells super-hands we are handling this releae
```
InfiniteLee: // This tells super-hands we are handling this ~~releae~~ release

src/components/super-spawner.js
```diff
@@ -1,112 +1,127 @@
...
  * @namespace network
  * @component super-spawner
  */
 AFRAME.registerComponent("super-spawner", {
   schema: {
-    template: { default: "" },
+    /**
+     * Source of the media asset the spawner will spawn when grabbed. This can be a gltf, video, or image, or a url that the reticiulm media API can resolve to a gltf, video, or image.
+     */
+    src: { default: "https://asset-bundles-prod.reticulum.io/interactables/Ducky/DuckyMesh-438ff8e022.gltf" },
```
InfiniteLee: having this defined here as the default seems wrong- this should get moved into the environment component data

src/hub.html
```diff
@@ -153,57 +152,20 @@
...
-                    stretchable="useWorldPosition: true; usePhysics: never"
-                    hoverable
-                    geometry="primitive: plane"
-                    image-plus
-                    sticky-object="autoLockOnLoad: true; autoLockOnRelease: true;"
-                    position-at-box-shape-border="target:.delete-button;dirs:forward,back"
+                    destroy-at-extreme-distances
                 >
-                    <a-entity class="delete-button" visible-while-frozen>
+                    <a-entity class="delete-button" visible-while-frozen scale="1.00 1.00 1.00">
```
InfiniteLee: isn't scale `1 1 1` by default?

src/utils/auto-box-collider.js
```diff
@@ -0,0 +1,20 @@
+export function getBox(entity, boxRoot) {
```
InfiniteLee: cache rotation and copy instead of clone
```
export function getBox() {
  const rotation = new THEE.Quaternion();
  return function (entity, boxRoot) {
    ...
    rotation.copy(entity.object3D.rotation);
...
```

src/components/position-at-box-shape-border.js
```diff
@@ -1,3 +1,5 @@
...
+
 const PI = Math.PI;
 const HALF_PI = PI / 2;
 const THREE_HALF_PI = 3 * PI / 2;
```
InfiniteLee: `const THREE_HALF_PI = 3 * HALF_PI;`

src/components/gltf-model-plus.js
netpro2k: the intent here is that setting undefined is the same as not setting it. GLTFLoader doesnt require a path to be set.

src/components/super-spawner.js
netpro2k: Yep, the plan would be to update the enviornments after we push this, then we can remove this default. This is just to allow us to push without breaking the existing environments during out push.

src/components/image-plus.js
netpro2k: Since we have already tested it a bunch with this in, going to leave it in for now to avoid missing some edge case. Put in a todo to evaluate though.

PR #466, opened 2018-07-20
Tidying up dependencies, package.json

mquander:
Now the package is named `hubs` instead of `mr-social-client`.


package.json
```diff
@@ -1,11 +1,11 @@
 {
-  "name": "mr-social-client",
+  "name": "hubs",
```
cvan: good fix! while you're at it, would be nice to add a `description` 👍

src/avatar-selector.js
```diff
@@ -26,22 +25,16 @@ addLocaleData([...en]);
 
 registerTelemetry();
 
+const hash = new URLSearchParams(location.hash.replace(/^#/, "?"));
```
cvan: it [won't break](https://caniuse.com/#search=URLSearchParams) for any existing users of Hubs, so feel free to ignore, but figured I'd mention anyway: `URLSearchParams` isn't supported in IE 11 (which I know is not a supported browser), and it was only recently [added in Edge](https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/8993198/) on April 10, 2018.

cvan: not sure if it's any more legible but you could do `(window.location.hash || "#").substr(1)`

src/index.js
```diff
@@ -4,12 +4,11 @@ import ReactDOM from "react-dom";
...
 import InfoDialog from "./react-components/info-dialog.js";
-import queryString from "query-string";
 
-const qs = queryString.parse(location.search);
+const qs = new URLSearchParams(location.search);
 registerTelemetry();
 
 ReactDOM.render(
-  <HomeRoot dialogType={qs.list_signup ? InfoDialog.dialogTypes.updates : null} />,
+  <HomeRoot dialogType={qs.get("list_signup") ? InfoDialog.dialogTypes.updates : null} />,
```
cvan: this seems fine. I'm not sure where this URL is advertised, but Google has https://hubs.mozilla.com/?list_signup=true indexed:

image

if you want to allow `?list_signup`, you could use `qs.has(…)`. figured would make sense to have parity with the `?room` behaviour. feel free to ignore this.

src/avatar-selector.js
```diff
@@ -26,22 +25,16 @@ addLocaleData([...en]);
...
 window.APP = new App();
-const hash = queryString.parse(location.hash);
-const isMobile = AFRAME.utils.device.isMobile();
-if (hash.quality) {
-  window.APP.quality = hash.quality;
-} else {
-  window.APP.quality = isMobile ? "low" : "high";
-}
+window.APP.quality = hash.get("quality") || AFRAME.utils.device.isMobile() ? "low" : "high";
```
cvan: thought: could/should there also be a `const qs = new URLSearchParams(location.search);` so users link to hubs URLs with `&quality=high` in the query-string without having to mess with the hash?

src/react-components/2d-hud.js
```diff
@@ -1,16 +1,15 @@
...
-import queryString from "query-string";
 
 import styles from "../assets/stylesheets/2d-hud.scss";
 
 import FontAwesomeIcon from "@fortawesome/react-fontawesome";
 import faPlus from "@fortawesome/fontawesome-free-solid/faPlus";
 
-const qs = queryString.parse(location.search);
+const qs = new URLSearchParams(location.search);
 function qsTruthy(param) {
```
cvan: I see this duplicated in [`hub.js`](https://github.com/mozilla/hubs/pull/466/files#diff-36e9875da4ba69fe26a79e004718180cR132). could put the method in a common utils file?

src/react-components/ui-root.js
```diff
@@ -281,12 +280,12 @@ class UIRoot extends Component {
...
       const location = window.location;
-      const qs = queryString.parse(location.search);
-      qs.vr_entry_type = "daydream"; // Auto-choose 'daydream' after landing in chrome
+      const qs = new URLSearchParams(location.search);
+      qs.set("vr_entry_type", "daydream"); // Auto-choose 'daydream' after landing in chrome
 
       const intentUrl =
         `intent://${location.host}${location.pathname || ""}?` +
-        `${queryString.stringify(qs)}#Intent;scheme=${(location.protocol || "http:").replace(":", "")};` +
+        `${qs.toString()}#Intent;scheme=${(location.protocol || "http:").replace(":", "")};` +
```
cvan: you can remove the `.toString()` - the template string literal will evaluate `qs` as a string

src/react-components/ui-root.js
```diff
@@ -281,12 +280,12 @@ class UIRoot extends Component {
 
       // We are not in mobile chrome, so launch into chrome via an Intent URL
       const location = window.location;
-      const qs = queryString.parse(location.search);
-      qs.vr_entry_type = "daydream"; // Auto-choose 'daydream' after landing in chrome
+      const qs = new URLSearchParams(location.search);
+      qs.set("vr_entry_type", "daydream"); // Auto-choose 'daydream' after landing in chrome
 
       const intentUrl =
         `intent://${location.host}${location.pathname || ""}?` +
```
cvan: when will `location.pathname` not default to `/`? think you can remove the `|| ""`

src/react-components/ui-root.js
cvan: since `location.protocol` will always be `https:` or `http:`, you could do `location.protocol.slice(0, -1)`; admittedly, that's not as expressive as the `.replace(…)` call. at the very least, you could remove the ` || "http:"` bit.

src/storage/store.js
```diff
@@ -4,7 +4,6 @@ import merge from "deepmerge";
 const LOCAL_STORE_KEY = "___hubs_store";
 const STORE_STATE_CACHE_KEY = Symbol();
 const validator = new Validator();
-import { EventTarget } from "event-target-shim";
```
cvan: awesome change 👍

src/utils/phoenix-utils.js
```diff
@@ -1,15 +1,14 @@
...
 
-  const socketProtocol = qs.phx_protocol || (document.location.protocol === "https:" ? "wss:" : "ws:");
+  const socketProtocol = qs.get("phx_protocol") || (document.location.protocol === "https:" ? "wss:" : "ws:");
   const [retHost, retPort] = (process.env.DEV_RETICULUM_SERVER || "").split(":");
   const isProd = process.env.NODE_ENV === "production";
-  const socketPort = qs.phx_port || (isProd ? document.location.port : retPort) || "443";
-  const socketHost = qs.phx_host || (isProd ? document.location.hostname : retHost) || "";
+  const socketPort = qs.get("phx_port") || (isProd ? document.location.port : retPort) || "443";
+  const socketHost = qs.get("phx_host") || (isProd ? document.location.hostname : retHost) || "";
   const socketUrl = `${socketProtocol}//${socketHost}${socketPort ? `:${socketPort}` : ""}/socket`;
```
cvan: would it make sense to allow passing in a `?phx_url`

src/utils/phoenix-utils.js
```diff
@@ -1,15 +1,14 @@
...
-  const qs = queryString.parse(location.search);
+  const qs = new URLSearchParams(location.search);
 
-  const socketProtocol = qs.phx_protocol || (document.location.protocol === "https:" ? "wss:" : "ws:");
+  const socketProtocol = qs.get("phx_protocol") || (document.location.protocol === "https:" ? "wss:" : "ws:");
   const [retHost, retPort] = (process.env.DEV_RETICULUM_SERVER || "").split(":");
   const isProd = process.env.NODE_ENV === "production";
-  const socketPort = qs.phx_port || (isProd ? document.location.port : retPort) || "443";
-  const socketHost = qs.phx_host || (isProd ? document.location.hostname : retHost) || "";
+  const socketPort = qs.get("phx_port") || (isProd ? document.location.port : retPort) || "443";
```
cvan: I'm probably just misreading this, but should the ` || "43"` be ` document.location.port || "443"` for dev environments?

package.json
mquander: That's true, I didn't go very overboard because I don't think we really have a use case for someone pulling down our package from NPM right now. Do you know whether there are other reasons we should care about having accurate metadata in package.json?

src/avatar-selector.js
mquander: > it won't break for any existing users of Hubs, so feel free to ignore, but figured I'd mention anyway: URLSearchParams isn't supported in IE 11 (which I know is not a supported browser), and it was only recently added in Edge on April 10, 2018.

Yeah, in general I think we are only interested in supporting the following:

- Latest stable Edge, Chrome, and Firefox on Linux, Windows, and OS X
- Latest stable Chrome and Firefox on Android w/VR
- Reasonably recent Android and iOS browsers w/magic window

So I'm not concerned about support.

Personally, I like the regex better, because it's clear to me what it does without having to think to myself "if I call `substr` with one argument is it the start or the end and is it inclusive or exclusive", but I guess it's going to depend on whether you remember what regex `^` is.

src/index.js
mquander: Interesting. I wonder how it got that URL? We should almost certainly take a day to fix up our main page SEO via webmaster tools, sitemap, better meta tags, etc. (I know you have #317 open related to this, supplying OG meta tags, and I intend to go make sure that or similar work gets into master soon.)

src/avatar-selector.js
mquander: I am not sure why this got put into the hash, actually -- I was just mechanically transforming it. It's a good question, though. I will poke around and figure out whether there is some reason for it not to be in the query string.

src/react-components/2d-hud.js
mquander: I think this is a weird and unpredictable technique that I would rather just get rid of, personally, so I'm hesitant to enshrine it somewhere. (I would prefer to declare that the query parameter is truthy if it is present at all, which seems simpler.)

src/react-components/ui-root.js
mquander: > you can remove the .toString() - the template string literal will evaluate qs as a string

Oh, I didn't know that, thanks!

> since `location.protocol` will always be `https:` or `http:`, you could do `location.protocol.slice(0, -1)`; admittedly, that's not as expressive as the `.replace(…)` call. at the very least, you could remove the ` || "http:"` bit.

Hmm, I wonder what the person writing this was thinking -- how could `location.protocol` ever be falsy? I will double-check and probably remove it if I can't find any reason. As for the replacement, I think it's a little clearer the way it is.

src/react-components/ui-root.js
mquander: Good question, reminiscent of the other place [here](https://github.com/mozilla/hubs/pull/466#discussion_r203915251). I will probably remove it.

src/utils/phoenix-utils.js
mquander: That does seem more straightforward. I see no good reason not to do it. I will consider performing that transformation after I dig up where anyone might be using these parameters.

src/utils/phoenix-utils.js
mquander: Honestly, this code doesn't make any sense to me, because I don't understand why the websocket port we're using should be correlated with `document.location.port`. I may try to understand it and interrogate Greg about why it is the way that it is.

mquander:
Thanks for the review, I didn't realize that someone would come and give me a nice feedback on what I did today!

mquander:
I don't want to increase the complexity of this PR with real functionality changes to e.g. questionable query string parameters and stuff, so while the rest of the suggestions are appreciated and will remain at the top of my mind I will put them off for now.

mquander:
I left the `EventTarget` him alone for now -- it turns out that although `EventTarget` generally has broad support, using it as a mixin is dodgier.

PR #467, opened 2018-07-20
Feature/oculus go tlc

gfodor:
This PR closes out a few odds-and-ends:

- Skips the initial step in the entry flow if you are wearing a VR headset already
- Adds a UA test for Firefox Reality
- Stops using `production` mode for webpack dev server in `yarn start`
- Fixes the code entry page, which broke due to an inadvertent dependency on A-Frame which was introduced. (this is OK I guess, because the flow for that page will have the user download A-Frame eventually anyway.)
- Adds the `initial_environment` query string parameter to the landing page, which we are going to need for the user testing in a few weeks

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -10,6 +10,9 @@ const qs = queryString.parse(location.search);
 registerTelemetry();
 
 ReactDOM.render(
-  <HomeRoot dialogType={qs.list_signup ? InfoDialog.dialogTypes.updates : null} />,
+  <HomeRoot
+    initialEnvironment={qs.initial_environment}
+    dialogType={qs.list_signup ? InfoDialog.dialogTypes.updates : null}
+  />,
   document.getElementById("home-root")
 );
--- a/src/link.html
+++ b/src/link.html
@@ -8,6 +8,12 @@
     <title>Enter Code | Hubs by Mozilla</title>
     <link href="https://fonts.googleapis.com/css?family=Zilla+Slab:300,300i,400,400i,700" rel="stylesheet">
     <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no">
+
+    <% if(NODE_ENV === "production") { %>
+    <script src="https://cdn.rawgit.com/aframevr/aframe/1be48d9/dist/aframe-master.min.js" integrity="sha384-SXrfoMHbXpA5RZhIyhgaR6tQ764dDZsbFk3PiokC/tc0+NnW1yaYQMUzWtL06hnq" crossorigin="anonymous"></script>
+    <% } else { %>
+    <script src="https://cdn.rawgit.com/aframevr/aframe/1be48d9/dist/aframe-master.js" integrity="sha384-AmjDGOMbvTrrUFdeVWcBIlXRINIWnO8iwj/4VS21OWbYDsa/7nheOIyPAPJSkR6J" crossorigin="anonymous"></script>
+    <% } %>
 </head>
 
 <body>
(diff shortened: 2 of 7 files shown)


src/index.js
```diff
@@ -10,6 +10,9 @@ const qs = queryString.parse(location.search);
 registerTelemetry();
 
 ReactDOM.render(
-  <HomeRoot dialogType={qs.list_signup ? InfoDialog.dialogTypes.updates : null} />,
+  <HomeRoot
+    initialEnvironment={qs.initial_environment}
+    dialogType={qs.list_signup ? InfoDialog.dialogTypes.updates : null}
+  />,
```
mquander: You're going to need to resolve a conflict with master here since `qs` became a `URLSearchParams`.

src/link.html
```diff
@@ -8,6 +8,12 @@
     <title>Enter Code | Hubs by Mozilla</title>
     <link href="https://fonts.googleapis.com/css?family=Zilla+Slab:300,300i,400,400i,700" rel="stylesheet">
     <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no">
+
+    <% if(NODE_ENV === "production") { %>
+    <script src="https://cdn.rawgit.com/aframevr/aframe/1be48d9/dist/aframe-master.min.js" integrity="sha384-SXrfoMHbXpA5RZhIyhgaR6tQ764dDZsbFk3PiokC/tc0+NnW1yaYQMUzWtL06hnq" crossorigin="anonymous"></script>
+    <% } else { %>
+    <script src="https://cdn.rawgit.com/aframevr/aframe/1be48d9/dist/aframe-master.js" integrity="sha384-AmjDGOMbvTrrUFdeVWcBIlXRINIWnO8iwj/4VS21OWbYDsa/7nheOIyPAPJSkR6J" crossorigin="anonymous"></script>
+    <% } %>
```
mquander: Why did this become necessary?

We really gotta get this into npm.

src/react-components/hub-create-panel.js
```diff
@@ -15,16 +15,25 @@ const HUB_NAME_PATTERN = "^[A-Za-z0-9-'\":!@#$%^&*(),.?~ ]{4,64}$";
...
   };
 
   constructor(props) {
     super(props);
 
+    let environmentIndex = Math.floor(Math.random() * props.environments.length);
+
+    if (props.initialEnvironment) {
+      environmentIndex = props.environments.findIndex(
+        e => e.name.toLowerCase() === props.initialEnvironment.toLowerCase()
```
mquander: I don't like code like this that tries to complicatedly fuzzy match what the inputs were supposed to mean, because I think it adds complexity with the result of making it harder to remember how the code works and more complicated to document. I wouldn't have done it and I would spend the time to actively work to undo it later by identifying and canonicalizing the actual parameters people use.

(This is a disagreement I have in other places in our code too, not just this one.)

mquander: I'm not sure what your use case is for setting the initial environment, but I would consider making an error message pop out here if you got an invalid input (i.e. an environment that doesn't exist) instead of having a mysterious failure down the line.

gfodor: This is intended for us to be able to tell the user study facilitators (perhaps verbally) how to initialize the landing page to a specific environment choice, and so I made it case insensitive (in part based upon the hilarious experiences of trying to get people to type "mediaTools=true" over the last couple of weeks and not getting the casing right)

gfodor: IOW I think a case insensitive match here for the name of the environment is the proper UX for this parameter.

PR #468, opened 2018-07-20
Add proper FxR check

gfodor:
My speculative FxR check is wrong since we are using the Fennec UA. This seems like a sane check but wanted to make sure I wasn't missing something.

To be clear, we are *not* using user agent checks to determine device capabilities -- we're using them to determine if the user is currently already in a HMD. My understanding is there's no other way to do this right now.

### DIFF
--- a/src/utils/vr-caps-detect.js
+++ b/src/utils/vr-caps-detect.js
@@ -50,7 +50,7 @@ export async function getAvailableVREntryTypes() {
   // This needs to be kept up-to-date with the latest browsers that can support VR and Hubs.
   // Checking for navigator.getVRDisplays always passes b/c of polyfill.
   const isWebVRCapableBrowser = window.hasNativeWebVRImplementation;
-  const isFirefoxReality = /Firefox for Android/.test(ua) && isWebVRCapableBrowser;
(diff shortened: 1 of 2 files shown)


src/utils/vr-caps-detect.js
```diff
@@ -46,12 +46,12 @@ export async function getAvailableVREntryTypes() {
   const ua = navigator.userAgent;
   const isSamsungBrowser = browser.name === "chrome" && /SamsungBrowser/.test(ua);
   const isOculusBrowser = /Oculus/.test(ua);
-  const isFirefoxReality = /Firefox Reality/.test(ua);
-  const isInHMD = isOculusBrowser || isFirefoxReality;
 
   // This needs to be kept up-to-date with the latest browsers that can support VR and Hubs.
   // Checking for navigator.getVRDisplays always passes b/c of polyfill.
   const isWebVRCapableBrowser = window.hasNativeWebVRImplementation;
+  const isFirefoxReality = "buildID" in navigator && isWebVRCapableBrowser;
```
cvan: there you go:
```js
const isFirefoxReality = window.orientation === 0 && ("buildID" in navigator) && isWebVRCapableBrowser;
```

PR #470, opened 2018-07-23
Upgrade aframe-physics-system slightly

mquander:
This doesn't have functionality changes that we care about, but I did it because I fixed a bug in aframe-physics-system that was causing our upstream build to pull in a ton of unnecessary packages.

This SHA of aframe-physics-system is of hubs/master. I changed it to a SHA because I think it's better to use a SHA than to peg it to a moving branch.


mquander:
Closing this for now in favor of a PR that does some more upgrades also.

PR #476, opened 2018-07-25
Fix and redesign implementation of bot mode

mquander:
How this worked two weeks ago:

The webpack build would include the bot recording files, which would be built under the /assets/avatars path. That build would get deployed to production and smoke in a way such that the /assets/avatars path was served via smoke-hubs.mozilla.com. Bot mode was hard-coded to fetch from that root-relative path at runtime in order to load the recordings for playback.

Why it didn't work last week:

The production deployment changes we made moved all the built assets to a different Cloudfront distribution, and stopped serving them from hubs.mozilla.com or smoke-hubs.mozilla.com. So the bots couldn't load the recordings. Actually, even if they knew the right URL, it still wouldn't have worked, because Chrome won't allow you to capture a stream from a cross-domain <audio> element and send it over WebRTC (no matter what CORS headers you make.)

How it works now:

Instead of putting the bot recording files into the webpack build, the bot recording files will get copied onto the bot machines along with the hubs repo. Then, puppeteer will upload those files from disk onto the page instead of fetching them remotely. If you run a bot from the browser, you can upload arbitrary files yourself using the file selector on the page.

See https://github.com/mozilla/hubs-ops/pull/26 for changes to the script that runs the squawkers.


brianpeiris:
This looks great! Thanks for digging into it.

PR #477, opened 2018-07-26
Switch to use npm instead of yarn

mquander:
This is appealing because yarn doesn't seem to be better in any appreciable way, it removes a dependency from our build toolchain, and it's consistent with the rest of our projects.


cvan:
nice change, btw. this simplified development and deployment both :+1:

PR #478, opened 2018-07-26
Support for sketchfab models

netpro2k:
This PR adds support for loading sketchfab zipped gltf models that are resolved by the reticulum media api. Zip parsing is done in a worker. Also adds a very simple media api resolver cache to deal with the fact that sketchfab urls may be resolved differently on different instances of reticulum, making duplicates of the same model actually bypass our GLTF caching.


src/utils/media-utils.js
```diff
@@ -5,16 +5,21 @@ if (process.env.NODE_ENV === "development") {
...
-  }).then(r => r.json());
+  const resolved =
+    (parsedUrl.protocol !== "http:" && parsedUrl.protocol !== "https:") || isHostWhitelisted(parsedUrl.hostname)
+      ? { raw: url, origin: url }
+      : await fetch(resolveMediaUrl, {
+          method: "POST",
+          headers: { "Content-Type": "application/json" },
+          body: JSON.stringify({ media: { url } })
+        }).then(r => r.json());
+  // farsparkCache.set(url, resolved);
```
johnshaughnessy: Uncomment to use cache

src/components/gltf-model-plus.js
```diff
@@ -181,14 +183,40 @@ function nextTick() {
...
+            resolve(model);
+          };
+        }
+
+        const gltfLoader = new THREE.GLTFLoader();
+        gltfLoader.path = basePath;
+        gltfLoader.preferredTechnique = preferredTechnique;
+        gltfLoader.load(gltfUrl, onLoad, onProgress, reject);
+      } catch (e) {
+        reject(e);
```
johnshaughnessy: Since we cache the rejected promise, will any future attempts to load this model fail?

src/utils/media-utils.js
```diff
@@ -5,16 +5,21 @@ if (process.env.NODE_ENV === "development") {
...
+  if (farsparkCache.has(url)) return farsparkCache.get(url);
 
-  return await fetch(resolveMediaUrl, {
-    method: "POST",
-    headers: { "Content-Type": "application/json" },
-    body: JSON.stringify({ media: { url } })
-  }).then(r => r.json());
+  const resolved =
+    (parsedUrl.protocol !== "http:" && parsedUrl.protocol !== "https:") || isHostWhitelisted(parsedUrl.hostname)
+      ? { raw: url, origin: url }
```
johnshaughnessy: In this case, we do not "resolve" the media url via farspark, which makes me think `farsparkCache` and `resolveFarsparkUrl` are misnomers. Are there better names for these? 

I haven't thought of better names, but here are some suggestions:
- `resolveUrl`/`urlsCache`
- `resolvedUrlsCache` (ew)
- `resolve` for the function and `urls` for the cache
- `farsparkifyIfNecessary` (ok not this one)

johnshaughnessy: I kind of like `resolve` and `urls` (or `urlsCache`)

johnshaughnessy: But `resolve` doesn't really tell you what's happening. I think it's the 2nd definition here:

resolve: 
(verb) settle or find a solution to (a problem, dispute, or contentious matter).
(verb) decide firmly on a course of action.
(noun) firm determination to do something.

netpro2k: Renamed to `resolveMedia` and `resolveMediaCache`

PR #481, opened 2018-07-26
Fix avatar selector

gfodor:
This fixes an issue that broke the state propagation from the avatar selector -> outer frame -> avatar selector by not re-parsing the URL hash when it changed.


mquander:
Thanks for figuring it out!

Issue #482, opened 2018-07-26
"Enter in VR" prompts to download Firefox in the case that vr hardware is not found


InfiniteLee:
https://i.gyazo.com/72f9beace96c965f2a13c6ae8e4917d4.png
When using Firefox, if you try to click "Enter in VR" when you don't have any VR devices installed (e.g. `navigator.getVRDisplays()`) finds nothing), the message that pops up prompts to download Firefox. 

This should probably be a different message in this case, or we should not show this button at all if no vr device is found.

gfodor:
https://github.com/mozilla/hubs/pull/483

PR #483, opened 2018-07-26
Bug/hmd check wrong for ff mobile

gfodor:
This PR deals with some more entry bugs:

- On Mac/Linux FF, stop showing VR prompt
- Stop misdetecting mobile Firefox as FxR
- Changes copy to for link flow to be more intentional around headset case


gfodor:
closes https://github.com/mozilla/hubs/issues/482

PR #487, opened 2018-07-27
Now we can use htmlhint directly on our HTML

mquander:
#479 caused our "HTML" to be actually valid HTML, so we don't need to jump through any hoops when linting it anymore.


mquander:
@brianpeiris I mentioned getting rid of the lodash templating entirely before, which I may still do (the origin trial token can be AFAIK injected at runtime just as well) but we can do this anyway.

package.json
```diff
@@ -19,7 +19,7 @@
     "doc": "node ./scripts/doc/build.js",
     "prettier": "prettier --write '*.js' 'src/**/*.js'",
     "lint:js": "eslint '*.js' 'scripts/**/*.js' 'src/**/*.js'",
-    "lint:html": "node ./scripts/lint-html.js 'src/**/*.html'",
+    "lint:html": "htmlhint src/**/*.html",
```
brianpeiris: The quotes around the glob might be significant because it will use node glob vs local shell glob

mquander: Interesting, I didn't know this detail. This [article](https://medium.com/@jakubsynowiec/you-should-always-quote-your-globs-in-npm-scripts-621887a2a784) seems to make a good case that I should be using Node glob if I want to aim for cross-platform compatibility, so I'll do that.

PR #493, opened 2018-07-31
Audio feedback fixes

mquander:
Contrary to some documentation, Safari doesn't support `getFloatTimeDomainData` (which is a shame since it's more efficient on Firefox and Chrome, but the efficiency doesn't matter much.)


mquander:
Merging this without further delay since it's a shame for iOS to be busted.

PR #495, opened 2018-08-02
Support file uploads

johnshaughnessy:
Users can drag-and-drop or use the create object dialog to upload files and display them in hubs.
TODO:
- [x] Support gltf/zip, glb
- [ ] Add a top-level button on mobile for uploading images. (Note: I can do this in a separate PR. I want to look into how changing the `accept` attribute of an `input` with `type=file` will change the file selection process on mobile. e.g. `accept=image/*` opens an image picker on android where the image could come from the camera app. Not sure what it looks like on iOS. This is _probably_ the attribute we want.) 
- [x] Adjust the icons and figure out how best to include font-awesome attribution. (I used svg's downloaded from their site, but maybe I should be including that library and using those icons as a font.)
- [x] Adjust the colors used and make sure font and font-size is consistent, reads well, and displays correctly on different screen sizes.


src/assets/stylesheets/create-object-dialog.scss
```diff
@@ -0,0 +1,64 @@
...
+  padding-right: 0%;
+  width: 80%;
+}
+
+:local(.form) {
+  width: 80%;
+}
+
+:local(.hideFileInput) {
+  width: 0.1px;
```
netpro2k: Is there a reason this cant just be `display: none` or `visibility: hidden`?

src/assets/stylesheets/create-object-dialog.scss
```diff
@@ -0,0 +1,64 @@
+@import 'shared';
+
+:local(.smallButton) {
+  padding-right: 5%;
```
netpro2k: These % based paddings seem odd. The buttons are of fixed sizes, why would their padding not be?

src/assets/stylesheets/create-object-dialog.scss
```diff
@@ -0,0 +1,64 @@
...
+  font-size: 2em;
+  align-self: center;
+}
+
+:local(.cancelIcon) {
+  padding-left: 2%;
+  color: white;
+  :hover {
```
netpro2k: Should this be `&:hover`? As it is now this is targeting children of something with the `cancelIcon` class that are hovered on.

src/assets/stylesheets/create-object-dialog.scss
```diff
@@ -0,0 +1,64 @@
...
+    color: #FF3D7F
+  }
+}
+
+:local(.uploadIcon) {
+  color: white;
+  :hover {
```
netpro2k: Same as above, should this be `&:hover`?

src/components/gltf-model-plus.js
```diff
@@ -200,7 +200,7 @@ function cachedLoadGLTF(src, basePath, contentType, preferredTechnique, onProgre
       try {
         let gltfUrl = src;
         let onLoad = resolve;
-        if (contentType === "model/gltf+zip") {
+        if (contentType === "model/gltf+zip" || contentType === "application/x-zip-compressed") {
```
netpro2k: I guess this is ok, its a bit odd though since currently the zip loading code assumes sketchfab's format (namely that there is a scene.gtlf in the root of the zip), so its not technically generally applicable to all zips.

src/components/media-loader.js
```diff
@@ -46,17 +48,23 @@ AFRAME.registerComponent("media-loader", {
...
+      const token = this.data.token;
 
-      this.showLoaderTimeout = setTimeout(() => {
-        const loadingObj = new THREE.Mesh(new THREE.BoxGeometry(), new THREE.MeshBasicMaterial());
-        this.el.setObject3D("mesh", loadingObj);
-        this.setShapeAndScale(true);
-      }, 100);
+      this.showLoaderTimeout =
+        this.showLoaderTimeout ||
+        setTimeout(() => {
```
netpro2k: Maybe move this into `this.showLoader` so this wrapps a little more cleanly and avoids creating this funciton several times even though it densest need to close over anything.

src/utils/media-utils.js
```diff
@@ -22,14 +22,38 @@ export const resolveMedia = async url => {
...
   const scene = AFRAME.scenes[0];
 
   const entity = document.createElement("a-entity");
   entity.id = "interactable-media-" + interactableId++;
   entity.setAttribute("networked", { template: "#interactable-media" });
-  entity.setAttribute("media-loader", { src, resize });
+  entity.setAttribute("media-loader", { resize, src: typeof src === "string" ? src : "" });
   scene.appendChild(entity);
+
+  if (typeof src === "object") {
```
netpro2k: Should probably check specifically for `src instanceof File` here.

src/utils/media-utils.js
```diff
@@ -22,14 +22,38 @@ export const resolveMedia = async url => {
...
 };
 
+export const upload = file => {
+  const formData = new FormData();
+  formData.append("media", file);
+  return fetch(mediaAPIEndpoint, {
+    method: "POST",
+    body: formData
+
+    // We do NOT specify a Content-Type header like so
```
netpro2k: Though I didn't know this off hand, its just documentation for how `FormData` works with the `fetch` API, seems a bit excessive. I would have looked it up if I was confused.

src/components/media-loader.js
```diff
@@ -6,6 +6,8 @@ const fetchContentType = async url => fetch(url, { method: "HEAD" }).then(r => r
 AFRAME.registerComponent("media-loader", {
   schema: {
     src: { type: "string" },
+    token: { type: "string" },
+    contentType: { type: "string" },
```
netpro2k: Hmm, feels like we shouldn't need contentType, part of media-loaders job is figuring that out so it feels a bit out of place to tell it. It should get this back from `resolveMedia` anyway.

src/components/media-loader.js
```diff
@@ -65,9 +73,24 @@ AFRAME.registerComponent("media-loader", {
           },
           { once: true }
         );
-        this.el.setAttribute("image-plus", { src: raw, contentType });
+        let blobUrl;
+        if (token) {
+          const imageResponse = await fetch(raw, {
```
netpro2k: Hmm why are we doing this idependantly for images/gltfs? Seems like it should just be part of determining what url to load. Those with tokens end up loading a blob url.

src/components/media-loader.js
```diff
@@ -65,9 +73,24 @@ AFRAME.registerComponent("media-loader", {
...
+            method: "GET",
+            headers: { Authorization: `Token ${token}` }
+          });
+          const blob = await imageResponse.blob();
+          blobUrl = window.URL.createObjectURL(blob);
```
netpro2k: Don't think these blob urls ever get cleaned up. When to do that is a bit tricky....

src/components/media-loader.js
```diff
@@ -65,9 +73,24 @@ AFRAME.registerComponent("media-loader", {
...
+        }
+        this.el.setAttribute("image-plus", { src: blobUrl || raw, contentType });
         this.el.setAttribute("position-at-box-shape-border", { target: ".delete-button", dirs: ["forward", "back"] });
-      } else if (contentType.startsWith("model/gltf") || url.endsWith(".gltf") || url.endsWith(".glb")) {
+      } else if (
+        contentType.indexOf("application/octet-stream") !== -1 ||
```
netpro2k: Not super in love with gltf's just being the fallback for any octet stream or zip file...

src/hub.js
```diff
@@ -330,11 +330,9 @@ const onReady = async () => {
 
       document.addEventListener("drop", e => {
         e.preventDefault();
-        const imgUrl = e.dataTransfer.getData("url");
-        if (imgUrl) {
-          console.log("Dropped: ", imgUrl);
-          spawnMediaInfrontOfPlayer(imgUrl);
-        }
+        const url = e.dataTransfer.getData("url");
+        const file = e.dataTransfer.files && e.dataTransfer.files[0];
```
netpro2k: I think almost this same exact code will work for pasting files from the clipboard as well. We should try it real quick.

src/react-components/create-object-dialog.js
```diff
@@ -22,52 +40,84 @@ export default class CreateObjectDialog extends Component {
...
     lastUrl = this.state.url;
   }
 
   onUrlChange = e => {
-    if (e && e.target.value && e.target.value !== "") {
+    let url;
+    try {
+      url = new URL(e.target.value);
+    } catch (_) {
       this.setState({
```
netpro2k: This control flow is a bit odd. I would do something like:
```
let attributionImage = this.state.attributionImage;
if(e.target.validity.valid){
      attributionImage = e.target && e.target.value && e.target.validity.valid && attributionHostnames[(new URL(e.target.value)).hostname]
}
this.setState({
      url: e.target && e.target.value,
      attributionImage
})
```

src/react-components/create-object-dialog.js
```diff
@@ -22,52 +40,84 @@ export default class CreateObjectDialog extends Component {
...
+        <FontAwesomeIcon icon={faTimes} />
+      </label>
+    );
+    const uploadButton = (
+      <label htmlFor={fileInputId} className={cx(styles.smallButton, styles.uploadIcon)}>
+        <FontAwesomeIcon icon={faPaperclip} />
+      </label>
+    );
+    const filenameLabel = (
+      <label className={cx(styles.fileNameLabel, "add-media-form__link_field")}>{this.state.fileName}</label>
```
netpro2k: I assume these non local classes are just a carryover? Seems like since they are classes specific to this we should just copy them over into `create-object-dialog.scss`

src/assets/stylesheets/create-object-dialog.scss
johnshaughnessy: No, `visibility: hidden` seems fine.

src/assets/stylesheets/create-object-dialog.scss
johnshaughnessy: Yes I think so. Woops!

src/assets/stylesheets/create-object-dialog.scss
johnshaughnessy: Switched to using `em` values, though the style in general is not looking so good as I vary the screen sizes. Hoping to get some help with this.

src/components/media-loader.js
johnshaughnessy: When `url` matches `uploads-dev.reticulum.io/uploads/`, `resolveMedia` only returns `original` and `raw` (which is the same value). It does not return `meta` with `expected_content_type`.
 
If alternatively I skip `resolveMedia` and call `fetchContentType` (with the authorization header added to the request), the `OPTIONS` returns a `204` and the `HEAD` returns a `401`. Neither of these have a `Content-Type` header.

Running into some difficulty tracing thru this in reticulum. I don't understand that code very well.

src/components/media-loader.js
johnshaughnessy: 👍 DRYed it up

src/components/media-loader.js
johnshaughnessy: If I could pass the `File` to `media-loader`, I could have it use `upload` instead of `resolveMedia`, which is basically what we want. I think we only consider using `resolveMedia`/`fetchContentType` because that's the existing code path inside `media-loader`. Since `media-loader` is an `aframe` `component` whose state we want to sync via `networked-aframe`, we eventually do need to set its `src` and `token` attributes. However, when we set the `src` to the `url` of an uploaded `File`, we don't need to call `resolveMedia` or `fetchContentType`, since the `upload` request returned the expected content type and the url we want to network. 
I'm not sure what the right way to think about this is. On Friday I was trying to think thru whether `media-loader` needs to be a component that does work, since the work we need to do changes depending on what media we want to load, but I didn't think of a clean way to avoid doing work. We want the loading cube to exist the whole time we're uploading or downloading media, which means we want it to be on the same entity that the eventually loaded image/video/model will be attached to.

src/assets/stylesheets/create-object-dialog.scss
```diff
@@ -0,0 +1,74 @@
+@import 'shared';
+
+:local(.addMediaForm) {
```
gfodor: i'm pretty sure we've been naming these styles in `snake-case`, and then the react/webpack stuff will case them in JS land as you have here.

src/components/media-loader.js
gfodor: the reticulum media API should return expected content type in the meta section of the response when you upload a file -- is that not working? was hard to tell from these comments.

src/react-components/create-object-dialog.js
```diff
@@ -1,19 +1,37 @@
...
+import cx from "classnames";
 
 const attributionHostnames = {
-  "giphy.com": giphyLogo
+  "giphy.com": giphyLogo,
+  "media.giphy.com": giphyLogo
 };
 
 const DEFAULT_OBJECT_URL = "https://asset-bundles-prod.reticulum.io/interactables/Ducky/DuckyMesh-438ff8e022.gltf";
+const isMobile = AFRAME.utils.device.isMobile();
```
gfodor: unfortunately this breaks the homepage, which doesn't include aframe. also it's a bad idea to use `AFRAME` without having `import "aframe"` in the file.

gfodor: https://github.com/mozilla/hubs/pull/504

src/components/media-loader.js
gfodor: OK I think I understand now. I will work on a reticulum patch to properly respond to HEAD from the uploads controller, which will include the content type.

gfodor: actually it looks to me like this already works. my guess is you are not passing the authorization header when you make the HEAD request?

gfodor: working on a PR

gfodor: https://github.com/mozilla/hubs/pull/505

src/react-components/create-object-dialog.js
gfodor: actually https://github.com/mozilla/hubs/pull/505

PR #496, opened 2018-08-02
Set things up for dev forks

gfodor:


mquander:
Is there a particular reason we still like to have a separate origin for GLTF bundles and other assets? I might be tempted to make them the same.

gfodor:
Yes, it makes sense to me to separate them -- the origin for the main assets is the bucket target that's part of our Hubs build pipeline (and incidentally assumes that everything is sitting under `/assets`.) Whereas the asset bundles origin is a separate S3 bucket that is tied to our artist pipeline tools, and comes with different requirements/permissions/etc. So they seem pretty conceptually different. This difference will become even more clear when we support hosted scenes, at which point our 1st party asset bundle server will likely go away completely and we will just become consumers of our own hosting services.

src/utils/media-utils.js
```diff
@@ -1,8 +1,9 @@
 const whitelistedHosts = [/^.*\.reticulum\.io$/, /^.*hubs\.mozilla\.com$/, /^hubs\.local$/];
 const isHostWhitelisted = hostname => !!whitelistedHosts.filter(r => r.test(hostname)).length;
 let mediaAPIEndpoint = "/api/v1/media";
-if (process.env.NODE_ENV === "development") {
-  mediaAPIEndpoint = `https://${process.env.DEV_RETICULUM_SERVER}${mediaAPIEndpoint}`;
+
+if (process.env.RETICULUM_SERVER) {
+  mediaAPIEndpoint = `https://${process.env.RETICULUM_SERVER}${mediaAPIEndpoint}`;
```
mquander: It seems like we expect `RETICULUM_SERVER` to always be set now -- can we simplify places like this to instead be e.g.
```Javascript
let mediaAPIEndpoint = "//${process.env.RETICULUM_SERVER}/api/v1/media"
```
?

gfodor: yeah I wasn't sure -- it may be useful to keep things so that if the environment variable isn't set it uses relative paths. (for example, perhaps there are situations where you don't know the domain name at build time?)

mquander:
Makes sense to me!

src/utils/media-utils.js
gfodor: eg i could imagine a model where there's a built version of Hubs that you can throw it behind two different reticulum servers and it works

mquander: Hmm, no idea whether I think that is the case or not, but then at least I would rewrite it without a branch, e.g.
```
let apiHost = process.env.RETICULUM_SERVER || location.host;
let apiEndpoint = `${apiHost}/api/v1/media`
```
I don't care if you change it now though.

mquander: Another random thought -- maybe it would be a nice time to move locations of our API endpoints and any associated utilities for making paths to them into a separate module, since there are a few of them.

gfodor: sounds like a good idea to me, i probably will leave that to a future PR though.

PR #498, opened 2018-08-03
Update GLTFLoader to dev/hubs-v2 + envMap and gamma

robertlong:
Fix all the GLTFLoader bugs!

This PR updates GLTFLoader to our dev branch of three.js: https://github.com/mozillareality/three.js/tree/hubs/dev-v2

It also uses the `gammaOutput` setting in the `WebGLRenderer`, adds a default environment map, and fixes normals.

Old and Busted:
screen shot 2018-08-02 at 7 09 59 pm

New Hotness:
screen shot 2018-08-02 at 7 09 49 pm


src/components/gltf-model-plus.js
```diff
@@ -274,7 +369,12 @@ AFRAME.registerComponent("gltf-model-plus", {
       }
 
       const gltfPath = THREE.LoaderUtils.extractUrlBase(src);
-      const model = await cachedLoadGLTF(src, this.data.basePath, this.data.contentType, this.preferredTechnique);
+
+      if (!GLTFCache[src]) {
+        GLTFCache[src] = loadGLTF(src, this.data.basePath, this.data.contentType, this.preferredTechnique);
+      }
+
+      const model = await GLTFCache[src].then(cloneGltf);
```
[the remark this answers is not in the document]
brianpeiris: Hmm, `delete` and then `throw` a new error, I guess.

src/components/gltf-model-plus.js
```diff
@@ -274,7 +369,13 @@ AFRAME.registerComponent("gltf-model-plus", {
...
 
       const gltfPath = THREE.LoaderUtils.extractUrlBase(src);
-      const model = await cachedLoadGLTF(src, this.data.basePath, this.data.contentType, this.preferredTechnique);
+
+      if (!GLTFCache[src]) {
+        GLTFCache[src] = loadGLTF(src, this.data.basePath, this.data.contentType, this.preferredTechnique);
+      }
+
+      const cachedModel = await GLTFCache[src];
+      const model = cloneGltf(cachedModel);
```
brianpeiris: This will still keep the broken promise in the `GLTFCache`. You need to evict it from the cache if it fails.

robertlong: Oh, right.

src/components/gltf-model-plus.js
```diff
@@ -194,33 +202,122 @@ function getFilesFromSketchfabZip(src) {
...
-          };
-        }
+async function loadEnvMap() {
+  const urls = [cubeMapPosX, cubeMapNegX, cubeMapPosY, cubeMapNegY, cubeMapPosZ, cubeMapNegZ];
+  const texture = await new THREE.CubeTextureLoader().load(urls);
+  texture.format = THREE.RGBFormat;
+  return texture;
+}
+
+function resolveURL(url, path) {
```
netpro2k: I would use URL for all this https://developer.mozilla.org/en-US/docs/Web/API/URL

src/components/gltf-model-plus.js
```diff
@@ -194,33 +202,122 @@ function getFilesFromSketchfabZip(src) {
...
+  const json = parser.json;
+  const images = json.images;
+  const buffers = json.buffers;
+  const materials = json.materials;
+
+  const pendingFarsparkPromises = [];
+
+  if (images) {
+    for (const image of images) {
+      const imagePromise = resolveMedia(resolveURL(image.uri, parser.options.path)).then(({ raw }) => {
```
netpro2k: I would pluck `({raw}) => image.url = raw` out to avoid creating the function repeatedly and the shadowing of `raw`

src/components/gltf-model-plus.js
```diff
@@ -194,33 +202,122 @@ function getFilesFromSketchfabZip(src) {
...
+
+  const json = parser.json;
+  const images = json.images;
+  const buffers = json.buffers;
+  const materials = json.materials;
+
+  const pendingFarsparkPromises = [];
+
+  if (images) {
+    for (const image of images) {
```
netpro2k: I would probably use map here. Even if it does create garbage this isn't really in a night loop so I would guess it would be fine, haven't ever profiled it to see how bad it is though.

src/components/gltf-model-plus.js
netpro2k: Strike me as a bit odd that we are reading out `parser.options.path`, we should probably be using `basePath` and then I wonder why we set path on the loader at all anymore.

src/components/gltf-model-plus.js
```diff
@@ -194,33 +202,122 @@ function getFilesFromSketchfabZip(src) {
...
+      const imagePromise = resolveMedia(resolveURL(image.uri, parser.options.path)).then(({ raw }) => {
+        image.uri = raw;
+      });
+
+      pendingFarsparkPromises.push(imagePromise);
+    }
+  }
+
+  if (buffers) {
```
netpro2k: was gunna suggest to DRY this up, but I think with my above suggestions it becomes a 1 liner anyway so probably fine.

src/components/gltf-model-plus.js
```diff
@@ -194,33 +202,122 @@ function getFilesFromSketchfabZip(src) {
...
+  if (/^data:.*,.*$/i.test(url)) return url;
+
+  // Blob URL
+  if (/^blob:.*$/i.test(url)) return url;
+
+  // Relative URL
+  return path + url;
+}
+
+async function loadGLTF(src, preferredTechnique, onProgress) {
```
netpro2k: I don't love how many (mostly unrelated) things this function ends up doing. It's not really any more than before, and this is cleaner than what we had but still feels a bit messy. I don't really know how I would break it up though since most of the work needs to get done in kind of awkward parts of the gltf loading process.

src/components/gltf-model-plus.js
```diff
@@ -274,7 +369,13 @@ AFRAME.registerComponent("gltf-model-plus", {
       }
 
       const gltfPath = THREE.LoaderUtils.extractUrlBase(src);
-      const model = await cachedLoadGLTF(src, this.data.basePath, this.data.contentType, this.preferredTechnique);
+
+      if (!GLTFCache[src]) {
+        GLTFCache[src] = loadGLTF(src, this.preferredTechnique);
+      }
+
+      const cachedModel = await GLTFCache[src];
```
netpro2k: I wold inline this and get rid of `cachedModel` since its not really valid to use if for anything but cloning. But that's mostly personal preference.

src/components/gltf-model-plus.js
```diff
@@ -301,6 +402,7 @@ AFRAME.registerComponent("gltf-model-plus", {
       this.el.setObject3D("mesh", object3DToSet);
       this.el.emit("model-loaded", { format: "gltf", model: this.model });
     } catch (e) {
+      delete GLTFCache[src];
```
netpro2k: Probably not part of this PR since we weren't doing it before, but we never actually clean up the GLTFCache in non-error cases. For images I do a reference counting cache, we should maybe do that here as well.

src/hub.html
```diff
@@ -24,7 +24,7 @@
     </audio>
 
     <a-scene
-        renderer="antialias: true"
+        renderer="antialias: true; gammaOutput: true; sortObjects: true; physicallyCorrectLights: true;"
```
netpro2k: Any perf costs to these? Particularly sortObjects and physicallyCorrectLights?

src/react-components/avatar-selector.js
```diff
@@ -122,6 +122,12 @@ class AvatarSelector extends Component {
     }
   }
 
+  componentDidMount() {
+    this.scene.addEventListener("loaded", () => {
+      this.scene.setAttribute("renderer", { gammaOutput: true, sortObjects: true, physicallyCorrectLights: true });
```
netpro2k: Any reason not to just put these on the scene element like we do for hub.html?

src/components/media-loader.js
```diff
@@ -78,9 +74,7 @@ AFRAME.registerComponent("media-loader", {
         );
         this.el.addEventListener("model-error", this.onError, { once: true });
         this.el.setAttribute("gltf-model-plus", {
-          src: raw,
-          contentType,
-          basePath: THREE.LoaderUtils.extractUrlBase(origin),
+          src: url,
```
netpro2k: Might warrant a comment here noting that we are passing `url` and not `raw` on purpose and why.

src/components/gltf-model-plus.js
brianpeiris: I see no benefit to using map here.

src/hub.html
robertlong: We're already using `sortObjects` I just moved the setting of that renderer option here.

I believe `physicallyCorrectLights` should have a negligible change in perf. The shader changes a little, but I don't think it's too much more expensive.

src/components/gltf-model-plus.js
```diff
@@ -194,33 +202,101 @@ function getFilesFromSketchfabZip(src) {
...
+  texture.format = THREE.RGBFormat;
+  return texture;
+}
+
+function resolveGLTFUri(gltfProperty, basePath) {
+  return resolveMedia(new URL(gltfProperty.uri, basePath).href).then(({ raw }) => (gltfProperty.uri = raw));
+}
+
+async function loadGLTF(src, preferredTechnique, onProgress) {
+  const { raw, origin, contentType } = await resolveMedia(src);
```
johnshaughnessy: `contentType` isn't returned in the request if the `src` is pointing to a file we uploaded to reticulum. See https://github.com/mozilla/hubs/pull/495#discussion_r208061694

src/components/gltf-model-plus.js
netpro2k: It is looping through an array and mapping it into another array. It just cleans up some of the ceremony. Not a big deal either way.

src/components/gltf-model-plus.js
netpro2k: Should be ok to push this PR before that issue is fixed since contentType is only used for detecting zips.

johnshaughnessy:
Looks good. We will need to adjust this for uploaded media.

PR #499, opened 2018-08-06
Properly set npm debug log filter

gfodor:
Now that we have moved A-Frame into a 1st party vendor bundle, we can set the npm debug logging without this hack. (In fact, the old hack regressed because now A-Frame loads after the head script tags.)

### DIFF
--- a/src/utils/logging.js
+++ b/src/utils/logging.js
@@ -1,15 +1,9 @@
 // A-Frame blows away any npm debug log filters so this allow the user to set the log filter
 // via the query string.
 import debug from "debug";
+import qsTruthy from "./qs_truthy";
 
 const qs = new URLSearchParams(location.search);
-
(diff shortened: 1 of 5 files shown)


src/utils/logging.js
```diff
@@ -0,0 +1,18 @@
...
+// via the query string.
+import debug from "debug";
+
+const qs = new URLSearchParams(location.search);
+
+function qsTruthy(param) {
+  const val = qs.get(param);
+  // if the param exists but is not set (e.g. "?foo&bar"), its value is the empty string.
+  return val === "" || /1|on|true/i.test(val);
+}
```
brianpeiris: `qsTruthy` is already defined in `hub.js`

gfodor: i'll just break it out. i've seen it copy/pasted elsewhere as well

PR #500, opened 2018-08-07
Add test environment env variables

robertlong:
When testing environments developed in `mr-social-assets` @j-conrad ran into some CORS issues with webpack-dev-server. I added two environment variables that allow anyone to test out custom environments on any server by proxying them through webpack-dev-server.

`TEST_ENVIRONMENT_HOST` is the server you wish to proxy.
`TEST_ENVIRONMENT_PATH` is the path to the environment you wish to add to the environment picker.

This should make testing custom environments as easy as modifying the environment variables and running `npm start` in both the hubs and mr-social-assets directories.

When building hubs with `NODE_ENV=production` this code is cleaned up by webpack's dead code elimination.


InfiniteLee:
lgtm

mquander:
Can you write more about the reason for this PR?

Assuming that this is the right solution to Jim's problem, some details:
- Please make sure that whatever problem you're solving, the solution still works given #497 -- that change is necessary to loosen the CORS enough to run local Reticulum and/or Janus.
- Why are there values present for this in .env.defaults?
- Why is `TEST_ENVIRONMENT_HOST` separate from `ASSET_BUNDLE_SERVER`? It doesn't seem like the application should need to know about two different hosts to look for environments on.
- Why `TEST_ENVIRONMENT_HOST` and `TEST_ENVIRONMENT_PATH` instead of `ENVIRONMENT_HOST` and `ENVIRONMENT_PATH`, and why host and path instead of one URL?

In the medium term, I guess that Greg's work to make custom environment hosting a first-class part of the Hubs UI is going to obsolete any decisions we make here.

mquander:
Obsoleted in favor of #529.

PR #501, opened 2018-08-07
Add a nag button for returning to VR on Oculus browser

netpro2k:
Stupid simple workaround for the user accidentally getting out of VR mode.


src/react-components/ui-root.js
```diff
@@ -54,6 +54,8 @@ async function grantedMicLabels() {
 
 const AUTO_EXIT_TIMER_SECONDS = 10;
 
+const isOculusBrowser = navigator.userAgent.match(/Oculus/);
```
gfodor: this should probably use the `isInHMD` check from vr device caps, which will be true for oculus browser and fxr

src/assets/translations.data.json
```diff
@@ -22,6 +22,7 @@
     "entry.invite-others": "invite others",
     "entry.invite-others-nag": "invite others to join",
     "entry.enable-screen-sharing": "Share my desktop",
+    "entry.return-to-vr": "Return to VR",
```
gfodor: I'd probably change this to "Enter in VR" to match the copy on the usual landing page

src/react-components/ui-root.js
netpro2k: Hmm, will `isInHMD` not also be true for Vive/Rift, and if not, should we rename it to something less confusing?

PR #505, opened 2018-08-07
Feature/file uploads gfodor fixes

gfodor:
- Fixes homepage by pulling in A-Frame
- Properly uses auth header to get HEAD so content type no longer needs to be networked


gfodor:
- Switches SCSS to use snake case for class names

gfodor:
- Add period to copy and add spacing between lines between tip and instructions.

PR #507, opened 2018-08-08
Feature/object spawn telemetry

gfodor:
This PR threads through two new A-Frame events `media_resolved` (on a `media-loader` bound entity) and `object_spawned` (on the scene) in order to allow us to send telemetry events for object spawning. Note that `object_spawned` is only spawned on the creator's instance. With this, we can capture the kinds of objects that were spawned in a room. A new phoenix channel message is PR'ed for reticulum to enable this telemetry:

https://github.com/mozilla/reticulum/pull/55


src/components/media-loader.js
```diff
@@ -67,6 +67,8 @@ AFRAME.registerComponent("media-loader", {
 
       const { raw, origin, contentType } = await resolveMedia(url, token);
 
+      this.el.emit("media_resolved", { src: this.data.src, raw, origin, contentType });
```
netpro2k: Replace `this.data.src` with `url`

src/hub.js
```diff
@@ -297,16 +299,25 @@ const onReady = async () => {
...
+
       entity.setAttribute("offset-relative-to", {
         target: "#player-camera",
         offset
       });
     };
 
     scene.addEventListener("add_media", e => {
-      spawnMediaInfrontOfPlayer(e.detail);
+      const origin = e.detail instanceof File ? ObjectContentOrigins.FILE : ObjectContentOrigins.URL;
```
netpro2k: Would call this `contentOrigin` to avoid confusion with `origin` from the media response.

src/object-types.js
```diff
@@ -0,0 +1,86 @@
+// Enumeration of spawned object content origins. URL means the content is
+// fetched from a URL, FILE means it was an uploaded/dropped file, and
+// CLIPBOARD means it was raw data pasted in from the clipboard, SPAWNER means
```
netpro2k: Hmm so ther eis no way to tell clipboard urls vs add media dialog url usage?

src/object-types.js
```diff
@@ -0,0 +1,86 @@
...
+  "application/pdf": [ObjectTypes.URL_PDF, ObjectTypes.FILE_PDF, ObjectTypes.CLIPBOARD_PDF, ObjectTypes.SPAWNER_PDF]
+};
+
+// Given an content origin and the resolved mime type of a piece of content, return
+// the ObjectType, if any, for that content.
+export function objectTypeForOriginAndContentType(origin, contentType) {
+  for (const prefix in objectTypeMimePrefixLookupMap) {
+    if (contentType.startsWith(prefix)) {
+      const types = objectTypeMimePrefixLookupMap[prefix];
+      return objectTypeForOrigin(origin, ...types);
```
netpro2k: why not jsut `return types[origin-1]`, or better yet make ObjectContentOrigins zero indexed.

src/react-components/create-object-dialog.js
```diff
@@ -50,7 +50,8 @@ export default class CreateObjectDialog extends Component {
   onUrlChange = e => {
     let attributionImage = this.state.attributionImage;
     if (e.target && e.target.value && e.target.validity.valid) {
-      attributionImage = attributionHostnames[new URL(e.target.value).hostname];
+      const url = new URL(e.target.value);
```
netpro2k: Not an issue, but was this change intended to be kept?

src/components/media-loader.js
```diff
@@ -67,6 +67,8 @@ AFRAME.registerComponent("media-loader", {
...
+
```
netpro2k: With PDFs this block of code is going to be hit for every page. Are we going to want to filter that?

src/utils/media-utils.js
```diff
@@ -71,5 +73,11 @@ export const addMedia = (src, resize = false) => {
         entity.setAttribute("media-loader", { src: "error" });
       });
   }
+
+  entity.addEventListener("media_resolved", ({ detail }) => {
```
johnshaughnessy: The entity's `media-loader`'s `src` can change, which may mean that this `contentOrigin` is no longer correct. So far it will be, but we should be aware of this. PDF's are the only thing that makes use of the changing `src` right now AFAIK

src/object-types.js
brianpeiris: No, looks like they are conflated in the telemetry. I assume we care more about measuring activity with this telemetry, vs UX anyway. Also, pasting a URL directly is kind of a shortcut to using the dialog. I'm going to leave it as it is.

src/object-types.js
brianpeiris: Since we're treating ObjectContentOrigins like an enum, probably best not to assume it is contiguous and sequential.

src/react-components/create-object-dialog.js
brianpeiris: 🤷‍♂️ Maybe a readability thing. I'll keep it.

src/utils/media-utils.js
brianpeiris: Hmm, I think contentOrigin is independent of src. In this context, contentOrigin is a value from the ObjectContentOrigins enum. i.e. one of URL, FILE, CLIPBOARD or SPAWNER.

PR #510, opened 2018-08-10
Add Spoke Light Components

robertlong:
Adds light components for Spoke environments.


src/components/ambient-light.js
```diff
@@ -0,0 +1,27 @@
...
+      this.light.color.set(this.data.color);
+    }
+
+    if (this.data.intensity !== prevData.intensity) {
+      this.light.intensity = this.data.intensity;
+    }
+  },
+
+  remove: function() {
+    this.el.removeObject3D("ambient-light");
```
netpro2k: Pretty sure aframe will handle doing this for you.

src/components/ambient-light.js
```diff
@@ -0,0 +1,27 @@
...
+  schema: {
+    color: { type: "color" },
+    intensity: { default: 1.0 }
+  },
+
+  init() {
+    const el = this.el;
+    this.light = new THREE.AmbientLight();
+    this.el.setObject3D("ambient-light", this.light);
+    this.el.sceneEl.systems.light.registerLight(el);
```
netpro2k: Since we dont own this system it feels a bit weird to be registering with it. From what I understand you are doing this to remove the default lights so maybe we should just set `defaultLightsEnabled` to `false` on the light system.  Not a huge deal either way.

src/components/spot-light.js
```diff
@@ -0,0 +1,47 @@
...
+
+    if (this.data.intensity !== prevData.intensity) {
+      this.light.intensity = this.data.intensity;
+    }
+
+    if (this.data.range !== prevData.range) {
+      this.light.distance = this.data.range;
+    }
+
+    if (this.data.innerConeAngle !== prevData.innerConeAngle || this.data.outerConeAngle !== prevData.outerConeAngle) {
```
netpro2k: Is there a reason we prefer `innerConeAngle` and `outerConeAngle` vs `angle` and `penumbra`? (No prefference from me, jsut wondering.

src/components/ambient-light.js
robertlong: Not on a per-component level. It does remove this object when removing an entity though.

src/components/spot-light.js
robertlong: I'm just mapping to the glTF spec for future implementation.

PR #512, opened 2018-08-13
Basic support for PDF files in media tools

netpro2k:
This PR adds basic support for PDF files to media tools. The UX leaves a lot to be desired but the basic functionality should be there.

This also contains a fix for the rotation netwokring issues we have been seeing with some media objects and cleans up the hanlding of upload tokens.


src/components/media-loader.js
```diff
@@ -80,35 +84,59 @@ AFRAME.registerComponent("media-loader", {
...
       } else if (
         contentType.includes("application/octet-stream") ||
         contentType.includes("x-zip-compressed") ||
         contentType.startsWith("model/gltf") ||
-        url.endsWith(".gltf") ||
-        url.endsWith(".glb")
+        src.endsWith(".gltf") ||
+        src.endsWith(".glb")
       ) {
+        this.el.removeAttribute("image-plus");
```
netpro2k: Really not loving this sort of "make sure to do all the inverse opperations when media type switches" things. This is indicating to me that we still need to split this up more, but not 100% clear on the best way to do that yet.

src/utils/media-utils.js
```diff
@@ -65,7 +65,7 @@ export const addMedia = (src, resize = false) => {
       .then(response => {
         const src = response.raw;
         const token = response.meta.access_token;
-        entity.setAttribute("media-loader", { src, token });
+        entity.setAttribute("media-loader", { src: `${src}?token=${token}` });
```
johnshaughnessy: If `src` has query params, this will break. Check for query params.

src/components/media-loader.js
```diff
@@ -53,62 +55,72 @@ AFRAME.registerComponent("media-loader", {
...
+      ) {
+        this.el.removeAttribute("gltf-model-plus");
         this.el.addEventListener(
           "image-loaded",
-          () => {
+          async () => {
             clearTimeout(this.showLoaderTimeout);
+            delete this.showLoaderTimeout;
+            if (isPDF) {
+              const maxIndex = await fetchMaxContentIndex(src, images.png);
```
johnshaughnessy: Is there a reason to prefer either `Promise.prototype.then` or (`async` and `await`) when either is available?

src/components/media-loader.js
```diff
@@ -53,62 +55,72 @@ AFRAME.registerComponent("media-loader", {
...
+        src.endsWith(".glb")
       ) {
+        this.el.removeAttribute("image-plus");
+        this.el.removeAttribute("media-pager");
         this.el.addEventListener(
           "model-loaded",
           () => {
             clearTimeout(this.showLoaderTimeout);
             this.setShapeAndScale(this.data.resize);
+            delete this.showLoaderTimeout;
```
johnshaughnessy: swap this line and the one above it so that `delete`ing the timeout immediately follows `clear`ing the timeout.

src/components/media-loader.js
```diff
@@ -121,3 +125,54 @@ AFRAME.registerComponent("media-loader", {
     }
   }
 });
+
+AFRAME.registerComponent("media-pager", {
```
johnshaughnessy: Are you purposefully putting the `media-loader` and the `media-pager` in the same file? That is fine if you are, just wondering if this is an accident.

src/components/media-loader.js
```diff
@@ -121,3 +125,54 @@ AFRAME.registerComponent("media-loader", {
...
+  schema: {
+    index: { type: "string" },
+    maxIndex: { type: "string" }
+  },
+
+  init() {
+    this.onNext = this.onNext.bind(this);
+    this.onPrev = this.onPrev.bind(this);
+
+    const tempalte = document.getElementById("paging-toolbar");
```
johnshaughnessy: typo: `tempalte` -> `template`

src/components/media-loader.js
```diff
@@ -121,3 +125,54 @@ AFRAME.registerComponent("media-loader", {
...
+    this.el.appendChild(document.importNode(tempalte.content, true));
+    this.toolbar = this.el.querySelector(".paging-toolbar");
+    // we have to wait a tick for the attach callbacks to get fired for the elements in a tempalte
```
johnshaughnessy: comment typo: `tempalte` -> `template`

src/hub.html
```diff
@@ -155,14 +155,30 @@
...
+                    <!-- HACK: rotation component above is required for its side effect of setting YXZ order -->
                     <a-entity class="delete-button" visible-while-frozen>
                         <a-entity mixin="rounded-text-button" remove-networked-object-button position="0 0 0"> </a-entity>
                         <a-entity text=" value:Delete; width:2.5; align:center;" text-raycast-hack position="0 0 0.01"></a-entity>
                     </a-entity>
                 </a-entity>
             </template>
 
+            <template id="paging-toolbar">
+                <a-entity class="paging-toolbar" visible-to-owner>
```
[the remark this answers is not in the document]
netpro2k: As it is currently written yes, since only the owner would network the page state. The alternative would be to making `media-pager` aware of the concept of ownership, but since design wise it seems kind of deisrable to not show the paging buttons for non-owners anyway, it seems better this way.

src/components/media-loader.js
netpro2k: Not really no. I usually use then if I can achieve the result in a point free way, or with a simple funciton with implicit return like `fetch().then(r => r.json())` but use async for anything else unless its particularly awkward for some reason like we saw with the constructor example Marshal shoed us.

src/components/media-loader.js
netpro2k: I did yeah, though I know I tend to be the only one that likes to do this.

johnshaughnessy: The end of an AFRAME component feels like an arbitrary boundary point to split up files so I don't mind it.

PR #514, opened 2018-08-17
Support for basic video synchronization (WIP)

netpro2k:


src/components/image-plus.js
```diff
@@ -135,41 +215,155 @@ errorImage.onload = () => {
...
 
-AFRAME.registerComponent("image-plus", {
+AFRAME.registerComponent("media-video", {
   schema: {
     src: { type: "string" },
-    contentType: { type: "string" },
+    time: { type: "number" },
+    videoPaused: { type: "boolean" },
+    tickRate: { default: 1000 }, // ms interval to send time interval updates
+    syncTolerance: { default: 2 }
```
johnshaughnessy: We may find we need to loosen the `syncTolerance`, especially given a `tickRate` of 1000 ms. If we run into trouble finding a tolerance that works well for variable network conditions, we can try increasing the `syncTolerance` anytime we are outside the range and require a seek. Then we can gradually shrink the tolerance as time goes on without being outside the bounds.

src/components/media-loader.js
```diff
@@ -65,18 +72,19 @@ AFRAME.registerComponent("media-loader", {
...
 
       const isPDF = contentType.startsWith("application/pdf");
-      if (
-        contentType.startsWith("image/") ||
-        contentType.startsWith("video/") ||
-        contentType.startsWith("audio/") ||
-        isPDF
-      ) {
+      if (contentType.startsWith("video/") || contentType.startsWith("audio/")) {
+        this.el.removeAttribute("gltf-model-plus");
```
johnshaughnessy: Do we want to write `if (anythingChanged) {removeAllTheThings();}`? Perhaps we don't, because if the `src` of a `media-image` or `gltf-model-plus` changes, we don't _need_ to `remove` those components and create new ones.

src/network-schemas.js
```diff
@@ -93,7 +93,15 @@ function registerNetworkSchemas() {
         requiresNetworkUpdate: vectorRequiresUpdate(0.5)
       },
       "scale",
-      "media-loader"
+      "media-loader",
+      {
+        component: "media-video",
+        property: "time"
```
johnshaughnessy: Nice

src/components/image-plus.js
```diff
@@ -135,41 +215,155 @@ errorImage.onload = () => {
...
+    this.updateTexture(src);
+  },
+
+  tick() {
+    if (this.data.videoPaused || !this.video || !this.networkedEl || !NAF.utils.isMine(this.networkedEl)) return;
+
+    const now = performance.now();
+    if (now - this.lastUpdate > this.data.tickRate) {
+      // console.log("sending time", this.video.currentTime);
+      this.el.setAttribute("media-video", "time", this.video.currentTime);
```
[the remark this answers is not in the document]
netpro2k: Yeah. Ideally I would just always set this and then handle network throttling in the schema, but it seems like a bad idea to thrash setAttribute.

src/components/image-plus.js
netpro2k: Yeah I think upping both the tickrate and syncTolerance would be fine.

src/components/media-loader.js
netpro2k: Yeah exactly. could maybe DRY it up into a function taht you give 1 component name and it remvoes all the others, but that doesnt really seem better.

src/components/media-views.js
```diff
@@ -307,19 +306,12 @@ AFRAME.registerComponent("media-video", {
 
     this.updatePlaybackState(true);
 
-    // TODO: verify if we actually need to do this
```
johnshaughnessy: What a relief

src/components/media-views.js
```diff
@@ -418,12 +409,6 @@ AFRAME.registerComponent("media-image", {
 
     fitToTexture(this.el, texture);
 
-    // TODO: verify if we actually need to do this
```
johnshaughnessy: A+

PR #516, opened 2018-08-20
When trying to make the mobile client fullscreen, keep it fullscreen.

johnshaughnessy:


netpro2k:
Not sure what the intent of this PR is, but I notice we are now doing this in two spots (the one introduced here, and in ui-root)

PR #517, opened 2018-08-20
remove velocity based haptics on interactables

InfiniteLee:
instead, pulse when grabbing and releasing. Also enable haptics for the cursor.

### DIFF
--- a/src/hub.html
+++ b/src/hub.html
@@ -329,6 +329,7 @@
                   missOpacity: 0.1;
                   curveShootingSpeed: 12;"
               haptic-feedback
+              event-repeater="events: haptic_pulse; eventSource: #cursor"
               body="type: static; shape: none;"
               mixin="controller-super-hands"
               controls-shape-offset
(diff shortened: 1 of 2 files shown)


src/hub.html
```diff
@@ -329,6 +329,7 @@
                   missOpacity: 0.1;
                   curveShootingSpeed: 12;"
               haptic-feedback
+              event-repeater="events: haptic_pulse; eventSource: #cursor"
```
brianpeiris: Add this to the left controller as well?

InfiniteLee: Actually that brings up a good point. On 6-dof, having this on the left largely doesn't matter as your cursor is only ever on the right hand. However, that is not the case for 3dof devices that may be left-handed only. That said, haptics only exist on 6-dof right now, and so this should be sufficient. (If we did add it to the left anyways, both controllers would get a haptic pulse when the cursor triggered one).

netpro2k: We currently only support right hand on 3dof anyway, so this sint an issue, though we obviously need to address the larger issue of handedness throughout.

InfiniteLee: >We currently only support right hand on 3dof anyway
Is not true, we respect if you have your controller set to left-handed on go, gearvr and daydream

PR #520, opened 2018-08-21
(WIP) Rotate images that have been uploaded.

johnshaughnessy:
Trying to fix issue in https://github.com/mozilla/hubs/issues/451:
>  Portrait photo uploads from phone are rotated

I _only_ need to rotate images that are uploaded from the camera; The images that come from the hard drive have the correct orientation. I'm not sure yet how to detect the difference.

### DIFF
--- a/src/utils/media-utils.js
+++ b/src/utils/media-utils.js
@@ -66,6 +66,12 @@ export const addMedia = (src, resize = false) => {
         const srcUrl = new URL(response.raw);
         srcUrl.searchParams.set("token", response.meta.access_token);
         entity.setAttribute("media-loader", { src: srcUrl.href });
+        // Need to only do this media uploaded from the CAMERA, but how can we tell?
+        entity.object3D.rotation.set(
+          entity.object3D.rotation.x,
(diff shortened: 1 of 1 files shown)


johnshaughnessy:
@netpro2k suggested the orientation data for the image may be embedded in the images themselves, e.g. in the EXIF data https://en.wikipedia.org/wiki/Exif.  I haven't read up on this yet. In order to get this data in the client, we may need to PARSE the images, which I think requires a bit of code for each image format we want to support. @netpro2k suggests we may do this in Farspark, since it is an image proxy and may have this kind of code built in, but I'm not sure it does:

> It would be great to be able to rotate, flip and apply masks to images, but in most of the cases, it is possible — and is much easier — to do that using CSS3.

Paging @gfodor to ask whether you know if we should try to get the image orientation in farspark and either rotate the actual image data (pixels), or just pass that info to the client and then set the initial orientation of the 3D object similar to the way I'm doing it in this code.

johnshaughnessy:
Closing this PR because this code is wrong, but I wanted a place to organize the discussion about this issue.

PR #523, opened 2018-08-22
Add a super-cool loading cube.

johnshaughnessy:


src/components/media-loader.js
```diff
@@ -48,14 +62,26 @@ AFRAME.registerComponent("media-loader", {
...
     this.el.setObject3D("mesh", loadingObj);
     this.setShapeAndScale(true);
     delete this.showLoaderTimeout;
   },
 
   clearLoadingTimeout() {
     clearTimeout(this.showLoaderTimeout);
+    if (this.loaderMixer) {
+      this.loadingClip.stop();
+      this.loaderMixer = null;
```
netpro2k: I would use `delete this.loadingMixer` instead. Doesn't actually make a difference in your code but I think the intend reads more cleanly.

src/components/media-loader.js
```diff
@@ -48,14 +62,26 @@ AFRAME.registerComponent("media-loader", {
   },
 
   showLoader() {
-    const loadingObj = new THREE.Mesh(new THREE.BoxGeometry(), new THREE.MeshBasicMaterial());
+    const useFancyLoader = !!loadingObject;
```
netpro2k: Was gpong to suggest just making a promise to loadingObject instead of having these 2 code paths but can't decide if that actually ends up being cleaner.

src/components/media-loader.js
johnshaughnessy: 👍

PR #524, opened 2018-08-23
Feature/drawing

InfiniteLee:
Adds a pen interactable that allows for 3d spatial drawing.

Generates procedural geometry using (modified) @fernandojsg's `sharedbuffergeometry.js` and `sharedbuffergeometrymanager.js` from A-Painter.

Adds a concept of "toggleable" for interactables. Objects that have a "toggle" class defined will not be dropped from your hand/cursor after initial grab. Only subsequent releases using the "primary" button will cause the object to be dropped. E.g. if grip is considered primary and trigger is secondary, grip or trigger will pick up the object, but after initial release, only the grip can cause the object to be dropped. (`grabbable-toggle` was added which is 90% the same as `super-hands` `grabbable` but has some additional logic to handle this concept.)

Relies on a new component `activatable` in `super-hands`: https://github.com/wmurphyrd/aframe-super-hands-component/pull/148 which allows secondary activations on grabbed/hovered objects. 

TODO:
- [x] remove pen interactable from being hard-coded into hub.html and instead add them to the environments
- [x] fix regression with how "toggle" objects are transitioned from one hand to another
- [x] update duck in mr-social-assets to not require `super-spawner` to have hard-coded duck src (maybe part of a different PR)


src/input-mappings.js
```diff
@@ -54,41 +54,38 @@ const config = {
...
         menuup: "thumb_up",
-        gripdown: ["action_grab", "middle_ring_pinky_down"],
-        gripup: ["action_release", "middle_ring_pinky_up"],
+        gripdown: ["primary_action_grab", "middle_ring_pinky_down"],
+        gripup: ["primary_action_release", "middle_ring_pinky_up"],
         trackpadtouchstart: "thumb_down",
         trackpadtouchend: "thumb_up",
-        triggerdown: ["action_grab", "index_down"],
-        triggerup: ["action_release", "index_up"],
-        scroll: { right: "move_duck" }
```
johnshaughnessy: RIP `"move_duck"`

src/network-schemas.js
```diff
@@ -72,15 +72,33 @@ function registerNetworkSchemas() {
...
-        component: "rotation"
+        component: "rotation",
+        requiresNetworkUpdate: vectorRequiresUpdate(0.5)
       },
       "visible"
     ]
   });
 
+  NAF.schemas.add({
+    template: "#interactable-template",
```
johnshaughnessy: I think this was replaced in master -- do we need this?

src/utils/sharedbuffergeometry.js
```diff
@@ -0,0 +1,135 @@
...
+    const colors = new Float32Array(this.maxBufferSize * 3);
+
+    const mesh = new THREE.Mesh(geometry, this.material);
+
+    mesh.drawMode = this.primitiveMode;
+
+    mesh.frustumCulled = false;
+    mesh.vertices = vertices;
+
+    this.object3D = new THREE.Object3D();
```
johnshaughnessy: I think this can be `const object3D = ...` instead of `this.object3D`. I don't see this being referred to anywhere else.

src/utils/mouse-events-handler.js
```diff
@@ -1,10 +1,14 @@
...
 const HORIZONTAL_LOOK_SPEED = 0.1;
 const VERTICAL_LOOK_SPEED = 0.06;
+const VERTICAL_SCROLL_TIMEOUT = 250;
+const HORIZONTAL_SCROLL_TIMEOUT = 250;
 
 export default class MouseEventsHandler {
   constructor(cursor, cameraController) {
     this.cursor = cursor;
+    const cursorController = this.cursor.el.getAttribute("cursor-controller");
+    this.superHand = cursorController.cursor.components["super-hands"];
```
johnshaughnessy: Ohh this line confused me until I realized you were referring to the `cursor`'s `cursor-controller`'s `cursor`'s `super-hands` component, where the `cursor` started as the `input-configurator`'s `cursorController` selector-ed element's `cursor-controller` component -- silly me.

package-lock.json
```diff
@@ -47,7 +47,8 @@
     },
     "@babel/helper-get-function-arity": {
       "version": "7.0.0-beta.44",
-      "resolved": "https://registry.yarnpkg.com/@babel/helper-get-function-arity/-/helper-get-function-arity-7.0.0-beta.44.tgz",
+      "resolved":
```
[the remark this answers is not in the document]
InfiniteLee: no, its just from running npm ci. I asked @mquander and he said to just commit whatever npm ci says to do. Pretty much all it did was add a "^" to each version number.

src/network-schemas.js
InfiniteLee: This got added back in intentionally, but will get removed again once the environments get updated (doing that now)

src/utils/sharedbuffergeometry.js
InfiniteLee: Ah yea, there's probably a few instances of stuff like this in this file since it came over from A-painter. I caught a few of them but thanks for finding this one.

src/components/tools/pen.js
```diff
@@ -0,0 +1,182 @@
...
+      default: [
+        "#FF0033",
+        "#FFFF00",
+        "#0099FF",
+        "#00FF33",
+        "#9900FF",
+        "#FF6600",
+        "#FFFFFF",
+        "#AAAAAA",
+        "#777777",
```
gfodor: probably due to the unlit shader but to actually get a gradient of greys I bet it has to be something like `#000000`, `#111111`, `#222222` or something

also we need brown so i can draw a proper goomba

https://www.mariowiki.com/images/thumb/9/9c/Goomba_-_MarioPartyStarRush.png/200px-Goomba_-_MarioPartyStarRush.png

gfodor: oh, we might want to include one or more flesh tones too

gfodor: so you can draw faces

gfodor:
truly amazing work! having it be a workable sculpting tool was a great surprise too.

media tools + pen tool = something magical is happening

PR #527, opened 2018-08-25
Remove Farspark resolution logic from gltf-model-plus

mquander:
**We can't merge this until the recent Farspark changes are configured in prod to provide Farspark URLs for subresources inside proxied GLTF files.**

Now this logic is applied in media-loader, which means we don't need to do it for GLTF components that don't need to be proxied (e.g. all of the normal Hubs assets) and we can throw out a bunch of complexity.

### DIFF
--- a/src/components/media-loader.js
+++ b/src/components/media-loader.js
@@ -152,8 +152,8 @@ AFRAME.registerComponent("media-loader", {
         );
         this.el.addEventListener("model-error", this.onError, { once: true });
         this.el.setAttribute("gltf-model-plus", {
-          src,
-          contentType,
+          src: raw,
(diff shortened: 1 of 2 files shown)


src/components/media-loader.js
```diff
@@ -151,9 +151,11 @@ AFRAME.registerComponent("media-loader", {
           { once: true }
         );
         this.el.addEventListener("model-error", this.onError, { once: true });
+
+        const resolved = await resolveMedia(src);
```
netpro2k: This url has already been resolved above on line 98, you should just be able to use `raw` and `contentType` already in this scope.

PR #535, opened 2018-08-30
Proxy raw GLTFs and GLBs used as environment

mquander:
This is useful to CORS/CSP proxy GLTFs and GLBs people paste into the scene creation environment box.


gfodor:
LGTM

PR #538, opened 2018-08-31
Dump aframe-extras, write nav system ourselves

mquander:
To date, we have represented navigation meshes in GLTFs like this:

```JSON
{
  ... some other nodes
  {
    "name": "NavMesh",
    "mesh": 1,
    "extras": { "components": { "visible": false, "nav-mesh": {} } }
  }
}
```

That is, the `nav-mesh` component is directly on a mesh node that is meant to serve as the navigation mesh. This worked with the `nav-mesh` A-Frame component packaged in aframe-extras.

In Spoke, we don't attach components to the inside of GLTFs right now, only their root. So we can't attach the `nav-mesh` component to the mesh. Instead, the mesh is somewhere underneath the node with the `nav-mesh` component. Thus, the `nav-mesh` A-Frame component needed to change, or else we needed to change how we were using it. The new rule needs to be to use the first mesh which is found in the scene subtree of the `nav-mesh` component.

It turns out that the `nav-mesh` A-Frame component and associated `nav` system had very little functionality, being thin wrappers for the underlying three-pathfinding library. So it's easiest to just throw them out entirely since they are not quite what we want anyway. That's the only thing we were using aframe-extras for, so now that is thrown out too.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -548,13 +548,6 @@
       "resolved": "https://registry.yarnpkg.com/aframe-billboard-component/-/aframe-billboard-component-1.0.0.tgz",
       "integrity": "sha1-EM4kgnKe73OGxYRNZZF1gaYtOtw="
     },
-    "aframe-extras": {
-      "version": "github:MozillaReality/aframe-extras#f545017da417468ffa3b41f57dd32643a32f65b1",
-      "from": "github:MozillaReality/aframe-extras#three-pathfinding-0.7",
-      "requires": {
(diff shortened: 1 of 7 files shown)


robertlong:
Is there any sort of cleanup we need to do with zone data? If so I'd maybe use a component and do cleanup in the `remove()` method. I guess we aren't swapping environments at the moment so it should be fine.

We should also add that zone property to Spoke, but we probably should come up with a better UI element so that you don't accidentally set a zone name with zero functionality.

Looks good to me!

PR #542, opened 2018-08-31
Fixed skybox, visible, and interactable components for spoke

robertlong:
- Set the skybox scale to the distance property (Note this is a breaking change. Environments will need the skybox node's scale set to 1, 1, 1)
- Adds the interactable component which just adds the interactable css-class
- All spoke components must be objects. The visible component inflator now also accepts an object


src/gltf-component-mappings.js
```diff
@@ -4,6 +4,9 @@ AFRAME.GLTFModelPlus.registerComponent("quack", "quack");
 AFRAME.GLTFModelPlus.registerComponent("sound", "sound");
 AFRAME.GLTFModelPlus.registerComponent("collision-filter", "collision-filter");
 AFRAME.GLTFModelPlus.registerComponent("css-class", "css-class");
+AFRAME.GLTFModelPlus.registerComponent("interactable", "css-class", (el, componentName) => {
```
netpro2k: Why not actually just use the css class component?

src/gltf-component-mappings.js
```diff
@@ -43,7 +46,13 @@ AFRAME.GLTFModelPlus.registerComponent(
     };
   })()
 );
-AFRAME.GLTFModelPlus.registerComponent("visible", "visible");
+AFRAME.GLTFModelPlus.registerComponent("visible", "visible", (el, componentName, componentData) => {
```
johnshaughnessy: How come the `visible` component needs this custom `inflator`?

src/gltf-component-mappings.js
brianpeiris: Spoke cannot export components with literal prop values. Prop values are always objects.

src/gltf-component-mappings.js
brianpeiris: Same as above, Spoke cannot export components with literal prop values. Prop values are always objects so we need to use a custom inflator to translate the value to a-frame but also maintain backwards compat for our existing environments (which we will eventually convert to the new format).

PR #543, opened 2018-08-31
Make web workers go into engine bundle

mquander:
The Sketchfab worker pulls in jszip, which is very large, so it's important that we don't have to download it separately in the avatar selector and hub pages.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -161,7 +161,7 @@ module.exports = (env, argv) => ({
     splitChunks: {
       cacheGroups: {
         engine: {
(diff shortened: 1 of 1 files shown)


netpro2k:
Does this actuall prevent the script from getting inlined? I would assume you would have to also modify the settings for worker-loader.

But in any case, how would this actually work without changing the CSP? The reason we ended up inlining the workers was because we were getting a CSP error when loading the workers in produciton.

mquander:
Either way, the script is inlined, but this change makes it get inlined into the engine bundle instead of the individual page bundles.

PR #548, opened 2018-09-04
Add hosted aframe-inspector to fix csp issues.

robertlong:


mquander:
Why do this instead of permitting the existing hosted inspector script on aframe.io?

PR #550, opened 2018-09-05
Don't import all of fontawesome

mquander:
This was doubling the vendor bundle size (about 500KB) to no end.


johnshaughnessy:
TIL. Thank you

Issue #503, opened 2018-08-07
Source Maps are no longer being built


robertlong:
When running `npm build` we are no longer generating source maps.

PR #552, opened 2018-09-05
Make source maps pop out in production again

mquander:
I think this regressed in https://github.com/mozilla/hubs/pull/461. When you specify your own `minimizer`, Webpack doesn't respect your `devtool` setting anymore.

Resolves https://github.com/mozilla/hubs/issues/503.


johnshaughnessy:
Aha, good to have source maps in prod. This default behavior confused some other people as well. Link for posterity: https://github.com/webpack/webpack/issues/2704

PR #554, opened 2018-09-05
Store artifacts from build in Circle CI

mquander:
This is me trying to get Circle CI to pop out a build size notification measured using these stored artifacts.


mquander:
Seems to work!

PR #556, opened 2018-09-06
Refactor dialog code to not all be one component

mquander:
Previously, we had the `InfoDialog` component, which was a sort of generic dialog whose `render` method would switch on a string type and produce different dialog contents depending on the type. Now we have a `DialogContainer` component with common dialog code and separate `SlackDialog`, `UpdatesDialog`, etc. components with a `DialogContainer` wrapping dialog contents for each type. Callers (who invariably know which dialog they want) can instantiate the appropriate dialog component explicitly.

This seems better just because it is common sense, but it also removes the weird fact that to use any of the dialogs you have to depend on the code for all of the dialogs, which we decided was a good reason to download a megabyte of Javascript on our landing page.


mquander:
I'm not 100% happy with this implementation because it annoys me that the pages with modal dialogs (UIRoot, HomeRoot) sort of have to do some ceremony wiring up `onClose` to set their dialog state. This behavior seems common to any modal dialog, so I feel like there's a way to rearrange things into another component and make those pages not have to do it.

src/index.js
```diff
@@ -3,17 +3,17 @@ import React from "react";
...
-    initialEnvironment={qs.get("initial_environment")}
-    dialogType={
-      qs.has("list_signup") ? InfoDialog.dialogTypes.updates : qs.has("report") ? InfoDialog.dialogTypes.report : null
-    }
-  />,
-  document.getElementById("home-root")
-);
+let component;
+const container = document.getElementById("home-root");
+const root = <HomeRoot ref={c => (component = c)} initialEnvironment={qs.get("initial_environment")} />;
```
netpro2k: This feels a bit odd, why not just do it on `HomeRoot`'s `componentDidMount`?

src/react-components/create-object-dialog.js
```diff
@@ -84,6 +78,14 @@ export default class CreateObjectDialog extends Component {
   };
 
   render() {
+    const { onCreate, onClose, ...other } = this.props;
+
+    const onCreateClicked = e => {
```
netpro2k: Any reason you moved this away from being a method? In generail my rule is to only define functions in render if they need to capture some things in scope that would be ugly pass around (like itterators). Beyond that, if it uses `this` it's a method (which I bind n the contructior) otherwiset its a function I define outside the component.

src/react-components/home-root.js
```diff
@@ -12,27 +12,31 @@ import { ENVIRONMENT_URLS } from "../assets/environments/environments";
...
+import UpdatesDialog from "./updates-dialog.js";
+import DialogContainer from "./dialog-container.js";
 
 addLocaleData([...en]);
 
 class HomeRoot extends Component {
   static propTypes = {
     intl: PropTypes.object,
-    dialogType: PropTypes.symbol,
+    dialog: PropTypes.string,
```
netpro2k: Why is this `string` and not `element` or `node`?

src/react-components/home-root.js
```diff
@@ -54,13 +58,33 @@ class HomeRoot extends Component {
...
-      e.stopPropagation();
-      this.setState({ dialogType });
-    };
-  };
+  closeDialog() {
+    this.setState({ dialog: null });
+  }
+
+  showSlackDialog() {
+    this.setState({ dialog: <SlackDialog onClose={this.closeDialog} /> });
```
netpro2k: Ah interesting, I was expecting you migh have ended up doing something like:
```
const Dialog = this.props.dialog
<Dialog onClose={this.closeDialog| />
```
In render. A bit less ceremony, but the way you have done it lets you vary the arguments which seems to be required.

src/react-components/home-root.js
```diff
@@ -80,9 +104,16 @@ class HomeRoot extends Component {
   render() {
     const mainContentClassNames = classNames({
       [styles.mainContent]: true,
-      [styles.noninteractive]: !!this.state.dialogType
+      [styles.noninteractive]: !!this.state.dialog
     });
-    const dialogTypes = InfoDialog.dialogTypes;
+
+    const onLinkClicked = trigger => {
```
netpro2k: Not a huge deal, but I would move this out of render. I think I only notice these becasue in some previous projects it was a linting rule to not define funcitons in render. Not sure how much it matters in practice.

gfodor:
thanks for this!

src/index.js
mquander: Good point, I may refactor this to pass in the parameters as props and do the update-dialog-making in there.

src/react-components/create-object-dialog.js
mquander: Honestly I think it was because otherwise lint complains that I didn't use `onCreate`. But I want to destructure `onCreate` so that `other` has the right props to forward onto `DialogContainer`.

How would you arrange this? Just make a lint ignore?

src/react-components/home-root.js
mquander: Good catch, in the present incarnation it should clearly be `node`. I don't think we even need this prop anymore though.

src/react-components/home-root.js
mquander: OK, if you like that convention I will follow it.

PR #557, opened 2018-09-06
MobileStandardMaterial

robertlong:
Replaces any StandardMaterials with the MobileStandardMaterial when on low quality. This improves performance of glTF models loaded via the media tools that do not use `MOZ_alt_materials` and `KHR_materials_unlit`.

The MobileStandardMaterial currently uses an unlit shading model. It uses the `color`, `map`, `aoMap`, `aoMapIntensity`, `emissive`, `emissiveMap`, and `opacity` fields from the StandardMaterial. The result is visually similar to the output of `gltf-unlit-generator`.

We could also use this material instead of using `gltf-unlit-generator` and `MOZ_alt_materials` in our art pipeline. It would result in downloading more textures and a slightly more expensive shader, but can be used in all situations without export-time generation of an unlit map.

Future versions of MobileStandardMaterial could also use the metallic, roughness, and normal maps in a more efficient manner than the StandardMaterial. Disabling dynamic lighting and using a cheaper approximate BRDF will improve the performance while still looking physically correct.


src/components/gltf-model-plus.js
```diff
@@ -256,8 +257,12 @@ async function loadGLTF(src, contentType, preferredTechnique, onProgress) {
 
   gltf.scene.traverse(object => {
     if (object.material && object.material.type === "MeshStandardMaterial") {
-      object.material.envMap = envMap;
-      object.material.needsUpdate = true;
+      if (preferredTechnique === "KHR_materials_unlit") {
```
gfodor: see this wasn't so hacky! :)

(thanks for the PR when you are technically on vacation!)

PR #560, opened 2018-09-08
Turn off collisions when grabbing objects.

netpro2k:
This helps when scaling objects or trying to stick them into other objects. It does make it so you can't hit objects with an object you are holding, but that was quite buggy anyway so it doesnt seem like a big loss. This might cause some issues with sticky-zones at some point but we will revisit that once they are actually being used sicne its likely they have regressed in other ways as well.

### DIFF
--- a/src/components/sticky-object.js
+++ b/src/components/sticky-object.js
@@ -47,7 +47,7 @@ AFRAME.registerComponent("sticky-object", {
   },
 
   _onRelease() {
-    // Happens if the object is still being heald by another hand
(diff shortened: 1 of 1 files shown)


src/components/sticky-object.js
```diff
@@ -47,17 +47,21 @@ AFRAME.registerComponent("sticky-object", {
   },
 
   _onRelease() {
+    // Happens if the object is still being heald by another hand
```
johnshaughnessy: held

Issue #421, opened 2018-06-05
Perf: Don't use aframe Line component in cursor-controller


johnshaughnessy:
We call `setAttribute` every frame in the `cursor-controller` in order to update the start and end of the aframe `line` component. We will see a performance benefit by not using the aframe `line` component and manipulating the underlying threejs line directly.

gfodor:
https://github.com/mozilla/hubs/pull/757

myeshiamorgan:
u all will get in trouble for identification theft an fraud also privacy
violations theives..

M.Morgan

PR #562, opened 2018-09-10
Stop using A-Frame raycaster component

mquander:
Roughly speaking, the A-Frame raycaster component was keeping track of the target entities, keeping track of the set of currently intersected elements, doing the raycast, and firing events when the set of intersected elements changed. Now we do that stuff in our cursor controller and we don't use an A-Frame raycaster component at all.

The result is much better, for a real big variety of reasons:

- The A-Frame raycaster was opinionated about how it sent the events. It sent events that we didn't care about, and it emitted them from the raycaster entity, which forced us to use an `event-repeater` component to get them onto the cursor. Now we can send them the way we want.

- The A-Frame raycaster had a lot of complexity associated with the fact that it was managing a whole set of intersections per raycast. In Hubs, we're only ever interested in the first intersected target.

- We were doing per-frame A-Frame property updates to communicate the target ray into the raycaster component, which are no longer necessary. As usual with A-Frame property updates, this was costing order of tenths of a millisecond per frame.

- The A-Frame raycaster had all of its own line drawing functionality that we weren't using, preferring to draw the line ourselves. This included calling `setTimeout` per frame, which is expensive (even though we didn't use it to draw lines!)

- The A-Frame raycaster was doing extra work gathering up the raycast targets from the entity hierarchy because it wanted to support non-recursive raycasting (i.e. on a fixed list of targets and not their children) which I think is not what we want.

- A lot of extra vector math was being done due to confusions between the A-Frame raycaster's expectations and our reality.

- The A-Frame raycaster was wholly responsible for interacting with the underlying `THREE.Raycaster`. We're about to change our raycasting mechanism to be something compatible with a spatially indexed data structure like an octree or BVH, so having that be controlled by A-Frame was not what we wanted.


src/components/cursor-controller.js
```diff
@@ -3,39 +3,113 @@ const TARGET_TYPE_INTERACTABLE = 2;
...
+  },
+
+  update: function() {
+    this.raycaster.far = this.data.far;
+    this.raycaster.near = this.data.near;
+    this.setDirty();
   },
 
+  play: function() {
+    this.observer = new MutationObserver(this.setDirty);
```
johnshaughnessy: TIL `MutationObserver` exists.

mquander: Me too. By the way, I don't know whether it's actually important for us to observe attribute changes or not -- that behavior is just inherited from the A-Frame raycaster. Do you know? That is, do we anticipate cases where an entity changes from raycast-targetable to not-raycast-targetable or vice versa?

InfiniteLee:
There appears to be bug: When I have my vive plugged in but I enter in 2D mode, the cursor is stuck to gaze mode when it should be in normal 2D cursor mode.

mquander:
Interesting. I didn't test with a desktop HMD yet so no surprise. I'll make
sure that case starts working.

src/components/cursor-controller.js
johnshaughnessy: Right now we do. For example when you interact with the spawner, it creates a new duck and hides the spawner and prevents you from interacting with the spawner a second time on accident by changing its css attribute. See https://github.com/mozilla/hubs/blob/master/src/components/super-spawner.js#L188

mquander: Cute! I didn't know exactly how spawners worked.

src/components/cursor-controller.js
```diff
@@ -45,42 +119,28 @@ AFRAME.registerComponent("cursor-controller", {
...
-        this.origin.setFromMatrixPosition(rayObject.matrixWorld);
-        this.updateRay();
+      if (this.dirty) {
+        this.populateEntities(this.data.objects, this.targets);
+        this.dirty = false;
       }
 
-      const isGrabbing = this.isInteracting();
-      if (isGrabbing) {
+      this.performRaycast(this.targets);
```
netpro2k: Yay, this actually happening in tick here makes everything so much claener!

src/components/cursor-controller.js
```diff
@@ -45,42 +119,28 @@ AFRAME.registerComponent("cursor-controller", {
...
         const distance = Math.min(
-          this.data.maxDistance,
-          Math.max(this.data.minDistance, this.currentDistance - this.currentDistanceMod)
+          this.data.far,
+          Math.max(this.data.near, this.currentDistance - this.currentDistanceMod)
         );
-        this.direction.multiplyScalar(distance);
-        this.data.cursor.object3D.position.addVectors(this.origin, this.direction);
+        this.data.cursor.object3D.position.copy(this.raycaster.ray.origin);
+        this.data.cursor.object3D.position.addScaledVector(this.raycaster.ray.direction, distance);
```
netpro2k: TIL `addScaledVector`

src/components/cursor-controller.js
```diff
@@ -96,7 +156,10 @@ AFRAME.registerComponent("cursor-controller", {
       }
 
       if (this.data.drawLine) {
-        this.el.setAttribute("line", { start: this.origin.clone(), end: this.data.cursor.object3D.position.clone() });
+        this.el.setAttribute("line", {
```
netpro2k: For all the reasons you stated re updated the raycaster aframe properties per frame we should just be directly using a `THREE.Line` here as well, but would be fine as its own PR.

johnshaughnessy:
What Dom said is true:
> Getting the raycasting happening in tick is going to clean up some of the input stuff John has been experemingting with as well.

I wrote some more thoughts trying to think through how to simplify the raycasting code further, although I'm not sure if it actually simplifies it or just moves the complexity somewhere else. If you're curious, I wrote [a (pretty raw) train of thought about it](https://github.com/mozilla/hubs/blob/actions/doc/touch-actions.org#input-based-branching-in-the-cursors-raycasting), which ends with me trying to think through how to implement a "virtual device" (which I also called `appdevice`) whose state depended on some app state (the camera position or whether a raycast sent at the time of a touchdown/mousedown event intersects a duck or pen or nothing).
Not sure if it makes sense (and sorry to crowd this PR with this info), but it would change the proposed `tick` to write the state of the app actions for [these "paths"](https://github.com/mozilla/hubs/blob/actions/src/systems/actions/paths.js#L77) in [three conceptual steps](https://github.com/mozilla/hubs/blob/actions/src/systems/actions.js#L65)

src/components/cursor-controller.js
johnshaughnessy: ditto https://github.com/mozilla/hubs/issues/421

mquander: Yes, that will improve performance in the case where the line is active. I will do it in a future PR.

src/components/cursor-controller.js
```diff
@@ -3,39 +3,113 @@ const TARGET_TYPE_INTERACTABLE = 2;
...
+        rayObjectRotation.setFromRotationMatrix(rayObject.matrixWorld);
+        this.raycaster.ray.origin.setFromMatrixPosition(rayObject.matrixWorld);
+        this.raycaster.ray.direction.set(0, 0, -1).applyQuaternion(rayObjectRotation);
+      } else {
+        this.raycaster.setFromCamera(this.mousePos, this.data.camera.components.camera.camera); // camera
+      }
+      const prevIntersection = this.intersection;
+      rawIntersections.length = 0;
+      this.raycaster.intersectObjects(targets, true, rawIntersections);
+      this.intersection = rawIntersections.find(x => x.object.el);
```
johnshaughnessy: This is surprising to me. If the targets are `object3D`, then should this be `x.el`?

PR #566, opened 2018-09-11
Bug/2d ui

johnshaughnessy:
- Fix user-select bug where you accidentally highlight the icons all the time.
- Fix bug where you can't touch spawn-media or choose-image buttons until you touch a joystick first.

### DIFF
--- a/src/assets/stylesheets/ui-root.scss
+++ b/src/assets/stylesheets/ui-root.scss
@@ -70,6 +70,10 @@
 
 .ui-interactive {
   pointer-events: auto;
+  -moz-user-select: none;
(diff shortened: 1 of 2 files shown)


src/assets/stylesheets/ui-root.scss
```diff
@@ -70,6 +70,10 @@
 
 .ui-interactive {
   pointer-events: auto;
+  -moz-user-select: none;
```
netpro2k: I was gunna suggest that we dont need prefixes because this is older tahn sin, but apparently we still do. Oh CSS.

PR #571, opened 2018-09-15
Use new comparator contract

gfodor:
Goes with: https://github.com/networked-aframe/networked-aframe/pull/149


gfodor:
Updated for https://github.com/networked-aframe/networked-aframe/pull/150

PR #572, opened 2018-09-15
Feature/phoenix channels

gfodor:
WIP, pending naf-janus-adapter update(s) for custom transports

### DIFF
--- a/src/hub.js
+++ b/src/hub.js
@@ -414,27 +414,20 @@ const onReady = async () => {
         return;
       });
 
-      const sendHubDataMessage = function(clientId, dataType, data, reliable) {
+      const sendHubDataMessage = function(clientId, dataType, data) {
         const payload = { dataType, data };
 
         if (clientId != null) {
           payload.clientId = clientId;
         }
 
-        if (reliable) {
-          hubChannel.channel.push("naf", { payload });
-        } else {
-          hubChannel.channel.pushUnreliable("naf", { payload });
-        }
+        hubChannel.channel.pushUnreliable("naf", { payload });
       };
 
       window.sendHubDataMessage = sendHubDataMessage;
 
-      NAF.connection.adapter.reliableTransport = (clientId, dataType, data) =>
-        sendHubDataMessage(clientId, dataType, data, true);
-
-      NAF.connection.adapter.unreliableTransport = (clientId, dataType, data) =>
-        sendHubDataMessage(clientId, dataType, data, false);
+      NAF.connection.adapter.reliableTransport = sendHubDataMessage;
+      NAF.connection.adapter.unreliableTransport = sendHubDataMessage;
 
       if (isDebug) {
         NAF.connection.adapter.session.options.verbose = true;
--- a/src/hub.js
+++ b/src/hub.js
@@ -414,16 +414,27 @@ const onReady = async () => {
         return;
       });
 
-      const sendHubDataMessage = function(clientId, dataType, data) {
+      const sendHubDataMessage = function(clientId, dataType, data, reliable) {
         const payload = { dataType, data };
+
         if (clientId != null) {
           payload.clientId = clientId;
         }
-        hubChannel.channel.push("naf", { payload });
+
+        if (reliable) {
+          hubChannel.channel.push("naf", { payload });
+        } else {
+          hubChannel.channel.pushUnreliable("naf", { payload });
+        }
       };
 
-      NAF.connection.adapter.reliableTransport = sendHubDataMessage;
-      NAF.connection.adapter.unreliableTransport = sendHubDataMessage;
+      window.sendHubDataMessage = sendHubDataMessage;
+
+      NAF.connection.adapter.reliableTransport = (clientId, dataType, data) =>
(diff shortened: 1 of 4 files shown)


src/hub.js
```diff
@@ -248,7 +248,12 @@ const onReady = async () => {
     document.body.removeEventListener("touchend", requestFullscreen);
   };
 
+  let hasEnteredScene = false;
+
   const enterScene = async (mediaStream, enterInVR, hubId) => {
+    if (hasEnteredScene) return;
+    hasEnteredScene = true;
```
mquander: What case is this handling? It's not obvious to me how it would happen that `enterScene` is called twice.

src/hub.js
```diff
@@ -521,22 +539,24 @@ const onReady = async () => {
...
+  const setRoom = (hubId, hubName) => {
+    remountUI({ hubId, hubName });
+  };
+
   if (qs.has("room")) {
     // If ?room is set, this is `yarn start`, so just use a default environment and query string room.
-    setRoom(qs.get("room") || "default");
+    const hubId = qs.get("room") || "default";
+    setRoom(hubId, hubId);
+    if (isBotMode) enterSceneWhenReady();
```
mquander: So this isn't going to work anymore if data is going over Phoenix channels, right? We should throw out this feature or fix it.

src/utils/phoenix-utils.js
```diff
@@ -20,7 +20,17 @@ export function connectToReticulum() {
...
-  const socket = new Socket(socketUrl, { params: { session_id: uuid() } });
+  const socketSettings = {
+    params: { session_id: uuid() }
+  };
+
+  if (debug) {
+    socketSettings.logger = (kind, msg, data) => {
+      console.log(`${kind}: ${msg}`, data);
+    };
+  }
```
mquander: IMO we should use our logging framework and its log levels, although we have historically been half-hearted about it.

src/hub.js
```diff
@@ -581,6 +614,12 @@ const onReady = async () => {
       console.error(res);
     });
 
+  channel.on("naf", data => {
+    if (NAF.connection.adapter) {
+      NAF.connection.adapter.onData(data.payload);
+    }
+  });
```
mquander: It seems like we are still subscribing to this channel when not yet in the scene. Why not wait to subscribe? Our loading process is already extremely slow and janky and the last thing we need is a hundred websocket frames per second flying in while we do it.

mquander:
Honestly, this code is kind of too much for me to figure out if it's really correct, and I haven't even understood rejoins yet. Since we are so unclear about which parts of this we endorse as a long-term solution, I'm OK with whatever seems to work for now and expecting that we will have to change it all later somehow anyway.

src/hub.js
mquander: (If the answer is "we need to be subscribed for the channel for those other things", I would suggest splitting them into two channels.)

gfodor: we could potentially break things up into multiple phoenix channels, but the current hub channel is used for a variety of things before this change added naf packets to its responsibilities:

- it manages the session tracking + telemetry (a session is considered 'active' while the channel is subscribed to)
- it sends over the initial hub room information in the join reply

soon this will make a lot more sense, as the new landing page design is going to incorporate:

- presence join/leave events to the channel as users land on or leave the page
- ephemeral chat messages pre-entry
- and, likely, a full view into the game state/voices on load

gfodor: i'm of the mindset it would probably be a reasonable idea to break the naf data out into its own channel, but i don't feel like there's a forcing function yet -- this will add a sizable amount of complexity if we ever want to do anything interesting with the data flowing over that channel (like persistence)

src/hub.js
gfodor: on the fence:

- removing it seems sane to me if its any effort to keep
- it seems easy to keep: we should just have a rule to use data channels for the relevant pipes iff there is no `RETICULUM_SERVER`

gfodor: actually nevermind that is wrong. if there is no `RETICULUM_SERVER` we assume those URLs are just going to be relative :/ i'll kill this feature for now until we feel some pain. thankfully we have dev set up properly now!

src/hub.js
gfodor: you're right I think this is leftover from earlier changes

src/utils/phoenix-utils.js
gfodor: transitioning to it seems like a dedicated PR -- we have these console logs all over

gfodor:
I updated it to no longer use `pushUnreliable`, so this should be compatible again with Phoenix 1.3 -- we can leave the binding to my gfodor branch until upstream has the other optimizations (hopefully)

src/hub.js
gfodor: oh, do you mean do not subscribe to the `naf` message type? that doesn't save much other than this function dispatch I think (we will still receive the messages, they just won't be routed because there's no match on the message type.) i think i'll leave it because I think this will all be *much* more sane soon given that you will be joining the room in full on load (except will be in spectator mode.)

PR #581, opened 2018-09-21
Camera Tool

netpro2k:
An MVP for a simple in world camera tool. Controls are like the pen. Using a placeholder model from sketchfab for now.

image

Todo:
- [ ] Add spawn camera button to the VR HUD


src/components/camera-tool.js
```diff
@@ -0,0 +1,141 @@
...
+  },
+
+  stateAdded(evt) {
+    if (evt.detail === "activated") {
+      this.takeSnapshotNextTick = true;
+    }
+  },
+
+  tock(_, __, tockCamera) {
+    // tock gets called after any render. We trigger a render by rendering the camera preview, so ignore those
```
brianpeiris: This comment is a bit confusing.

src/components/camera-tool.js
```diff
@@ -0,0 +1,141 @@
...
+    const renderer = this.renderer || sceneEl.renderer;
+    const now = performance.now();
+
+    if (!this.playerHead) {
+      const headEl = document.getElementById("player-head");
+      this.playerHead = headEl && headEl.object3D;
+    }
+
+    if (this.takeSnapshotNextTick || this.updateRenderTargetNextTick) {
+      const tempScale = new THREE.Vector3();
```
brianpeiris: Avoid allocating new vectors in tock

brianpeiris:
Oh, forgot to mention that the VR hud still appears in the snapshots.

johnshaughnessy:
+10 points to Dombledoor for putting a screenshot in the PR. I like this.

johnshaughnessy:
Brian said
> Oh, forgot to mention that the VR hud still appears in the snapshots.
and it reminded me that Dom and I talked about whether the floating DELETE and BLOCK buttons should appear in the snap if you try to take a photo while in pause mode. I think the answer isn't obvious one way or another, since you might pause in order to get a photo of stuff before anyone moves it (or their avatars) around too much, or you might pause it in order to show someone what something looks like while you're in pause mode, so that they know what you're talking about. This matters more on devices that don't have the PRINT SCREEN button on the keyboard. Relatedly: if we apply a filter to the player camera during pause mode (e.g. grayscale), should the camera also render the scene with that filter applied?

gfodor:
I feel like the "i want to freeze the scene for a photo" use case seems much more interesting to me than the other one -- if i had to pick i'd not render the buttons

gfodor:
pausing the app and then using "image generation" tools like the camera is kind of a fascinating idea. for example, you could pause the app and then use the camera to take a bunch of action shots of the same moment in time. or maybe you can snap a panorama or stereo shot. etc.

gfodor:
one q: if everyone spawns a camera in the room, are we all rendering the scene N times if those cameras are in view? if so that seems bad.

gfodor:
lots of ways that could be addressed: only render the nearest camera's viewports, pick a random camera to render each frame, etc

netpro2k:
> one q: if everyone spawns a camera in the room, are we all rendering the scene N times if those cameras are in view (and their frame step happen to line up)? if so that seems bad.

Yep, which is obviously not ideal. Each person is soft limited to 1 camera and the cameras only render at a few FPS, so we will have to see how it ends up feeling.

PR #589, opened 2018-09-26
Fix up media loading a lot

mquander:
Now that Farspark accepts requests from the client, we can do many nice things:

- You can now tell `media-loader` whether it needs to resolve the URL you give it into a canonical media URL using the Reticulum resolution API before loading. Otherwise, it doesn't have to do that. Basically the only time we do it is when someone pastes a mystery file into the create object box.
- As a result of ambient simplification, `media-pager` can now have all PDF-specific paging logic, drastically simplifying `media-loader`.
- We now try to infer the content type from the extension of the provided URL and don't bother making a HEAD request to test the waters if we think we know what it is. This is a nice optimization since it's handling common cases like "the ducks in our environment" and "the pens you spawn."
- The combination of these things makes PDF paging responsive and nice instead of terrible.

Next easy steps in this process:

- Clean up dead code in Reticulum, since it doesn't need to assemble Farspark URLs anymore.
- Consider making media URL resolution in Reticulum a GET request, which should eliminate the necessity of manual caching on the client.
- Clean up the Farspark API URL space to make it less annoying for the client to assemble Farspark URLs for subresources.

After that we should start thinking about how we want to unify the media resolution and proxying into one service, to further improve efficiency.

### DIFF
--- a/.env.defaults
+++ b/.env.defaults
@@ -8,6 +8,10 @@ JANUS_SERVER="wss://dev-janus.reticulum.io"
 # See here for the server code: https://github.com/mozilla/reticulum
 RETICULUM_SERVER="dev.reticulum.io"
 
+# The Farspark backend to connect to. Used as a CORS proxy and transformer for in-world media objects.
+# See here for the server code: https://github.com/MozillaReality/farspark
+FARSPARK_SERVER="farspark-dev.reticulum.io"
+
 # The root URL under which Hubs expects environment GLTF bundles to be served.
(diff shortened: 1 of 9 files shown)


mquander:
I also removed the client-side caching for max index and content-type -- I think these were mostly in there because our implementation was trying to do way too much work for PDFs and GLTFs and it would request them over and over. Now it doesn't, anyway, so they would only come into play if people paste a lot of the same thing (and in the case of content type, we can't infer it.)

PR #591, opened 2018-09-26
Run pngcrush

mquander:
Why not? The pngs got a lot smaller.


robertlong:
Can we do image minification in the production webpack build? That way people don't accidentally forget to minify their images. Or maybe that would add too much time to the build process?

This project looks interesting: https://github.com/tcoopman/image-webpack-loader

mquander:
Some arguments against putting it in the build:

- Image authors really should be thinking about this when they make the image, rather than as an afterthought (especially if they are making JPEGs, but even if they are making PNGs.)
- It seems to me that the checked-in version should be the real one we use. I don't see any benefit to checking in a less-optimized version and building a more-optimized version out of it.
- It shouldn't matter very much so I'm not inclined to add any build complexity. A marginal 50 KB of images shouldn't be a very big deal (compared to 50 KB of JS, which I actually care about.)

If you want, feel free to make something to do it, but I wouldn't bother.

PR #595, opened 2018-10-01
Feature/lobby design

gfodor:
This PR incorporates most of the design elements from Rian's redesign into the hub landing page, sans the presence/chat updates and some aspects to the visual styling (like color and font updates, which will be in another PR.) It also updates things to use the new lobby functionality and the new entry codes.


brianpeiris:
I think the avatar picker design feels broken and incomplete now. This is the first thing you see if you enter a room and you don't have a saved avatar name.

The background text and UI bleeding through the transparency makes the UI very confusing and hard to parse and the prompt is too terse for it to be the first thing you're greeted with.

I like that the avatar picker has more screen real estate though. 

image

brianpeiris:
Also on mobile, you can actually pinch to move still

brianpeiris:
Other changes look to me otherwise. Thanks for doing all the refactoring. Has been a long time coming.

brianpeiris:
Maybe something like this?

image

gfodor:
https://github.com/mozilla/hubs/pull/599

PR #597, opened 2018-10-02
Feature/alpha link codes

gfodor:
This PR reintroduces headset link codes -- the trick is that entry codes are numeric and link codes use the letters A-I. The keypad at hub.link lets the user toggle to letter mode, with a secondary CTA to flip to letter mode if they are linking a headset. The flow for getting a link code also looks very different then entry codes, to ensure users are not confused between the two.


gfodor:
https://github.com/mozilla/hubs/pull/599

PR #603, opened 2018-10-03
glTF Animation Support (WIP)

robertlong:
This PR adds support for animated glTF models.

- `gltf-model-plus` adds a `animation-mixer` component to models with 1 or more animation before the `model-loaded` event.
- If the model contains one or more animation and there are no `animation-loop` components defined in the model. An `animation-loop` component with default props will be added to the `gltf-model-plus` entity.
- By default the `animation-loop` component will infinitely loop the first animation clip.

I've found one issue with this implementation. The `animation-loop` component is added to any glTF models without an `animation-loop` component  specified. This seems like a good behavior for spawned objects, but maybe not for anything else. I noticed this caused the `idle-eyes` animation to play on bots in the avatar picker. That seems desirable, but I can imagine there may be other models that should not loop their first animation. Also, we'll want to be able to specify different types of animation than just infinite looping. I think we want to move adding the `animation-loop` component to some place where we handle spawning media. I'm not sure where best to put that.


gfodor:
nice! i think this concern could belong on GLTF model plus or elsewhere, its hard to say at this stage. my gut says that this should be an option on the GLTF component, because one could imagine (if we don't have it already) a variety of rules you can turn on/off during inflation to "augment" vanilla GLTF with smart behaviors like this.

gfodor:
(for example, theoretically "generate a nav mesh" would be in this set of configs)

gfodor:
so they would all be off by default (so no unexpected surprises) but then you could flip them on when you add the component. so the robot eyes would not have the unintended side effect of moving but we would start using the "generate animation component" bits in our media tools calls

gfodor:
similarly, we would use the "create nav mesh" config when loading the scene GLB, but not the media tools

robertlong:
I feel like we should be configuring it via composition of components rather than props which add those components. It seems like a more scalable pattern. The problem is I don't know if we have a centralized place where we define the entity that gets spawned for an interactable.

robertlong:
`nav-mesh-generator` could then be a component that could be added to any entity with a mesh in it's `object3DMap`. We could also define additional animation components.

gfodor:
potentially -- the thing that makes this case weird is you're not really adding behavior to the object, your adding "meta-behavior" about what behaviors or other scene nodes to create specifically at inflation time. once that event is over, the model of having these as components doesn't make much sense. it seems coming up with a scalable design pattern here is probably more about making the bits reusable and generic outside of the GLTF inflator where you might want to generate these kinds of secondary things

src/components/gltf-model-plus.js
```diff
@@ -31,57 +31,55 @@ AFRAME.GLTFModelPlus = {
...
+    clonedMesh.bind(clonedMesh.skeleton, sourceMesh.bindMatrix);
+  });
 
   return clone;
 }
 
+// From https://gist.github.com/cdata/f2d7a6ccdec071839bc1954c32595e87
+// Tracking glTF cloning here: https://github.com/mrdoob/three.js/issues/11573
+function cloneGltf(gltf) {
+  console.log("cloneGLTF");
```
johnshaughnessy: marking for removal

src/components/gltf-model-plus.js
```diff
@@ -312,8 +312,21 @@ AFRAME.registerComponent("gltf-model-plus", {
     if (useCache) {
       if (!GLTFCache[src]) {
         GLTFCache[src] = await loadGLTF(src, contentType, technique);
+
+        if (GLTFCache[src].canClone) {
+          return cloneGltf(GLTFCache[src]);
+        } else {
+          console.log("a");
```
johnshaughnessy: marking for removal

src/components/gltf-model-plus.js
```diff
@@ -312,8 +312,21 @@ AFRAME.registerComponent("gltf-model-plus", {
...
+          return GLTFCache[src];
+        }
+      } else {
+        if (GLTFCache[src].canClone) {
+          return cloneGltf(GLTFCache[src]);
+        } else {
+          console.log("b");
```
johnshaughnessy: marking for removal

src/components/gltf-model-plus.js
```diff
@@ -362,7 +375,17 @@ AFRAME.registerComponent("gltf-model-plus", {
...
         }
       }
+
       this.el.setObject3D("mesh", object3DToSet);
+
+      if (model.animations.length > 0) {
+        this.el.setAttribute("animation-mixer", "");
+
+        if (!this.el.querySelector("[loop-animation]")) {
+          this.el.setAttribute("loop-animation", "");
```
johnshaughnessy: Does this allow us to remove the animation mixer in `media-loader` that plays the loading cube's animation? 
https://github.com/mozilla/hubs/blob/master/src/components/media-loader.js#L74

robertlong: It would, yeah. Whether that should be the responsibility of `gltf-model-plus` or if we should be adding those components in `media-loader` or somewhere else, I don't know.

robertlong:
I'm not certain I understand what you're suggesting as an alternative, but I took another stab at it.

I moved the logic to add the `animation-mixer` and `loop-animation` components out of `gltf-model-plus` and into `hub.js` and `media-loader`. This way `gltf-model-plus` does not have to be knowledgeable of the animation components.

johnshaughnessy:
image
conflict in hub.js?

src/components/gltf-model-plus.js
```diff
@@ -31,57 +31,54 @@ AFRAME.GLTFModelPlus = {
...
-  const skinnedMeshes = {};
-  gltf.scene.traverse(node => {
-    if (!node.name) {
-      node.name = node.uuid;
-    }
-    if (node.isSkinnedMesh) {
-      skinnedMeshes[node.name] = node;
-    }
-  });
+function parallelTraverse(a, b, callback) {
```
gfodor: it took me a bit to figure out what was going on here. i'd probably push this into a domain-specific function that fills the map recursively unless this routine is going to be used elsewhere. alternatively, if you want to keep it generic, i'd write some comments.

is there a reason you couldn't just look this up by id in the cloned graph? just perf?

src/components/gltf-model-plus.js
```diff
@@ -31,57 +31,54 @@ AFRAME.GLTFModelPlus = {
...
-  };
+  for (let i = 0; i < a.children.length; i++) {
+    parallelTraverse(a.children[i], b.children[i], callback);
+  }
+}
 
-  const cloneBones = {};
-  const cloneSkinnedMeshes = {};
+function cloneSkinnedMesh(source) {
+  const cloneLookup = new Map();
```
gfodor: just a style thing, but maps like this I lke to name FooToBar so you know what the key/value space is. so sourceToCloneNode etc

src/components/gltf-model-plus.js
```diff
@@ -31,57 +31,54 @@ AFRAME.GLTFModelPlus = {
...
+  parallelTraverse(source, clone, function(sourceNode, clonedNode) {
+    cloneLookup.set(sourceNode, clonedNode);
   });
 
-  for (const name in skinnedMeshes) {
-    const skinnedMesh = skinnedMeshes[name];
-    const skeleton = skinnedMesh.skeleton;
-    const cloneSkinnedMesh = cloneSkinnedMeshes[name];
+  source.traverse(function(sourceMesh) {
+    if (!sourceMesh.isSkinnedMesh) return;
```
gfodor: I'm confused, so this assumes the root node is a skinned mesh?

robertlong:
Oops @gfodor in those comments above I forgot to document where I got the code from: https://github.com/mrdoob/three.js/pull/14494

src/components/gltf-model-plus.js
gfodor: oh nevermind, i forgot traverse doesn't terminate early when returning falsy

src/components/loop-animation.js
```diff
@@ -1,62 +1,78 @@
...
-      this.mixer = this.el.components["animation-mixer"].mixer;
-    } else {
-      this.onModelLoaded = this.onModelLoaded.bind(this);
-      this.el.addEventListener("model-loaded", this.onModelLoaded);
+    this.onMixerReady = this.onMixerReady.bind(this);
+
+    if (this.mixerEl) {
+      if (this.mixerEl.components["animation-mixer"]) {
+        this.mixer = this.mixerEl.components["animation-mixer"].mixer;
+        this.updateClip();
```
gfodor: probably can DRY this against `onMixerReady`

src/components/loop-animation.js
```diff
@@ -1,62 +1,78 @@
...
 
-        break;
-      }
+    if (!clipName) {
+      clip = animations[0];
+    } else {
+      clip = animations.find(({ name }) => name === clipName);
+    }
+
+    if (clip) {
```
gfodor: consider `if (!clip) return`

src/hub.js
```diff
@@ -541,6 +541,15 @@ const onReady = async () => {
   const initialEnvironmentEl = document.createElement("a-entity");
   initialEnvironmentEl.addEventListener("bundleloaded", () => {
     remountUI({ initialEnvironmentLoaded: true });
+
+    for (const modelEl of initialEnvironmentEl.querySelectorAll(":scope > [gltf-model-plus]")) {
```
gfodor: note this style selector wasn't working in edge in the one other place we do it (of course edge support is not going to happen anytime soon :))

src/hub.js
```diff
@@ -541,6 +541,15 @@ const onReady = async () => {
...
+      const model = modelEl.object3DMap.mesh;
+
+      if (model && model.animations.length > 0) {
+        modelEl.setAttribute("animation-mixer", "");
```
gfodor: I think this routine should be extracted into a plain utility function for now -- eg is there a reason we check for loop animation above but not here?

gfodor:
LGTM left some comments!

gfodor:
there is some non-trivial changes to `hub.js` you'll need to integrate. we no longer create the `environmentEl` node programatically, but grab one out of the dom called `#scene-environment`. you'll need to take your code that adds the animation mixer and manually insert it into the similar region in master's `hub.js` where we load the root scene.

src/hub.js
robertlong: I was thinking we probably don't want to automatically add the `loop-animation` component to the environment. They should probably be specified in the glTF model you are using as your environment. Idk, maybe we should use the same behavior? Would that mess any environments up?

gfodor: it seems like a good idea to make these consistent so we will naturally parameterize things properly as we hit edge cases -- I don't think any of our existing environments have one-shot animations so probably ok on that front -- any other concerns with adding the loop?

robertlong: Idk, let's turn it on and get feedback.

PR #609, opened 2018-10-05
Feature/spoke landing

gfodor:
v1 cut of the spoke landing page. still will need to be adjusted once we get the initial videos.


robertlong:
Could we organize the hero section vertically rather than horizontally?

screen shot 2018-10-08 at 4 26 38 pm

robertlong:
Here are my suggested changes: https://github.com/mozilla/hubs/tree/feature/spoke-landing-robert

robertlong:
I'd put the play video button directly on top of the video if you want to break out into a player or prevent loading it by default on mobile.

robertlong:
I feel like I need to see the video on the right hand side to really get a feel for it.

The left hand side is not vertically aligned on my monitor. This gives it too much vertical space above the logo. Something is forcing the page to overflow.
screen shot 2018-10-08 at 5 52 21 pm

screen shot 2018-10-08 at 5 57 30 pm

screen shot 2018-10-08 at 5 57 22 pm

Also the buttons on the iPhoneX and the S9 are using padding and probably should be using margins instead so they do not get clipped on mobile. The border radius on the background video can also be disabled when on the mobile breakpoint.

PR #615, opened 2018-10-08
Feature/new design ux tweaks

gfodor:
This PR improves the UX around invites based on testing:

- Remove the placeholder number of digits in the hub.link dialog which can be confusing if you have an alpha code
- Add both a Copy and Share button if you have the share API
- Always show an invite mechanism even if there are people in the room. On smaller screens, show a one-tap-to-copy button that is smaller than the usual invite button.
- In VR, show the invite link above the HUD
- Stop spillover in HUD tooltips
- Removes links on the environment titles since they are a confusing CTA on the home page


brianpeiris:
I had some concerns earlier about the front page so I'll bring them up now that we're doing this:

- The "Join Room" button is the first CTA you see since the "Create Room" button is only visible after all the carousel content has loaded. This takes about 360ms on production.
- The "Join the Conversation" link is quite prominent and can be positioned directly below the "Join Room" button, depending on your screen.

I think we either need to fix the carousel delay, or only show the "Join Room" button after the carousel is loaded. I would also reduce the prominence of the bottom links, either making them grey, reducing their font size and/or removing their outline.

image

brianpeiris:
Not for this PR but it seems we might need to auto hide the UI for landscape mode on phones. Feels pretty crowded now.

image

brianpeiris:
Also probably not for this PR, but I suspect adding a simple hyphen between the six digits would significantly help with cognitive load and recall:

image

gfodor:
i agree with all your feedback, i'll do another pass on this. worth noting the current carousel will likely be ditched for Rian's final design once we do a pass over our scene collection

gfodor:
instead of codifying the dash explicitly, I'll at least space the digits apart into two 3-digit segments which should get 80% of the value. adding the dash may cause a chain of unintended consequences :)

gfodor:
ok updated @brianpeiris

PR #620, opened 2018-10-11
Feature/tweet buttons

gfodor:
This PR adds two tweet buttons:

- The invite dialog will now offer a "tweet" button if there is no `navigator.share` capability
- The scene landing page now has a secondary CTA to tweet the scene. Focusing on Twitter sharing of scenes seems right vs having a generic `navigator.share` facility like the invite dialog given the desire for scenes to go wide vs rooms which are more for private shares.


src/react-components/invite-dialog.js
```diff
@@ -44,6 +44,11 @@ export default class InviteDialog extends Component {
     const entryCodeString = pad(entryCode, 6);
     const shareLink = `hub.link/${entryCodeString}`;
 
+    const tweetText = `Join me now in #hubs!`;
+    const tweetLink = `https://twitter.com/share?url=${encodeURIComponent(
+      "https://" + shareLink
```
brianpeiris: Shouldn't we use the full room URL for tweets? If a link expires, we send you to a completely blank 404 page. With a full URL, even if the room is now empty, at least you get to look around and (later) follow a link back to the original scene.

gfodor: I considered this but am not sure -- I think for the "copy" action it clearly makes sense to copy the link as it appears on screen. It may be confusing to show the short link and then pass through the longer link to twitter (and, presumably, this would be the same on the navigator.share flow as well.)

I'm leaning towards using the full link now that you've brought it up though, do you think it makes sense to change it here and also for the navigator.share flow?

brianpeiris: Yeah, might as well use the full link for navigator.share as well then.

PR #625, opened 2018-10-13
Presence + Chat

gfodor:
Adds user presence information behind presence info button.


src/react-components/presence-list.js
```diff
@@ -0,0 +1,61 @@
+import React, { Component } from "react";
+import PropTypes from "prop-types";
+import styles from "../assets/stylesheets/presence-list.scss";
+import classNames from "classnames";
+import PhoneImage from "../assets/images/mobile_screen_entry.svg";
+import DesktopImage from "../assets/images/desktop_screen_entry.svg";
+import HMDImage from "../assets/images/generic_vr_entry.svg";
```
brianpeiris: These icons do not look so good on a normal DPI desktop for some reason:

image

gfodor: ah bc these are svgs. i will bite the bullet and export better PNGs.

PR #626, opened 2018-10-14
Feature/permalink and mobile entry ux

gfodor:
This PR:

- Updates the invite dialog to use the new short permalinks from https://github.com/mozilla/reticulum/pull/75

- Updates some copy and flow for the mobile headset linking. Updates the copy on the device chooser. Changes the hub.link landing page to not show headset linking options unless you are in a headset, which should help reduce confusion for the mainline case of entry codes. (Note: this means that if we don't have proper detection for isInHMD for some browsers, like FxR, they will no longer be able to do headset linking. we need to fix this for FxR.)


src/react-components/link-root.js
```diff
@@ -208,15 +211,19 @@ class LinkRoot extends Component {
...
+                  onTouchStart={() => this.toggleMode()}
+                  onClick={() => {
+                    if (!hasTouchEvents) this.toggleMode();
+                  }}
+                >
+                  {this.state.isAlphaMode ? "123" : "ABC"}
+                </button>
+              ) : (
+                <div />
+              )}
```
brianpeiris: Not sure if you noticed, but the "Linking a Headset?" option in the footer is currently styled to disappear on short mobile screens. Even if it didn't just disappear, you wouldn't be able to scroll to it since we disable scrolling on this page.

brianpeiris:
This doesn't change the in-VR entry code url. That needs to change, right?

brianpeiris:
The `navigator.share` code also needs to change

src/react-components/link-root.js
gfodor: yeah that option should only be visible inside of an HMD

PR #627, opened 2018-10-17
Ignore pastes on active input elements

gfodor:
This PR fixes an issue where Ctrl-V stopped working with the introduction of a text field into the DOM (the chat box entry.) The reason is that the `paste` event seems to set `e.target` to the topmost target in the DOM that could be pasted into, regardless of if it is the focused element or not.  This additional check added will ensure that if that element is not the focused element (ie we're pasting into the canvas) then it will fall through.

### DIFF
--- a/src/scene-entry-manager.js
+++ b/src/scene-entry-manager.js
@@ -221,7 +221,7 @@ export default class SceneEntryManager {
     });
 
     document.addEventListener("paste", e => {
-      if (e.target.nodeName === "INPUT") return;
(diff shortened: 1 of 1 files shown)


mquander:
Looks OK, I guess this code might be pretty brittle (e.g. if we ever have a textarea or something) but your change seems right.

I wonder if we should be calling `preventDefault` in the case where we're pasting media?

gfodor:
Not sure about that one :/

PR #638, opened 2018-10-21
Remove keyboard shortcuts, fix up immediate entry mode

gfodor:
Removes the incredibly annoying F and ESC keybindings a-frame injects by default, and also fixes up the _now mode to enter in one click (ideally it would just enter on load, but this functionality regressed during the UX redesign so this is a half-way fix for now.)

### DIFF
--- a/src/hub.js
+++ b/src/hub.js
@@ -307,6 +307,8 @@ async function runBotMode(scene, entryManager) {
 
 document.addEventListener("DOMContentLoaded", async () => {
   const scene = document.querySelector("a-scene");
+  scene.removeAttribute("keyboard-shortcuts"); // Remove F and ESC hotkeys from aframe
+
   const hubChannel = new HubChannel(store);
   const entryManager = new SceneEntryManager(hubChannel);
   entryManager.init();
--- a/src/react-components/ui-root.js
+++ b/src/react-components/ui-root.js
@@ -440,7 +440,7 @@ class UIRoot extends Component {
     if (!this.props.forcedVREntryType || !this.props.forcedVREntryType.endsWith("_now")) {
       this.goToEntryStep(ENTRY_STEPS.audio_setup);
     } else {
-      setTimeout(this.onAudioReadyButton, 3000); // Need to wait otherwise input doesn't work :/
+      this.onAudioReadyButton();
     }
   };
 


src/hub.js
```diff
@@ -307,6 +307,8 @@ async function runBotMode(scene, entryManager) {
 
 document.addEventListener("DOMContentLoaded", async () => {
   const scene = document.querySelector("a-scene");
+  scene.removeAttribute("keyboard-shortcuts"); // Remove F and ESC hotkeys from aframe
```
johnshaughnessy: That looks good. 

If we didn't remove the component like this, we might set the `keyboard-shortcuts` component data's `enterVR` property to `false` (which would be a worse solution). 
https://github.com/aframevr/aframe/blob/master/src/components/scene/keyboard-shortcuts.js#L16

Removing the component like you did here is better because there's no point running a tick that isn't doing anything.
We could do the same for `vr-mode-ui`, which is also added by `a-scene` (https://github.com/aframevr/aframe/blob/master/src/core/scene/a-scene.js#L60).

In hubs we disable this thru the `html`.

```
src/react-components/avatar-selector.js:154:        <a-scene vr-mode-ui="enabled: false" ref={sce => (this.scene = sce)}>
src/hub.html:32:        vr-mode-ui="enabled: false"
src/scene.html:18:        vr-mode-ui="enabled: false"
```

It would be better if we could tell aframe not to add this default component in the first place.

src/react-components/ui-root.js
```diff
@@ -440,7 +440,7 @@ class UIRoot extends Component {
     if (!this.props.forcedVREntryType || !this.props.forcedVREntryType.endsWith("_now")) {
       this.goToEntryStep(ENTRY_STEPS.audio_setup);
     } else {
-      setTimeout(this.onAudioReadyButton, 3000); // Need to wait otherwise input doesn't work :/
+      this.onAudioReadyButton();
```
johnshaughnessy: Glad to see a mysterious `setTimeout` go away! Was there a race condition that got fixed?

src/hub.js
gfodor: I remember trying to do this and failing :(

src/react-components/ui-root.js
gfodor: No it's not a fix really, it just changes the UX for the `vr_entry_type=XXX_now` so that it actually doens't just enter, it just makes it one-click to enter.

PR #643, opened 2018-10-24
Feature/pinning

gfodor:
Adds a "pin" feature which allows you to lock an object's position in space (and prevent others from moving it accidentally.)

This PR also restyles the freeze menu to look more like the rest of our UI (though not perfect) and binds spacebar to freeze/unfreeze (controversial?) 

One nit I don't like about this branch right now is you can't pause/resume videos when they are pinned -- but I think the solution to this is to introduce a player UI of some kind to align it with the PDF paging. For now you'll just need to unpin to pause or resume.

This is *not* persistently saving these pinned objects on this branch, that will be for another branch.

Bonus fixes: dragging a video will not pause/play it, and fixed a CSS issue causing cursor to show up along the edges of the invite button.


src/hub.js
```diff
@@ -287,6 +290,8 @@ async function runBotMode(scene, entryManager) {
 
 document.addEventListener("DOMContentLoaded", async () => {
   const scene = document.querySelector("a-scene");
+  scene.removeAttribute("keyboard-shortcuts"); // Remove F and ESC hotkeys from aframe
```
gfodor: pulled from https://github.com/mozilla/hubs/pull/638

src/react-components/ui-root.js
```diff
@@ -440,7 +440,7 @@ class UIRoot extends Component {
     if (!this.props.forcedVREntryType || !this.props.forcedVREntryType.endsWith("_now")) {
       this.goToEntryStep(ENTRY_STEPS.audio_setup);
     } else {
-      setTimeout(this.onAudioReadyButton, 3000); // Need to wait otherwise input doesn't work :/
+      this.onAudioReadyButton();
```
gfodor: https://github.com/mozilla/hubs/pull/638

src/components/media-loader.js
```diff
@@ -42,7 +42,7 @@ AFRAME.registerComponent("media-loader", {
     if (this.el.body && this.shapeAdded && this.el.body.shapes.length > 1) {
       this.el.removeAttribute("shape");
       this.shapeAdded = false;
-    } else if (!this.hasBakedShapes) {
+    } else if (!this.hasBakedShapes && !box.isEmpty()) {
```
gfodor: these various `box.isEmpty()` checks are addressing a problem where if you paste a GLTF file that has nothing but `media-loader`, it fails on these code paths because the initial bounding box is empty

gfodor:
https://github.com/mozilla/hubs/pull/648

PR #645, opened 2018-10-24
Address issue with .el's being set recursively

gfodor:
See comment in code


gfodor:
merged into persistence pinning branch

PR #648, opened 2018-10-26
Feature/pinning persistence

gfodor:
Adds persistent pinned objects. (Really!)

This PR:

- Bumps to `networked-aframe` master which applies a heuristic to not remove "persisted" objects when the owner leaves, and has the whole room race to take ownership

- Fixes a bug in the GLTF loader which was leaving nested `el` properties on object3Ds mis-wired

- Takes ownership of scene-owned video objects a few seconds after connecting

- Flushes pinned object states to reticulum via pin/unpin channel messages

- Loads the `object.gltf` root scene at load time

- Adds the `media` and `pinnable` GLTF components. The schema for `media` is `{ id, src, pageIndex }` and for `pinnable` its `{ pinned }`

It also includes code from feature/pinning which:

Adds a "pin" feature which allows you to lock an object's position in space (and prevent others from moving it accidentally.)

This PR also restyles the freeze menu to look more like the rest of our UI (though not perfect) and binds spacebar to freeze/unfreeze (controversial?)

One nit I don't like about this branch right now is you can't pause/resume videos when they are pinned -- but I think the solution to this is to introduce a player UI of some kind to align it with the PDF paging. For now you'll just need to unpin to pause or resume.

Bonus fixes: dragging a video will not pause/play it, and fixed a CSS issue causing cursor to show up along the edges of the invite button.


package-lock.json
```diff
@@ -8984,7 +8984,7 @@
       "dev": true
     },
     "networked-aframe": {
-      "version": "github:mozillareality/networked-aframe#1dd7e0aa62bd119c214fec7e9137d4447f40cba0",
+      "version": "github:mozillareality/networked-aframe#f91ad0132e3622469b2f958a24d3cc7e43afac39",
```
johnshaughnessy: image
diff:
https://github.com/MozillaReality/networked-aframe/compare/96031430298335ffaba9eacf376cc0a398785288...MozillaReality:master?diff=split

src/components/block-button.js
```diff
@@ -14,11 +14,11 @@ AFRAME.registerComponent("block-button", {
   },
 
   play() {
-    this.el.addEventListener("click", this.onClick);
+    this.el.addEventListener("grab-start", this.onClick);
```
johnshaughnessy: I have a TODO item to revisit the user bindings for this so that click worked like before and to make sure block-button, remove-networked-icon-button, icon-button are DRY and consistent

src/utils/hub-channel.js
```diff
@@ -96,6 +96,14 @@ export default class HubChannel {
     this.channel.push("message", { body, type });
   };
 
+  pin = (id, gltfNode) => {
+    this.channel.push("pin", { id, gltf_node: gltfNode });
```
johnshaughnessy: so in order for persistent to outlast the last person who disconnects from janus, this could be caught with something in `hub_channel.ex` like
```
  def handle_in("pin", _payload, socket) do
    hub = socket |> hub_for_socket
    # do something to save the object in the payload for later use
    {:noreply, socket}
  end
```
is that right?

gfodor: yup! it's already in fact live :)

https://github.com/mozilla/reticulum/blob/master/lib/ret_web/channels/hub_channel.ex#L86

gfodor: https://github.com/mozilla/reticulum/pull/78

gfodor: note that PR includes the web push stuff too since i was working against that base branch

johnshaughnessy: Oh right, woops -- I was looking at reticulum from 100 commits ago O_o

src/components/pin-networked-object-button.js
```diff
@@ -0,0 +1,71 @@
...
+    this.el.sceneEl.removeEventListener("stateremoved", this._updateUI);
+
+    if (this.targetEl) {
+      this.targetEl.removeEventListener("pinned", this._updateUI);
+      this.targetEl.removeEventListener("unpinned", this._updateUI);
+    }
+  },
+
+  _updateUI() {
+    const isPinned = this.targetEl.components.pinnable && this.targetEl.components.pinnable.data.pinned;
```
netpro2k: I would use `getAttribute` here
```suggestion
    const isPinned = this.targetEl.getAttribute("pinnable", "pinned");
```

src/components/pin-networked-object-button.js
```diff
@@ -0,0 +1,71 @@
...
+
+    this.labelEl.setAttribute("text", { value: isPinned ? "un-pin" : "pin" });
```
netpro2k:
```suggestion
    this.labelEl.setAttribute("text", "value", isPinned ? "un-pin" : "pin" );
```

src/components/pin-networked-object-button.js
```diff
@@ -0,0 +1,71 @@
...
+
+    this.el.parentNode.querySelectorAll(this.data.hideWhenPinnedSelector).forEach(hideEl => {
```
netpro2k: This falls in line with my suggestion above about only caring about siblings, but given the current implementation it seems like this should be `this.uiEl.querySelectorAll`.

src/components/pinnable.js
```diff
@@ -0,0 +1,90 @@
+AFRAME.registerComponent("pinnable", {
+  schema: {
+    pinned: { default: false }
+  },
+
+  init() {
+    this._apply = this._apply.bind(this);
```
netpro2k: Apply is a bit of an overlaoded function name.. I would call something like `applyState` or `updateProperties`

src/components/pinnable.js
```diff
@@ -0,0 +1,90 @@
...
+    this.el.sceneEl.removeEventListener("stateremoved", this._apply);
+    this.el.removeEventListener("componentinitialized", this._allowApplyOnceComponentsReady);
+  },
+
+  update() {
+    this._apply();
+    this._fireEvents();
+  },
+
+  _fireEvents() {
```
netpro2k: Odd to break these into 3 methods. I would just have 1. Also since the events bubble I don't see the need for the extra event on the scene (just check the target of the "pinned" event)

src/components/pinnable.js
```diff
@@ -0,0 +1,90 @@
...
+  },
+
+  _fireUnpinnedEvents() {
+    if (!this.data.pinned) {
+      this.el.emit("unpinned");
+      this.el.sceneEl.emit("object_unpinned", { el: this.el });
+    }
+  },
+
+  _allowApplyOnceComponentsReady() {
```
netpro2k: Yuck..  But thems component dependencies.

src/components/pinnable.js
```diff
@@ -0,0 +1,90 @@
...
+    if (!this._allowApply && this.el.components.grabbable && this.el.components.stretchable) {
+      if (this.el.components.grabbable.data.maxGrabbers !== 0) {
+        this.prevMaxGrabbers = this.el.components.grabbable.data.maxGrabbers;
```
netpro2k: Seems like you would want to save this at the point you are about to overwrite it instead of just at load time, otherwise any changes that happened after the fact to maxGrabbers would be clobbered.

src/components/remove-networked-object-button.js
```diff
@@ -1,18 +1,23 @@
 AFRAME.registerComponent("remove-networked-object-button", {
   init() {
+    this.scene = document.querySelector("a-scene");
```
netpro2k: No need to keep an extra scene reference around, use `this.el.sceneEl`

src/components/set-yxz-order.js
```diff
@@ -0,0 +1,5 @@
+AFRAME.registerComponent("set-yxz-order", {
```
netpro2k: Ho boy, I know why we might want this but it scares me to have another thing doing this... I guess its better than adding a rotation component just for its side-effects...

src/components/sticky-object.js
```diff
@@ -60,6 +60,8 @@ AFRAME.registerComponent("sticky-object", {
   },
 
   _onGrab() {
+    if (!this.el.components.grabbable || this.el.components.grabbable.data.maxGrabbers === 0) return;
```
netpro2k: Was going to point out that this was weird since things that aren't `grabbable` shouldn't get grab events (not actually true) and things with `maxGrabbers` of 0 should also not get grab events (also not true). Fooled by logic!

src/components/stop-event-propagation.js
```diff
@@ -0,0 +1,21 @@
...
+
+AFRAME.registerComponent("stop-event-propagation", {
+  multiple: true,
+
+  schema: {
+    event: { type: "string" }
+  },
+
+  play() {
+    console.log("HELLO " + this.data.event);
```
netpro2k:
```suggestion
```

src/components/super-networked-interactable.js
```diff
@@ -82,7 +88,15 @@ AFRAME.registerComponent("super-networked-interactable", {
     this.el.setAttribute("body", { type: "static" });
     this.el.emit("grab-end", { hand: this.hand });
     this.hand = null;
-    this.counter.deregister(this.el);
+    this._syncCounterRegistration();
+  },
+
+  _onPinned: function() {
```
netpro2k: I would ditch these 2 methods and just bind `_syncCounterRegistration` to the events directly

src/components/super-networked-interactable.js
```diff
@@ -93,6 +107,19 @@ AFRAME.registerComponent("super-networked-interactable", {
     }
   },
 
+  _syncCounterRegistration: function() {
+    const el = this.networkedEl;
+    if (!el) return;
+
+    const isPinned = el.components["pinnable"] && el.components["pinnable"].data.pinned;
```
netpro2k:
```suggestion
    const isPinned = el.getAttribute("pinnable", "pinned");
```

src/gltf-component-mappings.js
```diff
@@ -72,3 +72,22 @@ AFRAME.GLTFModelPlus.registerComponent("nav-mesh", "nav-mesh", (el, _componentNa
   // nav-mesh-helper will query for it later.
   el.setAttribute("nav-mesh");
 });
+
+AFRAME.GLTFModelPlus.registerComponent("pinnable", "pinnable");
+
+AFRAME.GLTFModelPlus.registerComponent("media", "media", (el, componentName, componentData) => {
```
netpro2k: If we are going to make this sort of combo component I would propose we at least point it out a bit more by calling it something like `networked-media`

src/gltf-component-mappings.js
```diff
@@ -72,3 +72,22 @@ AFRAME.GLTFModelPlus.registerComponent("nav-mesh", "nav-mesh", (el, _componentNa
...
+  if (componentData.id) {
+    el.setAttribute("networked", {
+      template: "#interactable-media",
+      owner: "scene",
+      lastOwnerTime: 1,
```
netpro2k: `lastOwnerTime` is not actually a schema property of `networked` its just a value on it, so if you really want to do this you would need to do `el.components.networked.lastOwnerTime = 1`, though I suspect since it was working like this you don't actually need to set it at all.

src/gltf-component-mappings.js
```diff
@@ -72,3 +72,22 @@ AFRAME.GLTFModelPlus.registerComponent("nav-mesh", "nav-mesh", (el, _componentNa
...
+      networkId: componentData.id
+    });
+  }
+
+  el.setAttribute("media-loader", { src: componentData.src, resize: true, resolve: true });
+
+  if (componentData.pageIndex) {
```
netpro2k: should be fine to just always include ` index: componentData.pageIndex` in the previous setAttribute

src/hub.js
```diff
@@ -247,36 +256,39 @@ async function handleHubChannelJoined(entryManager, hubChannel, data) {
...
-        hubChannel.channel.push("naf", payload);
-      };
-    })
-    .catch(connectError => {
-      // hacky until we get return codes
-      const isFull = connectError.error && connectError.error.msg.match(/\bfull\b/i);
-      console.error(connectError);
-      remountUI({ roomUnavailableReason: isFull ? "full" : "connect_error" });
-      entryManager.exitScene();
+    while (!scene.components["networked-scene"] || !scene.components["networked-scene"].data) await nextTick();
```
netpro2k: Apparently this is not new code.. but wat?

src/hub.js
```diff
@@ -225,6 +229,11 @@ async function handleHubChannelJoined(entryManager, hubChannel, data) {
 
   console.log(`Scene URL: ${sceneUrl}`);
   const environmentScene = document.querySelector("#environment-scene");
+  const objectsScene = document.querySelector("#objects-scene");
```
netpro2k: Still a bit concerned about having all the pinned entities living in a few layers of wrapper object, surprised it hasn't caused more issues..

src/scene-entry-manager.js
```diff
@@ -213,6 +214,28 @@ export default class SceneEntryManager {
       spawnMediaInfrontOfPlayer(e.detail, contentOrigin);
     });
 
+    this.scene.addEventListener("object_pinned", e => {
```
netpro2k: as per suggestion above in `pinned` component, I would just listen for the `pinned` event and then use `e.target` as `el`

src/systems/userinput/userinput.js
```diff
@@ -66,7 +66,7 @@ AFRAME.registerSystem("userinput", {
   init() {
     this.frame = {};
 
-    this.activeSets = new Set([sets.global]);
+    this.activeSets = new Set([sets.global, sets.globalPost]);
```
netpro2k: Annoying that we had to add this, but we will be able to remove it soon with the work John is doing now on determining binding dependency

src/scene-entry-manager.js
```diff
@@ -213,6 +214,28 @@ export default class SceneEntryManager {
...
+      const el = e.detail.el;
+      const networkId = el.components.networked.data.networkId;
+      const gltfNode = pinnedEntityToGltf(el);
```
netpro2k: Hmm I wonder if it makes more sense to just send the data over to reticulum and have it do this part. Seems a bit odd that the wire protocol is gltf nodes... But maybe thats nice?

src/components/media-loader.js
```diff
@@ -224,16 +224,16 @@ AFRAME.registerComponent("media-pager", {
       // if this is the first image we ever loaded, set up the UI
       if (this.toolbar == null) {
         const template = document.getElementById("paging-toolbar");
-        this.el.appendChild(document.importNode(template.content, true));
+        this.el.querySelector(".interactable-ui").appendChild(document.importNode(template.content, true));
```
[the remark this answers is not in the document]
gfodor: I don't remember specifically, but overall I ended up moving all of the UI components into a nested `interactable-ui` subentity -- i'll reply here if i remember the motivating case. overall this PR really tightens up and deals with a variety of weirdness around the UI interaction events for the pause mode being ephemerally togglable etc and this was one of the things that I did along the way

src/components/media-views.js
```diff
@@ -230,13 +230,25 @@ AFRAME.registerComponent("media-video", {
...
 
     this.lastUpdate = 0;
 
     NAF.utils.getNetworkedEntity(this.el).then(networkedEl => {
       this.networkedEl = networkedEl;
       this.updatePlaybackState();
+
+      // For scene-owned videos, take ownership after a random delay if nobody
+      // else has so there is a timekeeper.
+      if (NAF.utils.getNetworkOwner(this.networkedEl) === "scene") {
```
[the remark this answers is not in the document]
gfodor: hmm -- basically this technique is only needed for objects which rely upon an ongoing "simulation step" to be run by a client for their network state to be sane. if we want to generalize this i think we would want to factor that concept out specifically and only perform this heuristic for objects that have that characteristic. (another theoretical example would be a scene owned object that requires physics simulation.)

src/components/media-views.js
```diff
@@ -252,12 +264,28 @@ AFRAME.registerComponent("media-video", {
...
   // aframe component pause, unrelated to video
   pause() {
-    this.el.removeEventListener("click", this.togglePlayingIfOwner);
+    this.el.removeEventListener("grab-start", this._grabStart);
+    this.el.removeEventListener("grab-end", this._grabEnd);
+  },
+
+  _grabStart() {
+    if (!this.el.components.grabbable || this.el.components.grabbable.data.maxGrabbers === 0) return;
+    if (this.el.sceneEl.is("frozen")) return; // Ignore clicks on video when clicking on buttons.
```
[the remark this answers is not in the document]
gfodor: oh, my latest set of changes (introducing a component that stops event propagation above the UI subtree) may have addressed this

src/components/pinnable.js
gfodor: most of the hairy bits of this have been GCed from my brain at this point, but I can assure you there were several revisions of this that went down that road :) I can take another go at it. A-Frame lifecycle fun

src/components/set-yxz-order.js
gfodor: yeah this is in the "step forward" category but not going to argue that this isn't fundamentally concerning. i didn't know until this week that information loss when converting through Eulers is axis-dependent

src/components/sticky-object.js
gfodor: /shrug

src/components/stop-event-propagation.js
gfodor: GOODBYE

src/gltf-component-mappings.js
gfodor: as it's written right now, it's parameterized so that it could be networked or non-networked -- the intent was this was a generic catch-all component for any URLs you want to put into the scene (and perhaps the contract includes 'is resolved at load time vs scene export time')

src/gltf-component-mappings.js
gfodor: the component below is `media-pager` -- so I think you may have just misread this. `media-loader` won't deal with `index` right?

src/hub.js
gfodor: it had some issues :) but i fixed the ones I found

src/hub.js
gfodor: this is waiting on the creation of the `networked-scene` component -- isn't this "idiomatic?" (heavy quotes)

src/scene-entry-manager.js
gfodor: yeah I feel like this is the proper separation of concerns. it's nice that reticulum can remain completely agnostic to aframe and networked-aframe schema.

src/components/media-views.js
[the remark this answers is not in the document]
gfodor: yes I can remove this

src/components/pin-networked-object-button.js
```diff
@@ -0,0 +1,71 @@
...
+
+    let uiElSearch = this.el;
+
+    do {
+      uiElSearch = uiElSearch.parentNode;
+      this.uiEl = uiElSearch.querySelector(this.data.uiSelector);
+    } while (uiElSearch && !this.uiEl);
+
+    this.uiEl = uiElSearch.querySelector(this.data.uiSelector);
+    this.labelEl = this.uiEl.querySelector(this.data.labelSelector);
```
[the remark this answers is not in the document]
gfodor: ah I believe I used to use `uiEl` in prior versions, so now I agree this looks unnecessary

src/components/pin-networked-object-button.js
gfodor: i changed to the sibling approach overall now

src/components/pinnable.js
gfodor: the dusty cobwebs of my java-isms using method names to document poking through. i think you're right in this case

src/components/super-networked-interactable.js
gfodor: this seems to not be a thing, the second arg isn't valid. but single arg works

src/hub.js
netpro2k: Probably would be more correct to wait for an `componentinitialized` event

PR #650, opened 2018-10-27
Feature/spawn chat messages

gfodor:
Adds a "spawn" button to messages, which lets you spawn messages into the space and also spawn singular emojis or media if the message is just a URL. Adds multline chat and the ability to wrap chat messages in backticks to make them render monospaced. It also adds the `t` hotkey to focus the chat text entry.

image

Also touches up the photo chat message and styling a bit.


src/assets/stylesheets/presence-log.scss
```diff
@@ -35,6 +64,29 @@
       max-width: 75%;
     }
 
+    :local(.spawn-message) {
```
netpro2k: IN my branch I split this off into an `icon-butotn` class. We should also technically probably do the hover effect in CSS instead of needing 2 images, especially since we don't spritesheet them

src/hub.html
```diff
@@ -32,6 +32,7 @@
         vr-mode-ui="enabled: false"
         stats-plus="false"
         action-to-event__mute="path: /actions/muteMic; event: action_mute;"
+        action-to-event__focus_chat="path: /actions/focusChat; event: action_focus_chat;"
```
netpro2k: Nice, `action-to-event` was mostly meant as a way to port old code/components, but I guess this is maybe the preferred way to interact with anything without an existing tick function.

src/hub.js
```diff
@@ -464,8 +465,9 @@ document.addEventListener("DOMContentLoaded", async () => {
   hubPhxChannel.on("message", ({ session_id, type, body }) => {
     const userInfo = hubPhxPresence.state[session_id];
     if (!userInfo) return;
+    const maySpawn = scene.is("entered");
 
-    addToPresenceLog({ name: userInfo.metas[0].profile.displayName, type, body });
+    addToPresenceLog({ name: userInfo.metas[0].profile.displayName, type, body, maySpawn });
```
netpro2k: Spawn state shouldn't really be a property of the message, it should probably be passed in from ui-root (which can get it from the scene, and know about changes with the `state-change` event). That way messages that were added before you joined still get buttons. Probably minor in practice, but feels a bit odd.

src/react-components/presence-log.js
```diff
@@ -5,6 +5,107 @@ import classNames from "classnames";
...
+function ChatMessage(props) {
+  const ref = React.createRef();
+
+  const handleSpawn = () => {
+    if (props.body.match(urlRegex)) {
+      document.querySelector("a-scene").emit("add_media", props.body);
+      return;
+    }
+
+    const el = ref.current;
```
netpro2k: Rather than try and render the current element and doing all this CSS trickery I would maybe just create a new element and then `ReactDOM.render`ing to that. Like 
```
const tmpEl = document.createElement("div");
ReactDOM.render(<ChatMessage {...props} renderForSpawn />, tmpEl, () => { // do SVG stuff with tmpEl })
```

src/react-components/ui-root.js
```diff
@@ -440,7 +440,7 @@ class UIRoot extends Component {
     if (!this.props.forcedVREntryType || !this.props.forcedVREntryType.endsWith("_now")) {
       this.goToEntryStep(ENTRY_STEPS.audio_setup);
     } else {
-      setTimeout(this.onAudioReadyButton, 3000); // Need to wait otherwise input doesn't work :/
+      this.onAudioReadyButton(); // Need to wait otherwise input doesn't work :/
```
netpro2k: Nice that we were able to get rid of this timeout (how?) but we should also remove the comment

src/scene-entry-manager.js
```diff
@@ -218,7 +218,8 @@ export default class SceneEntryManager {
     });
 
     document.addEventListener("paste", e => {
-      if (e.target.nodeName === "INPUT" && document.activeElement === e.target) return;
+      if ((e.target.nodeName === "INPUT" || e.target.nodeName === "TEXTAREA") && document.activeElement === e.target)
```
netpro2k: `e.matches("input, textarea")` would be a bit cleaner for the first part of this. (might technically be slower but this isn't hot code)

src/utils/serialize-element.js
```diff
@@ -0,0 +1,194 @@
+// https://stackoverflow.com/questions/6209161/extract-the-current-dom-and-print-it-as-a-string-with-styles-intact
```
netpro2k: I don't fully understand how the SVG hack works, will we still need this if we render the react component in a way where it can easily toggle its own styles?

src/react-components/ui-root.js
gfodor: haha whoops. `2d_now` no longer really works as it used to -- it doesn't just enter the scene now, it requires one-click to enter. this was a small change to make it less painful, but a separate PR will be necessary to fully restore the feature. (it regressed when I did the overhaul to the landing page)

src/utils/serialize-element.js
gfodor: the intention here was to make it so the internal styles and structure of the chat messages will be future proofed -- for example, if we were to add markdown support, it would "just work" without altering the SVG generation code. so this bit of code basically just inlines all the styles in a way that is agnostic to our overall CSS structure.

i realize I'm slightly breaking this contract right now by having some light styling applied by the wrapper div, but it still seems like a good design to make the inner parts of the chat message and the CSS styles be 1-1 against whatever we would render in the real DOM for the chat history. the current emoji library and blockquote functionality illustrate the versatility somewhat -- there was no new code needed in the SVG generation to pull that functionality in.

src/hub.js
gfodor: ah this was more important in earlier versions where I was tagging the message as spawnable based upon a more complex expression (if the room has been entered, and you were the origin of the message, and I was presuming we would not have it spawnable if it was a photo eventually etc.) i'll leave it for now since it's a minor refactor if/when it starts breaking down

src/react-components/presence-log.js
gfodor: going to try this today, if it works I'll add "ctrl-shift enter will do the same thing as enter + hit spawn button, but not echo to chat"

PR #653, opened 2018-10-30
Allow sharing photos from the chat log

netpro2k:
Adds a share button to the "Foo took a photo" chatlog messages.

This uses `navigator.share` if available, otherwise falls back to twitter. I also reworked how we were sharing to twitter in other places to use twitter intents popup, which is a bit nicer style wise.

todo:
- [x] Add proper share icon
- [x] Figure out correct copy for prefilled message


src/react-components/invite-dialog.js
```diff
@@ -89,9 +85,9 @@ export default class InviteDialog extends Component {
             )}
           {this.props.allowShare &&
             !navigator.share && (
-              <a href={tweetLink} className={styles.linkButton} target="_blank" rel="noopener noreferrer">
+              <button className={styles.linkButton} onClick={this.shareClicked.bind(this, shortLink)}>
```
gfodor: I would have written this to be `onClick=() => this.shareClicked(shortLink)` but this is probably OK -- i find i don't use `bind` anymore due to the property syntax. (currying the second arg is interesting tho, very FP :))

src/react-components/presence-log.js
```diff
@@ -4,11 +4,56 @@ import styles from "../assets/stylesheets/presence-log.scss";
 import classNames from "classnames";
 import { FormattedMessage } from "react-intl";
 import ChatMessage from "./chat-message";
+import { share } from "../utils/share";
+
+function SpawnPhotoMessage({ name, body: { src: url }, className, maySpawn, hubId }) {
+  let landingPageUrl = new URL(url);
+  const [hostname, port] = process.env.RETICULUM_SERVER.split(":");
+  console.log(hostname, port, landingPageUrl.port);
```
gfodor: console log

gfodor: I might consider rolling this routine into `phoenix-utils.js`, its where I've tried to centralize references to `RETICULUM_SERVER`

gfodor: any reason `getReticulumFetchUrl(landingPageUrl.pathname.replace(".png", ".html"))` wouldn't work? also might be worth future proofing this for other file extensions by just dropping the file extension instead of assumign `.png`

src/react-components/presence-log.js
```diff
@@ -4,11 +4,56 @@ import styles from "../assets/stylesheets/presence-log.scss";
...
+    </div>
+  );
+}
+SpawnPhotoMessage.propTypes = {
+  name: PropTypes.string,
+  maySpawn: PropTypes.bool,
+  body: PropTypes.object,
+  className: PropTypes.string,
+  hubId: PropTypes.string
+};
```
gfodor: `chat-message.js` is a thing so maybe `photo-message.js` should be too

src/utils/share.js
```diff
@@ -0,0 +1,17 @@
+export function share(opts) {
```
gfodor: `export default` perhaps?

src/utils/share.js
```diff
@@ -0,0 +1,17 @@
...
+  if (navigator.share) {
```
gfodor: I'd add a comment
`// Use navigator.share otherwise just post to Twitter`

src/utils/share.js
netpro2k: Eh I figured there might be other share utils at some point

PR #655, opened 2018-10-30
Feature/screensharing v2

gfodor:
Screen/window/camera sharing in the HUD. 

The new screensharing functionality sits on top of the media content spawning -- it introduces a new URL scheme `webrtc://<client id>` which will cause a video element to be created and will bind the first video stream of `client id` to that object's video texture.

This PR also reduces the size of the pause button since the wider HUD now with the new video sharing buttons is problematic on mobile otherwise.


src/assets/stylesheets/2d-hud.scss
```diff
@@ -144,3 +164,42 @@
 :local(.iconButton.mobile-media-picker) {
   background-image: url(../hud/spawn_photo.png);
 }
+
+:local(.iconButton.share_camera) {
```
netpro2k: We really do need to get around to spritesheeting these...

src/components/media-views.js
```diff
@@ -319,15 +327,17 @@ AFRAME.registerComponent("media-video", {
         return;
       }
 
-      texture.audioSource = this.el.sceneEl.audioListener.context.createMediaElementSource(texture.image);
-      this.video = texture.image;
+      if (!src.startsWith("webrtc://")) {
```
netpro2k: I know our video streams happen to not have audio, but why special case this?

src/components/media-views.js
```diff
@@ -79,19 +79,27 @@ async function createGIFTexture(url) {
...
+    videoEl.srcObject = new MediaStream(stream.getVideoTracks());
+  }
+
   return videoEl;
 }
 
 function createVideoTexture(url) {
-  return new Promise((resolve, reject) => {
-    const videoEl = createVideoEl(url);
+  return new Promise(async (resolve, reject) => {
```
netpro2k: Might as well just ditch promise and make `createVideoTexture` itself `async`

src/components/media-views.js
```diff
@@ -79,19 +79,27 @@ async function createGIFTexture(url) {
...
   videoEl.loop = true;
   videoEl.crossOrigin = "anonymous";
-  videoEl.src = src;
+
+  if (!src.startsWith("webrtc://")) {
+    videoEl.src = src;
+  } else {
+    const streamClientId = src.substring(9);
+    const stream = await NAF.connection.adapter.getMediaStream(streamClientId, "video");
+    videoEl.srcObject = new MediaStream(stream.getVideoTracks());
```
netpro2k: I would just use the `MediaStream` directly rather than making a new one with just video.

src/components/media-views.js
```diff
@@ -79,19 +79,27 @@ async function createGIFTexture(url) {
...
-function createVideoEl(src) {
+async function createVideoEl(src) {
   const videoEl = document.createElement("video");
   videoEl.setAttribute("playsinline", "");
   videoEl.setAttribute("webkit-playsinline", "");
   videoEl.loop = true;
   videoEl.crossOrigin = "anonymous";
-  videoEl.src = src;
+
+  if (!src.startsWith("webrtc://")) {
```
netpro2k: hmm, this made up protocol is a bit concerning, if we are going to make it up we might as well it something that is clearly app specific like `hubs://[hubid]/[clientId]/video`

src/react-components/webrtc-screenshare-dialog.js
```diff
@@ -0,0 +1,32 @@
+import React, { Component } from "react";
+import DialogContainer from "./dialog-container.js";
+
+export default class WebRTCScreenshareDialog extends Component {
```
netpro2k: Since this dialog is shown when screen sharing is unsupported, the copy should probably actually say something along the lines of "Your browser appears to not support screen sharing". Also the component name should maybe be something like `WebRTCScreenshareUnsupportedDialog`.

src/scene-entry-manager.js
```diff
@@ -268,6 +240,84 @@ export default class SceneEntryManager {
         }
       }
     });
+
+    let currentVideoShareEntity;
+    let isHandlingVideoShare = false;
```
netpro2k: Is `isHandlingVideoShare` ever false when `currentVideoShareEntity` is defined? If not I would ditch the extra bool.

src/utils/media-utils.js
```diff
@@ -22,6 +22,8 @@ function b64EncodeUnicode(str) {
 }
 
 export const proxiedUrlFor = (url, index) => {
+  if (url.startsWith("webrtc://")) return url;
```
netpro2k: We should probably more appropriately check for `!(http || https)` since that's all it makes sense to send farspark (The experiment I did with webtorrent also required a similar change to skip `magnet://` links)

src/utils/media-utils.js
```diff
@@ -44,6 +46,7 @@ export const resolveUrl = async (url, index) => {
 };
 
 export const guessContentType = url => {
+  if (url.startsWith("webrtc://")) return "video/webrtc";
```
netpro2k: I don't think `video/webrtc` is a valid mimetype. Since we probably can't easily know I would do `video/*`

src/utils/media-utils.js
```diff
@@ -126,6 +129,9 @@ export const addMedia = (src, template, contentOrigin, resolve = false, resize =
         entity.setAttribute("media-loader", { src: "error" });
       });
   }
+  if (src instanceof MediaStream) {
```
netpro2k:
```suggestion
  else if (src instanceof MediaStream) {
```

src/utils/pinned-entity-to-gltf.js
```diff
@@ -20,7 +20,14 @@ export default function pinnedEntityToGltf(el) {
   if (!equalArray(scale, [1, 1, 1])) gltfNode.scale = scale;
 
   if (components["media-loader"]) {
-    gltfComponents.media = { src: components["media-loader"].data.src, id: networkId };
+    const mediaSrc = components["media-loader"].data.src;
+
+    if (mediaSrc.startsWith("webrtc://")) {
```
netpro2k: Ideally I shouldn't be able to pin them in the first place, but I am guessing that was a bit more of a pain to network.

src/scene-entry-manager.js
```diff
@@ -268,6 +240,84 @@ export default class SceneEntryManager {
...
+      const newStream = await navigator.mediaDevices.getUserMedia(constraints);
+      const videoTracks = newStream ? newStream.getVideoTracks() : [];
+
+      if (videoTracks.length > 0) {
+        newStream.getVideoTracks().forEach(track => mediaStream.addTrack(track));
+        NAF.connection.adapter.setLocalMediaStream(mediaStream);
+        currentVideoShareEntity = spawnMediaInfrontOfPlayer(mediaStream, undefined);
+
+        // Wire up custom removal event which will stop the stream.
+        currentVideoShareEntity.setAttribute("emit-scene-event-on-remove", "event:action_end_video_sharing");
```
netpro2k: Aframe already emits a `child-detached` event which you could use, though I guess you are already listening to `action_end_video_sharing` for the UI anyway..

src/components/media-views.js
gfodor: this was based on the previous implementation, which had a comment I failed to carry over. will fix

src/components/media-views.js
gfodor: i'll verify but this was breaking if i didn't avoid it iirc

gfodor: yes get this error in firefox `The captured HTMLMediaElement is playing a MediaStream. Applying volume or mute status is not currently supported.`

gfodor: will leave a comment for now

src/utils/media-utils.js
gfodor: going with `video/vnd.hubs-webrtc` for now

src/components/media-views.js
gfodor: getting the hub id in these places was going to add more incidental complexity than i was comfortable with, so for now i went with `hubs://clients/<client id>/video` -- these URLs are not saved anywhere so its probably OK for now. i agree having the full hub id would be more correct

PR #660, opened 2018-10-31
Don't use Farspark to proxy CORS-accessible stuff

mquander:
**WIP - we need to iron out CSP stuff before merging this.**

We now pass the canonical URL for media down into all of the media view components and into `gltf-model-plus`, and each one will gracefully handle the case where the media turns out to not be CORS-accessible. (The methodology for each one looks quite different because of the different ways they load the media and try to get errors out.)

Tested with a variety of media sources. Note that, contrary to what we thought, Youtube URLs retrieved via youtube-dl are not CORS-accessible. However, Sketchfab and Google Poly URLs are, so this will reduce latency and Farspark load a lot in rooms where people are loading 3D models in particular.

### DIFF
--- a/src/components/gltf-model-plus.js
+++ b/src/components/gltf-model-plus.js
@@ -7,6 +7,7 @@ import cubeMapPosY from "../assets/images/cubemap/posy.jpg";
 import cubeMapNegY from "../assets/images/cubemap/negy.jpg";
 import cubeMapPosZ from "../assets/images/cubemap/posz.jpg";
 import cubeMapNegZ from "../assets/images/cubemap/negz.jpg";
+import { loadProxyable } from "../utils/media-utils.js";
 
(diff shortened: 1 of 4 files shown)


mqp:
Actually, Dom just made the very good point to me that some folks probably actively dislike us directly loading third-party media URLs without prompting in Hubs rooms, because it leaks their IP address to anyone in the room who wants to get it. Given that we are moving away from "Google Hangouts VR" and toward "Discord channel VR", I'm not even convinced this PR is a good idea anymore.

gfodor:
Also we are in a decent spot now wrt proxying stuff reliably, this original
PR was at least in part motivated by the expected burden of scaling up
farspark to support proxying all media traffic

gfodor:
I think we can close this.

PR #662, opened 2018-11-01
Add custom room name dialog to scene landing page

gfodor:
This addresses the current gap in functionality which prevents custom room names for custom scenes. Eventually, once auth is done, you'll be able to rename a room from within the room page, but this seemed like the quickest way to enable the use-case of a custom room name for a room for a custom scene. There's now an ellipsis on the create scene landing page create button that pops a dialog to prompt the user for the custom room name.

image

### DIFF
--- a/src/react-components/create-room-dialog.js
+++ b/src/react-components/create-room-dialog.js
@@ -4,8 +4,9 @@ import DialogContainer from "./dialog-container.js";
 
 const HUB_NAME_PATTERN = "^[A-Za-z0-9-'\":!@#$%^&*(),.?~ ]{4,64}$";
 
-export default class CreateObjectDialog extends Component {
+export default class CreateRoomDialog extends Component {
   static propTypes = {
+    includeScenePrompt: PropTypes.bool,
     onCustomScene: PropTypes.func,
     onClose: PropTypes.func
   };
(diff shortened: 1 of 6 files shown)


src/react-components/scene-ui.js
```diff
@@ -93,9 +97,16 @@ class SceneUI extends Component {
...
               </div>
-              <button onClick={this.createRoom}>
-                <FormattedMessage id="scene.create_button" />
-              </button>
+              <div className={styles.createButtons}>
+                <button className={styles.createButton} onClick={this.createRoom}>
+                  <FormattedMessage id="scene.create_button" />
+                </button>
+                <button className={styles.optionsButton} onClick={() => this.setState({ showCustomRoomDialog: true })}>
+                  <i>
```
brianpeiris: This `<i>` is redundant I think.

src/react-components/create-room-dialog.js
```diff
@@ -38,13 +44,15 @@ export default class CreateObjectDialog extends Component {
...
+                  type="url"
+                  placeholder="URL to Scene GLTF or GLB (Optional)"
+                  className="custom-scene-form__link_field"
+                  value={this.state.customSceneUrl}
+                  onChange={e => this.setState({ customSceneUrl: e.target.value })}
+                />
+              )}
               <div className="custom-scene-form__buttons">
                 <button className="custom-scene-form__action-button">
                   <span>create</span>
```
brianpeiris: Capitalize this while we're at it?

PR #663, opened 2018-11-01
Fix/649/continued

johnshaughnessy:
This changes the way we define bindings in the userinput system such that:
- Bindings are put in "dependency order", meaning that if binding A writes to a path in its `dest` that binding B lists in its `src`, then A will come before B.
- When registeredMappings change, for all bindings we compute the set of all OTHER bindings that might mask (aka "override") this one for some combination of active action sets.
- When active action sets change, we figure out which bindings are both active and unmasked.

So we no longer have to worry about writing the bindings in just the right order (and we remove the `postGlobal` action set), and we can provide better debugging information about how binding order and masks.


src/systems/userinput/userinput-debug.js
```diff
@@ -0,0 +1,96 @@
...
+    sb.push("in set");
+    sb.push(" : ");
+    sb.push(s);
+    sb.push("\n");
+  }
+  sb.push("\n");
+  sb.push("}\n");
+  return sb.join("");
+};
+AFRAME.registerSystem("userinput-debug", {
```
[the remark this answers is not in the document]
johnshaughnessy: I like this as an escape hatch to specify anything with order dependence... 
Order dependence in general is related to today's thoughts on bindings in that we want to specify a partial order on systems ticks or binding xforms. E.g. A scenario with systems A, B, C where you want A to run before B or C but don't care whether B or C runs first.

src/systems/userinput/userinput.js
```diff
@@ -145,21 +148,86 @@ AFRAME.registerSystem("userinput", {
...
+            this.actives.push(active);
+            runners.push(binding);
+            this.overrides.push([]);
+          }
+        }
+      }
+
+      const maxAmongActive = (path, map) => {
+        let max = { priority: -1 };
+        const bindings = map.get(path);
```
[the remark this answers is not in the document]
johnshaughnessy: ah, right

src/systems/userinput/userinput.js
```diff
@@ -165,21 +247,46 @@ AFRAME.registerSystem("userinput", {
...
+        this.prevSortedBindings = this.sortedBindings;
+      }
+      this.masks = computeMasks(this.sortedBindings);
+    }
 
+    resolveActionSets();
     for (const { set, value } of this.pendingSetChanges) {
       this.activeSets[value ? "add" : "delete"](set);
     }
+    const activeSetsChanged = this.pendingSetChanges.length; // TODO: correct this
```
johnshaughnessy: Need to check if the active sets from last frame != the active sets from this frame

src/systems/userinput/userinput.js
```diff
@@ -155,7 +235,9 @@ AFRAME.registerSystem("userinput", {
       e => {
         for (const device of this.activeDevices) {
           if (device.gamepad === e.gamepad) {
+            console.warn("NEED TO UPDATE REGISTERED MAPPINGS WHEN GAMEPAD DISCONNECTED!");
```
johnshaughnessy: Need to be able to figure out what should be in registeredMappings based on the active devices and whether or not you're in vr

src/systems/userinput/userinput.js
```diff
@@ -22,36 +21,113 @@ import { viveUserBindings } from "./bindings/vive-user";
...
 import { GamepadDevice } from "./devices/gamepad";
 import { gamepadBindings } from "./bindings/generic-gamepad";
 
-const priorityMap = new Map();
-function prioritizeBindings(registeredMappings, activeSets) {
-  const activeBindings = new Set();
-  priorityMap.clear();
-  for (const mapping of registeredMappings) {
+const satisfiesPath = (binding, path) => {
+  return Object.values(binding.dest) && Object.values(binding.dest).indexOf(path) !== -1;
```
mquander: Maybe a little suspicious performance here.

johnshaughnessy: The `indexOf` can't be replaced with `===` because we want a binding with `/foo/bar` to "satisfy" any path like `foo/bar/baz`. Dom pointed out that `indexOf` isn't right because we actually want to only check whether the two paths start with the same words between the `/`s... 

e.g. "/foo/ba" should not "satisfy" "/foo/bar" but right now it does

src/systems/userinput/userinput.js
```diff
@@ -22,36 +21,113 @@ import { viveUserBindings } from "./bindings/vive-user";
...
+    }
+  }
+
+  const sorted = [];
+  while (unsorted.length > 0) {
+    const binding = unsorted.shift();
+    if (satisfiedBy(binding, sorted)) {
+      sorted.push(binding);
+    } else {
+      unsorted.push(binding);
```
mquander: I guess you tested this and it turns out to be acceptably fast, but I'm honestly surprised that it is. Isn't this N^2 with respect to the number of bindings? And there can be quite a fair number of bindings active, like maybe one or two hundred, right? And the work for each step is not completely trivial, it's making some garbage and doing some string compares and so on.

It seems like you could do a lot of different ways that were faster, but maybe the simplest would be to maintain a version of the bindings that were indexed by their destination paths. Then checking whether a binding was satisfied would be fast.

src/systems/userinput/userinput.js
mquander: Oh, I see. I forgot about that part.

src/systems/userinput/userinput.js
```diff
@@ -22,36 +21,123 @@ import { viveUserBindings } from "./bindings/vive-user";
...
+function canMask(masker, masked) {
+  if (masker.priority === undefined) {
+    masker.priority = 0;
+  }
+  if (masked.priority === undefined) {
+    masked.priority = 0;
+  }
+  if (masked.priority >= masker.priority) return false;
+  for (const maskerPath of Object.values(masker.src)) {
+    for (const maskedPath of Object.values(masked.src)) {
```
mquander: This seems like another inner loop that is unnecessarily inefficient, unless the JIT knows how to optimize it.

src/systems/userinput/userinput.js
```diff
@@ -22,36 +21,123 @@ import { viveUserBindings } from "./bindings/vive-user";
...
+  const actives = [];
+  for (const row in sortedBindings) {
+    actives[row] = isActive(sortedBindings[row], activeSets);
+  }
+
+  const masked = [];
+  for (const row in sortedBindings) {
+    for (const col in sortedBindings) {
+      const rowMask = masked[row] || [];
+      if (masks[Number(row) * sortedBindings.length + Number(col)] && isActive(sortedBindings[col], activeSets)) {
```
mquander: I'm pretty sure it's considered impolite to iterate over an array with a `for...in` loop. I would write a normal `for` loop.

src/systems/userinput/userinput.js
johnshaughnessy: I share your concern that this code is not fast enough, and think that basically anything that takes any significant chunk of time can be made faster (with just a few exceptions like when we perform a raycast). The userinput system doesn't need to do a lot of work. That said, I was hesitant to start trying to remove some of these N^2 loops without first trying to measure how long the system tick was actually taking, whether the hotspots are where we think, and whether my changes actually sped the system up. For instance, this code is only run when active devices change, which I expect is very infrequent. This _can_ be made faster, but it doesn't matter to me as much as getting e.g. `computeExecutionStrategy` to be faster because it runs every time action sets change (very frequent).

To that end, I took several measurements using perf.html and focused on worst offenders / lowest hanging fruit. I *think* I managed to drop the average CPU time of the userinput system in half, from 4.6+-1% of the work done by javascript each frame to 2.3 +- 0.5%. Below I'll post my notes, the perf.html recordings, and the changes I made to help speed things up.

johnshaughnessy: First perf:
https://perfht.ml/2B2U5nE
System takes 4.6% total.
Resolve action sets: 18.694362017% : 18.9 / 101.1 
Compute execution strategy: 14.638971315% : 14.8 / 101.1

After removal of querySelectors from resolveActionSets:
https://perfht.ml/2B5dILP
System takes 4.2%
Resolve action sets        : 14.209320091% : 18.6 / 130.9
Compute execution strategy : 14.820473644% : 19.4/130.9

After replacing `for in` with `for of` and turn `masks` into an array of arrays, not a square matrix in a single array:
https://perfht.ml/2B2xSG7
System takes 2.7%
Resolve action sets        : 16.355140186% : 17.5 / 107
Compute execution strategy : 12.429906542%  : 13.3/107.0

After the next chunk of perf improvements:
https://perfht.ml/2B379JI
System takes 2.3%
Resolve action sets: 17.477477477% : 19.4/111.0
Compute execution strategy : 5.76576576577% : 6.4 / 111.0

johnshaughnessy: https://github.com/mozilla/hubs/pull/663/commits/c7eaf2ac36ba18560814f583b3f0481ec59b4ac2

johnshaughnessy: To be clear, I'm not content with the fact that the userinput system tick is still taking ~2.3% of the frame time, because I think we can get it to ~1% (general estimate, thinking that the raycasting that will definitely happen when you have a mouse takes about ~0.7% of the frametime). However, the userinput system in production now takes more like ~3.5% of the frame, I'm OK with pushing this as-is and making it faster as we move forward. There is other low-hanging perf fruit that we should be able to address when we prioritize them. For example, we could remove tracked-controls or superhands, since the event emitting that happens in both of those takes a surprisingly large amount of the CPU.

src/components/pin-networked-object-button.js
```diff
@@ -1,8 +1,5 @@
 AFRAME.registerComponent("pin-networked-object-button", {
   schema: {
-    // Selector for root of all UI that needs to be clickable when pinned
-    uiSelector: { type: "string" },
```
johnshaughnessy: Appears this was unused

src/components/pitch-yaw-rotator.js
```diff
@@ -5,8 +5,8 @@ const radToDeg = THREE.Math.radToDeg;
 
 AFRAME.registerComponent("pitch-yaw-rotator", {
   schema: {
-    minPitch: { default: -50 },
-    maxPitch: { default: 50 }
+    minPitch: { default: -65 },
+    maxPitch: { default: 65 }
```
johnshaughnessy: Unrelated to the rest of the PR, just better default values in my opinion.

src/components/stop-event-propagation.js
```diff
@@ -1,6 +1,5 @@
 const handler = e => {
   e.stopPropagation();
-  e.preventDefault();
```
johnshaughnessy: Fixes issues where when interacting with UI we end up with messed up superhands state.

johnshaughnessy:
Looks like my vive (OpenVR) controllers sometimes only have two buttons, touchpad and trigger. I think they might sometimes have even fewer (none at all?) because my buttons seem to work or not at random. 

image

PR #670, opened 2018-11-03
Firefox Reality video playback and VR entry

### LINKED ISSUE #652: No button to re-enter VR view on Mirage Solo with Firefox Reality
Device: Lenovo Mirage Solo
Browser Firefox Reality

Initial visit to a room or reload of the room url provides correct options to enter VR.
Hitting "back" exits to 2D mode with no option to re-enter VR view without completely reloading the URL to reload the room.  This makes adding any links in the room from the VR headset difficult.

brianpeiris:
Fixes #652 and #652

### DIFF
--- a/src/utils/vr-caps-detect.js
+++ b/src/utils/vr-caps-detect.js
@@ -23,8 +23,10 @@ function isMaybeDaydreamCompatibleDevice(ua) {
 const GENERIC_ENTRY_TYPE_DEVICE_BLACKLIST = [/cardboard/i];
 
 export function detectInHMD() {
-  const isOculusBrowser = /Oculus/.test(navigator.userAgent);
-  return isOculusBrowser;
+  const ua = navigator.userAgent;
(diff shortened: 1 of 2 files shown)


src/utils/vr-caps-detect.js
```diff
@@ -23,8 +23,10 @@ function isMaybeDaydreamCompatibleDevice(ua) {
 const GENERIC_ENTRY_TYPE_DEVICE_BLACKLIST = [/cardboard/i];
 
 export function detectInHMD() {
-  const isOculusBrowser = /Oculus/.test(navigator.userAgent);
-  return isOculusBrowser;
+  const ua = navigator.userAgent;
+  const isFirefoxReality = /Firefox/.test(ua) && /Android/.test(ua) && window.hasNativeWebVRImplementation;
```
gfodor: I think this ends up getting xr-polyfilled to always be `true`, oddly enough. You should confirm but I think this will regress us back to thinking firefox android is a VR capable browser. The proper check is what you have here but also for "VR" as well I think, as per the latest update for 1.1.

PR #671, opened 2018-11-05
Feature/desktop vr tlc

gfodor:
This PR makes a number of improvements to desktop VR mode:

- Stretches the canvas so only one eye buffer is visible on the screen
- Fixes the issue where the app automatically re-enters VR mode on refresh after a successful entry
- Enables the OS cursor so the desktop UI can be used
- Greatly improves the alignment of virtual hands on Oculus and adds a more forgiving box collider to the palm
- Sets the min cursor distance to a non zero value so it's clear that you need to physically grab objects and aren't grabbing them by pulling them to you
- Fixes a bug introduced by the camera rotation not being YXZ ordered that caused WASD to be wrong and the HUD to be mis-oriented. (This should eventually be replaced by the `set-rotation-yxz` component on the `feature/pinning-persistence` branch

Note that using 2d mode with VR controllers plugged in is still broken because the cursor pose path is written to twice, once by the vive/oculus control bindings and once by the keyboard/mouse bindings, regardless of how the user decides to enter.

### DIFF
--- a/src/components/controls-shape-offset.js
+++ b/src/components/controls-shape-offset.js
@@ -1,4 +1,4 @@
-import { CONTROLLER_OFFSETS } from "./hand-controls2.js";
+import { LEFT_CONTROLLER_OFFSETS, RIGHT_CONTROLLER_OFFSETS } from "./hand-controls2.js";
 
 /**
  * Sets the offset of the aframe-physics shape on this entity based on the current VR controller type
@@ -7,11 +7,12 @@ import { CONTROLLER_OFFSETS } from "./hand-controls2.js";
  */
 AFRAME.registerComponent("controls-shape-offset", {
   schema: {
-    additionalOffset: { type: "vec3", default: { x: 0, y: -0.03, z: -0.04 } }
+    additionalOffset: { type: "vec3", default: { x: 0, y: 0, z: -0.04 } }
   },
   init: function() {
     this.controller = null;
     this.shapeAdded = false;
+    this.isLeft = this.el.getAttribute("hand-controls2") === "left";
 
     this._handleControllerConnected = this._handleControllerConnected.bind(this);
     this.el.addEventListener("controllerconnected", this._handleControllerConnected);
@@ -24,8 +25,9 @@ AFRAME.registerComponent("controls-shape-offset", {
   tick: function() {
     if (!this.shapeAdded && this.controller) {
       this.shapeAdded = true;
-      const hasOffset = CONTROLLER_OFFSETS.hasOwnProperty(this.controller);
-      const offset = hasOffset ? CONTROLLER_OFFSETS[this.controller] : CONTROLLER_OFFSETS.default;
+      const offsets = this.isLeft ? LEFT_CONTROLLER_OFFSETS : RIGHT_CONTROLLER_OFFSETS;
+      const hasOffset = offsets.hasOwnProperty(this.controller);
+      const offset = hasOffset ? offsets[this.controller] : offsets.default;
       const position = new THREE.Vector3();
       const quaternion = new THREE.Quaternion();
       const scale = new THREE.Vector3();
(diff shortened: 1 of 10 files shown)


src/hub.html
```diff
@@ -26,6 +26,7 @@
         gamma-factor
         networked-scene="adapter: janus; audio: true; debug: true; connectOnLoad: false;"
         physics="gravity: -6; debug: false;"
+        tunneleffect="minRadius: 0.3"
```
johnshaughnessy: Since we authored this component, I suggest we change the default `minRadius` value to `0.3` rather than overwrite it just for this instance, now that we know we prefer the value `0.3` instead of `0.25`.

src/components/controls-shape-offset.js
```diff
@@ -34,8 +36,8 @@ AFRAME.registerComponent("controls-shape-offset", {
       quaternion.conjugate();
 
       const shape = {
-        shape: "sphere",
-        radius: "0.02",
+        shape: "box",
+        halfExtents: { x: 0.03, y: 0.04, z: 0.05 },
```
johnshaughnessy: Did a box end up feeling better than the sphere? Or is change better for perf (I thought sphere collision checks were fast)?

src/components/hand-controls2.js
```diff
@@ -12,17 +12,33 @@ const POSES = {
   mrpDown: "mrpDown"
 };
 
-// TODO: If the hands or controllers are mispositioned, then rightHand.controllerPose and rightHand.pose
-//       should be bound differently.
-export const CONTROLLER_OFFSETS = {
+export const LEFT_CONTROLLER_OFFSETS = {
```
johnshaughnessy: Nice -- these offsets have been just slightly wrong for so long. Once we get off tracked-controls, these offsets can just be set on a per-device basis in the bindings so that the correct poses are stored in the frame at the `hand` paths

src/hub.html
```diff
@@ -342,6 +343,7 @@
                 class="camera"
                 camera
                 personal-space-bubble="radius: 0.4;"
+                rotation
```
johnshaughnessy: Yikes. This is not the first time we've been burned by this. It took us a long time to correct it for the case of network-instantiated objects where we observed the RECEIVERS of an instantiation setting their euler order as YZX while the SENDER's euler was XYZ. 
I saw your message about how this regressed when I changed `pitch-yaw-rotator` so this fix is good.

src/hub.js
```diff
@@ -317,6 +317,19 @@ document.addEventListener("DOMContentLoaded", async () => {
...
   window.APP.scene = scene;
 
+  scene.addEventListener("enter-vr", () => {
+    document.body.classList.add("vr-mode");
+
+    if (!scene.is("entered")) {
+      // If VR headset is activated, refreshing page will fire vrdisplayactivate
+      // which puts A-Frame in VR mode, so exit VR mode whenever it is attempted
+      // to be entered and we haven't entered the room yet.
+      scene.exitVR();
```
johnshaughnessy: 👏

src/systems/userinput/userinput.js
```diff
@@ -73,23 +73,37 @@ AFRAME.registerSystem("userinput", {
     this.registeredMappings = new Set([keyboardDebuggingBindings]);
     this.xformStates = new Map();
 
+    let connectedGamepadBindings;
```
johnshaughnessy: Ah yes.. We were not ending up with the `registeredMappings` we wanted when e.g. touch was connected but you had not entered VR. I'll incorporate these changes to some others that I am working on for responding to active device changes correctly.

src/components/controls-shape-offset.js
gfodor: box felt better to me

gfodor: not a change for perf, but to better capture the sense that putting an object near your palm will grab it

src/hub.html
gfodor: yep, on the `pinned-persistence` branch i have a custom component to just correct this instead of using `rotation`

PR #674, opened 2018-11-08
Remove some frame time

gfodor:
Does a few optimizations on some fixed costs per frame:

- Disables a `tick` on `motion-player-capture` if we're not running in bot mode
- Avoids calling `setAttribute` on cursor components if not necessary
- Adds a missing translation

### DIFF
--- a/src/components/cursor-controller.js
+++ b/src/components/cursor-controller.js
@@ -98,7 +98,11 @@ AFRAME.registerComponent("cursor-controller", {
       const rightHandPose = userinput.get(paths.actions.rightHand.pose);
 
       this.data.cursor.object3D.visible = this.enabled && !!cursorPose;
-      this.el.setAttribute("line", "visible", this.enabled && !!rightHandPose);
+      const lineVisible = !!(this.enabled && rightHandPose);
+
+      if (this.el.getAttribute("line").visible !== lineVisible) {
+        this.el.setAttribute("line", "visible", lineVisible);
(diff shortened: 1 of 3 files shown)


src/scene-entry-manager.js
```diff
@@ -75,6 +75,8 @@ export default class SceneEntryManager {
       return;
     }
 
+    this.scene.setAttribute("motion-capture-replayer", "enabled", false);
```
johnshaughnessy: Should we remove the component entirely?

src/components/cursor-controller.js
```diff
@@ -98,7 +98,11 @@ AFRAME.registerComponent("cursor-controller", {
       const rightHandPose = userinput.get(paths.actions.rightHand.pose);
 
       this.data.cursor.object3D.visible = this.enabled && !!cursorPose;
-      this.el.setAttribute("line", "visible", this.enabled && !!rightHandPose);
+      const lineVisible = !!(this.enabled && rightHandPose);
+
+      if (this.el.getAttribute("line").visible !== lineVisible) {
+        this.el.setAttribute("line", "visible", lineVisible);
```
johnshaughnessy: Eventually we will ditch the aframe line component and draw it ourselves.

src/scene-entry-manager.js
gfodor: I believe its used for bots

PR #683, opened 2018-11-09
Feature/vr presence count

gfodor:
Adds a user count in VR, which shows the number of people in the room or lobby. It will be a bit weird until we get the full list in since it might make users think that is the number of people in the room.


gfodor:
actually I am changing this PR to include everyone. for privacy reasons you should know someone else may be listening, even if they are not in the room yet. it will be a weird UX until we get the full list in, but its better than nothing.

PR #688, opened 2018-11-09
Feature/pinning ux tweaks

gfodor:
This PR:

- Adjusts the action button colors in pause mode
- Makes it so the un-pin button is grey not red to make it more obvious which objects are pinned
- Fixes a bug where pinned videos disappear after the first person who joins the room leaves

### DIFF
--- a/src/components/remove-networked-object-button.js
+++ b/src/components/remove-networked-object-button.js
@@ -1,13 +1,7 @@
 AFRAME.registerComponent("remove-networked-object-button", {
   init() {
-    this.onClick = e => {
+    this.onClick = () => {
       if (!NAF.utils.isMine(this.targetEl) && !NAF.utils.takeOwnership(this.targetEl)) return;
-
-      // HACK currently superhands does not simulate -end events
-      // when an object is removed, so we do it here for now to ensure any
-      // super hands who have this element are cleared.
-      this.targetEl.dispatchEvent(new window.CustomEvent("hover-end", e));
-      this.targetEl.dispatchEvent(new window.CustomEvent("grab-end", e));
 
       this.targetEl.parentNode.removeChild(this.targetEl);
     };
--- a/src/components/remove-networked-object-button.js
+++ b/src/components/remove-networked-object-button.js
@@ -1,7 +1,14 @@
 AFRAME.registerComponent("remove-networked-object-button", {
   init() {
-    this.onClick = () => {
+    this.onClick = e => {
       if (!NAF.utils.isMine(this.targetEl) && !NAF.utils.takeOwnership(this.targetEl)) return;
+
+      // HACK currently superhands does not simulate a grab-end or a hover-end
+      // when an object is removed, so we do it here for now to ensure any
+      // super hands who have this element are cleared.
+      this.targetEl.dispatchEvent(new window.CustomEvent("stretch-end", e));
+      this.targetEl.dispatchEvent(new window.CustomEvent("hover-end", e));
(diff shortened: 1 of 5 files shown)


src/components/remove-networked-object-button.js
```diff
@@ -1,8 +1,14 @@
...
   init() {
-    this.onClick = () => {
+    this.onClick = e => {
       if (!NAF.utils.isMine(this.targetEl) && !NAF.utils.takeOwnership(this.targetEl)) return;
 
+      // HACK currently superhands does not simulate -end events
+      // when an object is removed, so we do it here for now to ensure any
+      // super hands who have this element are cleared.
+      this.targetEl.dispatchEvent(new window.CustomEvent("hover-end", e));
+      this.targetEl.dispatchEvent(new window.CustomEvent("grab-end", e));
```
johnshaughnessy: This looks like it tells the targeted object that it should no longer be hovered or grabbed by a superhand. Do we also need to tell the superhand that it shouldn't have any grabbers anymore?

johnshaughnessy: There's also the related issue of someone else taking ownership of the object in your superhand. Feels like both of those issues should be fixed by the same thing (and maybe could be a patch to superhands). Not sure

gfodor: What do you mean by tell the superhand it doesn't have any grabbers anymore? I thought grabbers were on the target not the superhand.

Also, this is definitely a hack -- it is specifically handling the case where the cursor is over the button and grabbing the underlying object, and then it's removed. In this case, `e` has the cursor hand, and this fires a `grab-end` event for the cursor hand.

I agree there should be a better way to deal with this. I think that superhands should properly fire the `-end` events on an object when it is removed to clear the necessary state. I looked into it and it seemed to be a non-trivial patch so I put this in for now to un-break the client (this bug is really bad since if you remove an object your cursor stops working.)

johnshaughnessy: Oops, I got confused. The superhand doesn't have grabbers. I think you're right in saying this code removes the cursor superhand from the target's grabbable's grabbers like you said (so the target won't think it's being grabbed as it gets removed from the scene). What's not clear to me is if the cursor's superhand needs some information to remove the target from its `hoverEls` or anything else. 
The bug I have in mind is the one where you can still move your cursor in and out as if there was an object attached to it after the object that was attached is removed from the scene or taken by someone else.

johnshaughnessy: I'm skeptical of this fix because I thought that the bug was not "if you remove an object your cursor stops working". I thought it was, "if you remove an object, your cursor still thinks that object is attached to it, and it won't come back until you send the drop object command (by squeezing grip on vive/oculustouch controllers)". If it's the second, it seems like it will still be broken but we won't notice on oculus/vive because we removed grab from trigger so we won't be attempting to grab the ui element in the first place. 
I get easily confused when talking thru superhands so I'm going to ping Dom and ask him to take a look at this fix. Maybe I'm mistaken about this.

gfodor: Also in my other PR moving off of trigger we still use trigger to start grabbing UI elements, since that is currently how we deal with clicks

gfodor:
This PR seems to address that issue specifically for the case where the
user clicks the remove button, but doesn't address it beyond that case -- I
agree a proper fix will involve updating superhands

gfodor:
I'd just pull the branch and test it -- on master if you remove the camera
the cursor doesn't work (because the action sets still think you are
holding the camera) but with this patch that no longer happens.

PR #689, opened 2018-11-09
Click on camera to take snapshot

gfodor:
This makes it so if you click on the camera it takes a snapshot. (Much like clicking on a video pauses it.)

This should be merged *after* we drop grab locking from camera.


gfodor:
This is wrong, closing

PR #691, opened 2018-11-10
Allow entering custom avatar urls in profile screen

netpro2k:
Just a really simple screen for setting custom avatar url on the profile screen. This can obviously be done much nicer but wanted to just get this feature in for people to play with. 

The first step to making this better would be to migrate away form avatar ids and make even the inbuilt ones just urls.

Also unrelated but simple fix to hide the avatars in the background of the picker (by setting camera far clip plane). No longer really makes sense to show them since we don't show the meeting room anymore.


src/react-components/profile-entry-panel.js
```diff
@@ -94,20 +96,46 @@ class ProfileEntryPanel extends Component {
...
+                </label>
+                <input
+                  id="custom-avatar-url"
+                  type="url"
+                  className={styles.formFieldText}
+                  value={this.state.avatarId}
+                  onChange={e => this.setState({ avatarId: e.target.value })}
+                />
+                <div className={styles.links}>
+                  <a onClick={() => this.setState({ customMode: false, avatarId: "botdefault" })}>cancel</a>
```
brianpeiris: It's a bit annoying that you reset to "botdefault" if you just wanted to take a look at what's behind the "options" link.

src/react-components/profile-entry-panel.js
```diff
@@ -94,20 +96,46 @@ class ProfileEntryPanel extends Component {
...
+                  src={`/avatar-selector.html#avatar_id=${this.state.avatarId}`}
+                  ref={ifr => (this.avatarSelector = ifr)}
+                />
+                <a
+                  className="custom-url-link"
+                  onClick={() =>
+                    this.setState({ customMode: true, avatarId: avatars.find(a => a.id === this.state.avatarId).model })
+                  }
+                >
+                  options
```
brianpeiris: I would actually prefer if this said "custom url" for now, since most people will click on an "options" link expecting to find generally useful options, whereas this url field will probably be niche.

src/react-components/profile-entry-panel.js
```diff
@@ -94,20 +96,46 @@ class ProfileEntryPanel extends Component {
...
             />
-            <div className={styles.avatarSelectorContainer}>
-              <div className="loading-panel">
-                <div className="loader-wrap">
-                  <div className="loader">
-                    <div className="loader-center" />
+            {this.state.customMode ? (
+              <div className={styles.avatarSelectorContainer}>
+                <label htmlFor="#custom-avatar-url" className={styles.title}>
+                  Avatar GLTF/GLB{" "}
```
brianpeiris: Can we maybe not make this look exactly like a `styles.title`?

PR #693, opened 2018-11-10
Remove hoveroff sounds. We "DISABLED" it but need to remove it.

johnshaughnessy:
Woops. 
Remove hoveroff sounds. We "DISABLED" it but need to remove it. #693


johnshaughnessy:
Already removed

PR #723, opened 2018-11-12
Add fly, resizing commands, and chat sounds

gfodor:
This PR:

- Adds chat commands, right now `/help` `/leave` `/fly` `'/bigger` and `/smaller` are supported. Easy to add more.
- Adds a chat sound effect
- Changes ratchet turn sound to be more subtle


src/hub.html
```diff
@@ -68,9 +68,10 @@
...
             <a-asset-item id="specialquack"                               src="./assets/sfx/specialquack.mp3"   response-type="arraybuffer" preload="auto"></a-asset-item>
+            <a-asset-item id="sound_asset-quack"                          src="./assets/sfx/quack.mp3"          response-type="arraybuffer" preload="auto"></a-asset-item>
             <a-asset-item id="sound_asset-teleport_start"                 src="./assets/sfx/teleportArc.wav"     response-type="arraybuffer" preload="auto"></a-asset-item>
             <a-asset-item id="sound_asset-teleport_end"                   src="./assets/sfx/quickTurn.wav"     response-type="arraybuffer" preload="auto"></a-asset-item>
-            <a-asset-item id="sound_asset-snap_rotate"                    src="./assets/sfx/quickTurn.wav"     response-type="arraybuffer" preload="auto"></a-asset-item>
+            <a-asset-item id="sound_asset-snap_rotate"                    src="./assets/sfx/tap_mellow.wav"     response-type="arraybuffer" preload="auto"></a-asset-item>
```
johnshaughnessy: Oh, I thought the reception of `quickTurn` was positive.

gfodor: It's still on teleport end, which feels right since it sounds like you are moving rapidly. Dom mentioned the ratchet turn as something worth refining (if I read his message in discord right) and having been play testing it a bit I tend to think it's used often enough that the current sound is a bit much -- not strongly opposed to it but I think if you try this PR it feels better. (The tap sound here is now unique since I moved to an alternative, so the ratchet turn sound is still distinct)

PR #724, opened 2018-11-12
Animation menu buttons

gfodor:
Follow-up to https://github.com/mozilla/hubs/pull/696. Animates the scaling of the freeze menus.

### DIFF
--- a/src/components/position-at-box-shape-border.js
+++ b/src/components/position-at-box-shape-border.js
@@ -69,7 +69,10 @@ AFRAME.registerComponent("position-at-box-shape-border", {
         }
       }
 
-      if (this.targetEl.getAttribute("visible") === false) return;
+      if (this.targetEl.getAttribute("visible") === false) {
+        this.prevScale = null;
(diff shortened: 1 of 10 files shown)


src/components/position-at-box-shape-border.js
```diff
@@ -122,7 +125,22 @@ AFRAME.registerComponent("position-at-box-shape-border", {
       const scale = this.halfExtents[inverseHalfExtents[targetHalfExtentStr]] * distance;
       const targetScale = Math.min(2.0, Math.max(0.5, scale * tempParentWorldScale.x));
 
-      this.target.scale.setScalar(targetScale / tempParentWorldScale.x);
+      if (!this.prevScale && Math.abs(this.prevScale - targetScale) > 0.01) {
```
brianpeiris: This logic breaks if you spawn a duck from a spawner while you're in pause mode. It also breaks the dynamic scaling behavior that currently exists on non-spawner media objects.

PR #725, opened 2018-11-12
Add camera focus and tracking

gfodor:
Adds the ability to focus and track objects on your current camera. 

Note this feature isn't going to be very useful until we fix the pause menu on avatars, which seems broken. Once fixed, we should add "focus" and "track" buttons to avatars as well.


src/components/camera-tool.js
```diff
@@ -100,12 +106,38 @@ AFRAME.registerComponent("camera-tool", {
...
     }
   },
 
+  focus(el, track) {
+    this.lookAt(el);
+
+    if (track) {
+      this.trackTarget = el;
+
+      this.trackTarget.addEventListener("componentremoved", e => {
```
gfodor: Is there a better way to do this? Trying to detect removal of the element.

src/systems/cameras.js
```diff
@@ -0,0 +1,28 @@
+// Used for tracking and managing camera tools in the scene
+AFRAME.registerSystem("cameras", {
```
netpro2k: ooh, while reading all the rest of the code I was assuming this was the system for the build in aframe `camera` component (which is also called `camera` so I am surprised this didn't break everything). We should call this `camera-tool`, it also then gets automatically bound in `camera-tool` component as `this.system`

src/systems/cameras.js
```diff
@@ -0,0 +1,28 @@
...
+    this.currentCamera = null;
+  },
+
+  deregister(el) {
+    this.cameraEls = this.cameraEls.filter(c => c !== el);
+    el.removeEventListener("ownership-changed", this._onOwnershipChange);
+    this.currentCamera = null;
+  },
+
+  getCurrent() {
```
netpro2k: What does it mean for a camera to be "current", maybe this should be `getMyCamera"

src/components/camera-tool.js
```diff
@@ -88,6 +88,12 @@ AFRAME.registerComponent("camera-tool", {
       selfieScreen.scale.set(-2, 2, 2);
       this.el.setObject3D("selfieScreen", selfieScreen);
 
+      this.cameraSystem = this.el.sceneEl.systems.cameras;
+
+      if (this.cameraSystem) {
```
netpro2k: I would ditch this defensive if, if the system doesn't exist something bad has happened, and this if would just put us in a weird state where this camera inst registered instead of complaining loudly.

src/components/camera-tool.js
netpro2k: The "correct" way is probably using a [MutationObserver](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) but in this case I would probably just check for `this.trackTarget && !this.trackTarget.parrentNode` in `tick`.

src/components/camera-tool.js
```diff
@@ -100,12 +106,38 @@ AFRAME.registerComponent("camera-tool", {
...
+      this.trackTarget.addEventListener("componentremoved", e => {
+        if (e.detail.name === "position" && e.target === this.trackTarget) {
+          this.trackTarget = null;
+        }
+      });
+    }
+  },
+
+  lookAt(el) {
+    const targetPos = new THREE.Vector3();
```
netpro2k: You are creating a new `THREE.Vector3` every tick here, I would rewrite this as:
```
lookAt: (function() {
    const targetPos = new THREE.Vector3();
    return function(el) {
        this.el.object3D.lookAt(targetPos.setFromMatrixPosition(el.object3D.matrixWorld));
    }
})(),

File: src/components/camera-focus-button.js [hunk 6]
```diff
@@ -0,0 +1,39 @@
...
+
+  init() {
+    this.cameraSystem = this.el.sceneEl.systems.cameras;
+
+    NAF.utils.getNetworkedEntity(this.el).then(networkedEl => {
+      this.targetEl = networkedEl;
+    });
+
+    this.onClick = () => {
+      if (!this.cameraSystem) return;
```
netpro2k: I would get rid of all these defensive checks for camera system existence.

src/components/camera-focus-button.js
```diff
@@ -0,0 +1,39 @@
...
+      currentCamera.components["camera-tool"].focus(this.targetEl, this.data.track);
+    };
+  },
+
+  tick() {
+    const isVisible = this.el.getAttribute("visible");
+    const shouldBeVisible = !!(this.cameraSystem && this.cameraSystem.getCurrent());
+
+    if (isVisible !== shouldBeVisible) {
+      this.el.setAttribute("visible", shouldBeVisible);
```
netpro2k: Would rewrite this to use threejs visibility directly `this.el.object3D.visible = shouldBeVisible`

PR #727, opened 2018-11-13
Fix issues detecting VR input devices

netpro2k:
This fixes a few issues which cause VR input devices not working correctly:
- [x] If `gamepadconnected` fires before the input system is initialized we may miss the event. Now queries all devices on startup to make sure
- [x] We were only keeping track of the bindings for the last gamepad that was connected. This would break things if a non-vr gamepad's `gamepadconnected` event fired after a VR device. Now turns on/off all bindings for connected VR devices.
- [x] bindings are not removed when a VR input device disconnects
- [x] if you exit VR after all calls to `getVRDisplays` in the app have been made, the Gamepad API does not report gampead state until `getVRDisplays` is called again. Now calls `getVRDisplays` every time we enter VR.

### DIFF
--- a/src/scene-entry-manager.js
+++ b/src/scene-entry-manager.js
@@ -51,6 +51,9 @@ export default class SceneEntryManager {
     }
 
     if (enterInVR) {
+      // HACK - A-Frame calls getVRDisplays at module load, we want to do it here to
+      // force gamepads to become live.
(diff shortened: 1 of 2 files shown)


gfodor:
pushed via https://github.com/mozilla/hubs/pull/728

PR #729, opened 2018-11-13
Add scale option to gltf model plus

gfodor:
Address the long standing bug where when you paste a GLTF model you see it at the original scale for one frame (which is potentially huge.) Allows setting a scale modifier on the gltf-model-plus node which will be multiplied against the root object node's scale, and sets that to a small value so the object starts out small (and then is animated to the proper scale.)


src/components/gltf-model-plus.js
```diff
@@ -136,10 +136,11 @@ const inflateEntities = function(node, templates, isRoot) {
     z: node.rotation.z * THREE.Math.RAD2DEG
   });
   el.setAttribute("scale", {
-    x: node.scale.x,
-    y: node.scale.y,
-    z: node.scale.z
+    x: node.scale.x * (scale !== undefined ? scale : 1),
```
robertlong: Should we make scale a vector rather than a scalar value so that the non-uniform scale of persisted models can be set?

robertlong: Or maybe we just don't set a scale property and use the normal transform scale to deserialize that data?

netpro2k: At first glance I thought this was going to scale all the nodes, but then i realized its just not passed through on recursion, probably ok got unexpected at first

src/components/media-loader.js
```diff
@@ -195,7 +195,8 @@ AFRAME.registerComponent("media-loader", {
         this.el.setAttribute("gltf-model-plus", {
           src: accessibleUrl,
           contentType: contentType,
-          inflate: true
+          inflate: true,
+          scale: 0.0001
```
netpro2k: So.. its a bit weird that we are going to have this scale property dangling around that actually doesnt reflect the current scale of anything. Maybe just renaming it to `initialScale` or something would help? but still feels pretty weird that we have this but then also modify the meshes scale later to reverse this.

robertlong: Yeah... to be honest, using this scale property to fix this seems like a bit of a hack. Can we do this scaling when the model is loaded but before it's added to the scene? I don't like adding more events, but I also don't really like pushing the complexity into the schema when there is already a scale component.

gfodor: we perform the initial sizing and scale in response to the `media-loaded` event, which happens > 1 frame after the gltf model has been spawned into the scene. if there's a straightforward way to resolve this issue that is cleaner than this approach then that sounds good to me, but my assumption was that this PR I wrote was the best way to do it given the way we deal with media

gfodor: i'm going to rename this to `modelToWorldScale` and push it tmw unless there are objections or suggestions on a cleaner solution

PR #738, opened 2018-11-14
Scale cursor size with player size

mquander:
This is an easy improvement to the status quo and makes the cursor look and feel reasonable at different scales, although the "shine" of the cursor on a media object still looks a little funny if you get really small.

It seems like it works fine if we just leave the physics body of the cursor the original size.

### DIFF
--- a/src/message-dispatch.js
+++ b/src/message-dispatch.js
@@ -36,6 +36,7 @@ export default class MessageDispatch {
     }
 
     const playerRig = document.querySelector("#player-rig");
+    const cursor = document.querySelector("#cursor");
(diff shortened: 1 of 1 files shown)


gfodor:
Oh, when we were chatting earlier my suggestion was to also scale the cursor based upon distance to ensure it has a fixed size in your visual field (technically steradians)

mquander:
I see, I didn't know we desired that effect -- I am not quite sure how "physical" we want the cursor to seem.

gfodor:
Here's a PR that does what I was thinking (and speeds things up too) https://github.com/mozilla/hubs/pull/757

gfodor:
ended up merging #757

PR #739, opened 2018-11-14
Sign In and Persistence Ownership

brianpeiris:
- Adds the ability to sign in to hubs
- You must now sign in to pin objects
- Pinned objects are associated with your account

image

image

Goes with https://github.com/mozilla/reticulum/pull/96


src/assets/translations.data.json
```diff
@@ -2,8 +2,15 @@
   "en": {
     "auth.verifying": "Verifying...",
     "auth.verified-title": "E-Mail Verified!",
-    "auth.verified": "Your email has been verified!",
+    "auth.verified": "Your email has been verified! {br} You can now close this browser tab and return to Hubs.",
     "auth.spoke-verified": "You email has been verified! {br} You can now close this browser tab and return to Spoke.",
+    "sign-in.prompt": "Signing in gives you the ability to pin objects in rooms.",
+    "sign-in.auth-started": "Email sent! {br} Waiting for you to click on the link in the email...",
+    "sign-in.pin": "You must sign in to pin objects.",
+    "sign-in.pin-complete": "You are now signed in!",
```
gfodor: I'd drop the `!` on this one

src/react-components/auth-dialog.js
```diff
@@ -22,7 +22,7 @@ class AuthDialog extends Component {
         ) : authOrigin === "spoke" ? (
           <FormattedMessage id="auth.spoke-verified" values={{ br: <br /> }} />
         ) : (
-          <FormattedMessage id="auth.verified" />
+          <FormattedMessage id="auth.verified" values={{ br: <br /> }} />
```
gfodor: interesting -- the approach I've been taking with this is use `white-space: pre` in the CSS. (though I don't recall if I've done it with translation tags.) any opinions on which is better?

src/react-components/ui-root.js
```diff
@@ -568,35 +628,42 @@ class UIRoot extends Component {
...
-    this.setState({
-      dialog: <CreateObjectDialog onCreate={this.createObject} onClose={this.closeDialog} />
-    });
-  }
+  showInviteTeamDialog = () => this.showDialog(InviteTeamDialog, { hubChannel: this.props.hubChannel });
 
-  showWebVRRecommendDialog() {
-    this.setState({
-      dialog: <WebVRRecommendDialog onClose={this.closeDialog} />
+  showCreateObjectDialog = () => this.showDialog(CreateObjectDialog, { onCreate: this.createObject });
```
gfodor: 👏

src/react-components/ui-root.js
```diff
@@ -1232,7 +1299,14 @@ class UIRoot extends Component {
             </div>
 
             {this.state.showPresenceList && (
-              <PresenceList presences={this.props.presences} sessionId={this.props.sessionId} />
+              <PresenceList
+                presences={this.props.presences}
+                sessionId={this.props.sessionId}
+                signedIn={this.state.signedIn}
+                email={this.props.store.state.credentials.email}
```
gfodor: could `credentials` be undefined here for existing users?

src/scene-entry-manager.js
```diff
@@ -220,12 +230,24 @@ export default class SceneEntryManager {
...
+      } else {
+        const wasInVR = this.scene.is("vr-mode");
+        if (wasInVR) this.scene.exitVR();
+        const continueTextId = wasInVR ? "entry.return-to-vr" : "dialog.close";
+
+        this.onRequestAuthentication("sign-in.pin", "sign-in.pin-complete", continueTextId, () => {
+          if (this.authChannel.authenticated) {
+            this._pinElement(e.detail.el);
+          } else {
+            // UI pins the entity optimistically, so we undo that here.
```
gfodor: it seems like this logic belongs in the pin button onclick (on grab, really) handler, and we should gate there and have no side effects if you click it and are signed in

src/utils/auth-channel.js
```diff
@@ -0,0 +1,43 @@
...
+    this.store = store;
+    this.socket = null;
+  }
+
+  setSocket = socket => {
+    this.socket = socket;
+  };
+
+  removeCredentials = () => {
+    return this.store.update({ credentials: { token: "", email: "" } });
```
gfodor: reason for `""` instead of `null`?

src/utils/link-channel.js
```diff
@@ -62,6 +62,7 @@ export default class LinkChannel {
               if (this.store.state.activity.hasChangedName) {
                 data.profile = { ...this.store.state.profile };
               }
+              data.credentials = { ...this.store.state.credentials };
```
gfodor: we probably need to bump the code to 5 or even 6 letters with this change now

gfodor: (yes non-scientific, but /shrug)

gfodor:
cut a small design/copy PR

https://github.com/mozilla/hubs/pull/741

gfodor:
specifically we should still log an error but not include any values ideally (or just the top level keys should work)

src/react-components/auth-dialog.js
brianpeiris: Ah. Using `white-space: pre` to enforce formatting feels a bit odd to me, but I think you're right, in this particular case it is probably a better solution since we control the white space in the string explicitly. I'll switch these over.

src/react-components/ui-root.js
brianpeiris: Good question. I'll check, but I think the change I made in `store.js` covers that.

src/scene-entry-manager.js
brianpeiris: Yeah, I wasn't sure about all of this since all our pinning logic is event-based and basically "fire and forget". I'll try to improve it but I may have to rework a lot of the existing code.
Also not sure what the correct way is to use "manager" code like this all the way inside an a-frame component, but I'll see if I can figure something out.

src/utils/auth-channel.js
brianpeiris: Hmm, I think the store schema validation didn't like nulls, but I'll check again.

src/utils/link-channel.js
brianpeiris: I dunno, already seems like a low enough probability for now given our traffic.

gfodor: oh, the mechanism we're missing actually that would help here is a second confirmation step on the link workflow back on the original device. that doesn't have to gate this functionality but that is relatively high priority to ensure people can't snipe credentials.  i'll add it to github

brianpeiris:
Superseded by #759. Review changes also applied to #759.

PR #740, opened 2018-11-14
Don't do meaningless stuff with empty media components

mquander:
When you have any networked-aframe schema for a component, that component will be instantiated every time a networked object with the given schema is instantiated, even if no data for it is present. In this case, that causes `media-pager` and `media-video` components to appear on all networked objects, regardless of whether the instantiator has a pager or a video. If the pager is actually on a plain image media object, it will even pick up on the `image-loaded` event and present the pager UI in some cases.

It seems like the easiest way to deal with this without changing networked-aframe's behavior is to just make no-op pager and video components behave politely.

### DIFF
--- a/src/components/media-loader.js
+++ b/src/components/media-loader.js
@@ -224,31 +224,36 @@ AFRAME.registerComponent("media-pager", {
     this.toolbar = null;
     this.onNext = this.onNext.bind(this);
     this.onPrev = this.onPrev.bind(this);
-    this.el.addEventListener("image-loaded", async e => {
-      // unfortunately, since we loaded the page image in an img tag inside media-image, we have to make a second
-      // request for the same page to read out the max-content-index header
-      this.maxIndex = await fetchMaxContentIndex(e.detail.src);
-      // if this is the first image we ever loaded, set up the UI
-      if (this.toolbar == null) {
-        const template = document.getElementById("paging-toolbar");
-        this.el.querySelector(".interactable-ui").appendChild(document.importNode(template.content, true));
-        this.toolbar = this.el.querySelector(".paging-toolbar");
-        // we have to wait a tick for the attach callbacks to get fired for the elements in a template
-        setTimeout(() => {
-          this.nextButton = this.el.querySelector(".next-button [text-button]");
-          this.prevButton = this.el.querySelector(".prev-button [text-button]");
-          this.pageLabel = this.el.querySelector(".page-label");
 
-          this.nextButton.addEventListener("grab-start", this.onNext);
-          this.prevButton.addEventListener("grab-start", this.onPrev);
+    // we kind of have to deal with the "empty case" for this component, because networked-aframe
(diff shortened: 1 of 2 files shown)


gfodor:
thank you for tracking this down!

src/components/media-loader.js
```diff
@@ -217,31 +217,36 @@ AFRAME.registerComponent("media-pager", {
...
-          this.nextButton = this.el.querySelector(".next-button [text-button]");
-          this.prevButton = this.el.querySelector(".prev-button [text-button]");
-          this.pageLabel = this.el.querySelector(".page-label");
-
-          this.nextButton.addEventListener("grab-start", this.onNext);
-          this.prevButton.addEventListener("grab-start", this.onPrev);
 
+    // we kind of have to deal with the "empty case" for this component, because networked-aframe
+    // will instantiate a null version of it for all media objects thanks to how it handles component schemas
+    if (this.data.src) {
```
johnshaughnessy: Two alternative fixes to this problem is 1) make a change to the networked component so that it doesn't call `setAttribute` for schema properties when `data` is `null` or 2) change the contract for what `el.setAttribute("foo", null)` will do in aframe.

The first suggestion (change networked-aframe) lets us do what we wanted to do with our network schemas better, because we don't really want to create a `media-pager` if no data ever comes in for it. If we wanted it to get created no matter what, we'd put `media-pager` into the template. We don't, and I think it's a mistake that we're doing it in this case. The scary thing about this suggestion to me is that if you ever purposefully sent `null` for the component data for a component you really do want to exist (but whose data needs to be null for some reason), then suddenly this change would break for you.

As for the second suggestion, it's surprising (at least to me) that calling `el.setAttribute("foo", null)` on an `el` that doesn't have a `foo` will add a `foo` to the `el`, but `el.setAttribute("foo", undefined)` will not.

What you did here is probably the fix with smallest splash zone, which is good, but it breaks in the case where you know you want a media-pager but don't fill the src in until later. For example, if we always wanted a `media-pager` on our interactable media, then we'd add `media-pager` to the template. The `media-pager` component would get created without a `src` when `networked` initialized and then setting the `src` after the image load would not cause `media-pager` to load correctly. (I tested this to confirm that it breaks.) 

Perhaps this fix will do: Change `media-pager` so that `image-loaded` needs to fire at some point AND `src` needs to be set. These can happen in any order, and when either happens we ask "has both happened yet?" If so, we go thru the rest of the work to actually create the pager controls.

netpro2k: I personally like fixing `networked-aframe` as you described, this is I think the expected behavior since right now you end up with inconsistent components on the local/remote side of a network instantiation, even without specifying a local specific template, this is a bug.

mquander: I definitely agree with these criticisms of networked-aframe, but since this fix is very harmless and that change is slightly scary, I am just going to put this fix in in the meantime because I'm sick of seeing funny arrows on images. I will add a TODO that points to this comment as an idea for future improvement.

PR #743, opened 2018-11-15
Show chat messages in HMD

gfodor:
Adds in-VR visibility into incoming chat messages.

image

(does not show spawn or presence messages atm but that should be easy to add if we want it. I thought it might be too noisy.)

There's one known rendering issue -- in Chrome on Oculus Go the text spills past the right side for long messages. I'll spend some cycles on it and update the PR if i can fix it.

edit: fixed the wrapping issue ^^


mquander:
Awkward, I already have a fix on my box to make the chat work with long usernames and messages.

I'll wait for you to merge this and if there are still any problems I'll fix them then.

gfodor:
Ah cool thanks

src/react-components/chat-message.js
```diff
@@ -60,38 +94,123 @@ export function spawnChatMessage(body) {
...
+      };
+
+      img.src = "data:image/svg+xml," + xhtml;
+    });
   });
 }
 
-export default function ChatMessage(props) {
-  const isOneLine = props.body.split("\n").length === 1;
+export async function createInWorldLogMessage({ name, type, body }) {
```
johnshaughnessy: Would like us to revisit this so that incoming messages don't render past message unreadable. I spent some time playing with the parameters of the animations configured here. I suggest at least changing the parameters such that 1) the messages disappear sooner (15 seconds is too long imo) and 2) they translate more aggressively (so that they are spread out a bit more). This will not fix the underlying problem (which is that new messages hide old ones, but is easy to do and makes it less likely to happen.

src/react-components/chat-message.js
```diff
@@ -60,38 +94,123 @@ export function spawnChatMessage(body) {
...
+        objectScale = "25%";
+        scale = 4;
+      }
+
+      messageCanvas.width = el.offsetWidth * (scale + 0.1);
+      messageCanvas.height = el.offsetHeight * (scale + 0.1);
+
+      const xhtml = encodeURIComponent(`
+          <svg xmlns="http://www.w3.org/2000/svg" width="${messageCanvas.width}" height="${messageCanvas.height}">
+            <foreignObject width="${objectScale}" height="${objectScale}" style="transform: scale(${scale});">
```
johnshaughnessy: TIL `foreignObject` is a thing you use to help you display SVGs

src/react-components/chat-message.js
```diff
@@ -60,38 +94,123 @@ export function spawnChatMessage(body) {
...
 
-export default function ChatMessage(props) {
-  const isOneLine = props.body.split("\n").length === 1;
+export async function createInWorldLogMessage({ name, type, body }) {
+  if (type !== "chat") return;
+
+  const lowResolution = AFRAME.utils.device.isMobile();
+  const blob = await renderChatMessage(body, name, false, lowResolution);
+  const entity = document.createElement("a-entity");
+  const meshEntity = document.createElement("a-entity");
```
johnshaughnessy: Could we be reusing any resources here? Not sure whether perf is sensitive enough to merit a pool of entities (or even a pool of animations, if they are reusable).

gfodor: I think that'd just reduce memory churn, the overall perf impact of this stuff is primarily in the rendering step. I'd be nervous to rig that up at this point unless it proves a real problem, since it'd be easy to get wrong. I know we've tried to avoid memory allocations in behaviors that run each tick, but these allocations only happen when a chat message comes in so I think it's going to be OK for now.

src/react-components/chat-message.js
gfodor: i sped them up, but increasing the distances they travel just strikes me as past the annoying threshold. i really think there's not going to be a way to show more than one message in your visual field without some kind of modality -- the amount of space needed (and the amount of motion needed to put on the chat messages) just for me personally isn't the right tradeoff. i think hover actions like you suggested are the way to go.

gfodor:
Def please take a UX pass on this when you have the bandwidth. It feels right to me but these things are very subjective, and also testing it manually vs actually using it with real chat is going to be different.

PR #746, opened 2018-11-17
Add queue for latency-tolerant components on media objects

gfodor:
Adds a simple queue mechanism where once a component is registered with the queue then it can request if it is its turn to `tick` on a given frame. Adds the two time-consuming, latency tolerant components on media objects to a single queue.


src/components/destroy-at-extreme-distances.js
```diff
@@ -8,9 +8,14 @@ AFRAME.registerComponent("destroy-at-extreme-distances", {
...
   },
 
+  init() {
+    this.el.sceneEl.systems["components-queue"].register(this, "media-components");
+  },
+
   tick: (function() {
     const pos = new THREE.Vector3();
     return function() {
+      if (!this.el.sceneEl.systems["components-queue"].shouldTick(this)) return;
```
netpro2k: Instead of this I would just remove the `tick` method entirely have have the `component-queue` system call a different method (either passed into `register` or just with a specific name). This way aframe never registers this component as having a behavior and avoiding the extra work and code noise of doing an early out form `tick`

src/systems/components-queue.js
```diff
@@ -0,0 +1,51 @@
+AFRAME.registerSystem("components-queue", {
+  init() {
+    // Registry a map from queue name to another map, which is a map from component object to its assigned index.
+    // Indicies range from 0 to N - 1 where N is the number of registered components in the queue.
+    // The shouldTick method returns true for the component whose index is the mod of the current frame index.
+    this.registry = new Map();
```
netpro2k: For my suggestion above this just becomes a Map from queue name to an array of components (or functions) in that queue. Then in tick you just go through each named queue and execute the function at `this.frameIndex % queue.length`

src/systems/components-queue.js
```diff
@@ -0,0 +1,51 @@
...
+      this.registry.get(queue, component).delete(component);
+      this.componentToQueue.delete(component);
+
+      let i = 0;
+
+      for (const k of components.keys()) {
+        components.set(k, i++);
+      }
+
+      component.el.removeEventListener("componentremoved", removalHandler);
```
netpro2k: I would just make this the responsibility of the component who registered to unregister themselves. This works nicely with the form of `register` that just takes a function, since then components could register themselves with multiple queues to do work without us needing to maintain an additional mapping.

src/systems/components-queue.js
```diff
@@ -0,0 +1,51 @@
...
+  init() {
+    // Registry a map from queue name to another map, which is a map from component object to its assigned index.
+    // Indicies range from 0 to N - 1 where N is the number of registered components in the queue.
+    // The shouldTick method returns true for the component whose index is the mod of the current frame index.
+    this.registry = new Map();
+    this.componentToQueue = new Map();
+    this.frameIndex = 0;
+  },
+
+  register(component, queue = component.name) {
```
netpro2k: This is nice, eventually I could see us adding different strategies for each queue or even each item such that we could do things like specify a maximum frame delay and such.

src/systems/components-queue.js
```diff
@@ -0,0 +1,51 @@
+AFRAME.registerSystem("components-queue", {
```
netpro2k: Don't love this name but I don't have better suggestions heh. `throttled-tick`, `tick-queue`, `behavior-queue`?

gfodor:

PR #749, opened 2018-11-19
Switch to non-positional audio

gfodor:
Updating the panner nodes' position/orientation every frame is expensive. For sfx that don't really need to be positional, turn it off.

### DIFF
--- a/src/hub.html
+++ b/src/hub.html
@@ -178,9 +178,9 @@
                     destroy-at-extreme-distances
                     set-yxz-order
                     pinnable
-                    sound__grab="src: #sound_asset-grab; on: grabbed; poolSize: 1;"
-                    sound__graboff ="src: #sound_asset-grab_off; on: ungrabbed; poolSize: 1;"
-                    sound__pinned ="src: #sound_asset-pinned; on: pinned; poolSize: 1;"
+                    sound__grab="src: #sound_asset-grab; on: grabbed; poolSize: 1; positional: false;"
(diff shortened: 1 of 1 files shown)


src/hub.html
```diff
@@ -453,48 +453,48 @@
                 rotation
                 pitch-yaw-rotator
                 set-yxz-order
-                sound__teleport_start="positional: false; src: #sound_asset-teleport_start; on: nothing; poolSize: 2; loop: true;"
+                sound__teleport_start="positional: false; src: #sound_asset-teleport_start; on: nothing; poolSize: 2; loop: true; positional: false;"
```
johnshaughnessy: This and the rest of the sounds on the player rig already have `position: false`.

src/hub.html
```diff
@@ -477,7 +477,7 @@
                 scene-sound__action_freeze="on: play_freeze_sound;"
                 sound__action_thaw="positional: false; src: #sound_asset-thaw; on: nothing; poolSize: 2;"
                 scene-sound__action_thaw="on: play_thaw_sound;"
-                sound__hud_action_space_bubble="positional: false; src: #sound_asset-toggle_space_bubble; on: nothing; poolSize: 2;"
+                sound__hud_action_space_bubble="positional: false; src: #sound_asset-toggle_space_bubble; on: nothing; poolSize: 2; positional: false;"
```
johnshaughnessy: ^ same as before

gfodor:

PR #750, opened 2018-11-19
Skip panner node position updates if not moved

gfodor:
This monkeypatches the `updateMatrixWorld` routine in the `PositionalAudio` node to not keep calling `setPosition`, which is expensive, unless the audio dynamics have changed.

### DIFF
--- a/src/utils/threejs-positional-audio-updatematrixworld.js
+++ b/src/utils/threejs-positional-audio-updatematrixworld.js
@@ -1,0 +1,55 @@
+// Monkeypatch to address performance issue with updating panner nodes
+// TODO this needs to be merged into 3.js
+
+THREE.PositionalAudio.prototype.updateMatrixWorld = (function() {
+  const position = new THREE.Vector3();
+  const quaternion = new THREE.Quaternion();
+
+  const scale = new THREE.Vector3();
+  const orientation = new THREE.Vector3();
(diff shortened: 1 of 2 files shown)


src/utils/threejs-positional-audio-updatematrixworld.js
```diff
@@ -0,0 +1,59 @@
...
+
+THREE.PositionalAudio.prototype.updateMatrixWorld = (function() {
+  const position = new THREE.Vector3();
+  const quaternion = new THREE.Quaternion();
+
+  const scale = new THREE.Vector3();
+  const orientation = new THREE.Vector3();
+
+  return function updateMatrixWorld(force) {
+    THREE.Object3D.prototype.updateMatrixWorld.call(this, force);
```
johnshaughnessy: Do we know why `updateMatrixWorld` is being called so often in the first place? If it's unavoidable this patch seems right. If we can avoid updating matrices when there's no change that may be better.

gfodor: `updateMatrixWorld` itself is called on every tick, and the details of how to know if something has changed (position/rot/scale, `.matrix`, or `.matrixWorld`) is not really possible to know atm

gfodor:

PR #752, opened 2018-11-19
Fix Oculus Go and Vive Focus bindings in Firefox Reality

netpro2k:
Firefox Reality names both the Vive Focus and Oculus Go controllers Gear VR Controller". This adds bindings for "Gear VR Controller" to handle those cases, as well as the actual GearVR Controller. If these get renamed later we can add new bindings to support them easily.

### DIFF
--- a/src/systems/userinput/devices/gear-vr-controller.js
+++ b/src/systems/userinput/devices/gear-vr-controller.js
@@ -1,0 +1,50 @@
+import { paths } from "../paths";
+import { Pose } from "../pose";
+
+export class GearVRControllerDevice {
+  constructor(gamepad) {
+    this.gamepad = gamepad;
+    this.buttonMap = [{ name: "touchpad", buttonId: 0 }, { name: "trigger", buttonId: 1 }];
+    this.axisMap = [{ name: "touchpadX", axisId: 0 }, { name: "touchpadY", axisId: 1 }];
+
+    this.rayObjectRotation = new THREE.Quaternion();
+    this.selector = `#player-${gamepad.hand}-controller`;
+    this.pose = new Pose();
+  }
+
+  write(frame) {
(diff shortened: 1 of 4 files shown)


src/systems/userinput/devices/gear-vr-controller.js
```diff
@@ -0,0 +1,50 @@
...
+        frame[outpath.pressed] = !!frame[paths.device.gamepad(this.gamepad.index).button(button.buttonId).pressed];
+        frame[outpath.touched] = !!frame[paths.device.gamepad(this.gamepad.index).button(button.buttonId).touched];
+        frame[outpath.value] = frame[paths.device.gamepad(this.gamepad.index).button(button.buttonId).value];
+      });
+      this.axisMap.forEach(axis => {
+        frame[paths.device.gearVRController.axis(axis.name)] =
+          frame[paths.device.gamepad(this.gamepad.index).axis(axis.axisId)];
+      });
+
+      // TODO ideally we should just be getting pose from the gamepad
```
johnshaughnessy: We can if we just add in an elbow model like in
 https://github.com/aframevr/aframe/blob/7f9c5bffeaa67a6b78b35f1401e336019914ac3b/src/components/tracked-controls.js#L130
or
https://github.com/MozillaReality/FirefoxReality/blob/b019f45785fbb4bee534811ffa38aa27509915f3/app/src/main/cpp/ElbowModel.cpp

johnshaughnessy:
In FxR: https://github.com/MozillaReality/FirefoxReality/blob/eaa4fc5920e3ac7a7dbf10204cf49f04baa52cd4/app/src/wavevr/cpp/DeviceDelegateWaveVR.cpp#L301
Wave support in other browsers to be handled later...

PR #755, opened 2018-11-20
Spoke Realtime Shadows Support

robertlong:
Scenes exported from Spoke can now cast and receive shadows.

screen shot 2018-12-17 at 3 36 48 pm

- Updates the light components with a new `shadowMapResolution` property.
- Removes the `scene-shadow` component (this never actually changed the renderer's shadow type, which is not possible at runtime)
- Sets the default shadow type to `pcfsoft`.
- Upgrades AFrame to the latest `mozillareality/hubs/dev` branch https://github.com/MozillaReality/aframe/tree/hubs/dev
  - This branch adds the `enabled` prop to the `shadow` system. This is used to disable shadow maps on low quality.

To Do:
- [x] Disable shadows on mobile.
- [x] Fix bug where alt-materials models are not rendered in low quality.


robertlong:
I still need to update our AFrame dev branch before merging this change.

src/utils/resizeShadowCameraFrustum.js
```diff
@@ -0,0 +1,76 @@
...
+  new THREE.Vector3(),
+  new THREE.Vector3(),
+  new THREE.Vector3(),
+  new THREE.Vector3(),
+  new THREE.Vector3(),
+  new THREE.Vector3(),
+  new THREE.Vector3()
+];
+
+export default function resizeShadowCameraFrustum(light, scene) {
```
netpro2k: Hmm, seems a bit scary to be doing this every frame for each (probably only one) directional light. Its very likely the environment is going to be the largest thing, so the frustum actually wont end up changing anyway... Maybe just do it on environment load, or when objects are added/removed from the scene... Both of these could be a bit buggy for a really small environment with objects moving outside its bounds.. hmm

robertlong: I can just disable it on mobile for now and we can tune it as we learn more about our shadow requirements.

PR #759, opened 2018-11-27
Sign In, persistence ownership and upload promotion

brianpeiris:
Uploads can now be pinned permanently. Previously, pinned uploads would only live for 48 hours. As a side effect, you'll need to sign in to Hubs in order to pin objects.

image

image

- Adds the ability to sign in to hubs
- You must now sign in to pin objects
- Pinned objects are associated with your account
- Media uploads are now associated with your account and persist permanently
- Media uploads now return a `promotion_token` which is stored in local storage.
- `media-loader` component now includes a `fileId` property for uploaded media.
- `promotion_token` is used to promote pinned uploads.
- Remote users are only allowed to pin/un-pin your uploads after you have pinned them once (i.e. claimed ownership).

Goes with https://github.com/mozilla/reticulum/pull/97

Relies on https://github.com/MozillaReality/networked-aframe/pull/20

Supersedes #739


src/components/super-networked-interactable.js
```diff
@@ -71,7 +71,6 @@ AFRAME.registerComponent("super-networked-interactable", {
         this.el.setAttribute("body", { type: "dynamic" });
         this._syncCounterRegistration();
       } else {
-        this.el.emit("grab-end", { hand: this.hand });
```
gfodor: mind explaining this change?

src/scene-entry-manager.js
```diff
@@ -202,11 +202,23 @@ export default class SceneEntryManager {
   };
 
   _pinElement = el => {
-    const networkId = el.components.networked.data.networkId;
+    const { networkId } = el.components.networked.data;
+
+    const { fileId, src } = el.components["media-loader"].data;
+
+    let fileToken, promotionToken;
```
gfodor: I think the term we've been using is `access_token`, perhaps we should update the query string arg as well

src/storage/store.js
```diff
@@ -47,6 +47,14 @@ export const SCHEMA = {
       properties: {
         lastUsedMicDeviceId: { type: "string" }
       }
+    },
+
+    uploadPromotionTokens: {
+      type: "array",
+      items: {
+        fileId: { type: "string" },
+        promotionTokens: { type: "string" }
```
gfodor: typo here (trailing "s") -- surprised you aren't getting field validation errors -- should probably confirm this is doing what we expect

src/storage/store.js
```diff
@@ -99,10 +109,10 @@ export default class Store extends EventTarget {
 
   update(newState) {
     const finalState = merge(this.state, newState);
-    const isValid = validator.validate(finalState, SCHEMA).valid;
+    const result = validator.validate(finalState, SCHEMA);
 
-    if (!isValid) {
-      throw new Error(`Write of ${JSON.stringify(finalState)} to store failed schema validation.`);
+    if (!result.valid) {
+      throw new Error(`Write of ${JSON.stringify(finalState)} to store failed schema validation. ${result}`);
```
gfodor: oh, please remove this `stringify` while you're here. it's a privacy issue

src/utils/hub-channel.js
```diff
@@ -104,17 +104,30 @@ export default class HubChannel {
     this.channel.push("message", { body, type });
   };
 
-  pin = (id, gltfNode) => {
+  pin = (id, gltfNode, fileId, fileToken, promotionToken) => {
```
gfodor: `accessToken`, etc

src/scene-entry-manager.js
```diff
@@ -230,6 +242,7 @@ export default class SceneEntryManager {
     });
 
     this.scene.addEventListener("pinned", e => {
+      if (!NAF.utils.isMine(e.detail.el)) return;
```
gfodor: what case causes this to need to be here? when you click the pin button, you should take ownership -- but maybe that's not working right?

gfodor:
looking good, one thing missing I think is hiding the pin button on uploaded files you do not have the promotion token for. though, to do that properly we would need to also ensure the pin button is visible on an object that has its file already promoted, which I don't think we have a way to know atm.

gfodor:
the best approach I can come up with for this:

- introduce a locally-maintained, non-networked flag `fileIsOwned` on media objects that is:
  - `false` when you spawn a new object (uploaded or URL) or see another user spawn an object
  - is flipped to `true` whenever you see an object with a file id become pinned (since it's assumed the underlying file has become owned by that point)
  - is flipped to `true` by default in the GLTF deserializer for media with file ids

and we show the pin button for objects you either have the promotion token for or whose `fileIsOwned` is true

gfodor:
ofc, we could promote the `fileIsOwned` flag to a fully networked piece of state as well. on the fence on that one -- it seems more semantically correct but perhaps bad to include this kind of incidental state in our persistent data structure/GLTF schema etc.

src/components/super-networked-interactable.js
brianpeiris: These changes uncovered an existing issue with pinning, where grab-end would be fired when a remote user took over ownership of a pinned entity. It would cause all other users to re-pin an entity before the new owner un-pinned it. This was especially a problem now that pinning is gated.

Removing this emit was a bad solution though. I've gone with a different solution with Kevin's advice: https://github.com/MozillaReality/networked-aframe/pull/20

src/components/media-loader.js
```diff
@@ -23,6 +23,7 @@ const boundingBox = new THREE.Box3();
 
 AFRAME.registerComponent("media-loader", {
   schema: {
+    fileId: { type: "string" },
```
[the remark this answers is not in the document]
brianpeiris: I don't think we should serialize `fileId`, it's really only needed by the uploader in conjunction with the promotion token.

PR #764, opened 2018-11-28
Set matrix auto update to default in gltf model plus

gfodor:
Fixes issue where gltf-model-plus is flipping auto matrix update off in three.js scenes where the default was not changed.

### DIFF
--- a/src/components/gltf-model-plus.js
+++ b/src/components/gltf-model-plus.js
@@ -258,7 +258,7 @@ async function loadGLTF(src, contentType, preferredTechnique, onProgress) {
   const envMap = await CachedEnvMapTexture;
 
   gltf.scene.traverse(object => {
-    object.matrixAutoUpdate = false;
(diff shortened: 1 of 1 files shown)


src/components/gltf-model-plus.js
```diff
@@ -258,7 +258,7 @@ async function loadGLTF(src, contentType, preferredTechnique, onProgress) {
   const envMap = await CachedEnvMapTexture;
 
   gltf.scene.traverse(object => {
-    object.matrixAutoUpdate = false;
+    object.matrixAutoUpdate = THREE.Object3D.DefaultMatrixAutoUpdate;
```
netpro2k: Was confused why we needed this line at all. Suggest adding a comment:
```suggestion
    // GLTFLoader sets matrixAutoUpdate on animated objects, we want to keep the defaults
    object.matrixAutoUpdate = THREE.Object3D.DefaultMatrixAutoUpdate;
```

PR #765, opened 2018-11-28
Fix bug where pagination causes scaling

gfodor:
The recent animation changes introduced a bug where pagination on PDFs would scale the object down by 50%. This was due to the preamble to the spawn animation being re-run. This PR skips the spawn animation if it's already run on the object.

### DIFF
--- a/src/utils/media-utils.js
+++ b/src/utils/media-utils.js
@@ -119,7 +119,7 @@ export const addMedia = (src, template, contentOrigin, resolve = false, resize =
     entity.addEventListener(eventName, () => {
       clearTimeout(fireLoadingTimeout);
 
-      if (!entity.classList.contains("pen")) {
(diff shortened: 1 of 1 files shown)


src/utils/media-utils.js
```diff
@@ -119,7 +119,7 @@ export const addMedia = (src, template, contentOrigin, resolve = false, resize =
     entity.addEventListener(eventName, () => {
       clearTimeout(fireLoadingTimeout);
 
-      if (!entity.classList.contains("pen")) {
+      if (!entity.classList.contains("pen") && !entity.getAttribute("animation__spawn-start")) {
```
netpro2k: Ah interesting, I expected that the animation component would remove itself. I guess it doesn't so this makes sense. Wonder if pagination should have its own loading animation at some point

PR #768, opened 2018-11-29
Fix distance check when not visible

gfodor:
This PR fixes an issue where the distance check is not working in the visible while frozen component since we no longer update world matrices every frame for invisible objects.

### DIFF
--- a/src/components/visible-while-frozen.js
+++ b/src/components/visible-while-frozen.js
@@ -42,8 +42,8 @@ AFRAME.registerComponent("visible-while-frozen", {
       getLastWorldPosition(this.cam, this.camWorldPos);
       this.objWorldPos.copy(this.el.object3D.position);
 
-      if (!isVisible && isFrozen) {
-        // Edge case, if the object is not visible, and the scene isFrozen, force a
+      if (!isVisible) {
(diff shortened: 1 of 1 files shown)


src/components/visible-while-frozen.js
```diff
@@ -36,10 +36,18 @@ AFRAME.registerComponent("visible-while-frozen", {
     const isFrozen = this.el.sceneEl.is("frozen");
 
     let isWithinDistance = true;
+    const isVisible = this.el.getAttribute("visible");
```
netpro2k: We try not to use these fake components anymore:
```suggestion
    const isVisible = this.el.object3D.visible;
```

src/components/visible-while-frozen.js
```diff
@@ -36,10 +36,18 @@ AFRAME.registerComponent("visible-while-frozen", {
...
 
     if (this.data.withinDistance !== undefined) {
       getLastWorldPosition(this.cam, this.camWorldPos);
       this.objWorldPos.copy(this.el.object3D.position);
+
+      if (!isVisible) {
```
netpro2k: Would probably move this above the `this.objWorldPos.copy` just so it reads a bit more clearly.

PR #770, opened 2018-11-30
Do not intersect non-visible objects

gfodor:
Setting the in-world HUD results in its matrix no longer updating, which results in it being stuck in the floor, which (until this PR) results in being able to click on it and mute yourself, etc. This PR filters out non-visible objects from the raycast intersection set so the cursor no longer interacts with non-visible objects. (Is this a bad idea?)


netpro2k:
In the past I think we have just removed the `interactable` or `ui` class from objects we did not want to raycast against. I think I like that better than having to walk the tree for all raycasts

PR #771, opened 2018-11-30
Add undo functionality for drawings

InfiniteLee:
Adds functionality and input bindings for drawings to allow "undoing" for drawn lines.

TODO:

- [x] Drawings are broken when a user enters after lines have been removed from the drawing.

d9245c4586db65d2d8179c0ba33cba60

Current bindings (while holding the pen):
Oculus Go: "dpadNorth"
Oculus Rift: "B" or "Y" button, on associated controller
Vive: "Top" button on associated controller
Keyboard: "U" key
Xbox controller: "Y" button


src/systems/userinput/bindings/vive-user.js
```diff
@@ -623,13 +623,13 @@ export const viveUserBindings = addSetsToBindings({
...
       dest: { value: paths.actions.leftHand.startDrawing },
       xform: xforms.rising,
-      priorty: 2
+      priority: 2
     },
     {
       src: { value: leftTriggerPressed2 },
       dest: { value: lTriggerFallingStopDrawing },
       xform: xforms.falling,
-      priorty: 2
```
johnshaughnessy: yikes, good catch

src/systems/userinput/bindings/oculus-go-user.js
```diff
@@ -76,10 +76,10 @@ export default function generate3DOFTriggerBindings(device) {
           west: dpadWest,
           center: dpadCenter
         },
-        xform: xforms.vec2dpad(0.8)
+        xform: xforms.vec2dpad(0.3)
       },
       {
-        src: [dpadNorth, dpadSouth, dpadCenter],
```
johnshaughnessy: Hmm, this also affects freeze/thaw... does the binding using `dpadNorth` (below) mask this minding? If so, I think you don't need to remove `dpadNorth` and `dpadCenter` and freeze/thaw will continue to work as it does now when not holding the pen.

PR #776, opened 2018-12-07
Fix video playback and fullscreen bugs on iOS Safari

brianpeiris:
It seems Safari 12 has tightened its auto-play policies on iOS. This fix allows video playback with an user gesture to unmute videos. Safari 12 also has new fullscreen behavior that was breaking our movement and interaction controls when pinching and dragging. Safari does not allow preventing the native fullscreen behavior, so I've disabled fullscreen mode as well.
 
- Sets `autoplay` and `muted` properties on videos
- Adds a UI button on videos to indicate that a user gesture is necessary to un-mute them. The un-mute happens on `touchstart` since `grab-start` is not recognized as a user-initiated gesture. The button is hidden in freeze mode
- Disables fullscreen mode on iOS.
- Minor refactor of a-frame UI HTML and components
- Adds a `debug_log` flag that shows an on-screen console and log which is useful on mobile devices


gfodor:
Thanks so much for fixing all these iOS issues! Will review ASAP

robertlong:
Looks pretty good. I did run into a bug. I'm not sure what the cause is yet.

My iPhone did not load a video that I added on my laptop:
img_1218

When adding the video on my phone it loaded just fine.
img_1219

robertlong:
This time I added the video on my laptop before joining on my phone. It loaded but did not play automatically. I think that's a limitation of iOS though.
img_53fb3efc6dfc-1

robertlong:
I reloaded and it worked. I'll try to find a reproducible bug.

brianpeiris:
I'll chalk some of the video loading bugs up to dev farspark being slow maybe, but I'll continue testing Safari/iOS once this is out. Thanks for the review.

PR #780, opened 2018-12-11
Media Loader Refactor + Spoke Image/Video Node Support

robertlong:


netpro2k:
We went with a less invasive solution to support spoke. We may revisit a refactor like this in the future.

PR #783, opened 2018-12-12
Add mediaOptions property to media-loader and video/image inflators

robertlong:
Media exported from Spoke can now be imported into Hubs. Spoke also includes some additional options for media like spatial audio options for videos.

screen shot 2018-12-17 at 1 56 55 pm

- Adds the `mediaOptions` property to `media-loader` properties on this object will be set on the media component on load.
- Adds the `static-media` and `static-controlled-media` network templates.
  - `static-controlled-media` is non-interactable (no physics) but can be grabbed to toggle play/pause on a video.
  - `static-media` has no grabbable behavior or cursor highlight effects.
- Adds the `image` and `video` gltf component inflators.
  - The `video` inflator accepts two properties:
    - `controls` determines what template to use. `image` always uses `static-media`.
    - `autoPlay` determines whether the first network owner sets `videoPaused` to false/true.
- Adds a number of new properties to the `media-video` component:
    - volume
    - loop
    - audioType (Currently accepted values are "stereo" or "pannernode")
    - distanceModel
    - rolloffFactor
    - refDistance
    - maxDistance
    - coneInnerAngle
    - coneOuterAngle
    - coneOuterGain
    - See https://developer.mozilla.org/en-US/docs/Web/API/PannerNode for more info on these properties.

### DIFF
--- a/src/components/gltf-model-plus.js
+++ b/src/components/gltf-model-plus.js
@@ -168,7 +168,7 @@ const inflateEntities = function(node, templates, isRoot, modelToWorldScale) {
     for (const prop in entityComponents) {
       if (entityComponents.hasOwnProperty(prop) && AFRAME.GLTFModelPlus.components.hasOwnProperty(prop)) {
         const { componentName, inflator } = AFRAME.GLTFModelPlus.components[prop];
-        inflator(el, componentName, entityComponents[prop]);
+        inflator(el, componentName, entityComponents[prop], entityComponents);
       }
     }
   }
--- a/src/components/media-loader.js
+++ b/src/components/media-loader.js
@@ -31,7 +31,7 @@ AFRAME.registerComponent("media-loader", {
     contentType: { default: null },
     mediaOptions: {
       default: {},
-      parse: v => (typeof v === "object" ? JSON.parse(v) : v),
+      parse: v => (typeof v === "object" ? v : JSON.parse(v)),
       stringify: JSON.stringify
     }
   },
--- a/src/components/media-loader.js
+++ b/src/components/media-loader.js
@@ -28,7 +28,18 @@ AFRAME.registerComponent("media-loader", {
     src: { type: "string" },
     resize: { default: false },
     resolve: { default: false },
-    contentType: { default: null }
+    contentType: { default: null },
+    mediaOptions: {
+      default: {},
+      parse: value => {
+        if (typeof value === "object") {
+          return value;
+        }
+
+        return JSON.parse(value);
+      },
+      stringify: JSON.stringify
+    }
   },
 
   init() {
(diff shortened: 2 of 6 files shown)


src/components/gltf-model-plus.js
```diff
@@ -168,7 +168,7 @@ const inflateEntities = function(node, templates, isRoot, modelToWorldScale) {
     for (const prop in entityComponents) {
       if (entityComponents.hasOwnProperty(prop) && AFRAME.GLTFModelPlus.components.hasOwnProperty(prop)) {
         const { componentName, inflator } = AFRAME.GLTFModelPlus.components[prop];
-        inflator(el, componentName, entityComponents[prop]);
+        inflator(el, componentName, entityComponents[prop], entityComponents);
```
netpro2k: Still would prefer we not pass an element of a collection AND the whole collection into a function like this

src/components/media-loader.js
```diff
@@ -28,7 +28,18 @@ AFRAME.registerComponent("media-loader", {
     src: { type: "string" },
     resize: { default: false },
     resolve: { default: false },
-    contentType: { default: null }
+    contentType: { default: null },
+    mediaOptions: {
+      default: {},
+      parse: value => {
+        if (typeof value === "object") {
```
netpro2k: super minor but I would write:
```
parse: v => (typeof v === "object" ? JSON.parse(v) : v)
```

src/components/media-views.js
```diff
@@ -336,12 +345,29 @@ AFRAME.registerComponent("media-video", {
         // TODO FF error here if binding mediastream: The captured HTMLMediaElement is playing a MediaStream. Applying volume or mute status is not currently supported -- not an issue since we have no audio atm in shared video.
         texture.audioSource = this.el.sceneEl.audioListener.context.createMediaElementSource(texture.image);
 
-        const sound = new THREE.PositionalAudio(this.el.sceneEl.audioListener);
-        sound.setNodeSource(texture.audioSource);
-        this.el.setObject3D("sound", sound);
+        let audio;
+
+        if (this.data.audioType === "pannernode") {
```
netpro2k: Hmm, does the fact that its a PannerNode actually matter here? I feel like it should just be "positional" or "spatialized"

src/hub.html
```diff
@@ -162,6 +162,34 @@
...
+                <a-entity
+                    class="interactable"
+                    body="type: static; shape: none; mass: 1;"
+                    auto-scale-cannon-physics-body
+                    set-sounds-invisible
+                >
+                </a-entity>
+            </template>
+
+            <template id="static-controlled-media">
```
netpro2k: Confused as to how this actually ends up being not movable?

src/components/media-views.js
robertlong: Aren't these properties just one model of spatial audio though? Like Google's resonance API has some additional/different properties but could fall back to the panner node: https://developers.google.com/resonance-audio/reference/web/Source#~SourceOptions

src/hub.html
robertlong: I made the body static and the network schema does not sync transform properties.

src/components/gltf-model-plus.js
robertlong: Our inflator inflates multiple components that are dependent on one another `networked` and `media-loader`. In glTF I also represent these as two components `image`/`video` and `networked` I use the id stored in the networked component when adding the `networked` component to the entity. I somewhat agree that it isn't ideal that these cant be separate concerns. However, I do not know of a better alternative.

src/components/media-loader.js
robertlong: In this case I think that's fine because it is short and readable. I generally avoid ternary operators when the cases include something other than a value (In this case, `JSON.parse`). They can be hard to read and minification will reduce the size of the function anyway.

src/components/gltf-model-plus.js
netpro2k: I just meant since we are passing in the collection we don't need to pass in the single item, just pass in the collection:
```
inflator(el, componentName, entityComponents); 
```

src/components/media-loader.js
netpro2k: It's not about being less code, its about avoiding multiple codepaths and making as many things possible an expression.

src/components/media-views.js
netpro2k: I was thinking of them as pretty generic, but I guess its true that not all spatialization is going to have properties that look like this.

PR #784, opened 2018-12-13
Add 360 Images and Videos

robertlong:
You can now export scenes with 360 Videos from Spoke.

360video

Adds the projection field to `media-image` and `media-video`. We currently support `flat` (default) and `equirectangular`. Spoke exports video and image nodes with this property.


robertlong:
I'll fix those up. Thanks for the review!

PR #789, opened 2018-12-17
Set matrixNeedsUpdate after loading glTF model (WIP)

robertlong:
This fix ensures that objects in the scene have their matrices initialized on load. I'm still investigating why the old way didn't work sometimes and this does.

This bug can be reproduced with this scene:
https://localhost:8080/scene.html?scene_id=uw4R5Jz

[ShadowExample.spoke.zip](https://github.com/mozilla/hubs/files/2688255/ShadowExample.spoke.zip)


netpro2k:
This is likely related to https://github.com/mozilla/hubs/pull/801

We originally had to use the `setAttribute` calls here for some reason, though that may have changed. If it has we may be able to avoid the added weirdness.

PR #791, opened 2018-12-18
Fix revoking SketchfabZipWorker's Object URLs (WIP)

robertlong:
Currently `revokeObjectURL` is being called with the original file name not the blob URL. However, based on the `chrome://blob-internals` tool it looks like these object URLs are still not being freed.

1. Load Atrium Scene
screen shot 2018-12-17 at 5 32 41 pm

2. Import [Sketchfab model](https://sketchfab.com/models/c51ca76e5c3e4108a3de3f9cfdd48d4e)
screen shot 2018-12-17 at 5 32 52 pm

3. Remove Sketchfab model
screen shot 2018-12-17 at 5 33 04 pm

I believe there should not be any blob being held onto at step 2. Instead all of the blobs created by the SketchfabZipLoader are never freed.

### DIFF
--- a/src/components/gltf-model-plus.js
+++ b/src/components/gltf-model-plus.js
@@ -275,7 +275,7 @@ async function loadGLTF(src, contentType, preferredTechnique, onProgress) {
 
   if (fileMap) {
     // The GLTF is now cached as a THREE object, we can get rid of the original blobs
-    Object.keys(fileMap).forEach(URL.revokeObjectURL);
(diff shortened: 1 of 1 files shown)


netpro2k:
Well actually the zip blobs should be immediately not held onto once the model is done loading. I would expect some blobs (mainly textures) to stick around though since we are keeping a copy of the loaded gltf around in memory.

robertlong:
I'm gonna close this since it's not the correct solution and it's pretty stale.

PR #798, opened 2018-12-20
Support Windows Mixed Reality Controllers

brianpeiris:
Re-enable support for Windows Mixed Reality with bindings in our user input system.

Also:
- `log-events` Debug helper that you can add to any element to log all of its events
- `userinput-debug` Will now show a live status of the current frame's actions when you hit the `l` key.
- New `all` and `deadzone` xforms


src/systems/userinput/paths.js
```diff
@@ -227,3 +227,27 @@ paths.device.vive.right = {
   },
   pose: `/device/vive/right/pose`
 };
+
+paths.device.wmr = {};
```
netpro2k: I think I like how I redid the paths for Oculus Go better, the generic functions kind of lose some of the benefits of making these variables in the first place.

src/systems/userinput/devices/windows-mixed-reality-controller.js
```diff
@@ -0,0 +1,62 @@
+import { paths } from "../paths";
+import { Pose } from "../pose";
+
+export class WindowsMixedRealityControllerDevice {
```
netpro2k: This looks fine and is how most of our controllers are done. I rewrote Oculus Go in a more straightforward way which you might consider doing here as well. Haven't 100% decided if its worth changing all of them.

src/systems/userinput/userinput-debug.js
```diff
@@ -3,11 +3,44 @@ AFRAME.registerSystem("userinput-debug", {
...
+        if (!["/wmr-user", "/actions/"].some(x => key.startsWith(x))) continue;
+        let val = JSON.stringify(frame[key], replacer);
+        if (val) val = val.replace(/{"(\w{3,})":/g, '{\n  "$1":').replace(/,"(\w{3,})":/g, ',\n  "$1":');
+        pathsOutput.push(`${key} -> ${val}`);
+      }
+      this.userinputFrameStatus.textContent = pathsOutput.join("\n");
+    }
+
     if (userinput.get(paths.actions.logDebugFrame)) {
+      if (!this.userinputFrameStatus) {
```
netpro2k: Nifty. I did a little experiment with rstats (the thing aframe uses for perf display) to draw graphs of all the values over time. Was pretty cool, but needed work to be useful. This is a good start. I really like how debuggable this system should be.

src/systems/userinput/userinput-debug.js
```diff
@@ -3,11 +3,44 @@ AFRAME.registerSystem("userinput-debug", {
...
+        if (typeof v === "number") {
+          return `${v >= 0 ? "+" : ""}${v.toFixed(3)}`;
+        }
+        return v;
+      };
+      const pathsOutput = [];
+      const { frame } = userinput;
+      for (const key in frame) {
+        if (!frame.hasOwnProperty(key)) continue;
+        if (!["/wmr-user", "/actions/"].some(x => key.startsWith(x))) continue;
```
netpro2k: Did you mean to leave `/wmr-user` in here? Seems like it should only be actions by default.

src/systems/userinput/bindings/windows-mixed-reality-user.js
```diff
@@ -0,0 +1,498 @@
+import { paths } from "../paths";
+import { sets } from "../sets";
+import { xforms } from "./xforms";
+import { addSetsToBindings } from "./utils";
+
+const v = name => {
+  return `/wmr-user/wmr-var/${name}`;
```
netpro2k: We have typically been prefixing this with `/vars`. Fine for the paths to be whatever after that but would prefer not adding more roots. Doesn't really matter much now, but seems cleaner to keep the root namespace clean.

src/systems/userinput/bindings/windows-mixed-reality-user.js
```diff
@@ -0,0 +1,498 @@
...
+      xform: xforms.copy
+    },
+    ...handPoseBindings("left"),
+    ...handPoseBindings("right"),
+    {
+      src: { value: paths.device.keyboard.key(" ") },
+      dest: { value: paths.actions.thaw },
+      xform: xforms.falling
+    },
+    ...freezeBindings("left"),
```
netpro2k: I do like how this ends up reading with the utility functions.

src/systems/userinput/bindings/windows-mixed-reality-user.js
```diff
@@ -0,0 +1,498 @@
...
+  [sets.cursorHoveringOnCamera]: [],
+  [sets.rightHandHoveringOnCamera]: [],
+
+  [sets.leftHandHoldingCamera]: holdingCameraBindings("left"),
+  [sets.cursorHoldingCamera]: holdingCameraBindings("right", true),
+  [sets.rightHandHoldingCamera]: holdingCameraBindings("right"),
+
+  [sets.inputFocused]: [
+    {
+      src: { value: "/device/keyboard" },
```
netpro2k: This was one of the motivating factors behind why we created the priority system based on src hierarchy.

src/systems/userinput/devices/windows-mixed-reality-controller.js
```diff
@@ -0,0 +1,62 @@
...
+      { name: "touchpad", buttonId: 0 },
+      { name: "trigger", buttonId: 1 },
+      { name: "grip", buttonId: 2 },
+      { name: "menu", buttonId: 3 }
+    ];
+
+    this.gamepad = gamepad;
+    this.pose = new Pose();
+    this.axisMap = [
+      { name: "padX", axisId: 0 },
```
netpro2k: I was going to suggest you call these `touchpadX` and `joystickX` for consistency, but looking at our other devices it seems like we did it both ways... I prefer using the same name as the corresponding button, but since we already have inconsistency, its your choice.

PR #803, opened 2018-12-27
Use dynamic janus host and failover

gfodor:
Depends on https://github.com/mozilla/reticulum/pull/102

Adds dynamic janus hosts and reconnect failover to a new janus host if it is re-assigned.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ npm run build
 
 and then upload the files in the `dist` folder to your hosting provider.
 
-If you are running your own servers, you can modify the environment variables `JANUS_SERVER` and
+If you are running your own servers, you can modify the environment variable
 `RETICULUM_SERVER` when building to point Hubs to your own infrastructure.
 
 See `.env.defaults` for the full set of environment variables that can modify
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -40,7 +40,7 @@ npm run build
 
 and then upload the files in the `dist` folder to your hosting provider.
 
-If you are running your own servers, you can modify the environment variables `JANUS_SERVER` and
+If you are running your own servers, you can modify the environment variable
 `RETICULUM_SERVER` when building to point Hubs to your own infrastructure.
 
 See `.env.defaults` for the full set of environment variables that can modify
```
brianpeiris: Should probably remove `JANUS_SERVER` from `.env.defaults` as well.

src/hub.js
```diff
@@ -286,15 +287,41 @@ async function handleHubChannelJoined(entryManager, hubChannel, messageDispatch,
...
       .connect()
       .then(() => {
+        // When reconnecting, update the server URL if necessary
+        NAF.connection.adapter.setReconnectionListeners(
+          () => {
+            const step = async () => {
+              if (!NAF.connection.adapter.isDisconnected()) return;
+
+              const currentServerURL = NAF.connection.adapter.serverUrl;
+              const newHubHost = await hubChannel.getHost();
```
brianpeiris: Hmm, I think this is impossible, but we may `await getHost` forever since `getHost` never rejects its promise.

gfodor: ah good point, i'll just thread the reject through on error at least

PR #806, opened 2018-12-28
Fix mic permissions handling on apple devices

brianpeiris:
- Restricts the hack introduced in #790 to Safari on iOS and macOS so that Firefox and Chrome aren't blocked
- Adds a info dialog insisting on mic permissions in Safari
- Also fixes a separate bug that broke hubs in Chrome on iOS


gfodor:
LGTM

PR #817, opened 2019-01-03
Video playback controls

netpro2k:
Adds controls for scrubbing videos, bindings for adjusting volume, and support for specifying a start time via url.

image

Start time is specified via `t=N` where N is the start time in seconds. This paramter can be specified as either a querystring or hash, which lines up with youtube's url structure:  https://youtu.be/vEvlWyb29ik?t=21 or https://www.youtube.com/watch?v=KAHKP313P2I#t=100

Followup pass will handle doing some sort of overlay to make the pause menu less cluttered and also hopefully show volume state in some way.

todo:
- Test volume bindings on:
  - [x] Mouse
  - [x] Oculus Touch
  - [x] Oculus Go
  - [x] Vive
  - [ ] ~Daydream~
  - [x] Windows MR (bindings need to be added first)


gfodor:
probably feature creep, but if we're rolling out this as the UX for volume control, may want to see how hard it is to add to avatars too

gfodor:
also obvious but some kind of visual indicator of the current volume seems like a good idea to follow up with (maybe just a speaker icon that appears with a few "sound waves" that expand on and off as you change volume up/down) -- i think if we had visual feedback and (eventually) educate users on how it works that'd probably be livable for a while vs spending time on some kind of slider UI

src/components/media-views.js
```diff
@@ -286,12 +288,34 @@ AFRAME.registerComponent("media-video", {
...
     this.el.removeEventListener("grab-start", this._grabStart);
     this.el.removeEventListener("grab-end", this._grabEnd);
+    this.seekForwardButton.removeEventListener("grab-start", this.seekForward);
+    this.seekBackButton.removeEventListener("grab-start", this.seekBack);
+    this.seekForwardButton.object3D.visible = false;
+    this.seekBackButton.object3D.visible = false;
+  },
+
+  seekForward() {
+    if (NAF.utils.isMine(this.networkedEl) || NAF.utils.takeOwnership(this.networkedEl)) {
```
robertlong: How does this work when you aren't the owner? takeOwnership is async right? I guess you sync the currentTime property so the other users will get that update?

netpro2k: takeOwnership is not actually async, you know if you are going to become the owner based on the last owner timestamp. You of course may lose ownership if someone else has a message in flight that has a newer owner time than yours, but that acts just like any other form of losing ownership. The only time it would ever return false is if you already own it (we already check for that) or if you somehow got a lastOwnerTime that is mroe recent than your current server time.

PR #820, opened 2019-01-04
Room and scene information dialog

gfodor:
You can now view the scene information + attributions for a room by clicking on the (i) icon next to the room name.

image

Depends on https://github.com/mozilla/reticulum/pull/104

In addition, uses the new farspark thumbnailing API to scale down/crop the scene image, and takes advantage of the new scene information in https://github.com/mozilla/reticulum/pull/104 

This PR also removes the "collapse" button on the landing page panel, which seemed to be confusing when collapsed since you're not "in the room" yet.


src/react-components/ui-root.js
```diff
@@ -1110,45 +1122,16 @@ class UIRoot extends Component {
...
-        <div className={entryStyles.entryDialog}>
-          <div>&nbsp;</div>
-          <WithHoverSound>
-            <button onClick={() => this.setState({ entryPanelCollapsed: false })} className={entryStyles.expand}>
-              <i>
-                <FontAwesomeIcon icon={faChevronUp} />
-              </i>
-            </button>
-          </WithHoverSound>
-        </div>
```
brianpeiris: I guess you know this already, but one downside to removing collapse is that the view is almost entirely blocked in landscape on iPhone 5/SE and you can't tap the user list or help buttons

image

brianpeiris:
Remember to make the PR title nicer before merging.

src/utils/media-utils.js
```diff
@@ -24,6 +24,23 @@ function b64EncodeUnicode(str) {
...
+  // farspark doesn't know how to read '=' base64 padding characters
+  // translate base64 + to - and / to _ for URL safety
+  return b64EncodeUnicode(url)
+    .replace(/=+$/g, "")
+    .replace(/\+/g, "-")
+    .replace(/\//g, "_");
+};
+
+export const scaledThumbnailUrlFor = (url, width, height) => {
+  if (process.env.FARSPARK_SERVER && process.env.NODE_ENV === "production") {
```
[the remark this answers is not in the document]
gfodor: oh, if there is a way to know if you are using a local reticulum node or not we could probably make it work. i was testing with the local reticulum instance, in which case you can't use dev farspark because the images are not internet accessible

src/react-components/ui-root.js
gfodor: yeah -- the problem is that the collapse button can fool the user into thinking they've entered the room -- it's actually worse on the devices where the panel takes up the most space because it feels like you've "clicked through" the dialog

src/utils/media-utils.js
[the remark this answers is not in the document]
brianpeiris: Oh right. Fair enough, we can change this later if we end up using local spoke + dev reticulum more often.

src/react-components/ui-root.js
brianpeiris: 👍 makes sense

src/utils/media-utils.js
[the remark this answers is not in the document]
gfodor: I'm just going to add an explicit check for hubs.local

src/react-components/ui-root.js
gfodor: I thought about this more and i think the right solution here is to add a grip point at the top the user can use to slide the panel up and down -- we can do that at some point

PR #821, opened 2019-01-04
Only time out session on mobile HMDs and firefox mobile

gfodor:
This PR updates the heuristic to determine if we should timeout the session on blur to only do so on mobile HMDs and on firefox mobile.


gfodor:
this behavior still does some stuff with oculus browser we'll need to figure out if we still need to do.

gfodor:
closing this for now, we will want to revisit this

PR #823, opened 2019-01-05
Add HTML history + back button to entry flow

gfodor:
Adds a back button to the entry flow and migrates the room page to use the HTML history API for all the various UI states.

image

This PR adds `react-router` and wires up the main room page to use routes. In dev mode we use hash history in prod it uses the real HTML history API.

Some notes:

- If you refresh the page during the entry flow, we don't want to start you off in the middle of it, so there's a special case that just resets the path to "/" in that situation

- If you enter the room and hit the back button, we don't want the entry flow to re-appear, and we also don't want you to end up on irrelevant URLs (eg the audio setup URL.) To get around this, if you hit back and end up being inside of the entry flow URL stack, the client just blows past it all and goes to the last history entry that was there before you landed on the page.

- All the modal dialogs are now entered and exited using HTML history, except for the sign in flow, since we don't want to have the back button re-enter the sign in flow.

- For modals, there's now the concept of the modal being successful or not (ie, was it not cancelled/closed with the X).

- When a modal is closed with the X, we navigate back in the browser history (so the modal is effectively not going to appear again if you hit back) but if the modal is successful, then we push `/` back onto the stack, so the back button will re-show it. (Note: you will currently *always* be brought to `/`, not what you were on before the modal appeared -- right now isn't relevant in the current flows since the only modal this happens for is the create dialog (and you are always at `/` before that) but we may need to revisit this.) I'm not convinced this is the ideal behavior, but the scenario I am thinking of is when we have a scene/avatar/object browser and you hit OK, the back button probably should bring you back into the browser.

One somewhat negative side effect of doing this using `react-router` idomatically results in us building virtual dom elements for all the non-visible dialogs. (Whereas before we would short circuit building these elements using our own internal state flags.) I'm not sure we care.

I also fixed up `vr_entry_type` to seemingly work better with `_now` again.


src/react-components/ui-root.js
```diff
@@ -195,6 +204,37 @@ class UIRoot extends Component {
...
+    if (ENTRY_FLOW_ROUTES.find(x => x === this.props.history.location.pathname)) {
+      this.props.history.replace("/");
+    }
+
+    // Hacky technique to skip over the entry flow history entries when we've entered the room and
+    // hit the back button.
+    //
+    // This makes it so if we are in the room and hit back in the browser, we go to the URL
+    // the browser was on before the entry flow, not back into the entry flow.
+    this.props.history.listen((newLocation, action) => {
```
netpro2k: Hmm this seems super fishy. The entered state should just be a route, and that route should `replace` all the entry flow routes in the history stack.

src/react-components/ui-root.js
```diff
@@ -195,6 +204,37 @@ class UIRoot extends Component {
...
     const scene = this.props.scene;
+
+    let preEntryHistoryLength = 0;
+
+    // If we landed on the page with a path in the middle of the entry flow (eg we refreshed the
+    // page on the audio setup dialog) then reset the history entry to /.
+    //
+    // Note this isn't perfect, if we refresh the page mid-entry flow and then hit back, we end
+    // up in a bad state unless we were on the first step. But this seems reasonable enough for now.
+    if (ENTRY_FLOW_ROUTES.find(x => x === this.props.history.location.pathname)) {
```
netpro2k: Not 100% sure but I think you can push states with the same path, so instead of making each entry flow step its own path we can just have a pushState that has the current flow step in it, that way back button still works, but the url doesn't change so refreshing still goes back to the beginning for the flow. This combined with a replace on the actual entry should fix these 2 hacks.

src/hub.js
```diff
@@ -214,18 +215,36 @@ function mountUI(props = {}) {
   const disableAutoExitOnConcurrentLoad = qsTruthy("allow_multi");
   const forcedVREntryType = qs.get("vr_entry_type");
 
+  const Router =
+    process.env.RETICULUM_SERVER && process.env.RETICULUM_SERVER !== document.location.host
+      ? HashRouter
+      : BrowserRouter;
```
netpro2k: I remember some nightmares with Android webviews and the BrowserRouter. Probably fine now, but worth giving extra care to testing Oculus Go.

src/react-components/profile-entry-panel.js
```diff
@@ -44,7 +46,14 @@ class ProfileEntryPanel extends Component {
         avatarId: this.state.avatarId
       }
     });
+
     this.props.finished();
+    this.props.history.goBack();
+
+    // We may need to go to a new path after saving.
+    if (this.props.location.state.postPushPath) {
+      this.props.history.push(this.props.location.state.postPushPath);
```
netpro2k: Hmm.. This goBack + conditional push feels odd. Being able to hit back to close the modal is maybe nice but I am not sure I expect to be able to go back to re-open a modal.

src/react-components/ui-root.js
```diff
@@ -44,17 +45,24 @@ import { faUsers } from "@fortawesome/free-solid-svg-icons/faUsers";
...
-const ENTRY_STEPS = {
-  start: "start",
-  device: "device",
-  mic_grant: "mic_grant",
-  mic_granted: "mic_granted",
-  audio_setup: "audio_setup",
-  finished: "finished"
-};
+// This needs to be updated as we add modal routes.
+const MODAL_ROUTES = [
```
netpro2k: Could alternatively define this as a state on the route or for bonus points make a `ModalRoute` component that does so, so you define this property as part of defining the route.

src/react-components/ui-root.js
```diff
@@ -228,17 +270,17 @@ class UIRoot extends Component {
       onContinueAfterSignIn();
     };
 
-    this.showDialog(SignInDialog, {
+    this.showNonHistoriedDialog(SignInDialog, {
```
netpro2k: Like I said above I would vote all the modals are non historied except that hitting back can close them. Seems simpler than having 2 different types.

src/react-components/ui-root.js
```diff
@@ -638,36 +670,37 @@ class UIRoot extends Component {
...
+      this.setState({ dialog: null });
+    } else {
+      // If dialog was successful (eg user hit "OK") then move forward in history, o/w go back.
+      //
+      // This makes it so if you create an object, back will re-show the create object dialog,
+      // but if you cancel, it will not.
+      if (success) {
+        this.props.history.push("/");
+      } else {
+        this.props.history.goBack();
```
netpro2k: This means I will be able to go "forward" into the dialog, not sure that is desirable.

src/react-components/ui-root.js
```diff
@@ -638,36 +670,37 @@ class UIRoot extends Component {
...
+  closeDialog = success => {
+    if (this.state.dialog) {
+      this.setState({ dialog: null });
+    } else {
+      // If dialog was successful (eg user hit "OK") then move forward in history, o/w go back.
+      //
+      // This makes it so if you create an object, back will re-show the create object dialog,
+      // but if you cancel, it will not.
+      if (success) {
+        this.props.history.push("/");
```
netpro2k: Again, not sure I really desire to hit back to re-open the dialog. I feel like if I am looking at a hubs room, hitting back should take me back to the lobby, if I am looking at a modal, it should close the modal.

src/hub.js
gfodor: note this is just for dev

brianpeiris:
Merging in #843

PR #824, opened 2019-01-07
Use skybox as environment map

robertlong:
Hubs now uses the skybox as the environment map. Previously there was a default environment map that made dark scenes look washed out. Now you can edit your scene's skybox in Spoke and the environment map will adjust accordingly.

beforeafterenvmap


src/components/environment-map.js
```diff
@@ -0,0 +1,43 @@
+import cubeMapPosX from "../assets/images/cubemap/posx.jpg";
+import cubeMapNegX from "../assets/images/cubemap/negx.jpg";
+import cubeMapPosY from "../assets/images/cubemap/posy.jpg";
+import cubeMapNegY from "../assets/images/cubemap/negy.jpg";
+import cubeMapPosZ from "../assets/images/cubemap/posz.jpg";
+import cubeMapNegZ from "../assets/images/cubemap/negz.jpg";
```
mquander: I personally dislike this style, because you don't usually look at imports and think "oh, right here is where I'm downloading N megabytes of stuff." I would rather make explicit fetches for them in `loadEnvMap`.

src/components/environment-map.js
```diff
@@ -0,0 +1,43 @@
...
+    source: { default: "skybox", type: "string" }
+  },
+
+  init() {
+    this.environmentMap = null;
+  },
+
+  update(oldData) {
+    if (oldData.source !== this.data.source && this.data.source === "defaultCubemap") {
+      // Used in the avatar selector scene because there is no skybox.
```
mquander: Is that really worth the effort of downloading the default environment map?

src/components/gltf-model-plus.js
```diff
@@ -407,6 +384,16 @@ AFRAME.registerComponent("gltf-model-plus", {
...
 
+      const environmentMapComponent = this.el.sceneEl.components["environment-map"];
+      const envMap = environmentMapComponent ? environmentMapComponent.environmentMap : null;
+
+      object3DToSet.traverse(object => {
+        if (object.material && object.material.isMeshStandardMaterial) {
+          object.material.envMap = envMap;
+          object.material.needsUpdate = true;
+        }
+      });
```
mquander: I would have made a function on the environment map component called `apply` which applies the current environment map to an object hierarchy, and you could call it both here and in `updateEnvironmentMap`.

src/components/skybox.js
```diff
@@ -231,6 +231,14 @@ AFRAME.registerComponent("skybox", {
   init() {
     this.sky = new THREE.Sky();
     this.el.setObject3D("mesh", this.sky);
+    this.skyScene = new THREE.Scene();
+    this.cubeCamera = new THREE.CubeCamera(1, 100000, 512);
```
mquander: Is 512 the best resolution to use? I see that the textures for our default environment map are 1024x1024. Perhaps we should inspect the environment map we have and decide at runtime?

src/components/directional-light.js
```diff
@@ -41,7 +41,7 @@ AFRAME.registerComponent("directional-light", {
     }
 
     if (this.data.shadowRadius !== prevData.shadowRadius) {
-      light.shadow.radius = this.data.radius;
+      light.shadow.radius = this.data.shadowRadius;
```
mquander: :thinking:

src/components/directional-light.js
```diff
@@ -41,7 +41,7 @@ AFRAME.registerComponent("directional-light", {
     }
 
     if (this.data.shadowRadius !== prevData.shadowRadius) {
-      light.shadow.radius = this.data.radius;
```
mquander: :thinking:

src/components/environment-map.js
robertlong: Idk, the bots do look darker and more menacing. So I'd say yes. It also only downloads if you open up the avatar selector scene modal. I think that modal needs a redesign anyway. We should look into improving it then.

src/components/skybox.js
```diff
@@ -280,6 +288,20 @@ AFRAME.registerComponent("skybox", {
 
       this.el.object3D.matrixNeedsUpdate = true;
     }
+
+    this.updateEnvironmentMap();
```
netpro2k: Hmm, update gets called on init, so it seems like you would end up rendering twice the first time. Perhaps we don't need the one in the requestAnimationFrame after all?

src/components/skybox.js
```diff
@@ -280,6 +288,20 @@ AFRAME.registerComponent("skybox", {
...
+  },
+
+  updateEnvironmentMap() {
+    this.skyScene.add(this.sky);
```
netpro2k: Shouldn't this just get added once in init? As it is you would end up adding it to the scene again each time any skymap property changes.

src/components/skybox.js
```diff
@@ -280,6 +288,20 @@ AFRAME.registerComponent("skybox", {
...
+    this.cubeCamera.update(this.el.sceneEl.renderer, this.skyScene);
+    this.el.setObject3D("mesh", this.sky);
```
netpro2k: This already happens in `init`, why does it need to get set again here?

src/components/skybox.js
```diff
@@ -280,6 +288,20 @@ AFRAME.registerComponent("skybox", {
...
+
+    const environmentMapComponent = this.el.sceneEl.components["environment-map"];
+
+    if (environmentMapComponent) {
```
netpro2k: I would just check this at the top and bail out of even rendering anything if there is no environment-map component.

src/components/environment-map.js
```diff
@@ -0,0 +1,46 @@
...
+async function loadEnvMap() {
+  const urls = [cubeMapPosX, cubeMapNegX, cubeMapPosY, cubeMapNegY, cubeMapPosZ, cubeMapNegZ];
+  const texture = await new THREE.CubeTextureLoader().load(urls);
+  texture.format = THREE.RGBFormat;
+  return texture;
+}
+
+AFRAME.registerComponent("environment-map", {
+  schema: {
+    source: { default: "skybox", type: "string" }
```
netpro2k: Its a bit weird that nothing ever actually looks for `skybox`, the only meaningful value for source is `defaultCubemap` and that even gets overwritten if there is a skybox. Maybe this should be nothing by default, then the skybox component would set it to `skybox` and then in `environment-map`'s `update` you could grab the skybox cubemap? 

Not too sure, but the way its split up now is a bit confusing.

src/components/skybox.js
robertlong: For some reason it needs to be called the frame after the skybox is initialized so I do need the requestAnimationFrame callback. I'm not 100% certain why though.

src/components/skybox.js
robertlong: add also removes it from the main scene https://github.com/mrdoob/three.js/blob/master/src/core/Object3D.js#L384 I create the skyScene so I can just render the skybox with the cube camera.

src/components/skybox.js
robertlong: When I call `this.skyScene.add(this.sky)` it is removed from the main scene. The `setObject3D` call adds it back to the main scene.

src/components/environment-map.js
robertlong: How about just changing this to a boolean `loadDefault` and default it to true. We'd then set it to false in hub.html and scene.html.

src/components/skybox.js
robertlong: Huh, it needs to be called in init and in the requestAnimationFrame callback... I don't know why.

PR #831, opened 2019-01-11
Support for HLS Video (Twitch, Crunchyroll, Youtube Live, Twitter, and more)

netpro2k:
Playing video from sites using HLS is now supported. This includes live streaming sites like Twitch and Youtube Live, as well as VOD streaming video sources like Crunchyroll and Twitter.

image

When the source is a live stream seek controls will be hidden and video syncing will be disabled (since the stream syncs itself). When native HLS support is available it will only be used if hls.js itself is not supported (due to media extensions not being supported) since it is still a useful interface for event and error handling.

Biggest hack is to do with proxying of relative urls. Since HLS.js internally tries to resolve the relative urls against the proxied url, we have to unpack the proxied url, and the resolve it against the original base url ourself. This would be more cleanly done by extending hls.js to expose its url resolving in an overridable way.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -518,9 +518,7 @@
       "requires": {
         "@tweenjs/tween.js": "^16.8.0",
         "browserify-css": "^0.8.2",
-        "debug": "github:ngokevin/debug#ef5f8e66d49ce8bc64c6f282c15f8b7164409e3a",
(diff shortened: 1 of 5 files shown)


mostafanimer89:
(momo_199017@yahoo.com)

PR #832, opened 2019-01-11
Add /scene and /rename commands

gfodor:
`/scene` and `/rename` commands are now available that let a user in the room change the scene or name of the room.

image

Depends upon https://github.com/mozilla/reticulum/pull/107 and https://github.com/mozilla/hubs/pull/827 to avoid memory leaks

The underlying mechanism is to send a `hub_update` or `scene_update` command to the hub channel, and clients will now recognize the `hub_refresh` message and act accordingly.

When refreshing the scene:

- The current scene is replaced by a (hardcoded) "loading scene" in order to free memory and ensure users can still communicate while the new scene is downloaded.

- The new scene is loaded by replacing the src on the environment entity's `gltf-model-plus` `src`.

This branch also refactors and rips out some more of the legacy gltf-bundle stuff, while not breaking backwards compatibility.

One note on memory: scene loading results in a number of `blob` `` resources to be stuck into the three.js cache. It seems the three.js image cache will hold these references even if the underlying objects are removed from the scene graph. So, for now, as a heuristic whenever we change the scene graph we also clear the entire three.js image cache. This may be problematic, I'm not sure if it means we will have to forcibly reload all the image resources for a scene or not as well (eg textures for other loaded models)

The alternative would be to get smarter about tracking which image cache entries are there due to the environment, and just purge those, but this would be quite tricky to track accurately given the asynchrony involved.


brianpeiris:
Merging in #843

PR #833, opened 2019-01-11
Feature/room ownership

brianpeiris:
- Adds a sign-in/out link to the home page
- Adds auth tokens to room creation and room join


src/hub.js
```diff
@@ -531,6 +531,10 @@ document.addEventListener("DOMContentLoaded", async () => {
 
   const pushSubscriptionEndpoint = await subscriptions.getCurrentEndpoint();
   const joinPayload = { profile: store.state.profile, push_subscription_endpoint: pushSubscriptionEndpoint, context };
+  const { token } = store.state.credentials;
+  if (token) {
+    joinPayload.token = token;
```
gfodor: notes from other PR, but these should prob be `auth_token` vs `perm_token`

src/react-components/home-root.js
```diff
@@ -41,12 +45,18 @@ class HomeRoot extends Component {
   state = {
     environments: [],
     dialog: null,
+    signedIn: null,
```
gfodor: nit, maybe make this `authenticated` or `signedIn` everywhere, not both

src/react-components/scene-ui.js
```diff
@@ -52,23 +51,8 @@ class SceneUI extends Component {
...
-
-    const hub = await res.json();
-
-    if (!process.env.RETICULUM_SERVER || document.location.host === process.env.RETICULUM_SERVER) {
-      document.location = hub.url;
-    } else {
-      document.location = `/hub.html?hub_id=${hub.hub_id}`;
-    }
+  createRoom = () => {
+    createAndRedirectToNewHub(this.state.customRoomName, this.props.sceneId);
```
gfodor: oops

gfodor:
LGTM mod the changes to move away from `is_owner` and towards using a `perm_token` (doesn't have to block but seems like we should just do it now.)

Also, nice to see the auth channel stuff migrated over to the home page nicely. 😅

brianpeiris:
Merging in #843

PR #835, opened 2019-01-12
Fix iOS HLS issues (probably)

netpro2k:
Fix video aspect ratio being broken, fix video being flipped, and an attempted fix for detecting live/VOD state (though I was not able to test this locally).

### DIFF
--- a/src/components/media-views.js
+++ b/src/components/media-views.js
@@ -136,7 +136,6 @@ function createVideoTexture(url, contentType) {
         hls.loadSource(url);
         hls.attachMedia(videoEl);
         hls.on(HLS.Events.ERROR, function(event, data) {
-          console.error(event, data);
           if (data.fatal) {
             switch (data.type) {
               case HLS.ErrorTypes.NETWORK_ERROR:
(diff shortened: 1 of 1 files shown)


src/components/media-views.js
```diff
@@ -174,7 +163,7 @@ function createVideoTexture(url, contentType) {
       videoEl.onerror = reject;
     }
 
-    videoEl.addEventListener("loadedmetadata", () => resolve(texture), { once: true });
+    videoEl.addEventListener("canplay", () => resolve(texture), { once: true });
```
netpro2k: Video size was not set yet when the `loadedmetadata` event fired for HLS videos. This is probably a more "correct" event for us to wait on for everything anyway

src/components/media-views.js
```diff
@@ -156,16 +155,6 @@ function createVideoTexture(url, contentType) {
       } else if (videoEl.canPlayType(contentType)) {
         videoEl.src = url;
         videoEl.onerror = reject;
-
-        // HACK aframe iOS HLS video hacks
-        if (isIOS) {
```
netpro2k: Looks like these hacks only break things, and things seem fine without them or any sort of special shader. Chalk it up to iOS updates?

src/components/media-views.js
```diff
@@ -499,13 +493,12 @@ AFRAME.registerComponent("media-video", {
         if (texture.hls.currentLevel >= 0) {
           updateLiveState();
         }
+      } else {
+        this.videoIsLive = this.video.duration === Infinity;
```
netpro2k: No clue if this is actually going to work, but I can't get a streaming source to work locally to try it out.

src/components/media-views.js
johnshaughnessy: down with the HACK

PR #838, opened 2019-01-15
Video Overlays (WIP)

netpro2k:
Move video playback controls into an overlay that is shown on hover instead of in the pause menu. Also add feedback for volume level and playback position.

This PR adds a component `hover-menu` which can specify a template to show on hover of the entity (or any child entities). The menu is automatically hidden when paused to not conflict with the pause menu. Components wishing to use the hover menu can call `getMenu()` on the component to get a reference to the instantiated menu element (and then query for whatever sub elements they want).

### DIFF
--- a/src/components/media-views.js
+++ b/src/components/media-views.js
@@ -287,15 +287,26 @@ AFRAME.registerComponent("media-video", {
     this.onPauseStateChange = this.onPauseStateChange.bind(this);
     this.tryUpdateVideoPlaybackState = this.tryUpdateVideoPlaybackState.bind(this);
 
-    this._grabStart = this._grabStart.bind(this);
-    this._grabEnd = this._grabEnd.bind(this);
     this.seekForward = this.seekForward.bind(this);
     this.seekBack = this.seekBack.bind(this);
+    this.togglePlaying = this.togglePlaying.bind(this);
 
     this.lastUpdate = 0;
 
-    this.seekForwardButton = this.el.querySelector(".video-seek-forward-button");
-    this.seekBackButton = this.el.querySelector(".video-seek-back-button");
+    this.el.setAttribute("hover-menu__video", { template: "#video-hover-menu", dirs: ["forward", "back"] });
+    this.el.components["hover-menu__video"].getHoverMenu().then(menu => {
+      this.hoverMenu = menu;
+
+      this.playPauseButton = this.el.querySelector(".video-playpause-button");
+      this.seekForwardButton = this.el.querySelector(".video-seek-forward-button");
+      this.seekBackButton = this.el.querySelector(".video-seek-back-button");
+
+      this.playPauseButton.addEventListener("grab-start", this.togglePlaying);
+      this.seekForwardButton.addEventListener("grab-start", this.seekForward);
+      this.seekBackButton.addEventListener("grab-start", this.seekBack);
+
+      this.updatePlaybackState();
+    });
 
     NAF.utils.getNetworkedEntity(this.el).then(networkedEl => {
       this.networkedEl = networkedEl;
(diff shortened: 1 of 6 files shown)


src/components/media-views.js
```diff
@@ -545,26 +549,39 @@ AFRAME.registerComponent("media-video", {
...
 
   tick() {
+    if (!this.video) return;
+
     const userinput = this.el.sceneEl.systems.userinput;
     const volumeMod = userinput.get(paths.actions.cursor.mediaVolumeMod);
-    if (volumeMod) {
+    if (this.el.is("hovered") && volumeMod) {
       this.el.setAttribute("media-video", "volume", THREE.Math.clamp(this.data.volume + volumeMod, 0, 1));
+      this.volumeLabel.setAttribute(
```
johnshaughnessy: You might avoid some work with:
```
      this.volumeLabel.components.text.updateGeometry(this.geometry, font);
      this.volumeLabel.components.text.updateLayout();
```
We have reservations for `setAttribute` and also the `update` for the `text` component does some checks that aren't necessary for you (because you aren't changing font or shader or texture here.

src/components/media-views.js
```diff
@@ -545,26 +549,39 @@ AFRAME.registerComponent("media-video", {
...
+        "value",
+        `${timeFmt(this.video.currentTime)} / ${timeFmt(this.video.duration)}`
+      );
     }
 
-    const now = performance.now();
-    if (now - this.lastUpdate > this.data.tickRate) {
-      this.el.setAttribute("media-video", "time", this.video.currentTime);
-      this.lastUpdate = now;
+    // If a none live video is currently playing and we own it, send out time updates
```
johnshaughnessy:
```suggestion
    // If a non-live video is currently playing and we own it, send out time updates
```

src/components/media-views.js
```diff
@@ -388,6 +376,11 @@ AFRAME.registerComponent("media-video", {
       this.video.removeEventListener("pause", this.onPauseStateChange);
       this.video.removeEventListener("play", this.onPauseStateChange);
     }
+    if (this.hoverMenu) {
```
johnshaughnessy: Might not exist if somehow this `remove` is called before the `hoverMenu` promise resolves

src/components/media-views.js
```diff
@@ -545,26 +549,39 @@ AFRAME.registerComponent("media-video", {
...
       this.el.setAttribute("media-video", "volume", THREE.Math.clamp(this.data.volume + volumeMod, 0, 1));
+      this.volumeLabel.setAttribute(
+        "text",
+        "value",
+        this.data.volume === 0 ? "MUTE" : `VOL: ${Math.round(this.data.volume * 100)}%`
+      );
+      this.volumeLabel.object3D.visible = true;
+      clearTimeout(this.hideVolumeLabelTimeout);
+      if (this.data.volume) {
+        this.hideVolumeLabelTimeout = setTimeout(() => (this.volumeLabel.object3D.visible = false), 1000);
```
johnshaughnessy: Ok to set a timeout every frame? I suppose people aren't messing with volume all the time... 

An alternative might be to write the time that the volume label should be made invisible here since tick is going to keep being called

src/components/media-views.js
```diff
@@ -399,6 +392,12 @@ AFRAME.registerComponent("media-video", {
   },
 
   updatePlaybackState(force) {
+    if (this.hoverMenu) {
+      this.playPauseButton.object3D.visible = !!this.video;
+      this.seekForwardButton.object3D.visible = !!this.video && !this.videoIsLive;
+      this.seekBackButton.object3D.visible = !!this.video && !this.videoIsLive;
+    }
+
     // Only update playback posiiton for videos you don't own
```
johnshaughnessy: posiiton -> position

src/components/media-views.js
netpro2k: This would leave the components data in a weird state. Since this is only when adjusting volume I am not too worried about it.

src/components/media-views.js
netpro2k: Yeah again, since this is only when manipulating volume I don't feel too worried about it.

PR #841, opened 2019-01-16
Authorize hub mutation commands

brianpeiris:
PR for review only. Based on a merge between #832 and #833. Goes with mozilla/reticulum#110

### DIFF
--- a/src/message-dispatch.js
+++ b/src/message-dispatch.js
@@ -39,6 +39,7 @@ export default class MessageDispatch {
     const playerRig = document.querySelector("#player-rig");
     const scales = [0.0625, 0.125, 0.25, 0.5, 1.0, 1.5, 3, 5, 7.5, 12.5];
     const curScale = playerRig.object3D.scale;
+    let err;
 
     switch (command) {
       case "fly":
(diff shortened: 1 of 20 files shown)


src/utils/hub-channel.js
```diff
@@ -97,10 +105,12 @@ export default class HubChannel {
   };
 
   updateScene = url => {
+    if (!this._permissions.update_hub) return "unauthorized";
```
gfodor: perms token expire in 5 minutes, so we'll need to add some plumbing to refetch a fresh one as needed

gfodor: we could potentially mitigate this issue for now by just making these tokens expire in 24 hours and revisiting, but it doesn't seem too hard to just refresh this (we'll just need another hub channel message to get a new token)

src/message-dispatch.js
```diff
@@ -78,10 +79,16 @@ export default class MessageDispatch {
         this.scene.emit("quack");
         break;
       case "scene":
-        this.hubChannel.updateScene(args[0]);
+        err = this.hubChannel.updateScene(args[0]);
+        if (err === "unauthorized") {
```
gfodor: ideally the way this would work is:
- every 2-3 minutes, we refresh a perms token and refresh the UI
- if you do not have permission to run these commands, we simply don't include them in the chat commands list

there would still be an edge case if you lost permissions and we hadn't updated perm token yet, but that seems like not a big concern.

gfodor:
this could go out as-is with a follow up to add perms token refreshing as mentioned, or now. either way lgtm!

src/utils/hub-channel.js
brianpeiris: yup, I'd imagined we'd need to request a new token before every kick/mute action anyway, so I'm going to leave it off to when I do that work

brianpeiris:
Gonna try to close and open a new PR for the final merge.

src/message-dispatch.js
brianpeiris: Arguably, I think we could postpone adding token refresh until we have more discord integration, where we would have to worry about people being added/removed to roles dynamically. Should be easy to do, as you said, but I'm going to punt for now.

PR #842, opened 2019-01-16
Feature/rotation buttons

johnshaughnessy:
You can now use the "rotate" and "align" buttons to reorient objects in the scene.

image


src/components/cursor-controller.js
```diff
@@ -89,11 +89,17 @@ AFRAME.registerComponent("cursor-controller", {
   emitIntersectionEvents: function(prevIntersection, currIntersection) {
     // if we are now intersecting something, and previously we were intersecting nothing or something else
     if (currIntersection && (!prevIntersection || currIntersection.object.el !== prevIntersection.object.el)) {
-      this.data.cursor.emit("raycaster-intersection", { el: currIntersection.object.el });
+      this.data.cursor.emit("raycaster-intersection", {
+        el: currIntersection.object.el,
+        intersection: currIntersection
```
johnshaughnessy: oops, this change is going away.

src/components/rotate-object-button.js
```diff
@@ -0,0 +1,334 @@
...
+    this.raycaster.far = far;
+    const intersection = intersections[0]; // point
+    if (!intersection) return;
+
+    normal.set(0, 0, -1).applyQuaternion(plane.quaternion);
+
+    Pc.copy(intersection.point);
+    PiPc.copy(Pc).sub(Pi);
+    if (this.mode === ROTATE_MODE.FREE || this.mode === ROTATE_MODE.GARY) {
+      const boost = AFRAME.scenes[0].systems.userinput.get(paths.actions.boost);
```
johnshaughnessy: TODO : move this out of right here and fix the "FREE" / "GARY" confusion where the two modes are smushed messily together here and swapped with the "boost" modifier key.

src/systems/userinput/resolve-action-sets.js
```diff
@@ -140,6 +145,7 @@ export function resolveActionSets() {
   userinput.toggleSet(sets.cursorHoldingPen, cursorHoldingPen);
   userinput.toggleSet(sets.cursorHoldingCamera, cursorHoldingCamera);
   userinput.toggleSet(sets.cursorHoldingInteractable, cursorHoldingInteractable);
+  userinput.toggleSet(sets.cursorHoldingUI, cursorHoldingUI);
```
johnshaughnessy: TODO: I was having issues getting this recognized with super-hands so didn't use this set yet. Not strictly necessary but want to figure out if the actions I'm polling are right and if I'm polling in the right spots

src/components/rotate-object-button.js
```diff
@@ -0,0 +1,334 @@
...
+      this.dualAxis = !boost;
+      this.gary = boost;
+      PpPc.copy(Pc).sub(Pp);
+      XYi.copy(PiPc)
+        .projectOnPlane(normal)
+        .applyQuaternion(this.q.copy(plane.quaternion).inverse())
+        .multiplyScalar(5 / cameraToPlaneDistance);
+      if (XYi.lengthSq() < 0.00001) {
```
netpro2k: This feels off. Either come up with a value that feels "good" or use `Number.EPSILON` if you really want to filter out "really small numbers". But honestly I doubt you need it.

src/components/rotate-object-button.js
```diff
@@ -0,0 +1,334 @@
...
+        return;
+      }
+
+      const stepLength = Math.PI / 6;
```
netpro2k: I would move this out of this function into a constant.

src/components/rotate-object-button.js
```diff
@@ -0,0 +1,334 @@
...
+      this.dyAll = this.dyStore + XYi.y;
+      this.dyApplied = this.dualAxis ? Math.round(this.dyAll / stepLength) * stepLength : this.dyAll;
+      this.dyStore = this.dyAll - this.dyApplied;
+
+      this.dzAll = this.dzStore + XYi.x;
+      this.dzApplied = this.dualAxis ? Math.round(this.dzAll / stepLength) * stepLength : this.dzAll;
+      this.dzStore = this.dzAll - this.dzApplied;
+
+      this.o.matrixWorld.decompose(this.v, this.objectWorldQuaternion, this.v);
+      this.right.set(1, 0, 0).applyQuaternion(this.dualAxis ? this.objectWorldQuaternion : this.cameraWorldQuaternion);
```
netpro2k: You are mutating your constant here. Don't.

src/components/rotate-object-button.js
```diff
@@ -0,0 +1,334 @@
...
+      this.q.setFromAxisAngle(this.right, this.dualAxis ? this.sign2 * this.sign * -this.dyApplied : -this.dyApplied);
```
netpro2k: `sign` and `sign2` should probably be computed every frame.

src/components/rotate-object-button.js
```diff
@@ -0,0 +1,334 @@
...
+
+      const rotationSnap = Math.PI / 8;
+      this.dzAll = this.dzStore + XYi.x;
+      this.dzApplied = Math.round(this.dzAll / rotationSnap) * rotationSnap;
+      this.dzStore = this.dzAll - this.dzApplied;
+
+      this.o.quaternion.multiply(this.q.setFromAxisAngle(this.axis, -this.sign * this.dzApplied));
+    }
+
+    Pi.copy(intersection.point);
```
netpro2k: This should probably be `Pp` since this is overwriting the initial point.

src/components/rotate-object-button.js
```diff
@@ -0,0 +1,334 @@
...
+  }
+});
+
+AFRAME.registerComponent("exist-if-threedof", {
```
netpro2k: Since all we actually care to do is set rotation mode I would call this something like `rotation-mode-selector` and check if hand poses exist each tick. If it has changed, change the `rotation-button` mode.

src/components/visibility-while-frozen.js
```diff
@@ -54,7 +54,10 @@ AFRAME.registerComponent("visibility-while-frozen", {
         this.camWorldPos.distanceToSquared(this.objWorldPos) < this.data.withinDistance * this.data.withinDistance;
     }
 
-    const shouldBeVisible = ((isFrozen && this.data.visible) || (!isFrozen && !this.data.visible)) && isWithinDistance;
+    const shouldBeVisible =
+      ((isFrozen && this.data.visible) || (!isFrozen && !this.data.visible)) &&
+      isWithinDistance &&
+      !AFRAME.scenes[0].systems["rotate-selected-object"].active;
```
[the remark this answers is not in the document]
johnshaughnessy: Is it actually less odd for the rotate button to be setting the visibility of the freeze menu? I thought the ideal situation be that the rotate button wouldn't know about the freeze menu and vice versa (e.g. visibility arbiter). Not sure if it's worth the effort to decomplect them like that, but seems like it's arbitrary whether visibility-while-frozen depends on rotate-button or vice versa.

src/components/rotate-object-button.js
```diff
@@ -0,0 +1,334 @@
...
+      this.o.matrixWorld.decompose(plane.position, q, v);
+      plane.matrixNeedsUpdate = true;
+      plane.updateMatrixWorld(true);
+
+      this.raycaster =
+        this.raycaster || document.querySelector("#cursor-controller").components["cursor-controller"].raycaster;
+      intersections.length = 0;
+      const far = this.raycaster.far;
+      this.raycaster.far = 1000;
+      plane.raycast(this.raycaster, intersections);
```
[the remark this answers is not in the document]
johnshaughnessy: postponing for now because i like being able to turn on the plane while debugging and also i don't know the point of the "constant" parameter of the three.js Plane object

PR #852, opened 2019-01-25
Handle failed pins due to invalid token

brianpeiris:
Check reply from pinning message and handle invalid token by signing user out and prompting for sign in

### DIFF
--- a/src/scene-entry-manager.js
+++ b/src/scene-entry-manager.js
@@ -18,8 +18,9 @@ function requestFullscreen() {
 }
 
 export default class SceneEntryManager {
-  constructor(hubChannel) {
+  constructor(hubChannel, authChannel) {
     this.hubChannel = hubChannel;
+    this.authChannel = authChannel;
     this.store = window.APP.store;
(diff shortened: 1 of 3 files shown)


src/scene-entry-manager.js
```diff
@@ -195,7 +196,46 @@ export default class SceneEntryManager {
     el.setAttribute("networked", { persistent: true });
     el.setAttribute("media-loader", { fileIsOwned: true });
 
-    this.hubChannel.pin(networkId, gltfNode, fileId, fileAccessToken, promotionToken);
+    try {
+      await this.hubChannel.pin(networkId, gltfNode, fileId, fileAccessToken, promotionToken);
+    } catch (e) {
+      if (e.reason === "invalid_token") {
+        await this.authChannel.signOut(this.hubChannel);
```
gfodor: currently the hub page load ignores `invalid_tokens` and just chugs along -- you might want to handle that properly too now. iirc the reason i punted was because we were in incident response and I wasn't sure how to properly update all the UI state, etc.

PR #860, opened 2019-01-29
Add optional GA tracking telemetry

gfodor:
This PR adds GA instrumentation (if configured) to track page landing loads for the main hubs entry pages.

- We do not track individual room URLs, but roll up every room page load into a single logical /hub path
- Scene URLs are collapsed similarly unless the scene is marked as promotable


brianpeiris:
Would you care to add it to `whats-new.html` as well?

PR #861, opened 2019-01-31
Room creators can now kick users

brianpeiris:
If you are the creator of a room, and you are logged in, you can now kick users. They will leave the room along with the media they own.

image

- We now set the NAF adapter's `clientId` explicitly before connecting. The clientId is set to the same uuid used for the Phoenix socket `session_id`, so that we can correlate NAF clients to phoenix sockets when kicking.
- NAF adapter now has a `joinToken`, which is just the `perms_token` (though joining is not enforced yet).
- Adds a "Kick" button in avatar menus
- Permissions are refreshed before dispatching a kick message to both reticulum and janus. Janus cuts off their RTC traffic and reticulum kills their phoenix socket. The clicked client is exited from the scene and its owned entities are removed.
- Adds a `permissions` system in A-Frame that allows you to fetch and query permissions
- `window.APP` now has `hubChannel`
- Adds a `visible-if-permitted` component which shows entities only if the given permission is granted
- Renamed "Block" button to "Hide"

Goes with:
https://github.com/mozilla/reticulum/pull/114
https://github.com/mozilla/hubs-ops/pull/44
https://github.com/mozilla/janus-plugin-sfu/pull/29
https://github.com/mozilla/naf-janus-adapter/pull/70
https://github.com/MozillaReality/networked-aframe/pull/22

### DIFF
--- a/src/components/kick-button.js
+++ b/src/components/kick-button.js
@@ -1,8 +1,3 @@
-/**
- * Registers a click handler and invokes the kick method on the NAF adapter for the owner associated with its entity.
- * @namespace network
- * @component kick-button
- */
 AFRAME.registerComponent("kick-button", {
   init() {
     this.onClick = () => {
--- a/src/components/kick-button.js
+++ b/src/components/kick-button.js
@@ -1,0 +1,29 @@
+/**
+ * Registers a click handler and invokes the kick method on the NAF adapter for the owner associated with its entity.
+ * @namespace network
+ * @component kick-button
+ */
+AFRAME.registerComponent("kick-button", {
+  init() {
+    this.onClick = () => {
+      this.kick(this.owner);
+    };
+    NAF.utils.getNetworkedEntity(this.el).then(networkedEl => {
+      this.owner = networkedEl.components.networked.data.owner;
+    });
+  },
+
+  play() {
+    this.el.addEventListener("grab-start", this.onClick);
+  },
+
+  pause() {
+    this.el.removeEventListener("grab-start", this.onClick);
+  },
+
+  async kick(clientId) {
+    const { permsToken } = await this.el.sceneEl.systems.permissions.fetchPermissions();
+    NAF.connection.adapter.kick(clientId, permsToken);
+    window.APP.hubChannel.kick(clientId);
+  }
+});
--- a/src/hub.js
+++ b/src/hub.js
@@ -356,6 +356,7 @@ async function handleHubChannelJoined(entryManager, hubChannel, messageDispatch,
     });
 
     while (!scene.components["networked-scene"] || !scene.components["networked-scene"].data) await nextTick();
+
     scene.components["networked-scene"]
       .connect()
       .then(() => {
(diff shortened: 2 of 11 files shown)


src/hub.js
```diff
@@ -569,6 +574,13 @@ document.addEventListener("DOMContentLoaded", async () => {
   });
 
   const socket = connectToReticulum(isDebug);
+
+  socket.onClose(() => {
+    // The socket should only close of the server has explicitly killed it.
+    entryManager.exitScene();
+    remountUI({ roomUnavailableReason: "left" });
```
brianpeiris: @gfodor I re-used the existing "You have left the room" message here, dunno if we want different messaging.

src/components/kick-button.js
```diff
@@ -0,0 +1,24 @@
...
+  },
+
+  pause() {
+    this.el.removeEventListener("grab-start", this.onClick);
+  },
+
+  async kick(clientId) {
+    const { permsToken } = await this.el.sceneEl.systems.permissions.fetchPermissions();
+    NAF.connection.adapter.kick(clientId, permsToken);
+    window.APP.hubChannel.kick(clientId);
```
brianpeiris: Note I've changed the NAF `clientId`, it is now just the phoenix socket `session_id`. Not sure if that's a bad idea, but I couldn't think of a better way of correlating a NAF client to a phoenix socket.

gfodor: that sounds right to me

src/hub.js
gfodor: i think they should know they were kicked because (since they can refresh anyway) the main thing kicking is doing is giving the owner a way to nudge them to behave better. (once we have OAuth in place, it can become a ban)

src/scene-entry-manager.js
```diff
@@ -167,6 +168,12 @@ export default class SceneEntryManager {
     this.scene.emit("username-changed", { username: displayName });
   };
 
+  _setupKicking = () => {
+    document.body.addEventListener("kicked", ev => {
+      NAF.connection.entities.removeEntitiesOfClient(ev.detail.clientId);
```
gfodor: I'm guessing this is only going to remove their owned objects, which seems OK for now -- but I think we probably want to do a pass where we start capturing the client id of the user who created the object, and remove the objects the kicked user created. (Obvious, but if you touch a kicked user's object, it still probably makes sense to remove it if they are kicked, and alternatively if they touch your objects before they are kicked probably doesn't make sense to remove them.)

src/utils/hub-channel.js
```diff
@@ -196,6 +200,22 @@ export default class HubChannel {
     this.channel.push("unpin", payload);
   };
 
+  fetchPermissions = () => {
+    return new Promise((resolve, reject) => {
+      this.channel
+        .push("refresh_perms_token")
+        .receive("ok", res => {
+          this.setPermissionsFromToken(res.perms_token);
+          resolve({ permsToken: res.perms_token, permissions: this._permissions });
```
gfodor: is the passed `permissions` used here?

brianpeiris: Ah, no but I imagined it will be used when we want to update more of the UI based on permissions

src/hub.js
brianpeiris: Made it "You have been kicked from the room"

Issue #810, opened 2018-12-29
Chat box inside the room


BhuvanaMeenakshiK:

brianpeiris:
In case you missed it, there is actually a chat box in the bottom left corner after you've entered the room:

![image](https://user-images.githubusercontent.com/79419/50569623-a2061b80-0d1e-11e9-81c6-f196d8d97f62.png)

If you're referring to using text chat from inside a VR headset, we've considered solutions to this but have avoided it so far since typing in VR can be difficult for chat especially. 

Thanks for the suggestion either way, it's good to know what problems people are running into.

BhuvanaMeenakshiK:
> In case you missed it, there is actually a chat box in the bottom left corner after you've entered the room:
> 
> ![image](https://user-images.githubusercontent.com/79419/50569623-a2061b80-0d1e-11e9-81c6-f196d8d97f62.png)
> 
> If you're referring to using text chat from inside a VR headset, we've considered solutions to this but have avoided it so far since typing in VR can be difficult for chat especially.
> 
> Thanks for the suggestion either way, it's good to know what problems people are running into.

Yes I went through the chat box here after a long time of search (because the UI was very light in color and could not be differentiated  until I started to type). But it could be better enough if it displayed against the avatar of each person. Also I saw the messages disappearing in a while, this is not effective when new users join the group. Because if anything important discussed or some links been shared over then the messages should be available everytime for future reference.

And absolutely fair enough to the fact that it becomes useless when VR heaset is used but a virtual keyboard can still be added to type the messages. The reason why I am insisting on the textual input is that last time I hosted a small conference among the Mozillians of my region and due to some problem of mic people were not able to interact with each other so we took the help of the message box to continue our converssation.

brianpeiris:
Cool, so there are a few things to unpack here, that may be split into separate issues:

1. Make the existing chat box easier to spot when you're in the room
    This is probably just a matter of adding some contrast to it so that it's visible in both light and dark scenes.
2. Add the ability to send a message to individual users. 
    I believe this will require some backend work, since we don't have a way to direct messages to a particular user at the moment, IIRC. 
    If the backend were in place already, we'd have to decide how to expose it in the client -- one idea I can think of is to add a "Message" button to each avatar in pause mode and that message button just focuses the existing chat box and populates it with a new `/whisper <avatar-name> <msg>` command, although that idea might fall apart if there are multiple avatars with the same name in the room.
3. Add the ability to persist and view the chat history of a room
[...]

PR #862, opened 2019-01-31
Making the chat box easier to spot

Mordax:
This refers to #810, one part of a series of enhancements to the chatbox.

- [x] Make the existing chat box easier to spot when you're in the room
This is probably just a matter of adding some contrast to it so that it's visible in both light and dark scenes.

I decided to follow the pink/red schema and just give it a coloured background: 
demo

Let me know your thoughts if you folks like it, or there's something I missed, etc.

### DIFF
--- a/src/assets/stylesheets/ui-root.scss
+++ b/src/assets/stylesheets/ui-root.scss
@@ -212,7 +212,7 @@ body.vr-mode {
   overflow: hidden;
   resize: none;
   background-color: transparent;
-  color: black;
+  color: $darkest-transparent;
   padding: 8px 1.25em;
   line-height: 28px;
   font-size: 1.1em;
@@ -223,7 +223,7 @@ body.vr-mode {
 }
 
 :local(.message-entry-input)::placeholder{
-  color: $dark-grey;
+  color: $light-grey;
   font-weight: 300;
   font-style: italic;
 }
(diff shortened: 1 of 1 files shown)


gfodor:
Hey @Mordax thank you for the PR! We have a redesign in the works for the chat area of Hubs (we're going to be repositioning and increasing the priority of the chat box overall) and so we'll take into account this variant you've shown here with the action button background as we determine what style changes to make.

I super appreciate you surfacing this usability issue and I hope we'll be able to get these design changes out soon. I'll close this PR for now -- once the updates are out (likely by late Feb, but not totally sure) then please feel free to resurface further contrast issues or provide additional design suggestions/PRs.

Thanks again!

gfodor:
Actually I re-opened this, let me take a quick look to see if we can make some tweaks together.

gfodor:
So here's what I'll suggest, which would get us a good step closer to some of the redesign. We'd like the chat box on the bottom to be similar to the one we show in the entry step here:

image

It should still likely be reduced in opacity when not focused, but the background should become white with black text (basically, the exact same style as the lobby chat box above) when it's focused. I think you can probably do this by either re-using the existing lobby chat box styles (`.message-entry`, `.message-entry-input`, etc) and then having the existing `-in-room` styles just apply slight changes to it to adjust the opacity a bit. Or for now you could just skip the opacity changes and just change the existing code to no longer use the `-in-room` styles at all and see how that looks.

Thanks again for looking into this, let me know if you think those style changes sound reasonable to try!

Mordax:
@gfodor (~~screwed up my branch sorry about that~~) - Here's the demo of the new updated code:
demo

gfodor:
this is going out now, hope it helps! https://github.com/mozilla/hubs/pull/896

PR #863, opened 2019-02-01
Feature/react admin

gfodor:
Adds a super basic admin console. Goes with https://github.com/mozilla/reticulum/pull/115.

To use it you need to be tunneled to PostgREST on port 3000. (Currently you can develop locally without too much trouble.)

You'll need to have previously logged into an admin account.

At the time of this writing PostgREST is live in dev but database is not migrated yet to support this. The new database migration scripts are here:

https://github.com/mozilla/hubs-ops/tree/master/admin/migrations


gfodor:
the build size bot output is actually worth looking at in this one. i fixed a bug where we were not packing THREE into engine but vendor, and then i tried to scoop of most of the react admin dependencies into a dedicated admindeps js file. the net effect is vendor went down some, engine went up some, but not by equal amounts because some of the dependencies of react admin did go into the vendor js. (since I didn't try to write a regex to capture all of them, just the bigger/most obvious ones)

src/utils/postgrest-data-provider.js
```diff
@@ -0,0 +1,308 @@
...
+        headers: response.headers,
+        body: text
+      }))
+    )
+    .then(({ status, statusText, headers, body }) => {
+      let json;
+      try {
+        json = json2ParseBigint(body);
+      } catch (e) {
+        // not json, no big deal
```
brianpeiris: Might be useful for debugging if we logged a warning here at least. Also maybe `reject()` here as well.

src/utils/postgrest-data-provider.js
```diff
@@ -0,0 +1,308 @@
...
+      }
+      if (status < 200 || status >= 300) {
+        return Promise.reject(new HttpError((json && json.message) || statusText, status, json));
+      }
+      return Promise.resolve({ status, headers, body, json });
```
brianpeiris: Pretty sure the `Promise.reject` and `Promise.resolve` are redundant here. You can just return the value directly and just `throw` the error.

src/utils/postgrest-data-provider.js
```diff
@@ -0,0 +1,308 @@
...
+ * DELETE       => DELETE http://my.api.url/posts?id=eq.123
+ */
+const postgrestClient = (apiUrl, httpClient = fetchJson) => {
+  const convertFilters = filters => {
+    const rest = {};
+
+    Object.keys(filters).map(function(key) {
+      switch (typeof filters[key]) {
+        case "string":
+          rest[key] = "ilike.*" + filters[key].replace(/:/, "") + "*";
```
brianpeiris: Is this removing colons in the query? Can't you escape them instead?

brianpeiris: Same below.

src/utils/postgrest-data-provider.js
```diff
@@ -0,0 +1,308 @@
...
+
+        case "number":
+          rest[key] = "eq." + filters[key];
+          break;
+
+        case "object":
+          if (filters[key].constructor === Array) {
+            rest[key] = "cs.{" + filters[key].toString().replace(/:/, "") + "}";
+          } else {
+            Object.keys(filters[key]).map(val => (rest[`${key}->>${val}`] = `ilike.*${filters[key][val]}*`));
```
brianpeiris: Forgot to account for colons in the `ilike` here, I think.

src/utils/postgrest-data-provider.js
```diff
@@ -0,0 +1,308 @@
...
+   * @param {String} type One of the constants appearing at the top if this file, e.g. 'UPDATE'
+   * @param {String} resource Name of the resource to fetch, e.g. 'posts'
+   * @param {Object} params The REST request params, depending on the type
+   * @returns {Object} { url, options } The HTTP request parameters
+   */
+  const convertRESTRequestToHTTP = (type, resource, params) => {
+    let url = "";
+    const options = {};
+    options.headers = new Headers();
+    const token = localStorage.getItem(localStorageKey);
```
brianpeiris: Can't we avoid using localStorage here? Seems pointless for a token that expires every 5 minutes.

gfodor: i'll confirm, i think you're right but we might end up seeing the login page anyway if we don't have that token there to be read immediately so the initial postgrest queries pass because of some react admin internals I didn't bother digging into yet. if it works without it i'll take it out otherwise i'll leave it, my plan was to try to polish up the login screen workflow at some point. (currently, you end up landing on a meaningless login screen when you first load the page)

PR #867, opened 2019-02-01
Use postMessage bidireactionally for managing avatar state

gfodor:
This fixes an issue due to the use of the URL hash as a way of transmitting avatar id state into and out of the avatar selector. Because we now do a history.pop() when closing a modal, it ended up breaking the "accept" button. This PR updates things to use postmessage to synchronize the state exclusively.

### DIFF
--- a/src/react-components/profile-entry-panel.js
+++ b/src/react-components/profile-entry-panel.js
@@ -56,6 +56,13 @@ class ProfileEntryPanel extends Component {
       return;
     }
     this.setState({ avatarId: e.data.avatarId });
+
+    // Send it back down to cause avatar picker UI to update.
+    this.sendAvatarStateToIframe();
+  };
+
+  sendAvatarStateToIframe = () => {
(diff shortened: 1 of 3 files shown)


src/react-components/profile-entry-panel.js
```diff
@@ -121,8 +128,18 @@ class ProfileEntryPanel extends Component {
...
                   className={styles.avatarSelector}
-                  src={`/avatar-selector.html#avatar_id=${this.state.avatarId}`}
-                  ref={ifr => (this.avatarSelector = ifr)}
+                  src={`/avatar-selector.html`}
+                  ref={ifr => {
+                    if (this.avatarSelector === ifr) return;
+
+                    this.avatarSelector = ifr;
+
+                    if (this.avatarSelector) {
```
brianpeiris: Hmm, is this ever falsy?

brianpeiris: Gonna push with this still in, but it's a bit suspicious

PR #874, opened 2019-02-06
Improve loading with less hitching and more waiting

johnshaughnessy:
image

image

This PR improves the loading experience by doing more work up front and giving the user information along the way. The specific changes include: 
- the loading screen now gives some indication of the progress being made.
- the loading screen remains until all objects in the room have finished loading 
- we call a new API on the three.js renderer called `compileAndUploadMaterials` which allows us to get the shaders compiled and textures uploaded so that this doesn't happen for the first time when objects come into the view frustum.
- similarly, the user's avatar is created before they enter the room so that it has the chance to compile its shader and upload its textures to the FPU.

Before this change, the app stutters as you first look around the room and objects are rendered for the first time (notice several long frames (each ~100-200ms) a while after the first long frames of loading):
hitching

After this change, the app only freezes  while the objects in the room are loading.


gfodor:
Is it turned back on somehow?

src/components/gltf-model-plus.js
```diff
@@ -325,7 +324,15 @@ AFRAME.registerComponent("gltf-model-plus", {
...
 
+      window.stuffToLoad += 1;
+      let text = `Loading ${window.stuffToLoad} more object${window.stuffToLoad > 1 ? "s" : ""}...`;
+      window.uiroot && window.uiroot.setState({ loadingText: text });
       const gltf = await this.loadModel(src, contentType, this.preferredTechnique, this.data.useCache);
+      window.stuffToLoad -= 1;
+      if (window.stuffToLoad > 0) {
+        text = `Loading ${window.stuffToLoad} more object${window.stuffToLoad > 1 ? "s" : ""}...`;
+        window.uiroot && window.uiroot.setState({ loadingText: text });
+      }
```
brianpeiris: Hmm, this code is pretty surprising to me.
Firstly, we should probably not just treat `window` like a grab-bag of properties. If we do have to use global state like this, we should at least namespace it under our existing `APP` global, maybe under an `APP.resourceLoading` namespace as well.

It's also surprising to have an a-frame component call `setState` directly on a react component, also via property on the window global

src/components/gltf-model-plus.js
```diff
@@ -407,8 +414,16 @@ AFRAME.registerComponent("gltf-model-plus", {
...
       });
 
+      this.el.object3D.traverse(obj => {
+        obj.frustumCulled = false;
+      });
       this.el.emit("model-loaded", { format: "gltf", model: this.model });
+      await AFRAME.scenes[0].systems.ticker.nextTick();
+      this.el.object3D.traverse(obj => {
+        obj.frustumCulled = true;
+      });
```
brianpeiris: Is it correct to emit `model-loaded` before we turn frustumCulling back on? If the aim here is to wait for the model to be uploaded to the GPU, perhaps it's better not to claim `model-loaded` until the GPU upload is complete.

src/assets/stylesheets/loader.scss
```diff
@@ -89,3 +89,13 @@
...
 }
+.loadedText {
+    margin: 1em 0 0 0;
+    color: green;
+    bottom: -10px;
+}
+.loadingText {
+    margin: 1em 0 0 0;
+    color: gray;
+    bottom: -10px;
```
mqp: Are these `bottom` rules even doing anything? I don't think that means anything outside of the context of elements with a specified `position`.

src/components/hand-poses.js
```diff
@@ -36,7 +36,13 @@ AFRAME.registerComponent("hand-pose", {
     this.from.play();
 
     const getNetworkedAvatar = el => {
-      const networkedAvatar = el.components["networked-avatar"];
+      if (!el) {
+        window.setTimeout(() => {
+          getNetworkedAvatar(this.el);
```
brianpeiris: What's this about? Why would `getNetworkedAvatar` call itself if it's just going to return `undefined` immediately?

src/systems/ticker.js
```diff
@@ -0,0 +1,23 @@
...
+    }
+    this.nextPromises.length = 0;
+  },
+
+  nextTick() {
+    return new Promise(resolve => {
+      this.nextPromises.push(resolve);
+    });
+  }
+});
```
mqp: Am I missing something, or does `nextTick` wait for the next next tick?

1. Initialize `promises` and `nextPromises` to be empty.
2. `nextTick` is called. `nextPromises` contains a resolver for the returned promise.
3. A tick happens. `nextPromises[0]` is moved into `promises[0]`.
4. Another tick happens. `promises[0]` is called and the promise resolves.

That's counterintuitive to me.

src/hub.js
```diff
@@ -364,13 +366,33 @@ async function handleHubChannelJoined(entryManager, hubChannel, messageDispatch,
...
       .then(() => {
         let newHostPollInterval = null;
 
+        let done = 0;
+        const i = window.setInterval(() => {
+          if (window.stuffToLoad === 0) {
+            done += 1;
+            if (done > 10) {
+              window.uiroot && window.uiroot.setState({ sceneObjectsLoaded: true });
+              window.clearInterval(i);
```
mqp: Why 10?

src/hub.js
```diff
@@ -364,13 +366,33 @@ async function handleHubChannelJoined(entryManager, hubChannel, messageDispatch,
...
+            } else {
+              const text = `Joining lobby...`;
+              window.uiroot && window.uiroot.setState({ loadingText: text });
+            }
+          } else {
+            done = 0;
+            const text = `Loading ${window.stuffToLoad} more object${window.stuffToLoad > 1 ? "s" : ""}...`;
+            window.uiroot && window.uiroot.setState({ loadingText: text });
+          }
+        }, 100);
```
mqp: I get the sense I would prefer this loading stuff to be in its own React component. hub.js is pretty crowded.

src/components/gltf-model-plus.js
brianpeiris: I think a less surprising way of doing this is to emit `loading-started` and `loading-ended` events, and then handle the incrementing/decrementing and UI updating in the react code instead, or are you really trying to avoid using events?

src/components/gltf-model-plus.js
mqp: It still kind of freaks me out that we're blowing away the not-frustum-culledness of anything that was not frustum culled. I would retain the previous setting on the objects (probably on an `{id: setting}` mapping in this code) and restore it afterward.

src/components/media-views.js
```diff
@@ -678,5 +697,13 @@ AFRAME.registerComponent("media-image", {
...
 
     this.el.emit("image-loaded", { src: this.data.src });
+
+    this.el.object3D.traverse(obj => {
+      obj.frustumCulled = false;
+    });
+    await AFRAME.scenes[0].systems.ticker.nextTick();
+    this.el.object3D.traverse(obj => {
+      obj.frustumCulled = true;
+    });
```
brianpeiris: Might be worth refactoring these lines into a single await call with a utility method, since we repeat it a few times.

src/components/gltf-model-plus.js
```diff
@@ -352,7 +359,7 @@ AFRAME.registerComponent("gltf-model-plus", {
         object3DToSet = this.inflatedEl.object3D;
         // TODO: Still don't fully understand the lifecycle here and how it differs between browsers, we should dig in more
         // Wait one tick for the appended custom elements to be connected before attaching templates
-        await nextTick();
+        await AFRAME.scenes[0].systems.ticker.nextTick();
```
brianpeiris: We should probably use `this.el.sceneEl` instead of `AFRAME.scenes[0]` where we can.

src/hub.js
mqp: (Once it becomes its own React component I will have suggestions about the state on that component and the interface for updating the loading text.)

src/hub.js
```diff
@@ -364,13 +366,33 @@ async function handleHubChannelJoined(entryManager, hubChannel, messageDispatch,
       debug: !!isDebug
     });
 
-    while (!scene.components["networked-scene"] || !scene.components["networked-scene"].data) await nextTick();
+    while (!scene.components["networked-scene"] || !scene.components["networked-scene"].data) {
+      await AFRAME.scenes[0].systems.ticker.nextTick();
```
brianpeiris: We already have a `scene` reference in this function.

src/assets/stylesheets/loader.scss
mqp: You should have three CSS classes, one called e.g. `status` which means "the thing positioned underneath the logo", which has the margin applied, and two more called `loaded` and `loading` which control the color. (Or, alternatively, have gray be the default color and a class called `finished` which applies the green color.) Then the element should always have the `status` class and it should choose which of the `loading`/`loaded`/`finished` classes it does or doesn't have at runtime based on what's going on.

src/scene-entry-manager.js
```diff
@@ -146,7 +151,7 @@ export default class SceneEntryManager {
     document.body.removeEventListener("touchend", requestFullscreen);
   };
 
-  _setupPlayerRig = () => {
+  _setupPlayerRig = async () => {
```
mqp: Seems not necessary.

src/react-components/ui-root.js
```diff
@@ -814,6 +819,10 @@ class UIRoot extends Component {
           </div>
 
           <img className="loading-panel__logo" src="../assets/images/hub-preview-light-no-shadow.png" />
+
+          <h4 className={this.state.loadingText.indexOf("lobby") !== -1 ? "loadedText" : "loadingText"}>
```
brianpeiris: Bit weird for this conditional to be based on the contents of the loading text. Would be better to add a boolean state field instead.

src/components/media-views.js
```diff
@@ -644,11 +656,18 @@ AFRAME.registerComponent("media-image", {
           return;
         }
       }
+      window.stuffToLoad -= 1;
     } catch (e) {
+      window.stuffToLoad -= 1;
```
mqp: More polite to put `finally`.

src/scene-entry-manager.js
```diff
@@ -78,7 +78,12 @@ export default class SceneEntryManager {
       this.playerRig.setAttribute("virtual-gamepad-controls", {});
     }
 
-    this._setupPlayerRig();
+    //    this._setupPlayerRig();
```
brianpeiris: delete

src/systems/ticker.js
johnshaughnessy: Yes it does wait for the next next aframe system tick, and the reason for that is because I expect most requests to come in from component updates/ticks, which come BEFORE this system's tick. However I realized this is the wrong approach. I was using this system's `tick` as the timer, which has a few problems:
- If I call nextFrame before aframe starts ticking the system (but after it exists), then the promise resolves before the first render.
- If I call nextFrame within a system, then the behavior is different depending on whether the system ticks before or after this one.
- If I call nextFrame from some async code, it's not clear whether it's going to wait 1 or 2 ticks.

Instead what I'll do is overwrite the scene's tick or the scene's render, because that's what calls the gl render which I need to ensure happens before these promises resolve.

src/hub.js
```diff
@@ -462,20 +463,21 @@ document.addEventListener("DOMContentLoaded", async () => {
...
-          navigator.serviceWorker.ready
-            .then(registration => subscriptions.setRegistration(registration))
-            .catch(() => subscriptions.setRegistrationFailed());
-        })
-        .catch(() => subscriptions.setRegistrationFailed());
-    } catch (e) {
-      subscriptions.setRegistrationFailed();
-    }
-  }
+  //if (navigator.serviceWorker) {
```
johnshaughnessy: This serviceWorker initialization seems to break entry on chrome. I don't know why. 
It seems that neither the `then` nor the `catch` attached to `navigator.serviceWorker.ready` is reached.

src/components/hand-poses.js
johnshaughnessy: This component is on our avatars before the `networked-avatar` component is attached, so if we go all the way up the ancestors and don't find one, we restart the search at `this.el` after 1 second.

johnshaughnessy: (The next iteration of `getNetworkedAvatar` won't _immediately_ return `undefined`. Note `this` in us calling `getNetworkedAvatar(this.el)`, not `getNetworkedAvatar(el)`.)

PR #876, opened 2019-02-06
Crush some pngs

mqp:
Nobody did this for a while, so I did it.


mqp:
They should look identical, PNGs are lossless. The optimization mechanism works by, uh, [magic](https://en.wikipedia.org/wiki/Pngcrush).

PR #882, opened 2019-02-08
Bug/scene transition issues

gfodor:
Fixes/updates two things w scene transitioning:

- Moves to the spawn point when switching scenes (probably can do something smarter later like distributing around that point)

- Fixes a bug where animations are retained if switching to a scene with no animations

### DIFF
--- a/src/components/gltf-model-plus.js
+++ b/src/components/gltf-model-plus.js
@@ -320,7 +320,7 @@ AFRAME.registerComponent("gltf-model-plus", {
       if (!src) {
         if (this.inflatedEl) {
           console.warn("gltf-model-plus set to an empty source, unloading inflated model.");
-          this.removeInflatedEl();
+          this.disposeLastInflatedEl();
         }
         return;
       }
@@ -332,7 +332,7 @@ AFRAME.registerComponent("gltf-model-plus", {
       if (src != this.lastSrc) return;
 
       // If we had inflated something already before, clean that up
-      this.removeInflatedEl();
+      this.disposeLastInflatedEl();
 
       this.model = gltf.scene || gltf.scenes[0];
       this.model.animations = gltf.animations;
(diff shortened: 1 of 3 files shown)


src/hub.js
```diff
@@ -317,8 +319,19 @@ async function updateEnvironmentForHub(hub) {
     THREE.Cache.clear();
 
     const onLoadingEnvironmentReady = () => {
-      environmentEl.setAttribute("gltf-model-plus", { src: sceneUrl });
       environmentEl.removeEventListener("model-loaded", onLoadingEnvironmentReady);
+
+      if (scene.is("entered")) {
+        // We've already entered, so move to new spawn point
```
gfodor: a somewhat nicer way to do this would be to fire an event, but then i'll need to revisit the way the event handling is happening in `spawn-controller`, which seemed like a potential rathole

src/components/gltf-model-plus.js
```diff
@@ -340,6 +340,8 @@ AFRAME.registerComponent("gltf-model-plus", {
       if (gltf.animations.length > 0) {
         this.el.setAttribute("animation-mixer", {});
         this.el.components["animation-mixer"].initMixer(gltf.animations);
+      } else {
+        this.el.removeAttribute("animation-mixer");
```
netpro2k: Probably makes sense to always do this cleanup in case the animation-mixer has any other state. Putting it in `removeInflatedEl` (maybe with a rename) seems like a good candidate, since that also gets called in the empty src case above

src/hub.js
```diff
@@ -317,8 +319,19 @@ async function updateEnvironmentForHub(hub) {
...
+        const spawnWhenEnvironmentReady = () => {
+          environmentEl.removeEventListener("model-loaded", spawnWhenEnvironmentReady);
```
netpro2k: You can use `{once: true}` on the `addEventListener` instead of dealing with this cleanup. That also lets you just inline the function since you dont need a ref to it anymore.

src/hub.js
```diff
@@ -278,6 +278,7 @@ async function updateEnvironmentForHub(hub) {
   let isLegacyBundle; // Deprecated
 
   const environmentScene = document.querySelector("#environment-scene");
+  const scene = document.querySelector("a-scene");
```
netpro2k: I might change this to `sceneEl` since `scene` is pretty overloaded in this function.

src/hub.js
gfodor: nice trick!

PR #885, opened 2019-02-12
Restore spawning animations

gfodor:
Restores springy spawning and loading cube appearance animations


robertlong:
Finally tested this out. I merged in #901 and tested spawners in Spoke scenes. The spawned object's scale was no longer preserved. It should spawn with the same scale as the spawner.

gfodor:
Ah, do you have a scene I can test with on dev?

robertlong:
https://localhost:8080/scene.html?scene_id=s81Suqs

PR #887, opened 2019-02-13
Upgrade webpack & friends, fix up minifier

mqp:
Self-explanatory. Shouldn't make much of a difference to anyone.


webpack.config.js
```diff
@@ -190,8 +189,6 @@ module.exports = (env, argv) => ({
   },
 
   optimization: {
-    // necessary due to https://github.com/visionmedia/debug/issues/547
-    minimizer: [new UglifyJsPlugin({ sourceMap: true, uglifyOptions: { compress: { collapse_vars: false } } })],
```
mqp: I mostly just did this so that we could get rid of this weird-looking hack. Nowadays it seems like people have forked off to a [better-maintained minifier](https://github.com/webpack-contrib/terser-webpack-plugin#terseroptions), which is the new Webpack default.

PR #888, opened 2019-02-13
Experimental trigger-volume component

robertlong:
This PR adds the `trigger-volume` component. This component is only enabled for use in Spoke when experimental features are enabled. **The API will probably change and those scenes may break. Use with caution!**

The trigger-volume component sets a property on a target component when an entity enters or leaves its bounding box.

I've exposed this component in the `gltf-component-mappings`. However, only two components `video` and `animation-loop` are exposed. And only their `paused` properties can be set. This is to reduce the surface area of the API for security sake and work on how we actually want this system to be implemented.

Note that the video's paused property is actually set through a `video-pause-state` component. This is a hack to get around the fact that calling `setAttribute("media-video", "videoPaused", true)` doesn't work for videos you own ([See here](https://github.com/mozilla/hubs/blob/master/src/components/media-views.js#L391)).

I've introduced the concept of a `propertyMapper` to filter out and map Spoke components/properties to Hubs components/properties. It's honestly a mess. I'd rather Spoke components and Hubs components match up one to one and mark individual components and their properties as public in their schemas.

Example (public means exposed. Not having the public property means private.):
```
AFRAME.GLTFModelPlus.exposeComponent("video");
AFRAME.registerComponent("video", {
  schema: {
    src: { type: "string" },
    paused: { type: "boolean", public: true },
    ...
  },
});
```

However, as we've seen with the inflators it's not that simple:
```
function mediaInflator(el, componentName, componentData, components) {
  if (components.networked) {
    el.setAttribute("networked", {
      template: componentData.controls ? "#static-controlled-media" : "#static-media",
      owner: "scene",
      persistent: true,
      networkId: components.networked.id
    });
  }

  const mediaOptions = {
    projection: componentData.projection
  };

  if (componentName === "video") {
    mediaOptions.videoPaused = !componentData.autoPlay;
    mediaOptions.volume = componentData.volume;
    mediaOptions.loop = componentData.loop;
    mediaOptions.audioType = componentData.audioType;

    if (componentData.audioType === "pannernode") {
      mediaOptions.distanceModel = componentData.distanceModel;
      mediaOptions.rolloffFactor = componentData.rolloffFactor;
      mediaOptions.refDistance = componentData.refDistance;
      mediaOptions.maxDistance = componentData.maxDistance;
      mediaOptions.coneInnerAngle = componentData.coneInnerAngle;
      mediaOptions.coneOuterAngle = componentData.coneOuterAngle;
      mediaOptions.coneOuterGain = componentData.coneOuterGain;
    }

    el.setAttribute("video-pause-state", { paused: mediaOptions.videoPaused });
  }

  el.setAttribute("media-loader", {
    src: componentData.src,
    resize: true,
    resolve: true,
    fileIsOwned: true,
    mediaOptions
  });
}

AFRAME.GLTFModelPlus.registerComponent("image", "image", mediaInflator);
AFRAME.GLTFModelPlus.registerComponent("video", "video", mediaInflator, (name, property, value) => {
  if (property === "paused") {
    return { name: "video-pause-state", property, value };
  } else {
    return null;
  }
});
```

I would like to merge this PR and work on improving this API over time. I think it will be hugely beneficial to design a robust system of exposing components and their properties to user generated content like scenes and eventually avatars and objects.

This sort of API can be used to enable adding basic interactivity to scenes. I think making our API surface area based on top of sanitized / scoped / permissioned calls to `el.setAttribute()` makes a lot of sense.

### DIFF
--- a/src/components/gltf-model-plus.js
+++ b/src/components/gltf-model-plus.js
@@ -87,7 +87,7 @@ function cloneGltf(gltf) {
 /// or templates associated with any of their nodes.)
 ///
 /// Returns the A-Frame entity associated with the given node, if one was constructed.
-const inflateEntities = function(node, templates, isRoot, modelToWorldScale = 1) {
+const inflateEntities = function(rootEl, node, templates, isRoot, modelToWorldScale = 1) {
   // TODO: Remove this once we update the legacy avatars to the new node names
   if (node.name === "Chest") {
     node.name = "Spine";
@@ -101,7 +101,7 @@ const inflateEntities = function(node, templates, isRoot, modelToWorldScale = 1)
   const childEntities = [];
   const children = node.children.slice(0); // setObject3D mutates the node's parent, so we have to copy
   for (const child of children) {
-    const el = inflateEntities(child, templates);
+    const el = inflateEntities(rootEl, child, templates);
     if (el) {
       childEntities.push(el);
     }
@@ -165,7 +165,7 @@ const inflateEntities = function(node, templates, isRoot, modelToWorldScale = 1)
     for (const prop in entityComponents) {
       if (entityComponents.hasOwnProperty(prop) && AFRAME.GLTFModelPlus.components.hasOwnProperty(prop)) {
         const { componentName, inflator } = AFRAME.GLTFModelPlus.components[prop];
-        inflator(el, componentName, entityComponents[prop], entityComponents);
+        inflator(el, componentName, entityComponents[prop], entityComponents, rootEl);
       }
     }
   }
@@ -345,7 +345,7 @@ AFRAME.registerComponent("gltf-model-plus", {
       let object3DToSet = this.model;
       if (
         this.data.inflate &&
-        (this.inflatedEl = inflateEntities(this.model, this.templates, true, this.data.modelToWorldScale))
+        (this.inflatedEl = inflateEntities(this.el, this.model, this.templates, true, this.data.modelToWorldScale))
       ) {
         this.el.appendChild(this.inflatedEl);
 
--- a/src/components/gltf-model-plus.js
+++ b/src/components/gltf-model-plus.js
@@ -8,7 +8,7 @@ THREE.Mesh.prototype.raycast = acceleratedRaycast;
 
 const GLTFCache = {};
 
-function inflateComponent(el, componentName, componentData) {
+function defaultInflator(el, componentName, componentData) {
   if (!AFRAME.components[componentName]) {
     throw new Error(`Inflator failed. "${componentName}" component does not exist.`);
   }
@@ -24,9 +24,9 @@ function inflateComponent(el, componentName, componentData) {
 AFRAME.GLTFModelPlus = {
   // eslint-disable-next-line no-unused-vars
   components: {},
-  registerComponent(componentKey, componentName, inflator, propertyMapper) {
-    inflator = inflator || inflateComponent;
-    AFRAME.GLTFModelPlus.components[componentKey] = { inflator, componentName, propertyMapper };
+  registerComponent(componentKey, componentName, inflator) {
+    inflator = inflator || defaultInflator;
+    AFRAME.GLTFModelPlus.components[componentKey] = { inflator, componentName };
   }
 };
 
@@ -81,13 +81,23 @@ function cloneGltf(gltf) {
   };
 }
 
+function getHubsComponents(node) {
+  const hubsComponents = node.userData.gltfExtensions && node.userData.gltfExtensions.HUBS_components;
+
+  // We can remove support for legacy components when our environment, avatar and interactable models are
+  // updated to match Spoke output.
+  const legacyComponents = node.userData.components;
+
+  return hubsComponents || legacyComponents;
+}
+
 /// Walks the tree of three.js objects starting at the given node, using the GLTF data
 /// and template data to construct A-Frame entities and components when necessary.
 /// (It's unnecessary to construct entities for subtrees that have no component data
 /// or templates associated with any of their nodes.)
 ///
 /// Returns the A-Frame entity associated with the given node, if one was constructed.
-const inflateEntities = function(rootEl, node, templates, isRoot, modelToWorldScale = 1) {
+const inflateEntities = function(indexToEntityMap, node, templates, isRoot, modelToWorldScale = 1) {
   // TODO: Remove this once we update the legacy avatars to the new node names
   if (node.name === "Chest") {
     node.name = "Spine";
@@ -101,19 +111,13 @@ const inflateEntities = function(rootEl, node, templates, isRoot, modelToWorldSc
   const childEntities = [];
   const children = node.children.slice(0); // setObject3D mutates the node's parent, so we have to copy
   for (const child of children) {
-    const el = inflateEntities(rootEl, child, templates);
+    const el = inflateEntities(indexToEntityMap, child, templates);
     if (el) {
       childEntities.push(el);
     }
   }
 
-  const hubsComponents = node.userData.gltfExtensions && node.userData.gltfExtensions.HUBS_components;
-
-  // We can remove support for legacy components when our environment, avatar and interactable models are
-  // updated to match Spoke output.
-  const legacyComponents = node.userData.components;
-
-  const entityComponents = hubsComponents || legacyComponents;
+  const entityComponents = getHubsComponents(node);
 
   const nodeHasBehavior = !!entityComponents || node.name in templates;
   if (!nodeHasBehavior && !childEntities.length && !isRoot) {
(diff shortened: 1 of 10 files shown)


src/components/trigger-volume.js
```diff
@@ -0,0 +1,64 @@
...
+    leaveProperty: { type: "string" },
+    leaveValue: {
+      default: "",
+      parse: v => (typeof v === "object" ? v : JSON.parse(v)),
+      stringify: JSON.stringify
+    }
+  },
+  init() {
+    this.size = new THREE.Vector3();
+    this.boundingBoxWorldPosition = new THREE.Vector3();
```
netpro2k: I like to only use instance properties for things that are actually carried over per frame and not just for these temp properties we create to avoid GC, for those I prefer using captured variables. This makes it easier to understand what is actually component state and has the added side benefit of sharing the temp variables across instances.

So I would keep `boundingBox` and `collidingLastFrame` as instance variables and make `boundingBox` and `colliderWorldPosition` into just `const`s outside the class (or if you like, captured in a closure on the function they are used in)

src/components/trigger-volume.js
```diff
@@ -0,0 +1,64 @@
...
+    const colliders = this.data.colliders;
+
+    this.el.object3D.getWorldPosition(this.boundingBoxWorldPosition);
+
+    for (let i = 0; i < colliders.length; i++) {
+      const collider = colliders[i];
+      const object3D = collider.object3D;
+
+      object3D.getWorldPosition(this.colliderWorldPosition);
+      const isColliding = this.boundingBox.containsPoint(this.colliderWorldPosition);
```
netpro2k: Hmm. Seems pretty not ideal to just be checking if the center point of the "collider" is in the bounding box here to determine collisions. Especially since we seem to be using the whole "player rig" as the default "collider". This means in VR you have to move the center of your playspace into the trigger volume to trigger it. Could also get pretty weird with fly mode. Seems like we should be using the physics system for this, since it already handles all this and even has collision enter/exit events (see `sticky-object-zone` https://github.com/mozilla/hubs/blob/master/src/components/sticky-object.js#L104)

src/gltf-component-mappings.js
```diff
@@ -181,3 +189,75 @@ AFRAME.GLTFModelPlus.registerComponent("spawner", "spawner", (el, componentName,
   });
   el.setAttribute("hoverable", "");
 });
+
+function propertyMapper(componentKey, componentProperty, componentValue) {
```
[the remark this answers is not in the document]
robertlong: Yeah I was trying to figure out what a generalized solution would look like and while I also don't like this, it's where I ended up.

I thought about using the inflators too. How would you go about it? I don't think a majority of our components will behave correctly if we call `setAttribute` on them. They'd have to handle things correctly in their update method.

src/gltf-component-mappings.js
```diff
@@ -181,3 +189,75 @@ AFRAME.GLTFModelPlus.registerComponent("spawner", "spawner", (el, componentName,
...
+      enterProperty,
+      enterValue,
+      leaveComponent,
+      leaveProperty,
+      leaveValue
+    } = componentData;
+
+    const enterAction = propertyMapper(enterComponent, enterProperty, enterValue);
+
+    if (!enterAction) {
```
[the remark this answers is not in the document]
robertlong: I disagree, these checks are for property sanitization. I'd like to keep them in one place so that if someone is editing the component they aren't confused and accidentally expose the wrong properties. See my point above about moving to a system that checks for a `public` attribute on the schema. I think that may be a better option for allowing you configure what properties are exposed in the component. I still think the system that checks that public properties are being set should live outside that component. Components should describe their public interfaces, but should not need to handle sanitization.

src/components/trigger-volume.js
robertlong: Oh yeah, this is going to break if I don't use the avatar position. Good catch 👍

Is the physics system really better here? I think checking for a single point inside a box collider is just about as fast as you can get.

src/gltf-component-mappings.js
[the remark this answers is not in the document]
netpro2k: This isn't about the public interface of the `trigger-volume` component, it is about that component caring about other components properties. Right now you could create a `trigger-volume` component directly and give it invalid properties since they are only checked by the inflator, so it makes sense to move that check upstream. As for the whole "public" property check, I wouldn't really bother with that personally, its not like we are doing an `eval` on the value or anything, so it doesnt scare me very much. I would rather find out novel ways people think to use things we didn't think of than manually have to whitelist things.

src/gltf-component-mappings.js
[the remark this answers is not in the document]
netpro2k: All I mean by "handle it in the inflators" is that any property remapping that has to happen from spoke -> hubs should happen in the inflator. 

In the case of `trigger-volume`, the fact that we want to have "media-video" "pause" get remapped to "video-pause-state" "paused" is a hack anyway so I would either hardcode that hack in `trigger-volume` or just actually set the real thing you want in spoke (which is to set the "paused" property on the "video-pause-state" component). Doesn't seem worth implementing a generic system just to support this hack.

As for:

>  I don't think a majority of our components will behave correctly if we call `setAttribute` on them. They'd have to handle things correctly in their update method.

That's partially true, though I would argue that's a bug anyway and we should fix it for the components we care to expose this way (though I don't particularly care much to make our components good aframe citizens anymore anyway which is why I might have just gone with an event or function call instead)

src/components/trigger-volume.js
robertlong: I think the only entity that I can use as the collider is the `player-camera`. `AvatarRoot` may not be there right?

netpro2k: Yeah, not suggesting physics here as a perf optimization but as a UX optimization. It might feel weird to have to get the center of my head inside the volume, but that might just be a matter of getting the volumes sized correctly. That said I think we are going to be computing these sorts of collisions anyway, so not sure it would be super expensive anyway.

robertlong: I'm not confident enough that moving to the physics system would provide much of a UX benefit. The scene creator will have to size their trigger volumes to account for the user's head, but that doesn't seem like a deal breaker. On the other hand, moving to the physics system will require physics a physics body on the avatar which I'm not certain I want to touch given that the new physics system is landing soon.

I think this whole feature needs another pass and a long incubation time before being exposed to users in Spoke. Can we ship with the simpler, more performant solution for the tutorial space and revisit in a few weeks after the physics system has landed?

src/components/video-pause-state.js
```diff
@@ -0,0 +1,40 @@
+// HACK: Setting videoPaused on media-video doesn't work if you are the owner.
+// This component gets around that by calling .togglePlaying() to play/pause the video based on a property change.
+AFRAME.registerComponent("video-pause-state", {
```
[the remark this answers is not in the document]
gfodor: i believe this is a backwards-compatibility breaking change for pinned, paused videos. if so, just wanted to flag it -- we can probably not worry about it but we should be mindful of these categories of schema migrations were we move state into new components -- the old data is effectively lost unless we migrate it.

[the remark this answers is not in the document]
gfodor: nevermind, looks like I am wrong about this

src/gltf-component-mappings.js
```diff
@@ -181,3 +189,75 @@ AFRAME.GLTFModelPlus.registerComponent("spawner", "spawner", (el, componentName,
...
+      );
+      return;
+    }
+
+    const leaveAction = propertyMapper(leaveComponent, leaveProperty, leaveValue);
+
+    if (!leaveAction) {
+      console.warn(
+        `GLTFModelPlus: leaveComponent "${leaveComponent}" on trigger-volume has an invalid property "${leaveProperty}". Skipping...`
+      );
```
gfodor: missing `return`

robertlong:
I've fixed a couple of the changes requested above by moving to a configuration object that stores public components, properties, and maps them to the correct components/properties/values. Also `trigger-volume`'s target property is now a glTF node index which is a unique identifier, unlike the the previous method which used the node name.

src/gltf-component-mappings.js
```diff
@@ -181,3 +190,74 @@ AFRAME.GLTFModelPlus.registerComponent("spawner", "spawner", (el, componentName,
   });
   el.setAttribute("hoverable", "");
 });
+
+const publicComponents = {
```
netpro2k: Nice, I think this is much better than adding an additional concept to all inflators.

src/components/gltf-model-plus.js
```diff
@@ -161,18 +165,30 @@ const inflateEntities = function(node, templates, isRoot, modelToWorldScale = 1)
...
+function inflateComponents(inflatedEntity, indexToEntityMap) {
+  inflatedEntity.object3D.traverse(object3D => {
+    const entityComponents = getHubsComponents(object3D);
+    const el = object3D.el;
+
+    if (entityComponents && el) {
+      for (const prop in entityComponents) {
+        if (entityComponents.hasOwnProperty(prop) && AFRAME.GLTFModelPlus.components.hasOwnProperty(prop)) {
+          const { componentName, inflator } = AFRAME.GLTFModelPlus.components[prop];
+          inflator(el, componentName, entityComponents[prop], entityComponents, indexToEntityMap);
```
netpro2k: Having this `indexToEntityMap` for inflators feels nice, that way all references inside the GLTF can now be index based, makes them feel much more "native" to the format.

test/unit/utils/component-mappings.test.js
```diff
@@ -0,0 +1,40 @@
+import test from "ava";
+import { getSanitizedComponentMapping } from "../../../src/utils/component-mappings";
+
+test("getSanitizedComponentMapping", t => {
```
netpro2k: Nice that we have the groundwork for writings some tests now. There are definitely a number of things we have that should be very easily testable.

src/components/gltf-model-plus.js
```diff
@@ -354,6 +392,9 @@ AFRAME.registerComponent("gltf-model-plus", {
         // Wait one tick for the appended custom elements to be connected before attaching templates
         await nextTick();
         if (src != this.lastSrc) return; // TODO: there must be a nicer pattern for this
+
+        inflateComponents(this.inflatedEl, indexToEntityMap);
```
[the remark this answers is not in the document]
robertlong: Yeah, I did some testing and couldn't find anything that broke yet, but I'm going to watch out for it. Let's see how the smoke test goes 🤞

PR #896, opened 2019-02-16
Feature/chatbox ux mobile

gfodor:
Chat box is updated to now be in the center of the screen and also is now available on mobile. You can now spawn objects from the top HUD in 2D.

image

Builds on https://github.com/mozilla/hubs/pull/890

This PR also:

- Adds some paths that allow detection of N finger taps and also toggling freeze mode

- Makes it so a two-finger tap gesture toggles pause mode on mobile since there is no more pause button (there will be onboarding tooltips in a subsequent PR)

- Adds a file upload or image upload button to the chat box

- Hides the joysticks on portrait orientation, hides chatbox and shows joysticks on landscape


src/components/virtual-gamepad-controls.css
```diff
@@ -19,10 +23,14 @@
...
   right: 0;
-  bottom: 10vh;
+  bottom: 15vh;
   display: flex;
   align-items: center;
   justify-content: space-around;
+
+  @media (orientation: portrait) {
+    display: none;
+  }
```
brianpeiris: I would much prefer if we kept the joysticks in portrait mode, if we could.

src/react-components/ui-root.js
```diff
@@ -1356,10 +1382,42 @@ class UIRoot extends Component {
...
+                    className={classNames([
+                      styles.messageEntryButton,
+                      styles.messageEntryButtonInRoom,
+                      styles.messageEntrySubmit
+                    ])}
+                  >
+                    <i>
+                      <FontAwesomeIcon icon={faPaperPlane} />
+                    </i>
+                  </button>
```
brianpeiris: Isn't it better to put the send button before the upload button? I would also suggest using the same cloud icon we use in the media browser.

firefox_2019-02-19_16-19-34

src/react-components/ui-root.js
```diff
@@ -1338,7 +1360,11 @@ class UIRoot extends Component {
                     ])}
                     value={this.state.pendingMessage}
                     rows={textRows}
-                    onFocus={e => e.target.select()}
+                    onFocus={e => {
+                      this.setState({ messageEntryOnTop: isMobile });
+                      e.target.select();
+                    }}
+                    onBlur={() => this.setState({ messageEntryOnTop: false })}
```
brianpeiris: This is buggy for me in Chrome for Android, the text box loses focus immediately after you tap on it and it moves to the top.

brianpeiris: Firefox for Android also gets into a bad state sometimes, where it doesn't exit/re-enter fullscreen mode correctly:

image

src/assets/stylesheets/ui-root.scss
```diff
@@ -228,34 +228,58 @@ body.vr-mode {
...
+  color: $action-color;
+}
+
+:local(.message-entry-submit) {
+  right: 15px;
+}
+
+:local(.message-entry-upload) {
+  margin-bottom: 1px;
+  right: 55px;
```
brianpeiris: Would also suggest using regular flex here if we can, instead of rearranging elements with `right`.

src/components/virtual-gamepad-controls.css
gfodor: I can see how it feels, but I figured it would be weird given the chatbox collides with it. We'd at least have to hide the preview sticks.

src/react-components/ui-root.js
gfodor: actually now that I think about it, i should reorder these completely. the reason for the + was to match slack and discord. i think the right layout would be:

```
[+ Send a message [spawn] [send]]
```

src/react-components/ui-root.js
gfodor: ah I forgot to test chrome on android, thanks for checking!

gfodor: weird couldn't get my local working on android. going to push to smoke and hope for the best, i think i figured this out :)

PR #898, opened 2019-02-17
Feature/hover freeze menus

gfodor:
This PR changes the menus to appear on mouse over, and also accents the hover effect to make it obvious when holding pause mode on that you should mouse over some objects.

TODO: need to test this out in VR and mobile


johnshaughnessy:
Pretty awkward on mobile touchscreen. I think it could be made somewhat better if we move the cursor to the center of your screen whenever you aren't touching down, or somehow was just smarter about the cursor on touchscreen in general. The problems right now are 1) it's not obvious that the user should touch one of the glowing objects to show a menu, and when they do, they might immediately (and repeatedly) click the button under their finger. 2) As soon as the user touches the object (to show the menu), they end up moving / rotating the object, usually in a surprising way, because the constraint seems to be created before the cursor is on the object, which means the object is told by the physics system to move and rotate.

gfodor:
thanks for trying that. i'll hold this open until we get a better design for mobile

gfodor:
i'm pushing this so it just happens on non-mobile

PR #905, opened 2019-02-18
Add room rename dialog and clean up settings menu

gfodor:
You can now rename the room if you have permission via the "Set Room Name" menu option.

image

Builds on https://github.com/mozilla/hubs/pull/904

This PR also updates the styling of the settings menu to break it up into sections, and also masks out user interface elements that require hub update permissions.

### DIFF
--- a/src/react-components/rename-room-dialog.js
+++ b/src/react-components/rename-room-dialog.js
@@ -1,0 +1,50 @@
+import React, { Component } from "react";
+import PropTypes from "prop-types";
+import { FormattedMessage } from "react-intl";
+
+import styles from "../assets/stylesheets/rename-room-dialog.scss";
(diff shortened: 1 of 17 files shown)


src/react-components/rename-room-dialog.js
```diff
@@ -0,0 +1,50 @@
...
+    return (
+      <DialogContainer title="Set Room Name" {...this.props}>
+        <form onSubmit={this.onSubmit} className={styles.renameRoomForm}>
+          <span>
+            <FormattedMessage id="rename-room.message" />
+          </span>
+          <input
+            name="name"
+            type="text"
+            required
```
brianpeiris: This should include name length and character validation.

PR #910, opened 2019-02-20
Update IK controller to skip unneeded work

gfodor:
This PR reduces the work needed to perform IK updates every frame. It has the following optimizations:

- If an avatar is out of view, we only run IK on it via the frame scheduler. We update the `this.isInView` bit for one avatar per frame by doing a frutum check against the player camera and (if available) mirror camera. Note that this means that the in-game camera viewport camera is not included in the set of cameras we care about, which is fine because it has a low frame rate anyway.

- If an avatar's head position has not moved and no further movement on the rest of the avatar is expected, we avoid running an IK step.

### DIFF
--- a/src/components/ik-controller.js
+++ b/src/components/ik-controller.js
@@ -1,4 +1,14 @@
 const { Vector3, Quaternion, Matrix4, Euler } = THREE;
+
+function quaternionAlmostEquals(epsilon, u, v) {
+  return (
+    (Math.abs(u.x - v.x) < epsilon || Math.abs(-u.x - v.x) < epsilon) &&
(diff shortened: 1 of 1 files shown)


src/components/ik-controller.js
```diff
@@ -218,5 +267,38 @@ AFRAME.registerComponent("ik-controller", {
...
+      frustum.setFromMatrix(frustumMatrix);
+      return frustum.containsPoint(pos);
+    };
+
+    return function() {
+      // Take into account the mirror camera if it is enabled.
+      const mirrorCamera = this.cameraMirrorSystem.mirrorCamera;
+
+      const camera = this.ikRoot.camera.object3D;
+      camera.updateMatrices(true, true);
```
netpro2k: doesn't `getWorldPosition` take care of this for you?

PR #912, opened 2019-02-20
Upgrade Babel and related things

mqp:
Self-explanatory. It's nice to be up to date. It doesn't seem like there were any difficult changes involved.


mqp:
Don't ask me why our stuff shrunk as much as it did, I guess I'm not complaining though.

PR #925, opened 2019-02-22
Fix up a bunch of layout and full screen issues on mobile

gfodor:
This PR deals with a number of ugly issues discovered around UX on firefox mobile:

- A bunch of flexbox constraints were wrong, using `fit-content` and also `width: 100%` -- the former is just straight up unsupported and the latter seemed to be computing the wrong min width. I worked around these by using `100vw` and setting `min-heights` and explicit `heights`. (I feel like this isn't really right, but it seems to result in the right layout.)

- Fixes a bunch of full screen transition issues. The main thing that I've encountered is that Firefox Android, when dealing with focused input fields, can in certain cases (like when we auto focus I think, or when the user uses the copy/paste functionality) exit full screen but no longer be able to re-enter it even if we request it. The result is that we need to exit full screen explicitly when interacting with text boxes, and re-enter it when that interaction has concluded.

- Auto-select on focus is kind of weird on mobile because it pops up the copy/paste options, so I stopped doing it.

### DIFF
--- a/src/assets/stylesheets/audio.scss
+++ b/src/assets/stylesheets/audio.scss
@@ -150,6 +150,8 @@
     background: none;
     border: none;
     cursor: pointer;
+    width: 111px;
+    height: 111px;
   }
 
   &__next-container {
--- a/src/assets/stylesheets/media-browser.scss
+++ b/src/assets/stylesheets/media-browser.scss
@@ -121,17 +121,17 @@
   :local(.header) {
     display: flex;
     align-items: flex-start;
-    width: 100%;
+    width: 100vw;
     margin: 32px;
     margin-bottom: 24px;
-    min-height: fit-content;
+    height: 77px;
   }
 
   :local(.nav) {
-    width: 100%;
+    width: 100vw;
     display: flex;
     position: relative;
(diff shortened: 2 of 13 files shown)


brianpeiris:
Tried this PR on my Galaxy S8 on FxA and ran into the following issues:
1. Tapping on the message entry exits fullscreen, but does not focus the textarea, so you have to tap on the message entry again.
2. Tapping on the message entry after exiting fullscreen causes the text input to focus, but it also disappears, perhaps hiding it behind the address bar when my keyboard pops open.
3. Even attempting to type blind doesn't work since the text input is cleared when I dismiss the keyboard
4. Hubs does not exit fullscreen when focusing the text input inside the "create object" dialog, so the phone's statusbar and/or addressbar cover the top of the app.
5. You can tap on the message entry text input, then tap on the create object button, focus the inner text input and you are left in a state where the app does not re-enter fullscreen

gfodor:
thanks for testing. #1 is unavoidable due to the bug/issue i mentioned where firefox enters a "half full screen" mode it seems when a text field is focused. we can't force re-focus without a user gesture. will look into the rest.

PR #929, opened 2019-02-22
Feature/settings links v2

gfodor:
Updates the links to include some missing ones, and adds "New Room" as a top line option.

image


gfodor:
Ah good point. I think keeping your mic hot if you hit that is wrong, so i'll make it open in the same tab.

PR #930, opened 2019-02-22
Prevent compileAndUpload from continuing to run

gfodor:
It looks like compileAndUploadMaterials keeps getting called throughout a session, I'm guessing this in unintentional, not sure how problematic it is.

### DIFF
--- a/src/react-components/ui-root.js
+++ b/src/react-components/ui-root.js
@@ -335,10 +335,14 @@ class UIRoot extends Component {
 
     if (this.loadingTimeout) window.clearTimeout(this.loadingTimeout);
     this.loadingTimeout = window.setTimeout(() => {
-      if (!this.props.isBotMode) {
-        AFRAME.scenes[0].renderer.compileAndUploadMaterials(AFRAME.scenes[0].object3D, AFRAME.scenes[0].camera);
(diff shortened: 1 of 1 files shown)


src/react-components/ui-root.js
```diff
@@ -335,10 +335,14 @@ class UIRoot extends Component {
 
     if (this.loadingTimeout) window.clearTimeout(this.loadingTimeout);
     this.loadingTimeout = window.setTimeout(() => {
-      if (!this.props.isBotMode) {
-        AFRAME.scenes[0].renderer.compileAndUploadMaterials(AFRAME.scenes[0].object3D, AFRAME.scenes[0].camera);
+      if (!this.state.hideLoader) {
```
johnshaughnessy: There's a race condition here since `setState` is async. I am bringing my branch up to speed where there's this same fix but flagging that you've been to this code path at least once is not done with react state.

johnshaughnessy: I'm fixing merge conflicts with that branch and HEAD at the moment. Will post in a bit.

johnshaughnessy:
I fixed this issue in https://github.com/mozilla/hubs/pull/921/files#diff-706aa9bb80768a301061a401774b665fR215

PR #933, opened 2019-02-22
Add support for ETC2 compressed textures (WIP)

robertlong:
To Do:
- [x] Add ETC2 support to Three.js
- [x] Add ETC2 support to PVRLoader
- [x] Add WebGL2 support to AFrame
- [x] Rewrite AFrame's builtin shaders to GLSL ES 3.0 at runtime if WebGL2 is enabled
- [x] Add PVRLoader to `media-image` component in Hubs
- [x] Upload images with the original content type so we can test PVR textures (Can be removed later)
- [x] Merge AFrame WebGL2 PR into the hubs/dev-9.0 branch
- [x] Merge Three.js ETC2 support PR into the hubs/master branch
- [ ] Proxy images through the new compressed images service @mqp is working on (Only if extension is supported on the target platform)


package.json
```diff
@@ -79,7 +79,7 @@
     "react-youtube": "^7.8.0",
     "screenfull": "^3.3.2",
     "super-hands": "github:mozillareality/aframe-super-hands-component#feature/drawing",
-    "three": "github:mozillareality/three.js#hubs/master",
+    "three": "github:mozillareality/three.js#7c0164305625dabe7e89746109e19661711543f3",
```
johnshaughnessy: perhaps should land this onto `hubs/master`

robertlong:
We eventually plan to add support for KTX2 and Basis, not EC2

PR #934, opened 2019-02-23
Links to scenes

gfodor:
You can now paste in links to scenes and the room owner can change to that scene via the menu.

image

This PR also adds a missing referenced overload to history.js.

### DIFF
--- a/src/scene-entry-manager.js
+++ b/src/scene-entry-manager.js
@@ -428,7 +428,7 @@ export default class SceneEntryManager {
     this.scene.addEventListener("action_selected_media_result_entry", e => {
       // TODO spawn in space when no rights
       const entry = e.detail;
-      if (entry.type === "scene_listing") return;
+      if (entry.type === "scene_listing" && this.hubChannel.permissions.update_hub) return;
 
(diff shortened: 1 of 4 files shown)


gfodor:
v2 will be a mouse over menu for these kinds of links, that will also check permissions. right now this just shows for anyone even those who do not have permission to do it.

src/scene-entry-manager.js
```diff
@@ -428,7 +428,7 @@ export default class SceneEntryManager {
     this.scene.addEventListener("action_selected_media_result_entry", e => {
       // TODO spawn in space when no rights
       const entry = e.detail;
-      if (entry.type === "scene_listing") return;
+      if (entry.type === "scene_listing" && this.hubChannel.permissions.update_hub) return;
```
gfodor: this logic here will make it so users who do not have permission to change the scene will paste links to scenes into the room when using the scene browser, vs having nothing happen at all, which is what happens now.

gfodor:
closed in favor of https://github.com/mozilla/hubs/pull/936

PR #936, opened 2019-02-25
Change scenes and rooms using links

gfodor:
Pasting room and scene links into a room will now give you an overlay menu to visit the room or change to the scene.

image

This PR also fixes an issue where we were not setting up the box menu computation right on 2d objects and fixes up a warning.


src/hub.js
```diff
@@ -505,6 +510,10 @@ document.addEventListener("DOMContentLoaded", async () => {
 
   const authChannel = new AuthChannel(store);
   const hubChannel = new HubChannel(store);
+
+  // hacky - we want to access permissions from throughout our A-Frame behaviors and code
+  window.APP.getPermissions = () => hubChannel.permissions;
```
brianpeiris: FYI, we already have access to the `hubChannel` on `APP` and we also have an A-Frame `permissions` system that exposes a `can` method: https://github.com/mozilla/hubs/blob/master/src/systems/permissions.js
Might be nice to reuse a similar pattern for our other JS code. Maybe just have a `window.APP.permissions.can()` method.

src/utils/media-utils.js
```diff
@@ -334,3 +334,8 @@ export function injectCustomShaderChunks(obj) {
 export function getPromotionTokenForFile(fileId) {
   return window.APP.store.state.uploadPromotionTokens.find(upload => upload.fileId === fileId);
 }
+
+const hubsSceneRegex = /https?:\/\/(hubs.local(:\d+)?|(smoke-)?hubs.mozilla.com)\/scenes\/(\w+)\/?\S*/;
+const hubsRoomRegex = /https?:\/\/(hubs.local(:\d+)?|(smoke-)?hubs.mozilla.com)\/(\w+)\/?\S*/;
```
netpro2k: Seems like we should restrict this a bit. Our hub ids are of a known length, so enforcing that might help cut down false positives a bit.

src/components/media-loader.js
```diff
@@ -190,14 +197,29 @@ AFRAME.registerComponent("media-loader", {
...
         this.el.removeAttribute("media-video");
-        this.el.addEventListener("image-loaded", this.onMediaLoaded, { once: true });
         this.el.removeAttribute("media-pager");
+        this.el.addEventListener(
+          "image-loaded",
+          () => {
+            const mayChangeScene = !!window.APP.getPermissions().update_hub;
+
+            if (isHubsRoomUrl(src) || (isHubsSceneUrl(src) && mayChangeScene)) {
+              this.el.setAttribute("hover-menu__hubs-item", {
```
netpro2k: technically if this `media-loader` changes type this will stick around, this is also true of the video menus though. Probably not for this PR but I vote we just stop pretending media-loaders can change type and just throw an error if it tries to. That allows us to delete a lot of the ugly `removeAttribute` code littered about in this `update` function.

src/components/open-media-button.js
```diff
@@ -1,18 +1,27 @@
...
     this.updateSrc = () => {
       this.src = this.targetEl.components["media-loader"].data.src;
-      this.label.setAttribute("text", "value", isHubsUrl(this.src) ? "visit" : "open link");
+
+      let label = "open link";
+
+      if (isHubsSceneUrl(this.src)) {
+        label = "use scene";
+      } else if (isHubsRoomUrl(this.src)) {
+        label = "visit";
```
netpro2k: We should maybe say "visit room" or "visit hub"

src/hub.js
gfodor: didn't know about the system, will refactor

src/utils/history.js
```diff
@@ -45,6 +45,10 @@ export function pushHistoryState(history, k, v) {
   pushOrUpdateHistory(history, false, k, v);
 }
 
+export function replaceHistoryState(history, k, v) {
```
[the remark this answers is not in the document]
gfodor: `state-link` uses it, we have a warning on master bc I forgot to include it (though the code path afaik is never exercised)

src/utils/media-utils.js
gfodor: unfortunately we have some legacy ones that are longer :(

i'll see if i can tighten this up at all though (on the next branch, which moves these utilities somewhere else)

src/components/open-media-button.js
gfodor: yep good point, since right now it's not clear it's a room

src/utils/media-utils.js
gfodor: i'm a bit nervous about this being future proofed if i set the length. there's a lot of coupling with the reticulum routing logic. main issue here is going to be false positives, but the only side effect is that it will say "visit room" instead of "open url" -- so nbd for now.

PR #937, opened 2019-02-25
Show lift HMD warning when pinning

gfodor:
In desktop VR we were previously calling `exitVR` when the user clicked on "pin" and they weren't signed in. This PR updates this flow to show an in-VR prompt telling the user to lift their headset to continue.

### DIFF
--- a/src/components/follow-in-lower-fov.js
+++ b/src/components/follow-in-lower-fov.js
@@ -2,7 +2,8 @@ AFRAME.registerComponent("follow-in-lower-fov", {
   schema: {
     target: { type: "selector" },
     offset: { type: "vec3" },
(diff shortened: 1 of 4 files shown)


src/components/follow-in-lower-fov.js
```diff
@@ -2,7 +2,8 @@ AFRAME.registerComponent("follow-in-lower-fov", {
   schema: {
     target: { type: "selector" },
     offset: { type: "vec3" },
-    speed: { type: "number", default: 0.003 }
+    speed: { type: "number", default: 0.003 },
+    angle: { type: "number", default: Math.PI / 4 }
```
brianpeiris: Would be nice to allow specifying this in degrees and convert to radians in `update`

PR #940, opened 2019-02-25
Camera snapshot button

gfodor:
You can now take photos at a distance using the camera snap button. The camera also no longer locks to your cursor when using a mouse.

image

This PR also fixes up the in-game action button color.


src/hub.html
```diff
@@ -46,6 +46,7 @@
     >
         <a-assets>
             <img id="rotation-symbol" crossorigin="anonymous" src="./assets/rotation.png">
+            <img id="button" crossorigin="anonymous" src="./assets/hud/button.9.png">
             <img id="action-button" crossorigin="anonymous" src="./assets/hud/action_button.9.png">
```
netpro2k: Weird, was this just broken before? Seems like this PR is adding the image but the reference was already there.

src/hub.html
```diff
@@ -335,6 +338,19 @@
                 </a-entity>
             </template>
 
+            <template id="camera-hover-menu">
+                <a-entity class="ui interactable-ui hover-container camera-snap-menu" mixin="button-container">
+                    <a-entity mixin="rounded-text-action-button" snap-camera-button position="0 -0.145 0.0" scale="0.75 0.75 0.75">
+                        <a-entity text="value:; width:1; align:center;" text-raycast-hack position="0 0 0.02"></a-entity>
+                        <a-image
+                            icon-button="image: #snap-camera; hoverImage: #snap-camera;"
```
netpro2k: You created a hover image but didn't add it to assets or use it here.

src/components/snap-camera-button.js
```diff
@@ -0,0 +1,8 @@
+AFRAME.registerComponent("snap-camera-button", {
```
netpro2k: Since these are so tightly coupled anyway I would just have `camera-tool` handle this instead and ditch this component. This is how `media-video` handles its overlay.

src/hub.html
```diff
@@ -46,6 +46,7 @@
     >
         <a-assets>
             <img id="rotation-symbol" crossorigin="anonymous" src="./assets/rotation.png">
+            <img id="button" crossorigin="anonymous" src="./assets/hud/button.9.png">
```
netpro2k: The 4 images added in this PR are all non power of 2 so will get resized at runtime. The ".9" ones in particular are only 1px off in each dimension.

src/components/position-at-box-shape-border.js
```diff
@@ -59,6 +59,8 @@ AFRAME.registerComponent("position-at-box-shape-border", {
   tick() {
     if (!this.target) {
       this.targetEl = this.el.querySelector(this.data.target);
```
netpro2k: Hmm, I know this wasnt just added but we really shouldn't be doing a querySelector on every tick

src/components/position-at-box-shape-border.js
```diff
@@ -59,6 +59,8 @@ AFRAME.registerComponent("position-at-box-shape-border", {
...
+      if (!this.targetEl) return;
```
netpro2k: Hmm, I assume this is to deal with async targets, but it feels wrong that we would just be silently ignoring this. Combined with my above comment about running querySelector in tick, seems like the contract should be that its up to whatever is adding the `position-at-box-shape-border` to ensure that target exists before adding this component, then the `targetEl` can just be resolved in `init`

src/hub.html
gfodor: no, this is a new image and new reference I added to make the buttons more pink

src/hub.html
gfodor: ah that image is dead code, will remove

src/components/position-at-box-shape-border.js
gfodor: `this.target` will eventually converge to non-null (assuming the author wrote a sane selector)

src/components/position-at-box-shape-border.js
gfodor: these are mostly in `hub.html`, i'll see if i can diagnose what caused this to start breaking for this specific feature

src/hub.html
gfodor: actually specifically i updated `action_button.9.png` to be pink, and now the `button` mixin references this `img` which is the old, grey button background.

src/components/position-at-box-shape-border.js
gfodor: I couldn't repro this, might have been an issue on my home machine. I'll add a warning and revisit later if I see it hitting again.

PR #941, opened 2019-02-25
Fix pen menu from clipping and also recenter to box

gfodor:
This PR fixes up the pen freeze menu to be positioned properly and also no longer be clipped inside of the pen.


gfodor:
simple change, pushing (feel free to review post-commit)

PR #942, opened 2019-02-25
Rename align to recenter and clean up layout

gfodor:
This PR renames "align" to "re-center" on the menu and also fixes up the camera rotation/re-center buttons. The re-center button on the camera was also not working because of the mode being wrong (I'm guessing the mode was renamed and this part was not refactored.)


gfodor:
this is simple enough, pushing

PR #946, opened 2019-02-26
Disable frustum culling on the avatar

gfodor:
For some still unknown reason it is now easy to accidentally get the avatar body to intersect the near clipping plane of the camera frustum, which results in the entire avatar being clipped out. This PR changes things so that the player's avatar is never subject to frustum culling.


gfodor:
Also I checked, and the near plane on the player camera did not change -- but something about the eye camera frustums may have.

PR #947, opened 2019-02-26
Rate limit camera to 1 shot per second

gfodor:
Button mashing the camera button (or, in the case of mobile, having the grab events fire a few times when you tap, which is weird) causes a bunch of photos to be taken. Rate limiting it seems sane enough.


gfodor:
eh, closing this. i fixed the underlying cause in the touch screen controls in https://github.com/mozilla/hubs/pull/976

PR #948, opened 2019-02-26
Add camera icon to focus and track buttons

gfodor:
Adds a camera icon to focus + track to make it more obvious they are related to the camera in the scene.

image


gfodor:
this is small enough change, feel free to review post-commit

PR #951, opened 2019-02-26
Add scene change dialog

gfodor:
Adds a dialog to set to a custom scene from the scene picker.

image

### DIFF
--- a/src/assets/stylesheets/change-scene-dialog.scss
+++ b/src/assets/stylesheets/change-scene-dialog.scss
@@ -13,11 +13,8 @@
   @extend %bottom-action-button;
   margin-left: 6px;
   margin-right: 6px;
-  appearance: none;
   width: 128px;
   text-align: center;
-  -moz-appearance: none;
-  -webkit-appearance: none;
 }
 
 :local(.buttons) {
@@ -32,7 +29,7 @@
   display: flex;
   @extend %rounded-border;
   @extend %default-font;
-  margin: 1em;
+  margin: 12px 0px;
   padding: 0.5em 0.75em;
   width: 100%;
   box-sizing: border-box;
--- a/src/assets/stylesheets/change-scene-dialog.scss
+++ b/src/assets/stylesheets/change-scene-dialog.scss
@@ -1,6 +1,6 @@
 @import 'shared';
 
-:local(.add-media-form) {
+:local(.change-scene-button) {
   display: flex;
   flex-direction: column;
   align-items: center;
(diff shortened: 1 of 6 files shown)


src/assets/stylesheets/change-scene-dialog.scss
```diff
@@ -0,0 +1,53 @@
...
+:local(.action-button) {
+  @extend %unselectable;
+  @extend %bottom-action-button;
+  margin-left: 6px;
+  margin-right: 6px;
+  appearance: none;
+  width: 128px;
+  text-align: center;
+  -moz-appearance: none;
+  -webkit-appearance: none;
```
brianpeiris: `%action-button` already sets appearance to none.

src/react-components/change-scene-dialog.js
```diff
@@ -0,0 +1,90 @@
...
+                  placeholder="Scene URL"
+                  onFocus={e => handleTextFieldFocus(e.target)}
+                  onBlur={() => handleTextFieldBlur()}
+                  onChange={this.onUrlChange}
+                  type="url"
+                  value={this.state.url}
+                />
+              </div>
+              <div className={styles.buttons}>
+                <WithHoverSound>
```
brianpeiris: I think the plan was to remove these `WithHoverSound` wrappers entirely at some point, since we're not actually playing audio on 2D UI anymore, so probably best not to add more of them.

src/react-components/change-scene-dialog.js
```diff
@@ -0,0 +1,90 @@
...
+
+    if (this.state.url) {
+      this.props.onChange(this.state.url);
+    }
+
+    this.props.onClose();
+  };
+
+  render() {
+    const { onChange, onClose, ...other } = this.props; // eslint-disable-line no-unused-vars
```
brianpeiris: I dunno how I feel about this catch-all props pattern. I'd rather be explicit where we can, as is the case with this dialog. For one thing, we probably shouldn't encourage use of `eslint-disable-line`.

src/react-components/media-browser.js
```diff
@@ -195,6 +196,7 @@ class MediaBrowser extends Component {
     const urlSource = searchParams.get("media_source") || sluglessPath(this.props.history.location).substring(7);
     const apiSource = this.state.result && this.state.result.meta.source;
     const isVariableWidth = this.state.result && ["bing_images", "tenor"].includes(apiSource);
+    const showCustomOption = apiSource !== "scene_listings" || this.props.hubChannel.permissions.update_hub;
```
brianpeiris: Maybe not in this PR but I was thinking we should have more explicit API for permissions check. A `permissions.can("update_hub")` method would allow us to validate permission names and warn on typos or name changes. (though I suppose we could use a Proxy if we wanted to get fancy)

brianpeiris: Also wonder if we should just start using a `window.APP.permissions` global at this point, instead of continuing to pass a hubChannel prop everywhere.

src/react-components/change-scene-dialog.js
gfodor: i've been just 'going with the flow' assuming there will be some monster PR ripping these out eventually :) i will remove the ones I added though.

src/react-components/change-scene-dialog.js
gfodor: ah this was copied from elsewhere. i'll remove it though

src/react-components/media-browser.js
gfodor: i was using the A-Frame system outside of React and using hubChannel -- I agree a global API may make sense.

PR #952, opened 2019-02-26
Feature/content search last search restore

gfodor:
This PR introduces a "stash" of the previous query/page/filter from the media browser, so if you re-open it, it picks up where you last left off. (And if you switch to another tab using the hotkeys, it retains the query.)

### DIFF
--- a/src/react-components/media-browser.js
+++ b/src/react-components/media-browser.js
@@ -125,12 +125,8 @@ class MediaBrowser extends Component {
       clearTimeout(this._sendQueryTimeout);
     }
 
-    const lastQuery = query;
-
     // Don't update search on every keystroke, but buffer for some ms.
     this._sendQueryTimeout = setTimeout(() => {
-      if (query !== lastQuery) return;
-
       // Drop filter for now, so entering text drops into "search all" mode
       this.props.mediaSearchStore.filterQueryNavigate("", query);
     }, 500);
--- a/src/react-components/media-browser.js
+++ b/src/react-components/media-browser.js
@@ -125,8 +125,12 @@ class MediaBrowser extends Component {
       clearTimeout(this._sendQueryTimeout);
     }
 
+    const lastQuery = query;
+
     // Don't update search on every keystroke, but buffer for some ms.
     this._sendQueryTimeout = setTimeout(() => {
+      if (query !== lastQuery) return;
+
       // Drop filter for now, so entering text drops into "search all" mode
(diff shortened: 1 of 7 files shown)


src/react-components/media-browser.js
```diff
@@ -125,8 +125,12 @@ class MediaBrowser extends Component {
       clearTimeout(this._sendQueryTimeout);
     }
 
+    const lastQuery = query;
+
     // Don't update search on every keystroke, but buffer for some ms.
     this._sendQueryTimeout = setTimeout(() => {
+      if (query !== lastQuery) return;
```
brianpeiris: Pretty sure this is impossible, since the value of `query` is captured by the closure, it should always remain equal to lastQuery. Did you mean to compare `lastQuery` to `this.state.query`? Even then, the `clearTimeout` should make that impossible.

src/storage/media-search-store.js
```diff
@@ -113,33 +115,68 @@ export default class MediaSearchStore extends EventTarget {
...
     searchParams.delete("media_nav");
 
+    if (!keepSource) {
+      searchParams.delete("media_source");
+    }
+
     return searchParams;
   };
 
+  stashLastSearchParams = location => {
```
brianpeiris: Probably should be `_stashLastSearchParams`

src/storage/media-search-store.js
```diff
@@ -113,33 +115,68 @@ export default class MediaSearchStore extends EventTarget {
...
+    }
+  };
+
   sourceNavigateToDefaultSource = () => {
-    this.sourceNavigate(SOURCES[0]);
+    this.sourceNavigate(this._stashedSource ? this._stashedSource : SOURCES[0], false, true);
   };
 
-  sourceNavigate = (source, hideNav) => {
+  sourceNavigate = (source, hideNav, useLastStashedParams) => {
```
brianpeiris: Maybe make this an internal `_sourceNavigate` method and make public `sourceNavigateHiddenNavWithStash` and `sourceNavigateWithNav` methods or use a `params` object.

src/react-components/media-browser.js
gfodor: ah yes, I was trying to resolve a fairly hard to reproduce issue where the state.value was getting clobbered, in part i thought because there was some bug in logic here. you're right that i intended to use `state` though. i'll just take this out for now until i get a better repro

PR #955, opened 2019-02-26
Feature/content search tile links

gfodor:
This PR updates the media browser to use actual links in the tiles, so you can right click and get the underlying URL or open it in a new tab to preview it.

### DIFF
--- a/src/react-components/media-browser.js
+++ b/src/react-components/media-browser.js
@@ -138,6 +138,7 @@ class MediaBrowser extends Component {
     if (!this.props.onMediaSearchResultEntrySelected) return;
     this.props.onMediaSearchResultEntrySelected(entry);
     this.close();
+    return false;
   };
 
   handleSourceClicked = source => {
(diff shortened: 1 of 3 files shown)


src/react-components/media-browser.js
```diff
@@ -138,6 +138,7 @@ class MediaBrowser extends Component {
     if (!this.props.onMediaSearchResultEntrySelected) return;
     this.props.onMediaSearchResultEntrySelected(entry);
     this.close();
+    return false;
```
brianpeiris: Would be nice to use `e.preventDefault()` instead to be more explicit.

src/react-components/media-browser.js
```diff
@@ -342,18 +343,27 @@ class MediaBrowser extends Component {
...
+          href={entry.url}
+          target="_blank"
+          rel="noreferrer noopener"
           onClick={() => this.handleEntryClicked(entry)}
           className={styles.image}
           style={{ width: `${imageWidth}px`, height: `${imageHeight}px` }}
         >
           <img src={scaledThumbnailUrlFor(imageSrc, imageWidth, imageHeight)} />
-        </div>
+        </a>
```
brianpeiris: 💯 hooray for real links

src/react-components/media-browser.js
gfodor: ah yes, i'm too old school

PR #957, opened 2019-02-26
Avatar skinning

netpro2k:
Adds support for uploading custom avatar skinning (currently behind the `customSkin` query param). The first iteration only allows reskinning of a single GLTF model who's ID is hardcoded into the client code. It will remain this way until we roll out an avatar selector and avatars_listings (similar to how scenes work today). A PR will follow to flip the feature on by default once we have completed legal review and have prepared a better base avatar.

Another flag (`advancedAvatarEditor`) is also available to flip on further debug options in the editor, including uploading a glb. This is primarily used to create future base avatars.

See associated reticulum PR for details on how avatar records are being stored.
https://github.com/mozilla/reticulum/pull/118


src/scene-entry-manager.js
```diff
@@ -155,7 +159,7 @@ export default class SceneEntryManager {
     const { avatarId, displayName } = this.store.state.profile;
     this.playerRig.setAttribute("player-info", {
       displayName,
-      avatarSrc: avatarId && avatarId.startsWith("http") ? proxiedUrlFor(avatarId) : `#${avatarId || "botdefault"}`
+      avatarSrc: getAvatarSrc(avatarId, avatarCachebuster++)
```
gfodor: surprised by this -- we shouldn't be HTTP caching the reticulum URLs right?

src/react-components/profile-entry-panel.js
```diff
@@ -105,57 +176,35 @@ class ProfileEntryPanel extends Component {
...
                 <a
-                  className="custom-url-link"
-                  onClick={() =>
-                    this.setState({ customMode: true, avatarId: avatars.find(a => a.id === this.state.avatarId).model })
-                  }
+                  onClick={() => this.setState({ avatarType: "skinnable", avatarId: null })}
+                  className={classNames({ selected: this.state.avatarType === "skinnable" })}
                 >
-                  custom url
+                  Custom Skin
```
gfodor: this label and the others should go into the strings file

src/react-components/profile-entry-panel.js
```diff
@@ -85,6 +93,69 @@ class ProfileEntryPanel extends Component {
...
+            <input
+              id="custom-avatar-url"
+              type="url"
+              required
+              className={styles.formFieldText}
+              value={this.state.avatarId}
+              onChange={e => this.setState({ avatarId: e.target.value })}
+            />
+            <div className={styles.info}>
+              Resources for creating custom avatar models can be found
```
gfodor: strings file

src/react-components/profile-entry-panel.js
```diff
@@ -105,57 +176,35 @@ class ProfileEntryPanel extends Component {
...
+                  onClick={() => this.setState({ avatarType: "skinnable", avatarId: null })}
+                  className={classNames({ selected: this.state.avatarType === "skinnable" })}
                 >
-                  custom url
+                  Custom Skin
                 </a>
-              </div>
+              )}
+              <a
+                onClick={() => this.setState({ avatarType: "url", avatarId: "" })}
```
gfodor: does `null` not work here? i tried it locally and it at least initialized OK

src/storage/store.js
```diff
@@ -18,7 +18,8 @@ export const SCHEMA = {
       additionalProperties: false,
       properties: {
         displayName: { type: "string", pattern: "^[A-Za-z0-9-]{3,32}$" },
-        avatarId: { type: "string" }
+        avatarId: { type: "string" },
+        personalAvatarId: { type: "string" }
```
gfodor: I'd consider maybe breaking this up to be a bit more future proof since these store data structures are schema checked. Perhaps an `avatars` array with right now an expected single value of `{ avatarId: XXX, type: 'personal' }`

src/react-components/avatar-editor.js
```diff
@@ -0,0 +1,247 @@
...
+    });
+
+    this.props.onAvatarChanged(avatar.avatar_id);
+
+    this.setState({ ...this.state, avatar });
+  };
+
+  getAvatar = avatarId =>
+    fetch(getReticulumFetchUrl(`/api/v1/avatars/${avatarId}`), {
+      method: "GET",
```
gfodor: isn't `GET` assumed?

gfodor: it seems like a good idea to DRY this into `phoenix-utils` (which should really be called `reticulum-utils`, but I digress)

src/react-components/avatar-editor.js
```diff
@@ -0,0 +1,247 @@
...
+    this.props.onAvatarChanged(avatar.avatar_id);
+
+    this.setState({ uploading: false });
+    this.props.saveStateAndFinish();
+  };
+
+  fileField = (name, label, accept, disabled = false, title) => (
+    <div className={styles.fileInputRow} key={name} title={title}>
+      <label htmlFor={`avatar-file_${name}`}>
+        <div className="img-box">{this.state.avatar.files[name] && <img src={this.state.avatar.files[name]} />}</div>
```
gfodor: should use style imports for classes

src/react-components/avatar-editor.js
```diff
@@ -0,0 +1,247 @@
...
+          {this.fileField("normal_map", "Normal Map", "image/*")}
+
+          {this.fileField("orm_map", "ORM Map", "image/*", false, "Occlussion (r), Roughness (g), Metallic (b)")}
+
+          {/* {this.fileField("ao_map", "AO Map", "images/\*", true)} */}
+          {/* {this.fileField("metallic_map", "Metallic Map", "image/\*", true)} */}
+          {/* {this.fileField("roughness_map", "Roughness Map", "image/\*", true)} */}
+
+          <div className={styles.info}>
+            Resources for creating custom avatar skins can be found
```
gfodor: strings

src/react-components/avatar-editor.js
```diff
@@ -0,0 +1,247 @@
...
+        <div className="form-body">
+          {advanced && this.textField("avatar_id", "Avatar ID", true)}
+          {advanced && this.textField("parent_avatar_id", "Parent Avatar ID")}
+          {advanced && this.textField("name", "Name")}
+          {advanced && this.textField("description", "Description")}
+          {advanced && this.checkbox("allow_remixing", "Allow Remixing")}
+          {advanced && this.checkbox("allow_promotion", "Allow Promotion")}
+          {advanced && this.fileField("glb", "Avatar GLB", "model/gltf+binary,.glb")}
+
+          {this.fileField("base_map", "Base Map", "image/*")}
```
gfodor: i suggest breaking this up so that the latter 3 fields are hidden by default and revealed via an "Advanced" link

next to the base map, we should have a link to the raw template base image. and then if you click advanced there should be links to the photoshop and substances projects. i think we can also hide the github link unless you click advanced if we wanted to keep it super simple.

as a stretch goal you could make the base map one big, centered image as a poor man's preview of the avatar, with the additional advanced maps, if enabled, being the current size below it.

src/assets/stylesheets/profile.scss
```diff
@@ -30,13 +44,59 @@
...
+  }
+
   :local(.avatar-selector-container) {
     flex: 1;
     position: relative;
     width: 95%;
     text-align: center;
     margin: 16px;
 
+    &.avatar-editor {
```
gfodor: should use `:local` for these nested ones too, its what I've been doing at least :)

src/react-components/profile-entry-panel.js
```diff
@@ -85,6 +93,69 @@ class ProfileEntryPanel extends Component {
...
+              id="custom-avatar-url"
+              type="url"
+              required
+              className={styles.formFieldText}
+              value={this.state.avatarId}
+              onChange={e => this.setState({ avatarId: e.target.value })}
+            />
+            <div className={styles.info}>
+              Resources for creating custom avatar models can be found
+              <a target="_blank" rel="noopener noreferrer" href="https://github.com/j-conrad/hubs-avatar-pipelines">
```
gfodor: we should include a direct link to the blender file (with a blender logo probably) and eventually probably a link to a 1m tutorial

src/scene-entry-manager.js
netpro2k: Yeah.. though we do GLTF cache it, and we want to broadcast the network update so that other clients fetch it, this does both, though I reworked how this is done since this was causing some issues.

src/storage/store.js
netpro2k: hmm I was assuming this would be completely throwaway and we would just be fetching this from the db. Since it represents something owned by an account I don't think its ever going to exist just in localStorage

src/react-components/avatar-editor.js
netpro2k: Usually only do that for the root element of a component, since they end up being scoped anyway, though this was just sloppy.

src/storage/store.js
gfodor: right now the schema validator complains if you have an unrecognized key, so once it's here, we have to keep it forever (or loosen the schema validator up somehow)

src/react-components/profile-entry-panel.js
netpro2k: React complains if you try to set a text input to null since the `.value` of ""

src/storage/store.js
netpro2k: The way our schema validation works right now seems too strict. The fact that the application will simply not start if there is a superfluous key seems wrong to me.

gfodor: i might be misremembering though

PR #959, opened 2019-02-27
Suport ultrawide and hi-res monitors.

robertlong:
@johnshaughnessy do you mind testing this out on your 4k monitor?


robertlong:
I'm gonna close this and perhaps we can work out a better solution. Perhaps in the settings page?

PR #965, opened 2019-02-27
Replace aframe-teleport-controls with bespoke teleporter.

johnshaughnessy:
This brings `aframe-teleport-contols` into our project, removes all the pieces we don't use, and changes it to poll `userinput`. I initially was just trying to track down a tricky issue I was facing when trying to remove `tracked-controls`, but I want to do this anyway to remove some code and uses of `action-to-event`.


src/components/teleporter.js
```diff
@@ -0,0 +1,377 @@
...
+    const userinput = AFRAME.scenes[0].systems.userinput;
+    const { start, confirm, speed } = this.data;
+    const object3D = this.el.object3D;
+
+    if (!this.isTeleporting && userinput.get(start)) {
+      this.isTeleporting = true;
+      this.timeTeleporting = 0;
+      this.hit = false;
+      this.rayCurve.mesh.visible = true;
+      this.rayCurve.mesh.updateMatrixWorld();
```
netpro2k: Realize this is probably copypaste from teleport controls, but unclear why this needs to get updated

src/components/teleporter.js
```diff
@@ -0,0 +1,377 @@
...
+      this.rayCurve.mesh.material.needsUpdate = true;
+      this.playSound("teleport-sound-start");
+    }
+
+    if (!this.isTeleporting) {
+      return;
+    }
+
+    if (userinput.get(confirm)) {
+      this.hitEntity.setAttribute("visible", false);
```
netpro2k:
```suggestion
      this.hitEntity.object3D.visible = false;
```

src/components/teleporter.js
```diff
@@ -0,0 +1,377 @@
...
+      parabolicCurve(this.p0, this.v0, t, vecHelper);
+      this.rayCurve.setPoint(i, vecHelper);
+    }
+
+    const color = this.hit ? HIT_COLOR : MISS_COLOR;
+    const opacity = this.hit && this.timeTeleporting >= DRAW_TIME_MS ? HIT_OPACITY : MISS_OPACITY;
+    this.rayCurve.material.color.set(color);
+    this.rayCurve.material.opacity = opacity;
+
+    this.hitEntity.setAttribute("visible", this.hit);
```
netpro2k:
```suggestion
    this.hitEntity.object3D.visible = this.hit;
```

src/components/teleporter.js
```diff
@@ -0,0 +1,377 @@
...
+    if (this.hit) {
+      this.hitEntity.setAttribute("position", this.hitPoint);
```
netpro2k:
```suggestion
      this.hitEntity.object3D.position = this.hitPoint;
```

src/components/teleporter.js
```diff
@@ -0,0 +1,377 @@
...
+      this.hitEntity.object3D.traverse(o => {
+        o.matrixNeedsUpdate = true;
```
netpro2k: This seems really surprising. The matrix updating code should handle updating children itself, you should only need to set it on hitEntity

src/components/teleporter.js
```diff
@@ -0,0 +1,377 @@
...
+      });
+      const hitEntityOpacity = HIT_OPACITY * easeOutIn(percentToDraw);
+      const dRadii = this.data.outerRadius - this.data.hitCylinderRadius;
+      const outerScale = (this.data.outerRadius - easeIn(percentToDraw) * dRadii) / this.data.outerRadius;
+      this.outerTorus.object3D.scale.set(outerScale, outerScale, 1);
+      this.torus.setAttribute("material", "opacity", hitEntityOpacity);
```
netpro2k: would like to get rid of use of the material component here, but might make sense as its own PR

src/components/teleporter.js
```diff
@@ -0,0 +1,377 @@
...
+      this.cylinder.setAttribute("material", "opacity", hitEntityOpacity);
+    }
+  },
+
+  playSound(sound) {
+    this.el.emit(sound);
+  },
+
+  createHitEntity() {
```
netpro2k: As mentioned above, rewriting this to just use Object3Ds would be nice, especially since using the position/rotation/scale components is extra scary with our matrixUpdate changes... That said teleport-controls was mostly working so probably fine for that to be a second pass if it turns out to be not a quick change.

src/components/teleporter.js
```diff
@@ -0,0 +1,377 @@
...
+const DRAW_TIME_MS = 400;
+const q = new THREE.Quaternion();
+const vecHelper = new THREE.Vector3();
+
+AFRAME.registerComponent("teleporter", {
+  schema: {
+    start: { type: "string" },
+    confirm: { type: "string" },
+    speed: { default: 12 },
+    collisionEntities: { default: "" },
```
netpro2k: We only care about navmesh and we end up having a whole component dedicated to getting this to look at navmesh at the right time, I would vote we just do it more directly and less configurable.

src/components/teleporter.js
```diff
@@ -0,0 +1,377 @@
...
...1weaHOwvSFpxapLhIsOpZATIhOOJTwQRAqqBaMJfITdyWOCnnCHcJnIi/RNtGI2ENcKh5O8kgqTXqS7JG8NXkkxTOlLOW5hCepkLxMDUzdmzqeFpp2IG0yPTq9MYOSkZBxQqohTZO2Z+pn5mZ2y6xlhbL+xW6Lty8elQfJa7OQrAVZLQq2QqboVFoo1yoHsmdlV2a/zYnKOZarnivN7cyzytuQN5zvn//tEsIS4ZK2pYZLVy0dWOa9rGo5sjxxedsK4xUFK4ZWBqw8uIq2Km3VT6vtV5eufr0mek1rgV7ByoLBtQFr6wtVCuWFfevc1+1dT1gvWd+1YfqGnRs+FYmKrhTbF5cVf9go3HjlG4dvyr+Z3JS0qavEuWTPZtJm6ebeLZ5bDpaql+aXDm4N2dq0Dd9WtO319kXbL5fNKNu7g7ZDuaO/PLi8ZafJzs07P1SkVPRU+lQ27tLdtWHX+G7R7ht7vPY07NXbW7z3/T7JvttVAVVN1WbVZftJ+7P3P66Jqun4lvttXa1ObXHtxwPSA/0HIw6217nU1R3SPVRSj9Yr60cOxx++/p3vdy0NNg1VjZzG4iNwRHnk6fcJ3/ceDTradox7rOEH0x92HWcdL2pCmvKaRptTmvtbYlu6T8w+0dbq3nr8R9sfD5w0PFl5SvNUyWna6YLTk2fyz4ydlZ19fi753GDborZ752PO32oPb++6EHTh0kX/i+c7vDvOXPK4dPKy2+UTV7hXmq86X23qdOo8/pPTT8e7nLuarrlca7nuer21e2b36RueN87d9L158Rb/1tWeOT3dvfN6b/fF9/XfFt1+cif9zsu72Xcn7q28T7xf9EDtQdlD3YfVP1v+3Njv3H9qwHeg89HcR/cGhYPP/pH1jw9DBY+Zj8uGDYbrnjg+OTniP3L96fynQ89kzyaeF/6i/suuFxYvfvjV69fO0ZjRoZfyl5O/bXyl/erA6xmv28bCxh6+yXgzMV70VvvtwXfcdx3vo98PT+R8IH8o/2j5sfVT0Kf7kxmTk/8EA5jz/GMzLdsAAAAgY0hSTQAAeiUAAICDAAD5/wAAgOkAAHUwAADqYAAAOpgAABdvkl/FRgAAADJJREFUeNpEx7ENgDAAAzArK0JA6f8X9oewlcWStU1wBGdwB08wgjeYm79jc2nbYH0DAC/+CORJxO5fAAAAAElFTkSuQmCC";
```
netpro2k: Might make this an external image
image

johnshaughnessy:
Thanks for the feedback @netpro2k -- I'll aim to have a fast followup addressing these comments. For now I'm going to push as-is because I want this and https://github.com/mozilla/hubs/pull/970 in more than I want to fix these.

PR #968, opened 2019-02-28
Help Tips

gfodor:
There are now help tips throughout the experience in Hubs.

image

This PR also reshuffles the settings menu again to surface tips in case a user accidentally closes it.


src/systems/tips.js
```diff
@@ -0,0 +1,274 @@
...
+
+    if (prevTop !== this.activeTips.top || prevBottom !== this.activeTips.bottom) {
+      this.el.emit("tips_changed", this.activeTips);
+    }
+  },
+
+  _performStep: function(tips, scope) {
+    if (finishedScopes[scope]) return;
+
+    const scene = AFRAME.scenes[0];
```
netpro2k: In systems, `this.el` is the scene

src/systems/tips.js
```diff
@@ -0,0 +1,274 @@
...
+let finishedScopes = {}; // Optimization, lets system skip scopes altogether once finished.
+
+const isMobile = AFRAME.utils.device.isMobile();
+
+const tipPlatform = () => {
+  if (detectInHMD()) return "standalone";
+  return isMobile ? "mobile" : "desktop";
+};
+
+const platformTips = () => {
```
netpro2k: Since platform can not change, instead of doing this I might pull the current platforms tips out into another const. Then, remove all the finished ones. And in `markTipFinished` remove it from the array. That way you never bother looping though tips in the (very common) case you have completed them. And in the case where we determine at startup that you have no tips to complete we can just not bother doing any work at all.

src/systems/tips.js
```diff
@@ -0,0 +1,274 @@
...
+
+    if (localStorage.getItem(LOCAL_STORAGE_KEY) === null) {
+      localStorage.setItem(LOCAL_STORAGE_KEY, JSON.stringify({}));
+    }
+
+    this._tickCount = 0;
+  },
+
+  tick: function() {
+    this._tickCount++;
```
netpro2k: Don't see this getting used at all. Left in by mistake?

src/systems/tips.js
```diff
@@ -0,0 +1,274 @@
...
+      switch (VALIDATORS[tip](this._userinput, scene, this._mediaCounter, window.APP.store)) {
+        case FINISH:
+          markTipFinished(tip);
+          break;
+        case VALID:
+          chosenTip = `${tipPlatform()}.${tip}`;
+          break;
+      }
+
+      if (chosenTip) break;
```
netpro2k: Hmm. If I finished a tip this frame (even if I wasnt seeing the tip), seems like we still want to mark it as finished. This early out might skip some.

src/systems/tips.js
```diff
@@ -0,0 +1,274 @@
...
+    this._performStep(tips.top, "top");
+    this._performStep(tips.bottom, "bottom");
+
+    if (prevTop !== this.activeTips.top || prevBottom !== this.activeTips.bottom) {
+      this.el.emit("tips_changed", this.activeTips);
+    }
+  },
+
+  _performStep: function(tips, scope) {
+    if (finishedScopes[scope]) return;
```
netpro2k: Micro-optimization, but might just move this check out of the function so you can avoid making the call at all

src/systems/tips.js
```diff
@@ -0,0 +1,274 @@
...
+    this._performStep = this._performStep.bind(this);
+
+    if (localStorage.getItem(LOCAL_STORAGE_KEY) === null) {
+      localStorage.setItem(LOCAL_STORAGE_KEY, JSON.stringify({}));
+    }
+
+    this._tickCount = 0;
+  },
+
+  tick: function() {
```
netpro2k: Seems a shame we are going to be running this tick function in the very common case that you are continuing to use hubs and have already completed all the tips. Made some suggestions above about avoiding the work if we detect that we have no tips to display this session.

src/systems/tips.js
```diff
@@ -0,0 +1,274 @@
...
+          chosenTip = `${tipPlatform()}.${tip}`;
+          break;
+      }
+
+      if (chosenTip) break;
+    }
+
+    this.activeTips[scope] = chosenTip;
+
+    if (finishCount === tips.length) {
```
netpro2k: Per my suggestion above, maybe instead of finishCount we might want to just remove tips when they are finished.

src/systems/tips.js
```diff
@@ -0,0 +1,274 @@
...
+  },
+  standalone: { top: [], bottom: [] }
+};
+
+let localStorageCache = null;
+let finishedScopes = {}; // Optimization, lets system skip scopes altogether once finished.
+
+const isMobile = AFRAME.utils.device.isMobile();
+
+const tipPlatform = () => {
```
netpro2k: Platform can't change, no need for this to be a function.

src/systems/tips.js
gfodor: The first part makes sense, but mutating the array could get ugly because of the "Show Tips" reset button

src/systems/tips.js
```diff
@@ -0,0 +1,274 @@
...
+    this._tickCount++;
+
+    if (!this._userinput) {
+      this._userinput = AFRAME.scenes[0].systems.userinput;
+
+      if (!this._userinput) return;
+    }
+
+    if (!this._mediaCounter) {
+      this._mediaCounter = document.querySelector("#media-counter").components["networked-counter"];
```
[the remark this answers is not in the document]
gfodor: yep fails in init, query selector fails. (interesting: this is similar behavior to what i was seeing with the default camera where the full DOM is not available in system `init`)

src/systems/tips.js
gfodor: that's a tough one actually. i think i'm going to leave it as is for now until we see if its annoying without doing this. if I did this, then there's probably some flows of FTUE where the user incidentally hits some feature without really grokking it and we blow past the tip. (eg, if they accidentally hit Q or something, they'll never actually be told Q's functionality.)

src/systems/tips.js
gfodor: the caller side calls this twice so I'm going to leave this since it's slightly more DRY

src/systems/tips.js
gfodor: the top tips are always "in-play", since we never FINISH them. (arguably we could explore that later, I was not sure if it made sense or not since the top tips basically give you a clear way to escape out of "modes" if you never learned or forgot how to.) the bottom tips, which are more sequences and one-time, have an early-exit optimization where it just checks `finishedScopes` at least so it doesn't have to walk + validate them all once you've completed them.

src/systems/tips.js
netpro2k: Ah makes sense, missed that feature

src/systems/tips.js
netpro2k: > they'll never actually be told Q's functionality
They had something even better, they experienced it, and now they don't need to be told.

src/systems/tips.js
```diff
@@ -145,6 +158,18 @@ const VALIDATORS = {
     if (userinput.activeSets.has(sets.cursorHoldingInteractable)) return FINISH;
     return VALID;
   },
+  object_rotate: function(userinput, scene, mediaCounter, store) {
+    if (!scene.is("frozen")) return INVALID;
+    if (mediaCounter.count() === 0) return INVALID;
+    if (store && store.state.activity.hasRotated) return FINISH;
```
netpro2k: Probably makes sense to do this check first otherwise you are keeping this tip unfinished longer than it needs to be

PR #970, opened 2019-02-28
Remove use of tracked-controls.

johnshaughnessy:
Removes uses of `tracked-controls` components in our code, since it does a lot more work than what we were using it for.

Although we aren't removing `tracked-controls` from our `aframe` fork, I counted the number of lines we now avoid running and it's a little over 2000.


src/systems/userinput/devices/daydream-controller.js
```diff
@@ -10,6 +13,15 @@ export class DaydreamControllerDevice {
     this.rayObjectRotation = new THREE.Quaternion();
     this.selector = `#player-${gamepad.hand}-controller`;
     this.pose = new Pose();
+    this.sittingToStandingMatrix = new THREE.Matrix4().makeTranslation(0, 1.6, 0);
+    navigator.getVRDisplays().then(d => {
```
netpro2k: We could technically use
`scene.renderer.vr.getStandingMatrix()` here instead since its already a Matrix4 version of this, but I think decoupling from `WebVRManager` makes sense. 

We should add guards for `navigator.getVRDisplays` and `stageParameters` existing though. 

Just a note from checking the spec, it seems like this should not change out from under us unless we call `resetPose` so doing this only once here seems correct.

src/systems/userinput/devices/daydream-controller.js
```diff
@@ -10,6 +13,15 @@ export class DaydreamControllerDevice {
...
+      const m = d[0].stageParameters.sittingToStandingTransform;
+      for (let i = 0; i < 16; i++) {
```
netpro2k: Matrix4 has `fromArray` which does this loop for you.

src/systems/userinput/devices/daydream-controller.js
```diff
@@ -10,6 +13,15 @@ export class DaydreamControllerDevice {
...
+        this.sittingToStandingMatrix.elements[i] = m[i];
+      }
+    });
+    this.matrix = new THREE.Matrix4();
```
netpro2k: This are just caches and not actual state so they should be pulled out into consts or as a closure in `write` (slight preference for the former). This is both for style (so you know they are not actual persistent sate) and for controllers with multiple instances actually saves constructing another object (minor given how many exist)

src/systems/userinput/devices/daydream-controller.js
```diff
@@ -1,5 +1,8 @@
 import { paths } from "../paths";
 import { Pose } from "../pose";
+import { applyArmModel } from "../arm-model.js";
+
+const ONES = new THREE.Vector3(1, 1, 1);
```
netpro2k: If only `THREE.Vector3.ONE` was a thing, but then again I guarantee we would get clobbered at some point and cause all sorts of bugs, so I take it back.

src/systems/userinput/devices/daydream-controller.js
netpro2k: Since we do this in all the controllers, might pull this out into a function that resolves to a Promise to a Matrix4

src/systems/userinput/devices/daydream-controller.js
```diff
@@ -45,6 +57,19 @@ export class DaydreamControllerDevice {
       this.pose.direction.set(0, 0, -1).applyQuaternion(this.rayObjectRotation);
       this.pose.fromOriginAndDirection(this.pose.position, this.pose.direction);
       frame[paths.device.daydream.pose] = this.pose;
+
+      this.headObject3D = this.headObject3D || document.querySelector("#player-camera").object3D;
+
+      // TODO: hand arm model
```
netpro2k: Unless I am missing something, this TODO is done

src/systems/userinput/devices/gear-vr-controller.js
```diff
@@ -45,6 +58,16 @@ export class GearVRControllerDevice {
       this.pose.direction.set(0, 0, -1).applyQuaternion(this.rayObjectRotation);
       this.pose.fromOriginAndDirection(this.pose.position, this.pose.direction);
       frame[paths.device.gearVRController.pose] = this.pose;
+      // TODO: hand arm model
+      if (this.gamepad.pose.orientation) {
+        frame[paths.device.gearVRController.matrix] = this.matrix
+          .compose(
+            applyArmModel(this.gamepad, this.gamepad.hand, this.headObject3D, 1.6),
```
netpro2k: Don't see where `this.headObject3D` gets set.

src/systems/userinput/devices/daydream-controller.js
```diff
@@ -45,6 +57,19 @@ export class DaydreamControllerDevice {
       this.pose.direction.set(0, 0, -1).applyQuaternion(this.rayObjectRotation);
       this.pose.fromOriginAndDirection(this.pose.position, this.pose.direction);
       frame[paths.device.daydream.pose] = this.pose;
+
+      this.headObject3D = this.headObject3D || document.querySelector("#player-camera").object3D;
```
netpro2k: Since we are already getting the VRDisplay we might as well do the right thing and directly `getPose()` from it and pass the `VRPose` and `GamepadPose` to `applyArmModel`

src/systems/userinput/devices/vive-controller.js
```diff
@@ -59,16 +72,22 @@ export class ViveControllerDevice {
...
-    }
     const rayObject = el.object3D;
     rayObject.updateMatrixWorld();
     this.rayObjectRotation.setFromRotationMatrix(rayObject.matrixWorld);
     this.pose.position.setFromMatrixPosition(rayObject.matrixWorld);
     this.pose.direction.set(0, 0, -1).applyQuaternion(this.rayObjectRotation);
     this.pose.fromOriginAndDirection(this.pose.position, this.pose.direction);
     frame[this.path.pose] = this.pose;
+
+    if (this.gamepad.pose.position && this.gamepad.pose.orientation) {
```
netpro2k: Even if we don't have both position and rotation we should still probably update what we do have.

src/systems/userinput/paths.js
```diff
@@ -193,7 +197,8 @@ paths.device.daydream = {
   axis: axisName => {
     return `${daydream}axis/${axisName}`;
   },
-  pose: `${daydream}pose`
+  pose: `${daydream}pose`,
+  matrix: `${daydream}matrix`
```
netpro2k: As we talked about in person, I think this should replace `pose` and not be called `matrix` but seems fine to do that as a fast follow since I agree that `pose` is used in some mysterious ways at the moment, so its more to debug

src/components/hand-controls2.js
```diff
@@ -127,8 +100,13 @@ AFRAME.registerComponent("hand-controls2", {
   tick() {
     const hand = this.data;
     const userinput = AFRAME.scenes[0].systems.userinput;
+    const shouldBeVisible = userinput.get(
+      hand === "left" ? paths.actions.leftHand.matrix : paths.actions.rightHand.matrix
+    );
+    if (this.el.object3D.visible !== shouldBeVisible) {
+      this.el.setAttribute("visible", shouldBeVisible);
```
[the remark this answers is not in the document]
johnshaughnessy: It seems we should always be setting visibility this way. I see how `getAttribute` checks to see if the attribute you're checking is `visible` and the `visible` component just sets the visibility of the object3D. There are about 30 places where we do this. I'll change all of them in one go in a separate PR.

src/components/copy-matrix.js
```diff
@@ -0,0 +1,15 @@
+AFRAME.registerComponent("copy-matrix", {
```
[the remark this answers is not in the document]
johnshaughnessy: 👍

src/systems/userinput/devices/daydream-controller.js
johnshaughnessy: I added a guard for `stageParameters` (and in fact this is `undefined` in chrome on android with daydream, so that's good). Why guard against `getVRDisplays`?

PR #971, opened 2019-02-28
Deal with default camera issue

gfodor:
This PR deals with a long standing issue where there is a race condition on the A-Frame `camera` system running a `querySelector` for `[camera]` tags in the scene DOM. If there is no such tag, it assumes there is no camera, and so creates its default camera + look controls. In practice, it seems like this query selector can fail in Hubs perhaps due to the complex hierarchy surrounding the `#player-camera` in the DOM, but I'm not sure.

In any case, there doesn't seem to be a way to flip this behavior off or set things up to prevent the default camera from getting injected when it does trigger. So, this PR introduces a component which will properly clean up the default camera if it was injected.

### DIFF
--- a/src/components/set-active-camera.js
+++ b/src/components/set-active-camera.js
@@ -5,7 +5,7 @@ AFRAME.registerComponent("set-active-camera", {
     const cameraSystem = this.el.sceneEl.systems.camera;
     const currentActiveCamera = cameraSystem.activeCameraEl;
 
-    if (currentActiveCamera && currentActiveCamera !== this.el) {
+    if (currentActiveCamera !== this.el) {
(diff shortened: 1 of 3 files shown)


gfodor:
Yeah I considered that -- the reason I decided against this is this bug is *very* non-deterministic for me (1 in 15-20 runs.) So if we accidentally forgot to re-patch this system it'd probably just quietly regress for many months before we noticed.

The "roll back the camera" thing is still pretty exposed, but I think it is going to be more future-proofed than a patch to the A-Frame system.

gfodor:
the CI fail seems to be a CI timeout. I'll ignore for now and see if master passes.

PR #973, opened 2019-02-28
Set service worker timeout

gfodor:
This adds a 5s timeout to the registration of the service worker completing. I thought I had all the relevant error handling in place, but it seems that in some cases the browser is not completing the promise for the registration, at all, when testing on a mobile device on a LAN. I am not positive what is going on, it seems related to the service worker loading over a slow LAN, but this timeout seems sane anyway (and the result is that the checkbox will just be hidden for notifications.)


johnshaughnessy:
Ah nice, I was going to ask about this. Had done something similar locally.
image

PR #976, opened 2019-03-01
Fix up touch controls

gfodor:
This fixes up a bunch of things with the touch controls:

- Tapping and holding on the screen (on android) no longer makes a fake selection
- Fix for the user input bindings that was resulting in holding your finger doing a grab/ungrab every frame :)
- Fixes a bunch of behaviors with the touch handling:
  - If you put two fingers down quickly to pinch or freeze, you will not accidentally grab the object on the first finger
  - We delay the grab an extra input frame to eliminate the weird physics side effect
  - You can now tap off of objects to de-hover/select them. This is done by always moving the cursor on touch start.

We also now don't do the query selector for the player camera every tick

### DIFF
--- a/src/systems/userinput/bindings/touchscreen-user.js
+++ b/src/systems/userinput/bindings/touchscreen-user.js
@@ -87,7 +87,7 @@ export const touchscreenUserBindings = addSetsToBindings({
     {
       src: { value: paths.device.touchscreen.isTouchingGrabbable },
       dest: { value: paths.actions.cursor.grab },
-      xform: xforms.copy
+      xform: xforms.rising
     },
     {
       src: { value: paths.device.hud.penButton },
--- a/src/systems/userinput/devices/app-aware-touchscreen.js
+++ b/src/systems/userinput/devices/app-aware-touchscreen.js
@@ -13,6 +13,18 @@ function distance(x1, y1, x2, y2) {
   return Math.sqrt(dx * dx + dy * dy);
 }
 
+const getPlayerCamera = (() => {
+  let playerCamera;
+
+  return function() {
+    if (!playerCamera) {
+      playerCamera = document.querySelector("#player-camera").components.camera.camera;
+    }
+
+    return playerCamera;
+  };
+})();
+
 function shouldMoveCursor(touch, raycaster) {
   const cursorController = document.querySelector("[cursor-controller]").components["cursor-controller"];
   const isCursorGrabbing = cursorController.data.cursor.components["super-hands"].state.has("grab-start");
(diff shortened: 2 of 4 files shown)


src/systems/userinput/devices/app-aware-touchscreen.js
```diff
@@ -149,44 +161,51 @@ export class AppAwareTouchscreenDevice {
...
+      let assignment;
+
+      // First touch
+      if (shouldMoveCursor(touch, this.raycaster)) {
+        assignment = assign(touch, MOVE_CURSOR_JOB, this.assignments);
+
+        // Grabbing objects is delayed by several frames:
+        // - We don't want the physics constraint to be applied too early, which results in crazy velocities
+        // - We don't want to mis-trigger grabs if the user is about to put down a second finger.
+        assignment.framesUntilGrab = 2;
```
johnshaughnessy: This is a scary hack but I don't have a better idea for how to fix things. Glad it works. I expect that we're going to have to revisit this as the app code changes. (It may be affected if we change physics, cursor-controller, or tick order.)

src/systems/userinput/devices/app-aware-touchscreen.js
```diff
@@ -13,6 +13,18 @@ function distance(x1, y1, x2, y2) {
   return Math.sqrt(dx * dx + dy * dy);
 }
 
+const getPlayerCamera = (() => {
```
johnshaughnessy: Ah good

src/systems/userinput/bindings/touchscreen-user.js
```diff
@@ -87,7 +87,7 @@ export const touchscreenUserBindings = addSetsToBindings({
     {
       src: { value: paths.device.touchscreen.isTouchingGrabbable },
       dest: { value: paths.actions.cursor.grab },
-      xform: xforms.copy
+      xform: xforms.rising
```
johnshaughnessy: Surprised to see that this was `copy`. I guess it never mattered because we would immediately change action sets?

gfodor: i think typically you'd be tapping and releasing for clicks (so you end up "clicking" a few times) and yes otherwise you're going to change action sets. so it just affected buttons like video chrome and camera

src/systems/userinput/devices/app-aware-touchscreen.js
gfodor: yep in the end this was a cleaner way to deal the grab for the pinch case, so i think this will stick around even if we dropped the need to delay for the physics issue.

PR #977, opened 2019-03-01
Room lobby modal CSS cleanup

gfodor:
It now looks like this:

image

### DIFF
--- a/src/react-components/ui-root.js
+++ b/src/react-components/ui-root.js
@@ -960,6 +960,9 @@ class UIRoot extends Component {
 
     return (
       <div className={entryStyles.entryPanel}>
+        <div className={entryStyles.header}>
+          <FormattedMessage id="entry.room" />
+        </div>
         <div className={entryStyles.name}>
           <span>{this.props.hubName}</span>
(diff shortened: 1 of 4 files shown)


src/react-components/ui-root.js
```diff
@@ -1020,7 +1023,7 @@ class UIRoot extends Component {
                     e.target.blur();
                   }
                 }}
-                placeholder="Send to room..."
+                placeholder="Send message to others..."
```
johnshaughnessy: Hmm... this copy feels awkward to me but am not sure if I can think of something better (that also fits on small screens). 
How about "Message the room..."?

src/react-components/ui-root.js
```diff
@@ -960,6 +960,9 @@ class UIRoot extends Component {
 
     return (
       <div className={entryStyles.entryPanel}>
+        <div className={entryStyles.header}>
+          <FormattedMessage id="entry.room" />
```
johnshaughnessy: I don't think this is a good addition

gfodor:
PR is updated to improve responsiveness of the chatbox

gfodor:
latest design:

image

PR #985, opened 2019-03-01
Ammo.js Physics

InfiniteLee:
Switch over to a new (still WIP) version of `aframe-physics-system` that uses `ammo.js` (https://github.com/kripken/ammo.js/) instead of `cannon.js` for physics.
e89ee80aa8268cb5bc0fa09063cae0d6

This enables many features that were not available before:
- convex hull shapes that can be dynamically generated, now the default for gltf models pasted into hubs
- static trimesh collision for environments (currently always enabed, but eventually can be configured via spoke). This replaces how heightmaps are currently used.
- sleeping/waking physics bodies
- enable/disable debug rendering at runtime via the `/debug` command. (note: performance and physics collision accuracy are significantly reduced while debug is on).
- collision masking/layers

b0e4014bc60dd828f593c114afdf2d5c

Future features to be supported:
- physics-based character controller (instead of navmesh)
- physics-based raycasting


package.json
```diff
@@ -44,10 +44,11 @@
     "aframe": "github:mozillareality/aframe#hubs/master",
     "aframe-billboard-component": "^2.0.0",
     "aframe-inspector": "^0.8.3",
-    "aframe-physics-extras": "github:mozillareality/aframe-physics-extras#bugfix/physics-collider-world",
-    "aframe-physics-system": "github:mozillareality/aframe-physics-system#93b1dfe64c00587a9d09e1a8071750e75f340815",
+    "aframe-motion-capture-components": "github:mozillareality/aframe-motion-capture-components#aframe090",
```
johnshaughnessy: I removed motion-capture-components intentionally. Do we need it again?

package.json
```diff
@@ -44,10 +44,11 @@
...
+    "aframe-physics-system": "github:mozillareality/aframe-physics-system#de74d40f71bc07de415dc0e1fc335055019cdf6a",
```
johnshaughnessy: Prefer pointing to a named branch in our other repos (#hubs/master) when possible, then following the same policy of "whatever is the head of this branch should be the code being run in hubs". If they are ever out of sync we can see that by seeing the shah in `package.lock`

package.json
```diff
@@ -75,7 +76,7 @@
     "react-router-dom": "^4.3.1",
     "react-youtube": "^7.8.0",
     "screenfull": "^3.3.2",
-    "super-hands": "github:mozillareality/aframe-super-hands-component#feature/drawing",
+    "super-hands": "github:mozillareality/aframe-super-hands-component#5a7ffe9805870f4821e6e636486a7cce01c6cf64",
```
johnshaughnessy: Same preference for pointing at a meaningfully named branch rather than a commit sha. Not sure if this preference is justified, to be honest. But figured I'd mention it anyway.

src/assets/translations.data.json
```diff
@@ -171,6 +171,7 @@
     "commands.help": "Show help.",
     "commands.leave": "Disconnect from the room.",
     "commands.duck": "The duck tested well. Quack.",
+    "commands.debug": "Toggle physics debug rendering.",
```
johnshaughnessy: perhaps should be called "physics debug" since "debug" alone is a bit vague

src/components/sticky-object.js
```diff
@@ -51,22 +51,22 @@ AFRAME.registerComponent("sticky-object", {
   _onRelease() {
     // Happens if the object is still being held by another hand
     if (this.el.is("grabbed")) return;
-
     if (
       this.data.autoLockOnRelease &&
       (this.data.autoLockSpeedLimit === 0 ||
-        this.el.body.velocity.lengthSquared() < this.data.autoLockSpeedLimit * this.data.autoLockSpeedLimit)
+        this.el.components["ammo-body"].getVelocity() < this.data.autoLockSpeedLimit)
```
johnshaughnessy: Perhaps `getVelocity` could be called `getSpeed` to indicate that it returns a scalar value

src/components/sticky-object.js
```diff
@@ -51,22 +51,22 @@ AFRAME.registerComponent("sticky-object", {
...
     }
-    this.el.body.collisionResponse = true;
+
+    this.el.setAttribute("ammo-body", { collisionFilterGroup: 1 });
   },
 
   _onGrab() {
     if (!this.el.components.grabbable || this.el.components.grabbable.data.maxGrabbers === 0) return;
 
+    this.el.setAttribute("ammo-body", { collisionFilterGroup: this.locked ? 2 : 1 });
```
johnshaughnessy: I suggest naming and exporting these filter groups as named constants rather than refer to them by their numerical value. (Except in the html, where we can't avoid it).

src/components/sticky-object.js
```diff
@@ -88,10 +88,11 @@ AFRAME.registerComponent("sticky-object-zone", {
     getLastWorldPosition(this.el.object3D, this.worldPosition);
     getLastWorldQuaternion(this.el.object3D, this.worldQuaternion);
 
-    const dir = new THREE.Vector3(0, 0, 5).applyQuaternion(this.el.object3D.quaternion);
-    this.bootImpulsePosition = new CANNON.Vec3(0, 0, 0);
-    this.bootImpulse = new CANNON.Vec3();
-    this.bootImpulse.copy(dir);
+    //TODO: support this with ammo
```
johnshaughnessy: Should this be removed or does this still need to be done?

src/components/super-networked-interactable.js
```diff
@@ -66,24 +68,30 @@ AFRAME.registerComponent("super-networked-interactable", {
...
     this.hand.emit("haptic_pulse", { intensity: "high" });
-    if (this.networkedEl && !NAF.utils.isMine(this.networkedEl)) {
-      if (NAF.utils.takeOwnership(this.networkedEl)) {
-        this.el.setAttribute("body", { type: "dynamic" });
-        this._syncCounterRegistration();
+    if (this.networkedEl) {
+      if (!NAF.utils.isMine(this.networkedEl)) {
+        if (NAF.utils.takeOwnership(this.networkedEl)) {
+          this.el.setAttribute("ammo-body", { type: "dynamic" });
+          this.el.body.forceActivationState(4);
```
johnshaughnessy: Same as collision flags - I suggest you export them and then refer to them by name

src/components/super-spawner.js
```diff
@@ -210,6 +219,7 @@ AFRAME.registerComponent("super-spawner", {
 
   activateCooldown() {
     if (this.data.spawnCooldown > 0) {
+      //TODO: set collision flags on ammo body
```
johnshaughnessy: Still needs doing?

src/message-dispatch.js
```diff
@@ -71,7 +72,15 @@ export default class MessageDispatch {
         break;
       case "duck":
         spawnChatMessage(DUCK_URL);
-        this.scene.emit("quack");
+        if (Math.random() < 0.01) {
```
johnshaughnessy: 👌 ✔️ ✔️ Gettin the 💯 important changes in right here 💯 💯 👌 👌

src/react-components/chat-command-help.js
```diff
@@ -11,7 +11,7 @@ export default class ChatCommandHelp extends Component {
   };
 
   render() {
-    const commands = ["leave", "fly", "grow", "shrink", "duck", "scene <scene url>", "rename <new name>"];
+    const commands = ["leave", "fly", "grow", "shrink", "duck", "debug", "scene <scene url>", "rename <new name>"];
```
johnshaughnessy: "debug" seems too general

package.json
mqp: I agree with this. The old behavior of referring to a specific commit worked better with yarn IIRC, but now referring to a branch works well. (A specific commit will wind up in the lock file, and you can do `npm install github:foo/bar#baz` again to update which commit that is to the tip of the branch.)

src/components/sticky-object.js
InfiniteLee: This still needs to be done, but I didn't do it because none of this code is live right now.

package.json
InfiniteLee: bad merge, thanks

package.json
InfiniteLee: For some reason I thought we preferred sha's but ok I can switch it to a branch name

src/components/media-loader.js
```diff
@@ -111,9 +115,9 @@ AFRAME.registerComponent("media-loader", {
       this.loadingClip.play();
     }
     this.el.setObject3D("mesh", mesh);
-    this.hasBakedShapes = !!(this.el.body && this.el.body.shapes.length > 0);
-    this.setShapeAndScale(true);
+    this.setScale(true);
     delete this.showLoaderTimeout;
+    this.addShape("box", "loader");
```
johnshaughnessy: May be nice to export constants AMMO_BOX, AMMO_HULL, etc to help indicate that these are specific strings that are meaningful to ammo components

src/react-components/chat-command-help.js
InfiniteLee: imo, we can change it if we ever introduce a different kind of debug, but for now it's the only one and typing anything longer would be frustrating

package.json
mqp: Yeah, I used to prefer SHAs also back when the tools worked less well, but now branches seem to work well.

src/components/media-loader.js
```diff
@@ -26,6 +27,16 @@ const fetchMaxContentIndex = url => {
   return fetch(url).then(r => parseInt(r.headers.get("x-max-content-index")));
 };
 
+const getNumMeshes = sceneRoot => {
```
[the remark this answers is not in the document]
InfiniteLee: this method is no longer used, removing

PR #986, opened 2019-03-02
Remove some words I prefer not to have in my room name

johnshaughnessy:


src/utils/name-generation.js
```diff
@@ -619,7 +616,6 @@ const nouns = [
   "get together",
   "conclave",
   "party",
-  "powwow",
```
InfiniteLee: keep `powwow`, remove `milky`

InfiniteLee: nvm, i was not aware of the cultural implications of "powwow"

PR #988, opened 2019-03-02
Bug/mobile hud ux cleanup

gfodor:
This PR (finally) fixes the hover issues on mobile in the HUD so that we can have hover states on desktop but not end up with the weird "stuck" hover images on mobile. It also tweaks the invite CSS so that the mini-invite link is not shown on landscape mode, where vertical real estate is as a premium.


src/assets/stylesheets/2d-hud.scss
```diff
@@ -172,57 +172,82 @@
   position: relative;
 }
 
+// This roundabout pattern is the only way I can get hovers to not happen on touch devices
 :local(.iconButton.mute) {
   background-image: url(../hud/mute_off.png);
 }
 :local(.iconButton.mute:hover) {
-  background-image: url(../hud/mute_off-hover.png);
+  @media (any-pointer: fine) {
```
johnshaughnessy: Does this really work? When I run the codepen example about `any-pointer` on mdn, it seems that both the `fine` and `coarse` rules take effect. Whichever rule is listed last gets priority. I tested this in both chrome and firefox.

https://developer.mozilla.org/en-US/docs/Web/CSS/@media/any-pointer

gfodor: For me, on desktop only `fine` selects, on phone then `fine` and `coarse` select. maybe you were running with mobile mode turned on (the intended behavior is on mobile there will be no hover behavior.)

PR #989, opened 2019-03-02
Remove tracked controls webvr/xr systems

johnshaughnessy:
The `tracked-controls-webvr` and `tracked-controls-webxr` systems were still being loaded and running. In order to remove them from our project, I also needed to update the bot playback code. Since we did not use the "stroke" or "events" features of the bot recording, I removed `aframe-motion-capture-components` entirely and wrote a `replay` component that is compatible with our existing recordings but only uses the saved `poses`. 

Will merge this to our aframe fork: https://github.com/MozillaReality/aframe/pull/8/files


package.json
```diff
@@ -41,10 +41,9 @@
     "@fortawesome/fontawesome-svg-core": "^1.2.2",
     "@fortawesome/free-solid-svg-icons": "^5.2.0",
     "@fortawesome/react-fontawesome": "^0.1.0",
-    "aframe": "github:mozillareality/aframe#afa764e5f515784e7349f834a53142bcadcd0ffe",
+    "aframe": "github:mozillareality/aframe#remove-tracked-controls",
```
johnshaughnessy: Will change this to point to `aframe#hubs/master` after merging https://github.com/MozillaReality/aframe/pull/8

src/components/replay.js
```diff
@@ -0,0 +1,24 @@
...
+  init: function() {
+    this.playhead = 0;
+    this.poseIndex = 0;
+  },
+
+  tick: function(t, dt) {
+    let overflow = false;
+    while (!overflow && this.playhead >= this.poses[this.poseIndex].timestamp) {
+      this.el.setAttribute("position", this.poses[this.poseIndex].position);
+      this.el.setAttribute("rotation", this.poses[this.poseIndex].rotation);
```
johnshaughnessy: We can avoid calling `setAttribute` here but I didn't know if the `rotation` euler needed to be a component so that its `order` was changed. I decided to just copy what the original code did. We can address it later if we need to optimize the bot code (which I doubt will be especially important anytime soon).

netpro2k: Yeah don't think bot perf matters to much

src/components/replay.js
```diff
@@ -0,0 +1,24 @@
...
+      this.el.object3D.updateMatrix();
```
netpro2k: Probably not a huge deal since its just bots, but why not just set the flag here?

src/components/replay.js
```diff
@@ -0,0 +1,24 @@
+AFRAME.registerComponent("replay", {
+  init: function() {
+    this.playhead = 0;
+    this.poseIndex = 0;
+  },
+
+  tick: function(t, dt) {
+    let overflow = false;
+    while (!overflow && this.playhead >= this.poses[this.poseIndex].timestamp) {
```
netpro2k: This seems odd, it appears to be setting the position/rotation/scale of the same element multiple times in a single frame. Seems like it should find the pose for the current playhead position and then set the position/rotation/scale once.

src/components/visibility-by-path.js
```diff
@@ -0,0 +1,12 @@
+AFRAME.registerComponent("visibility-by-path", {
+  schema: {
+    path: { type: "string" }
+  },
+  tick() {
+    const userinput = AFRAME.scenes[0].systems.userinput;
+    const shouldBeVisible = !!userinput.get(this.data.path);
+    if (this.el.object3D.visible !== shouldBeVisible) {
+      this.el.setAttribute("visible", shouldBeVisible);
```
netpro2k: Did you determine that you had to use the setAttribute version for networking?

src/components/replay.js
johnshaughnessy: No reason besides that I copied the code

src/components/replay.js
johnshaughnessy: Yes that seems more correct

src/components/visibility-by-path.js
johnshaughnessy: I have not yet determined that. I plan to figure out that issue and then decide what to do about all the times we do this in the code. A little nervous to just set the visibility on the object3D until I figure that out.

src/components/replay.js
johnshaughnessy: But did you consider that writing that version will require _slightly_ more work 😜

PR #995, opened 2019-03-02
Fix up media browser feeling lucky, re-order tabs, fix bug with initi…

gfodor:
This fixes issues with the media browser:

- Bug: the initial filters were not being set since the stashed params were being cleared by setting to empty object, not null

- Feature: Ctrl-enter is now the standard "spawn" hotkey in both chat and media browser

- Feature: Ctrl-enter can now be hit in the media browser without waiting for results (great for blind entry)

- Feature: Tabs are re-ordered -- images are shifted to the right, with models to the left.


gfodor:
nice catch!

PR #997, opened 2019-03-04
Add hacky override for oculus go left handed users

gfodor:
This improves the situation on Go for users who have selected the controller to be left handed (though its still not ideal.) Previously, the ray did not work at all so you couldn't really use the app. Now the ray is at least attached to the hand so the main effect is you just have a backwards (right) hand for your left hand.


johnshaughnessy:
Not pretty, but I agree it's better than things not working

PR #1006, opened 2019-03-05
Bug/grab flicker

gfodor:
This PR addresses a bug that manifests due to the fact that if you quickly tap on an object on touchscreen, we end up transitioning to a new action set, but only write `isTouchingGrabbable` to the frame once, so the `falling` xform does not ever fire, causing the object remained un-dropped. 

This is addressed by adding an additional edge case if the grab was being delayed (due to the pinch detection) -- in that case we now also delay the unassignment of the job by two input frames, which will ensure `true` gets written to the path at least twice: the first time will transition the action set, and the second time will result in the `true/false` falling pair to appear in the stream so the object is dropped.

### DIFF
--- a/src/systems/userinput/devices/app-aware-touchscreen.js
+++ b/src/systems/userinput/devices/app-aware-touchscreen.js
@@ -68,15 +68,15 @@ export class AppAwareTouchscreenDevice {
       switch (assignment.job) {
         case MOVE_CURSOR_JOB:
         case MOVE_CAMERA_JOB:
-          // If grab was being delayed, we should fire the initial grab to ensure
-          // clicks will work.
-          if (assignment.framesUntilGrab > 0) {
+          // If grab was being delayed, we should fire the initial grab and also delay the unassignment
+          // to ensure we write at least two frames with the grab down (since the action set will change)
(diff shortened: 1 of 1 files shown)


gfodor:
I've updated this PR with one more fix: this fixes up cases with three fingers -- either if you pinch first to start, or pinch after grabbing an object.

src/systems/userinput/devices/app-aware-touchscreen.js
```diff
@@ -278,8 +299,17 @@ export class AppAwareTouchscreenDevice {
 
     if (hasCursorJob) {
       const assignment = findByJob(MOVE_CURSOR_JOB, this.assignments);
+
       frame[path.isTouchingGrabbable] = assignment.framesUntilGrab <= 0;
       assignment.framesUntilGrab--;
+
+      if (assignment.framesUntilUnassign >= 0) {
```
johnshaughnessy: Odd that until touch end, this check is `false` because it's asking `undefined >= 0`

gfodor:
Yep it's really gnarly at this point. Hopefully we won't have to touch it
much for a while :) will fix that expression

PR #1008, opened 2019-03-05
Fix the ray pull gesture to bring objects near properly

gfodor:
Fixes a regression where pulling an object along the ray to your hand puts it inside your hand again. I am not sure what changed, perhaps the origin of the ray?


gfodor:
saw you already cut a PR for this :)

PR #1009, opened 2019-03-05
Fix for stretching of canvas when preemptively exiting VR

gfodor:
We have some logic to exit VR if VR mode is erroneously retained across refreshes, but the logic doesn't properly avoid setting the CSS class that stretches the canvas if that condition is hit.

### DIFF
--- a/src/hub.js
+++ b/src/hub.js
@@ -538,7 +538,7 @@ document.addEventListener("DOMContentLoaded", async () => {
   window.APP.scene = scene;
   window.APP.hubChannel = hubChannel;
 
-  const preEmptVRMode = () => {
+  const preemptVRMode = () => {
     // If VR headset is activated, refreshing page will fire vrdisplayactivate
     // which puts A-Frame in VR mode, so exit VR mode whenever it is attempted
     // to be entered and we haven't entered the room yet.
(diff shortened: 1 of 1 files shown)


src/hub.js
```diff
@@ -538,24 +538,34 @@ document.addEventListener("DOMContentLoaded", async () => {
...
-      console.log("Pre-emptively exiting VR mode.");
-      scene.exitVR();
-    }
   });
 
-  scene.addEventListener("exit-vr", () => document.body.classList.remove("vr-mode"));
+  preEmptVRMode();
+
+  scene.addEventListener("exit-vr", () => {
+    document.body.classList.remove("vr-mode");
```
johnshaughnessy: Seems odd that we have to remove the `vr-mode` class. I thought this is what `exitVR` in aframe did: https://github.com/MozillaReality/aframe/blob/hubs/master/src/core/scene/a-scene.js#L376

src/hub.js
```diff
@@ -538,24 +538,34 @@ document.addEventListener("DOMContentLoaded", async () => {
   window.APP.scene = scene;
   window.APP.hubChannel = hubChannel;
 
+  const preEmptVRMode = () => {
```
johnshaughnessy: `preEmpt` -> `preempt`

gfodor:
A-Frame sets the vr-mode state on the scene, which is (afaik) unrelated to
CSS classes

gfodor:
Also this is working around the default A-frame behavior which I believe is
intended, where it draws both eye buffers.

PR #1010, opened 2019-03-05
Put interstitial near HUD

gfodor:
When clicking the spawn icon the interstitial notice in VR to lift your headset is outside of your FoV. This sets things up to put the notice near the HUD where you are looking. The component which manages this is also now renamed to `follow-in-fov`.

### DIFF
--- a/src/hub.html
+++ b/src/hub.html
@@ -759,8 +759,8 @@
                 cursorSuperHand: #cursor;"
         ></a-entity>
 
-        <a-entity matrix-auto-update class="vr-notice" follow-in-lower-fov="target: #player-camera; offset: 0 0 -0.8; angle: 39;" visible="false">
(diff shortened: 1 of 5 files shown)


src/hub.html
```diff
@@ -759,8 +759,8 @@
                 cursorSuperHand: #cursor;"
         ></a-entity>
 
-        <a-entity matrix-auto-update class="vr-notice" follow-in-lower-fov="target: #player-camera; offset: 0 0 -0.8; angle: 39;" visible="false">
-            <a-entity hoverable slice9="width: 0.9; height: 0.2; left: 64; top: 64; right: 66; bottom: 66; opacity: 1.0; src: #action-button"
```
johnshaughnessy: Did `#action-button` change to `#button` on purpose?

gfodor: yep. #button is now the grey, tintable buton. #action-button is the pink one.

PR #1011, opened 2019-03-05
Fix cursor color issue with env map

gfodor:
This works around an issue where if you set `material` A-Frame component attributes, the environment map on the underlying material is wiped out. We do this in two places, the teleporter and the cursor when updating the color. I've left the teleporter out for now since it probably isn't that big of a deal, but the cursor should probably always reflect the actual environment map.

### DIFF
--- a/src/components/cursor-controller.js
+++ b/src/components/cursor-controller.js
@@ -241,8 +241,14 @@ AFRAME.registerComponent("cursor-controller", {
           ? cursorColorHovered
           : cursorColorUnhovered;
 
-      if (this.data.cursor.components.material.data.color !== cursorColor) {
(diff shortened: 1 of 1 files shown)


johnshaughnessy:
Alternatively we can not call `setAttribute` and or ditch the aframe material altogether. The `setAttribute` here causes a LOT of extra work. All we want to do here is update a shader uniform, which we can accomplish by accessing the THREE material. I have a patch for this I can share later today

gfodor:
that sounds smarter, i wasn't sure how to do that but if you have a PR that's great.

johnshaughnessy:
Bump. I forgot to share the patch. Will get it out today

johnshaughnessy:
https://github.com/mozilla/hubs/pull/1024

PR #1015, opened 2019-03-05
Scene remixing WIP

robertlong:
To Do:
- [ ] Add Spoke scene url to Reticulum's scene API https://github.com/mozilla/reticulum/pull/123
- [ ] Test PR with downloaded Spoke scenes.
- [ ] Clean up copy and add any images.


src/assets/stylesheets/scene-ui.scss
```diff
@@ -169,3 +169,11 @@
   top: -12px;
   color: white;
 }
+
+:local(.remixButton) {
```
gfodor: probably `remix-button`?

src/assets/translations.data.json
```diff
@@ -188,6 +189,8 @@
     "settings.help": "Help",
     "settings.row-profile": "Profile",
     "settings.row-room": "Room",
+    "settings.remix-scene": "Remix scene in Spoke",
+    "video_share.notify": "Select again to un-share video.",
```
gfodor: not sure why this line is added?

src/react-components/remix-scene-dialog.js
```diff
@@ -0,0 +1,99 @@
...
+
+const Steps = {
+  DownloadSpoke: "download-spoke",
+  DownloadScene: "download-scene",
+  OpenScene: "open-scene"
+};
+
+export default class RemixSceneDialog extends Component {
+  static propTypes = {
+    sceneUrl: PropTypes.string,
```
gfodor: i believe you are using `spokeUrl` elsewhere, i'd probably call this `sceneProjectUrl` everywhere as we are now using that term in the API

src/react-components/remix-scene-dialog.js
```diff
@@ -0,0 +1,99 @@
...
+    onClose: PropTypes.func
+  };
+
+  state = {
+    step: Steps.DownloadSpoke,
+    platform: getPlatform(),
+    downloadUrl: releasesLink
```
gfodor: i'd call this `downloadSpokeUrl` to avoid confusing between this and the project download url

gfodor:
On the scene listing page, the button area seems a bit crowded. I think the Remix button should be lower in the hierarchy than both the Create Room and Tweet buttons -- a good place for it may be to the left of the Spoke logo in the bottom right, or drop the Spoke logo altogether and replace it with the remix button. I was also thinking we may want to adopt a unique color for the remix button, since it's going to represent a significant departure from the app into content-creation flows. I originally designed it as a burnt-orange.

gfodor:
In Hubs, I'd recommend we move the remix button out of the settings menu and into the scene + room info dialog, which is where I think we should be directing the user to go to in order to see or do anything related to the currently selected room scene. I think eventually when we can export scenes into hosted spoke we probably should have an "Export Scene..." modal in settings that will let the user export it directly, download a .spoke project file (which will incidentally be a remix) or go into hosted spoke to remix.

gfodor:
For the remix dialog itself -- i think a better approach will just be to have a single modal:

- At the top, says "To remix scenes, you'll first need to download [Spoke](https://hubs.mozilla.com/spoke)", or perhaps direct link to the download as you are doing

- The main panel in the dialog has left-justified steps you have in the last part of the modal

- There's a single CTA at the bottom to "Download Spoke Project"

- There's an X to close, no need for a button at the bottom. You'll want to confirm back button closes it out too.

gfodor:
can this be closed @robertlong ?

PR #1017, opened 2019-03-06
Fix showing of player hud when camera active

gfodor:
In 2D mode turning on the camera or mirror mode would inadvertently show the in-world HUD even if it was previously not visible.

### DIFF
--- a/src/components/camera-tool.js
+++ b/src/components/camera-tool.js
@@ -195,7 +195,10 @@ AFRAME.registerComponent("camera-tool", {
           this.playerHead.updateMatrixWorld(true, true);
         }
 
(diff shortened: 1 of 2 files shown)


gfodor:
We currently have a different (generic) component managing the visibility,
which hides it in 2d mode. I'm not sure if adding a bespoke component and
replacing the generic one is right or not, vs this current approach. The
camera code is basically more of a "culling" algorithm being applied to the
head and HUD for that render pass and had a bug which failed to restore the
state. I'm not sure if a component is the right abstraction here, I think
if we wanted to do this "right" it would be to use layers or something.

Issue #1016, opened 2019-03-06
Non-signed in users should not be able to unpin objects


InfiniteLee:
Users should be prompted to sign-in if they try to unpin an object and are not already signed-in. 
Currently, a non-signed-in user can unpin and delete an object, deleting that object for any user currently in the space, however any new users entering the space will see the pinned object still pinned in its original location.

InfiniteLee:
fixed by #1018

PR #1018, opened 2019-03-06
Force user to sign in to un pin

gfodor:
Forces users to sign in to un-pin, also fixes a bug where things would get messed up because of the re-entrancy of the sign in flow if the user cancelled the sign in prompt. (The user-facing result is you would never see the sign in prompt again unless you refresh.)

Addresses https://github.com/mozilla/hubs/issues/1016


InfiniteLee:
lgtm

PR #1020, opened 2019-03-07
Add deadzone and increase sensitivty on pen resize

gfodor:
This fixes up the oculus touch UX for changing the pen settings -- pen resizing now has nicer sensitivity for finer brush scaling, and color changing is now subject to a deadzone so you won't resize the tip while you swap colors.


gfodor:
oops, i missed the left hand for the scale change!

PR #1023, opened 2019-03-07
Lower the unmute button UI

gfodor:
Lower the unmute button on iOS so it doesn't cover up the center of the video.


gfodor:
i think i just want to remove this

gfodor:
actually this still seems necessary

PR #1029, opened 2019-03-08
Remove chat box on top UX, no longer needed

gfodor:
After the app was updated to always exit full screen on focus of text fields, the mechanic where we move the chat entry to the top is no longer necessary, since the keyboard no longer covers it up (since we are not in full screen mode.)

This also bumps `screenfull` which changes the API to return promises, so we can get rid of a setTimeout.


gfodor:
oh, looks like this PR fixed chrome fullscreen for me. i can't repro that issue above however.

PR #1032, opened 2019-03-08
Add camera photo spawning animation and fix orientation

gfodor:
This rotates the photos 180 on Y so they reflect the view in the viewfinder, and also adds a little animation so they pop out of the camera.


gfodor:
Added an arrangement alg so photos appear around the camera.

src/components/camera-tool.js
```diff
@@ -254,9 +263,35 @@ AFRAME.registerComponent("camera-tool", {
...
+          const { entity, orientation } = addMedia(file, "#interactable-media", undefined, false);
+
+          const pos = this.el.object3D.position;
+
+          entity.object3D.position.set(pos.x, pos.y, pos.z);
           entity.object3D.rotation.copy(this.el.object3D.rotation);
+          entity.object3D.rotateY(Math.PI);
+          entity.object3D.scale.set(0.1, 0.1, 0.1);
+
+          // Generate photos in a circle around camera, starting from the bottom.
```
netpro2k: I like it! I would make the first photo go directly under the camera instead. Feels like a bug when you only take 1 photo.

src/components/camera-tool.js
```diff
@@ -254,9 +263,35 @@ AFRAME.registerComponent("camera-tool", {
...
+
+          photoPos.set(
+            Math.cos(Math.PI * 2 * (idx / 6.0)) * 0.75,
+            Math.sin(Math.PI * 2 * (idx / 6.0)) * 0.75,
+            -0.05 + idx * 0.001
+          );
+
+          this.el.object3D.localToWorld(photoPos);
+
+          entity.setAttribute("animation__photo_pos", {
```
netpro2k: We should technically have an option in media-loader for it.. but a quick hack to wait till after the image is loaded to do the aniation. Feels much nicer than animating out the cube:
```
          entity.object3D.visible = false;
          entity.addEventListener(
            "image-loaded",
            () => {
              entity.object3D.visible = true;
              entity.setAttribute("animation__photo_pos", {
                property: "position",
                dur: 800,
                from: { x: pos.x, y: pos.y, z: pos.z },
                to: { x: photoPos.x, y: photoPos.y, z: photoPos.z },
                easing: "easeOutElastic"
              });
            },
            { once: true }
          );
```

src/components/camera-tool.js
gfodor: Ah I considered that but the problem with that is that unless I got fancier with the algorithm, it'd force placement of another one directly above the camera, behind the viewfinder. Basically the ideal arrangement isn't a linear polar walk but that's what I hacked out here

Issue #606, opened 2018-10-03
hub.link buttons don't work on touchscreen laptops


gfodor:
due to the hack put in to work around this issue:

https://github.com/facebook/react/issues/9809

if the device supports both touch and mouse, the mouse is ignored. the proper 'fix' is to break the button into separate components and use ref's as is done in the react issue above, which will re-enable the ability to use `preventDefault` on the event.

gfodor:
PR is also updated to address https://github.com/mozilla/hubs/issues/606

gfodor:
PR out

johnshaughnessy:
The fix regressed entry on Go. See https://github.com/mozilla/hubs/pull/1080

gfodor:
now fixed

PR #1042, opened 2019-03-09
Fix bug where pasting doens't work on hub.link

gfodor:
Previously we weren't handling situations where the user pasted into the textbox, now the `onChange` will always check if it's time to redirect.


gfodor:
updated to also address https://github.com/mozilla/hubs/issues/606

PR #1044, opened 2019-03-09
Bug/ios video audio corruption

gfodor:
This adds two changes that seem to mitigate a large part of the scenarios where Safari on iOS has audio die or videos can misbehave:

- We weren't clearing out audio sources on removal, oops
- Prevent the user from unmuting for a few seconds.

### DIFF
--- a/src/components/media-views.js
+++ b/src/components/media-views.js
@@ -330,6 +330,7 @@ AFRAME.registerComponent("media-video", {
     this.togglePlaying = this.togglePlaying.bind(this);
 
     this.lastUpdate = 0;
+    this.videoMutedAt = 0;
 
     this.el.setAttribute("hover-menu__video", { template: "#video-hover-menu", dirs: ["forward", "back"] });
     this.el.components["hover-menu__video"].getHoverMenu().then(menu => {
@@ -556,6 +557,10 @@ AFRAME.registerComponent("media-video", {
     }
 
     this.updatePlaybackState(true);
+
+    if (this.video.muted) {
+      this.videoMutedAt = performance.now();
+    }
 
     this.el.emit("video-loaded");
   },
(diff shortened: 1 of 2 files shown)


src/components/media-views.js
```diff
@@ -617,6 +622,11 @@ AFRAME.registerComponent("media-video", {
   remove() {
     this.cleanUp();
 
+    const sound = this.el.getObject3D("sound");
```
netpro2k: I would do this instead to ensure the THREE.Audio node also gets GCed.

````
if(this.audio) {
  this.el.removeObject3D("sound");
  this.audio.disconnect();
  delete this.audio;
}
```

In practical terms it shouldn't matter either way since if we likely removing this `media-video` component in response to removing the whole entity, but good to always `removeObject3D` for any `setObject3D` you do in a component . (I bet we have other places where we forgot to do this).

src/components/unmute-video-button.js
```diff
@@ -5,9 +5,19 @@ AFRAME.registerComponent("unmute-video-button", {
...
       const mediaVideo = videoEl.components["media-video"];
-      if (!mediaVideo || !mediaVideo.video) return;
-      mediaVideo.video.muted = false;
-      this.el.setAttribute("visible", false);
+      if (!mediaVideo || !mediaVideo.video || !mediaVideo.videoMutedAt) return;
+
+      // iOS initially plays the sound and *then* mutes it, and sometimes a second video playing
+      // can break all sound in the app. (Likely a Safari bug.) Adding a delay before the unmute
+      // occurs seems to help with reducing this.
+      if (performance.now() - mediaVideo.videoMutedAt < 3000) {
```
netpro2k: I might normally complain about this sort of thing, but having done iOS web audio debugging in a past life, this is a very tame hack, so.. nothing to see here.. lol

Issue #1049, opened 2019-03-10
Ammo issues with pinned objects


gfodor:
When entering a room with previously pinned objects, I noticed the following (may not be due to ammo, but seems likely):

- Pinned objects cannot be moved with menus open
- If pinned objects are un-pinned they still cannot be moved
- In debug mode, pinned objects appear white (active) instead of green (sleeping) after you try to move them

Oddly this seems to not be totally deterministic: the smoke test breaks consistently but only on mobile, and desktop the smoke test seems fine but a simple test breaks.

This room seems to break consistently in Chrome: go over to the duck floating in space and try to move it with the menus open, you won't be able to: https://hubs.mozilla.com/3FQhJrJ/honorable-ample-land

Note that pinning and unpinning works as expected if you are in the room at the time the pin happens. This is specifically for the case where you join a room with pinned objects.

gfodor:
I checked and it certainly looks like the `ammo-constraint attribute is being set properly on the object, so probably something within the physics system is preventing the constraint, or something is overriding everything.

gfodor:
After more digging i'm still stumped -- the only difference afacit between the two objects:

- the pinned object is part of a nested hierarchy because it is loaded as part of `objects.gltf`
- the pinned object has `addCollideEventListener` to `true` instead of `false` (flipping it doesn't fix anything)

gfodor:
OK -- I believe this is a race condition on startup. It seems to have been introduced in the ammo merge -- if I add a `setTimeout` before spawning the `objects.gltf` entity then everything works. Perhaps there's some initialization that needs to block before we start spawning stuff.

gfodor:
Yep: https://github.com/mozilla/hubs/pull/1050

PR #1050, opened 2019-03-10
Wait for physics

gfodor:
See https://github.com/mozilla/hubs/issues/1049


InfiniteLee:
lgtm. Hopefully when webpack supports WASM better in the future we won't need to do this anymore.

PR #1051, opened 2019-03-10
Fix up UX for pen and camera on Oculus Go

gfodor:
Fixes up a variety of issues with the camera & pen tools on Oculus Go:

- Changes camera to no longer do the sticky grab, now that we have the take picture button
- Removes the viewport preview (which is drawn black)
- Adjusts the pause gesture on Go to be the full (up/center/down) strip
- Adds snapping/deadzoning to the pen tool on X and Y to keep you from accidentally changing size/distance as you use it


src/systems/userinput/bindings/oculus-go-user.js
```diff
@@ -243,15 +261,25 @@ export default function generate3DOFTriggerBindings(device) {
...
+          value: touchpadYForPen,
+          touching: touchpadButton.touched
+        },
+        dest: { value: paths.actions.cursor.modDelta },
+        xform: xforms.touch_axis_scroll(1, 0.1),
+        priority: 200
+      },
+      {
+        src: {
+          value: dpadBottomCenterStrip,
```
johnshaughnessy: Would be cool to be able to teleport while holding an object by using the top of the touchpad, since here we are dropping with the bottom + center.

src/components/camera-tool.js
```diff
@@ -56,6 +56,9 @@ async function pixelsToPNG(pixels, width, height) {
   return new File([blob], "snap.png", { type: "image/png" });
 }
 
+// Don't show camera viewports on mobile VR to same framerate.
+const enableCameraViewport = !AFRAME.utils.device.isMobileVR();
```
johnshaughnessy: Might be nice to add a qs parameter or something here too. On desktop I am mildly annoyed when there is an abandoned camera in my view frustum that I know is doing a render I don't care about, but it may be that we should just be more careful about knowing when someone cares about seeing the camera preview. (e.g. perhaps there's a way for me to explicitly hide the preview, so that I don't have to destroy the camera when someone else wants it around and we can both get what we wanted)

src/systems/userinput/bindings/oculus-go-user.js
gfodor: we're only using the bottom center strip when holding the pen (since we use north to undo) -- non-pen mode uses the full strip.

src/components/camera-tool.js
gfodor: yeah this seems like something we should address at some point -- the preview is nice but we def show it too aggressively.

PR #1052, opened 2019-03-10
Improve HUD interactions with camera and pen

gfodor:
This PR updates the HUD semantics and UX for the camera and pen to be more consistent and usable:

- The camera button in the HUD on all platforms now is a toggle, and is highlighted when the user has an active camera.

- Similarly, the pen is now a toggle on 2D modes.

- In 3dof, dropping the pen removes it.

- In 6dof, spawning the pen from the HUD shows an animation of it going into your hand


gfodor:
Added P hotkey to spawn/toggle pen. Doing camera on a hotkey would be trickier since we don't have action paths set up for that.

gfodor:
added C for camera

gfodor:
added X for take snapshot

src/systems/userinput/paths.js
```diff
@@ -12,12 +12,15 @@ paths.actions.boost = "/actions/boost";
 paths.actions.startGazeTeleport = "/actions/startTeleport";
 paths.actions.stopGazeTeleport = "/actions/stopTeleport";
 paths.actions.spawnPen = "/actions/spawnPen";
+paths.actions.removePen = "/actions/removePen";
```
johnshaughnessy: Why is there `removePen` and `pen.remove`?

src/systems/userinput/paths.js
```diff
@@ -42,6 +45,8 @@ paths.actions.cursor.scalePenTip = "/actions/cursorScalePenTip";
 paths.actions.cursor.scaleGrabbedGrabbable = "/actions/cursorScaleGrabbedGrabbable";
 paths.actions.cursor.mediaVolumeMod = "/actions/cursor/mediaVolumeMod";
 paths.actions.cursor.takeSnapshot = "/actions/cursorTakeSnapshot";
+paths.actions.pen = {};
+paths.actions.pen.remove = "/actions/pen/remove";
```
johnshaughnessy: Does this need to be broken out for leftHand, rightHand, and cursor?

src/assets/stylesheets/2d-hud.scss
```diff
@@ -199,32 +199,58 @@
...
+
+:local(.iconButton.camera) {
+  background-image: url(../hud/camera_off.png);
+}
+:local(.iconButton.camera:hover) {
+  @media (pointer: fine) {
+    background-image: url(../hud/camera_off-hover.png);
+  }
+
+  @media (pointer: coarse) {
```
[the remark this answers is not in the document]
gfodor: I don't really know what else we can do here -- the CSS selectors we are using I think are correct and the cascade is mitigating some browser issues. On the touchscreen laptop, you want to avoid showing the hover if they used their finger to click, otherwise show it if they used the mouse. I don't think we should be wiring up onHover functions, that seems like not worth the additional complexity.

src/assets/stylesheets/2d-hud.scss
```diff
@@ -199,32 +199,58 @@
   }  
 }
 
-:local(.iconButton.spawn_pen) {
-  background-image: url(../hud/spawn_pen.png);
+:local(.iconButton.pen) {
```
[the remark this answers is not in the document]
gfodor: I exported these out of Figma at 8x to reduce aliasing, which is what we've been doing -- my guess is they were exported smaller before.

src/systems/userinput/paths.js
gfodor: oops, forgot to remove this one

src/systems/userinput/paths.js
gfodor: nope, the removal only happens via the HUD interaction on 2D screens and on 3dof on Go

PR #1053, opened 2019-03-10
Make pen sticky on vive

gfodor:
This updates the vive controls to make the pen grip lock by default, so a second grip click is needed to drop it. The reason is that the pen is unique in only actually being usable as a tool when gripped, so this seems like the right tradeoff in comfort vs consistency wrt the grip and trigger working the same as other objects.

### DIFF
--- a/src/systems/userinput/bindings/vive-user.js
+++ b/src/systems/userinput/bindings/vive-user.js
@@ -73,8 +73,10 @@ const keyboardBoost = k("boost");
 
 const leftGripPressed1 = v("leftGripPressed1");
 const leftGripPressed2 = v("leftGripPressed2");
+const leftGripPressed3 = v("leftGripPressed2");
 const rightGripPressed1 = v("rightGripPressed1");
 const rightGripPressed2 = v("rightGripPressed2");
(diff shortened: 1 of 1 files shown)


src/systems/userinput/bindings/vive-user.js
```diff
@@ -73,8 +73,10 @@ const keyboardBoost = k("boost");
 
 const leftGripPressed1 = v("leftGripPressed1");
 const leftGripPressed2 = v("leftGripPressed2");
+const leftGripPressed3 = v("leftGripPressed2");
```
johnshaughnessy: typo 2->3

src/systems/userinput/bindings/vive-user.js
```diff
@@ -73,8 +73,10 @@ const keyboardBoost = k("boost");
...
 const rightGripPressed1 = v("rightGripPressed1");
 const rightGripPressed2 = v("rightGripPressed2");
+const rightGripPressed3 = v("rightGripPressed2");
```
johnshaughnessy: typo 2->3

gfodor:
Sloppy on my part. Will verify these are necessary.

PR #1067, opened 2019-03-12
Add support for removing scenes

gfodor:
Goes with https://github.com/mozilla/reticulum/pull/126

Updates the admin UI to support removing scenes, and easily looking up a scene based upon its listing.

Updates the scene page and hub page to properly handle removed or delisted scenes.

### DIFF
--- a/src/hub.js
+++ b/src/hub.js
@@ -303,7 +303,8 @@ async function updateEnvironmentForHub(hub) {
   if (hub.scene) {
     isLegacyBundle = false;
     sceneUrl = hub.scene.model_url;
-  } else if (hub.scene === null) { // delisted/removed scene
+  } else if (hub.scene === null) {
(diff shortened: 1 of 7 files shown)


src/hub.js
```diff
@@ -303,6 +303,9 @@ async function updateEnvironmentForHub(hub) {
   if (hub.scene) {
     isLegacyBundle = false;
     sceneUrl = hub.scene.model_url;
+  } else if (hub.scene === null) {
```
brianpeiris: Does this actually work for existing hubs that point to a scene that is now removed/delisted? AFAICT, `join_with_hub` will still preload the scene data associated with the hub.

brianpeiris: Ah, hub_channel jsonification goes through scene_view

PR #1070, opened 2019-03-13
Support for removing maps from an avatar skin

netpro2k:
Associated reticulum change https://github.com/mozilla/reticulum/pull/127


src/react-components/avatar-editor.js
```diff
@@ -164,6 +164,25 @@ export default class AvatarEditor extends Component {
           });
         }}
       />
+      {this.state.avatar.files[name] && (
+        <a
```
gfodor: small UX tweak if easy I'd change this to just be an X (`faPlus`) `<button>`

gfodor:
clearing maps yields error:

image


other than that (and small UX suggestion) lgtm!

PR #1075, opened 2019-03-13
Handle cases where event handlers come in late in VR startup

gfodor:
The code currently has two hacks dealing with WebVR startup in A-Frame when a user refreshes the page after a VR session on desktop:

- It's always been the case that A-Frame/WebVR will retain VR mode across refreshes, we have a hack that will exit VR mode on startup if this case is hit

- There was a regression in 0.9.0 that causes A-Frame to miss this condition in certain conditions, resulting in a situation where the display is presenting but A-Frame did not run its relevant "enter VR" routines (to tag the scene metadata, set up cameras, etc.)

The previous hacks handled these cases by subscribing to `enter-vr `and `vrdisplaypresentchange` to get out ahead of these conditions, but in practice it looks like in certain cases these event subscriptions are happening too late, the events have already fired (for both conditions.) This PR updates them to both be handled when these events fire but also pre-emptively run the necessary logic at event subscription time in case the event has already fired.

### DIFF
--- a/src/hub.js
+++ b/src/hub.js
@@ -583,7 +583,7 @@ document.addEventListener("DOMContentLoaded", async () => {
   window.APP.scene = scene;
   window.APP.hubChannel = hubChannel;
 
-  scene.addEventListener("enter-vr", () => {
+  const handleEarlyVRMode = () => {
     // If VR headset is activated, refreshing page will fire vrdisplayactivate
     // which puts A-Frame in VR mode, so exit VR mode whenever it is attempted
     // to be entered and we haven't entered the room yet.
@@ -593,6 +593,12 @@ document.addEventListener("DOMContentLoaded", async () => {
       return true;
     }
 
+    return false;
+  };
+
+  scene.addEventListener("enter-vr", () => {
+    if (handleEarlyVRMode()) return true;
+
     document.body.classList.add("vr-mode");
 
     // Don't stretch canvas on cardboard, since that's drawing the actual VR view :)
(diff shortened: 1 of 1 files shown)


dmarcos:
A hack should not be necessary. `vrdisplayactivate` handler is attached early before the scene initializes: https://github.com/aframevr/aframe/blob/master/src/core/scene/loadingScreen.js#L13

It this is not working as expected we should fix it. Would you mind opening an issue on A-Frame with steps to reproduce?

gfodor:
Yup, I will try to isolate the root cause -- this PR is meant to un-break hubs. The first "hack" we have is not working around a bug (though it'd be nice to have an option to disable this behavior in A-Frame) -- the latter is definitely a bug.

gfodor:
Also, if you'd like to look into this before I am able to try to isolate a test case, the behavior we are seeing specifically is that when refreshing the page after a VR session, with some probability A-Frame will have not flipped the scene into `vr-mode`.

dmarcos:
Try A-Frame master. This might help: https://github.com/aframevr/aframe/pull/4093

gfodor:
Ah great, will report back here later with results!

gfodor:
OK good news, if I cherry pick that commit it looks like the issue we were seeing is resolved and the scene is properly in `vr-mode` if the page is refreshed and the display is active. The other part of this PR is still relevant (for our application to exit VR mode in this condition) but I'll cut another PR for that. Thanks for the support!

PR #1076, opened 2019-03-13
Fix bug where vr users do not get teleported to the correct spot.

johnshaughnessy:
image
image

And before you ask... Yes, this graphic exists here almost entirely because I want to play with my tablet computer. But also I wanted to illustrate the problem we've always had where you could accidentally end up teleporting somewhere invalid and not end up where you expected.

### DIFF
--- a/src/components/character-controller.js
+++ b/src/components/character-controller.js
@@ -90,14 +90,17 @@ AFRAME.registerComponent("character-controller", {
   teleportTo: (function() {
     const rig = new THREE.Vector3();
     const head = new THREE.Vector3();
-    const deltaFromHeadToTarget = new THREE.Vector3();
+    const deltaFromHeadToTargetForHead = new THREE.Vector3();
+    const targetForHead = new THREE.Vector3();
     const targetForRig = new THREE.Vector3();
     return function teleportTo(targetWorldPosition) {
       const o = this.el.object3D;
       o.getWorldPosition(rig);
       this.data.pivot.object3D.getWorldPosition(head);
-      deltaFromHeadToTarget.copy(targetWorldPosition).sub(head);
-      targetForRig.copy(rig).add(deltaFromHeadToTarget);
+      targetForHead.copy(targetWorldPosition);
+      targetForHead.y += this.data.pivot.object3D.position.y;
+      deltaFromHeadToTargetForHead.copy(targetForHead).sub(head);
+      targetForRig.copy(rig).add(deltaFromHeadToTargetForHead);
 
       const pathfinder = this.el.sceneEl.systems.nav.pathfinder;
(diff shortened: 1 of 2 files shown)


src/components/character-controller.js
```diff
@@ -88,11 +86,29 @@ AFRAME.registerComponent("character-controller", {
...
+      deltaFromHeadToTarget.copy(targetWorldPosition).sub(head);
+      targetForRig.copy(rig).add(deltaFromHeadToTarget);
+
+      const pathfinder = this.el.sceneEl.systems.nav.pathfinder;
+      this.navGroup = pathfinder.getGroup(NAV_ZONE, targetForRig, true, true);
+      this.navNode = null;
+      this._setNavNode(targetForRig);
+      pathfinder.clampStep(rig, targetForRig, this.navNode, NAV_ZONE, this.navGroup, o.position);
+      o.matrixNeedsUpdate = true;
+      o.updateMatrices();
```
johnshaughnessy: not sure if I should call updateMatrices and decompose here or if i should do it in tick instead, since that's where we need this to be up to date.

netpro2k: You are setting position here, so the only thing you need to do is set matrixNeedsUpdate. In tick where you area actually reading `root.matrix` is where `updateMatricies` needs to get called. I don't think a decompose is needed until later in tick where it is already being done ` root.matrix.decompose(root.position, root.quaternion, root.scale);`. Here its a noop since yo literally just set vector3 position, then updated the matrix with that position, then decomposed the position you just set back into the vector3 position.

src/components/teleporter.js
```diff
@@ -231,7 +198,9 @@ AFRAME.registerComponent("teleporter", {
         return;
       }
 
-      this.teleport();
+      this.characterController =
```
netpro2k: Hmm, I would think this would be safe to resolve in init, at least the entity itself. And insofar as this remains a component that lives on an entity we maybe might as well use a schema property for this. Not particularly fond of it, but also don't love these self-caching queryselectors either so...

src/components/character-controller.js
```diff
@@ -88,11 +86,30 @@ AFRAME.registerComponent("character-controller", {
     this.el.emit("snap_rotate_right");
   },
 
-  handleTeleport: function(event) {
-    const position = event.detail.newPosition;
-    const navPosition = event.detail.hitPoint;
-    this.resetPositionOnNavMesh(position, navPosition, this.el.object3D);
-  },
+  // We assume the rig is at the root, and its local position === its world position.
+  teleportTo: (function() {
```
netpro2k: A quick not about what it means to teleport with regards to the rig and the players head might be helpful.

PR #1077, opened 2019-03-14
Remove a bunch of little recurring work

netpro2k:
- Don't update the FPS counter when in VR
- Don't poll for "support" availability. This also stops the error log spam for 404s (which presumably also then later sends another network request to Sentry)
- Don't poll to detect concurrent loads, instead use the `storage` event.

### DIFF
--- a/src/utils/concurrent-load-detector.js
+++ b/src/utils/concurrent-load-detector.js
@@ -1,44 +1,12 @@
-// Detects if another instance of ConcurrentLoadDetector is start()'ed by in the same local storage
-// context with the same instance key. Once a duplicate run is detected this will not fire any additional
-// events.
-
-const LOCAL_STORE_KEY = "___concurrent_load_detector";
-import { EventTarget } from "event-target-shim";
-
-export default class ConcurrentLoadDetector extends EventTarget {
-  constructor(instanceKey) {
-    super();
-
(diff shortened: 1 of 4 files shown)


src/utils/concurrent-load-detector.js
```diff
@@ -1,44 +1,12 @@
...
-    const maxStartedAt = new Date(currentState.started_at);
-
-    if (maxStartedAt.getTime() !== this.startedAt.getTime()) {
-      this.dispatchEvent(new CustomEvent("concurrentload"));
-      this.stop();
-    }
-  };
+// Stores a key in localStorage and registeres a listener for the same key being set (by another tab).
+// When detected, fire a "concurrentload" event.
+export default function detectConcurrentLoad(instanceKey = "global") {
```
johnshaughnessy: We don't use this `instanceKey` so I suggest we remove it until we want it.

src/utils/concurrent-load-detector.js
```diff
@@ -1,44 +1,12 @@
...
+  const key = `___concurrent_load_detector_${instanceKey}`;
+  localStorage.setItem(key, JSON.stringify({ started_at: new Date() }));
```
johnshaughnessy: question: do we need to store the date?

PR #1078, opened 2019-03-15
Remove proxying for the 'loading' scene

mqp:
There's no reason to CORS proxy this, and the code as written would be incorrect with `HUBS_PROXY` set to anything else.


gfodor:
Does it still work in dev and via hubs.local if you remove this? I believe I added the proxying to support a non-prod case.

mqp:
You're right, although IMO that's a bug and all content on uploads-prod.reticulum.io should be CORS accessible to anyone.

Changed to do the "right thing" so that it proxies only if necessary.

Issue #1079, opened 2019-03-15
Oculus Go: hub.link alpha buttons are interpretted as numbers


johnshaughnessy:
When you are on hubs.link and press "A B C", the numbers "1 2 3" are entered instead.

PR #1080, opened 2019-03-15
Disable link page touchscreen fix because it breaks Go

johnshaughnessy:
Fixes https://github.com/mozilla/hubs/issues/1079

I don't understand the details of how https://github.com/mozilla/hubs/pull/1042 fixed https://github.com/mozilla/hubs/issues/606, but it appears that it regressed the entry flow on Go. Since this is the primary flow for Go users, I am disabling the touchscreen fix which can be solved later.


johnshaughnessy:
Closed in favor of a real fix https://github.com/mozilla/hubs/pull/1081

PR #1087, opened 2019-03-19
Add Discord UI accoutrements

mqp:
These help users notice when a Discord bridge is in the room, so that they know to expect their chat to be bridged to Discord.

Still to do: UI indicating that pinned media is also bridged.


src/react-components/ui-root.js
```diff
@@ -963,6 +973,10 @@ class UIRoot extends Component {
     const pendingMessageFieldHeight = textRows * 28 + 20 + "px";
     const hasPush = navigator.serviceWorker && "PushManager" in window;
     const promptForNameAndAvatarBeforeEntry = !this.props.store.state.activity.hasChangedName;
+    const discordBridges = this.discordBridges();
```
gfodor: perhaps this stuff should be factored in a way that will deal nicely with subsequent non-discord bridges?

src/react-components/ui-root.js
```diff
@@ -1037,11 +1051,7 @@ class UIRoot extends Component {
                     e.target.blur();
                   }
                 }}
-                placeholder={
-                  this.occupantCount() <= 1
-                    ? "Nobody is here yet..."
-                    : `Send message to ${this.occupantCount() - 1} other${this.occupantCount() - 1 > 1 ? "s" : ""}...`
-                }
+                placeholder={this.occupantCount() <= 1 ? "Nobody is here yet..." : messageEntryPlaceholder}
```
gfodor: this regresses the UX which informs the user of the number of others in the room in the placeholder. this message should likely say "Send message to X others (+ #channel)"

src/react-components/ui-root.js
```diff
@@ -1414,6 +1424,11 @@ class UIRoot extends Component {
     // Allow scene picker pre-entry, otherwise wait until entry
     const showMediaBrowser = mediaSource && (mediaSource === "scenes" || this.state.entered);
 
+    const discordBridges = this.discordBridges();
+    const messageEntryPlaceholder = discordBridges
+      ? `Send to hub (+ ${discordBridges.map(ch => "#" + ch).join(", ")})...`
+      : "Send to hub...";
```
gfodor: We don't use the term `hub`, so should be `room` as before. Also, my 2c is the message should be:

`Send to room and #foo...`

src/hub.js
```diff
@@ -810,6 +810,7 @@ document.addEventListener("DOMContentLoaded", async () => {
 
     hubPhxPresence.onJoin((sessionId, current, info) => {
       const meta = info.metas[info.metas.length - 1];
+      console.log(sessionId, current, info);
```
gfodor: Intended?

src/react-components/ui-root.js
gfodor: (this text, along with the 'nobody is here yet' state, was put in since user feedback was it was confusing previously where there was no indication of who the messages were going to go to)

src/react-components/ui-root.js
mqp: How it is seems good to me. If you wanted to make an abstraction on top of the presence entries that provides some functions for getting out useful information, I'm sure you could do so, but I don't know the best way to do it right now.

src/react-components/ui-root.js
mqp: I would address that sentiment by making the visual presentation of the box look more like a chat instead of a mystery text box, but if you like the other text then I will change it back.

src/react-components/ui-root.js
mqp: Oops, I can never remember which we use.

src/hub.js
mqp: Nope.

src/react-components/ui-root.js
gfodor: i found the old text to be nicer because it conveyed some presence info beyond 0 or 1, so i'd advocate to switching it back

gfodor:
one thing i thought of -- the tip you've added in the pause menu may collide with the "focus" and "track" buttons, which are only visible when there is an active camera. we may just for now want to shift the tip upwards to unblock but those buttons probably can go somewhere else sometime.

PR #1091, opened 2019-03-21
Add interaction system

johnshaughnessy:
This PR introduces the `interaction` system, as well as several small systems that take advantage of it. The `interaction` system determines which entity are `held` and `hovered` by the player's right hand, left hand, and right remote. Each of our interactions (creating constraints, stretching, clicking buttons, hovering on menus, getting haptic feedback, making objects out of spawners) are handled in a separate system which is controlled by the `hubs-system` tick: https://github.com/mozilla/hubs/pull/1091/files#diff-9621a9166e4415759db3c35dd3b2b8ccR1


src/hub.html
```diff
@@ -208,32 +209,25 @@
                 >
                 </a-entity>
             </template>
-
+/
```
johnshaughnessy: delete

johnshaughnessy:
Failed attempt to simplify and also to implement two handed scaling:
https://github.com/mozilla/hubs/compare/try-replacing-super-hands...try-replacing-super-hands-bad-commits

johnshaughnessy:
Some things I know I want to improve, but that I don't think block this merge:

- [ ] Reintroduce grab / release sound effects
- [ ] Move `is-remote-hover-target` and `is-hand-collision-target` into `tags`.
- [ ] Move super-spawner code out of interaction system
- [ ] Add `isGrabbable` to tags for use in interaction system
- [ ] Reintroduce haptic feedback
- [ ] Don't need to use hover events -- can just do the work directly

src/systems/hubs-systems.js
```diff
@@ -0,0 +1,30 @@
+import { CursorTargettingSystem } from "./cursor-targetting-system";
+import { ConstraintsSystem } from "./constraints-system";
+import { TwoPointStretchingSystem } from "./two-point-stretching-system";
+import { SingleActionButtonSystem, HoldableButtonSystem, HoverButtonSystem } from "./button-systems";
+import { HoverMenuSystem } from "./hover-menu-system";
```
johnshaughnessy: There's no point in calling all of these "systems" and then calling their "tick" function. They might as well just be functions who are passed the data they need in order to have the side effects they want to have. (Usually, that data is the interaction system state.)

src/systems/interactions.js
```diff
@@ -0,0 +1,199 @@
...
+      }
+    } else {
+      state.hovered = options.hoverFn.call(this, options.entity.body);
+      if (state.hovered) {
+        if (userinput.get(options.grabPath)) {
+          const offersConstraint =
+            state.hovered.components.tags && state.hovered.components.tags.data[options.constraintTag];
+          const isHoldableButton = state.hovered.components.tags && state.hovered.components.tags.data.holdableButton;
+          const superSpawner = state.hovered.components["super-spawner"];
+          if (offersConstraint || isHoldableButton) {
```
johnshaughnessy: This could instead check a tag called `holdable`

src/systems/interactions.js
```diff
@@ -0,0 +1,199 @@
...
+            if (
+              this.el.is("frozen") ||
+              !state.hovered.components.pinnable ||
+              !state.hovered.components.pinnable.data.pinned
```
johnshaughnessy: This could be implemented in `pinnable` if it knew to toggle a tag called `holdingBlocked`. I would say that it could toggle `holdable`, but the thought of more than one thing wanting to toggle `holdable` gives me cold sweats at night.

src/systems/userinput/userinput.js
```diff
@@ -163,7 +163,8 @@ AFRAME.registerSystem("userinput", {
   },
 
   toggleSet(set, value) {
-    this.pendingSetChanges.push({ set, value });
+    const b = !!value;
```
johnshaughnessy: (We only care about whether the `value` passed in is truthy.)

src/components/networked-counter.js
```diff
@@ -5,77 +5,54 @@
  */
 AFRAME.registerComponent("networked-counter", {
   schema: {
-    max: { default: 3 },
-    ttl: { default: 0 },
```
johnshaughnessy: I removed `ttl` because we weren't using it and that made it easier to rework, but I can bring it back if it's something we care about.

src/components/scalable-when-grabbed.js
```diff
@@ -1,21 +1,14 @@
...
-
     const userinput = AFRAME.scenes[0].systems.userinput;
-    const deltaScale = userinput.get(grabberToPath[grabber.id]);
+    const interaction = AFRAME.scenes[0].systems.interaction;
+    if (interaction.state.rightRemote.held !== this.el) return;
+    const deltaScale = userinput.get(paths.actions.cursor.scaleGrabbedGrabbable);
     if (!deltaScale) return;
 
-    this.el.object3D.scale.addScalar(deltaScale).clampScalar(0.01, 1000);
+    this.el.object3D.scale.addScalar(deltaScale).clampScalar(0.1, 100);
```
johnshaughnessy: These were the original clamp values but I changed them two weeks ago without anyone noticing and for no particular reason.

src/components/sticky-object.js
```diff
@@ -82,70 +93,3 @@ AFRAME.registerComponent("sticky-object", {
     }
   }
 });
-
-AFRAME.registerComponent("sticky-object-zone", {
```
johnshaughnessy: I removed this because I'm a bad person but if you reprimand me I will bring it back.

src/systems/button-systems.js
```diff
@@ -0,0 +1,83 @@
...
+
+function getHoverableButton(hovered) {
+  if (!hovered) return null;
+  if (
+    hovered.components["icon-button"] ||
+    hovered.components["text-button"] ||
+    hovered.components["pin-networked-object-button"]
+  )
+    return hovered;
+  // TODO: fix this so that we aren't looping thru children here. I just did this to accomodate the new rounded buttons
```
johnshaughnessy: The rounded buttons I'm referring to here are the ones that got added here: https://github.com/mozilla/hubs/pull/1124/files#diff-a796e11be3d0dfc0f893148f1086736dR245

There is probably a smarter way I could be doing this but I don't know what it is.

src/systems/constraints-system.js
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  tickInteractor(options, state, prevState) {
+    if (prevState.held === state.held) {
+      if (
+        state.held &&
+        state.held.components.tags &&
+        state.held.components.tags.data[options.constraintTag] &&
+        prevState.spawning &&
+        !state.spawning
```
johnshaughnessy: It would be great to get rid of this `spawning` tag, but I don't know how to do it.

src/systems/cursor-targetting-system.js
```diff
@@ -0,0 +1,44 @@
...
+      this.populateEntities(this.targets);
+      this.dirty = false;
+    }
+
+    this.rightRemote.components["cursor-controller"].tick2(t);
+  }
+
+  populateEntities(targets) {
+    targets.length = 0;
+    // TODO: Do not querySelectorAll on the entire scene every time anything changes!
```
johnshaughnessy: This isn't new, this was just taken from cursor controller... Still, this is slow and it happens all the time!!

src/systems/two-point-stretching-system.js
```diff
@@ -0,0 +1,55 @@
...
+        leftHand.held !== this.stretched
+      ) {
+        this.initialStretchDistance = distanceBetweenStretchers(stretcherLeft, stretcherRight);
+        this.stretched = leftHand.held;
+        this.initialScale.copy(this.stretched.object3D.scale);
+      }
+
+      this.stretched.object3D.scale
+        .copy(this.initialScale)
+        .multiplyScalar(distanceBetweenStretchers(stretcherLeft, stretcherRight) / this.initialStretchDistance);
```
johnshaughnessy: Marshall suggested that the amount you stretch an object should be such that the two points you begin touching are still the same two points on the model after you're done stretching. I suspect that would feel even better than this, but I'm not sure how to do that.

src/systems/userinput/resolve-action-sets.js
```diff
@@ -1,150 +1,142 @@
...
+  userinput.toggleSet(sets.rightHandHoldingCamera, rightHand.held && rightHand.held.components["camera-tool"]);
+  userinput.toggleSet(sets.rightHandHoldingInteractable, rightHand.held);
 
-  const cursorHoldingCamera =
-    !rightTeleporter.isTeleporting &&
-    cursorHand.has("grab-start") &&
-    cursorHand.get("grab-start").matches(".icamera, .icamera *");
+  userinput.toggleSet(
+    sets.cursorHoveringOnNothing,
+    !rightHand.held && !rightHand.hovered && !rightRemote.held && !rightRemote.hovered
```
johnshaughnessy: The contract for these action sets is pretty surprising. I only know it because I wrote them. Perhaps the interaction system should enforce these rules such that (for example) `rightRemote.hovered` is guaranteed to be `null` if anything is in your right hand or right remote, or your right hand is hovering on anything. That way this line becomes,
`userinput.toggleSet(sets.cursorHoveringOnNothing, !rightRemote.hovered)`

src/systems/interactions.js
```diff
@@ -0,0 +1,205 @@
...
+      this.rightHandTeleporter || document.querySelector("#player-right-controller").components["teleporter"];
+
+    this.tickInteractor(this.options.leftHand, this.state.leftHand);
+    if (!this.state.rightRemote.held) {
+      this.tickInteractor(this.options.rightHand, this.state.rightHand);
+    }
+
+    if (!this.state.rightHand.held && !this.state.rightHand.hovered) {
+      this.tickInteractor(this.options.rightRemote, this.state.rightRemote);
+    }
```
johnshaughnessy: perhaps should be:
```
if (this.state.rightHand.held || this.state.rightHand.hovered) {
  this.state.rightRemote.held = null;
  this.state.rightRemote.hovered = null;
} else {
  this.tickInteractor(this.options.rightRemote, this.state.rightRemote);.
}
```

src/systems/interactions.js
```diff
@@ -0,0 +1,205 @@
...
+        this.state.leftHand.hovered,
+        "\nrightHand held",
+        this.state.rightHand.held,
+        "\nrightHand hovered",
+        this.state.rightHand.hovered,
+        "\nrightRemote held",
+        this.state.rightRemote.held,
+        "\nrightRemote hovered",
+        this.state.rightRemote.hovered
+      );
```
johnshaughnessy: I don't want 
```
console.log(this.state);
```
because I want to make sure I see a _snapshot_ of the values, rather than getting a "live" reference that will only become a snapshot once clicked on in the console.

src/systems/interactions.js
johnshaughnessy: Did it

src/systems/userinput/userinput.js
netpro2k: Not awful the sanitize this input, but I would ditch the temp variable

johnshaughnessy:
The name "right remote" might be changed to "right cursor". I just chose a new name to make it clear that it is part of the new system so that it is easy to tell what came before and after as we evolve the code in the near future.

src/components/media-views.js
```diff
@@ -634,8 +634,9 @@ AFRAME.registerComponent("media-video", {
     if (!this.video) return;
 
     const userinput = this.el.sceneEl.systems.userinput;
+    const interaction = this.el.sceneEl.systems.interaction;
     const volumeMod = userinput.get(paths.actions.cursor.mediaVolumeMod);
-    if (this.el.is("hovered") && volumeMod) {
+    if (interaction.rightRemoteHoverTarget === this.el && volumeMod) {
```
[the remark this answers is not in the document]
johnshaughnessy: ✔️

src/systems/interactions.js
```diff
@@ -0,0 +1,162 @@
+/* global AFRAME Ammo NAF */
+import { paths } from "./userinput/paths";
+
+const handCollisionTargets = new Map();
+AFRAME.registerComponent("is-hand-collision-target", {
+  init: function() {
+    handCollisionTargets.set(this.el.object3D.uuid, this.el);
+  }
+});
```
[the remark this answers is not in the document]
johnshaughnessy: ✔️

src/systems/interactions.js
```diff
@@ -0,0 +1,162 @@
...
+        hoverFn: findHandCollisionTargetForHand
+      },
+      rightHand: {
+        entity: document.querySelector("#player-right-controller"),
+        grabPath: paths.actions.rightHand.grab,
+        dropPath: paths.actions.rightHand.drop,
+        constraintTag: "offersHandConstraint",
+        hoverFn: findHandCollisionTargetForHand
+      },
+      rightRemote: {
```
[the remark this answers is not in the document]
johnshaughnessy: verified

src/components/visibility-while-frozen.js
```diff
@@ -23,13 +23,16 @@ AFRAME.registerComponent("visibility-while-frozen", {
...
-      if (hoverableSearch.getAttribute("hoverable") !== null) {
+      if (hoverableSearch.getAttribute("is-remote-hover-target") !== null) {
         this.hoverable = hoverableSearch;
         break;
       }
 
       hoverableSearch = hoverableSearch.parentNode;
     }
+    if (!this.hoverable) {
+      console.error("Didn't find a remote hover target.");
```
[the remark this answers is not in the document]
johnshaughnessy: Apparently yes

src/systems/hubs-systems.js
```diff
@@ -0,0 +1,36 @@
...
+    systems.interaction.tick2();
+    this.superSpawnerSystem.tick();
+    this.cursorTargettingSystem.tick(t);
+    this.constraintsSystem.tick();
+    this.twoPointStretchingSystem.tick();
+    this.singleActionButtonSystem.tick();
+    this.holdableButtonSystem.tick();
+    this.hoverButtonSystem.tick();
+    this.hoverMenuSystem.tick();
+    this.hapticFeedbackSystem.tick(this.twoPointStretchingSystem, this.singleActionButtonSystem.didInteractThisFrame);
```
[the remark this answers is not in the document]
johnshaughnessy: Agreed. The best thing I can think of is to pass the state they read into them. I don't want to write a comment for the usual reasons. I don't think putting some of them in a block (as mqp suggested) communicates anything about them in particular.

src/systems/interactions.js
```diff
@@ -0,0 +1,162 @@
...
+        state.held = null;
+      }
+    } else {
+      state.hovered = options.hoverFn.call(this, options.entity.body);
+      if (
+        state.hovered &&
+        state.hovered.components.tags &&
+        state.hovered.components.tags.data.isHoldable &&
+        userinput.get(options.grabPath) &&
+        (this.el.is("frozen") || !state.hovered.components.pinnable || !state.hovered.components.pinnable.data.pinned)
```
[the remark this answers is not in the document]
johnshaughnessy: Conceptually I like the idea that `isHoldable` should just be made to be false or true depending on whether we actually want it to be holdable or not this frame. That approach seems a bit tricky to get right, since it depends on the state's frozen state, the object's pinnable state, and whether the object was `isHoldable` to begin with, so for now I'm going to do this check here. As soon as another part of the app wants to check the `isHoldable` tag for some reason, let's revisit this decision.

As for breaking up the boolean, I'm not convinced that buys us anything.

src/systems/hubs-systems.js
```diff
@@ -0,0 +1,36 @@
...
+    this.hoverButtonSystem = new HoverButtonSystem();
+    this.hoverMenuSystem = new HoverMenuSystem();
+    this.superSpawnerSystem = new SuperSpawnerSystem();
+    this.hapticFeedbackSystem = new HapticFeedbackSystem();
+  },
+
+  tick(t) {
+    const systems = AFRAME.scenes[0].systems;
+    systems.userinput.tick2();
+    systems.interaction.tick2();
```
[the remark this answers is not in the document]
johnshaughnessy: Will do

src/systems/interactions.js
```diff
@@ -0,0 +1,162 @@
...
+function findHandCollisionTarget(o) {
+  if (!o) return null;
+  const target = handCollisionTargets.get(o.uuid);
+  return target || findHandCollisionTarget(o.parent);
+}
+function findHandCollisionTargetForHand(body) {
+  const driver = AFRAME.scenes[0].systems.physics.driver;
+  const numManifolds = driver.dispatcher.getNumManifolds();
+  const handPtr = Ammo.getPointer(body);
+  for (let i = 0; i < numManifolds; i++) {
```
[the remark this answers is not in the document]
johnshaughnessy: I would (and did), but aframe-physics-system stores its collisions in a Map whose values are Sets, which means that the code would look like this:
```
function findHandCollisionTargetForHand(body) {
  const driver = AFRAME.scenes[0].systems.physics.driver;
  const numManifolds = driver.dispatcher.getNumManifolds();
  const handPtr = Ammo.getPointer(body);
  if (driver.collisions.has(handPtr)) {
    const collisions = driver.collisions.get(handPtr);
    if (collisions.size) {
      const collidedPtr = collisions.values().next().value;
      const object3D = driver.els.get(collidedPtr).object3D;
      if (object3D.el && object3D.el.components.tags && object3D.el.components.tags.data.isHandCollisionTarget) {
        return object3D.el;
      }
      return null;
    }
  }
  return null;
}
```
The necessary call to `collisions.values().next()` creates garbage (a set iterator and an object), so instead I will leave it as is.

src/systems/hubs-systems.js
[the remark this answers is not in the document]
johnshaughnessy: Noted in https://github.com/mozilla/hubs/pull/1203

src/components/hoverable-visuals.js
```diff
@@ -25,20 +25,18 @@ AFRAME.registerComponent("hoverable-visuals", {
   tick(time) {
```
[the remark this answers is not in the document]
johnshaughnessy: Noted in https://github.com/mozilla/hubs/pull/1203

PR #1093, opened 2019-03-22
Fix scaling pen/spawner bug attempt #3

gfodor:
This PR addresses the initial scale issues with spawners + pen, fixes a bug where sometimes the first duck spawn can be mis-scaled and tightens up the animations as well.

- If the template has a `scale` component, we need to wait a tick for it to be applied (this seems to be why the pen scale was wrong.)

- Fixes up the condition where the box animation ends early, but there has been no component (scale or spawner) which has set the final object scale

I confirmed this seems to result in proper convex hulls on the pen, spawned objects, and /duck's, but not sure if there are other things to test.


InfiniteLee:
lgtm

PR #1097, opened 2019-03-22
Fix a few issues with new avatar skinning.

netpro2k:
- Fix caching issues by requesting the avatar url from the server, which now contains a cache busting timestamp from when the avatar was last updated. Requires https://github.com/mozilla/reticulum/pull/129
- Fix head dispreading in camera tool view when changing avatars (skinned or otherwise)
- Fix an unrelated issue with camera tool spamming GL errors since it was trying to render its own viewfinder.

### DIFF
--- a/src/systems/camera-tools.js
+++ b/src/systems/camera-tools.js
@@ -17,7 +17,7 @@ AFRAME.registerSystem("camera-tools", {
   },
 
   avatarUpdated() {
-    this.cameraEls.forEach(el => delete el.components["camera-tool"].playerHead);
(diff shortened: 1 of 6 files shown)


src/systems/camera-tools.js
```diff
@@ -16,6 +16,10 @@ AFRAME.registerSystem("camera-tools", {
     this.myCamera = null;
   },
 
+  avatarUpdated() {
+    this.cameraEls.forEach(el => delete el.components["camera-tool"].playerHead);
```
gfodor: this feels a little magical, maybe push it into a function `onAvatarUpdated` in the camera tool component?

PR #1098, opened 2019-03-23
Fix for issue that broke delayed spawning from media browser

gfodor:
n/t


src/scene-entry-manager.js
```diff
@@ -452,7 +456,7 @@ export default class SceneEntryManager {
       if (entry.type === "scene_listing" && this.hubChannel.permissions.update_hub) return;
 
       // If user has HMD lifted up, delay spawning for now. eventually show a modal
-      const delaySpawn = this._in2DInterstitial && !isMobileVR;
+      const delaySpawn = isIn2DInterstitial() && !isMobileVR;
```
johnshaughnessy: Ah woops. Got `this.`ed

PR #1102, opened 2019-03-26
Discord OAuth flow

brianpeiris:
Adds a discord OAuth dialog and flow for hubs bound to discord channels

firefox_2019-03-25_17-39-19

Goes with https://github.com/mozilla/reticulum/pull/131


src/hub.js
```diff
@@ -779,14 +781,30 @@ document.addEventListener("DOMContentLoaded", async () => {
...
-      hubChannel.setPermissionsFromToken(data.perms_token);
+
+      let permsToken;
+      const oauthFlowPermsToken = Cookies.get(OAUTH_FLOW_PERMS_TOKEN_KEY);
+      if (oauthFlowPermsToken) {
+        permsToken = oauthFlowPermsToken;
+        Cookies.remove(OAUTH_FLOW_PERMS_TOKEN_KEY);
+      } else {
+        permsToken = data.perms_token;
+      }
```
gfodor: I guess I don't understand the point of this perms token cookie being set by the OAuth controller, vs the usual `perms_for_account` evaluator in `Ret.Hub`, mind explaining? Also, will the 5 minute token refresh work OK still?

brianpeiris: As discussed, the secondary cookie is for discord users that don't have a verified email. This cookie will grant join_hub permissions and you're right that the user will lose that permission when the token is refreshed, but it works out since they only need the join_hub permission at join time. Not sure there's a cleaner way to solve that though.

brianpeiris: Turns out I was wrong about how the SFU used the join token as we discovered on Monday. I accounted for this by assigning the discord user id to the socket so that we can refresh these temporary perms token as well: https://github.com/mozilla/reticulum/pull/131/commits/14ac36fcdde1a173653d4ebdc5d359d98f94b52b

PR #1106, opened 2019-03-26
Fix hand rotations for new skinned avatar skeleton

netpro2k:
The old avatar skeleton had incorrect bone rotations for the right hand. This has been fixed in the updated skeleton we use for skinning, but this requires a different set of rotations to orient each hand correctly. This contains a bit of a hack to switch these rotations on the fly, which can go away once we remove the legacy avatars (when we ship an avatar picker that aligns with the new scene picker).

### DIFF
--- a/src/systems/userinput/bindings/keyboard-mouse-user.js
+++ b/src/systems/userinput/bindings/keyboard-mouse-user.js
@@ -3,6 +3,8 @@ import { sets } from "../sets";
 import { xforms } from "./xforms";
 import { addSetsToBindings } from "./utils";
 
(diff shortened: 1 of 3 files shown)


src/systems/userinput/bindings/keyboard-mouse-user.js
```diff
@@ -3,6 +3,8 @@ import { sets } from "../sets";
 import { xforms } from "./xforms";
 import { addSetsToBindings } from "./utils";
 
+// import { Pose } from "../pose";
```
johnshaughnessy: Remove

PR #1109, opened 2019-03-26
Fix bug where visual effects bounding sphere not up-to-date

gfodor:
The visual hover effect relies upon a bounding sphere for the object being hovered over, which previously was set once at media loading time. If the object was scaled, the bounding sphere was out-of-date. This updates the `hoverable-visuals` component to maintain the bounding sphere by checking if the scale has changed on the object on a given tick.

### DIFF
--- a/src/components/hoverable-visuals.js
+++ b/src/components/hoverable-visuals.js
@@ -1,8 +1,6 @@
 const interactorOneTransform = [];
 const interactorTwoTransform = [];
 
-const boundingBox = new THREE.Box3();
-
 /**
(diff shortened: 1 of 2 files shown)


brianpeiris:
This seems like a bad idea. I intentionally tried to limit the bounding box calculation to a one-time operation when I wrote the shader. Re-calculating the bounding box on tick would be very expensive for high-poly objects, especially if the user was continuously scaling it with their controllers, since setFromObject iterates over verticies.

gfodor:
closed in favor of #1113

PR #1111, opened 2019-03-26
Allow skipping code paths relying upon webRTC

gfodor:
This PR adds the `skip_webrtc` flag which can be used to bypass WebRTC/UserMedia API calls in the startup/entry flow of Hubs, so Hubs can be tested in Servo.


gfodor:
Closing for now

PR #1112, opened 2019-03-26
Custom Avatar Skins

netpro2k:
It is now possible to to upload custom skins to the default Hubs avatar! For a super easy way to customize your avatar, you can simply upload a new base color map - or dig in a bit deeper and change the emissive, normal, and ORM maps for even more control. This is only the first step in our planned avatar customization features, with more features planned to grant you even greater control over your appearance in Hubs! As always, you can still use a completely custom GLTF file hosted anywhere on the web as your avatar if you want the greatest control over your avatar's look and feel.

image

This PR just removes the feature flag so that its always on. It also renames `advancedAvatarEditor` to `avatarEditorDebug`.

### DIFF
--- a/src/react-components/avatar-editor.js
+++ b/src/react-components/avatar-editor.js
@@ -18,7 +18,7 @@ export default class AvatarEditor extends Component {
     onSignIn: PropTypes.func,
     onSignOut: PropTypes.func,
     signedIn: PropTypes.bool,
-    advanced: PropTypes.bool,
+    debug: PropTypes.bool,
     onAvatarChanged: PropTypes.func,
     saveStateAndFinish: PropTypes.func
   };
@@ -26,6 +26,7 @@ export default class AvatarEditor extends Component {
   constructor(props) {
     super(props);
     this.state = {};
+    // Blank avatar, used to create base avatar
     // this.state = { avatar: { name: "Base bot avatar", files: {} } };
 
     this.inputFiles = {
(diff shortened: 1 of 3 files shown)


src/react-components/avatar-editor.js
```diff
@@ -108,6 +109,27 @@ export default class AvatarEditor extends Component {
       URL.revokeObjectURL(gltfUrl);
 
       const { content, body } = parser.extensions.KHR_binary_glTF;
+
+      // Inject hubs components on upload. Used to create base avatar
+      // const gltf = parser.json;
```
johnshaughnessy: Uncomment or delete?

misslivirose:
I can't wait to play around with this! 

Copy suggestions for What's New: 
It is now possible to to upload custom skins to the default Hubs avatar! For a super easy way to customize your avatar, you can simply upload a new base color map - or dig in a bit deeper and change the emissive, normal, and ORM maps for even more control. This is only the first step in our planned avatar customization features, with more features planned to grant you even greater control over your appearance in Hubs! As always, you can still use a completely custom GLTF file hosted anywhere on the web as your avatar if you want the greatest control over your avatar's look and feel.

src/react-components/avatar-editor.js
netpro2k: There is manual work involved in updating the "base" avatar and the simplest way I know how to do it is to just hack up the form. I could extract this elsewhere but given its temporary nature it doesn't seem worth the effort, so its relegated to a comment based tool heh.

PR #1116, opened 2019-03-27
Add hack to deal with videos in Oculus Browser when exiting VR mode

gfodor:
This PR adds a hack because Oculus Browser pauses all videos in the scene when exiting VR mode. It also flips on a 1s delay on spawning on for mobile VR browsers because when you exit VR mode the head orientation is locked (so if you were looking up at the hud, the object you spawn will be above your head).

### DIFF
--- a/src/scene-entry-manager.js
+++ b/src/scene-entry-manager.js
@@ -458,10 +458,11 @@ export default class SceneEntryManager {
       if (entry.type === "scene_listing" && this.hubChannel.permissions.update_hub) return;
 
       // If user has HMD lifted up, delay spawning for now. eventually show a modal
-      const delaySpawn = isIn2DInterstitial();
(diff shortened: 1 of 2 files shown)


src/scene-entry-manager.js
```diff
@@ -458,10 +458,11 @@ export default class SceneEntryManager {
       if (entry.type === "scene_listing" && this.hubChannel.permissions.update_hub) return;
 
       // If user has HMD lifted up, delay spawning for now. eventually show a modal
-      const delaySpawn = isIn2DInterstitial() && !isMobileVR;
+      const spawnDelay = isIn2DInterstitial() ? (isMobileVR ? 1000 : 3000) : 0;
```
johnshaughnessy: If I recall correctly the problem was that things would spawn up in the air above the user if their headset was on their forehead. If that's the case, fixing the spawn location seems like a better solution than adding a delay

PR #1118, opened 2019-03-27
Feature/media volume for static media

gfodor:
This PR updates the behavior of Spoke-embedded media so that videos will have volume controls -- it seems right that we should always allow users to adjust the volume of videos in the room even if they can't control them otherwise.

This PR also removes the visual effects on hovering of controllable static media -- on balance this visual effect is likely not necessary and would be distracting for large, fixed videos etc where only the volume controls are accessible.


src/gltf-component-mappings.js
```diff
@@ -131,13 +131,24 @@ AFRAME.GLTFModelPlus.registerComponent("media", "media", (el, componentName, com
...
 function mediaInflator(el, componentName, componentData, components) {
+  let isControlled = true;
+
   if (components.networked) {
     // TODO: When non-hubs links can be traversed, make all link components controlled so you can open them.
-    const isControlled =
+    isControlled =
       componentData.controls || isHubsDestinationUrl(componentData.src) || isHubsDestinationUrl(componentData.href);
 
+    const hasVolume = componentName === "video";
```
gfodor: eventually `sound`

src/gltf-component-mappings.js
```diff
@@ -131,13 +131,18 @@ AFRAME.GLTFModelPlus.registerComponent("media", "media", (el, componentName, com
 });
 
 function mediaInflator(el, componentName, componentData, components) {
+  let isControlled = true;
```
netpro2k: Seems like you are only using this within the conditional, why define it as a let out here instead of a const inside like before?

src/hub.html
```diff
@@ -397,27 +396,29 @@
...
-                        visible="false"
-                    ></a-image>
-                    <a-image
-                        icon-button="image: #video-seek-forward; hoverImage: #video-seek-forward-hover;"
-                        scale="0.1 0.1 0.1"
-                        position="0.3 -0.125 0"
-                        class="ui video-seek-forward-button"
-                        visible="false"
-                    ></a-image>
+                    <a-entity class="video-playback">
```
netpro2k: Oof I hate to add another level here just for this. You already have a ref to back/forward/play.. Seems ok to just hide them directly.

src/gltf-component-mappings.js
gfodor: its used later on line 173

PR #1121, opened 2019-03-28
Three 103 + ImageBitmapLoader and texture cleanup after GPU upload

netpro2k:
TODO: 
- [x] merge https://github.com/MozillaReality/three.js/pull/20
- [x] merge https://github.com/MozillaReality/three.js/pull/21
- [x] merge https://github.com/MozillaReality/aframe/pull/13
- [x] update package.lock to point at new versions of `three` and `aframe`

### DIFF
--- a/src/components/camera-tool.js
+++ b/src/components/camera-tool.js
@@ -255,7 +255,9 @@ AFRAME.registerComponent("camera-tool", {
         delete sceneEl.object3D.onAfterRender;
         renderer.vr.enabled = false;
 
-        renderer.render(sceneEl.object3D, this.camera, this.renderTarget, true);
+        renderer.setRenderTarget(this.renderTarget);
+        renderer.render(sceneEl.object3D, this.camera);
+        renderer.setRenderTarget(null);
 
         renderer.vr.enabled = tmpVRFlag;
         sceneEl.object3D.onAfterRender = tmpOnAfterRender;
--- a/src/components/camera-tool.js
+++ b/src/components/camera-tool.js
@@ -105,7 +105,7 @@ AFRAME.registerComponent("camera-tool", {
       this.el.setObject3D("mesh", mesh);
 
       const width = 0.28;
-      const geometry = new THREE.PlaneGeometry(width, width / this.camera.aspect);
+      const geometry = new THREE.PlaneBufferGeometry(width, width / this.camera.aspect);
 
       if (enableCameraViewport) {
         const screen = new THREE.Mesh(geometry, material);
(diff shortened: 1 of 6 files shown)


src/components/gltf-model-plus.js
```diff
@@ -259,16 +259,7 @@ async function loadGLTF(src, contentType, preferredTechnique, onProgress) {
...
-    parser.parse(
-      (scene, scenes, cameras, animations, json) => {
-        resolve({ scene, scenes, cameras, animations, json });
-      },
-      e => {
-        reject(e);
-      }
-    )
-  );
+  const gltf = await new Promise(parser.parse.bind(parser));
```
johnshaughnessy: Is this due to THREE 103?

src/components/camera-tool.js
```diff
@@ -254,7 +254,9 @@ AFRAME.registerComponent("camera-tool", {
         delete sceneEl.object3D.onAfterRender;
         renderer.vr.enabled = false;
 
-        renderer.render(sceneEl.object3D, this.camera, this.renderTarget, true);
+        renderer.setRenderTarget(this.renderTarget);
+        renderer.render(sceneEl.object3D, this.camera);
+        renderer.setRenderTarget(null);
```
johnshaughnessy: Is this just a change due to Three 103?

src/components/media-views.js
```diff
@@ -758,7 +758,7 @@ AFRAME.registerComponent("media-image", {
         // invert the geometry on the x-axis so that all of the faces point inward
         geometry.scale(-1, 1, 1);
       } else {
-        geometry = new THREE.PlaneGeometry();
+        geometry = new THREE.PlaneBufferGeometry(1, 1, 1, 1, texture.flipY);
```
johnshaughnessy: See also:
```
src/react-components/chat-message.js:192:    const geometry = new THREE.PlaneGeometry();
src/components/camera-tool.js:108:      const geometry = new THREE.PlaneGeometry(width, width / this.camera.aspect);
```

src/components/camera-tool.js
netpro2k: Yep

src/components/gltf-model-plus.js
netpro2k: Yep

src/components/media-views.js
netpro2k: Oh nice, good catch. I had actually broken the chat messages since they were using texture loaded and not accounting for flipY. Camera tool wasn't broken since it doesn't use a TextureLoader (since this PlaneGeometry is for the preview view). It should be using PlaneBufferGeometry anyway, fixed.

PR #1122, opened 2019-03-28
Feature/scale tool

gfodor:
Adds a scale tool to the menu and also renames the rotate components to `xform` since it is now used for scaling.


src/components/xform-object-button.js
```diff
@@ -238,72 +244,86 @@ AFRAME.registerSystem("rotate-selected-object", {
...
+          v.set(0, 1, 0).applyQuaternion(CAMERA_WORLD_QUATERNION);
+        } else {
+          v.set(0, 1, 0);
+        }
+        q2.setFromAxisAngle(v, this.dxApplied);
+
+        this.target.quaternion.premultiply(q).premultiply(q2);
       } else {
-        v.set(0, 1, 0);
+        const scaleFactor =
```
gfodor: This is the main new control flow, everything else here is just renames

src/components/xform-object-button.js
```diff
@@ -16,12 +21,12 @@ const v2 = new THREE.Vector3();
 const q = new THREE.Quaternion();
 const q2 = new THREE.Quaternion();
 
-AFRAME.registerComponent("rotate-button", {
+AFRAME.registerComponent("xform-button", {
```
johnshaughnessy: Would prefer `transform` to `xform`. We would have called `xforms` in the `userinput` system `transforms` or `transformFunctions` but we didn't want to risk confusion with the "position/rotation/scale" interpretation of "transform". Here that IS what we mean, so I think it fits better.

gfodor: ah good point. i was trying to save characters since this term is used in many places, but i agree the conflict with the user input concept makes it worth changing.

PR #1148, opened 2019-03-30
Extract chatboxes into separate React components

mqp:
Similar to #1147. Cleans up the root React component so that it's less scary, and prevents a full re-render on each character typed into chat.


gfodor:
thanks for doing this! please make sure to test on mobile

PR #1150, opened 2019-03-30
Rewrite spawning control flow to deal with animations properly

gfodor:
Makes a bunch of changes to the spawning control flow to ensure animations are dealt with properly. Also cleans up the spawner UX.

### DIFF
--- a/src/components/media-loader.js
+++ b/src/components/media-loader.js
@@ -9,11 +9,12 @@ import {
   generateMeshBVH
 } from "../utils/media-utils";
 import { addAnimationComponents } from "../utils/animation";
-
 import "three/examples/js/loaders/GLTFLoader";
 import loadingObjectSrc from "../assets/LoadingObject_Atom.glb";
 
-const SHAPE = require("aframe-physics-system/src/constants").SHAPE;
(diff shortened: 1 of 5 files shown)


src/components/media-loader.js
```diff
@@ -102,27 +98,56 @@ AFRAME.registerComponent("media-loader", {
...
+      delete this.loadingClip;
     }
     delete this.showLoaderTimeout;
+    this.removeShape("loader");
+  },
 
-    this.waitForMediaScaleReady(() => {
-      this.removeShape("loader");
-    });
+  updateShape(shapeType) {
```
InfiniteLee: Would prefer this to take an `options` or `props` object like so:

```
    const options = {type: SHAPE.BOX,
        halfExtents: { x: 0.5, y: 0.5, z: 0.02 },
        margin: 0.1,
        fit: FIT.MANUAL
    }
    updateShape(options);
...
   updateShape(shapeOptions) {
        if (this.el.getAttribute("ammo-shape")) {
    });
      this.el.removeAttribute("ammo-shape");
    }
      this.el.setAttribute("ammo-shape", shapeOptions);
    }
```
Obviously also do the same where this is being called in media-views

PR #1156, opened 2019-04-01
Add "Leave Room" button to VR HUD

gfodor:
Adds the "Leave Room" UX from 2d mode into VR, and also re-arranges the HUD in VR a bit.


src/components/leave-room-button.js
```diff
@@ -0,0 +1,16 @@
+AFRAME.registerComponent("leave-room-button", {
+  init() {
+    this.onClick = () => {
+      console.log("click");
```
brianpeiris: Remove debug log

src/components/leave-room-button.js
```diff
@@ -0,0 +1,16 @@
...
+      this.el.sceneEl.emit("leave_room_requested");
```
brianpeiris: nit: We seem to alternate between using hyphens and underscores in our event names. Would be nice to settle on one of them. Looks like underscores are winning, so just something to keep an eye out for in other reviews.

src/components/leave-room-button.js
gfodor: oops, nice catch

src/components/leave-room-button.js
gfodor: ah yeah i've noticed this too

PR #1159, opened 2019-04-01
Update copy for desktop VR interstitial

gfodor:
The current copy in the notification when in desktop VR mode implies you have to lift your headset to proceed, when in fact you can also switch over to virtual desktop in Dash, etc. This updates the copy.

### DIFF
--- a/src/hub.html
+++ b/src/hub.html
@@ -881,7 +881,7 @@
                     textColor: #fff;
                     backgroundColor: #2f80ed;"
                 position="0 0 0.01">
-                <a-entity text="value:View desktop to continue; width:1; align:center;" position="0 0 0.01"></a-entity>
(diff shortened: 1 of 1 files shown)


src/hub.html
```diff
@@ -881,7 +881,7 @@
                     textColor: #fff;
                     backgroundColor: #2f80ed;"
                 position="0 0 0.01">
-                <a-entity text="value:Lift your headset to continue; width:1; align:center;" position="0 0 0.01"></a-entity>
+                <a-entity text="value:View your desktop to continue; width:1; align:center;" position="0 0 0.01"></a-entity>
```
brianpeiris: Suggestion: "Continue in your browser tab"

gfodor: I think I like this copy better for now, super hard to know what connects without user testing ofc. We should figure out a way to see people go through this flow

PR #1168, opened 2019-04-02
Add feedback button

gfodor:
Adds a feedback button going to google forms.

### DIFF
--- a/src/react-components/settings-menu.js
+++ b/src/react-components/settings-menu.js
@@ -151,7 +151,7 @@ export default class SettingsMenu extends Component {
               <a href="https://discord.gg/wHmY4nd" target="_blank" rel="noreferrer noopener">
                 <FormattedMessage id="settings.community" />
               </a>
-              <a target="_blank" href="https://forms.gle/wYyAowiQyK9EDdcDA" rel="noopener noreferrer">
+              <a target="_blank" href="https://forms.gle/1g4H5Ayd1mGWqWpV7" rel="noopener noreferrer">
                 <FormattedMessage id="settings.send-feedback" />
               </a>
               <a className={styles.bottomLink} href="/?report" target="_blank" rel="noreferrer noopener">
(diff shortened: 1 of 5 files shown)


gfodor:
Gated on legal approval.

src/react-components/settings-menu.js
```diff
@@ -151,8 +151,8 @@ export default class SettingsMenu extends Component {
               <a href="https://discord.gg/wHmY4nd" target="_blank" rel="noreferrer noopener">
                 <FormattedMessage id="settings.community" />
               </a>
-              <a href="/spoke" target="_blank" rel="noreferrer noopener">
-                <FormattedMessage id="settings.create-scene" />
+              <a target="_blank" href="https://forms.gle/wYyAowiQyK9EDdcDA" rel="noopener noreferrer">
```
johnshaughnessy: I would make the response field in this form a text box so that people can type more than 1 sentence:
image

src/react-components/ui-root.js
```diff
@@ -1889,17 +1889,11 @@ class UIRoot extends Component {
                   onEndShareVideo={this.endShareVideo}
                   onShareVideoNotCapable={() => this.showWebRTCScreenshareUnsupportedDialog()}
                 />
-                {this.props.isSupportAvailable && (
```
johnshaughnessy: Is this feature deprecated and if so, has the rest of the code supporting it been removed?

gfodor: still TBD, I'll remove it in a week or so after this goes live if we see this new mechanism actually working -- good chance we will revisit if not

src/react-components/settings-menu.js
gfodor: cc @misslivirose

misslivirose: The UI isn't great at displaying it, but it actually is a long-form answer for paragraph responses

src/react-components/ui-root.js
netpro2k: fwiw we did actually remove the polling, so its already disabled

src/react-components/settings-menu.js
johnshaughnessy: Ohh I see -- I didn't even try. My bad!

PR #1174, opened 2019-04-03
Add SoundEffectsSystem

johnshaughnessy:
This introduces the `SoundEffectsSystem` to reduce our use of the `sound` component. Each `sound` component creates a `THREE.Group` and `n=poolSize` `THREE.Audio` nodes, so it's nice to not make those anymore. (We set them to be invisible so it's not much of a performance problem, but it's still nice to get rid of them.) The other reason we like to get rid of them is to reduce our use of DOM events. The `sound` components rely on events that trigger them to play, emits `sound-loaded` events, and (most importantly) emits `sound-ended` events whenever they finish playing.

Two new sound effects were sourced from: 
https://freesound.org/people/tim.kahn/sounds/233800/
https://freesound.org/people/SieuAmThanh/sounds/397674/


src/systems/sound-effects-system.js
```diff
@@ -0,0 +1,75 @@
+/* global fetch */
+import TICK from "../assets/sfx/tick.mp3";
+
+function getBuffer(url, context) {
```
netpro2k: This can be simplified as
```
function getBuffer(url, context) {
    return fetch(url)
        .then(response => response.arrayBuffer())
        .then(arrayBuffer => context.decodeAudioData(arrayBuffer))
}
```

src/systems/sound-effects-system.js
```diff
@@ -0,0 +1,245 @@
...
+    this.tickTeleportSounds(this.teleporters.leftHand, this.teleporterState.leftHand);
+    this.tickTeleportSounds(this.teleporters.rightHand, this.teleporterState.rightHand);
+    this.tickTeleportSounds(this.teleporters.rightRemote, this.teleporterState.rightRemote);
+
+    const userinput = AFRAME.scenes[0].systems.userinput;
+    if (userinput.get(paths.actions.muteMic)) {
+      playSound(this.sounds.tick, this.ctx);
+    }
+
+    for (let i = 0; i < this.pendingEffects.length; i++) {
```
netpro2k: Would be nice for SoundSystem to only do this part and have another system (or maybe even the interaction system) just queue up sounds like everything else. The teleport state feels particularly out of place since its sort of trying to piece together what the teleporter component already knows...

src/systems/sound-effects-system.js
```diff
@@ -0,0 +1,245 @@
...
+    scene.addEventListener("media-loaded", this.onMediaLoaded);
+    scene.addEventListener("media-loading", this.onMediaLoading);
+  }
+
+  cleanUp(scene) {
+    scene.removeEventListener("media-loaded", this.onMediaLoaded);
+    scene.removeEventListener("media-loading", this.onMediaLoading);
+  }
+
+  shouldTick() {
```
netpro2k: I would ditch this and have a promise that resolve once all sounds are ready set `soundsAreReady`

src/systems/sound-effects-system.js
```diff
@@ -0,0 +1,245 @@
...
+      },
+      rightRemote: {
+        teleporting: false,
+        teleportArcSource: null
+      }
+    };
+    this.ctx = THREE.AudioContext.getContext();
+    this.pendingEffects = [];
+    this.soundFor = new Map();
+    this.sounds = {};
```
netpro2k: unclear why this is needed when you have `soundFor`

src/systems/sound-effects-system.js
```diff
@@ -0,0 +1,245 @@
...
+    });
+    getBuffer(MEDIA_LOADING, this.ctx).then(buffer => {
+      this.sounds.mediaLoading = buffer;
+    });
+    getBuffer(MEDIA_LOADED, this.ctx).then(buffer => {
+      this.sounds.mediaLoaded = buffer;
+      this.soundFor.set("media_loaded", buffer);
+    });
+
+    this.mediaLoadingSources = [];
```
netpro2k: Hmm surprised these ones are not spatialized sounds. In any case would be nice to figure out a way to handle these sort of looping sounds in the `pendingEffects` queue, but may not be worth bothering with until we have some more examples.

src/systems/sound-effects-system.js
```diff
@@ -0,0 +1,245 @@
...
+      rightRemote: {
+        teleporting: false,
+        teleportArcSource: null
+      }
+    };
+    this.ctx = THREE.AudioContext.getContext();
+    this.pendingEffects = [];
+    this.soundFor = new Map();
+    this.sounds = {};
+    getBuffer(TICK, this.ctx).then(buffer => {
```
netpro2k: All these getBuffer/then blocks are a lot of boilerplate code. Seems like after getting rid of `this.sounds` it should be possible to rework this into a `Promise.all` and a loop. This would also help get rid of the kind of ugly `shouldTick`.

src/systems/userinput/resolve-action-sets.js
```diff
@@ -38,7 +38,10 @@ export function resolveActionSets() {
   userinput.toggleSet(sets.leftHandHoldingCamera, leftHand.held && leftHand.held.components["camera-tool"]);
   userinput.toggleSet(sets.leftHandHoldingInteractable, leftHand.held);
 
-  userinput.toggleSet(sets.rightHandHoveringOnNothing, !rightRemote.held && !rightRemote.hovered && !rightHand.held && !rightHand.hovered);
+  userinput.toggleSet(
```
netpro2k: Did you actually mean to make this change in this PR, thought this would already be accounted for by gregs menu fix PR

src/hub.html
```diff
@@ -702,46 +653,6 @@
                 set-yxz-order
                 matrix-auto-update
                 set-active-camera
-                sound__teleport_start="positional: false; src: #sound_asset-teleport_start; on: nothing; poolSize: 2; loop: true;"
```
netpro2k: 🎉

src/hub.html
```diff
@@ -312,17 +282,6 @@
...
                     scale="0.5 0.5 0.5"
-                    sound__next_pen_color="src: #sound_asset-next_pen_color; on: next_pen_color; poolSize: 2; positional: false;"
-                    sound__prev_pen_color="src: #sound_asset-prev_pen_color; on: prev_pen_color; poolSize: 2; positional: false;"
-                    sound__start_draw="src: #sound_asset-start_draw; on: start_draw; poolSize: 2;"
-                    sound__stop_draw="src: #sound_asset-stop_draw; on: stop_draw; poolSize: 2; positional: false;"
-                    sound__increase_pen_radius="src: #sound_asset-increase_pen_radius; on: increase_pen_radius; poolSize: 2; positional: false;"
-                    sound__decrease_pen_radius="src: #sound_asset-decrease_pen_radius; on: decrease_pen_radius; poolSize: 2; positional: false;"
-                    sound__grab="src: #sound_asset-grab; on: grabbed; poolSize: 1; positional: false;"
-                    sound__graboff ="src: #sound_asset-grab_off; on: ungrabbed; poolSize: 1; positional: false;"
-                    set-sounds-invisible
```
netpro2k: We should be able to delete the `set-sound-invisible` component now as well, since there should no longer be any non-positional sound components.

src/systems/userinput/resolve-action-sets.js
johnshaughnessy: This was to fix a linting issue but Greg just fixed this on master.

src/hub.html
johnshaughnessy: Do we want it on the template for loaded media?

johnshaughnessy: removed

PR #1180, opened 2019-04-04
Moderately improve cursor target acquisition

mqp:
Populating the targets here seems like it can take a substantial fraction of a second. This cuts the number of times it has to happen on object spawn down to 2 from a handful, and removes the need to call it at all in cases where we're using `setAttribute` but not modifying the class.

In the future we should do something totally different with more mechanical sympathy, like making a single class, A-Frame component, or key on the tags component be responsible for determining whether something is a valid target or not, and making the raycasting code take as input a list of A-Frame entities instead of a list of `Object3D`s.

### DIFF
--- a/src/systems/cursor-targetting-system.js
+++ b/src/systems/cursor-targetting-system.js
@@ -3,23 +3,19 @@ import { waitForDOMContentLoaded } from "../utils/async-utils";
 export class CursorTargettingSystem {
   constructor() {
     this.targets = [];
-    this.setDirty = this.setDirty.bind(this);
     this.dirty = true;
-
+    this.onMutation = this.onMutation.bind(this);
+    this.onObject3DSet = this.onObject3DSet.bind(this);
+    this.onObject3DRemove = this.onObject3DRemove.bind(this);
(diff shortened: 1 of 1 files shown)


src/systems/cursor-targetting-system.js
```diff
@@ -25,6 +27,21 @@ export class CursorTargettingSystem {
...
+    this.dirty = true;
+  }
+
+  onObject3DRemove() {
+    this.dirty = true;
+  }
+
+  onMutation(records) {
+    // let's try to avoid re-querying the targets on attribute changes we know we can ignore
+    if (records.some(r => r.type === "childList" || (r.type === "attributes" && r.attributeName === "class"))) {
```
johnshaughnessy: MDN does not know whether desktop chrome, firefox, or safari support `MutationRecord.type`, so I suggest putting a breakpoint here on each of those browsers just to make sure `r.type` works as expected.
https://developer.mozilla.org/en-US/docs/Web/API/MutationRecord

johnshaughnessy: ^ This comment is out of date. This looks to be supported across all browsers.

PR #1185, opened 2019-04-04
Always show in world HUD until first hover on children

gfodor:
This PR improves the discoverability of the in-VR HUD by ensuring it's always visible until the user first hovers over any of the buttons, at which point it begins doing the hide-on-not-looking behavior for all subsequent sessions.

### DIFF
--- a/src/components/hud-controller.js
+++ b/src/components/hud-controller.js
@@ -20,6 +20,7 @@ AFRAME.registerComponent("hud-controller", {
   },
   init() {
     this.onChildHovered = this.onChildHovered.bind(this);
+    this.removeHoverEvents = this.removeHoverEvents.bind(this);
     this.isYLocked = false;
     this.lockedHeadPositionY = 0;
     this.lookDir = new THREE.Vector3();
@@ -88,6 +89,10 @@ AFRAME.registerComponent("hud-controller", {
   },
 
   pause() {
+    this.removeHoverEvents();
+  },
+
+  removeHoverEvents() {
     for (let i = 0; i < this.hoverableChildren.length; i++) {
       this.hoverableChildren[i].object3D.removeEventListener("hovered", this.onChildHovered);
     }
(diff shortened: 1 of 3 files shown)


src/components/hud-controller.js
```diff
@@ -19,10 +19,13 @@ AFRAME.registerComponent("hud-controller", {
     showTip: { type: "bool" }
   },
   init() {
+    this.onChildHovered = this.onChildHovered.bind(this);
     this.isYLocked = false;
     this.lockedHeadPositionY = 0;
     this.lookDir = new THREE.Vector3();
     this.lookEuler = new THREE.Euler();
+    this.store = window.APP.store;
```
johnshaughnessy: No need to store a reference here. I'd just use `window.APP.store` everywhere instead.

src/components/hud-controller.js
```diff
@@ -69,9 +75,27 @@ AFRAME.registerComponent("hud-controller", {
...
+  },
+
+  pause() {
+    for (let i = 0; i < this.hoverableChildren.length; i++) {
+      this.hoverableChildren[i].object3D.removeEventListener("hovered", this.onChildHovered);
+    }
+  },
+
+  onChildHovered() {
+    if (!this.store.state.activity.hasHoveredInWorldHud) {
```
johnshaughnessy: Can remove event listeners here

src/components/hud-controller.js
```diff
@@ -34,11 +37,14 @@ AFRAME.registerComponent("hud-controller", {
     const pitch = head.rotation.x * THREE.Math.RAD2DEG;
     const yawDif = deltaAngle(head.rotation.y, hud.rotation.y) * THREE.Math.RAD2DEG;
 
+    // HUD is always visible until first hover, to increase discoverability.
+    const forceHudVisible = !this.store.state.activity.hasHoveredInWorldHud;
```
johnshaughnessy: Don't need to make a name for this value -- just use in the `if`'s predicate directly

src/components/hud-controller.js
gfodor: maybe subjective, but i'd prefer to not be dereferencing the window object on every tick since that's kind of a deeper dependency on globals

src/components/hud-controller.js
gfodor: this is "variable name as documentation" :)

PR #1189, opened 2019-04-05
Make it so that you can produce chats with images

mqp:
This is designed so that we can bridge chats with images from Discord. It works but it's kind of WIP because I want to be able to put both text and images in the same chat message.


mqp:
OK, I decided I'm happy with this, and that I want to implement Discord messages with both chat and images by creating separate Hubs messages for the chat and for each image. That is simple to implement, it's not ugly, and it cleanly resolves some questions like "what should the spawn button do".

Issue #1183, opened 2019-04-04
loading screen ends early for heavy scenes


gfodor:
the loading screen often ends before the scene has been fully initialized, resulting in a black screen that sits around for a while. easy repro:

create a room
go to choose scene
choose custom scene
put in this URL: https://s3-us-west-1.amazonaws.com/asset-bundles.reticulum-dev-7f8d39c45878ee2e/files/Traffic4.glb
refresh, room will be black while scene downloads

PR #1197, opened 2019-04-05
Require environment to finish loading before loader is complete

johnshaughnessy:
Fixes https://github.com/mozilla/hubs/issues/1183

### DIFF
--- a/src/react-components/loader.js
+++ b/src/react-components/loader.js
@@ -24,6 +24,14 @@ class Loader extends Component {
     this.props.scene.addEventListener("model-loaded", this.onObjectLoaded);
     this.props.scene.addEventListener("image-loaded", this.onObjectLoaded);
     this.props.scene.addEventListener("model-error", this.onObjectLoaded);
+    this.props.scene.addEventListener(
+      "environment-scene-loaded",
+      () => {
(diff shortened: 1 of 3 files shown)


src/react-components/loader.js
```diff
@@ -51,12 +52,14 @@ class Loader extends Component {
 
     if (!this.doneWithInitialLoad && this.loadingTimeout) window.clearTimeout(this.loadingTimeout);
 
-    this.loadingTimeout = window.setTimeout(() => {
-      this.doneWithInitialLoad = true;
-      if (this.props.onLoaded) {
-        this.props.onLoaded();
-      }
-    }, 1500);
+    if (this.props.environmentSceneLoaded) {
```
gfodor: does `onObjectLoaded` happen when the environment loads? if not, this change would imply this `if` could never be entered in the case where the environment loads last (but i'm assuming this was the main breaking case)

johnshaughnessy: Oh, I hadn't consider that. 
I updated the PR such that the loading screen will definitely finish no matter the order of the two relevant event handlers (`onObjectLoaded` in the `Loader` and `onFirstEnvironmentLoad` in hub.js)

PR #1200, opened 2019-04-06
Check for scene in interaction system

gfodor:
leaving the room throws an error because `AFRAME.scenes[0]` becomes undefined on exit but the interaction system doesn't know to check if the scene is missing.

### DIFF
--- a/src/systems/button-systems.js
+++ b/src/systems/button-systems.js
@@ -1,8 +1,11 @@
 export class SingleActionButtonSystem {
   tick() {
     this.didInteractThisFrame = false;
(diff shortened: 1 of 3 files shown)


johnshaughnessy:
How do I reproduce the error? I don't see it when I press leave room. Also, confused why these changes would fix it. Wouldn't all of these systems be prevented from running by an error on the first line of `hubs-systems`? https://github.com/mozilla/hubs/blob/master/src/systems/hubs-systems.js#L25

gfodor:
i can no longer repro this, will close for now

PR #1201, opened 2019-04-06
Fix issues with invisible buttons in HUD

gfodor:
This PR manages to fix the invisible buttons in the HUD by making it so the `.visible` flag is set to false on the same object(s) the raycaster hits (the ones tagged with the `ui` class.)

However we should understand what is going on before merging. My only working theory right now is that the raycaster checks the visible flag on the hit object and its parent, which would explain why they icon buttons don't have the bug (they are direct children of the invisible root HUD object) but the other buttons do, which have additional layers of nesting between them and the invisible root HUD object.

### DIFF
--- a/src/hub.html
+++ b/src/hub.html
@@ -602,7 +602,6 @@
                 id="player-hud"
                 hud-controller="head: #player-camera;"
                 vr-mode-toggle-class="class: ui"
-                vr-mode-toggle-class__ui="class: interactable-ui"
                 vr-mode-toggle-visibility
                 vr-mode-toggle-playing__hud-controller
             >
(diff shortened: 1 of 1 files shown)


johnshaughnessy:
Now that I'm staring at it, the problem makes sense. The entity marked with `ui` is added to the `CursorTargettingSystem`'s list of eligible raycast targets. When the cursor looks for an `intersection`, it goes through all the raycast targets and calls `raycast` on them. It's _true_ that calling `raycast` on an object will check if it's `visible` and will skip the children, but in this case, it's the _parent_ that was set to be `invisible`, which means that the parent never had the opportunity (during the raycasting process) to prevent its children from being raycasted on. Collapsing these entities (or just ensuring that the `ui` class is on the entity whose `visibility` you need to affect the raycasting code) makes sense and is the correct solution.

src/hub.html
```diff
@@ -605,20 +605,18 @@
                 vr-mode-toggle-visibility
                 vr-mode-toggle-playing__hud-controller
             >
-                <a-entity visibility-while-frozen="requireHoverOnNonMobile: false;" rotation="30 0 0">
-                    <a-entity class="ui interactable-ui">
-                        <a-entity class="leave-btn" position="0 -0.2 0" scale="0.7 0.7 0.7" leave-room-button is-remote-hover-target tags="singleActionButton: true;">
-                            <a-entity mixin="rounded-text-action-button" slice9="width: 0.8" scale="0.8 0.8 1.0">
-                                <a-entity text=" value:leave room; width:2; align:center;" text-raycast-hack position="0 0 0.01"></a-entity>
-                            </a-entity>
+                <a-entity class="ui interactable-ui" visibility-while-frozen="requireHoverOnNonMobile: false;" rotation="30 0 0">
```
johnshaughnessy: I think you liked the separation of these layers for architectural reasons, but they seem undesirable from a perf perspective. The outer three layers seem easy to combine into one. Combining those three with the inner-most entity will involve more work that doesn't seem important to do now.

src/hub.html
```diff
@@ -605,20 +605,18 @@
...
                         </a-entity>
                     </a-entity>
                 </a-entity>
-                <a-entity in-world-hud visibility-while-frozen="visible: false; requireHoverOnNonMobile: false;" rotation="30 0 0">
+                <a-entity in-world-hud visibility-while-frozen="visible: false; requireHoverOnNonMobile: false;" rotation="30 0 0"  class="ui interactable-ui">
                     <a-rounded height="0.08" width="0.2" color="#000000" position="-0.44 -0.045 0" radius="0.040" opacity="0.35" class="hud bg"></a-rounded>
                     <a-image scale="0.07 0.06 0.06" position="-0.38 -0.005 0.001" src="#presence-count" material="alphaTest:0.1;"></a-image>
                     <a-entity id="hud-presence-count" text="value:; width:1.1; align:center;" position="-0.30 -0.005 0" matrix-auto-update="target: text;"></a-entity>
-                    <a-entity class="ui interactable-ui">
+                    <a-entity>
```
johnshaughnessy: Remove this layer. Consider combining the next two layers.

gfodor:
How is it that the icon buttons didn't exhibit this problem?

johnshaughnessy:
All the `icon-buttons` that I see in `hub.html` are nested in an `entity` that is on or above a level that is toggling visibility.

src/hub.html
```diff
@@ -602,23 +602,22 @@
                 id="player-hud"
                 hud-controller="head: #player-camera;"
                 vr-mode-toggle-class="class: ui"
+                vr-mode-toggle-class__ui="class: interactable-ui"
```
johnshaughnessy: The only place I see `.interactable-ui` being used is in `media-pager`. Is that relevant here?

PR #1202, opened 2019-04-06
Replace jsonschema with a tiny piece of validation logic

mqp:
This was a pretty heavy dependency (a few dozen minified kilobytes) so it's unreasonable to take it and then barely use any of its features.

(It was actually even bigger than it looked, because it also secretly depended on Webpack polyfilling the Node.js `url` standard library module.)

### DIFF
--- a/test/unit/utils/validate-storage-schema.test.js
+++ b/test/unit/utils/validate-storage-schema.test.js
@@ -1,0 +1,64 @@
+import test from "ava";
+import validate from "../../../src/storage/validate.js";
+
+function isString(value) {
+  return typeof value === "string" || value instanceof String;
+}
+
+function isBoolean(value) {
+  return typeof value === "boolean" || value instanceof Boolean;
+}
+
+test("validate local storage schema", t => {
+
+  t.notThrows(() => validate("foo", isString));
+  t.throws(() => validate({}, isString));
+  t.throws(() => validate("foo", isBoolean));
+
+  const exampleSchema = {
+    profile: {
+      displayName: isString,
+      avatarId: isString,
+    },
+    settings: {
+      isBot: isBoolean
+    },
+    confirmedDiscordRooms: [isString],
+    uploadPromotionTokens: [{ fileId: isString, promotionToken: isString }]
+  };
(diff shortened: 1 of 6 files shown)


gfodor:
This seems like the wrong trade-off to me. The downside here is now we are exposed to bugs or regressions in the new validation logic. If there were a bunch of unit tests to ensure we are not going to get burned by this I'd feel better, but it certainly doesn't seem worth a few kilobytes to lose confidence in the schema validation routines.

mqp:
I'm happy to make some unit tests, so if that changes your mind then we agree.

(I didn't know we had a mechanism for tests, but I see that we do now!)

gfodor:
Great thanks!

gfodor:
Also I don't know if we added the test runner to npm build -- can you do that if not?

test/unit/utils/validate-storage-schema.test.js
```diff
@@ -0,0 +1,64 @@
...
+    settings: {},
+    confirmedDiscordRooms: [],
+    uploadPromotionTokens: []
+  }, exampleSchema));
+
+  t.throws(() => validate({
+    profile: { displayName: "John" },
+    settings: { isBot: true },
+    confirmedDiscordRooms: ["general"],
+    uploadPromotionTokens: [{ fileId: "whiz", promotionToken: 94329 }]
```
johnshaughnessy: bang

mqp:
> Also I don't know if we added the test runner to npm build -- can you do that if not?

Well, right now, it's rather the opposite -- `npm test` runs the build as a "kind of test". Do you know a whole other way you endorse?

If I were designing it, I would have `test` and `build` be totally separate and neither one does the other.

gfodor:
don't have a strong opinion -- don't know npm culture. in java land,
typically the happy path build script runs the tests for example. in any
case, main thing is to ensure that our actual CI build runs the tests and
fails the build if it fails.

netpro2k:
I would keep it as `npm test`, but I agree our CI scripts should run the tests if they don't already.

PR #1207, opened 2019-04-08
More perf fixes for physics

InfiniteLee:
-don't only check for scaling when object is held (check for ownership instead)
-fix issue with duplicate shapes being created for the environment

### DIFF
--- a/src/components/sticky-object.js
+++ b/src/components/sticky-object.js
@@ -48,7 +48,7 @@ AFRAME.registerComponent("sticky-object", {
     this.heldRightRemote = heldRightRemote;
 
     if (!almostEquals(0.001, this.prevScale, this.el.object3D.scale)) {
-      if ((this.heldLeftHand || this.heldRightHand || this.heldRightRemote) && !this.wasScaled) {
+      if ((!this.el.components.networked || NAF.utils.isMine(this.el)) && !this.wasScaled) {
         this.wasScaled = true;
         this.el.setAttribute("ammo-body", { collisionFilterMask: COLLISION_LAYERS.HANDS });
       }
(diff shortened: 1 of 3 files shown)


src/utils/media-utils.js
```diff
@@ -362,6 +362,7 @@ export const traverseMeshesAndAddShapes = (function() {
         o.name !== "Floor_Plan" &&
         o.name !== "Ground_Plane"
       ) {
+        o.updateMatrices();
```
netpro2k: Hmm, whats this about? This should only get called if you are about to read an objects matrix. Guessing that maybe happens in ammo-shape? If so would be good to do this there instead. (we are trying to keep upadteMatricies/matrixWorldNeedsUpdate as close as possible to the code that makes them needed so we can spot places where we are breaking the contract by mistake)

src/components/sticky-object.js
```diff
@@ -48,7 +48,7 @@ AFRAME.registerComponent("sticky-object", {
     this.heldRightRemote = heldRightRemote;
 
     if (!almostEquals(0.001, this.prevScale, this.el.object3D.scale)) {
-      if ((this.heldLeftHand || this.heldRightHand || this.heldRightRemote) && !this.wasScaled) {
+      if ((!this.el.components.networked || NAF.utils.isMine(this.el)) && !this.wasScaled) {
```
johnshaughnessy: Rather than have each `sticky-object` check to see if it's owned and scaled, how about we set the appropriate attribute on the `ammo-body` at the same time that we begin scaling an (owned) object: https://github.com/mozilla/hubs/blob/master/src/systems/two-point-stretching-system.js#L39

johnshaughnessy: (And likewise set the appropriate attribute(s) when using the scaling button in the menu)

Issue #1208, opened 2019-04-09
Deleting an object that is mid-spawning resulted in audio playing on loop


misslivirose:
**Repro Steps:**
1. Start adding a 3D model from Sketchfab (I reproduced this with what was likely a high-poly model) 
2. While the object is in it's 'adding' state, delete the object through the overlay menu

The audio was repeating for me (and I believe that @gfodor also heard it on a loop) until refreshing the page. 

FF Nightly on Win 10

johnshaughnessy:
Fixed in https://github.com/mozilla/hubs/pull/1209

PR #1209, opened 2019-04-09
Stop sound effect when media-loader's object is removed

johnshaughnessy:
Fixes https://github.com/mozilla/hubs/issues/1208

Note however that `gltf-model-plus` will still emit `model-loaded`, and the loaded sound will still play because the event listener is still hooked up. Both `media-loader` and `gltf-model-plus` need to be patched to take this possibility into account. I would not be surprised, for example, if the event listeners in `media-loader` are holding references to nodes that have been removed from the dom and are thus preventing them from getting GC'ed. I have not tested this.

### DIFF
--- a/src/components/media-loader.js
+++ b/src/components/media-loader.js
@@ -82,6 +82,13 @@ AFRAME.registerComponent("media-loader", {
     }
   },
 
+  remove() {
+    if (this.loadingSoundNode) {
+      this.el.sceneEl.systems["hubs-systems"].soundEffectsSystem.stopSoundNode(this.loadingSoundNode);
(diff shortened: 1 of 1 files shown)


src/components/media-loader.js
```diff
@@ -407,10 +414,6 @@ AFRAME.registerComponent("media-pager", {
     if (this.toolbar) {
       this.toolbar.parentNode.removeChild(this.toolbar);
     }
-    if (this.loadingSoundNode) {
```
johnshaughnessy: I stuck this in the wrong component's `remove`! 😳

PR #1211, opened 2019-04-09
Presence-authoritative identity

brianpeiris:
We want phoenix presence to be the authority for identity. 
This PR:
- Marks "naf" messages with a "phx-reliable" source so that networked-aframe can identify them.
- Sets networked-aframe to ignore "first sync" messages from anything other than "phx-reliable" sources.
- Changes our phoenix connection process such that we receive a server-generated session id after joining a hub channel, instead of generating the session id ourselves. This prevents others from spoofing our session id in "naf" phoenix messages
- Moves displayName syncing out of networked-aframe.
- Changes the profile editing flow such that a change to the local store only triggers a presence update.
- Changes the player-info component to respond to display name changes in presence, correlating session ids in presence with NAF clientIds (which are now authoritative).

More info here: https://discourse.mozilla.org/t/discord-authoratative-identity/37511

Goes with:
https://github.com/mozilla/naf-janus-adapter/pull/74
https://github.com/MozillaReality/networked-aframe/pull/25
https://github.com/mozilla/reticulum/pull/152


src/hub.js
```diff
@@ -819,83 +820,99 @@ document.addEventListener("DOMContentLoaded", async () => {
...
+
       const hubPhxPresence = hubChannel.presence;
-      let isInitialSync = true;
       const vrHudPresenceCount = document.querySelector("#hud-presence-count");
 
-      hubPhxPresence.onSync(() => {
-        remountUI({ presences: hubPhxPresence.state });
-        const occupantCount = Object.entries(hubPhxPresence.state).length;
-        vrHudPresenceCount.setAttribute("text", "value", occupantCount.toString());
+      if (isInitialJoin) {
```
brianpeiris: Mostly just wrapped this block in an if statement. Take a look at the diff with whitespace ignored.

src/components/player-info.js
```diff
@@ -7,29 +7,58 @@ import { AVATAR_TYPES } from "../assets/avatars/avatars";
...
     avatarSrc: { type: "string" },
     avatarType: { type: "string", default: AVATAR_TYPES.LEGACY }
   },
   init() {
+    this.displayName = null;
     this.applyProperties = this.applyProperties.bind(this);
+    this.updateDisplayName = this.updateDisplayName.bind(this);
+    this.applyDisplayName = this.applyDisplayName.bind(this);
+
+    this.isLocalPlayerInfo = this.el.id === "player-rig";
```
gfodor: perhaps this should be a schema property instead of looking at the id?

src/components/player-info.js
```diff
@@ -7,29 +7,58 @@ import { AVATAR_TYPES } from "../assets/avatars/avatars";
...
+    this.playerSessionId = null;
+
+    if (!this.isLocalPlayerInfo) {
+      NAF.utils.getNetworkedEntity(this.el).then(networkedEntity => {
+        this.playerSessionId = NAF.utils.getCreator(networkedEntity);
+        const playerPresence = window.APP.hubChannel.presence.state[this.playerSessionId];
+        if (playerPresence) {
+          this.displayName = playerPresence.metas[0].profile.displayName;
```
gfodor: maybe this and the two-liner on 51 should be turned into a function that receives the data structure at `playerPresence.metas[0]`

src/hub.js
```diff
@@ -819,83 +820,99 @@ document.addEventListener("DOMContentLoaded", async () => {
...
+          });
         });
-      });
+      }
+      isInitialJoin = false;
 
       const permsToken = oauthFlowPermsToken || data.perms_token;
       hubChannel.setPermissionsFromToken(permsToken);
+
+      hubChannel.sendProfileUpdate();
```
gfodor: what's the reason for this? isn't the initial presence info going to go out on join automatically?

brianpeiris: Oh. Hmm, I think I did this because it seems we only update presence in reticulum after entry, not after join. I'll change reticulum instead.

brianpeiris: Weird. I can't repro the issue. It seems we already update presence on join, so maybe I was mistaken.

src/components/player-info.js
brianpeiris: Oh, actually, I can't do that, since player-info is still networked, so we can't allow others to override that prop.

PR #1213, opened 2019-04-10
Video snapshotting  

gfodor:
Click the camera button on a video or screen share to pop out a still snapshot.

image


src/components/media-views.js
```diff
@@ -646,6 +670,17 @@ AFRAME.registerComponent("media-video", {
...
+    const heldRightHand = interaction.state.rightHand.held === this.el;
+    const heldRightRemote = interaction.state.rightRemote.held === this.el;
+    if (
+      (heldLeftHand && userinput.get(interaction.options.leftHand.grabPath)) ||
+      (heldRightHand && userinput.get(interaction.options.rightHand.grabPath)) ||
+      (heldRightRemote && userinput.get(interaction.options.rightRemote.grabPath))
+    ) {
+      this.localSnapCount = 0;
+    }
+
```
johnshaughnessy: I recognize this blob because I wrote a similar one. I started writing this kind of check in a nicer way, using a helper caled `isHeld` on the interaction system. What we care about here is whether the object went from not held to held, so now I'd write this as:

```
    const isHeld = interaction.isHeld(this.el);
    if (isHeld && !this.wasHeld) {
      this.localSnapCount = 0;
    }
    this.wasHeld = isHeld;
```

(You will have to pull in the code that has that helper into your branch, which I'm deploying this afternoon in #1203 . See https://github.com/mozilla/hubs/pull/1203/files#diff-519394e566e5f85b23992db2e1b8f609R26 )

src/components/media-views.js
```diff
@@ -421,6 +427,23 @@ AFRAME.registerComponent("media-video", {
...
+    const canvas = document.createElement("canvas");
+    canvas.width = this.video.videoWidth;
+    canvas.height = this.video.videoHeight;
+    canvas.getContext("2d").drawImage(this.video, 0, 0, canvas.width, canvas.height);
+    const blob = await new Promise(resolve => canvas.toBlob(resolve));
+    const file = new File([blob], "snap.png", { type: "image/png" });
+
+    this.localSnapCount++;
+    const { entity } = spawnMediaAround(this.el, file, this.localSnapCount);
+    entity.addEventListener("image-loaded", () => (this.isSnapping = false), { once: true });
```
johnshaughnessy: Write `this.onImageLoaded = () => { this.isSnapping = false; }; ` in `init`, then replace this line with
`entity.addEventListener("image-loaded", this.onImageLoaded, { once: true });` 
to avoid creating garbage.
If you want to, you can even write
`const ONCE_TRUE = { once: true }` at the top of the file and then use that when adding an event listener to avoid creating an object. 

Not a big deal because this snapping happens rarely, but I've started to avoid allocations whenever possible unless it adds significant complexity.

src/components/media-views.js
```diff
@@ -421,6 +427,23 @@ AFRAME.registerComponent("media-video", {
...
+    if (this.isSnapping) return;
+    this.isSnapping = true;
+    this.el.sceneEl.systems["hubs-systems"].soundEffectsSystem.playSoundOneShot(SOUND_CAMERA_TOOL_TOOK_SNAPSHOT);
+
+    const canvas = document.createElement("canvas");
+    canvas.width = this.video.videoWidth;
+    canvas.height = this.video.videoHeight;
+    canvas.getContext("2d").drawImage(this.video, 0, 0, canvas.width, canvas.height);
+    const blob = await new Promise(resolve => canvas.toBlob(resolve));
+    const file = new File([blob], "snap.png", { type: "image/png" });
```
johnshaughnessy: Could reuse this `{ type: "image/png" }` object by declaring it at the top of the file if you feel so inclined.

src/hub.html
```diff
@@ -416,22 +418,30 @@
             <!-- TODO: All of these positions are assuming 16:9 - need to actually reposition based on video resolution -->
             <template id="video-hover-menu">
                 <a-entity class="ui interactable-ui hover-container" visible="false">
+                    <a-image
+                        is-remote-hover-target
+                        tags="singleActionButton: true; isHoverMenuChild: true;"
+                        icon-button="image: #video-snap; hoverImage: #video-snap-hover; activeImage: #videosnapplay; activeHoverImage: #video-snap-hover"
+                        scale="0.075 0.075 0.075"
+                        position="0 0.225 0"
+                        class="ui video-snap-button"
```
johnshaughnessy: I think you can lose the `ui` class here.

src/utils/media-utils.js
```diff
@@ -405,6 +405,61 @@ export const traverseMeshesAndAddShapes = (function() {
   };
 })();
 
+const mediaPos = new THREE.Vector3();
+
+export function spawnMediaAround(el, media, snapCount, mirrorOrientation = false) {
+  const { entity, orientation } = addMedia(media, "#interactable-media", undefined, false);
+
```
johnshaughnessy: Do you need
```
el.object3D.updateMatrices();
el.object3D.matrix.decompose(el.object3D.position, el.object3D.rotation, el.object3D.scale)
```
here?

src/utils/media-utils.js
```diff
@@ -405,6 +405,61 @@ export const traverseMeshesAndAddShapes = (function() {
...
+    },
+    { once: true }
+  );
+
+  entity.object3D.matrixNeedsUpdate = true;
+
+  entity.addEventListener(
+    "media_resolved",
+    () => {
+      el.emit("photo_taken", entity.components["media-loader"].data.src);
```
johnshaughnessy: What listens to this event?

johnshaughnessy: Nevermind - I found it
```
src/scene-entry-manager.js:495: 
    this.scene.addEventListener("photo_taken", e => {
      this.hubChannel.sendMessage({ src: e.detail }, "spawn");
    });
```

src/utils/media-utils.js
johnshaughnessy: Nevermind about this. I got confused.

PR #1214, opened 2019-04-10
Room owners can now mute other users

gfodor:
Room owners can now apply a soft mute to other users. Goes with https://github.com/mozilla/reticulum/pull/153

image

Note that this isn't a tool for dealing with abuse -- kicking is that facility. This is soft mute for non-bad actors who have open mics, etc.

### DIFF
--- a/src/components/mute-button.js
+++ b/src/components/mute-button.js
@@ -1,0 +1,22 @@
+AFRAME.registerComponent("mute-button", {
+  init() {
+    this.onClick = () => {
+      this.mute(this.owner);
+    };
+    NAF.utils.getNetworkedEntity(this.el).then(networkedEl => {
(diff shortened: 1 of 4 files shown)


gfodor:
UX question: perhaps we should hide the "hide" button if you are a moderator, to reduce confusion between local and global actions. (The volume control still has this problem, but less so.) Not to mention that if a moderator hides someone and then that person starts causing trouble, they'll have to reload to kick.

brianpeiris:
I think we'd want to resolve any confusion with some other mechanism. Maybe global actions should always have a confirmation dialog, or at least the first time you use them. I would also rather add an "unhide" button to the presence list UI.

gfodor:
ah yeah I think that can be a follow up. at a glance that seems somewhat complicated given the threading of state that would be necessary

PR #1216, opened 2019-04-10
Fix scaling of media-images with shared textures

netpro2k:
When creating a `media-image` that ends up sharing a texture (since it was in the texture cache) the image does not get the correct aspect ratio, since images are now deleted after uploading to the GPU. Fix this by also storing width and height in the texture cache.

### DIFF
--- a/src/components/media-views.js
+++ b/src/components/media-views.js
@@ -199,10 +199,7 @@ function createVideoTexture(url, contentType) {
   });
 }
 
-function fitToTexture(el, texture) {
-  const ratio =
-    (texture.image.videoHeight || texture.image.height || 1.0) /
-    (texture.image.videoWidth || texture.image.width || 1.0);
+function scaleToAspectRatio(el, texture, ratio) {
   const width = Math.min(1.0, 1.0 / ratio);
   const height = Math.min(1.0, ratio);
   el.object3DMap.mesh.scale.set(width, height, 1);
@@ -247,11 +244,14 @@ class TextureCache {
   cache = new Map();
 
   set(src, texture) {
+    const image = texture.image;
     this.cache.set(src, {
       texture,
-      count: 0
+      count: 0,
+      width: image.width || image.imageWidth || image.videoWidth,
+      height: image.height || image.imageHeight || image.videoHeight
     });
-    this.retain(src);
+    return this.retain(src);
   }
 
   has(src) {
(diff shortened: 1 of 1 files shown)


src/components/media-views.js
```diff
@@ -199,10 +199,7 @@ function createVideoTexture(url, contentType) {
   });
 }
 
-function fitToTexture(el, texture) {
-  const ratio =
-    (texture.image.videoHeight || texture.image.height || 1.0) /
-    (texture.image.videoWidth || texture.image.width || 1.0);
+function scaleToAspectRatio(el, texture, ratio) {
```
johnshaughnessy: `texture` is unused. Can pass in height and width here like
`function scaleToAspectRatio(el, height, width)`
or remove this function and set the `scale` and set `matrixNeedsUpdate` at the call sites.

src/components/media-views.js
```diff
@@ -706,6 +706,8 @@ AFRAME.registerComponent("media-image", {
 
   async update(oldData) {
     let texture;
+    let width = 1;
```
johnshaughnessy: Is it a bug if this default value is used? If so, would prefer that they are not given any default width/height so that we can notice that bug if it happens.

src/components/media-views.js
```diff
@@ -733,7 +738,10 @@ AFRAME.registerComponent("media-image", {
           throw new Error(`Unknown image content type: ${contentType}`);
         }
 
-        textureCache.set(src, texture);
+        const cacheItem = textureCache.set(src, texture);
```
johnshaughnessy: `GIFTexture` extends `THREE.Texture` and writes:
```
    this.image.width = frames[0].width;
    this.image.height = frames[0].height;
```
Are GIFs being scaled correctly?

netpro2k: It doesn't use TextureLoader to load gifs, so the texture cleanup after upload change doesn't apply to them (it couldn't anyway since the images are actually needed on the CPU side). The width/height here are width/height of the frame data provided by the gif worker.

src/components/media-views.js
netpro2k: Ah I thought the error image was square, but actually its not, so it should default to the size of that.

src/components/media-views.js
netpro2k: Ah yeah woops. In fact we actually only care about ratio, so just going to store that instead

PR #1218, opened 2019-04-11
Fix spawn animations; improve collision shapes for flat media; update three-to-ammo

InfiniteLee:
-fix spawn animations so that they scale the mesh instead of the object3D
-use default fit for "flat" media collision shapes
-use an updated version of three-to-ammo that calls updateMatrices appropriately


src/components/media-loader.js
```diff
@@ -182,9 +182,7 @@ AFRAME.registerComponent("media-loader", {
       } else {
         el.setAttribute("ammo-shape", {
           type: SHAPE.BOX,
-          halfExtents: { x: 0.5, y: 0.5, z: 0.02 },
```
netpro2k: So what is now responsible for setting the right half extents?

InfiniteLee: three-to-ammo does it automatically

PR #1219, opened 2019-04-12
More efficient AABB for object hierarchies

mqp:
We will typically already have the geometry's `boundingBox` cached ([for example, when building a BVH](https://github.com/gkjohnson/three-mesh-bvh/pull/102)), so this is much faster than the previous implementation, which looks at every vertex in the hierarchy fresh every time.

This implementation should get a less tight AABB in some cases (where there are rotations inside the object hierarchy) but it shouldn't be that much worse, and we only rely on this for things that need to be roughly correct (object scale and menu offset), not precisely correct.

### DIFF
--- a/src/utils/auto-box-collider.js
+++ b/src/utils/auto-box-collider.js
@@ -1,3 +1,8 @@
+// Computes an AABB that surrounds the object in question and all of its children.
+// Note that if the children have rotations, this probably won't return the tightest
+// possible AABB -- we could return a tighter one if we examined all of the vertices
+// of the geometry for ourselves, but we don't care enough for what we're using this
+// for to do so much work.
 const computeObjectAABB = (function() {
   const bounds = new THREE.Box3();
   return function(root, target) {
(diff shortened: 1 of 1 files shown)


src/utils/auto-box-collider.js
```diff
@@ -7,8 +24,9 @@ export function getBox(entity, boxRoot) {
 
   entity.object3D.updateMatrices(true, true);
   boxRoot.updateMatrices(true, true);
+  boxRoot.updateMatrixWorld(true);
```
johnshaughnessy: This is surprising. I thought that `boxRoot.updateMatrices(true, true)` will force a world update. Why is `updateMatrixWorld(true)` necessary here?

(The forced updates above are already surprising by the way. I don't know why we need to force local or world updates here.)

src/utils/auto-box-collider.js
```diff
@@ -7,8 +24,9 @@ export function getBox(entity, boxRoot) {
...
 
-  box.setFromObject(boxRoot);
+  computeObjectAABB(boxRoot, box);
```
johnshaughnessy: Would like it if this function name indicated that the box that is returned is not the smallest AABB that contains the object.
I know I don't like `computeObjectAABBButNotNecessarilyTheSmallestOneThatContainsTheObject` but I haven't thought of another one. 

Perhaps a comment instead of a rename would help, since we all guessed wrong about whether this algorithm would return the smallest AABB.

PR #1220, opened 2019-04-12
Don't recreate new workers for every message

mqp:
Pretty self-explanatory. This interface should be suitable for any worker that takes work items and processes them independently into some kind of results.

I'm not sure exactly how inefficient the old version was (does it make a thread for every worker? do the threads stay around forever? does it have to re-run all the worker Javascript, which is a lot in the case of the Sketchfab worker, every time the worker is created?) but now we don't have to think about it.

### DIFF
--- a/src/utils/promisify-worker.js
+++ b/src/utils/promisify-worker.js
@@ -1,0 +1,36 @@
+// Wrapper for a worker which accepts work items and responds with results.
+//
+// The worker must receive messages with data of the shape { id, payload }, and then respond
+// with messages containing matching IDs of the shape { id, result, err }.
+//
+// Returns a function that accepts a work item, sends it to the worker, and returns a promise that
+// will resolve with the result or reject with the error.
+//
+export function promisifyWorker(worker) {
+  let nextItemId = 0;
+  const outstanding = {}; // item ID: { resolve, reject }
+
+  worker.onmessage = msg => {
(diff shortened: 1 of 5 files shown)


src/utils/promisify-worker.js
```diff
@@ -0,0 +1,36 @@
...
+// will resolve with the result or reject with the error.
+//
+export function promisifyWorker(worker) {
+  let nextItemId = 0;
+  const outstanding = {}; // item ID: { resolve, reject }
+
+  worker.onmessage = msg => {
+    const { id, result, err } = msg.data;
+    const handlers = outstanding[id];
+    if (handlers == null) {
```
johnshaughnessy: Do you want `===` or is this `==` on purpose?

src/utils/promisify-worker.js
```diff
@@ -0,0 +1,36 @@
...
+      console.error(`Unknown message with ID ${id} received from ${worker}.`);
+    } else {
+      delete outstanding[id];
+      if (err != null) {
```
johnshaughnessy: Do you want `!==` or is this `!=` on purpose?

mqp:
I conventionally use `x == null` in Javascript. It is equivalent to `x === null || x === undefined`. It seems so rare that I find a good reason to treat `null` and `undefined` as having different semantics that I prefer to consistently treat them the same.

PR #1237, opened 2019-04-17
Add video capture command and system

brianpeiris:
Add an experimental chat command to capture a 15 second video from your point of view, using the MediaRecorder API. 

There are a few downsides to this method:
- Firefox and Chrome will only produce webm videos, so Safari can't play them. 
- This implementation uses the main canvas (not our in-world camera tool) so the resulting video will have the canvas' dimensions. It also means that a video's dimensions will change if you resize your window in the middle of a capture.


src/systems/capture-system.js
```diff
@@ -0,0 +1,64 @@
...
+        this.chunks.push(e.data);
+      };
+      this.recorder.onstop = () => {
+        if (this.chunks.length === 0) return;
+        const mimeType = this.chunks[0].type;
+        const blob = new Blob(this.chunks, { type: mimeType });
+        this.chunks.length = 0;
+        this.scene.emit("add_media", new File([blob], "capture", { type: mimeType }));
+      };
+    });
```
johnshaughnessy: A `"loaded"` event should not reach the scene more than once, but do you want to add, `{once: true}` to `addEventListener` anyway?

src/systems/hubs-systems.js
```diff
@@ -19,6 +20,7 @@ AFRAME.registerSystem("hubs-systems", {
     this.superSpawnerSystem = new SuperSpawnerSystem();
     this.hapticFeedbackSystem = new HapticFeedbackSystem();
     this.soundEffectsSystem = new SoundEffectsSystem();
+    this.captureSystem = new CaptureSystem(this.el);
```
johnshaughnessy: Sure why not. It's as good a spot as any other 😆

src/react-components/chat-command-help.js
```diff
@@ -11,7 +11,17 @@ export default class ChatCommandHelp extends Component {
...
+    const commands = [
+      "leave",
+      "fly",
+      "grow",
+      "shrink",
+      "duck",
+      "debug",
+      "scene <scene url>",
+      "rename <new name>",
+      "capture [stop]"
```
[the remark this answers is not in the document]
johnshaughnessy: If people don't understand `/capture stop`, a separate `/stopCapture` command might work.

src/systems/capture-system.js
```diff
@@ -0,0 +1,64 @@
+export class CaptureSystem {
+  constructor(scene) {
+    this.scene = scene;
+
+    this.gotAudioTrack = false;
+    this.stream = new MediaStream();
+
+    this.scene.addEventListener("loaded", () => {
```
johnshaughnessy: I think it may be possible for the scene to have already been loaded. I know that is surprising but humor me for a moment.

The `a-scene` element's prototype (aka `AScene.prototype`) is a an `AEntity.prototype`, which in turn is an `ANode.prototype`. The `ANode` is responsible for emitting the `loaded` event, which happens once it notices that all of its children have been `loaded`. This typically happens in an `AEntity`'s `attachedCallback`, but doesn't in the case that the `AEntity` `isScene`. In the case of an `AScene`, `load` is called from a `setTimeout` in `play`, which is called from `attachedCallbackPostCamera`, which happens when the `cameraready` event handler fires....

This happens in the `camera` system's `init` which is.... 
Ok it's after the `hubs-systems` `init`, so nevermind, I think you're fine.

If the camera system came before... then... would it break? I thiinknk not... because the `timeout` in `AScene`'s `play` would make all the system `init`s get called before its `load`. But if that timeout wasn't there, then could it be?

OK feel free to ignore this. I was going to suggest you check if the scene `hasLoaded` but I think that isn't required after all.

misslivirose:
Is there any kind of visual indicator when someone is using this feature?

brianpeiris:
Nope, there is no visual indicator.

PR #1260, opened 2019-04-23
Close Room

gfodor:
Room owners can now close rooms.

image

Goes with https://github.com/mozilla/reticulum/pull/160


src/react-components/close-room-dialog.js
```diff
@@ -0,0 +1,35 @@
+import React, { Component } from "react";
+import PropTypes from "prop-types";
+import { FormattedMessage } from "react-intl";
+
+import styles from "../assets/stylesheets/close-room-dialog.scss";
+import DialogContainer from "./dialog-container";
+
+export default class RenameRoomDialog extends Component {
```
brianpeiris: Rename class

src/assets/translations.data.json
```diff
@@ -15,6 +15,9 @@
     "sign-in.out": "Sign Out",
     "rename-room.message": "Enter the new name for the room.",
     "rename-room.rename": "Rename Room",
+    "close-room.message": "Closing this room will remove others from the room and prevent further entry. Are you sure?",
```
brianpeiris:
```suggestion
    "close-room.message": "Closing this room will remove yourself and others from the room and prevent further entry. This cannot be undone. Are you sure?",
```

If we are going to ship this feature without the ability for admins/owners to re-open a room, we should make it clear that they cannot undo the action and that they themselves will lose access to the room. In case they close it accidentally or they change their mind.

gfodor: Yup good point. I will make this warning scarier. I started implementing re-open but it seemed non-trivial since we are blocking room joins once room is closed, so I'd need to either make an exception to the owner(s) (which may be a bad idea from a security standpoint) or use a REST API.

Issue #1238, opened 2019-04-17
"Use scene" should either not show up or tell the user they need to be the room owner


johnshaughnessy:
People think the "use scene" button is broken because they click it and nothing happens. We need the button to either not show up or we need to tell them that they can't change it unless they're the owner. We could do the latter using the same mechanism by which we show the "pinning this object will send it to discord" message when you hover on the pin button.

gfodor:
https://github.com/mozilla/hubs/pull/1417

PR #1261, opened 2019-04-24
Show use scene permission tip

johnshaughnessy:
Fixes https://github.com/mozilla/hubs/issues/1238

We now show the "use scene" button on hover for scene links, whether or not you have permission to change the scene. When you do not have permission, we explain that with a text label that appears.
image

In the freeze menu for rooms and scenes, the "visit room" and "use scene" buttons are replaced with "open link", since "visit room" and "use scene" are available in the hover menu.
image

### DIFF
--- a/src/components/open-media-button.js
+++ b/src/components/open-media-button.js
@@ -3,8 +3,6 @@ import { guessContentType } from "../utils/media-utils";
 
 AFRAME.registerComponent("window-open-button", {
   init() {
-    this.label = this.el.querySelector("[text]");
-
     this.updateSrc = () => {
       this.src = this.targetEl.components["media-loader"].data.src;
       this.el.object3D.visible = !!this.src;
--- a/src/components/open-media-button.js
+++ b/src/components/open-media-button.js
@@ -1,9 +1,40 @@
 import { isHubsSceneUrl, isHubsRoomUrl } from "../utils/media-utils";
 import { guessContentType } from "../utils/media-utils";
 
-AFRAME.registerComponent("open-media-button", {
+AFRAME.registerComponent("window-open-button", {
   init() {
     this.label = this.el.querySelector("[text]");
+
+    this.updateSrc = () => {
+      this.src = this.targetEl.components["media-loader"].data.src;
+      this.el.object3D.visible = !!this.src;
+    };
+    this.onClick = () => {
+      window.open(this.src);
+    };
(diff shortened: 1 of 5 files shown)


src/utils/media-utils.js
```diff
@@ -438,7 +438,7 @@ export function spawnMediaAround(el, media, snapCount, mirrorOrientation = false
   return { entity, orientation };
 }
 
-const hubsSceneRegex = /https?:\/\/(hubs.local(:\d+)?|(smoke-)?hubs.mozilla.com)\/scenes\/(\w+)\/?\S*/;
+const hubsSceneRegex = /https?:\/\/((dev.reticulum.io)|(hubs.local(:\d+)?|(smoke-)?hubs.mozilla.com))\/scenes\/(\w+)\/?\S*/;
```
gfodor: i don't think we want to commit this change -- this would mean links to dev scenes would be recognized as scene links in production. if we want to get this working in dev it should be conditional at some point i think

gfodor:
We have a similar disclaimer for the discord pinning -- perhaps we should unify this under a single tip system? the "view your desktop" tip we give users when they click the media browser was factored somewhat generically (in naming, etc.)

src/components/open-media-button.js
```diff
@@ -1,9 +1,40 @@
 import { isHubsSceneUrl, isHubsRoomUrl } from "../utils/media-utils";
 import { guessContentType } from "../utils/media-utils";
 
+AFRAME.registerComponent("window-open-button", {
```
gfodor: what's the motivation for breaking this out vs just incorporating the necessary logic into `open-media-button`?

johnshaughnessy: We use this to open links to anything (including scenes and rooms) in another tab, rather than to change the scene or navigate to a room. They have two things in common: 
- Both want to get the `src` from the `media-loader`
- Both do something on click
But besides that they serve different purposes.

gfodor:
Note: we'll want to update this PR once https://github.com/mozilla/hubs/pull/1305 lands to hide the tip and then properly ask the user to sign in if they can become the room creator and change the scene if they click.

gfodor:
i believe you no longer need this permission tip aspect to this PR -- the latest code will either not show the button if you don't permission, or will show it and prompt you to login if you need to become owner

Issue #1259, opened 2019-04-22
UX: Provide an affordance to unhide a user when they have been hidden


misslivirose:
The current flow for this requires leaving the room before the user shows up again. There should be some flow inside of Hubs that doesn't force a user to leave and return to unhide someone. I would imagine this being somewhere in the user list (perhaps as an icon when users are hidden that can be clicked to unhide them, sketch below - I don't know that the 'unhidden' state should be shown all the time for all users)

![image](https://user-images.githubusercontent.com/4313320/56512852-6ff31200-64e5-11e9-8343-4fa6ecd35aff.png)

gfodor:
https://github.com/mozilla/hubs/pull/1310

misslivirose:
🎉

PR #1262, opened 2019-04-24
Unhide users from the presence list

johnshaughnessy:
Resolves https://github.com/mozilla/hubs/issues/1259

### DIFF
--- a/src/components/block-button.js
+++ b/src/components/block-button.js
@@ -23,11 +23,12 @@ AFRAME.registerComponent("block-button", {
 
   block(clientId) {
     NAF.connection.adapter.block(clientId);
+    window.APP.blockedIds.add(clientId);
   },
 
   // Currently unused
   unblock(clientId) {
     NAF.connection.adapter.unblock(clientId);
-    NAF.connection.entities.completeSync(clientId);
+    NAF.connection.entities.completeSync(clientId, true);
   }
 });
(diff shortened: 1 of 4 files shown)


gfodor:
Unfortunately I think this design needs to be revisited, since we need a similar mechanic to also allow kicking and muting from the presence list. My guess is we will need some kind of menu you can expect on the items in the presence list to reveal actions you can take on that person's avatar. (Though I don't mind us shipping this as a v1)

Also, how does this deal with objects the user has spawned?

johnshaughnessy:
Regarding objects the user has spawned:
Alice spawns a duck.
Bob hides Alice. The duck disappears because Alice owns it.
Bob unhides Alice. The duck reappears because it is part of Alice's "full sync" message (as long as Alice still owns the duck).


Likewise, Alice sees Bob's objects come back because Bob also sends Alice a "full sync".

If Charlie takes ownership of Alice's duck while Alice is hidden from Bob, then Bob never sees the duck. This is how hiding currently works in production. If Charlie deletes the duck, Bob will receive this message in the console:
image

gfodor:
please hold off on merging this, i'm going to work on a modal that lets users perform actions on others via the presence list, and will include 'unhide' as a verb (extracting the relevant bits from this PR)

gfodor:
Closing in favor of https://github.com/mozilla/hubs/pull/1310

PR #1281, opened 2019-04-30
Prepare pen, camera, and avatar before initial render

johnshaughnessy:
This PR attempts to prepare the materials and programs of the pen, camera, and avatar models for render before rendering them, so that there is less of a delay.

- [ ] Merge https://github.com/MozillaReality/three.js/pull/27
- [ ] Update package.json and package.lock
- [ ] Merge this


src/systems/hubs-systems.js
```diff
@@ -6,9 +6,28 @@ import { HoverMenuSystem } from "./hover-menu-system";
...
+
+async function preloadPenAndCamera(sceneEl) {
+  const objects = [];
+
+  const pen = await load(
+    proxiedUrlFor("https://asset-bundles-prod.reticulum.io/interactables/DrawingPen/DrawingPen-34fb4aee27.gltf")
+  );
+  pen.traverse(o => objects.push(o));
+
+  const camera = await load(cameraModelSrc);
```
netpro2k: Note that the camera-tool currently does not load models through the GLTF cache, so I don't think this preloading will actually help it very much, since it will have its own instances of geometry/textures

src/systems/hubs-systems.js
```diff
@@ -6,9 +6,28 @@ import { HoverMenuSystem } from "./hover-menu-system";
...
 import { SoundEffectsSystem } from "./sound-effects-system";
+import { proxiedUrlFor } from "../utils/media-utils";
+import { load, prepareForRender } from "../utils/preload";
+import cameraModelSrc from "../assets/camera_tool.glb";
+
+async function preloadPenAndCamera(sceneEl) {
+  const objects = [];
+
+  const pen = await load(
+    proxiedUrlFor("https://asset-bundles-prod.reticulum.io/interactables/DrawingPen/DrawingPen-34fb4aee27.gltf")
```
netpro2k: If you trace the crazy path by which pen gets loaded, you will see that we don't need to proxy this.

src/components/player-info.js
```diff
@@ -65,6 +66,18 @@ AFRAME.registerComponent("player-info", {
 
     const modelEl = this.el.querySelector(".model");
     if (this.data.avatarSrc && modelEl) {
+      modelEl.addEventListener(
```
netpro2k: Rather than doing the preloading here I would just do it in ui-root where you are early-updating player-info.

src/scene-entry-manager.js
```diff
@@ -143,6 +143,7 @@ export default class SceneEntryManager {
   };
 
   _setupPlayerRig = () => {
+    this.playerRig.querySelector(".model").object3D.visible = true;
```
netpro2k: You wont need this if you follow the above suggestion about preloading the avatar

src/systems/hubs-systems.js
johnshaughnessy: I don't need to _resolve_ this but do still want to proxy it

johnshaughnessy:
I think I got the camera preloading, but its lighting is wrong. I thought it would create a new program when the lights loaded.

PR #1282, opened 2019-04-30
Adds graceful channel migration

gfodor:
This PR makes it so the client transparently migrates to a new reticulum server when notified of a deploy. A few highlights:

- There's now a generic `migrateChannelToSocket` routine in `phoenix-utils` which lets you copy over the various bindings/handlers to a new channel instance for a new socket

- `HubChannel` now exposes a `migrateToSocket` function which can do the necessary internal munging of state to transition the channel and presence to the new server

- Fixed up a number of edge cases in presence event handling that were straight up bugs, and also made the handlers compatible with the scenario where a new `Presence` object may be attached to the hub channel across event handler calls. (This is what happens when we migrate to a new server.)

- Updates the presence event handlers to recognize the scenario where a user is leaving while still having other presence (eg, they are reconnecting) and so do not show a leave event

### DIFF
--- a/src/hub.js
+++ b/src/hub.js
@@ -21,7 +21,12 @@ import "./utils/threejs-positional-audio-updatematrixworld";
 import "./utils/threejs-world-update";
 import patchThreeAllocations from "./utils/threejs-allocation-patches";
 import { detectOS, detect } from "detect-browser";
-import { getReticulumFetchUrl, getReticulumMeta, invalidateReticulumMeta, migrateChannel } from "./utils/phoenix-utils";
+import {
+  getReticulumFetchUrl,
+  getReticulumMeta,
+  invalidateReticulumMeta,
+  migrateChannelToSocket
+} from "./utils/phoenix-utils";
 
 import nextTick from "./utils/next-tick";
 import { addAnimationComponents } from "./utils/animation";
@@ -715,7 +720,6 @@ document.addEventListener("DOMContentLoaded", async () => {
 
   getReticulumMeta().then(reticulumMeta => {
     console.log(`Reticulum @ ${reticulumMeta.phx_host}: v${reticulumMeta.version} on ${reticulumMeta.pool}`);
-    window.meta = reticulumMeta;
 
     if (
       qs.get("required_ret_version") &&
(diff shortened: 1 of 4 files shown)


src/hub.js
```diff
@@ -908,19 +942,26 @@ document.addEventListener("DOMContentLoaded", async () => {
...
 
-          if (!isInitialSync) return;
-          // Wire up join/leave event handlers after initial sync.
-          isInitialSync = false;
+          // HACK - Set a flag on the presence object indicating if the initial sync has completed,
+          // which is used to determine if we should fire join/leave messages into the presence log.
+          presence.__hadInitialSync = true;
+
+          presence.onJoin((sessionId, current, info) => {
+            if (!hubChannel.presence.__hadInitialSync) return;
```
brianpeiris: Might be worth adding a comment here that these onSync, onJoin, and onLeave handlers are reused when migrating servers, since this early exit doesn't make sense if you don't know about that context.

brianpeiris: Specifically that this early exit should only ever happen if we are in the middle of a migration (if I understand correctly)

src/utils/hub-channel.js
```diff
@@ -29,6 +30,33 @@ export default class HubChannel extends EventTarget {
     return this._permissions;
   }
 
+  // Migrates this hub channel to a new phoenix channel and presence
+  async migrateToSocket(socket, params) {
+    let presenceBindings;
+
+    // Unbind presence, and then set up bindings after reconnect
+    if (this.presence) {
```
brianpeiris: I suppose something would have gone very wrong if we end up here and the channel doesn't have a presence instance. Is it even possible? Maybe at least `console.warn` in an else clause?

src/utils/phoenix-utils.js
```diff
@@ -141,3 +145,35 @@ export function getPresenceProfileForSession(presences, sessionId) {
   const presence = entry[1];
   return (presence && presence.metas && presence.metas[0].profile) || {};
 }
+
+// Takes the given channel, and creates a new channel with the same bindings
+// with the given socket, joins it, and leaves the old channel after joining.
+export function migrateChannelToSocket(oldChannel, socket, params) {
```
brianpeiris: Oh boy. It looks like we're reaching into phoenix library internals here? Dunno if it's worth adding a comment to warn our future selves to watch out for breakages in new versions of the library (though I guess we use our own fork at the moment)

src/utils/phoenix-utils.js
```diff
@@ -141,3 +145,35 @@ export function getPresenceProfileForSession(presences, sessionId) {
...
+  }
+
+  return new Promise(resolve => {
+    const oldJoinPush = oldChannel.joinPush;
+    const joinPush = channel.join();
+
+    for (let i = 0, l = oldJoinPush.recHooks.length; i < l; i++) {
+      const item = oldJoinPush.recHooks[i];
+      joinPush.receive(item.status, item.callback);
+    }
```
brianpeiris: nit. I don't see a reason the above code needs to be inside the promise.

src/hub.js
gfodor: the early exit actually happens for the first sync event for each `Presence` object -- typically just the one we create at startup. but if you reconnect, we create another, and skip onJoins as well here in that scenario (so you don't see the whole room join when you reconnect.) i'll update the comment

src/hub.js
```diff
@@ -802,51 +802,83 @@ document.addEventListener("DOMContentLoaded", async () => {
...
+  };
+
+  retPhxChannel.on("notice", async data => {
+    // On Reticulum deploys, reconnect after a random delay until pool + version match deployed version/pool
+    if (data.event === "ret-deploy") {
+      await migrateToNewReticulumServer(data);
+    }
+  });
+
+  const hubPhxChannel = socket.channel(`hub:${hubId}`, createHubChannelParams());
```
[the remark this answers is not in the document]
gfodor: nice catch, I had it backwards -- I just wanted to include it in the main flow, but omit it from the reconnect flow, since my understanding is it is a one-time use token

src/utils/hub-channel.js
gfodor: yep this seems impossible more or less, since you won't receive the message to migrate unless you managed to get connected to the `ret` channel. so failure mode would be a busted `hub` channel with a working `ret` channel, but at that point client is hosed anyway.

src/utils/phoenix-utils.js
gfodor: Yep I can add a comment. This seemed to be the better way to do it than making it so we can re-materialize all of the event handlers in `hub.js` into a new channel. I can add a comment. Also, i forgot we were on a fork, we can move off of it at some point: my changes were merged into master 😎

src/hub.js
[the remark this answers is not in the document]
brianpeiris: Hmm, you're right, it is a one-time use token. Now that I think about it, discord users with unverified emails will not be able to re-join successfully after a reconnect. I'm not sure if they will be properly redirected to oauth either, so it's something we should test. This may just be a good reason to remove that flow, as you suggested, and just inform users that they need accounts with verified emails.

src/hub.js
brianpeiris: I don't think I understand how the early exit applies on the very first sync, given we set `__hadInitialSync` to true immediately before registering this handler.

PR #1286, opened 2019-04-30
Ensure we wire up websocket NAF handlers before initial syncs

gfodor:
The latest changes to filter on `phx-reliable` sources has a race condition that is easily exposed when using `vr_entry_type=2d_now`. It's unclear how this race condition could be exposed otherwise, but the symptoms are that other users will not see that person's avatar, which is similar to what we've started seeing happen anecdotally since around the time the last discord identity push happened. So perhaps there's some way this condition is being hit.

The root cause is that the `reliableTransport` adapter property was previously being set in the promise handler after `connect()` is called on the `networked-scene` -- this can fire after the initial sync messages are sent out for any existing `networked` objects since both of these are event handlers on the `connected` event emitted by `networked-scene`.

So, in practice, those messages would go out over data channels, and be dropped on the floor because they did not have the proper `source`. This PR updates these adapter property settings to be wired up in the `onConnected` event handler on the `NetworkConnection`, which is earlier in the flow than the `connected` event emission, and so ensure that no network data is sent before it happens.

### DIFF
--- a/src/hub.js
+++ b/src/hub.js
@@ -429,55 +429,56 @@ async function handleHubChannelJoined(entryManager, hubChannel, messageDispatch,
     });
 
     while (!scene.components["networked-scene"] || !scene.components["networked-scene"].data) await nextTick();
-    scene.components["networked-scene"]
-      .connect()
-      .then(() => {
-        let newHostPollInterval = null;
 
-        scene.emit("didConnectToNetworkedScene");
-        // When reconnecting, update the server URL if necessary
-        NAF.connection.adapter.setReconnectionListeners(
-          () => {
-            if (newHostPollInterval) return;
+    NAF.connection.onConnect(() => {
+      let newHostPollInterval = null;
 
-            newHostPollInterval = setInterval(async () => {
-              const currentServerURL = NAF.connection.adapter.serverUrl;
-              const newHubHost = await hubChannel.getHost();
-              const newServerURL = `wss://${newHubHost}`;
+      // When reconnecting, update the server URL if necessary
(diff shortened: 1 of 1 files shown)


src/hub.js
```diff
@@ -429,55 +429,56 @@ async function handleHubChannelJoined(entryManager, hubChannel, messageDispatch,
...
+        const payload = { dataType, data };
+
+        if (clientId) {
+          payload.clientId = clientId;
+        }
+
+        hubChannel.channel.push("naf", payload);
+      };
+
+      scene.emit("didConnectToNetworkedScene");
```
brianpeiris: Does it make sense to emit this on `connection.onConnect` now, when it was on `networked-scene.connect` before?

src/hub.js
```diff
@@ -429,55 +429,56 @@ async function handleHubChannelJoined(entryManager, hubChannel, messageDispatch,
...
-          hubChannel.channel.push("naf", payload);
-        };
-      })
-      .catch(connectError => {
-        // hacky until we get return codes
-        const isFull = connectError.error && connectError.error.msg.match(/\bfull\b/i);
-        console.error(connectError);
-        remountUI({ roomUnavailableReason: isFull ? "full" : "connect_error" });
-        entryManager.exitScene();
+    NAF.connection.onConnect(() => {
```
brianpeiris: Actually, I think it may be better to do this on the `"adapter-ready"` event I added recently, since this onConnect callback is essentially just `document.body.addEventListener("connected", ...)`.

gfodor: oh, you're right I guess this just puts it at the head of the list. adapter-ready seems smarter!

PR #1305, opened 2019-05-05
Feature/room owner change

gfodor:
Goes with https://github.com/mozilla/reticulum/pull/171

Updates the client to properly incorporate room creator assignment tokens. If you create a room and have a token, when you sign in, you'll become the room owner. Updates all the relevant flows in the app to ask you to sign in now instead of just not being visible, if you have a token to take ownership of the room.


src/react-components/settings-menu.js
```diff
@@ -89,14 +97,21 @@ export default class SettingsMenu extends Component {
...
                 </div>
                 <div className={styles.listItem}>
-                  <StateLink
-                    stateKey="modal"
-                    stateValue="rename_room"
-                    history={this.props.history}
-                    onClick={this.props.hideSettings}
+                  <a
+                    href="#"
+                    onClick={() =>
```
brianpeiris: Went on a pedantry google journey when I saw this. We use this `<a href="#" onClick=...` because we want buttons that look like links but don't actually link to anything. In fact the "Choose a Scene" link/button just above this is just a `<div>`.

I only care in this case because you forgot to `preventDefault` in the link variant here.

Ideally, we should be using `<button>`s here, that are styled like links. But I would be okay with using a `<span>` instead, so that we don't have to worry about `href` and `preventDefault`, and it's annoying to have to reset browser button styles just to make them look like links.

src/react-components/settings-menu.js
```diff
@@ -108,14 +123,21 @@ export default class SettingsMenu extends Component {
...
-                    stateValue="close_room"
-                    history={this.props.history}
-                    onClick={this.props.hideSettings}
+                  <a
+                    href="#"
+                    onClick={() =>
+                      this.props.performConditionalSignIn(
+                        () => this.props.hubChannel.can("update_hub"),
+                        () => {
+                          this.props.pushHistoryState("modal", "close_room");
```
brianpeiris: Aside: I'm increasingly inclined to stick functions like `performConditionalSignIn` and `pushHistoryState` into our `APP` global (under appropriate namespaces) since it's getting a bit silly to pass them along in props, like we do, as if they are ever going to mean different things to different react components.

src/react-components/ui-root.js
```diff
@@ -951,17 +951,21 @@ class UIRoot extends Component {
       <div className={entryStyles.entryPanel}>
         <div className={entryStyles.name}>
           <span>{this.props.hubName}</span>
-          {this.props.hubChannel.permissions.update_hub && (
-            <StateLink
-              stateKey="modal"
-              stateValue="rename_room"
-              history={this.props.history}
+          {this.props.hubChannel.canOrWillIfCreator("update_hub") && (
+            <button
```
brianpeiris: A button!

src/react-components/settings-menu.js
brianpeiris: I guess React has Contexts to solve this. TIL the problem is common enough to have been coined "prop drilling".

src/react-components/settings-menu.js
brianpeiris: I'll list some benefits of semantic HTML while I'm in rumination mode:
- Consistency and predictability within our own code base. Especially useful when styling elements to be able to apply styles consistently on particular element types.
- Accessibility, so that screen readers know how to read our app. (Arguably we could just put `role`/ARIA attributes on everything instead, but why do that when we get it for free if we are semantic from the start)
- Generally being nice to machines that might want to understand our app for whatever reason.
- Predictability for people that might attempt to write user scripts against our UI.

src/scene-entry-manager.js
```diff
@@ -217,45 +217,15 @@ export default class SceneEntryManager {
...
-            this._disableSignInOnPinAction = true;
-            el.setAttribute("pinnable", "pinned", !pin);
-            this._disableSignInOnPinAction = false;
-          }
-
-          handleReEntryToVRFrom2DInterstitial();
-        }
-      );
-    }
+    const action = pin ? async () => await this._pinElement(el) : async () => await this._unpinElement(el);
```
brianpeiris: Pretty sure these `async`s and `await`s are redundant. The Promises returned from `_pinElement` and `_unpinElement` can just be allowed to pass upwards.

src/utils/hub-channel.js
```diff
@@ -65,6 +77,7 @@ export default class HubChannel extends EventTarget {
   setPermissionsFromToken = token => {
     // Note: token is not verified.
     this._permissions = jwtDecode(token);
+    this.dispatchEvent(new CustomEvent("permissions-updated"));
```
brianpeiris: Maybe let's try shifting towards underscores in our event names.

src/react-components/settings-menu.js
gfodor: Yup, I actually had already replaced a few of these with buttons on my next branch :)

src/react-components/settings-menu.js
gfodor: ah yeah I agree a Context seems a good choice for a few of the things we have. i think it probably deserves its own refactor PR vs me just adding this one function, since I think there are several candidates.

src/scene-entry-manager.js
gfodor: looks like `_unpinElement` doesn't return anything, so I'll leave that one

PR #1306, opened 2019-05-06
Add basic badging to moderator

gfodor:
Adds a simple star badge to moderators (those with kick permissions) in presence, and adds an asterisk to their nametag. Goes with https://github.com/mozilla/reticulum/pull/172


src/components/player-info.js
```diff
@@ -50,7 +50,8 @@ AFRAME.registerComponent("player-info", {
     this.updateDisplayNameFromPresenceMeta(e.detail);
   },
   updateDisplayNameFromPresenceMeta(presenceMeta) {
-    this.displayName = presenceMeta.profile.displayName;
+    const isModerator = presenceMeta.perms && presenceMeta.perms.kick_users;
```
brianpeiris: Related to the comment I left on the reticulum PR, Our discord-based roles technically allow for users that have `update_hub` and `close_hub` permissions but not `kick_users` or `mute_users` permissions, so this logic would not display a badge for the former.

PR #1310, opened 2019-05-07
Perform user actions from presence list

gfodor:
Clicking on a user in the presence list will now give you a list of actions you can perform.

image

(Not a huge fan of this design, but decided it was easier than trying to figure out how to do context menus in the presence list in a way that was cross platform, etc.) Eventually the avatar image here can be replaced with their actual profile image, etc.


gfodor:
need to get badges out before pushing this, since this PR pulled in the badges PR inadvertently

PR #1312, opened 2019-05-08
New gravity settings

gfodor:
Experimental changes to in-game media physics:

- No more 'auto lock' based upon velocity
- By default, objects are not under the influence of gravity, and have extreme damping. So a light toss or push will cause them to float away slightly, and just letting go will keep them basically in place. Once an object comes to rest in this modality it will be made static so will not move when hit by other objects
- Tossing an object with some velocity will result in reduced damping and slight gravity at -1 m/s^2 for 2d objects and -2 m/s^2 for 3d objects (intended to make it easy to throw objects between users)
- There is now a "drop" button in the menu below the remove button which will cause the object to be permanently subject to Earth-style kinematics
- Pinned objects are always static


gfodor:
Oops, intended this to be WIP for now subject to in-game testing etc

src/components/drop-object-button.js
```diff
@@ -0,0 +1,33 @@
+AFRAME.registerComponent("drop-object-button", {
+  init() {
+    this.onClick = () => {
+      if (!NAF.utils.isMine(this.targetEl) && !NAF.utils.takeOwnership(this.targetEl)) return;
+
+      this.targetEl.setAttribute("sticky-object", { modifyGravityOnRelease: false });
+      this.targetEl.setAttribute("ammo-body", {
+        type: "dynamic",
+        gravity: { x: 0, y: -9.79, z: 0 }, // HACK setting to normal world gravity has bug in ammo-body not resetting via setGravity
```
InfiniteLee: this should be fixed now if you `npm update aframe-physics-system` again

src/components/sticky-object.js
```diff
@@ -3,9 +3,24 @@ const COLLISION_LAYERS = require("../constants").COLLISION_LAYERS;
 
 AFRAME.registerComponent("sticky-object", {
```
InfiniteLee: Can we call this something else now that its changed so much?

PR #1313, opened 2019-05-08
Mute microphone before entering

gfodor:
You can now mute your microphone before you enter a room. In the final confirmation page while in the room's lobby, you can check the 'Mute my Microphone' option to enter the room without an open mic. You will be able to mute and un-mute yourself as usual once in the room.

image

### DIFF
--- a/src/react-components/ui-root.js
+++ b/src/react-components/ui-root.js
@@ -1198,12 +1198,14 @@ class UIRoot extends Component {
                 srcSet="../assets/images/level_background@2x.png 2x"
                 className="audio-setup-panel__levels__icon-part"
               />
-              <img
-                src="../assets/images/level_fill.png"
-                srcSet="../assets/images/level_fill@2x.png 2x"
-                className="audio-setup-panel__levels__icon-part"
-                style={micClip}
-              />
+              {!this.state.muteOnEntry && (
+                <img
+                  src="../assets/images/level_fill.png"
+                  srcSet="../assets/images/level_fill@2x.png 2x"
+                  className="audio-setup-panel__levels__icon-part"
+                  style={micClip}
+                />
+              )}
               {this.state.audioTrack && !this.state.muteOnEntry ? (
                 <img
                   src="../assets/images/mic_level.png"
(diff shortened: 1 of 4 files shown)


src/react-components/ui-root.js
```diff
@@ -700,7 +701,7 @@ class UIRoot extends Component {
       showFullScreenIfAvailable();
     }
 
-    this.props.enterScene(this.state.mediaStream, this.state.enterInVR, this.props.hubId);
+    this.props.enterScene(this.state.mediaStream, this.state.enterInVR, this.state.muteOnEntry);
```
gfodor: I have no idea what this last arg was doing here. AFAICT it was never used/referenced

misslivirose:
Copy suggestion for the PR since this is going into the What's New page: 

"You can now mute your microphone before you enter a room. In the final confirmation page while in the room's lobby, you can check the 'Mute my Microphone' option to enter the room without an open mic. You will be able to mute and un-mute yourself as usual once in the room."

src/scene-entry-manager.js
```diff
@@ -110,6 +110,10 @@ export default class SceneEntryManager {
     })();
 
     this.scene.addState("entered");
+
+    if (muteOnEntry) {
+      this.scene.emit("action_mute");
```
johnshaughnessy: Maybe can just disable the microphone directly here, to avoid the "toggle mic" sound effect that is played in the `mute-mic` component's `onToggle`:
```
      NAF.connection.adapter.enableMicrophone(false);
      if (this.scene.is("muted")) this.scene.removeState("muted");
```

src/react-components/ui-root.js
johnshaughnessy: 🤷‍♂ 🤷‍♀

src/scene-entry-manager.js
gfodor: i'd be nervous about moving this code elsewhere from a future-proofing perspective -- currently the onMute routine in this component is the general function called for all muting. i agree the sound effect is not ideal

PR #1331, opened 2019-05-13
Avatar Browser

brianpeiris:
The new avatar browser lets you create and choose from multiple custom avatars.

image

- Replaces previous avatar-selector UI with a simpler profile panel with a simple preview, and an avatar media browser.
- The media browser now has an "avatars" source that pulls from featured avatars or our legacy static avatar list, and a user's saved avatars.
- The avatar editor is now accessed though the media browser, or through the profile panel, when you've selected a saved avatar.
- The avatar editor now creates and uploads a thumbnail for saved avatars. It also allows you to delete avatars.
- The avatar browser shows featured avatars by default, unless you have any saved avatars.
- Your selected avatar is now cleared from local storage when you sign out.
- StateLinks now have a `stateDetail` prop that can be used to pass extra state to routes.

PR Haiku:
identify now
embody many feelings 
be unique today


gfodor:
design nit: can "delete avatar" be made into the action color and not have the "..." -- i think the "..." idiom should be reserved for menu items

src/react-components/profile-entry-panel.js
```diff
@@ -62,28 +82,27 @@ class ProfileEntryPanel extends Component {
...
-
-  sendAvatarStateToIframe = () => {
-    this.avatarSelector.contentWindow.postMessage({ avatarId: this.state.avatarId }, location.origin);
+  setAvatarFromMediaResult = ({ detail: entry }) => {
+    this.setState({
+      avatarId: entry.id,
+      avatarType: getAvatarType(entry.id),
+      avatarGltfUrl: entry.gltfs.avatar
+    });
+    // Replace history state with the current avatar id since this component gets destroyed when we open the
```
netpro2k: Confused why you can't just load the profile from localstorage like normal?

src/react-components/profile-entry-panel.js
```diff
@@ -62,28 +82,27 @@ class ProfileEntryPanel extends Component {
...
+  async componentDidMount() {
+    this.setState({ avatarGltfUrl: await getAvatarGltfUrl(this.state.avatarId) });
     if (this.nameInput) {
       // stop propagation so that avatar doesn't move when wasd'ing during text input.
       this.nameInput.addEventListener("keydown", this.stopPropagation);
       this.nameInput.addEventListener("keypress", this.stopPropagation);
       this.nameInput.addEventListener("keyup", this.stopPropagation);
     }
-    window.addEventListener("message", this.setAvatarStateFromIframeMessage);
+    this.scene.addEventListener("action_selected_media_result_entry", this.setAvatarFromMediaResult);
```
netpro2k: Would have been nice for this to be an `onItemSelected` prop for the mediaBrowser, but maybe is more work than its worth.

netpro2k: On second thought that does't really work since the media browser is invoked by url not actually embedded as a component.

src/react-components/profile-entry-panel.js
```diff
@@ -189,32 +143,32 @@ class ProfileEntryPanel extends Component {
...
+              className={styles.chooseAvatar}
+              onClick={() => this.props.mediaSearchStore.sourceNavigateWithNoNav("avatars")}
+            >
+              <FormattedMessage id="profile.choose_avatar" />
+            </a>
+
+            <div className={styles.preview}>
+              <AvatarPreview avatarGltfUrl={this.state.avatarGltfUrl} />
+
+              {this.state.avatarType === AVATAR_TYPES.SKINNABLE && (
```
netpro2k: We will need to remember to update this when we add the owner id to the avatar response so that "real" featured avatars dont get an edit button.

src/react-components/ui-root.js
```diff
@@ -273,7 +276,7 @@ class UIRoot extends Component {
     this.props.scene.addEventListener("stateremoved", this.onAframeStateChanged);
     this.props.scene.addEventListener("share_video_enabled", this.onShareVideoEnabled);
     this.props.scene.addEventListener("share_video_disabled", this.onShareVideoDisabled);
-    this.props.scene.addEventListener("exit", this.exit);
+    this.props.scene.addEventListener("exit", () => this.exit());
```
netpro2k: Super minor since this is an exit event, but we won't be able to clean it up this way. Should just be bound in the constructoror when defined like most of the other functions in this class.

src/storage/media-search-store.js
```diff
@@ -66,21 +67,67 @@ export default class MediaSearchStore extends EventTarget {
...
+    const body = await res.text();
+
+    let result;
+    try {
+      result = JSON.parse(body);
+    } catch (e) {
+      result = body;
+    }
+
+    if (source === "avatar_listings" && result.entries.length === 0) {
```
netpro2k: I think there are some other changes we are going to need to make to have featured avatars work, but it would be nice to be able to start setting them up in the admin interface, so we should avoid making the call at all unless you have a flag set or something.

src/react-components/avatar-editor.js
```diff
@@ -233,88 +214,123 @@ export default class AvatarEditor extends Component {
...
+                {debug && this.textarea("description", "Description")}
+                {debug && this.fileField("glb", "Avatar GLB", "model/gltf+binary,.glb")}
+
+                {this.mapField("base_map", "Base Map", "image/*")}
+                {this.mapField("emissive_map", "Emissive Map", "image/*")}
+                {this.mapField("normal_map", "Normal Map", "image/*")}
+
+                {this.mapField("orm_map", "ORM Map", "image/*", false, "Occlussion (r), Roughness (g), Metallic (b)")}
+
+                {debug && this.checkbox("allow_promotion", "Allow Promotion")}
```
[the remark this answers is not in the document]
brianpeiris: Not sure if this would require legal sign off given we have specific [legal copy for Spoke scene promotion](https://github.com/mozilla/Spoke/blob/master/PROMOTION.md). 
Would it make sense to surface promotion alone, or remixing as well? Spoke also has separate legal copy for that. 
Not sure if it should hold this PR up. @gfodor?

src/react-components/media-browser.js
```diff
@@ -422,7 +501,7 @@ class MediaBrowser extends Component {
               className={styles.name}
               onClick={e => this.handleEntryClicked(e, entry)}
             >
-              {entry.name}
+              {entry.name || "\u00A0"}
```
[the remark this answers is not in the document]
brianpeiris: I wanted the link to occupy space even if there wasn't a name. Didn't wasn't to futz with CSS to make it behave well in all cases.

src/react-components/profile-entry-panel.js
brianpeiris: This is referring to the ephemeral state, before you actually mutate localstorage.

src/react-components/profile-entry-panel.js
brianpeiris: Yeah, it's not great. Might be worth rethinking how the whole media browser works if we keep running into problems with it.

src/react-components/avatar-editor.js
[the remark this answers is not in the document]
gfodor: shouldn't hold it up, i'll surface it in our next meeting but in general we are gtg on avatar customization. promotion/remixing should be off by default for avatars.

src/react-components/ui-root.js
brianpeiris: This was actually to fix a bug with scene exit. The exit method is already bound like other methods, but it takes a parameter, so I didn't want the event argument being passed along. Good point about the clean-up though, this does the wrong thing. I'll fix it.

src/storage/media-search-store.js
brianpeiris: Oh. Okay, I'll put this behind a query parameter so that we can test it when we're preparing the actual set of featured avatars.

src/react-components/media-browser.js
```diff
@@ -66,6 +71,10 @@ const DEFAULT_FACETS = {
     { text: "People", params: { filter: "people" } },
     { text: "Scenes", params: { filter: "scenes" } },
     { text: "Transport", params: { filter: "transport" } }
+  ],
+  avatars: [
+    { text: "Featured", params: { filter: "featured" } },
```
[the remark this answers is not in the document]
gfodor: I think we'll want to have an Avatars tab in media search (for spawning and sharing avatars with others) so we'll need them in facets there. The main thing that's a bit weird about facets right now is that there's no way to know which facet you recently filtered into, but if we made it feel like tabs it'd imply the search query was scoped to the facet (it isn't)

src/react-components/avatar-editor.js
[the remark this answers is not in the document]
brianpeiris: Unclear if you're saying that this PR *should* show the Allow Promotion and Allow Remixing checkboxes. If it should, shall I just copy and tweak the Scene PROMOTION.md and Scene REMIXING.md files from Spoke?

PR #1332, opened 2019-05-14
My Scenes now in scene picker

gfodor:
You can now browse the scenes you've created with Spoke from the scene picker, under "My Scenes".

### DIFF
--- a/src/react-components/media-browser.js
+++ b/src/react-components/media-browser.js
@@ -75,7 +75,8 @@ const DEFAULT_FACETS = {
   avatars: [
     { text: "Featured", params: { filter: "featured" } },
     { text: "My Avatars", params: { filter: "my-avatars" } }
-  ]
+  ],
+  scenes: [{ text: "Featured", params: { filter: "featured" } }, { text: "My Scenes", params: { filter: "my-scenes" } }]
 };
 
 class MediaBrowser extends Component {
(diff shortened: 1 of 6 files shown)


src/react-components/media-browser.js
```diff
@@ -387,23 +390,32 @@ class MediaBrowser extends Component {
...
+                      <div className={styles.tileContent}>
+                        <FontAwesomeIcon icon={faPlus} />
+                        <FormattedMessage id="media-browser.create-scene" />
+                      </div>
+                    </a>
+                  ) : (
+                    <a
+                      onClick={e => {
+                        e.preventDefault();
+                        window.dispatchEvent(new CustomEvent("action_create_avatar"));
```
robertlong: This should probably be another action?

src/react-components/media-browser.js
```diff
@@ -387,23 +390,32 @@ class MediaBrowser extends Component {
...
+                      }}
+                      className={styles.tileLink}
+                    >
+                      <div className={styles.tileContent}>
+                        <FontAwesomeIcon icon={faPlus} />
+                        <FormattedMessage id="media-browser.create-avatar" />
```
robertlong: Maybe this as well?

src/react-components/media-browser.js
brianpeiris: Looks okay to me. I think you just misread the diff.

PR #1334, opened 2019-05-14
Read Discord presence data from new profile field

mqp:
See https://github.com/MozillaReality/hubs-discord-bot/pull/49. It was necessary to move this into the "profile" field so that it could be easily updated by the bot without reconnecting. While we're at it, now there's lots more information about the bridges in there (both channel and guild name and ID) so we can make much nicer UI with it.

### DIFF
--- a/src/components/pin-networked-object-button.js
+++ b/src/components/pin-networked-object-button.js
@@ -80,7 +80,11 @@ AFRAME.registerComponent("pin-networked-object-button", {
     } else {
       const channels = [];
       for (const p of Object.values(presences)) {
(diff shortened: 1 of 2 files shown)


gfodor:
Do we need to update here too:
https://github.com/mozilla/hubs/blob/master/src/react-components/presence-list.js#L23

and

https://github.com/mozilla/hubs/blob/master/src/react-components/presence-list.js#L61

mqp:
My intention was to keep that key in as `true` (matching the other `context` keys `hmd` and `mobile`) because it feels reasonable to have that as a flag in there. After all, it could be reasonable in the future to have the Discord bot be in rooms even when they aren't currently bridged. I actually have to go make sure that stays put in my Discord bot PR though -- so nice catch.

PR #1336, opened 2019-05-14
Draw in world HUD with ShaderMaterial

johnshaughnessy:
This uses a texture sheet to draw the in-world hud in fewer calls (at the cost of a webgl program change).

From here it should be straightforward to include the rest of the buttons in the HUD (not the text). There are a few things to learn to do the object menus, but I think we can apply the same technique. 

Would like to check in here to get feedback and deploy this part before doing more.


src/assets/images/hud/spritesheet.json
```diff
@@ -0,0 +1,289 @@
+{
```
johnshaughnessy: I only need "frame" and "size" data from this. I can tweak the generator script to only give me this if need be.

src/components/hud-controller.js
```diff
@@ -76,6 +76,9 @@ AFRAME.registerComponent("hud-controller", {
     }
 
     hud.visible = !hudOutOfView || forceHudVisible;
+    if (hud.el.object3DMap.mesh) {
```
johnshaughnessy: temporary, will remove

src/systems/hud-system.js
```diff
@@ -0,0 +1,119 @@
...
+
+    if (hoverZone !== NONE && userinput.get(interaction.options.rightRemote.grabPath) && GRAB_EFFECT[hoverZone]) {
+      scene.systems["hubs-systems"].soundEffectsSystem.playSoundOneShot(SOUND_HOVER_OR_GRAB);
+      GRAB_EFFECT[hoverZone](scene);
+    }
+  };
+})();
+
+// This is written very component like, not system like...
+// Didn't bother to do multiple of these, but could do that if we do that for object menus
```
mqp: So why didn't you make it a component and put it on the hud-button-row entity?

src/systems/hud/memoize-sprites.js
```diff
@@ -0,0 +1,23 @@
...
+  return function getSprite(name) {
+    const memoizedSprite = memo.get(name);
+    if (memoizedSprite) {
+      return memoizedSprite;
+    }
+    const sprite = readNorm(name, spritesheet);
+    memo.set(name, sprite);
+    return sprite;
+  };
+};
```
mqp: If the goal is to establish a mapping from sprite name to `{x, y, w, h}` without creating the `{x, y, w, h}` objects every time, I would have just made an object that is the mapping. The spritesheet isn't very big so doing it lazily doesn't seem important.

mqp:
I like this approach about infinity times more than the "green channel means the offset" approach.

src/systems/hud/enum.js
```diff
@@ -0,0 +1,7 @@
+let h = 0;
+export const NONE = h++;
+export const BUTTON_MIC = h++;
+export const BUTTON_CREATE = h++;
+export const BUTTON_PEN = h++;
+export const BUTTON_CAMERA = h++;
+//export const BUTTON_INVITE = h++;
```
mqp: h+ https://humanityplus.org/

src/systems/hud/hud.frag
```diff
@@ -0,0 +1,36 @@
+precision mediump float;
+
+uniform sampler2D u_spritesheet;
+uniform vec4 u_stencils[5];
+uniform vec4 u_sprites[5];
```
mqp: Should the sprite coordinates array be a vector of integers instead of a vector of floats?

(It seems the stencil code could also be rewritten to represent the stencil coordinates as integers instead of floats, if we wanted.)

src/systems/hud/stencils.js
```diff
@@ -0,0 +1,77 @@
...
+export const determineHoverZone = (function() {
+  const WIDTH_IN_PIXELS = 512;
+  const MIC_X1 = (128 + 16) / WIDTH_IN_PIXELS;
+  const MIC_X2 = (128 + 16 + 64) / WIDTH_IN_PIXELS;
+  const CREATE_X1 = (128 + 16 + 64 + 16) / WIDTH_IN_PIXELS;
+  const CREATE_X2 = (128 + 16 + 64 + 16 + 128) / WIDTH_IN_PIXELS;
+  const PEN_X1 = (128 + 16 + 64 + 16 + 128 + 16) / WIDTH_IN_PIXELS;
+  const PEN_X2 = (128 + 16 + 64 + 16 + 128 + 16 + 64) / WIDTH_IN_PIXELS;
+  const CAM_X1 = (128 + 16 + 64 + 16 + 128 + 16 + 64 + 16) / WIDTH_IN_PIXELS;
+  const CAM_X2 = (128 + 16 + 64 + 16 + 128 + 16 + 64 + 16 + 64) / WIDTH_IN_PIXELS;
```
mqp: These constants line up with the offsets written into the stencil array, right? Why not write the one in terms of the other? e.g.

```javascript
const CAM_X1 = CAMERA.x;
const CAM_X2 = CAMERA.x + 64 / WIDTH_IN_PIXELS;
```

src/systems/hud/stencils.js
```diff
@@ -0,0 +1,77 @@
...
+
+  function contains1D(p, min, max) {
+    return p > min && p < max;
+  }
```
mqp: It doesn't seem right that this is exclusive on both ends.

src/systems/hud-system.js
```diff
@@ -0,0 +1,119 @@
...
+      return;
+    }
+    const frozen = scene.is("frozen");
+    this.el.object3D.visible = !frozen;
+    //this.el.object3DMap.mesh.visible = !frozen;
+
+    const userinput = scene.systems.userinput;
+    const interaction = scene.systems.interaction;
+    const point = interaction.intersectionPoint;
+    const xZeroToOne = (vec3.copy(point).add(OFFSET).x + WIDTH / 2) / WIDTH;
```
mqp: I would have just typed `(point.x + offset.x + WIDTH / 2) / WIDTH` and eliminated the temporary vector.

mqp:
It's difficult for me to mentally keep track of what units different stuff in this code is in -- i.e. "pixel offset relative to the whole HUD", "pixel offset relative to the button edge", "fraction of distance across the whole HUD", "fraction of distance across the button", etc.

johnshaughnessy:
Thank you for the review @mqp . This feedback and our conversations about it led me to a different approach that applies to more of the `a-image`s in the scene and does not need the (burdensome) "check which of the five buttons on this quad I am intersecting" part. I am closing this PR in favor of the new approach found here https://github.com/mozilla/hubs/pull/1379 .

PR #1344, opened 2019-05-16
Friendly display names

brianpeiris:
- Show new `communityIdentifier` presence field in avatar nametag in pause mode, and in user info panel.
- Fix flow for unverified discord users
- Show avatar thumbnail in user info panel

Goes with https://github.com/mozilla/reticulum/pull/178

### DIFF
--- a/src/assets/stylesheets/client-info-dialog.scss
+++ b/src/assets/stylesheets/client-info-dialog.scss
@@ -1,9 +1,21 @@
 @import 'shared';
 
+:local(.client-info-dialog) {
+  .dialog__box__contents__title {
+    margin-bottom: 10px;
+  }
+}
+
 :local(.title) {
   text-align: center;
   font-weight: bold;
   margin: 0 12px;
+}
+
+:local(.community-identifier) {
+  font-size: 12pt;
+  font-weight: normal;
+  margin-top: 3px;
 }
 
 :local(.client-info) {
--- a/src/assets/stylesheets/client-info-dialog.scss
+++ b/src/assets/stylesheets/client-info-dialog.scss
@@ -30,6 +30,8 @@
 }
 
 :local(.client-profile-image) {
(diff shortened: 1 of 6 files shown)


src/assets/stylesheets/client-info-dialog.scss
```diff
@@ -1,11 +1,23 @@
 @import 'shared';
 
+:local(.client-info-dialog) {
+  .dialog__box__contents__title {
```
gfodor: /shrug

src/react-components/client-info-dialog.js
```diff
@@ -61,26 +65,44 @@ export default class ClientInfoDialog extends Component {
...
+  render() {
+    const profile = this.getPresenceProfile();
+    if (!profile) return null;
+
+    const { displayName, communityIdentifier } = profile;
+    const { hubChannel, clientId, onClose } = this.props;
+    const title = (
+      <div className={styles.title}>
+        {displayName}
+        <div className={styles.communityIdentifier}>{communityIdentifier}</div>
```
gfodor: can `communityIdentifier` be null?

gfodor:
lgtm, one minor comment

src/assets/stylesheets/client-info-dialog.scss
brianpeiris: Heh, I'm exploiting our old BEM style CSS. I think I've become partial to @netpro2k's CSS structure now -- with `:local()` containers but plain `.foo-bar` sub class names. That would better allow for cases like this where we want to override child component styling. Not sure if there's a better way.

src/react-components/client-info-dialog.js
brianpeiris: Yup, it could be null or undefined. But this will just display nothing, as intended.

PR #1345, opened 2019-05-16
Add preview to avatar admin

netpro2k:
Adds preview as a column for avatars in the admin panel, and exposes account, order and attribution columns.

### DIFF
--- a/src/react-components/admin/avatars.js
+++ b/src/react-components/admin/avatars.js
@@ -68,13 +68,15 @@ const rowStyle = record => ({
 
 export const AvatarList = props => (
   <List {...props} filters={<AvatarFilter />} bulkActionButtons={false}>
-    <Datagrid rowStyle={rowStyle} expand={<Preview />}>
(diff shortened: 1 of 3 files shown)


src/react-components/admin/avatars.js
```diff
@@ -38,13 +45,30 @@ export const AvatarEdit = props => (
...
+    overflow: "hidden",
+    width: 200,
+    height: 200 * (16 / 9)
+  }
+};
+
+const Preview = withStyles(styles)(({ record, classes }) => (
+  <AvatarPreview
+    className={classes.preview}
+    avatarGltfUrl={getReticulumFetchUrl(`/api/v1/avatars/${record.avatar_sid}/avatar.gltf?v=${record.updated_at}`)}
```
gfodor: maybe we should DRY up this path generation?

netpro2k: Yeah I mean it should technically hit the avatar endpoint and et it from the response... But I didn't want to add another asnyc step here.

PR #1354, opened 2019-05-21
Support selecting real avatar listings

netpro2k:
Requires https://github.com/mozilla/reticulum/pull/179

Allows actually selecting avatar listings as an avatar. Only show the edit button when you have an avatar (not an avatar listing) selected, and you own it.

Using real avatar listings instead of legacy featured avatars still remains behind a flag (`use_avatar_listings`) so that we can test against the production database after configuring it correctly.

On production we will need to make sure all the featured avatar listings have their parent_avatar_listing_id set to the base avatar listing.

### DIFF
--- a/src/react-components/profile-entry-panel.js
+++ b/src/react-components/profile-entry-panel.js
@@ -9,10 +9,9 @@ import { SCHEMA } from "../storage/store";
 import styles from "../assets/stylesheets/profile.scss";
 import hubLogo from "../assets/images/hub-preview-white.png";
 import { WithHoverSound } from "./wrap-with-audio";
-import { AVATAR_TYPES, getAvatarGltfUrl, getAvatarType } from "../utils/avatar-utils";
+import { fetchAvatar } from "../utils/avatar-utils";
 import { handleTextFieldFocus, handleTextFieldBlur } from "../utils/focus-utils";
 import { replaceHistoryState } from "../utils/history";
-import { proxiedUrlFor } from "../utils/media-utils";
 import StateLink from "./state-link";
 
 import AvatarPreview from "./avatar-preview";
@@ -31,9 +30,8 @@ class ProfileEntryPanel extends Component {
 
   state = {
     avatarId: null,
-    avatarType: null,
     displayName: null,
-    avatarGltfUrl: null
+    avatar: null
   };
 
   constructor(props) {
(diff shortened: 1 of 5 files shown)


src/react-components/profile-entry-panel.js
```diff
@@ -84,26 +76,31 @@ class ProfileEntryPanel extends Component {
...
-    this.setState({ avatarGltfUrl: await getAvatarGltfUrl(this.state.avatarId) });
+  componentDidMount() {
     if (this.nameInput) {
       // stop propagation so that avatar doesn't move when wasd'ing during text input.
       this.nameInput.addEventListener("keydown", this.stopPropagation);
       this.nameInput.addEventListener("keypress", this.stopPropagation);
       this.nameInput.addEventListener("keyup", this.stopPropagation);
     }
     this.scene.addEventListener("action_selected_media_result_entry", this.setAvatarFromMediaResult);
+    // This handles editing avatars in the entry_step, since this component remains mounted with the same avatarId
```
brianpeiris: This comment is a bit misleading, since the profile panel _does_ get unmounted when you edit an avatar, but surprisingly, the action_avatar_saved event is emitted _after_ the profile panel is re-mounted, so this handler ends up working anyway.

src/scene-entry-manager.js
```diff
@@ -142,9 +142,9 @@ export default class SceneEntryManager {
 
   _setupPlayerRig = () => {
     this._setPlayerInfoFromProfile();
+    this.store.addEventListener("statechanged", this._setPlayerInfoFromProfile);
```
brianpeiris: This is a little dangerous, since it could result in an infinite loop if you have a 404'd avatar in local storage. `_setPlayerInfoFromProfile` will call `resetToRandomLegacyAvatar`, which updates store state.
This will be fixed in my PR anyway #1355

src/react-components/profile-entry-panel.js
```diff
@@ -9,10 +9,9 @@ import { SCHEMA } from "../storage/store";
 import styles from "../assets/stylesheets/profile.scss";
 import hubLogo from "../assets/images/hub-preview-white.png";
 import { WithHoverSound } from "./wrap-with-audio";
-import { AVATAR_TYPES, getAvatarGltfUrl, getAvatarType } from "../utils/avatar-utils";
+import { fetchAvatar } from "../utils/avatar-utils";
```
brianpeiris: Nice. I see what you mean now.
Looks like we can get rid of `getAvatarGltfUrl` now. Nothing else uses it.

src/react-components/profile-entry-panel.js
netpro2k: It actually does not get unmounted in the entry flow (when you haven't set your name) case

src/scene-entry-manager.js
netpro2k: This was already being done before, I just moved it so that it didn't get added multiple times as it is now in production

PR #1355, opened 2019-05-21
Avatar Bugs cont'd

brianpeiris:
- Fix another scenario where a 404'd avatar will break the profile panel
- Fix webgl context leaks
- Tweak avatar preview lighting
- Tweak create tile styling for FF on Android

### DIFF
--- a/src/react-components/avatar-preview.js
+++ b/src/react-components/avatar-preview.js
@@ -76,10 +76,10 @@ export default class AvatarPreview extends Component {
     this.controls = new THREE.OrbitControls(this.camera, this.canvas);
     this.controls.enablePan = false;
 
-    const light = new THREE.DirectionalLight(0xfdf5c2, 3);
+    const light = new THREE.DirectionalLight(0xf7f6ef, 1);
     light.position.set(0, 10, 10);
     this.scene.add(light);
-    this.scene.add(new THREE.HemisphereLight(0xb1e3ff, 0xb1e3ff, 3));
+    this.scene.add(new THREE.HemisphereLight(0xb1e3ff, 0xb1e3ff, 2.5));
 
     this.camera.position.set(-0.2, 0.5, 0.5);
     this.camera.matrixAutoUpdate = true;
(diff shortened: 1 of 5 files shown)


src/react-components/avatar-preview.js
```diff
@@ -107,7 +107,6 @@ export default class AvatarPreview extends Component {
       const dt = clock.getDelta();
       this.mixer && this.mixer.update(dt);
       this.previewRenderer.render(this.scene, this.camera);
-      this.snapshotRenderer.render(this.scene, this.snapshotCamera);
```
netpro2k: Woops, this was silly

src/react-components/avatar-preview.js
```diff
@@ -148,8 +147,24 @@ export default class AvatarPreview extends Component {
...
   componentWillUnmount = () => {
+    // Gotta be particularly careful about disposing things here since we will likely create many avatar
+    // previews during a session and Chrome will eventually discard the oldest webgl context if we leak
+    // contexts by holding on to them directly or indirectly.
+
     this.scene && this.scene.traverse(disposeNode);
-    this.previewRenderer && this.previewRenderer.dispose();
+    if (this.previewRenderer) {
+      this.previewRenderer.dispose();
+      this.previewRenderer.context = null;
```
netpro2k: Surprised this doesn't just get handled by GC. Is this just to push the GC along? If not, wonder if something else might still be holding onto the renderer...

brianpeiris: I think it does get handled by GC, I just wanted to be more certain about all the context references.
Interestingly, it seems that Chrome ended up hanging on to the contexts if you didn't dispose the OrbitControls, not only because orbit controls references the canvas, but because it attaches event listeners to the canvas. So that was the primary leak in Chrome at least.

src/react-components/avatar-preview.js
brianpeiris: Heh, yeah it was my fault. Just a debug line that I missed.

PR #1356, opened 2019-05-21
Adds missing cannon dependency to package.json

liebeskind:
Hubs not compiling because missing cannon in package.json.  It's possible you have this installed locally.


InfiniteLee:
This is a result of an unfortunate hack I introduced to prevent loading `cannon.js` as it is no longer a required dependency for Hubs. Unfortunately, `aframe-physics-system` still lists it as a dependency, although it technically is not required anymore. That should get resolved soon, but until then, we want to avoid bundling `cannon.js` as we now use `ammo.js` for the underlying physics.
If you're having issues with running Hubs, I recommend making sure that you are using `npm ci` to install the latest packages as listed in `package-lock.json` and if you happen to need to install new package or update an existing one, to make sure to remove any added `cannon.js` lines in `package-lock.json`.
This is not really an ideal solution, so either we'll revert it back to how it was, or `aframe-physics-system` will get updated to not actually require `cannon.js` anymore.
This issue will be tracked in #1357.

PR #1358, opened 2019-05-22
Spoke defined collision (heightfields/trimesh)

InfiniteLee:
Adds support for spoke-defined scene collision and support for heightfields.
We will now look for either heightfield data, or a trimesh mesh to use for creating environment collision. If neither are present, a heuristic based on triangle density (calculated from the bvh tree of the scene) will determine if it is safe to use a run-time generated trimesh or not (if not, it backs out to using either the navmesh if available or a simple box floor).

spoke_heightfield


misslivirose:
A blog post about physics for Hubs would probably be really interesting for Mozilla Hacks

PR #1360, opened 2019-05-23
Improved gamepad support

### LINKED ISSUE #1283: Gamepad bindings do not work with a standard Xbox controller
The [gamepad tester tool](https://html5gamepad.com/) seems to work, but the userinput system is not picking it up.

The gamepad shows up as "xinput" in Firefox and "Xbox 360 Controller (XInput STANDARD GAMEPAD)" in Chrome.

We should also test using the gamepad with a VR headset.

brianpeiris:
Fixes support for gamepads (Xbox Controller, etc). Works in 2D and VR mode. In 2D mode, input dynamically switches between mouse/keyboard and gamepad controller when either device is detected to be in use.

Closes #1283

### DIFF
--- a/src/components/character-controller.js
+++ b/src/components/character-controller.js
@@ -73,8 +73,8 @@ AFRAME.registerComponent("character-controller", {
     this.accelerationInput.set(axes[0], 0, axes[1]);
   },
 
-  setAngularVelocity: function(event) {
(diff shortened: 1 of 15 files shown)


johnshaughnessy:
Thank you for working on this! Glad that gamepads are working again :D

PR #1372, opened 2019-05-29
Ensure we only remove entities that are owned

gfodor:
It's a bad idea to remove entities from the scene which are networked and not owned -- until https://github.com/MozillaReality/networked-aframe/pull/29 this manifests in a bug but in general this is a bad idea since you should either take ownership of an object before removing it, so all users see it removed, or you should skip removal of unowned objects in situations (like networked counter) where you are trying to rate limit.

### DIFF
--- a/src/components/remove-networked-object-button.js
+++ b/src/components/remove-networked-object-button.js
@@ -14,6 +14,7 @@ AFRAME.registerComponent("remove-networked-object-button", {
       this.el.parentNode.setAttribute("visible", false);
 
       this.targetEl.addEventListener("animationcomplete", () => {
+        NAF.utils.takeOwnership(this.targetEl);
         this.targetEl.parentNode.removeChild(this.targetEl);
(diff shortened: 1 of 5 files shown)


src/components/remove-networked-object-button.js
```diff
@@ -14,6 +14,7 @@ AFRAME.registerComponent("remove-networked-object-button", {
       this.el.parentNode.setAttribute("visible", false);
 
       this.targetEl.addEventListener("animationcomplete", () => {
+        NAF.utils.takeOwnership(this.targetEl);
```
brianpeiris: I guess you probably already know this, but `takeOwnership` does not always guarantee ownership after it is called.

gfodor: yep, I considered that -- it seems right to remove it anyway if it fails for this case since at that point the object has scaled down to be tiny

PR #1374, opened 2019-05-30
Fix migration for really old spoke scenes

InfiniteLee:
Migrate even older versions of spoke scenes with heightfields properly.
Older spoke scenes used "HUBS_components" extension without a version number, but still contained heightfield components that can be inflated. This causes bad collision geometry in old scenes with this format.

### DIFF
--- a/src/components/gltf-model-plus.js
+++ b/src/components/gltf-model-plus.js
@@ -232,15 +232,20 @@ function runMigration(version, json) {
   if (version < 2) {
     //old heightfields will be on the same node as the nav-mesh, delete those
     const oldHeightfieldNode = json.nodes.find(node => {
-      return (
-        node.extensions &&
-        node.extensions.MOZ_hubs_components &&
-        node.extensions.MOZ_hubs_components.heightfield &&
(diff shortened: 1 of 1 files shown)


src/components/gltf-model-plus.js
```diff
@@ -232,15 +232,20 @@ function runMigration(version, json) {
...
-      return (
-        node.extensions &&
-        node.extensions.MOZ_hubs_components &&
-        node.extensions.MOZ_hubs_components.heightfield &&
-        node.extensions.MOZ_hubs_components["nav-mesh"]
-      );
+      let components = null;
+      if (node.extensions && node.extensions.MOZ_hubs_components) {
+        components = node.extensions.MOZ_hubs_components;
+      } else if (node.extensions && node.extensions.HUBS_components) {
```
robertlong: HUBS_components won't have the version set. You'll want to run that separately.

InfiniteLee: version defaults to `0` below in the case a version number isn't found.

PR #1379, opened 2019-05-30
Render sprites in one draw call.

johnshaughnessy:
This PR introduces the `SpriteSystem`, which renders many sprites from a spritesheet in one draw call.

good things:
 - one draw call for all sprites
 - one texture upload for all sprites
 - small change from "a-image" to "a-entity sprite"

areas ripe for further development:
 - slice9's
 - text
 - media-images

Sprites on:
image

Sprites off:
image

Note: This screenshot was taken before I added in the highlight effect so the sprites are not lit up. The highlight effect has since been added.


src/systems/sprites.js
```diff
@@ -0,0 +1,255 @@
...
+      }
+    }
+    return intersects;
+  }
+  constructor(scene) {
+    this.missingSprites = [];
+    this.spriteComponents = [];
+    this.maxSprites = 100;
+    Promise.all([createImageTexture(spritesheetPng), waitForDOMContentLoaded()]).then(([spritesheetTexture]) => {
+      const geometry = new THREE.BufferGeometry();
```
mqp: I would break off these next bunch of lines into a separate function `createSpriteGeometry`.

src/systems/sprites/sprite.vert
```diff
@@ -0,0 +1,18 @@
...
+in vec2 a_uvs;
+uniform mat4 projectionMatrix;
+uniform mat4 modelViewMatrix;
+
+out vec2 v_uvs;
+
+void main() {
+  mat4 mv = mat4(mvCol0, mvCol1, mvCol2, mvCol3);
+  gl_Position = projectionMatrix * modelViewMatrix * mv * vec4(a_vertices, 1.0);
+  v_uvs = a_uvs;
```
mqp: I actually didn't know you could pass arbitrary computed values from a vertex shader into a fragment shader like this.

johnshaughnessy: `v_uvs` is a `varying` but since this is written with `version 300 es` we replace that keyword with `out` in the vertex shader and `in` in the fragment shader.

src/components/sprite-button.js
```diff
@@ -0,0 +1,67 @@
...
+    const image = active ? (hovering ? "activeHoverImage" : "activeImage") : hovering ? "hoverImage" : "image";
+    if (this.el.components.sprite) {
+      if (this.data[image]) {
+        this.el.components.sprite.data.name = this.data[image];
+      } else {
+        console.warn(`No ${image} image on me.`, this);
+      }
+    } else {
+      console.warn("no sprite");
+    }
```
mqp: Could we have just extended `icon-button` to optionally use sprites, if the `sprite` component is also hanging around? It annoys me to have `icon-button` and `sprite-button` both in tree with duplicated code for all the other stuff they do.

src/components/sprite-button.js
```diff
@@ -0,0 +1,67 @@
+/**
+ * A button with an image, tooltip, hover states.
+ * @namespace ui
+ * @component icon-button
```
mqp: Fix if we plan on keeping this as a separate component.

src/systems/sprites.js
johnshaughnessy: 👍

src/components/sprite-button.js
johnshaughnessy: 👍

src/systems/sprites/sprite.frag
```diff
@@ -0,0 +1,49 @@
+#version 300 es
```
robertlong: I think this needs to be converted to GLSL ES 2.0 for compatibility with WebGL 1.0. I'd test on iOS to be certain

src/systems/sprites/sprite.vert
```diff
@@ -0,0 +1,28 @@
+#version 300 es
+
+uniform mat4 modelViewMatrix;
+uniform mat4 projectionMatrix;
+
+in vec4 mvCol0;
```
robertlong: Matrix transforms per vertex is a bit too heavy. I'd store matrix transforms in a uniform array and store an index into that array per vertex.

src/systems/sprites/sprite.vert
```diff
@@ -0,0 +1,28 @@
...
+in vec4 mvCol1;
+in vec4 mvCol2;
+in vec4 mvCol3;
+in vec3 a_vertices;
+in vec2 a_uvs;
+out vec2 v_uvs;
+in float a_hubs_EnableSweepingEffect;
+flat out float v_hubs_EnableSweepingEffect;
```
robertlong: Kinda wish this effect was done in another forward rendered pass instead of having to be injected into each of our shaders, but that's out of the scope of this PR.

src/systems/post-physics.js
```diff
@@ -0,0 +1,16 @@
+import { SpriteSystem } from "./sprites";
+
+// wait for aframe physics system to be registered...
```
robertlong: 🤔

src/systems/camera-mirror.js
```diff
@@ -90,6 +90,8 @@ AFRAME.registerSystem("camera-mirror", {
       if (playerHud) {
         playerHudWasVisible = playerHud.visible;
         playerHud.visible = false;
+        // playerHud.el.object3DMap.mesh.visible = false;
+        // TODO: playerHud sprites should be invisible to camera
```
robertlong: Is this out of scope for this PR?

src/systems/sprites/sprite.vert
netpro2k: Better yet since all of these are all always just flat quads, we can send way less data by using instancing, and sending these transforms as instance attributes.

src/systems/sprites/sprite.frag
netpro2k: Yeah sadly we should write all our shaders such that they run on WebGL1, and then rewrite them at runtime to work on WebGL2 when we are using it (this is what we are doing now for all our other shaders). You will also need to make sure multiview is accounted for (likely just by letting the shader code get patched at runtime like we are doing for our other shaders)

src/systems/sprites/sprite.vert
```diff
@@ -0,0 +1,28 @@
...
+in vec2 a_hubs_SweepParams;
+flat out vec2 v_hubs_SweepParams;
+out vec3 hubs_WorldPosition;
+
+void main() {
+  mat4 mv = mat4(mvCol0, mvCol1, mvCol2, mvCol3);
+  gl_Position = projectionMatrix * modelViewMatrix * mv * vec4(a_vertices, 1.0);
```
netpro2k: traditionally the modelview matrix is for going from model space to view space. I assume mv is just from going from model to world space, then you want to take that into view space. So what you really want here is just the inverse camera matrix (view matrix) and the model world matrix. Since the sprite's worldMatrix ends up being identity anyway, this is the same thing as modelViewMatrix, albiet with a slightly more confusing name.

I think renderObject computes modelViewMatrix for every object regardless of if we use it or not, so its makes sense to keep as is, though I would maybe change the name of `mv` to make it look less like the view matrix is being applied twice.

src/systems/sprites/sprite.vert
```diff
@@ -0,0 +1,28 @@
+#version 300 es
+
+uniform mat4 modelViewMatrix;
+uniform mat4 projectionMatrix;
+
+in vec4 mvCol0;
+in vec4 mvCol1;
+in vec4 mvCol2;
+in vec4 mvCol3;
+in vec3 a_vertices;
```
netpro2k: I would make these singular, and I would call this one `pos`, as `vertex` is a bit ambiguous

src/components/hoverable-visuals.js
```diff
@@ -8,7 +8,6 @@ const interactorTwoTransform = [];
  */
 AFRAME.registerComponent("hoverable-visuals", {
   schema: {
-    cursorController: { type: "selector" },
```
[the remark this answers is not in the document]
johnshaughnessy: Correct

src/systems/sprites.js
```diff
@@ -0,0 +1,321 @@
...
+        side: THREE.DoubleSide,
+        transparent: true
+      });
+      this.mesh = new THREE.Mesh(createGeometry(this.maxSprites), material);
+      const el = document.createElement("a-entity");
+      el.classList.add("ui");
+      scene.appendChild(el);
+      el.setObject3D("mesh", this.mesh);
+      this.mesh.frustumCulled = false;
+      this.mesh.renderOrder = window.APP.RENDER_ORDER.HUD_ICONS;
```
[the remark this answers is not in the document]
johnshaughnessy: No, transparency issues arise here because the sprites conflict with the slice9's. If they aren't sorted correctly then they look wrong.

src/systems/sprites/sprite.vert
johnshaughnessy: Leaving instancing / reducing per-vertex data as a TODO for now, especially given that we will likely work to combine our "batching" efforts in the near future.

src/systems/sprites/sprite.vert
johnshaughnessy: I would like to learn more about how this can be done!

src/systems/camera-mirror.js
johnshaughnessy: nope. did it

src/systems/post-physics.js
johnshaughnessy: Leaving as a TODO for now

PR #1380, opened 2019-05-31
Feature/pwa

gfodor:
Adds a PWA manifest.


gfodor:
Thanks @cvan for the original version of this work!

PR #1383, opened 2019-05-31
Add optimized billboarding

gfodor:
This PR:

- Stops calling `lookAt` for billboarding on non-visible objects
- Sets the community identifier to invisible so its not updated unless needed
- Stop doing auto matrix update for billboarded objects
- Only performs look at billboarding for objects whose origins are visible to the camera, in mobile VR mode

Because of the latter performance optimizations, I forked.


src/components/billboard.js
```diff
@@ -0,0 +1,66 @@
...
+
+  _updateIsInView: (function() {
+    const frustum = new THREE.Frustum();
+    const frustumMatrix = new THREE.Matrix4();
+    const cameraWorld = new THREE.Vector3();
+    const isInViewOfCamera = (screenCamera, pos) => {
+      frustumMatrix.multiplyMatrices(screenCamera.projectionMatrix, screenCamera.matrixWorldInverse);
+      frustum.setFromMatrix(frustumMatrix);
+      return frustum.containsPoint(pos);
+    };
```
mqp: It's kind of frustrating that we aren't reusing the work that the renderer is doing to determine whether things are frustum-culled. I can see that it's not so easy, though, because if we reuse the renderer's `_frustum` property we'll be one frame behind. I guess in the ideal world we would need to rework the order of operations to first find the view frustum for the upcoming frame, then run the application tick, respecting that frustum, and then run the renderer, respecting that frustum.

gfodor: for the purposes of this kind of optimization being a frame behind wouldn't be a big deal -- but i think a three.js patch to do this would be iffy because it'd mean at least one more branch and a bunch of memory writes each frame per object

src/components/billboard.js
```diff
@@ -0,0 +1,66 @@
+const isMobileVR = AFRAME.utils.device.isMobileVR();
+
+// Billboard component that only updates visible objects and only those in the camera view on mobile VR.
```
mqp: Why did you enable this optimization only on mobile?

src/components/billboard.js
```diff
@@ -0,0 +1,66 @@
...
+AFRAME.registerComponent("billboard", {
+  init: function() {
+    this.target = new THREE.Vector3();
+    this._updateBillboard = this._updateBillboard.bind(this);
+    this._updateIsInView = this._updateIsInView.bind(this);
+
+    if (isMobileVR) {
+      this.el.sceneEl.systems["frame-scheduler"].schedule(this._updateIsInView, "billboards");
+    }
```
mqp: Doing this only one per frame seems pretty bad to me. It means that if there are 20 people in the room, you could spend half a second maybe looking at someone before their nametag suddenly rotates to face you. I feel that will seem like a bug.

src/components/billboard.js
gfodor: Basically I've been adding optimizations that technically degrade the experience (eg this one causes weird billboarding behavior if you move your camera quickly, since there's latency before it is corrected) so I think we should only apply these to mobile VR contexts since that is where we have the main sensitivity and susceptibility to lower frame rates

src/components/billboard.js
gfodor: I agree this is ultimately the tradeoff in exchange for a higher frame rate on mobile VR, we are taking on that visual impact. Having multiple pieces of work per frame or a better regulator would be an improvement but in general our approach to spreading across work is a one-per-frame within a scope model right now

src/components/billboard.js
mqp: You wouldn't need a three.js patch to do it. You would just use the `_frustum` property that already gets set every frame on the renderer object.

I have yet to be convinced that being a frame behind on things is not a big deal. It seems like common sense to me that maximally crisp-feeling responsiveness, state transitions, and behavior is important in making games and other software "feel good", and I have heard plenty of design stories about smart game developers working hard to cut frame delays on things, so I see no reason to doubt my common sense. I think we just wish it wasn't a big deal because it's hard.

src/components/billboard.js
mqp: I slightly prefer consistency between platforms, but it probably doesn't matter much.

src/components/billboard.js
```diff
@@ -0,0 +1,66 @@
...
+        this.isInView = false;
+        return;
+      }
+
+      if (!this.playerCamera) {
+        this.playerCamera = document.querySelector("#player-camera").getObject3D("camera");
+      }
+
+      if (!this.playerCamera) return;
+      this.el.object3D.getWorldPosition(cameraWorld);
```
mqp: `cameraWorld` seems like a misnomer here.

src/components/billboard.js
```diff
@@ -0,0 +1,66 @@
...
+  },
+
+  _updateIsInView: (function() {
+    const frustum = new THREE.Frustum();
+    const frustumMatrix = new THREE.Matrix4();
+    const cameraWorld = new THREE.Vector3();
+    const isInViewOfCamera = (screenCamera, pos) => {
+      frustumMatrix.multiplyMatrices(screenCamera.projectionMatrix, screenCamera.matrixWorldInverse);
+      frustum.setFromMatrix(frustumMatrix);
+      return frustum.containsPoint(pos);
```
mqp: This seems kind of suspicious -- doesn't it mean that the thing will be considered in view only iff the center is in view? It seems better to check whether the bounding box or sphere of the entity intersects the frustum.

src/components/billboard.js
gfodor: I was referring to billboarding specifically, and our use of it at that. A frame delay of the nametags being rotated to face you would be imperceptible.

src/components/billboard.js
gfodor: that's true. this code is lifted from the ik controller check, i don't remember if i had a reason to avoid the box check. perhaps concerns about performance or complexity. in this case, the visual artifact is more profound than in the avatar case so i think a box check is probably warranted.

gfodor: updated to do this. not worried about performance here because the box check happens in the scheduled frame.

PR #1387, opened 2019-06-01
Fix issues with scaling animation with extremely large objects

InfiniteLee:
update animejs with a fix to handle values with exponents (see: https://github.com/MozillaReality/anime/commit/3bec50077f08b242a84c6d0584472060e41ea6e2)
handle case when initalScale is smaller than mesh.scale

This should fix issues caused with certain objects (e.g. https://poly.google.com/view/51T1pmZ73vP) that are very large.


package.json
```diff
@@ -41,12 +41,13 @@
     "@fortawesome/fontawesome-svg-core": "^1.2.2",
     "@fortawesome/free-solid-svg-icons": "^5.2.0",
     "@fortawesome/react-fontawesome": "^0.1.0",
-    "aframe": "github:mozillareality/aframe#remove-text",
+    "aframe": "github:mozillareality/aframe#hubs/master",
```
gfodor: I think this is due to @mqp 's latest changes -- probably needs to be revertted?

gfodor:
lgtm other than the package json comment. doesn't leave a warm fuzzy feeling, but the chance this code ever would has long past :)

package.json
InfiniteLee: `remove-text` already got merged into master afaict

PR #1396, opened 2019-06-03
Support for in world avatar links

netpro2k:
Currently avatar links get incorrectly detected as room urls,  this fixes them to be correctly identified as avatar urls and adds a "use avatar" button to instantly switch to the linked avatar.


gfodor:
lgtm

PR #1386, opened 2019-06-01
Degredation optimizations

gfodor:
This does a bunch of optimizations, with a slight focus on slightly degrading the user experience on mobile VR in the name of performance:

- Stop running the tips tick on mobile VR
- Stop doing audio analysis on silent avatars
- Update one bubble invader per tick on mobile VR
- Skip work in hover visuals
- Fix a few other small things

PR #1398, opened 2019-06-03
Reopen https://github.com/mozilla/hubs/pull/1386

gfodor:
Reopen https://github.com/mozilla/hubs/pull/1386


gfodor:
i didn't do a full profile comparison -- the profiler is noisy but these appeared in most frames -- particularly the ones which are linear with the number of avatars seem good to have sped up (tho some of these are pretty speculative like the hover stuff)

PR #1399, opened 2019-06-04
Install Hubs Progressive Web App

gfodor:
In compatible browsers (Firefox Android, Chrome PC or Android), you'll now be prompted to install the Hubs Progressive Web App.

image


arthurmougin:
Now PWA install is supported on the Oculus Browser

PR #1404, opened 2019-06-04
Bug/environment collision while holding

gfodor:
This PR updates the behavior so that floaty objects do not intersect with other objects unless they have some form of gravity. (This fixes the issue where, for example, trying to place objects using the cursor results in environment collision.)


InfiniteLee:
lgtm

PR #1405, opened 2019-06-04
Handle avatar loading errors

### LINKED ISSUE #1315: Using an unresolvable avatar url in the lobby flow makes avatar invisible but does not block
I think that I've now got a repro case related to #1302 - if you try to change to a custom GLB avatar in the entry flow, you do not have a fallback avatar when the avatar fails, and you become invisible (and potentially not heard) by other people in the scene.

Repro steps (I'm seeing a 100% repro): 
1) Create or enter a room
2) While in the lobby, open the user list and hit the pencil to edit your avatar
3) In the 'custom URL' field, enter https://fake.glb
4) Observe that this is non blocking and you are able to enter the room without an avatar

brianpeiris:
- Catch model loading errors when loading avatars
- Display an error message in the avatar preview
- Reset to random legacy avatar if avatar in local storage is broken
- Also fixes issue with custom avatar url dialog where it would display the wrong dialog on the "my avatars" screen if you were not signed in.

image

Fixes #1315

### DIFF
--- a/src/react-components/media-browser.js
+++ b/src/react-components/media-browser.js
@@ -212,9 +212,8 @@ class MediaBrowser extends Component {
   };
 
   showCustomMediaDialog = source => {
-    const isSceneApiType = source === "scene_listings" || source === "scenes";
-    // Note: The apiSource for avatars *is* actually singular. We should probably fix this on the backend.
-    const isAvatarApiType = source === "avatar_listings" || source === "avatar";
+    const isSceneApiType = source === "scenes";
+    const isAvatarApiType = source === "avatars";
     this.pushExitMediaBrowserHistory(!isAvatarApiType);
     const dialog = isSceneApiType ? "change_scene" : isAvatarApiType ? "avatar_url" : "create";
     pushHistoryState(this.props.history, "modal", dialog);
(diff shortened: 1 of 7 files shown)


gfodor:
Thanks for fixing these up! one small design nit, I'd just make the error text dark grey, and perhaps make the warning icon yellow if it's easy.

src/react-components/media-browser.js
```diff
@@ -212,9 +212,8 @@ class MediaBrowser extends Component {
   };
 
   showCustomMediaDialog = source => {
-    const isSceneApiType = source === "scene_listings" || source === "scenes";
-    // Note: The apiSource for avatars *is* actually singular. We should probably fix this on the backend.
```
johnshaughnessy: Were these the back-end change that needed to happen? https://github.com/mozilla/reticulum/pull/179

johnshaughnessy: Never mind about this; You cleared it up for me in person. Maybe check in with gfodor about why he wrote this to inspect the search results' types rather than the urlSource.

brianpeiris: Greg preferred apiSource over urlSource since urlSource was more of a UI implementation detail where as apiSource was theoretically more concrete and less likely to change. However, it seems urlSource is now pretty solidified, but it would be nice to rely on apiSource if our REST responses were consistent in failure cases.

Issue #1413, opened 2019-06-05
Spawn points in Firefox doesn't work


smeybi:
Hi,
Spawn points created in spoke do not work in Firefox 67. In Chrome they works.

I've tested with 2 Computer, the same issue. You spawn in FF somewhere in the middle.
Here an example, you should spawn at one of the two spawn points at the side.
https://hubs.mozilla.com/K4GiBqf/spry-alarmed-nation

gfodor:
interesting, can confirm!

gfodor:
should be fixed! let us know if you still see this issue.

PR #1416, opened 2019-06-06
Fix missing update matrices

gfodor:
We were missing calls to updateMatrices in three-utils, which (I think) was causing spawn points to sometimes not be applied properly.


gfodor:
might fix https://github.com/mozilla/hubs/issues/1413

PR #1418, opened 2019-06-06
Fix custom glb avatars

brianpeiris:
This fixes Custom GLB avatars, which don't have a Hubs skeleton, so that they have a voice, correct avatar movement and avatar menus. Note, we treat the GLB as a head instead of a body.

### DIFF
--- a/src/components/gltf-model-plus.js
+++ b/src/components/gltf-model-plus.js
@@ -265,6 +265,10 @@ export async function loadGLTF(src, contentType, preferredTechnique, onProgress,
 
   const parser = await new Promise((resolve, reject) => gltfLoader.createParser(gltfUrl, resolve, onProgress, reject));
 
+  if (jsonPreprocessor) {
+    parser.json = jsonPreprocessor(parser.json);
+  }
+
   let version = 0;
   if (
     parser.json.extensions &&
@@ -303,10 +307,6 @@ export async function loadGLTF(src, contentType, preferredTechnique, onProgress,
 
       node.extras.gltfIndex = i;
     }
-  }
-
-  if (jsonPreprocessor) {
-    parser.json = jsonPreprocessor(parser.json);
   }
 
   const gltf = await new Promise(parser.parse.bind(parser));
--- a/src/components/gltf-model-plus.js
+++ b/src/components/gltf-model-plus.js
@@ -250,7 +250,7 @@ function runMigration(version, json) {
   }
 }
 
-export async function loadGLTF(src, contentType, preferredTechnique, onProgress) {
+export async function loadGLTF(src, contentType, preferredTechnique, onProgress, jsonPreprocessor) {
   let gltfUrl = src;
   let fileMap;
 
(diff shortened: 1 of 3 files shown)


src/components/player-info.js
```diff
@@ -1,5 +1,28 @@
...
+  const { nodes } = json;
+  if (!nodes.some(node => node.name === "Head")) {
+    // If the avatar model doesn't have a Head node. The user has probably chosen a custom GLB.
+    // So, we need to construct a suitable hierarchy for avatar functionality to work.
+    // We re-parent the original root node to the Head node and set the scene root to a new AvatarRoot.
+    nodes.push({ name: "LeftEye", extensions: { MOZ_hubs_components: {} } });
+    nodes.push({ name: "RightEye", extensions: { MOZ_hubs_components: {} } });
+    nodes.push({
+      name: "Head",
+      children: [0, nodes.length - 1, nodes.length - 2],
```
netpro2k: I don't think we should assume that 0 is the root of the scene. We should use the `json.scenes[json.scene]` nodes if present.

src/components/player-info.js
```diff
@@ -1,5 +1,28 @@
...
+      extensions: { MOZ_hubs_components: { "scale-audio-feedback": "" } }
+    });
+    nodes.push({ name: "Neck", children: [nodes.length - 1] });
+    nodes.push({ name: "Spine", children: [nodes.length - 1] });
+    nodes.push({ name: "Hips", children: [nodes.length - 1] });
+    nodes.push({ name: "AvatarRoot", children: [nodes.length - 1] });
+    json.scenes[0].nodes[0] = nodes.length - 1;
```
netpro2k: We should use the `json.scene`th scene if its defined, otherwise default to 0th.

src/components/gltf-model-plus.js
```diff
@@ -305,6 +305,10 @@ export async function loadGLTF(src, contentType, preferredTechnique, onProgress)
     }
   }
 
+  if (jsonPreprocessor) {
+    parser.json = jsonPreprocessor(parser.json);
```
netpro2k: Don't think it matters too much for our current set of preprocessing but it probably makes sense to do this before our other modifications.

src/components/ik-controller.js
```diff
@@ -245,8 +245,8 @@ AFRAME.registerComponent("ik-controller", {
       this.ikRoot.el.components["player-info"].data.avatarType === AVATAR_TYPES.LEGACY
         ? LEGACY_HAND_ROTATIONS
         : HAND_ROTATIONS;
-    this.updateHand(handRotations.left, leftHand, leftController.object3D, true, this.isInView);
-    this.updateHand(handRotations.right, rightHand, rightController.object3D, false, this.isInView);
+    if (leftHand) this.updateHand(handRotations.left, leftHand, leftController.object3D, true, this.isInView);
```
netpro2k: Hmm forgot that a custom avatar means they aren't going to get hands. Wonder if we should create some default hands, at least locally.

brianpeiris: Oh, yeah, good point. Not sure if I want to tackle it in this PR though.

PR #1430, opened 2019-06-07
Only reset your avatar on failure

brianpeiris:


brianpeiris:
thanks for the quick review

PR #1433, opened 2019-06-10
Upgrade three and restructure three patches

robertlong:
- Upgrade ThreeJS to the current v105 dev branch -> [PR Here](https://github.com/MozillaReality/three.js/pull/33)
  - Adds updated fixes for GLTFLoader
  - Adds back in fix for multiscattering: https://github.com/mrdoob/three.js/pull/16559
  - Spoke and Hubs will be on the same branch again
- Reorganized patches to Three.js by moving them into `aframe-entry` and using `require()` for deterministic patching
- Removes `threejs-allocation-patches` (@gfodor said this may not actually be necessary and we can add it back in if we think otherwise)

The plan is to merge this branch and ensure stability before merging in the ThreeJS [UBO](https://github.com/mrdoob/three.js/pull/15562) and [VAO](https://github.com/mrdoob/three.js/pull/16287) PRs. We will build the batching support on top of ThreeJS with those PRs.

To Do:
- [ ] Merge https://github.com/MozillaReality/three.js/pull/33
- [ ] Update `three` dependency in package.json


robertlong:
Here's a picture of the bundle-analyzer output. It doesn't look like we are including three multiple times.

Screen Shot 2019-06-10 at 11 27 02 AM

PR #1440, opened 2019-06-12
Surface Drawing

InfiniteLee:
Enable drawing on surfaces with the pen. This will be enabled by default when in 2D mode (desktop or mobile) or holding a pen with the cursor in VR. It is toggle-able when holding the pen in a hand in 6-dof enabled devices (mapped to the "undo" button on the hand holding the pen. Existing undo functionality is now mapped to the "other" hand).

7e373a322398c4f36f9988fd1238cd68

Future TODO:
- [ ] Network the drawing "laser" to indicate a user is in this drawing mode.


src/systems/interactions.js
```diff
@@ -176,7 +176,13 @@ AFRAME.registerSystem("interaction", {
...
       !this.rightHandTeleporter.isTeleporting &&
       !this.gazeTeleporter.isTeleporting;
-    this.cursorController.components["cursor-controller"].enabled = enableRightRemote;
+    const enableCursor =
+      !this.state.rightRemote.held ||
+      !this.state.rightRemote.held.components ||
+      !this.state.rightRemote.held.components.tags ||
+      !this.state.rightRemote.held.components.tags.data.isPen ||
+      this.state.rightRemote.held.children[0].components.pen.cursorVisible;
+    this.cursorController.components["cursor-controller"].enabled = enableRightRemote && enableCursor;
```
johnshaughnessy: I think it would be good to move the logic that determines `cursorVisible` into this line, so the whole predicate is made up of "the state of the world" at one point in time, rather than half of the predicate state being set when the pen component runs and the other half being determined here. Just seems

src/components/tools/pen.js
```diff
@@ -20,19 +22,30 @@ const pathsMap = {
...
     penNextColor: paths.actions.leftHand.penNextColor,
     penPrevColor: paths.actions.leftHand.penPrevColor,
     scalePenTip: paths.actions.leftHand.scalePenTip
   },
   cursor: {
+    pose: paths.actions.cursor.pose,
     startDrawing: paths.actions.cursor.startDrawing,
     stopDrawing: paths.actions.cursor.stopDrawing,
     undoDrawing: paths.actions.cursor.undoDrawing,
+    // no way to switchDrawModes with mouse cursor yet
```
johnshaughnessy: Oops, I left this comment in but we probably do not want it.

PR #1451, opened 2019-06-14
Sets the camera model's environment map, fixes #1450

### LINKED ISSUE #1450: Camera model currently not getting environment map
Screen Shot 2019-06-12 at 9 15 35 PM

Started working to fix but unsure — 

is this something that should happen in 

https://github.com/mozilla/hubs/blob/a7b597ed5f61b86c74084e089d0fe1b5cca9ae92/src/systems/camera-tools.js

or is it part of the gltf extension work youre doing / the model should get tagged with an environment-map component ?

chrisnovello:
Screen Shot 2019-06-14 at 4 14 01 PM


gfodor:
grazi! does this look good @brianpeiris ?

PR #1453, opened 2019-06-20
Pull request to add Chrome extension for screen sharing

srigaurav1986:
Hi Sir/ Madam,

I have created one folder for extension and in hubs/scripts directory added two files
- getScreenId.js
- getScreenId.html

These two files talk to extension.

- In getScreenId.html we need to provide extension id which is received from chrome web store only
- In getScreenId.js we need to provide getScreenId.html

I have also created file sample-file-for-getScreedId-usage.js to show usage of function getScreenId() before calling webrtc’s  getUserMedia() api

Extension folder can be directly uploaded after zipping it to Chrome Web Store
Requesting to revert in case of any concern.


johnshaughnessy:
This seems really useful because it adds support for desktop streaming to chrome. I'm not familiar with chrome extensions or its web store so I haven't made much effort to push this along.

scripts/sample-file-for-getScreedId-usage.js
```diff
@@ -0,0 +1,187 @@
...
+			  return stPSD;
+ 			}
+                    getScreenId(function(error, sourceId, screen_constraints) {
+                        // error    == null || 'permission-denied' || 'not-installed' || 'installed-disabled' || 'not-chrome'
+                        // sourceId == null || 'string' || 'firefox'
+                        // getUserMedia(screen_constraints, onSuccess, onFailure);
+
+                        document.getElementById('capture-screen').disabled = false;
+
+                        if (IsAndroidChrome) {
```
johnshaughnessy: Which platforms' builds of chrome support extensions (and screen sharing)? Do we need to do different things for windows, mac, linux, android and iOS?

srigaurav1986: Windows,mac linux has no problem "only" for Android chrome i made exception. 

Also this is just a sample file to explain usage of getScreenId() in application where we are receiving the screenid from chrome extension  and passing it in getUserMedia API as constraints. This should be replaced in HUBs code where we are directly calling getUserMedia for screensharing.

srigaurav1986:
Regarding publishing in Chrome store we can do it as mentioned below :
https://developer.chrome.com/webstore/publish
which is quite easy.One has to login to chrome dashboard and upload zip.
https://chrome.google.com/webstore/developer/dashboard?pli=1

In manifest.json I already have provided Hubs domain if we change domain we need to just update it.

gfodor:
screen sharing is now enabled natively in chrome -- thanks for the PR though!

PR #1456, opened 2019-06-24
Camera Mode

johnshaughnessy:
Room owners can now activate Camera Mode from the menu. Camera Mode is designed for use when you are running an event and you'd like to stream, or would like to broadcast a specific view to spectators in the lobby. When in Camera Mode, you can fly around, hide the UI, and your vantage point will be streamed to viewers in the lobby.

image

This PR introduces streamer mode. When a user enters streamer mode:
 - The favorite button, settings menu, share button, and feedback button are hidden.
 - The scene-preview-camera of lobby participants is synced to the streamer's camera location.

TODO:
 - "Currently Spectating" should be tied to react state so that it update automatically when the streamer toggles streamer mode.
 - Make the tip reactive to different screen sizes so it doesn't conflict with the 2D Hud. The "Currently Spectating" info has the same problem with the enter-room modal.
- Add a red dot in the nametag to indicate when someone is the streamer.
- Change the icon in the presence list to indicate when someone is the streamer.
- Change the "exit streamer mode" icon so that it looks more like exiting streamer mode (e.g. a video camera with an x thru it.)


src/message-dispatch.js
```diff
@@ -37,7 +37,7 @@ export default class MessageDispatch {
     }
 
     const playerRig = document.querySelector("#player-rig");
-    const scales = [0.0625, 0.125, 0.25, 0.5, 1.0, 1.5, 3, 5, 7.5, 12.5];
+    const scales = [0.03125, 0.0625, 0.125, 0.25, 0.5, 1.0, 1.5, 3, 5, 7.5, 12.5, 25];
```
gfodor: 🐉

gfodor:
Streamer mode needs to be soft-blocked client side from being entered by non-owners. For now, I'd just check for `kick` permission.

src/react-components/settings-menu.js
```diff
@@ -213,6 +215,27 @@ export default class SettingsMenu extends Component {
                 </a>
               </div>
             </div>
+            {showRoomSettings ? (
```
gfodor: I'd probably stick this under a new section in the menu `TOOLS`

src/react-components/ui-root.js
```diff
@@ -737,6 +743,16 @@ class UIRoot extends Component {
     });
   };
 
+  toggleStreaming = () => {
```
gfodor: should this be `enableStreaming`?

src/react-components/ui-root.js
```diff
@@ -737,6 +743,16 @@ class UIRoot extends Component {
...
+    takeOrCreateLobbyCamera();
+    this.setState({ showSettingsMenu: false, hide: true });
+
+    const playerRig = document.querySelector("#player-rig");
+    if (playerRig.getAttribute("character-controller").fly !== true) {
```
gfodor: this seems weird, don't we typically do `.components["character-controller"]`?

src/assets/translations.data.json
```diff
@@ -246,7 +246,8 @@
     "settings.close-room": "Close Room...",
     "settings.room-info": "Room & Scene Info",
     "settings.create-room": "New Room",
-    "settings.streamer-mode": "Streamer Mode",
+    "settings.enable-streamer-mode": "Enable Streamer Mode",
```
gfodor: suggest `Streaming Mode` and `Exit Streaming Mode`

src/components/scene-preview-camera.js
```diff
@@ -1,14 +1,34 @@
...
+  const inv = new THREE.Matrix4();
+  return function setMatrixWorld(object3D, m) {
+    object3D.matrixWorld.copy(m);
+    object3D.matrix = inv.getInverse(object3D.parent.matrixWorld).multiply(object3D.matrixWorld);
+    object3D.matrix.decompose(object3D.position, object3D.quaternion, object3D.scale);
+  };
+})();
+
+function getStreamerCamera() {
+  const playerInfos = document.querySelectorAll("[player-info]");
```
gfodor: any way to avoid this happening every frame?

src/components/scene-preview-camera.js
```diff
@@ -1,14 +1,34 @@
-/**
- * Nicely pans the camera for previewing a scene. There's some weirdness with this right now
- * since it ends up panning in a direction dependent upon the start camera orientation,
- * but it's good enough for now.
- */
+const setMatrixWorld = (function() {
```
gfodor: this probably should go into utils somewhere

src/react-components/settings-menu.js
```diff
@@ -215,22 +217,33 @@ export default class SettingsMenu extends Component {
                 </a>
               </div>
             </div>
-            {showRoomSettings ? (
+            {this.props.scene.is("entered") && showRoomSettings ? (
```
gfodor: you'll want to check `this.props.hubChannel.can("kick_users")` to soft-prevent non-mods from going into streaming mode

src/react-components/settings-menu.js
```diff
@@ -215,22 +217,33 @@ export default class SettingsMenu extends Component {
...
                 <div className={styles.listItem}>
-                  <div
-                    className={styles.listItemLink}
-                    onClick={() => {
-                      this.props.toggleStreaming();
-                    }}
-                  >
-                    <FormattedMessage id="settings.streamer-mode" />
-                  </div>
+                  {document.querySelector("#player-rig").components["player-info"].data.isStreaming ? (
```
gfodor: copying from other PR, this should go into a `TOOLS` section after `PROFILE` and `ROOM`

src/react-components/ui-root.js
```diff
@@ -743,14 +739,21 @@ class UIRoot extends Component {
...
     }
+    playerRig.setAttribute("player-info", "isStreaming", true);
+  };
+
+  disableStreamerMode = () => {
+    const playerRig = document.querySelector("#player-rig");
+    if (playerRig.components["character-controller"].data.fly) {
+      playerRig.setAttribute("character-controller", "fly", false);
+    }
+    playerRig.setAttribute("player-info", "isStreaming", false);
```
gfodor: nit: this could be DRYer

src/systems/lobby-camera-system.js
```diff
@@ -1,79 +1,8 @@
...
-}
-
 export class LobbyCameraSystem {
   tick() {
-    const state = document.getElementById(NETWORKED_LOBBY_CAMERA_STATE);
-    if (AFRAME.scenes[0].is("entered") || !state || !state.parentNode || NAF.utils.isMine(state)) {
-      return;
+    const el = document.querySelector("[scene-preview-camera]");
+    if (el) {
+      el.components["scene-preview-camera"].tick2();
```
gfodor: not a strong opinion here, but it does seem a bit weird to refactor this into this ECS pattern when there is just a single instance

src/react-components/ui-root.js
```diff
@@ -1643,7 +1663,8 @@ class UIRoot extends Component {
                 >
                   {!embed &&
                     !showVREntryButton &&
-                    !hasTopTip && (
+                    !hasTopTip &&
+                    !this.state.isStreaming && (
```
gfodor: since the logic here has gotten haywire, i've been putting these in vars, i'd suggest `streaming`

src/react-components/settings-menu.js
```diff
@@ -35,6 +39,7 @@ export default class SettingsMenu extends Component {
     const showCloseRoom = !!this.props.hubChannel.canOrWillIfCreator("close_hub");
     const showRoomInfo = !!this.props.hubScene;
     const showRoomSection = showRoomSettings || showRoomInfo || showCloseRoom;
+    const showStreamerMode = this.props.scene.is("entered") && !!this.props.hubChannel.canOrWillIfCreator("kick_users");
```
gfodor: I suppose `canOrWillIfCreator` is probably OK vs `can`

gfodor:
looking good! this seems a bit easier to reason about, thanks for changing it up

gfodor:
Do you think we should include an informational label in the lobby noting who you are viewing through? I was thinking only on wider screens, in the lower left corner saying "Watching via Jomb" or something

gfodor:
Also as part of this I think we should move the Hide UI button from lobby viewing mode into this mode.

src/systems/lobby-camera-system.js
johnshaughnessy: I didn't put this code here because I care about the ECS pattern. I only care that the tick happens after the `networked` component `tick`, which lerps the remote avatar's head transform. The same can be said of some of the other systems in `hubs-systems`. Some of them were written with an ECS pattern in mind but a lot of it is just about wanting a defined execution order.

johnshaughnessy: Open to suggestions for how else to accomplish the desired tick order. One way is to just change `scene-preview-camera` into an aframe system instead of this class with a `tick` method.

gfodor: ah that makes sense, thanks for clarifying -- seems OK to leave as-is.

PR #1497, opened 2019-07-04
Detect transparent pngs

robertlong:
Use the png header and transparency chunk to check whether a png is in fact transparent. Set the pixel format to `RGB` `RGBA` accordingly.


src/components/media-views.js
```diff
@@ -203,23 +204,22 @@ function scaleToAspectRatio(el, ratio) {
...
-        reject(`'${url}' could not be fetched (Error code: ${xhr.status}; Response: ${xhr.statusText})`);
-      }
-    );
-  });
+async function createImageTexture(url, contentType) {
+  const texture = new THREE.Texture();
+
+  await textureLoader.loadTextureAsync(texture, url, contentType);
+
+  // try {
```
robertlong: Uncomment before merging

src/loaders/HubsTextureLoader.js
```diff
@@ -0,0 +1,132 @@
+/**
+ * @author mrdoob / http://mrdoob.com/
+ */
```
mqp: ?

src/loaders/HubsTextureLoader.js
```diff
@@ -0,0 +1,132 @@
...
+  return url.search(/\.png($|\?)/i) > 0 || url.search(/^data:image\/png/) === 0;
+}
+
+function getChunkType(arrayBuffer, byteOffset) {
+  const arr = new Uint8Array(arrayBuffer, byteOffset, 4);
+  let str = "";
+  for (let i = 0; i < 4; i++) {
+    str += String.fromCharCode(arr[i]);
+  }
+  return str;
```
mqp: Seems like this can be `return String.fromCharCode.apply(null, arr)`.

src/loaders/HubsTextureLoader.js
```diff
@@ -0,0 +1,132 @@
...
+  return new Promise((resolve, reject) => loader.load(url, resolve, onProgress, reject));
+}
+
+function isJPEG(url) {
+  return url.search(/\.jpe?g($|\?)/i) > 0 || url.search(/^data:image\/jpeg/) === 0;
+}
+
+function isPNG(url) {
+  return url.search(/\.png($|\?)/i) > 0 || url.search(/^data:image\/png/) === 0;
+}
```
mqp: It would be nice if this used `guessContentType` from media-url-utils, into which it would be totally reasonable to put the data-URL logic.

src/loaders/HubsTextureLoader.js
robertlong: We should probably just add the threejs MIT header instead? HubsTextureLoader is a modified THREE.TextureLoader.

Issue #1490, opened 2019-06-26
[UX Enhancement] Enter Code: No Cancel or X Button


flowpoly:
As a user, when I am adding a room's code, there is no way for me to press cancel to go back to the previous page. This also isn't a modal experience as most of the similar experiences throughout the system.

![EnterCode](https://user-images.githubusercontent.com/38146102/60178459-9a18c700-97e9-11e9-96a3-7411a7cf3e07.jpg)

gfodor:
For this one, what was the previous page you're referring to? That page typically should be navigated to directly by typing hub.link into the browser -- and to go back obviously you'd use the browser back button. Is there a flow where we direct you there some other way? (I think maybe the invite dialog is the one?)

flowpoly:
@gfodor Yeah, I see what you mean and that makes sense, but would it be helpful to make the "Hubs by Mozilla Logo" a link to Hubs' main page?

gfodor:
ah yup I can do that

PR #1504, opened 2019-07-04
Linkify logo

gfodor:
https://github.com/mozilla/hubs/issues/1490


gfodor:
I don't think we want that logo to be a link -- in this case its a way to get somewhere useful if you landed on the wrong page.

PR #1507, opened 2019-07-05
Don't do a ton of stuff whenever local storage changes

mqp:
Sending the new presence profile caused the code right below responding to presence updates to also run, which caused everyone in the room to, e.g., do a full React re-render because you clicked some button that updated its bit to say that you didn't need to see a tutorial tooltip anymore.

It's still really fishy that we re-render the whole React UI (multiple times, actually) whenever anything in presence changes, but this starts to minimize the damage.


gfodor:
cc @netpro2k

PR #1508, opened 2019-07-06
Settings menu is responsible for its own expanded state

mqp:
This pattern is more efficient with regard to re-rendering and moves some conceptually settings-menu-related stuff (its non-expanded look and behavior) outside of the root component, which feels nice.


gfodor:
lgtm!

PR #1509, opened 2019-07-06
Presence list is responsible for its own expanded state

mqp:
Same as https://github.com/mozilla/hubs/pull/1508. Rendering efficiency, cleanliness.

We really need to fix up the CSS on these -- right now the styles and markup are kind of redundant and weird. (For example, the presence list and presence list button are positioned totally independently of each other.)


gfodor:
lgtm! you'll have a small obvious merge conflict between this and the other branch fyi

PR #1516, opened 2019-07-08
Turn 'kicked' message into 'disconnected' message

mqp:
As we talked about a while ago, this is probably a better default assumption to present to the user. Examples of times I definitely get this message:

- If I leave the debugger paused for a little bit
- If a tab gets backgrounded too hard by the browser (seems especially common on mobile, but sometimes happens on desktop)
- If my wireless connection dies for a little bit


netpro2k:
I vote we also get rid of the "You can also [create a new room]" part of this screen since that's almost never the right thing to do to resolve a connection issue and just adds confusion. (@johnshaughnessy saw this happen when using Hubs with a friend)

mqp:
I'm not very opinionated about that, but you can get rid of it if you want.

PR #1517, opened 2019-07-09
Hub object permissions

brianpeiris:
You can now set permissions on a room that prevent visitors from creating objects, using the camera and pen, or pinning objects.

firefox_2019-07-08_17-16-50

Adds UI for setting permissions and enforces those permissions on actions and UI elements.

image

image
(note the missing pin button)

Goes with:
- https://github.com/mozilla/reticulum/pull/201
- https://github.com/MozillaReality/networked-aframe/pull/32


gfodor:
Nice work getting this all together! Will do a full integration test today but took a quick look. Two things off the top of my head:

- With this we need to no longer respect any data channel traffic. I believe it's as simple as just filtering out any messages not coming from the phx source

- Where did you land on the policy/semantics around pager/video controls? The use case I have in mind is if someone puts a video element or a PDF element into a spoke scene (or, alternatively, has one that is pinned in the room) it seems likely we want to support the use case where visitors can come in and control that media but not move, unpin, or remove it.

gfodor:
Also, I think we probably need to set up a real stage of the smoke test that directs people into a "fully denied" room, and asks you to attempt a few actions. It'll add a good 30-45 seconds to the test, but I think it's worth it to ensure we don't have any security regressions.

gfodor:
Notice what seems to be a bug:

- Sign in, create a room, enter room
- Sign out
- Main HUD actions etc are disabled unexpectedly
- Refreshing resolves

gfodor:
We should disable moving of pinned objects in pause mode if you do not have object pinning permission

gfodor:
If I deny camera permissions, it looks like a non-privileged user can cause a few bad things to happen:
- They can remove cameras, which seems wrong I think
- They can grab and move cameras. Which might be OK, but doing so seems to break things -- the other client does not see them move the camera, and also from then forward vice versa.

gfodor:
Nice to have if it's easy:

- If a user has created a camera/pen/media and then the relevant permission is denied, it'd be nice to have client side logic just remove all the relevant spawned objects since likely the intent of the room owner is to deny these things retroactively

gfodor:
If pen permissions are denied, I still seem to be able to grab another person's pen. If I draw with it, it works locally (but not remotely) but the remote client crashes with some errors:

image

gfodor:
We should also get rid of remove button on relevant objects if not already

gfodor:
It looks like even if media spawning permission is denied, I am able to spawn a screenshot off of a video. This points to some server-level security hole in addition to us not hiding the button.

gfodor:
jfyi, it looks like non-privileged users can always control playback controls on videos. i think that's probably OK for now but I bet we want to add more fine grained permissions for that at some point

gfodor:
It looks like if I deny and then grant drawing permission (or maybe just in general?) then drawings are no longer appearing across clients from a non-privileged user. The receiving client gets a bunch of errors (similar to the ones posted above)

gfodor:
It looks like the video screenshot thing actually isn't a server bug, just a client one. the photo spawns locally but does not network

misslivirose:
Quick question - it looks like in some cases, when permissions are denied to a user, the relevant UI is in a grayed-out state, as opposed to others where the UI is simply not present. Would it make more sense to have all of those elements use the same behavior?

gfodor:
I think the greyed out options are just the HUD (correct me if I'm wrong
brian) -- the reason for that vs removal is the HUD doesn't provide a
visual design that'd gracefully degrade if you removed arbitrary buttons.
(Eg if the pen is removed, there's just a weird gap in the horizontal
layout)

brianpeiris:
Yup, that was my reasoning. Happy to do a design pass once this is out. This is more about getting the mechanisms in place with a decent UX.

misslivirose:
Totally makes sense, I wouldn't view it as blocking at all. 👍  really looking forward to these.

src/components/hide-when-pinned-and-forbidden.js
```diff
@@ -0,0 +1,47 @@
...
+  _updateUIOnStateChange(e) {
+    if (e.detail !== "frozen") return;
+    this._updateUI();
+  },
+
+  _updateUI() {
+    if (!this.targetEl) return;
+    const isPinned = this.targetEl.components.pinnable && this.targetEl.components.pinnable.data.pinned;
+
+    if (this.data.whenPinned) {
```
gfodor: i think we've converged on doing `this.el.object3D.visible` to avoid attribute accessor overhead

src/hub.html
```diff
@@ -302,15 +313,27 @@
...
+                            <a-entity
+                                mixin="rounded-button"
+                                is-remote-hover-target
+                                tags="singleActionButton: true;"
+                                hide-when-pinned-and-forbidden="whenPinned: true"
+                                remove-networked-object-button
+                                position="0 -0.375 0.01">
                                 <a-entity
                                     sprite="name: remote-action.png"
                                     icon-button="image: remote-action.png; hoverImage: remote-action.png"
```
gfodor: huh, `s/remote/remove/` here, not this PR but probably my typo if you feel like fixing :P

src/components/in-world-hud.js
```diff
@@ -31,15 +36,18 @@ AFRAME.registerComponent("in-world-hud", {
     };
 
     this.onSpawnClick = () => {
+      if (!window.APP.hubChannel.can("spawn_and_move_media")) return;
```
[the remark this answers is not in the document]
brianpeiris: Yes, this is necessary, and it's not hotkey related. `icon-button` does nothing except visuals. Perhaps an improvement for another day.

PR #1520, opened 2019-07-09
Feature/feedback dialog

gfodor:
Updates the feedback UX for collecting user surveys:

- Puts the new survey in a modal dialog
- Adds a tip asking for the user to complete the survey after their third session

Also extracts some more stuff out of UI root into a separate component, the lower tip.


src/assets/translations.data.json
```diff
@@ -323,7 +327,7 @@
     "change-scene-dialog.new-spoke-project": "Launch Spoke",
     "avatar-url-dialog.apply": "Apply",
     "interstitial.prompt": "Continue",
-    "feedback.prompt": "Feedback",
+    "feedback.prompt": "🦆 Feedback ",
```
johnshaughnessy: Wonder if the duck from the favicon would look better than an emoji here

src/react-components/ui-root.js
```diff
@@ -1646,49 +1655,15 @@ class UIRoot extends Component {
...
-                  </button>
-                  <div className={styles.tip}>
-                    {hasDiscordBridges ? (
-                      <span>{`Chat in this room is being bridged to ${discordSnippet} on Discord.`}</span>
-                    ) : (
-                      <FormattedMessage id="embed.presence-warning" />
-                    )}
-                  </div>
-                </div>
+                <Tip
```
johnshaughnessy: I'm glad you pulled this stuff out into its own component

src/assets/translations.data.json
gfodor: its intentional, the feedback duck is a mallard :duck: :)

PR #1522, opened 2019-07-10
Update three to vao-ubo-prs

netpro2k:
This updates package.lock to point at our latest three.js fork that includes vao+ubo patches. Also fixes an issue introduced yesterday for relative url textures.

### DIFF
--- a/src/utils/media-url-utils.js
+++ b/src/utils/media-url-utils.js
@@ -99,8 +99,7 @@ export const guessContentType = url => {
       matches[1];
     }
   }
-  const path = url.startsWith("/") ? url : new URL(url).pathname;
-  const extension = path.split(".").pop();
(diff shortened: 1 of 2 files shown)


src/utils/media-url-utils.js
```diff
@@ -99,7 +99,8 @@ export const guessContentType = url => {
       matches[1];
     }
   }
-  const extension = new URL(url).pathname.split(".").pop();
+  const path = url.startsWith("/") ? url : new URL(url).pathname;
```
robertlong: `new URL(url, window.location).pathname` will work as well

robertlong:
I added one suggestion, otherwise lgtm

src/utils/media-url-utils.js
netpro2k: Ah yeah, I like that better.

PR #1524, opened 2019-07-10
Fix content type check in HubsTextureLoader

robertlong:
This fix uses the file header instead of the inferred content type from farspark or the url extension. This allows us to better check for file types of blob urls, urls without extensions, and resolved urls with incorrectly inferred content types.

### DIFF
--- a/src/loaders/HubsTextureLoader.js
+++ b/src/loaders/HubsTextureLoader.js
@@ -5,7 +5,31 @@
 // https://github.com/mrdoob/three.js/blob/master/LICENSE
 //
 
-import { guessContentType } from "../utils/media-url-utils";
+// http://www.libpng.org/pub/png/spec/1.2/PNG-Structure.html
+const PNGFileSignature = new Uint8Array([137, 80, 78, 71, 13, 10, 26, 10]);
+
+// https://en.wikipedia.org/wiki/JPEG_File_Interchange_Format#File_format_structure
+// SOI Chunk Header: FF D8
+// JFIF / EXIF Chunk Header: FF ?? (either E0 or E1)
+const JPEGFileSignature = new Uint8Array([255, 216, 255]);
+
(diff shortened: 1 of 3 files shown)


src/loaders/HubsTextureLoader.js
```diff
@@ -76,13 +96,15 @@ export default class HubsTextureLoader {
...
-      url = URL.createObjectURL(blob);
-    } else if (finalContentType === "image/jpeg") {
+    } else if (signaturesEqual(JPEGFileSignature, bytes)) {
       // JPEGs can't have an alpha channel, so memory can be saved by storing them as RGB.
       transparent = false;
+    } else {
+      console.warn(`Couldn't detect image type for: "${src}". Using RGBA pixel format.`);
     }
 
+    url = URL.createObjectURL(blob);
```
netpro2k: I think we only need to bother with this in the case we have to fallback to Image. We already have the data here we need to create an `ImageBitmap`, so I don't think we need to go through the trouble of creating, fetching, and revoking this object url only to convert it back to an TypedArray to then create an ImageBitmap.

robertlong: Unfortunately there isn't a good way of doing that 😞 at least not that I can find. So I guess we'll stick with this for now.

src/loaders/HubsTextureLoader.js
```diff
@@ -26,42 +50,38 @@ export default class HubsTextureLoader {
...
     let url = src;
     let transparent = true;
 
-    const finalContentType = contentType || guessContentType(src);
-
-    if (finalContentType === "image/png") {
-      const fileLoader = new THREE.FileLoader(this.manager);
-      fileLoader.setResponseType("blob");
+    const fileLoader = new THREE.FileLoader(this.manager);
+    fileLoader.setResponseType("blob");
```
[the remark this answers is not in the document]
robertlong: See below

PR #1536, opened 2019-07-16
Feature/glowing mic

johnshaughnessy:
These changes add a visual effect to the mic icons that reacts proportionally to the volume coming in from the user's microphone.


src/assets/stylesheets/2d-hud.scss
```diff
@@ -393,3 +367,204 @@
     background-image: url(../hud/share_window_on.png);
   }
 }
+
+// Discrete frames of the mic "animation"
+:local(.mic-0) {
```
johnshaughnessy: Can remove these now that the css spritesheet is working without the "bouncing icon" issue

src/react-components/mic-level-widget.js
```diff
@@ -3,80 +3,55 @@ import PropTypes from "prop-types";
...
     const micLevelMovingAverage = MovingAverage(100);
+    let max = 0;
     this.micUpdateInterval = setInterval(() => {
-      analyser.getByteTimeDomainData(levels);
-      let v = 0;
-      for (let x = 0; x < levels.length; x++) {
-        v = Math.max(levels[x] - 128, v);
-      }
-      const level = v / 128.0;
-      // Multiplier to increase visual indicator.
```
johnshaughnessy: I made it so that the volume indicators glow in proportion to the loudest sound you've made in the microphone so far. This max should probably be a "max in the last minute or so" so that if you make a very loud noise that's well above your speaking voice, then your glow bars will recover after a minute. It's fine as-is for now and fills up well for non-degenerate cases like described above.

src/react-components/ui-root.js
```diff
@@ -595,6 +595,9 @@ class UIRoot extends Component {
       if (micDeviceId) {
         this.props.store.update({ settings: { lastUsedMicDeviceId: micDeviceId } });
       }
+      const mediaStreamForMicAnalysis = new MediaStream();
+      mediaStreamForMicAnalysis.addTrack(this.state.audioTrack.clone());
+      this.props.scene.emit("local-media-stream-created", { mediaStream: mediaStreamForMicAnalysis });
```
johnshaughnessy: I didn't think of a good way to "future proof" accidental mis-use of this mediaStream. When toggling your microphone, the janus adapter makes sure to disable the audio streams it knows about, so that's good. In general no one should be piping audio from this mediaStream into anything that is sent over the networked or saved to a file (like videos from the camera tool).

src/react-components/mic-level-widget.js
johnshaughnessy: Changed micLevelsForVolume such that the icon sees a wide range when talking.

PR #1540, opened 2019-07-18
Improve concurrent media resolving

netpro2k:
Right now if multiple concurrent requests are made for the same piece of media (for example you are joining a room where someone has cloned an image multiple times) your client may make multiple requests to reticulum, make multiple requests to get the image, and create multiple THREE.Textures (incorrectly messing up the retain count in the texture cache in the process). This is made worse by the fact that image loading is currently more expensive due to trying to detect the datatype of the image.

This PR fixes these issues by:
1. Removing image type detection
It turns out that modern GPUs like store things in RGBA format anyway, so we aren't actually getting any memory savings, and actually taking a performance hit for our trouble. This does mean that we no longer know if an image may be transparent. For now I have just made all images be considered opaque except for gifs. This means all non-gif images will be batched. To help with the common case of images with cutout alpha (like our chat bubbles), I have added support to the batch shader for this. We lose anti-aliasing of the chat messages, but gain batching of chat messages. We can expand on this later, likely doing transparency detection on the backend by actually looking through the alpha channel (also possible to do this on the client in a GPU accelerated way).
2. Preventing concurrent resolve media requests by updating resolveUrlCache to contain promises to the resolved json which are immediately cached at the start of the first request
3. Preventing concurrent image texture creation by adding an inflightTextures map to keep track of inflight requests.

We might be able to do something to DRY up 2 and 3 a bit, but it's so little code I am not really bothered by it. That said I would be interested to hear opinions on which technique people prefer.

### DIFF
--- a/src/components/media-views.js
+++ b/src/components/media-views.js
@@ -730,7 +730,7 @@ AFRAME.registerComponent("media-image", {
     }
     if (this._hasRetainedTexture) {
       textureCache.release(this.data.src);
-      this._hasRetainedTexture = false;
+      this.currentSrcIsRetained = false;
     }
   },
 
@@ -749,6 +749,7 @@ AFRAME.registerComponent("media-image", {
         this.mesh.material.needsUpdate = true;
         if (this.mesh.map !== errorTexture) {
           textureCache.release(oldData.src);
+          this.currentSrcIsRetained = false;
         }
       }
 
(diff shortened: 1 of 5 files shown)


src/components/media-views.js
```diff
@@ -750,31 +752,40 @@ AFRAME.registerComponent("media-image", {
...
+          let promise;
+          if (contentType.includes("image/gif")) {
+            promise = createGIFTexture(src);
+          } else if (contentType.startsWith("image/")) {
+            promise = createImageTexture(src);
+          } else {
+            throw new Error(`Unknown image content type: ${contentType}`);
+          }
+          inflightTextures.set(src, promise);
+          texture = await promise;
```
johnshaughnessy: This isn't a concern that is new with your changes, but if a `media-image`'s `entity` gets removed after `createImageTexture` is called, but before the `promise` resolves, then will `_hasRetainedTexture` be `false` when checked in `remove`? (If so, does than mean we'll retain and never release this texture?)

netpro2k: Good catch, added a fix

PR #1544, opened 2019-07-19
Fix some random bugs seen during 7/19 meetup Sentry

gfodor:
- Oculus Browser 6 no longer pauses videos when exiting VR mode, so we can remove an ugly hack that breaks videos easily
- It seems the favorites API can sometimes return an invalid response, handle this case properly on the homepage
- Deal with the case where we can't parse a URL that we'd like to thumbnail

### DIFF
--- a/src/hub.js
+++ b/src/hub.js
@@ -809,12 +809,12 @@ document.addEventListener("DOMContentLoaded", async () => {
 
     remountUI({ hide: false });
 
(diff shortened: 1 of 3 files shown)


gfodor:
Fixed the root cause of the favorites issue I think: https://github.com/mozilla/reticulum/commit/db6b88cfa07e54e2222dd3b9ae7f794cc89aac0b

favorited the smoke test, broke the homepage 🤦‍♂

PR #1547, opened 2019-07-19
Add particle-emitter component

robertlong:


src/components/particle-emitter.js
```diff
@@ -0,0 +1,147 @@
...
+        console.error
+      );
+    }
+
+    if (prevData.startColor !== data.startColor) {
+      particleEmitter.startColor.set(data.startColor);
+      this.updateParticles = true;
+    }
+
+    if (prevData.middleColor !== data.middleColor) {
```
johnshaughnessy: Unless there are perf concerns (that are worse than doing lots of if checks), the only `if` statements that seem necessary here are the ones that cause you to set `this.updateParticles = true` or load a texture from a new `src`. Perhaps replace the former with

```
if ( predicateA || predicateB || predicateC || predicateD...) {
  this.updateParticles = true;
}
particleEmitter.foo.copy(data.foo);
particleEmitter.bar = data.bar
(etc)
```

PR #1556, opened 2019-07-23
Apply stashed first sync on videos before playback and ownership

### LINKED ISSUE #1543: Scene owned videos have a race condition
If a scene is complex enough, if you join a room with a scene owned video and someone else is the time keeper, NAF will process the initial sync message from them (causing an instantiation) before the glTF parser has instantiated the video. The result is two videos.

here's what I think is the proper fix for the bug and also doesn't break the contract on the new server authorization wrt instantiations:
- the server skips authorization for NAF messages marked `isFirstSync: true, persistent: true`
- the server authorizes other `isFirstSync: true` messages
- the client, when it sees a isFirstSync message:
  - if it's `persistent: true`, stashes the message until the object is instantiated during loading, at which time it's applied
  - otherwise, instantiates as usual

instantiations occur iff both flags are true, so as long as those messages are authorized, the server is still authorizing all instantiations. this addresses the bug by ensuring we allow the glTF loader to initialize all scene owned objects before processing the incoming first sync message from the current owner. (and as usual, non first sync messages are dropped on the floor for objects not yet instantiated.) this does leave an additional race condition behind -- if during loading the owner makes some other kind of networked state change after the initial sync is sent, it may be dropped if the user has yet to instantiate the object. in practice, this will not happen for videos since the timekeeper continually flushes updated information.

brianpeiris:
Fixes #1543 by having networked-aframe stash first-syncs for persistent entities, instead of creating said entities. We then apply the stashed first-sync before deciding on playback state and ownership.

Goes with:
- https://github.com/MozillaReality/networked-aframe/pull/33
- https://github.com/mozilla/reticulum/pull/203


src/components/media-views.js
```diff
@@ -361,6 +361,7 @@ AFRAME.registerComponent("media-video", {
 
     NAF.utils.getNetworkedEntity(this.el).then(networkedEl => {
       this.networkedEl = networkedEl;
+      this.networkedEl.components.networked.applyPersistentFirstSync();
```
gfodor: as noted in the other PR, any reason we can't push this policy into NAF itself?

brianpeiris: We can't apply the stashed sync in `init` on `networked`, the rest of the entity's components are not initialized at that point. If we did apply it then, it would result in those components being created and then overwritten by the rest of Hub's scene and media initialization process.

brianpeiris: In particular, we have to apply the first sync after component inits, but before we do this very next bit, where we try to take ownership of scene-owned media.

gfodor: ah interesting. i guess i can't come up with a better idea. renaming the function to be less implementation-centric would be nice, since it's a public API, but i can't think of a much better name. eg `notifyPeristentEntityReady` or something

brianpeiris: Hmm, actually I think I would like to keep the name explicit since I want it to be clear to us when reading this code.

Issue #335, opened 2018-04-28
Non-working link with “Add to Home Screen” on iOS 


j-conrad:
In Safari (iOS 11), while in a hub, using the “Add To Home Screen” function creates a shortcut that resides with other apps as expected. However, it creates an (A-frame?) icon and subsequently opening that app appears to open a webview instead of opening up an actual bookmark to that hub, which means the hub space doesn't open.

Note: using “Add To Home Screen” on the hubs.mozilla.com landing page makes a usable app button.
It would be useful to be able to make a series of app icons to easily return to previously-created/visited hub spaces.

Platform: iOS 11 / Safari / iPhone 6s

gfodor:
this is (finally) getting looked at now via https://github.com/mozilla/hubs/pull/1380. sorry it took so long :(

misslivirose:
Since 1380 was merged, @j-conrad, can you verify that this is no longer an issue?

j-conrad:
This is still not working. Just tested on Safari with iOS 12.4.
Same result as described in the original issue— app icon gets created, but only for hubs.mozilla.com, not the actual room you created. Icon is also still the A-Frame logo.

misslivirose:
Thanks for verifying!

gfodor:
Having it add a link to home screen for hubs.mozilla.com (not the room) is
intended -- that's the landing page for the PWA. By introducing the PWA
manifest for the site, it's no longer possible to create home screen links
to specific rooms. (The web manifest dictates the launch URL.) We could
support this if we wanted by serving up a room-level manifest on the room
page, but I don't think that's desirable since the concept is that you are
adding the "hubs app" to your phone.

The icon is weird. It might just be cached on your phone, but we should try
on another device to confirm.

gfodor:
BTW, the flow to save rooms is intended to be favorite the room and launch
the PWA to get the list of favorites. You could imagine this is how it'd
work if you had a "Hubs app" you installed from the app store.

misslivirose:
Great, that makes sense. So the bug (in this case) would be if the A-Frame logo is still what shows up for iOS users when they save it to the home screen, if I'm understanding correctly. The rest is by design with the favorites feature and manifest.

gfodor:
Yup. and we should confirm that A-Frame icon on another device. I'm pretty
sure I tested it and the manifest is correct, so it might just be a cache
on Jim's phone from the last time he installed it.

j-conrad:
[...]

PR #1557, opened 2019-07-24
fix icon for iOS Safari's "Add to Homescreen" (PWA): add required `<link rel="apple-touch-icon">` tag (fixes issue #335)

cvan:
see discussion: https://github.com/mozilla/hubs/issues/335#issuecomment-514016281

## Context
- Apple shipped support in [Safari for iOS 11.1](https://developer.apple.com/library/archive/releasenotes/General/WhatsNewInSafari/Articles/Safari_11_1.html#//apple_ref/doc/uid/TP40014305-CH14-SW6) for parsing [web-app manifests](https://w3c.github.io/manifest/) via `<link rel="manifest">` but (for some reason) ignores the `icons` defined in the manifest. A `<meta name="apple-touch-icon">` tag appears to be still required for the time being. (I assume that'll change in an upcoming release.)
- This also replaces the icon of the scene pages, since A-Frame injects, if missing, [a few Apple-proprietary `<meta>` tags](https://github.com/aframevr/aframe/blob/c871d04e/src/core/scene/metaTags.js#L14-L16).

### DIFF
--- a/src/hub.html
+++ b/src/hub.html
@@ -12,6 +12,7 @@
     <link rel="manifest" href="/manifest.webmanifest">
     <!-- iOS (for "Add to Homescreen") -->
     <meta name="mobile-web-app-capable" content="yes">
+    <link rel="apple-touch-icon" sizes="180x180" href="https://assets-prod.reticulum.io/assets/images/pwaicon-180-hubs.png">
     <!-- Microsoft Edge (Windows) -->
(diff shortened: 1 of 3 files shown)


gfodor:
thanks for this! this had to get added to the reticulum side, so I did that here: https://github.com/mozilla/reticulum/pull/227/files

PR #1562, opened 2019-07-24
Fix disappearing pinned objects due to a doubled animation on first sync

brianpeiris:
Recent changes with NAF cause `media-loader`'s `onMediaLoaded` method to be called twice -- once when the object is loaded from the room's objects.gltf and again when NAF receives a firstSync from the pinned object's owner.
This PR fixes onMediaLoaded to be idempotent with regard to the scaling animation.


InfiniteLee:
lgtm- though I think we should also fix the root issue here which is that onMediaLoaded is getting called twice for the same object back to back.

brianpeiris:
Yup, I'll probably look into that further when I start working on more NAF update authorization stuff. Thanks

PR #1563, opened 2019-07-24
Fix flipY on 360 images

robertlong:
Fixes https://github.com/mozilla/Spoke/issues/693

I'm still not really sure why the conditional needs to be `if (!texture.flipY) {` but it does.


robertlong:
I'm fine with merging this, but I still think it's kind of a hack and doesn't fix the underlying issue. @netpro2k can we either move this fix to ThreeJS in some way that will also fix 360 videos, or modify the batch manager so that it handles the texture flipping on upload? That's the reasoning for the flipY parameter in PlaneBufferGeometry right? I'd rather not litter our code with this UV flipping code. If the PlaneBufferGeometry needs it and now the SphereBufferGeomerty, should we just move this to SphereBufferGeometry, or should we make a fix that doesn't involve modifying each BufferGeometry class?

PR #1567, opened 2019-07-25
Better Valve Index Support

InfiniteLee:
Adds better joystick support for Index controllers (movement, rotation, scrolling).
Also changes the grab-toggle behavior to only affect Vive wands.


src/systems/userinput/bindings/vive-user.js
```diff
@@ -315,21 +348,31 @@ export const viveUserBindings = addSetsToBindings({
...
     },
+    {
+      src: {
+        value: rDpadEast2
+      },
+      dest: {
+        value: rSnapRight2
+      },
+      xform: xforms.rising,
+      priority: 1
```
johnshaughnessy: I think this priority value is unnecessary

InfiniteLee: Ah yes, forgot about that one, thanks.

johnshaughnessy:
When me come across more things that are different between index / vive wands / vive focus plus then we ought to do what Dom suggested -- namely, create a new device for each of those, change `vrGamepadMappings` so that its values are arrays-of-objects-full-of-bindings instead of just objects-full-of-bindings.

johnshaughnessy:
Another thing I thought of after submitting that review is that the bindings that use the Index/Knuckles controller axes (or buttons) ought to be separate from the vive-user bindings, since the vive wands don't want/need those (and we'll have to separate them if ever we want those axes/buttons to do different things).

PR #1579, opened 2019-07-26
Load fewer images

johnshaughnessy:
Unless I'm misunderstanding something here, we don't need to load all these images.


netpro2k:
These are all now in the spritesheet yeah?

johnshaughnessy:
> These are all now in the spritesheet yeah?

yes

PR #1582, opened 2019-07-26
Fix avatar switching when recording videos

### LINKED ISSUE #1555: Changing your avatar 2x during video recording crashes
In smoke test, start video camera and then switch between the avatars pasted in the room. Error:
image

netpro2k:
Fixes #1555 
Also fixes error found in the process where we release the assets for the avatar you are switching from while we are still rendering it (resulting in some log spam as it attempts to re-upload an empty texture).
Lastly makes this actually testable on dev by fixing regexes to recognize in world avatar (and room/scene) links on dev.

### DIFF
--- a/src/utils/media-url-utils.js
+++ b/src/utils/media-url-utils.js
@@ -110,9 +110,9 @@ export const guessContentType = url => {
   const extension = new URL(url, window.location).pathname.split(".").pop();
   return commonKnownContentTypes[extension];
 };
-const hubsSceneRegex = /https?:\/\/(hubs.local(:\d+)?|(smoke-)?hubs.mozilla.com)\/scenes\/(\w+)\/?\S*/;
-const hubsAvatarRegex = /https?:\/\/(hubs.local(:\d+)?|(smoke-)?hubs.mozilla.com)\/avatars\/(\w+)\/?\S*/;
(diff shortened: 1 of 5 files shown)


src/utils/media-url-utils.js
```diff
@@ -110,9 +110,9 @@ export const guessContentType = url => {
   const extension = new URL(url, window.location).pathname.split(".").pop();
   return commonKnownContentTypes[extension];
 };
-const hubsSceneRegex = /https?:\/\/(hubs.local(:\d+)?|(smoke-)?hubs.mozilla.com)\/scenes\/(\w+)\/?\S*/;
-const hubsAvatarRegex = /https?:\/\/(hubs.local(:\d+)?|(smoke-)?hubs.mozilla.com)\/avatars\/(\w+)\/?\S*/;
-const hubsRoomRegex = /https?:\/\/(hubs.local(:\d+)?|(smoke-)?hubs.mozilla.com)\/(\w+)\/?\S*/;
+const hubsSceneRegex = /https?:\/\/(hubs.local(:\d+)?|(smoke-)?hubs.mozilla.com|(dev\.)?reticulum.io)\/scenes\/(\w+)\/?\S*/;
```
johnshaughnessy: I think Greg prefers that dev links don't get recognized in prod

PR #1587, opened 2019-07-29
Resolve images in particle-emitter component

robertlong:
I forgot to resolve / proxy urls in the particle-emitter component. This PR adds those. You can turn it off by setting `resolve: false`

### DIFF
--- a/src/components/particle-emitter.js
+++ b/src/components/particle-emitter.js
@@ -1,8 +1,11 @@
 import { ParticleEmitter } from "@mozillareality/three-particle-emitter";
 import { textureLoader } from "../utils/media-utils";
+import { resolveUrl } from "../utils/media-utils";
+import { proxiedUrlFor } from "../utils/media-url-utils";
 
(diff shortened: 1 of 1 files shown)


src/components/particle-emitter.js
```diff
@@ -32,22 +35,37 @@ AFRAME.registerComponent("particle-emitter", {
...
+
+      // todo: we don't need to proxy for many things if the canonical URL has permissive CORS headers
+      accessibleUrl = proxiedUrlFor(canonicalUrl, null);
+    }
+
+    const texture = new THREE.Texture();
+
+    await textureLoader.loadTextureAsync(texture, accessibleUrl);
+
+    this.particleEmitter.material.uniforms.map.value = texture;
```
netpro2k: Can particle emitters change src? If so we need to check here that src didn't change while the request was in flight.

robertlong: I'd like to make it so they could since the change is pretty simple. I'll add that, thanks 👍

PR #1593, opened 2019-07-31
Big sign-in refactoring

mqp:
**WIP because I still want to clean things up a bit and it's too stupid to merge this without testing carefully every flow.**

Goals of this refactoring:

- Prevent sign-in control flow from constantly jumping back and forth between 2-3 big god files, which makes it impossible to follow (ui-root.js, hub.js, and to a lesser extent scene-entry-manager.js.)
- Eliminate the unnecessary use of React state and props as an temporary message bus between parts of the sign-in flow.
- Incidentally fix ~most~ some of the random sign-in bugs by removing all the parts of the process where the bugs were (i.e. corrupting said state and props.)

Note that I intend to completely rework the pinning stuff in a future PR, so this is a relatively temporary state for the pin/unpin code to be in.


src/components/pinnable.js
```diff
@@ -18,13 +18,17 @@ AFRAME.registerComponent("pinnable", {
   },
 
   _fireEventsAndAnimate(oldData, force) {
-    // We need to guard against _fireEventsAndAnimate being called during entity initialization,
-    // when the networked component isn't initialized yet.
-    if (this.el.components.networked && this.el.components.networked.data && !NAF.utils.isMine(this.el)) return;
+    // during object initialization, we may not have a networked component, and we
+    // also have no desire to fire pinned events or animate anything
+    if (!this.el.components.networked) {
```
mqp: @brianpeiris you may have an opinion about this. The failure state of the previous code was during object inflation, when the pinnable component may be instantiated before the networked component. In production right now, `pinned` events do fire during inflation of room objects as a result of this, and we just happen to ignore them because we inflate room objects surprisingly early in the initial load.

mqp:
The fact that `signedIn` is redundant between the React props and state and gets automatically updated sometimes to be in sync is weird and probably broken but I haven't worked out what to do about it yet.

gfodor:
Thanks for looking into this -- can you give an example of a scenario where these props end up getting corrupted?

The incidental complexity we had before in the data flow was in-part due to the intentional move of the event handlers into the `scene-entry-manager (a misnomer) which was basically intended to move stuff around to cut down the size of `hub.js` and `ui-root.js`. i agree if it causes bugs though it should be reconsidered.

mqp:
The severe bug this fixes is the one where if you ever close the dialog that says "Email sent" and then in the same Hubs session sign in later, stuff will get all screwed up, because the `showSignInDialog` prop will get out of sync with the state of the dialog actually being on the screen. Now the `showSignInDialog` prop doesn't exist, so that is impossible. Like I mentioned above, there are also ways you can get the `signedIn` state to disagree with the fact of whether you are signed in or not, but I'm not sure that causes any negative user-visible side effects except maybe rendering some inaccurate UI.

I would be happy to find nice ways to split up this code, but putting parts of it on opposite sides of the React component boundary seems like a very awkward way to do it. For example, having the code in `performConditionalSignIn` live inside of the `SignInDialog` React component class seems very reasonable, and having the OAuth code and state live inside the `OAuthDialog` React component class would be likewise reasonable.

mqp:
This last commit also changes the inconsistent behavior about post-sign-in notifications ("You are now signed in" dialog) to always present the dialog.

src/components/pinnable.js
brianpeiris: It would be nice if we didn't do this implicit pinning based on events, but rather did an explicit "pin" operation triggered directly from user actions, but you know that already. Other than that, I have no opinion on this. This seems like a fine change to make.

mqp:
A few related salient things that were already problems with this code, but were not yet fixed in this PR, whose scope I am trying not to expand to infinity:

- Bugs where if you trigger `closeDialog` at a time when you don't have a dialog open, you get booted back out of Hubs.
- The odd fact that the close button on each sign-in dialog still tries to do the thing, so if the state preventing you from doing it changed, the thing will happen. For example, if you try to kick someone, and you get the sign-in dialog because you don't have permissions, and then you click X out of the sign-in dialog, and in the meantime you got permissions, it will kick them right then and there.
- Suspicious UI behavior related to the `signedIn` React state.
- Extreme lack of error handling during the sign-in process. For example, we don't log exceptions that happen on the "action" requiring sign-in.
- Bugs involving state in the Twitter OAuth flow.
- No mechanism to clean up the authentication continuation if you send a link and don't use it, which is fundamentally kind of busted -- it just doesn't make any sense that you can e.g. kick someone, X out when you change your mind, 10 minutes later you click the link in your email to sign in, and then they get kicked.

Some of these I will fix soon and some maybe I will shrug and move on.

brianpeiris:
Did a quick read-through. This looks much better than the previous situation, especially with regard to the hops we were going through for `performConditionalSignIn`, it does make sense for that code to live in ui-root. 
I think it made sense for the related code to live outside ui-root at an early point during its evolution, but that quickly became false and we didn't see the forest for the trees enough to realize that it ought to be refactored.
It's unfortunate that ui-root has now ballooned in size again, but I think the answer to that might be to carve some of its more independent functionality into smaller individual react components. Perhaps a task for another day.

Thanks for doing this. Nice work!

PR #1595, opened 2019-07-31
Migrate to pdf.js

gfodor:
This greatly improves performance + quality of PDFs in hubs by moving the pdf rendering to the client using pdf.js.

Also fixes up what seem to be some bugs in texture management in `media-image`.

Also, for uploads, drops the use of the `raw` response from the media resolver, which lets us remove references to farspark from reticulum. (see https://github.com/mozilla/reticulum/pull/208)


src/components/media-views.js
```diff
@@ -837,3 +846,150 @@ AFRAME.registerComponent("media-image", {
...
+    let texture;
+    let ratio = 1;
+
+    try {
+      const { src, index } = this.data;
+      if (!src) return;
+
+      this.el.emit("pdf-loading");
+
+      if (!oldData.src || src !== oldData.src) {
```
netpro2k: src will be truthy here because of the check above, so the check for `!oldData.src` seems redundnat

src/components/media-views.js
```diff
@@ -761,7 +766,11 @@ AFRAME.registerComponent("media-image", {
 
       let cacheItem;
       if (textureCache.has(src)) {
-        cacheItem = textureCache.retain(src);
+        if (this.currentSrcIsRetained) {
```
netpro2k: Hmm, a bit surprised if this is ever happening...

src/utils/pinned-entity-to-gltf.js
```diff
@@ -32,7 +32,10 @@ export default function pinnedEntityToGltf(el) {
 
     gltfComponents.media = { src: mediaSrc, contentSubtype: mediaContentSubtype, id: networkId };
 
-    if (components["media-pager"]) {
+    if (components["media-pdf"]) {
+      gltfComponents.media.pageIndex = components["media-pdf"].data.index;
+    } else if (components["media-pager"]) {
+      // Legacy
```
netpro2k: Since this is only for serializing, do we need to support legacy?

src/components/media-views.js
```diff
@@ -837,3 +846,150 @@ AFRAME.registerComponent("media-image", {
...
+        this.canvas = document.createElement("canvas");
+        this.canvasContext = this.canvas.getContext("2d");
+        this.canvas.width = pw;
+        this.canvas.height = ph;
+      }
+
+      const renderingSrc = this.data.src;
+      const renderingIndex = index;
+
+      if (this.renderTask) {
```
netpro2k: RenderTask has a `cancel` method. https://mozilla.github.io/pdf.js/api/draft/RenderTask.html, if you in fact still want to wait its probably better to just await on `this.renderTask.promise` instead of generating this new polling promise to wait on.

src/components/media-views.js
```diff
@@ -837,3 +846,150 @@ AFRAME.registerComponent("media-image", {
...
+    if (this.currentPageTextureIsRetained) {
+      const cacheKey = this.currentTextureCacheKey();
+      textureCache.release(cacheKey);
+      this.currentPageTextureIsRetained = false;
+    }
+  },
+
+  currentTextureCacheKey() {
+    // We ensure a unique texture for each PDF entity, because they are drawn over during pagination.
+    return `${this.el.object3D.uuid}_${this.canvas.width}_${this.canvas.height}`;
```
netpro2k: The main point of the texture cache is to re-use textures across multiple object3Ds. If we key it light this it would probably be better just to have a single `CanvasTexture` per `media-pdf` and just resize its canvas as needed, avoiding the texture cache al together. 

If we wanted to actually share textures across multiple instances of the same PDF, its possible, but you would probably end up needing one texture per page, which seems probably worse overall, just to handle the uncommon case of displaying multiple instances of the same page of a PDF.

src/components/media-views.js
```diff
@@ -8,6 +8,11 @@ import { proxiedUrlFor } from "../utils/media-url-utils";
 import { buildAbsoluteURL } from "url-toolkit";
 import { SOUND_CAMERA_TOOL_TOOK_SNAPSHOT } from "../systems/sound-effects-system";
 import { promisifyWorker } from "../utils/promisify-worker.js";
+import pdfjs from "pdfjs-dist";
+
+// Using external CDN to reduce build size
+pdfjs.GlobalWorkerOptions.workerSrc =
```
netpro2k: In the interest of self hosting, might be better to point at one of the general cdn versions of this instead https://github.com/mozilla/pdf.js#including-via-a-cdn

src/components/media-views.js
```diff
@@ -734,7 +739,7 @@ AFRAME.registerComponent("media-image", {
     if (this.data.batch && this.mesh) {
       this.el.sceneEl.systems["hubs-systems"].batchManagerSystem.removeObject(this.mesh);
     }
-    if (this._hasRetainedTexture) {
+    if (this.currentSrcIsRetained) {
```
netpro2k: Woops thats a nasty bug, means images are not getting released.

src/components/media-views.js
```diff
@@ -837,3 +846,150 @@ AFRAME.registerComponent("media-image", {
...
+      const cacheKey = this.currentTextureCacheKey();
+      let cacheItem;
+      if (textureCache.has(cacheKey)) {
+        if (this.currentPageTextureIsRetained) {
+          cacheItem = textureCache.get(cacheKey);
+        } else {
+          cacheItem = textureCache.retain(cacheKey);
+        }
+      } else {
+        if (src === "error") {
```
netpro2k: Dont think you need to handle this case. Its dont for images since a `media-image` with `src: "error"`is used for displaying errors in general (ex when failing to load a GLTF, or when loading an unknown content type). A `media-pdf` should never get created with`src: "error"`.

src/components/media-views.js
```diff
@@ -837,3 +846,150 @@ AFRAME.registerComponent("media-image", {
...
+          cacheItem = textureCache.set(cacheKey, texture);
+        }
+      }
+
+      texture = cacheItem.texture;
+      ratio = cacheItem.ratio;
+
+      this.currentPageTextureIsRetained = true;
+    } catch (e) {
+      console.error("Error loading image", this.data.src, e);
```
netpro2k:
```suggestion
      console.error("Error loading PDF", this.data.src, e);
```

src/components/media-views.js
```diff
@@ -837,3 +846,150 @@ AFRAME.registerComponent("media-image", {
...
+
+    this.mesh.material.transparent = texture == errorTexture;
+    this.mesh.material.map = texture;
+    this.mesh.material.map.needsUpdate = true;
+    this.mesh.material.needsUpdate = true;
+
+    scaleToAspectRatio(this.el, ratio);
+
+    if (texture !== errorTexture && this.data.batch) {
+      this.el.sceneEl.systems["hubs-systems"].batchManagerSystem.addObject(this.mesh);
```
netpro2k: Hard to say if we should bother batching PDF pages. Feels a bit of a shame to bother atlasing the textures while quickly paging through a document, but maybe its fine.

package.json
```diff
@@ -73,6 +73,8 @@
     "naf-janus-adapter": "^2.0.2",
     "networked-aframe": "github:mozillareality/networked-aframe#master",
     "nipplejs": "github:mozillareality/nipplejs#mr-social-client/master",
+    "node-ensure": "0.0.0",
```
netpro2k: Whats with this dependency? Dont see it being imported and doesn't appear to be a peer dependency of pdfjs

src/components/media-views.js
```diff
@@ -837,3 +846,150 @@ AFRAME.registerComponent("media-image", {
...
+AFRAME.registerComponent("media-pdf", {
+  schema: {
+    src: { type: "string" },
+    projection: { type: "string", default: "flat" },
+    contentType: { type: "string" },
+    index: { default: 0 },
+    batch: { default: false }
+  },
+
+  init() {},
```
netpro2k: No need to include empty methods, I believe aframe checks for them before attempting to call them.

src/components/media-views.js
```diff
@@ -837,3 +846,150 @@ AFRAME.registerComponent("media-image", {
...
+    try {
+      const { src, index } = this.data;
+      if (!src) return;
+
+      this.el.emit("pdf-loading");
+
+      if (!oldData.src || src !== oldData.src) {
+        const loadingSrc = this.data.src;
+        this.pdf = await pdfjs.getDocument(src);
+        if (loadingSrc !== this.data.src) return;
```
netpro2k: This check should technically happen before changing `this.pdf`.. Though I am kind of thinking now that pdfs no long mutate `media-images` we just make all the `media-*` components (including media-loader) not support changing src after they have been initialized. Seems like it would drastically simplify the code.

src/components/media-views.js
gfodor: ah yeah, I tried doing this at first. the reason this is difficult to do in another way is because the user can potentially slam on the pagination and this results in `update()` being called many times in succession, while a render is in-flight. to properly handle cancelling, you need to cancel and then wait for the original promise to resolve. so the orchestration here is quite complicated: you need to submit cancel, and then the originating call stack and your new call stack are both waiting on the same promise, and then if another request comes in, you don't want to re-cancel but still need to wait on the promise.

i was ready to wave the white flag until i realized I can just serialize the render requests, and the last one submitted will always be the one we want anyway, which is what it does now.

gfodor: i'll take another stab at it real quick

src/utils/pinned-entity-to-gltf.js
gfodor: oh good catch, i mis-interpreted it as I was going through the code that this was a deserializer

src/components/media-views.js
gfodor: OK, so I managed to avoid the interval and redundant renders by changing the control flow-- but cancelling seems problematic. If I cancel and handle it, it seems to corrupt the rendering somehow where the page ends up rendering into a small region on the canvas. Odds are it's something I'm doing wrong, but the net effect now is that the worst case is you need to wait for a single render to complete before seeing your page.

src/components/media-views.js
gfodor: i guess for images it would only happen if update gets called for the same src, which seems unlikely currently -- because it'd have to be if the content type changed. i think this code is more correct though, and covers the future potential cases where there is more state on this component that can be mutated. (eg perhaps an image filter)

src/components/media-views.js
gfodor: yep that makes sense. earlier versions of this code was trying to cache at the page level and now that I landed on a texture-per-entity the cache makes no sense.

package.json
gfodor: the build was failing complaining about it due to `pdfjs-dist` needing it. i've seen this happen in the past with other libraries, not sure

gfodor:
Oh, I guess my logic was wrong and these PDFs are not being batched. My guess is that the atlas is not being updated in response to the texture's `needsUpdate` being flipped. Probably something for a future PR.

src/components/media-views.js
gfodor: oh, good catch

gfodor:
good review, thanks!

PR #1596, opened 2019-07-31
Update authorization

brianpeiris:
- Wait for initial presence sync before having NAF connect to room
- Extend NAF schemas with `authorizedSchemas` structure that authorizes components and properties regardless of permissions (for media syncing, in this case)
- Use permissions in presence to authorize or sanitize incoming NAF update and removal messages.

Goes with https://github.com/mozilla/reticulum/pull/209


gfodor:
one nit (this is totally my fault, I should have mentioned it sooner) -- calling these things `authorized` is a bit confusing since it could mean what it does mean "this component is always authorized to be updated" vs "this is a component subject to authorization" -- if I were a first time reader I might mis-interpret this to mean that the components marked as authorized are the components we apply security policies against, when in fact its the opposite.

alternative names:
- `publicComponents`
- `nonAuthorizedComponents`
- `alwaysGrantedComponents`
- `insecureComponents`

src/utils/permissions-utils.js
```diff
@@ -19,3 +19,129 @@ export function canMove(entity) {
...
+  for (const index in data.components) {
+    if (!data.components.hasOwnProperty(index)) continue;
+    if (!authorizedIndices.includes(index)) {
+      data.components[index] = null;
+    }
+  }
+  return data;
+}
+
+function authorizeEntityManipulation(entity, sender, senderPermissions) {
```
gfodor: maybe `Sync` instead of `Manipulation`?

src/utils/permissions-utils.js
```diff
@@ -19,3 +19,129 @@ export function canMove(entity) {
...
+    return true;
+  } else {
+    const { template } = entity.components.networked.data;
+    sanitizeMessageData(template, data);
+    return true;
+  }
+}
+
+const emptyObject = {};
+export function authorizeOrSanitizeMessage(message) {
```
gfodor: maybe `s/Message/Sync/`

src/utils/permissions-utils.js
```diff
@@ -19,3 +19,129 @@ export function canMove(entity) {
...
+    const entity = NAF.entities.getEntity(message.data.networkId);
+    if (!entity) return emptyObject;
+    if (authorizeEntityManipulation(entity, from_session_id, senderPermissions)) {
+      return message;
+    } else {
+      return emptyObject;
+    }
+  } else {
+    // Pass through for other data types. Namely, "drawing-<networkId>" messages
+    return message;
```
gfodor: seems like in this context we should whitelist + reject unknown message types

src/utils/permissions-utils.js
```diff
@@ -19,3 +19,129 @@ export function canMove(entity) {
...
+    const authorizedOrSanitized = authorizeOrSanitizeMessageData(message.data, from_session_id, senderPermissions);
+    if (authorizedOrSanitized) {
+      return message;
+    } else {
+      return emptyObject;
+    }
+  } else if (dataType === "r") {
+    const entity = NAF.entities.getEntity(message.data.networkId);
+    if (!entity) return emptyObject;
+    if (authorizeEntityManipulation(entity, from_session_id, senderPermissions)) {
```
gfodor: ah I see now `Sync` isn't the only context

src/utils/permissions-utils.js
gfodor: nvmd, forgot about removals

src/utils/permissions-utils.js
gfodor: nvmd, forgot about removals

src/utils/permissions-utils.js
brianpeiris: We decided not to bother with a whitelist on the server side either. I think it's safe not to bother with a whitelist for `dataTypes`, since they need explicit subscribers already.

PR #1598, opened 2019-08-01
Add inspect button to interactable media.

johnshaughnessy:
This PR replaces the "align" button with an "inspect" button. Inspecting an object means to transform your camera such that you are looking directly at the object from in front of it.


src/components/ik-controller.js
```diff
@@ -168,9 +168,10 @@ AFRAME.registerComponent("ik-controller", {
     }
 
     const root = this.ikRoot.el.object3D;
+    root.updateMatrices();
     const { camera, leftController, rightController } = this.ikRoot;
 
-    camera.object3D.updateMatrix();
```
gfodor: for this to be semantically the same, do you need to flip `matrixNeedsUpdate = true`? iirc `updateMatrix` forces a local matrix update.

src/components/pitch-yaw-rotator.js
```diff
@@ -1,55 +1,50 @@
...
+      const scene = AFRAME.scenes[0];
+      const userinput = scene.systems.userinput;
+      const cameraDelta = userinput.get(
+        scene.is("entered") ? paths.actions.cameraDelta : paths.actions.lobbyCameraDelta
+      );
+      if (cameraDelta) {
+        rotatePitchAndYaw(this.el.object3D, this.pendingXRotation + cameraDelta[1], cameraDelta[0]);
+      } else if (this.pendingXRotation) {
+        rotatePitchAndYaw(this.el.object3D, this.pendingXRotation, 0);
+      }
```
gfodor: was it intentional to leave `this.on === true` at the end here?

src/components/visibility-while-frozen.js
```diff
@@ -18,8 +18,10 @@ AFRAME.registerComponent("visibility-while-frozen", {
   init() {
     this.updateVisibility = this.updateVisibility.bind(this);
     this.camWorldPos = new THREE.Vector3();
+    this.cam2WorldPos = new THREE.Vector3();
     this.objWorldPos = new THREE.Vector3();
-    this.cam = this.el.sceneEl.camera.el.object3D;
+    this.cam = document.getElementById("player-camera").object3D;
+    this.cam2 = this.el.sceneEl.camera;
```
gfodor: i think we can check if these cameras are equal and skip the extra work if so

src/hub.html
```diff
@@ -987,7 +984,7 @@
                 </template>
 
                 <template data-name="Head">
-                    <a-entity id="player-head" visible="false" bone-visibility></a-entity>
```
gfodor: does this cause head to be visible for a frame when spawning?

src/systems/camera-system.js
```diff
@@ -0,0 +1,182 @@
...
+function getInspectable(child) {
+  let el = child;
+  while (el) {
+    if (el.components && el.components.tags && el.components.tags.data.inspectable) return el;
+    el = el.parentNode;
+  }
+  return null;
+}
+
+AFRAME.registerComponent("inspect-button", {
```
gfodor: new file?

src/systems/hubs-systems.js
```diff
@@ -41,10 +44,25 @@ AFRAME.registerSystem("hubs-systems", {
...
-    // batchManager is ticked in "post-physics" system
   },
 
   remove() {
     this.cursorTargettingSystem.remove();
   }
 });
+
+const i = window.setInterval(() => {
+  if (AFRAME.scenes[0] && AFRAME.scenes[0].systems && AFRAME.scenes[0].systems.physics) {
```
gfodor: this seems pretty hacky -- can't we always register and move this check lazily into `tick`?

src/systems/interactions.js
```diff
@@ -182,7 +182,16 @@ AFRAME.registerSystem("interaction", {
       this.state.rightRemote.held.components.tags.data.isPen &&
       this.state.rightRemote.held.children[0].components.pen.intersection;
 
-    const enableRightRemote = !rightHandInteracting && !rightHandTeleporting && !rightRemotePenIntersectingInVR;
+    const enableRightRemote =
+      !rightHandInteracting &&
+      !rightHandTeleporting &&
+      !rightRemotePenIntersectingInVR &&
+      !(
```
gfodor: i'd extract into `rightHandInspecting` var

src/systems/hubs-systems.js
gfodor: ah i see this is how it was before -- still hacky :)

src/systems/sprites.js
```diff
@@ -47,22 +47,22 @@ AFRAME.registerComponent("sprite", {
   schema: { name: { type: "string" } },
   tick() {
     // TODO when we run out of sprites we currently just stop rendering them. We need to do something better.
-    if (!(this.didRegisterWithSystem || this.didFailToRegister) && this.el.sceneEl.systems["post-physics"]) {
-      const registered = this.el.sceneEl.systems["post-physics"].spriteSystem.add(this);
+    if (!(this.didRegisterWithSystem || this.didFailToRegister) && this.el.sceneEl.systems["hubs-systems"]) {
+      const registered = this.el.sceneEl.systems["hubs-systems"].spriteSystem.add(this);
```
gfodor: if you do refactor that `setInterval` business, I guess you'll need a `.ready` flag or something here.

src/utils/auto-box-collider.js
```diff
@@ -21,7 +21,7 @@ const computeObjectAABB = (function() {
 })();
 
 const rotation = new THREE.Euler();
-export function getBox(entity, boxRoot) {
+export function getBox(entity, boxRoot, dontLocalize) {
```
gfodor: suggest: `dontLocalize` -> `worldSpace`

gfodor:
I'd ask @netpro2k to review the input system changes, I still don't feel well versed enough to recognize potential improvements or bugs

src/systems/camera-system.js
```diff
@@ -0,0 +1,182 @@
...
+      .multiplyScalar(calculateViewingDistance(camera, object))
+      .applyQuaternion(owq);
+    v2.copy(camera.position).applyQuaternion(rig.quaternion);
+    rig.position.subVectors(v1, v2).add(owp);
+
+    rig.matrixNeedsUpdate = true;
+  };
+})();
+
+let numCameraModes = 0;
```
gfodor: this is a bit weird to initialize constants this way -- I'd just write them

src/systems/camera-system.js
```diff
@@ -0,0 +1,182 @@
...
+const offset = new THREE.Vector3();
+export class CameraSystem {
+  constructor() {
+    this.mode = CAMERA_MODE_FIRST_PERSON;
+    waitForDOMContentLoaded().then(() => {
+      this.playerCamera = document.getElementById("player-camera");
+      this.playerRig = document.getElementById("player-rig");
+      this.cameraEl = document.getElementById("experimental-camera");
+      this.rigEl = document.getElementById("experimental-rig");
+      this.rigEl.object3D.matrixWorldNeedsUpdate = true;
```
gfodor: why is this here and not at the call site which modifies the matrix? also the idiomatic way to do this is just `.matrixNeedsUpdate = true`, the world flag is ideally just used internally (the `matrixNeedsUpdate` flag should deal with the world and local cases)

src/systems/camera-system.js
```diff
@@ -0,0 +1,182 @@
...
+      this.rigEl.object3D.matrixIsModified = true;
+    });
+  }
+  nextMode() {
+    if (!qsTruthy("thirdPerson")) return;
```
gfodor: I'd move this into a constant at the top instead of checking qs vars in the system

src/systems/camera-system.js
gfodor: oic, you need `numCameraModes` -- not a huge deal seems fine all things considered. if it were me i'd probably declare the constants as numbers and make an explicit array that lists them in the order we want them cycled.

src/systems/camera-system.js
```diff
@@ -0,0 +1,182 @@
...
+    if (this.mode !== CAMERA_MODE_INSPECT) {
+      this.preInspectMode = this.mode;
+    }
+    this.mode = CAMERA_MODE_INSPECT;
+    this.inspected = o;
+    moveRigSoCameraLooksAtObject(this.rigEl.object3D, this.cameraEl.object3D, this.inspected);
+  }
+  uninspect() {
+    this.inspected = null;
+    if (this.mode !== CAMERA_MODE_INSPECT) {
```
gfodor: nit: can make one liner `if (this.mode !== CAMERA_MODE_INSPECT) return`, also probably would put as first line of the function to show it's a guard clause

src/systems/camera-system.js
```diff
@@ -0,0 +1,182 @@
...
+      return;
+    }
+    this.mode = this.preInspectMode || CAMERA_MODE_FIRST_PERSON;
```
gfodor: `preInspectMode` presumably should be reset to `null` here

src/systems/camera-system.js
```diff
@@ -0,0 +1,182 @@
...
+  }
+  tick() {
+    this.playerHead = this.playerHead || document.getElementById("player-head");
+    if (!this.playerHead) return;
+    if (!this.playerCamera) return;
```
gfodor: if this is ever `true`, will it ever become `false`?

src/systems/camera-system.js
```diff
@@ -0,0 +1,182 @@
...
+    this.playerCamera.components["pitch-yaw-rotator"].on = true;
+    this.cameraEl.components["pitch-yaw-rotator"].on = true;
+
+    const userinput = AFRAME.scenes[0].systems.userinput;
```
gfodor: consider moving this into a `this` field

src/systems/camera-system.js
```diff
@@ -0,0 +1,182 @@
...
+    if (userinput.get(paths.actions.nextCameraMode)) {
+      this.nextMode();
+    }
+
+    const visible = this.mode !== CAMERA_MODE_FIRST_PERSON;
```
gfodor: `s/visible/headVisible` or `headShouldBeVisible`

src/systems/camera-system.js
```diff
@@ -0,0 +1,182 @@
...
+    if (visible !== this.playerHead.object3D.visible) {
+      this.playerHead.object3D.visible = visible;
+    }
+
+    this.playerRig.object3D.updateMatrices();
```
gfodor: `updateMatrices` likely should be called only if you need to ensure `matrixWorld` is up-to-date, doesn't seem needed for `this.rigEl` whose matrix is written to, not read

src/systems/camera-system.js
```diff
@@ -0,0 +1,182 @@
...
+    this.rigEl.object3D.updateMatrices();
+    if (this.mode === CAMERA_MODE_FIRST_PERSON) {
+      this.cameraEl.components["pitch-yaw-rotator"].on = false;
+      m2.copy(this.playerRig.object3D.matrixWorld);
```
gfodor: I don't think you need to copy into `m2` here, can just do `setMatrixWorld(this.rigEl.object3D, this.playerRig.object3D.matrixWorld)`

src/systems/camera-system.js
```diff
@@ -0,0 +1,182 @@
...
+      setMatrixWorld(this.rigEl.object3D, m2);
+    }
+
+    this.playerCamera.object3D.matrixWorldNeedsUpdate = true;
```
gfodor: this flag should not be necessary -- it's used internally and the contract is "if you update the position/rotation/scale of an object you need to set `matrixNeedsUpdate = true`"

src/systems/camera-system.js
```diff
@@ -0,0 +1,182 @@
...
+    if (this.mode === CAMERA_MODE_FIRST_PERSON) {
+      if (!AFRAME.scenes[0].is("vr-mode")) {
+        setMatrixWorld(this.cameraEl.object3D, this.playerCamera.object3D.matrixWorld);
+      } else {
+        setMatrixWorld(this.playerCamera.object3D, this.cameraEl.object3D.matrixWorld);
```
gfodor: `this.cameraEl.object3D.updateMatrices()` should be called before reading `matrixWorld` here

src/systems/camera-system.js
gfodor: after reading the rest of the code, I think you might misunderstand the contract of these various bits. here's the contract (might want to verify with @netpro2k, even though I wrote the code I tend to forget it ;))

- If you update the position/rotation/scale of an object3D, you *must* set `matrixNeedsUpdate = true`
- You shouldn't ever have to touch `matrixWorldNeedsUpdate`
- If you're using `getWorldPosition`, etc, you don't need to do anything special, just use those functions
- If you're reading `matrix` or `matrixWorld` directly, you need to call `updateMatrices()` on the object first

src/systems/hubs-systems.js
johnshaughnessy: This will go away will Kevin's upcoming changes to the physics system. The issue we have before then is that we don't want to register this system until after the physics system registers itself (because registration order == tick order), and the physics system registration happens asynchronously (hence this hack). Planning to leave it as-is since it'll get removed soon.

src/systems/camera-system.js
johnshaughnessy: I think the check was unnecessary because the body of the `waitForDomContentLoaded.then( ()=> ` will run before `tick`

src/systems/camera-system.js
johnshaughnessy: I still want `this.inspected = null` to run in the case this is called from `nextMode` but I did make the check a one liner.

src/systems/camera-system.js
johnshaughnessy: Yep, corrected these

PR #1608, opened 2019-08-05
Physics/three-ammo

InfiniteLee:
Replace `aframe-physics-system` with [`three-ammo`,](https://github.com/InfiniteLee/three-ammo) a non-aframe, three.js-only version of the Ammo.js portion of the library. This provides better control over the lifecycle of physics and physics related components and will make future attempts at moving physics to workers easier. 

TODO:
- [x] shapes for pinned objects are offset incorrectly


src/components/body-helper.js
```diff
@@ -0,0 +1,50 @@
...
+    emitCollisionEvents: { default: false },
+    disableCollision: { default: false },
+    collisionFilterGroup: { default: 1 }, //32-bit mask,
+    collisionFilterMask: { default: 1 }, //32-bit mask
+    scaleAutoUpdate: { default: true }
+  },
+
+  init: function() {
+    this.system = this.el.sceneEl.systems["hubs-systems"].physicsSystem;
+    this.body = new Body(this.data, this.el.object3D, this.system.world);
```
johnshaughnessy: Is there a race condition where `this.system.world` may be undefined given its asynchronous definition in `physics-system.js`: 
```
    AmmoModule().then(() => {
      this.world = new World({
        debugDrawMode: THREE.AmmoDebugConstants.DrawWireframe,
        gravity: { x: 0, y: -9.8, z: 0 }
      });
    });
```
?

src/systems/hubs-systems.js
```diff
@@ -23,16 +26,18 @@ AFRAME.registerSystem("hubs-systems", {
...
-  tick(t) {
+  tick(t, dt) {
     const systems = AFRAME.scenes[0].systems;
     systems.userinput.tick2();
-    systems.interaction.tick2(this.soundEffectsSystem);
     this.superSpawnerSystem.tick();
     this.cursorTargettingSystem.tick(t);
     this.constraintsSystem.tick();
+    this.physicsSystem.tick(dt);
+    systems.interaction.tick2(this.soundEffectsSystem);
```
johnshaughnessy: Not sure what the optimal tick order is (physics before interaction or interaction before physics), but since `superSpawnerSystem`, `cursorTargettingSystem` and `constraintsSystem` read from `interaction` system state, I think they should come after the `interaction` system tick. In the current order, those three systems have the previous frame's interaction system state but the current frame's userinput state. I suggest `tick`ing physics either first or after all interactions have happened, unless there's a reason I'm missing for why we want it to happen in the middle.

InfiniteLee: I didn't know those systems read from `interaction` system, just wanted to make sure `interaction` ticked after physics did. In that case, I agree, `physics` tick should just happen first.

InfiniteLee: Oh except `constraintSystem` needs to tick before `physicsSystem`

src/components/body-helper.js
InfiniteLee: This is fixed now.

src/systems/physics-system.js
```diff
@@ -0,0 +1,101 @@
...
+  tick(dt) {
+    if (this.world) {
+      if (this.debugRequested !== this.debugEnabled) {
+        this.debugEnabled = this.debugRequested;
+        if (this.debugEnabled) {
+          this.world.getDebugDrawer(this.scene).enable();
+        } else {
+          this.world.getDebugDrawer(this.scene).disable();
+        }
+      }
```
mqp: Why did you choose to do this in `tick`, instead of having `setDebug` call functions on the debug drawer directly?

src/systems/physics-system.js
```diff
@@ -0,0 +1,101 @@
...
+
+      while (this.bodyHelpers.length > 0) {
+        this.bodyHelpers.shift().init2();
+      }
+      while (this.shapeHelpers.length > 0) {
+        this.shapeHelpers.shift().init2();
+      }
```
mqp: Is there a reason to do it this way instead of
```javascript
for (const bodyHelper of bodyHelpers) {
  bodyHelper.init2();
}
for (const shapeHelper of shapeHelpers) {
  shapeHelper.init2();
}
this.shapeHelpers = [];
this.bodyHelpers =[];
```
My way looks a little more normal to me.

src/systems/physics-system.js
```diff
@@ -0,0 +1,101 @@
...
+  addBody(body) {
+    this.bodies.push(body);
+  }
+
+  removeBody(body) {
+    const idx = this.bodies.indexOf(body);
+    if (idx !== -1) {
+      this.bodies.splice(idx, 1);
+    }
+  }
```
mqp: It's not clear to me whether the order of the bodies is important to the operation of physics -- if it's not then you could use a set and this would be nicer.

src/systems/physics-system.js
mqp: Is there a reason to do this in `tick`, instead of doing it when Ammo is loaded and the `world` is created?

src/systems/interactions.js
```diff
@@ -5,28 +5,16 @@ import { waitForDOMContentLoaded } from "../utils/async-utils";
...
-        if (object3D.el && object3D.el.components.tags && object3D.el.components.tags.data.isHandCollisionTarget) {
-          return object3D.el;
-        }
-        return null;
-      }
+  const world = this.el.sceneEl.systems["hubs-systems"].physicsSystem.world;
+  const handPtr = Ammo.getPointer(body.physicsBody);
+
+  if (world.collisions.has(handPtr) && world.collisions.get(handPtr).length > 0) {
+    const object3D = world.object3Ds.get(world.collisions.get(handPtr)[0]);
```
mqp: Since this is reasonably hot code, maybe it would be nicer to do less work here? e.g. 
```javascript
const handCollisions = world.collisions.get(handPtr);
if (handCollisions != null && handCollisions.length > 0) {
  const object3D = world.object3Ds.get(handCollisions[0]);
  // ...
}

Comment (mqp) [reply src/systems/interactions.js]: Is this code correct? What if the hand is colliding with more than one thing, but the first one you look at isn't a collision target? The old version would continue looking at the other things, but the new version gives up.

File: src/systems/hubs-systems.js [hunk 7]
```diff
@@ -44,25 +48,13 @@ AFRAME.registerSystem("hubs-systems", {
...
-      },
-      tick(t, dt) {
-        this.hubsSystems.spriteSystem.tick(t, dt);
-        this.hubsSystems.batchManagerSystem.tick(t);
-        this.hubsSystems.cameraSystem.tick();
-      }
-    });
-    window.clearInterval(i);
-  }
-}, 1000);
```
mqp: This system code is all starting to look downright sane.

mqp:
I don't think the body/shape helper "registration" code is correct. Consider the following case:

1. Ammo starts loading.
2. A body helper component is initialized. `registerBodyHelper` is called.
3. The body helper component is removed. `removeBody` is called.
4. Ammo finishes loading.
5. A physics system tick occurs. `init2` is called. `addBody` is called. This body will be leaked forever.

If you swap the order of 3 and 4, the same result happens.

Also:

1. Ammo starts loading.
2. Body helper component X is initialized. `registerBodyHelper` is called.
3. Ammo finishes loading.
4. Body helper component Y is initialized. `registerBodyHelper` is called. `this.world` is non-null, so `init2` on Y is called directly.
5. A physics system tick occurs. `init2` on X is called.

This violated the invariant that bodies are initialized in the order the components are created. I don't know if that creates any problems, but it's certainly a little surprising.

There are many ways you can fix this, but here is how I believe I would have written the code:

- Store the promise returned by `AmmoModule().then(...)` in the physics system.
- Create a physics system function `whenReady` which returns that promise.
- Remove the `bodyHelpers` and `shapeHelpers` arrays in the physics system.
- In the body and shape helper, write:
```javascript
init: function() {
  physicsSystem.whenReady().then(world => {
    if (!this.removed) {
      this.body = world.addBody(new Body(/* ... */));
    }
  });
}
remove: function() {
  this.removed = true;
  if (this.body != null) {
    physicsSystem.removeBody(this.body);
  }
}
```

src/components/shape-helper.js
```diff
@@ -0,0 +1,81 @@
...
+    for (let i = 0; i < this.shapes.length; i++) {
+      this.bodyHelper.body.addShape(this.shapes[i]);
+    }
+  },
+
+  remove: function() {
+    if (this.shapes) {
+      for (let i = 0; i < this.shapes.length; i++) {
+        if (this.bodyHelper.body) {
+          this.bodyHelper.body.removeShape(this.shapes[i]);
```
mqp: It seems like a lot of this code, like this part, can be calling functions on a `body` that has already been destroyed, if the body helper component was removed. That seems bad.

mqp: I think it's very confusing that three-to-ammo's `removeShape` also destroys the shape. It's very not-RAII, it looks like a bug, and it prevents you from reusing a shape in multiple bodies, which it seems like you could otherwise do.

src/systems/interactions.js
mqp: Is this code correct? In the old version, if the hand collides with multiple things and the first one examined isn't a collision target, it continues to examine the other things. In this version, we give up.

src/systems/physics-system.js
InfiniteLee: This is due to a possible race condition where if you use the `?debug=true` flag, `setDebug` can get called before `world` exists.

src/systems/physics-system.js
InfiniteLee: In your example I would personally do:
```
this.shapeHelpers.length = 0;
this.bodyHelpers.length = 0;
```
But I find little difference doing it with a while loop vs a for loop. I agree though this can be moved into the promise callback where `world` is created.

src/systems/physics-system.js
InfiniteLee: The order doesn't matter but you can't iterate a set without generating garbage because of iterator instantiation, so I tend to avoid using sets (and maps) when I have to iterate over the data frequently.

src/components/shape-helper.js
InfiniteLee: > 
> 
> I don't think the body/shape helper "registration" code is correct. Consider the following case:
> 
>     1. Ammo starts loading.
> 
>     2. A body helper component is initialized. `registerBodyHelper` is called.
> 
>     3. The body helper component is removed. `removeBody` is called.
> 
>     4. Ammo finishes loading. `world` is created.
> 
>     5. A physics system tick occurs. `init2` is called. `addBody` is called. This body will be leaked forever.
> 
> 
> If you swap the order of 3 and 4, the same result happens.
> 
> Also:
> 
>     1. Ammo starts loading.
> 
>     2. Body helper component X is initialized. `registerBodyHelper` is called.
> 
>     3. Ammo finishes loading. `world` is created.
> 
>     4. Body helper component Y is initialized. `registerBodyHelper` is called. `world` is non-null, so `init2` on Y is called directly.
> 
>     5. A physics system tick occurs. `init2` on X is called.
> 
> 
> This violated the invariant that bodies are initialized in the order the components are created. I don't know if that creates any problems, but it's certainly a little surprising.
> 
> There are many ways you can fix this, but here is how I believe I would have written the code:
> 
>     * Store the promise returned by `AmmoModule().then(...)` in the physics system.
> 
>     * Create a physics system function `whenReady` which returns that promise.
> 
>     * Remove the `bodyHelpers` and `shapeHelpers` arrays in the physics system, and the `registerBodyHelper` and `registerShapeHelper` functions.
> 
>     * In the body and shape helper components, write something like:
> 
> 
> ```js
> init: function() {
>   physicsSystem.whenReady().then(world => {
>     if (!this.removed) {
>       this.body = world.addBody(new Body(/* ... */));
>     }
>   });
> }
> remove: function() {
>   this.removed = true;
>   if (this.body != null) {
>     physicsSystem.removeBody(this.body);
>   }
> }
> ```
> 
> (The shape helper one is more complicated, but I don't see any reason it should be a problem.)

You bring up a good point about what happens if these helper components get destroyed before world is created. In practice, this never really happens because the ammo .wasm loads very quickly before any of these components get initialized- this is all just trying to avoid any issues that might come in the case where something else terrible has gone wrong and the wasm somehow doesn't load at all.

That said, there is a reason I didn't do this with promises as you suggested:
1. The promise returned by `AmmoModule()` is not a "real" promise. This is not a huge deal because you can create a "real" promise and return that instead, and have at promise resolved when this fake one does.
2. It is important that bodies are initialized before shapes do, since they have to add themselves to the body.

I think though that in practice what will happen is you will get an error because you will try to call `tick2` on a component that no longer exists. So I think the correct fix is to check to make sure the "helper" component is still valid. 

As for the second concern about the order in which `init2` is called on bodies, it doesn't really matter in practice. However, the problem is resolved if the iteration of `this.bodyHelpers` and `this.shapeHelpers` is done in the `AmmoModule().then()` promise as opposed to in `tick` as you  suggested in another comment.

src/systems/physics-system.js
mqp: I see, that makes sense. I might have implemented `setDebug` in terms of a loading promise instead, such as
```javascript
setDebug: function(enabled) {
  this.whenReady().then(world => {
    world.getDebugDrawer(this.scene).enabled = enabled;
  });
}
```
But it's a matter of taste.

src/systems/physics-system.js
mqp: I see, that makes sense, I forgot about that.

src/components/shape-helper.js
mqp: I see, the issue of ordering between the bodies and their corresponding shapes does make it a lot less straightforward, and it's not obvious to me anymore how to improve on the way you did it (besides fixing correctness with some more explicit checks as you suggested.) Maybe I will think about it a little more.

mqp: (My first thought is that if you care a lot about the initialization ordering of separate components, you probably shouldn't have modelled them as A-Frame components. Components are supposed to be kind of conceptually independent of each other with independent lifecycles. So maybe you could remove the shape component and implement shapes by calling a function on the body component instead.)

PR #1610, opened 2019-08-06
Update bug issue template to use 'needs triage' label

misslivirose:
New bugs will be tagged as needing triage


johnshaughnessy:
Woops I did not mean to self-request a review but :+1: 
:bug: :eyes:

PR #1614, opened 2019-08-06
Fix avatar interaction on touchscreens

johnshaughnessy:
Fix https://github.com/mozilla/hubs/issues/1421


gfodor:
Do we really want to move the cursor onto stuff you can't interact with when using the touch screen? I think that seems wrong, because it will leave a little floating cursor behind your finger on the screen which serves no purpose. I think we just need to update the check here to not skip over the avatar menus somehow.

johnshaughnessy:
We do hide the cursor on touchscreen so that it's not left behind on objects, but we forgot to hide the hover effect that it leaves on the objects. I'll include a fix for that problem: https://github.com/mozilla/hubs/pull/1614/commits/0f3f1215d8053f33b9c72eec7f3e7dd5ad799803#diff-844fec38f85559beb2ecdcb97f48cff8R86

The main purpose of this change is to make sure that the cursor still hovers on hoverable items, which is are those children of the cursor targetting system's targets that the interaction system considers to be a hover target: https://github.com/mozilla/hubs//blob/bug%2Ftouchscreen-buttons/src/systems/userinput/devices/app-aware-touchscreen.js#L56 , which backs out to those items that have the `is-remote-hover-target` component

PR #1624, opened 2019-08-09
Fix issues with camera spawning

gfodor:
This attempts to fix the hard-to-repro issue around the spawn animation not running or running too late on the camera (leaving the scale at 0.5) -- it defers running the animation until the model has been loaded, etc.

### DIFF
--- a/src/components/camera-tool.js
+++ b/src/components/camera-tool.js
@@ -89,6 +89,7 @@ AFRAME.registerComponent("camera-tool", {
   },
 
   init() {
+    this.el.object3D.visible = false; // Make invisible until model ready
     this.lastUpdate = performance.now();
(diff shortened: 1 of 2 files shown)


src/components/camera-tool.js
```diff
@@ -130,6 +131,17 @@ AFRAME.registerComponent("camera-tool", {
       mesh.matrixNeedsUpdate = true;
       this.el.setObject3D("mesh", mesh);
 
+      this.el.object3D.visible = true;
+      this.el.object3D.scale.set(0.75, 0.75, 0.75);
```
johnshaughnessy: Should we toggle `this.el.object3D.matrixNeedsUpdate` here?

gfodor:
ah yes, good catch. i can update it in my next push. the animation should
deal with it too but we might be frame delayed.

PR #1627, opened 2019-08-09
In-VR emoji hud

brianpeiris:
Add's an in-VR emoji hud that corresponds to the new 2D UI emoji bar. Also various clean up things.


src/components/freeze-controller.js
```diff
@@ -7,22 +7,10 @@ import { SOUND_FREEZE, SOUND_THAW } from "../systems/sound-effects-system";
...
     this.onToggle = this.onToggle.bind(this);
   },
 
-  play: function() {
-    this.el.addEventListener(this.data.toggleEvent, this.onToggle);
-  },
-
-  pause: function() {
-    this.el.removeEventListener(this.data.toggleEvent, this.onToggle);
-  },
```
brianpeiris: These were unused remnants from when we used to have a pause button in the hud

src/hub.html
```diff
@@ -416,7 +416,7 @@
                     </a-entity>
                     <a-entity class="label-background" mixin="rounded-button" is-remote-hover-target tags="singleActionButton: true;" position="0 0.15 0.08" scale="0.75 0.75 0.75">
                     </a-entity>
-                    <a-entity class="ui interactable-ui camera-snap-menu" mixin="button-container" visible-if-permitted="spawn_camera">
```
brianpeiris: The `button-container` mixin was removed at some point in the past, so this was doing nothing.

src/react-components/2d-hud.js
```diff
@@ -51,7 +51,6 @@ class TopHUD extends Component {
     activeTip: PropTypes.string,
     history: PropTypes.object,
     onToggleMute: PropTypes.func,
-    onToggleFreeze: PropTypes.func,
```
brianpeiris: More unused code.

package.json
```diff
@@ -24,7 +24,9 @@
     "test": "npm run lint && npm run test:unit && npm run build",
     "test:unit": "ava",
     "stats": "rimraf ./dist && webpack --mode=production --json",
-    "spritesheet": "spritesheet-js -f custom-scale-css-modules -p src/assets/images/spritesheets/ --customCanvasWidth 840 --customCanvasWidth 840 -n css-spritesheet src/assets/images/css-sprites/* --hubsFlag anything --customHeight 40 --customWidth 40 && spritesheet-js -f json -p src/assets/images/spritesheets/ --padding 8 --divisibleByTwo -n sprite-system-spritesheet --powerOfTwo src/assets/images/sprites/*"
+    "spritesheet": "npm run spritesheet:css && npm run spritesheet:system",
+    "spritesheet:css": "spritesheet-js -f custom-scale-css-modules -p src/assets/images/spritesheets/ --customCanvasWidth 840 --customCanvasWidth 840 -n css-spritesheet src/assets/images/css-sprites/* --hubsFlag anything --customHeight 40 --customWidth 40",
+    "spritesheet:system": "spritesheet-js -f json -p src/assets/images/spritesheets/ --padding 8 --divisibleByTwo -n sprite-system-spritesheet --powerOfTwo src/assets/images/sprites/*"
```
brianpeiris: Just split the commands here, for readability

src/hub.js
```diff
@@ -624,6 +626,11 @@ document.addEventListener("DOMContentLoaded", async () => {
   const scene = document.querySelector("a-scene");
   scene.setAttribute("shadow", { enabled: window.APP.quality !== "low" }); // Disable shadows on low quality
 
+  // HACK - Trigger initial batch preparation with an invisible object
```
johnshaughnessy: Why is this necessary / desirable?

brianpeiris: This is only tangentially related to the emoji feature. Currently, the batching system compiles its shader code the very first time you load an image into a room. This causes a large (~3s) hitch on both Firefox and Chrome. Obviously this affects the emoji picture, since we're spawning the emoji in the world as a regular old `media` object, which uses the batching system. So if an emoji is the first image in the room, it would cause that hitch.

This hack just adds an invisible image asset to the scene so that the batching shaders get initialized during room load.

@netpro2k Perhaps you have strong opinions about this hack.

PR #1629, opened 2019-08-09
Split record and snap buttons

gfodor:
The new camera UI is pretty confusing wrt knowing how to take a photo vs a video. This updates the UI to split those into two buttons.

### DIFF
--- a/src/components/camera-tool.js
+++ b/src/components/camera-tool.js
@@ -39,8 +39,8 @@ const CAPTURE_WIDTH = isMobileVR ? 640 : 1280;
 const CAPTURE_HEIGHT = isMobileVR ? 360 : 720;
 const RENDER_WIDTH = 1280;
 const RENDER_HEIGHT = 720;
-const CAPTURE_DURATIONS = allowVideo ? [0, 3, 7, 15, 30, 60, Infinity] : [0];
-const DEFAULT_CAPTURE_DURATION = allowVideo ? 3 : 0;
+const CAPTURE_DURATIONS = [3, 7, 15, 30, 60, Infinity];
+const DEFAULT_CAPTURE_DURATION = 3;
 const COUNTDOWN_DURATION = 3;
 const VIDEO_LOOPS = 3; // Number of times to loop the videos we spawn before stopping them (for perf)
 const MAX_DURATION_TO_LIMIT_LOOPS = 31; // Max duration for which we limit loops (eg GIFs vs long form videos)
@@ -155,7 +155,6 @@ AFRAME.registerComponent("camera-tool", {
       this.labelBackground = this.el.querySelector(".label-background");
       this.durationLabel = this.el.querySelector(".duration");
 
-      this.snapIcon = this.el.querySelector(".snap-icon");
       this.recordIcon = this.el.querySelector(".record-icon");
       this.recordAlphaIcon = this.el.querySelector(".record-alpha-icon");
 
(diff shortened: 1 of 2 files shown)


src/components/camera-tool.js
```diff
@@ -275,7 +277,7 @@ AFRAME.registerComponent("camera-tool", {
 
     const label = this.data.label;
     const isFrozen = this.el.sceneEl.is("frozen");
-    const hasDuration = this.data.captureDuration !== 0 && this.data.captureDuration !== Infinity;
+    const hasDuration = this.data.captureDuration !== Infinity;
     const isPhoto = this.data.captureDuration === 0;
```
johnshaughnessy: Looks like `this.data.captureDuration` won't be `0` anymore so this `isPhoto` check is suspicious.

src/components/camera-tool.js
```diff
@@ -39,8 +39,8 @@ const CAPTURE_WIDTH = isMobileVR ? 640 : 1280;
 const CAPTURE_HEIGHT = isMobileVR ? 360 : 720;
 const RENDER_WIDTH = 1280;
 const RENDER_HEIGHT = 720;
-const CAPTURE_DURATIONS = allowVideo ? [0, 3, 7, 15, 30, 60, Infinity] : [0];
-const DEFAULT_CAPTURE_DURATION = allowVideo ? 3 : 0;
+const CAPTURE_DURATIONS = [3, 7, 15, 30, 60, Infinity];
+const DEFAULT_CAPTURE_DURATION = 3;
 const COUNTDOWN_DURATION = 3;
```
johnshaughnessy: I suggest losing the countdown especially for photos. The delay between when I click to when it starts feels frustrating/unnecessary to me and it prevents quick/candid capture of dynamic events (snapping while a video is on a particular frame, snapping while something is moving, snapping while people have configured themselves in a particular way). Might be useful as a user pref. Otherwise I suggest we gather user feedback to get evidence of one or the other if we disagree.

src/components/camera-tool.js
```diff
@@ -295,14 +297,14 @@ AFRAME.registerComponent("camera-tool", {
...
-    this.snapIcon.object3D.visible = isPhoto && !isFrozen;
+    this.durationLabel.object3D.visible = hasDuration && !this.data.isSnapping && !isFrozen && allowVideo;
+    this.recordIcon.object3D.visible = !hasDuration && !isPhoto && !isFrozen && allowVideo;
+    this.recordAlphaIcon.object3D.visible = hasDuration && !isPhoto && !isFrozen && allowVideo;
     this.snapButton.object3D.visible = !this.data.isSnapping && !isFrozen;
+    this.recordButton.object3D.visible = !this.data.isSnapping && !isFrozen && allowVideo;
     this.cancelButton.object3D.visible = this.data.isSnapping && !isFrozen;
     this.prevDurationButton.object3D.visible = this.nextDurationButton.object3D.visible =
-      !this.data.isSnapping && allowVideo && !isFrozen;
+      !this.data.isSnapping && allowVideo && !isFrozen && allowVideo;
```
johnshaughnessy: Can't see countdown label from behind the camera, so it seems like nothing happened after clicking the photo/video buttons.

johnshaughnessy: (Problem is not specifically at this line)

src/components/camera-tool.js
gfodor: the reason i added the countdown is because there is no way to pose for a picture, since you need to be using your hand to click the button. right now, the way to take a point in time photo is to point-and-shoot. i guess i could add an option for the timer in the menu tho, like the audio toggle.

johnshaughnessy: Ah, that makes sense. Would be nice to have both options, but not worth blocking this PR if it's difficult

src/components/camera-tool.js
gfodor: fixed

PR #1630, opened 2019-08-09
Bug/permissions error

brianpeiris:
I couldn't reproduce the conditions that caused this, but I assume network latency could cause a race condition where we receive a NAF message before we receive the presence state for a user that has recently joined a room. This change just guards against that.


gfodor:
seems sane to me, might be nice to buffer it but since this is a manipulation message the downside risk seems low it'll materialize in some kind of de-sync. (main case I can think of is if we get a video play/pause/seek message before presence)

brianpeiris:
Yeah, I agree it seems low impact to ignore these. Not so sure it makes sense to buffer either, since we may end up replaying messages out of order.

PR #1631, opened 2019-08-09
Reintroduce limits for cursor move job in touchscreen

johnshaughnessy:
This fixes a regression made in #1614 where pinned media gets touch precedence over camera movement. This includes a fix for the driving issue #1421 by adding the `interactable` class to the avatar hover menu.


gfodor:
this seems better!

PR #1641, opened 2019-08-10
Ux/inspect in vr

johnshaughnessy:
This PR improves the UX of inspect button in the following ways:

- We use camera layers to render the inspected object in an isolated environment (a gray sphere, for now).
- When an inspected object has an audio node, we move the audio node to the exact spot where the audio listener is.

You cannot yet inspect the avatars of other users. (I ran into some complications while adding an inspect button to them.)


gfodor:
As part of this, think we can add the B button binding to inspect mode, and maybe right click-and-hold if it's easy? (https://github.com/mozilla/hubs/issues/1650)

src/hub.html
```diff
@@ -1054,7 +1054,9 @@
                 rotation
                 set-yxz-order
                 pitch-yaw-rotator
-                matrix-auto-update></a-entity>
+                matrix-auto-update>
+                <a-sphere id="inspect-sphere"></a-sphere>
```
gfodor: i know we were burned by using primitives before in the teleporter stuff, is there any reason to be concerned here about poly count etc?

src/systems/camera-system.js
```diff
@@ -86,33 +132,79 @@ export class CameraSystem {
 
   inspect(o) {
     if (this.mode !== CAMERA_MODE_INSPECT) {
-      this.preInspectMode = this.mode;
+      this.snapshot.mode = this.mode;
     }
     this.mode = CAMERA_MODE_INSPECT;
     this.inspected = o;
+
+    (getBatch(o, this.batchManagerSystem) || o).traverse(enableInspectLayer);
```
gfodor: i'm confused by this, won't this put everything in the batch into the layer, not just the inspected object? i'm not familiar with the batching code

gfodor: i guess in practice it doesn't do that, but can you explain why if it's obvious?

gfodor:
can we try just disabling rolloff attentuation on the audio source instead of moving audio listeners around? that would be better for two reasons: it'll be less subject to weird artifacts or bugs, and it will also allow you to be able to still hear nearby conversations to your avatar.

src/systems/camera-system.js
```diff
@@ -86,33 +132,79 @@ export class CameraSystem {
...
+      this.inspected = null;
+    }
     if (this.mode !== CAMERA_MODE_INSPECT) return;
-    this.mode = this.preInspectMode || CAMERA_MODE_FIRST_PERSON;
-    this.preInspectMode = null;
+    this.mode = this.snapshot.mode || CAMERA_MODE_FIRST_PERSON;
+    this.snapshot.mode = null;
+
+    if (this.snapshot.audio) {
+      setMatrixWorld(this.snapshot.audio, this.snapshot.audioTransform);
```
gfodor: i suggested we switch to using rolloff in my other comments, but if we stick with this, you should dereference this.snapshot.audio so we don't pin it to memory if the object is removed.

src/systems/camera-system.js
```diff
@@ -86,33 +132,79 @@ export class CameraSystem {
...
+    }
+    const vrMode = AFRAME.scenes[0].is("vr-mode");
+    const scene = AFRAME.scenes[0];
+    const camera = vrMode ? scene.renderer.vr.getCamera(scene.camera) : scene.camera;
+    camera.layers.mask = this.snapshot.mask;
+    if (vrMode) {
+      camera.cameras[0].layers.mask = this.snapshot.mask0;
+      camera.cameras[1].layers.mask = this.snapshot.mask1;
+    }
+    if (this.inspected) {
```
gfodor: this check doesn't seem necessary

src/systems/camera-system.js
```diff
@@ -86,33 +132,79 @@ export class CameraSystem {
...
+
   tick = (function() {
     const m2 = new THREE.Matrix4();
     const m3 = new THREE.Matrix4();
     const offset = new THREE.Vector3();
-    return function tick() {
+    return function tick(scene) {
+      if (!scene.is("entered")) return;
       this.playerHead = this.playerHead || document.getElementById("avatar-head");
-      if (!AFRAME.scenes[0].is("entered")) return;
```
gfodor: this `tick` method is pretty bulky and does a bunch of work -- we should at least extract the common subexpressions (like the raw rotator) and avoid necessary work. have you tried profiling this?

gfodor: also for the common/perf-sensitive case of 1st person, VR mode, is it possible to avoid some of the matrix copies? it seems like we copy a matrix to rigEl needlessly since we can just read the avatar rig matrix in 1st person mode, and it's a bit unclear to me why we need avatarPOV in VR mode.

gfodor: nevermind re: rig copy, i guess we now need to do that because camera is on that viewing rig

gfodor: i took a quick look and the profile doesn't jump out or anything. but would be nice to eliminate some of this work esp if it's cuttable in VR mode

src/systems/camera-system.js
gfodor: oh, i see, you clobber the matrix :)

src/systems/render-manager/hubs-batch-raw-uniform-group.js
```diff
@@ -60,7 +69,11 @@ export default class HubsBatchRawUniformGroup extends BatchRawUniformGroup {
       // TODO need to account for nested visibility deeper than 1 level
       this.setInstanceTransform(
         instanceId,
-        mesh.visible && (mesh.parent && mesh.parent.visible) ? mesh.matrixWorld : HIDE_MATRIX
+        mesh.visible &&
+        (mesh.parent && mesh.parent.visible) &&
+        (!cameraSystem.inspected || inParentHierarchyOf(cameraSystem.inspected, mesh))
```
gfodor: this is a hotspot, so i would push `!!cameraSystem.inspected` into a `const` boolean variable above to ensure this expression doesn't need to deref `cameraSystem` on every iteration and also have a mixture of types in the expression.

you also might want to just consider splitting it into two branches, to avoid evaluating the ternary operator on every instance.

src/systems/camera-system.js
johnshaughnessy: Yes -- Made it hide all the other batched objects by setting their (batched) transforms

src/hub.html
johnshaughnessy: Yes I want to replace this with a specific kind of "inspect mode environment object" that we like. At the very least, I'll create the THREE.js object directly instead of using this primitive.

src/systems/camera-system.js
johnshaughnessy: :+1:

src/systems/camera-system.js
johnshaughnessy: :+1:

src/systems/render-manager/hubs-batch-raw-uniform-group.js
johnshaughnessy: :+1:

src/systems/camera-system.js
johnshaughnessy: I'll simplify it a bit, but I'm surprised that you find this an expensive `tick`. We can probably get rid of a matrix copy somehow but I imagined this as a tiny blip compared to e.g. copying all the matrices of physics objects to and from the physics world every frame.

The most expensive work is done when first inspecting an object, when we calculate the bounding box, viewing transform, and toggle the camera layers. Besides that is it the number of conditionals that is worrying?

src/systems/render-manager/hubs-batch-raw-uniform-group.js
johnshaughnessy: Ok I'll pull the dereference out of the loop, and I can find which `mesh` in the batch will be affected once when transitioning to inspect mode and keep that state around, rather than checking `inParentHierarchyOf` for every object in the batch each frame.

src/systems/render-manager/hubs-batch-raw-uniform-group.js
```diff
@@ -71,7 +65,7 @@ export default class HubsBatchRawUniformGroup extends BatchRawUniformGroup {
         instanceId,
         mesh.visible &&
         (mesh.parent && mesh.parent.visible) &&
-        (!cameraSystem.inspected || inParentHierarchyOf(cameraSystem.inspected, mesh))
+        (!inspected || inspectedMeshesFromBatch.indexOf(mesh) !== -1)
```
gfodor: depending on how hot the loop is here, it may be smarter to put a top level branch like:
```
if (!inspected) {
  // Common code path, with mesh.matrixWorld and no conditional on inspectedMeshesFromBatch
} else {
  // the code you have now (minus any redundant checking on inspected
}
```
since most of the time the first condition will be hit, skipping the array walks through inspectedMeshesFromBatch etc

johnshaughnessy:
> As part of this, think we can add the B button binding to inspect mode, and maybe right click-and-hold if it's easy? (#1650)

This is a good idea.  I haven't added it in yet but we can do it the same way we do other hoverables so that B becomes the inspect button when your cursor is hovering on something inspectable.

PR #1644, opened 2019-08-12
Make sure you can't overshoot and flip your head

johnshaughnessy:
Following up on a bug partially fixed in https://github.com/mozilla/hubs/pull/1639

(We use the [dot product](https://en.wikipedia.org/wiki/Dot_product) to solve this issue.)
image


misslivirose:
🙃

Issue #1437, opened 2019-06-10
Discord landing page should have preview video surfaced


misslivirose:
It would help provide more context to the bot and the capabilities on the landing page
https://hubs.mozilla.com/discord - Landing Page
https://www.youtube.com/watch?v=5HtRJolThZ8 - Video Introduction

PR #1651, opened 2019-08-13
First pass at replacing Discord splash with video files

misslivirose:
Adds the video files to the Discord landing page instead of the current splash screen to better capture the features that the bot enables. I don't love the styling on this so definitely open to suggestions. I'm a bit rusty. 🙂

Implements #1437


misslivirose:
Thanks for reviewing @brianpeiris ! I agree it would be nice to have on mobile, I think Spoke does that.

PR #1657, opened 2019-08-14
Fix avatar views not loading in admin and expose tag editing

netpro2k:
We started using internationalization stuff in the avatar preview, but didn't update the admin panel for these so the avatar pages in admin fail to load. Also exposes tag editing in the avatar listings edit view and links to the avatar landing pages instead of the API response when opening an avatar.


gfodor:
lgtm!

PR #1664, opened 2019-08-16
Load assets locally instead of from S3

brianpeiris:
The pen, duck and loading environment models were previously being loaded from a couple of S3 buckets when they didn't need to be. This PR just moves them to local assets.


src/utils/resolveURL.js
```diff
@@ -1,25 +0,0 @@
...
-  return path + url;
-}
-
-export function extractUrlBase(url) {
-  const index = url.lastIndexOf("/");
-
-  if (index === -1) return "./";
-
-  return url.substr(0, index + 1);
-}
```
brianpeiris: This code was unused.

src/utils/url-utils.js
```diff
@@ -0,0 +1,6 @@
+export function getAbsoluteUrl(baseUrl, relativeUrl) {
+  return new URL(relativeUrl, baseUrl);
+}
+export function getAbsoluteHref(baseUrl, relativeUrl) {
+  return getAbsoluteUrl(baseUrl, relativeUrl).href;
+}
```
gfodor: perhaps these should go into `media-url-utils`?

src/hub.html
```diff
@@ -359,7 +361,7 @@
                 <a-entity
                     class="pen interactable"
                     body-helper="type: dynamic; mass: 0.001; collisionFilterGroup: 1; collisionFilterMask: 15; gravity: 0 0 0;"
-                    media-loader="src: https://asset-bundles-prod.reticulum.io/interactables/DrawingPen/DrawingPen-34fb4aee27.gltf;"
+                    gltf-model-plus="src: #drawing-pen;"
```
gfodor: interesting, doesn't this change the behavior a bunch? (no loading sound, cube, spawn animation, etc?) probably for the better if it feels good.

brianpeiris: Yeah, that was my thinking as well, it does feel snappier without the loading and the unnecessary URL resolution.

PR #1666, opened 2019-08-19
Fix up a few UX issues around send to device and entry

gfodor:
This fixes up a few things that have been bugging me around UX and also in light of a quick user test I did:

- Change the copy from "Send to Device" to "Enter on a Device" to reduce confusion around "Send" of the chat message button

- Increases the height slightly of the "Choose Device" panel to reduce user-blindness of the lower panel after clicking "Enter Room"

- Update the chat send message icon to be less metaphorical and closer to what you see on other chat services

- After sending over a letter link code, exit the session on the sending machine (reasoning: if you send to device, you probably don't want to hear an echo of the room coming from the original device, yourself still in the lobby there, etc.)

- Adds a bunch of incidental cruft needed to properly add custom svg icons that can be colorized via CSS like fontawesome


src/react-components/in-world-chat-box.js
```diff
@@ -131,11 +131,8 @@ class InWorldChatBox extends Component {
               styles.messageEntryButtonInRoom,
               styles.messageEntrySubmit
             ])}
-          >
-            <i>
-              <FontAwesomeIcon icon={faPaperPlane} />
-            </i>
-          </button>
+            dangerouslySetInnerHTML={{ __html: sendMessageIcon }}
```
brianpeiris: If we start using this method more often for icons, it would be nice to wrap it in a small react component so that we don't litter `dangerouslySetInnerHTML` all over the code base.

PR #1676, opened 2019-08-20
Fix a variety of HLS live stream playback issues

gfodor:
This fixes or works around a bunch of gnarly HLS playback issues:

- There can be a delay between the video starting to play back and the time we determine the video is a live stream. For example, in Oculus Browser, we don't actually see the HLS level load until we get a user gesture. As such we should avoid setting `currentTime` (which defaults to 0 because of `this.data.time` defaulting to zero) until we know for sure if the video is live or not.

- Fixes an issue where in Oculus Browser, `LEVEL_SWITCHED` is never fired for live streams, so we do not ever tag the stream as such.

- Avoid flushing time stamps over the network until we've determine that the video is not a live stream. We *never* want time syncing with live streams.

- Disable the pause button for live streams, which leads to all kinds of confusing semantics.

- Partially work around a bug in firefox reality causing HLS seeking to fail.  

- Deal with potential latencies introduced by the need for a user gesture to begin playback by seeking to the end of the stream once we determine that it is a live stream.

### DIFF
--- a/src/components/media-views.js
+++ b/src/components/media-views.js
@@ -18,6 +18,10 @@ const ONCE_TRUE = { once: true };
 const TYPE_IMG_PNG = { type: "image/png" };
 const parseGIF = promisifyWorker(new GIFWorker());
 
+const isIOS = AFRAME.utils.device.isIOS();
+const isMobileVR = AFRAME.utils.device.isMobileVR();
+const isFirefoxReality = isMobileVR && navigator.userAgent.match(/Firefox/);
+
 export const VOLUME_LABELS = [];
 for (let i = 0; i <= 20; i++) {
   let s = "|";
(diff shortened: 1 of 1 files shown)


src/components/media-views.js
```diff
@@ -582,9 +588,22 @@ AFRAME.registerComponent("media-video", {
...
-          this.updateHoverMenuBasedOnLiveState();
+          if (texture.hls.currentLevel >= 0) {
+            const videoWasLive = !!this.videoIsLive;
+            this.videoIsLive = texture.hls.levels[texture.hls.currentLevel].details.live;
+            this.updateHoverMenuBasedOnLiveState();
+
+            if (!videoWasLive && this.videoIsLive) {
+              // We just determined the video is live (there can be a delay due to autoplay issues, etc)
+              // so catch it up to HEAD.
+              if (!isFirefoxReality) {
```
netpro2k: yuck

gfodor: this is fixed in 1.4

PR #1678, opened 2019-08-20
Add soft room capping

gfodor:
Adds a soft room cap of 24 users who are either in the room or going through the entry flow. This is not server enforced, so could be bypassed somewhat easily by a user, but it should help with reducing the room capacity somewhat until a server-authoritative approach can be incorporated. Goes with: https://github.com/mozilla/reticulum/pull/218

When the room is full, the landing page CTA switches to a "Watch from Lobby" button which has the same effect as the X.

### DIFF
--- a/src/hub.js
+++ b/src/hub.js
@@ -156,7 +156,7 @@ const NOISY_OCCUPANT_COUNT = 12; // Above this # of occupants, we stop posting j
 
 // Maximum number of people in the room/entering before users are forced to observer mode.
 // Eventually this should be moved to a room setting.
-const MAX_OCCUPIED_ROOM_ENTRY_SLOTS = 1;
+const MAX_OCCUPIED_ROOM_ENTRY_SLOTS = 24;
 
 const qs = new URLSearchParams(location.search);
 const isMobile = AFRAME.utils.device.isMobile();
--- a/src/hub.js
+++ b/src/hub.js
@@ -154,6 +154,10 @@ const mediaSearchStore = window.APP.mediaSearchStore;
 const OAUTH_FLOW_PERMS_TOKEN_KEY = "ret-oauth-flow-perms-token";
 const NOISY_OCCUPANT_COUNT = 12; // Above this # of occupants, we stop posting join/leaves/renames
 
+// Maximum number of people in the room/entering before users are forced to observer mode.
+// Eventually this should be moved to a room setting.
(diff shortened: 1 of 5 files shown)


src/react-components/ui-root.js
```diff
@@ -1043,44 +1043,62 @@ class UIRoot extends Component {
...
+            </div>
+          )}
+        {this.props.entryDisallowed &&
+          !this.state.waitingOnAudio && (
+            <div className={entryStyles.buttonContainer}>
               <a
                 onClick={e => {
                   e.preventDefault();
-                  this.handleForceEntry();
+                  this.setState({ watching: true });
```
brianpeiris: Was this `watching` state meant to be used somewhere?

src/hub.js
```diff
@@ -154,6 +154,10 @@ const mediaSearchStore = window.APP.mediaSearchStore;
 const OAUTH_FLOW_PERMS_TOKEN_KEY = "ret-oauth-flow-perms-token";
 const NOISY_OCCUPANT_COUNT = 12; // Above this # of occupants, we stop posting join/leaves/renames
 
+// Maximum number of people in the room/entering before users are forced to observer mode.
+// Eventually this should be moved to a room setting.
+const MAX_OCCUPIED_ROOM_ENTRY_SLOTS = 24;
```
brianpeiris: Would it be useful to make this optionally configurable with a query string param, for debugging at least?

src/react-components/ui-root.js
gfodor: yeah this is used to push you into the observer lobby mode, which is also triggered by the X in the panel

src/hub.js
gfodor: good call

src/react-components/ui-root.js
brianpeiris: Ah, okay, I just thought it was a new state variable 👍

src/hub.js
gfodor: on second thought, I think this might be a bad idea since it makes it trivial to bypass this check. not going to argue this check is the most robust thing in the world, but having it be reduced down to 'someone can just share a URL to bypass it' seems excessively weak.

gfodor: i think insofar as we get complaints about this, we should escalate pushing this into a room setting. i was going to do that for v1 but didn't want to get bogged down since I'd like to get this out ahead of tomorrow's event.

PR #1680, opened 2019-08-20
Fix env map error on loading cube

brianpeiris:
Fix for https://sentry.prod.mozaws.net/operations/hubs-prod/issues/6099086/

Couldn't repro this manually, but I imagine it happens if someone in a room spawns something while you're in the middle of initializing your scene.


gfodor:
oh, maybe it's the pinned objects loading or something. we def show the loading cube for those. PR lgtm.

PR #1684, opened 2019-08-21
networked pen laser

InfiniteLee:
Network the laser the laser that projects from the pen when in projection-drawing mode.

pen-laser


src/components/tools/pen.js
```diff
@@ -256,106 +276,114 @@ AFRAME.registerComponent("pen", {
...
 
-      if (!almostEquals(0.005, this.worldPosition, this.lastPosition)) {
-        this.direction.subVectors(this.worldPosition, this.lastPosition).normalize();
-        this.lastPosition.copy(this.worldPosition);
+      laserEndPosition.copy(intersection.point);
+
+      if (this.el.sceneEl.is("vr-mode")) {
+        remoteLaserOrigin.copy(laserStartPosition);
+      } else {
+        camerWorldPosition.copy(this.data.camera.object3D.position);
```
johnshaughnessy: Looks like `cameraWorldPosition` is immediately overwritten in the next line.

src/utils/permissions-utils.js
```diff
@@ -80,7 +80,7 @@ function authorizeEntityManipulation(entity, sender, senderPermissions) {
   } else if (template.endsWith("-camera")) {
     return isCreator || senderPermissions.spawn_camera;
   } else if (template.endsWith("-pen")) {
-    return isCreator || senderPermissions.spawn_pen;
+    return isCreator || senderPermissions.spawn_drawing;
```
johnshaughnessy: Did the name of this change elsewhere?

InfiniteLee: no, this was incorrectly named in the first place, and was causing ownership transfer to get busted with pens.

src/components/tools/pen.js
```diff
@@ -185,68 +167,106 @@ AFRAME.registerComponent("pen", {
...
+            ? AFRAME.scenes[0].systems.userinput.get(pathsMap.cursor.pose)
+            : null;
+
+        const intersection = this._getIntersection(cursorPose);
+
+        this._updatePenTip(intersection);
+
+        const remoteLaserVisible = this.data.drawMode === DRAW_MODE.PROJECTION && !!intersection;
+        const laserVisible = this.el.sceneEl.is("vr-mode") && !!intersection && remoteLaserVisible;
+        this.el.setAttribute("pen-laser", {
```
johnshaughnessy: May be good to check whether these attributes changed, since calling `setAttribute` is expensive (so we want to avoid doing it in `tick` whenever possible)

src/components/tools/pen.js
```diff
@@ -185,68 +167,106 @@ AFRAME.registerComponent("pen", {
...
+          remoteLaserVisible: remoteLaserVisible
+        });
+
+        if (remoteLaserVisible || laserVisible) {
+          this._updateLaser(cursorPose, intersection);
         }
+
+        const penVisible = this.grabberId !== "cursor" || !intersection;
+        this._setPenVisible(penVisible);
+        this.el.setAttribute("pen", { penVisible: penVisible });
```
johnshaughnessy: similarly here - check if this has changed to avoid calling `setAttribute` whenever possible. also consider using the `clobber` option in order to avoid creating a new object each time

johnshaughnessy:
Overall would be good to reduce `setAttribute` calls (and use the optimized path to avoid creating new objects to pass in each time)

src/components/tools/pen.js
InfiniteLee: Ah yea, i think that used to be `worldToLocal`, thanks

PR #1692, opened 2019-08-22
Hacky fix for safari WebRTC audio

gfodor:
After a full day's worth of investigating, here's what I've uncovered about iOS Safari:

- If you join a room and there are avatars already in the room, there's a non-deterministic chance that you will have one avatar whose audio you will not hear. You will always hear the audio from the other avatars besides that one.

- Additionally, if you are in a room by yourself, the first avatar to join the room will also non-deterministically not be heard.

- When debugging this state, the track and everything is wired up right in the client, but there is just no audio data being received over the track.

- I don't have a solution to the problem seen when joining a room with avatars already present. However, this PR addresses the case when you are in a room by yourself and the first other user joins. It looks like the Safari behavior can be mitigated by ensuring that when a user joins as a peer in Janus, that the Safari peers in the room immediately receive audio over their track. If they do not, then there's a non-deterministic chance the track will die.

This PR mitigates the issue by sending a brief tone over the audio track when joining the lobby (!). This fully mitigates the issue for Safari when you are in a room by yourself. Without this patch, about 80% of the time the first avatar to join you will not be heard. After this patch, they'll be heard 100% of the time!

### DIFF
--- a/src/hub.js
+++ b/src/hub.js
@@ -1266,9 +1266,41 @@ document.addEventListener("DOMContentLoaded", async () => {
       hubChannel.setPermissionsFromToken(permsToken);
 
       scene.addEventListener("adapter-ready", ({ detail: adapter }) => {
+        // HUGE HACK Safari does not like it if the first peer seen does not immediately
+        // send audio over its media stream. Otherwise, the stream doesn't work and stays
+        // silent. (Though subsequent peers work fine.)
+        //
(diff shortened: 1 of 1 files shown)


gfodor:
The "avatars already in room" case seems to be mitigated by this hack: https://github.com/mozilla/naf-janus-adapter/pull/78

joshmarinacci:
good research!

PR #1696, opened 2019-08-23
Add screen sharing to chrome and edge

gfodor:
Hilariously (or not), this is all it took to enable screen sharing in Chrome


gfodor:
this ain't ready yet. want to update it so the camera still has the "click to disable" behavior but make it so you can share multiple windows (which totally works with this API)

gfodor:
OK, this is now ready to go.

gfodor:
(ended up not being able to do multiple windows, because of Janus limitations)

PR #1713, opened 2019-08-30
Remove S3-based default avatars

brianpeiris:
This removes code related to "legacy" avatars in favor of choosing from a set of "default" avatars via the reticulum avatar search API.
If you have a legacy avatar set in your localstorage profile, it will error out on loading and be replaced by a random choice of default avatar.


src/hub.js
```diff
@@ -574,6 +572,8 @@ async function runBotMode(scene, entryManager) {
 }
 
 document.addEventListener("DOMContentLoaded", async () => {
+  await store.init();
```
netpro2k: Are we sure nothing else ends up trying to access store before this runs?

brianpeiris: Currently, init really just initializes the avatarId in the profile. I'll change the method name to make that more clear

brianpeiris: There might be other parts of the code that relies on the store before this happens, but afaict, they do not rely on the avatarId specifically

PR #1726, opened 2019-09-04
Fix pinned object physics

### LINKED ISSUE #1675: Can't move pinned images on Quest
**Description**
If you enter a room and there are pinned images or videos, you can't move them in pause mode on Quest (but works fine on PC VR)

**To Reproduce**
Create a room
Pin an image from image search
Reload
Try to move the image in paused mode on quest in oculus browser. Doesn't work.

**Hardware** 
Quest + Oculus Browser

**Additional context**
It looks like `state.held` is being set properly in the interaction system so the bug might be in setting up the constraints.

InfiniteLee:
Fixes #1675
bump three-ammo to fix issue with zero-mass kinematic bodies getting stuck as immovable; update floaty-object and pinnable to use "kinematic" instead of "static" because static objects should always remain static

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17339,7 +17339,7 @@
       "from": "github:mozillareality/three.js#hubs/master"
     },
     "three-ammo": {
-      "version": "github:infinitelee/three-ammo#c613987b7ae126fdb570f10477399a5cac4d70f6",
+      "version": "github:infinitelee/three-ammo#fe57353506b79627a5b2e9b0feba4a2fe52b745c",
       "from": "github:infinitelee/three-ammo#master"
     },
     "three-bmfont-text": {
--- a/src/components/floaty-object.js
+++ b/src/components/floaty-object.js
@@ -52,7 +52,7 @@ AFRAME.registerComponent("floaty-object", {
         this.bodyHelper.body.physicsBody.getAngularVelocity().length2() < angularThreshold * angularThreshold;
 
       if (isAtRest && isMine) {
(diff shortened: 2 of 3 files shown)


package-lock.json
```diff
@@ -17339,7 +17339,7 @@
       "from": "github:mozillareality/three.js#hubs/master"
     },
     "three-ammo": {
-      "version": "github:infinitelee/three-ammo#c613987b7ae126fdb570f10477399a5cac4d70f6",
+      "version": "github:infinitelee/three-ammo#fe57353506b79627a5b2e9b0feba4a2fe52b745c",
```
johnshaughnessy: https://github.com/InfiniteLee/three-ammo/commit/fe57353506b79627a5b2e9b0feba4a2fe52b745c#diff-f0c01727e5628483090afb5bc8c36be6L377

This is the change that affects the behavior we can see

src/components/floaty-object.js
```diff
@@ -52,7 +52,7 @@ AFRAME.registerComponent("floaty-object", {
         this.bodyHelper.body.physicsBody.getAngularVelocity().length2() < angularThreshold * angularThreshold;
 
       if (isAtRest && isMine) {
-        this.el.setAttribute("body-helper", { type: "static" });
+        this.el.setAttribute("body-helper", { type: "kinematic" });
```
johnshaughnessy: We expect this change not to have a noticeable effect for end-users, but is more in line with our current understanding of the physics object state

PR #1728, opened 2019-09-05
Allow remxing from avatar_listing landing pages and hover menues

netpro2k:
Adds remix buttons to the landing pages and hover menus for avatar listings. 
image

We probably want to expand this to avatars as well, but starting with listing to match what we have in other places we surface them.

### DIFF
--- a/src/avatar.js
+++ b/src/avatar.js
@@ -72,7 +72,7 @@ class AvatarUI extends React.Component {
     const { avatar } = this.state;
     this.setState({ copyMessage: "copying..." });
     await remixAvatar(avatar.avatar_id, avatar.name);
-    this.setState({ copyMessage: "Coppied!" });
+    this.setState({ copyMessage: "Copied!" });
     setTimeout(() => this.setState({ copyMessage: false }), 2000);
   };
 
--- a/src/avatar.js
+++ b/src/avatar.js
@@ -41,7 +41,7 @@ class AvatarUI extends React.Component {
 
   constructor(props) {
     super(props);
-    this.state = { copyMessage: false };
+    this.state = { copyMessage: null };
   }
 
   componentDidMount() {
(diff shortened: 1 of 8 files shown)


src/avatar.js
```diff
@@ -64,8 +67,17 @@ class AvatarUI extends React.Component {
     });
   };
 
-  render() {
+  handleCopyAvatar = async e => {
+    e.preventDefault();
     const { avatar } = this.state;
+    this.setState({ copyMessage: "copying..." });
+    await remixAvatar(avatar.avatar_id, avatar.name);
+    this.setState({ copyMessage: "Coppied!" });
```
brianpeiris:
```suggestion
    this.setState({ copyMessage: "Copied!" });
```

src/avatar.js
```diff
@@ -38,7 +41,7 @@ class AvatarUI extends React.Component {
 
   constructor(props) {
     super(props);
-    this.state = {};
+    this.state = { copyMessage: false };
```
brianpeiris: nit: Bit odd that this value starts as a boolean and changes to a string. Maybe just use `null` for the initial value? Same one line 76 below.

src/components/remix-avatar-button.js
```diff
@@ -0,0 +1,54 @@
...
+    this.onClick = async () => {
+      if (this.copying || !this.avatar) return;
+
+      try {
+        this.copying = true;
+        this.label.setAttribute("text", "value", "Copying...");
+
+        await remixAvatar(this.avatar.avatar_id, this.avatar.name);
+
+        this.label.setAttribute("text", "value", "Coppied!");
```
brianpeiris:
```suggestion
        this.label.setAttribute("text", "value", "Copied!");
```

PR #1730, opened 2019-09-05
Drawing Menu

InfiniteLee:
drawing-menu

Adds a pause-mode menu when hovering your cursor over drawings. Gives anyone the ability to delete drawings, and exposes the undo function for the creator of the drawing.
This is the first pass on what will hopefully be several other options for interacting with drawings.


src/systems/drawing-menu-system.js
```diff
@@ -0,0 +1,200 @@
...
+        this.populateEntities(this.targets);
+        this.dirty = false;
+      }
+
+      const hovered = this.getHovered();
+      if (hovered) {
+        for (let i = 0; i < this.cursorControllers.length; i++) {
+          if (this.cursorControllers[i].components["cursor-controller"].intersection) {
+            const intersectionPoint = this.cursorControllers[i].components["cursor-controller"].intersection.point;
+            this.showMenu(hovered, intersectionPoint);
```
johnshaughnessy: Can this end up showing the menu twice, or on the wrong intersection point?

src/systems/drawing-menu-system.js
```diff
@@ -0,0 +1,200 @@
...
+            interaction.options[remote] &&
+            userinput.get(interaction.options[remote].grabPath) &&
+            hovered.components.tags &&
+            hovered.components.tags.data.singleActionButton &&
+            hovered.parentEl.parentEl.components["networked-drawing"]
+          ) {
+            if (hovered.classList.contains("undo-drawing")) {
+              hovered.parentEl.parentEl.components["networked-drawing"].undoDraw();
+            } else if (hovered.classList.contains("delete-drawing")) {
+              NAF.utils.takeOwnership(hovered.parentEl.parentEl);
```
johnshaughnessy: Having to crawl up the hierarchy here is fine I think, but might prefer to have a helper to build the association between the hovered entity and the networked entity, so that this code is a bit more robust to changes in the `html` structure. (e.g. if there was another child inserted between the networked el and the hovered el).

src/components/cursor-controller.js
```diff
@@ -69,7 +70,7 @@ AFRAME.registerComponent("cursor-controller", {
       }
 
       const interaction = AFRAME.scenes[0].systems.interaction;
-      let intersection;
+      this.intersection = null;
```
johnshaughnessy: Should this be moved up before the early out on line 69 (https://github.com/mozilla/hubs/pull/1730/files#diff-bafcf688d672a56b722fa967f5b16acbR69) ?

src/systems/drawing-menu-system.js
```diff
@@ -0,0 +1,200 @@
...
+          deleteButton.object3D.position.y = 0.125;
+          deleteButton.object3D.matrixNeedsUpdate = true;
+        }
+
+        const dist = cameraWorldPos.distanceTo(menu.object3D.position);
+        const finalScale = this.getMenuScale(dist);
+
+        menu.removeAttribute("animation__show");
+
+        menu.setAttribute("animation__show", {
```
johnshaughnessy: Nice I keep forgetting that we can add animations to the interactions. Would like to add juice like this more often!

src/systems/drawing-menu-system.js
```diff
@@ -0,0 +1,200 @@
...
+      const interaction = AFRAME.scenes[0].systems.interaction;
+      const userinput = AFRAME.scenes[0].systems.userinput;
+
+      for (let i = 0; i < remotes.length; i++) {
+        const remote = remotes[i];
+        if (interaction.state[remote]) {
+          const hovered = interaction.state[remote].hovered;
+          if (
+            hovered &&
+            interaction.options[remote] &&
```
johnshaughnessy: When does `interaction.options[remote]` yield a falsey value?

src/components/tools/drawing-manager.js
```diff
@@ -6,42 +6,45 @@
...
+        this.drawingEl.setAttribute("networked", "template: #interactable-drawing");
+        this.el.sceneEl.appendChild(this.drawingEl);
+
+        this.drawingEl.addEventListener(
+          "componentinitialized",
+          e => {
+            if (e.detail.name == "networked-drawing") {
+              this.drawing = this.drawingEl.components["networked-drawing"];
+              this.drawingToPen.set(this.drawing, pen);
+              resolve(this.drawing);
```
johnshaughnessy: Seems unfortunate that the async initialization affects things at this level, since this will be called for each start draw / stop draw. No specific suggestion here - but would be interested in thinking thru how we might refrain from `getDrawing` calls until the initialization is done so that it is synchronous.

src/components/tools/drawing-manager.js
```diff
@@ -6,42 +6,45 @@
...
 
   getDrawing(pen) {
     //TODO: future handling of multiple drawings
-    if (this.drawing && (!this.drawingToPen.has(this.drawing) || this.drawingToPen.get(this.drawing) === pen)) {
-      this.drawingToPen.set(this.drawing, pen);
-      return this.drawing;
-    }
-
-    return null;
+    return new Promise((resolve, reject) => {
```
johnshaughnessy: I know the TODO above was added before but -- any thoughts on how multiple drawings can be handled (in terms of technical requirements and UX)?

src/systems/drawing-menu-system.js
```diff
@@ -0,0 +1,200 @@
...
+    this.sceneEl.addEventListener("stateremoved", this.hideMenus.bind(this));
+
+    this.lastIntersection = new THREE.Vector3();
+
+    this.drawingMenus = [];
+    this.setDirty = this.setDirty.bind(this);
+    this.dirty = true;
+
+    // TODO: Use the MutationRecords passed into the callback function to determine added/removed nodes!
+    this.observer = new MutationObserver(this.setDirty);
```
johnshaughnessy: Do you want to be observing the whole scene? It seems like you only care when new `.drawing-menu`'s get added, so I wonder if they should have a component similar to `is-remote-hover-target` on it that adds itself to a collection in `init` and then removes itself in `remove`

src/systems/drawing-menu-system.js
johnshaughnessy: Oh, I think you need to pull master into this branch to test it with the left hand changes

src/components/tools/networked-drawing.js
```diff
@@ -471,6 +475,9 @@ AFRAME.registerComponent("networked-drawing", {
     this.currentPointCount = 0;
     this.lineStarted = false;
     this.drawStarted = false;
+
+    AFRAME.scenes[0].systems["hubs-systems"].cursorTargettingSystem.setDirty();
```
johnshaughnessy: Do you think we should add a second API here that only updates the cursorTargettingSystem's knowledge of this specific drawing?

src/systems/drawing-menu-system.js
InfiniteLee: since you can't have two cursors active at the same time (at least while in pause mode) this shouldn't be an issue, but even if you could, it would just move the menu to whichever cursor is second in the list (left-cursor I think), and only if that cursor is at least a certain distance away from the first cursor.

src/systems/drawing-menu-system.js
InfiniteLee: This was written before the left handed changes got merged

src/components/tools/drawing-manager.js
InfiniteLee: This is only a problem now because drawings can be deleted. Before, this initialization only happened once whenever you spawned a pen for the first time. 
Probably the best way to handle this long term would be if `networked-drawing` wasn't an a-frame component so `componentinitialized` wouldn't need to be waited for, but as it is now, that's the only sane way to actually have networking for something.

src/components/tools/drawing-manager.js
InfiniteLee: Haven't given it too much thought yet because I'm not sure we even need it. It depends on how future functionality around manipulating/pinning drawings pans out.

src/systems/drawing-menu-system.js
InfiniteLee: I opted against doing that because this runs in `tick` (though tbf, only while in pause-mode). As such, doing any kind of hierarchy search, while being more robust, will be significantly more expensive. If you have any ideas of how to do this and still be efficient, please let me know.

johnshaughnessy: I think if we do the search once we can re-use the result (rather than redo the search)

InfiniteLee: There's N number of drawings, where N is up to the number of users in the space. This system runs for all of them, not just one drawing.

src/components/tools/networked-drawing.js
InfiniteLee: Maybe, but turns out I don't actually need this line here at all.

gfodor:
I haven't read the PR, but we should make sure this menu is disabled on drawings other than your own, if you are not granted drawing permissions. (Specifically, if the owner of the room can draw, but you can't, it shouldn't be possible to remove/undo strokes on the owner's drawings.)

PR #1744, opened 2019-09-09
Replace S3 pen model with local asset, accounting for physics shape component for VR interactions

brianpeiris:
My [previous attempt](https://github.com/mozilla/hubs/pull/1664/files#diff-a796e11be3d0dfc0f893148f1086736dR364) to replace the pen S3 asset in #1664 broke grabbing in VR since the ammo shape wasn't being set correctly on the pen model. This PR fixes that by changing the GLB to include the `ammo-shape` hubs component (which maps to `shape-helper`) and setting the `gltf-model-plus` component to inflate the model correctly.


brianpeiris:
Took a loot at this today. AFAIK, I can't just add an `shape-helper` to the entity since the model needs to be loaded first. Since we only have a gltf mapping for `ammo-shape`, I'm going to ship this as is.

PR #1748, opened 2019-09-11
Add support for using morph targets for audio feedback

netpro2k:
Adds a `morph-audio-feedback` component that can be added to a mesh with morph targets. It will look for a `network-audio-analyser` in its parent and animate a morph target with `name` between `minValue` and `maxValue` based on the audio volume. This also attempts to consolidate all the different methods we had for trying to normalize the analyser output. We can further refine how we normalize, but at least we can do it in one spot.

### DIFF
--- a/src/components/audio-feedback.js
+++ b/src/components/audio-feedback.js
@@ -1,3 +1,5 @@
+import { findAncestorWithComponent } from "../utils/scene-graph";
+
 // This computation is expensive, so we run on at most one avatar per frame, including quiet avatars.
 // However if we detect an avatar is seen speaking (its volume is above DISABLE_AT_VOLUME_THRESHOLD)
 // then we continue analysis for at least DISABLE_GRACE_PERIOD_MS and disable doing it every frame if
@@ -44,7 +46,7 @@ AFRAME.registerComponent("networked-audio-analyser", {
       const ctx = THREE.AudioContext.getContext();
       this.analyser = ctx.createAnalyser();
       this.analyser.fftSize = 32;
-      this.levels = new Uint8Array(this.analyser.frequencyBinCount);
+      this.levels = new Uint8Array(this.analyser.fftSize);
       event.detail.soundSource.connect(this.analyser);
     });
   },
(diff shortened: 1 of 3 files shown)


src/components/audio-feedback.js
```diff
@@ -196,24 +267,8 @@ const SPRITE_NAMES = {
...
-            ? 3
-            : volume < max * 0.16
-              ? 4
-              : volume < max * 0.32
-                ? 5
-                : volume < max * 0.5
-                  ? 6
-                  : 7;
+export function micLevelForVolume(volume) {
+  return Math.ceil(THREE.Math.mapLinear(volume, 0, 1, 0, 7));
```
johnshaughnessy: Wonder if this will feel good enough or if it would feel better if either we hand-wrote this map (which I did before by trying different values until it "felt right" to me) or figuring out how to map the 0-1 in "raw volume" (or... "average square amplitude across all the levels") to "perceived volume"

src/components/audio-feedback.js
```diff
@@ -1,3 +1,5 @@
+import { findAncestorWithComponent } from "../utils/scene-graph";
```
johnshaughnessy: Wow, this utility function existed for a year before I noticed

src/components/audio-feedback.js
```diff
@@ -28,23 +30,51 @@ export function getAudioFeedbackScale(fromObject, toObject, minScale, maxScale,
...
 }
 
+function updateVolume(component) {
+  const newVolume = calculateVolume(component.analyser, component.levels);
+
+  component.loudest = Math.max(component.loudest, newVolume);
+  component.volume = component.loudest === 0 ? 0 : THREE.Math.mapLinear(newVolume, 0, component.loudest, 0, 1);
+
+  const s = 0.3;
+  component.volume = s * component.volume + (1 - s) * component.prevVolume;
```
johnshaughnessy: Is this lerp a noticeable improvement, and is 0.3 a fine-tuned value?  (I know this code existed before this PR, just wondering if you messed with this at all)

src/components/audio-feedback.js
```diff
@@ -28,23 +30,51 @@ export function getAudioFeedbackScale(fromObject, toObject, minScale, maxScale,
...
+  component.loudest *= 0.999;
+
+  component.prevVolume = component.volume;
+}
+
+// function updateVolume(component) {
```
johnshaughnessy: remove

src/components/audio-feedback.js
netpro2k: I think we should tune for that at the analyser level, since I think that's what we want everywhere we are reading that data.

src/components/audio-feedback.js
netpro2k: Did not mess with the value, but did notice it feels better to me with the lerp than without.

src/components/audio-feedback.js
netpro2k: Yep, just wanted to show a different method I tried as well, and maybe give you a chance to play with it.

Issue #1462, opened 2019-06-26
Add room object list to a Hubs room


flowpoly:
As an advanced user, I would like to see a list of all my objects in my scene so that I can delete/bring them to me if they fall under the boundaries.

kwadroke:
Maybe allow resetting of the objects' positions to a former state? Possibly a "Soft Pin"?

misslivirose:
What do folks think about linking this with #1461 in the same UI panel?

Other open questions: 
- What is the permission level that can access and use the object list? Given the object permissions (#1517) feature, I would think that we'd only want the object list to appear for people who can create objects in the room, but there's been some discussion about whether this should be an owner-only feature or not. 

- What kind of actions can be taken from inside the list? My take is delete / 'bring to me' / maybe 'pin'. 

- Where does the UI for this live? We've been thinking that it might combine with the presence list (so that becomes a record of who/what is in the room) but open to some other ideas.

misslivirose:
Another open question: What do we show in the list to make it useful? The raw url for the media probably wouldn't provide enough context, so we might want to combine the object list with @johnshaughnessy 's new inspect mode when you click on an object in the list to get a preview of what it is that object is linked to (as opposed to trying to do thumbnails).

flowpoly:
Hey all, I did some mocks for this feature: https://www.figma.com/file/MV6fCW3schGwXrfbtmVbyQxf/Hubs?node-id=27%3A184

johnshaughnessy:
@flowpoly 's mockup informed the work I'm doing in #1749 . Thanks a lot -- those visuals made it a lot easier to start the work!

There are some differences in my first-pass implementation due to a few factors: 
 - In order to keep a more consistent feel with the way the participants list entries work, I moved the operations you can do on each object into a modal dialog that pops up when clicking the row, rather than having each of the operations represented with small icons in the list.
 - I do not have nice names for the media objects in the room, so I truncate their URL in the list and then display them in full in the modal that pops up. 

The icons I used were taken from font-awesome for convenience but I'd be happy to swap them out for different ones. I see one icon in the mockup for 3D models and one for videos -- if we like those better I'll swap them in. We also want icons for images, pdfs, and unknown media types.

[...]

PR #1749, opened 2019-09-11
Basic object list

johnshaughnessy:
You can now review a list of media items that have been shared in the room.

Note -- I will push this behind a feature flag and will save the "What's New" label for when it's ready to go fully live.

image

image

image


johnshaughnessy:
- [x] - Put behind feature flag
- [x] - Inspect on hover.
- [x] - Pop up a modal on click.
- [x] - Delete object from the modal.
- [x] - Disallow both "participants" and "objects" lists being expanded at once
- [x] - Get a scrubbed version of the url to display
- [x] - Handle object permissions.
- [x] - Fix placement to float to the top right properly on different screen sizes.
- [x] - Replace icons for item type.
- [x] - Test on mobile and in VR.
- [x] - Trigger react updates when objects load.
- [x] - Give the objects list icon hover text.
- [x] - Allow pinning/unpinning.
- [ ] - Add option to jump to inspected item.
- [ ] - Display a screenshot of inspected object?

src/assets/stylesheets/info-dialog.scss
```diff
@@ -49,6 +62,10 @@
         margin: 10px 0;
         white-space: pre-wrap;
 
+        @media (max-width: 801px) {
```
johnshaughnessy: This width was chosen to include mobile and mobile vr browsers, but feels a bit arbitrary.

src/react-components/object-info-dialog.js
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  updatePinnedState() {
+    this.setState({ pinned: this.props.el.components["networked"].data.persistent });
+  }
+
+  pin() {
+    if (!NAF.utils.isMine(this.props.el) && !NAF.utils.takeOwnership(this.props.el)) return;
+    this.props.el.emit("pinned", { el: this.props.el });
+    this.props.el.setAttribute("pinnable", "pinned", true);
```
johnshaughnessy: This line is suspicious. The line above will pin the entity, but the UI on `pin-networked-object-button`s will not update. Not sure which codepath is endorsed for pinning/unpinning these entities.

johnshaughnessy:
See also https://github.com/mozilla/hubs/issues/1462

src/react-components/object-info-dialog.js
```diff
@@ -0,0 +1,97 @@
...
+      property: "scale",
+      dur: 200,
+      to: { x: 0.01, y: 0.01, z: 0.01 },
+      easing: "easeInQuad"
+    });
+
+    targetEl.addEventListener("animationcomplete", () => {
+      this.props.scene.systems["hubs-systems"].cameraSystem.uninspect();
+      NAF.utils.takeOwnership(targetEl);
+      targetEl.parentNode.removeChild(targetEl);
```
InfiniteLee: I think you might have a race condition here you might need to guard for: If someone else deletes this object before this animation completes,  targetEl.parentNode may not be set when this callback fires.

src/scene-entry-manager.js
```diff
@@ -226,7 +226,11 @@ export default class SceneEntryManager {
   };
 
   _signInAndPinOrUnpinElement = (el, pin) => {
-    const action = pin ? () => this._pinElement(el) : async () => await this._unpinElement(el);
+    const action = pin
+      ? () => this._pinElement(el)
+      : async () => {
+          await this._unpinElement(el);
+        };
```
InfiniteLee: This is a nit but I find that when these ternary operators become multi-line, you're better off just re-writing it to be not a ternary anymore, for the sake of readability

src/react-components/object-list.js
```diff
@@ -0,0 +1,181 @@
...
+  if (httpIndex !== -1 && split.length > httpIndex + 3) {
+    host = split[httpIndex + 2];
+    const hostSplit = host.split(".");
+    if (hostSplit.length > 1) {
+      lessHost = `${hostSplit[hostSplit.length - 2]}.${hostSplit[hostSplit.length - 1]}`;
+    }
+  }
+
+  return `${lessHost} ... ${resourceName.substr(0, 30)}`;
+}
```
InfiniteLee: I think this entire function would be easier to understand/maintain if you converted it into a regex instead of all these string splits.

src/react-components/object-info-dialog.js
```diff
@@ -0,0 +1,97 @@
...
+    targetEl.setAttribute("animation__remove", {
+      property: "scale",
+      dur: 200,
+      to: { x: 0.01, y: 0.01, z: 0.01 },
+      easing: "easeInQuad"
+    });
+
+    targetEl.addEventListener("animationcomplete", () => {
+      this.props.scene.systems["hubs-systems"].cameraSystem.uninspect();
+      NAF.utils.takeOwnership(targetEl);
```
InfiniteLee: It feels really odd to me that we would take ownership _after_ the animation. Ideally I think we would immediately take ownership and delete the object, but still somehow play the animation locally. Unfortunately I don't think we have an easy way to delete objects over the network but _not_ delete them locally.

johnshaughnessy: Here we take ownership before AND after the animation, copying this from `remove-networked-object-button`. (If taking ownership fails above, then we early out.) Perhaps though I should be checking that the `targetEl` still exists and I still have ownership (or can take ownership back) in the case that I leave pause mode during the animation and a delete message came across the network while I was paused.

src/react-components/object-info-dialog.js
johnshaughnessy: Yes I think you're right. Networked messages are paused until I leave the inspect mode, so the likelihood of this happening is very low, but it IS technically possible to quickly hit "delete" and then "cancel" to uninspect (thereby leaving pause mode) before the animation completes. I'll fix this.

src/scene-entry-manager.js
johnshaughnessy: ah sure. This was an accident of running `prettier` on this file -- not a change I thought about.

johnshaughnessy:
Wondering if we should just lose the URL / display name from the list:

image

Issue #1727, opened 2019-09-04
Using transform controls on pinned objects doesn't persist


gfodor:
**Description**
If you use the scale or rotate buttons on a pinned object, they work, but do not re-save the pinned object state when letting go.

**To Reproduce**
1. Pin an object
2. Scale or rotate it with the menu
3. Refresh
4. The rotation/scale has not been applied

**Expected behavior**
Rotate + scale should re-pin the object when completed (similar to moving it)

misslivirose:
This is probably closer to a P1.5 (I was torn between P1/2 here) but losing media because it isn't re-pinned after changing one of these properties is definitely a broken effect and I would guess the fix is fairly straightforward?

misslivirose:
@gfodor Did you happen to fix this already? I am not seeing this behavior right now

gfodor:
nope, didn't try to fix this.

johnshaughnessy:
Fixed in https://github.com/mozilla/hubs/pull/1757. Sorry for my slow.

PR #1757, opened 2019-09-17
Retain transformations of pinned objects

johnshaughnessy:
Fixes https://github.com/mozilla/hubs/issues/1727

### DIFF
--- a/src/components/pinnable.js
+++ b/src/components/pinnable.js
@@ -69,7 +69,9 @@ AFRAME.registerComponent("pinnable", {
   tick() {
     const held = this.isHeld(this.el);
 
+    let didFireThisFrame = false;
     if (!held && this.wasHeld) {
+      didFireThisFrame = true;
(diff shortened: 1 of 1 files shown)


src/components/pinnable.js
```diff
@@ -74,5 +74,12 @@ AFRAME.registerComponent("pinnable", {
     }
 
     this.wasHeld = held;
+
+    this.transformObjectSystem = this.transformObjectSystem || AFRAME.scenes[0].systems["transform-selected-object"];
+    const transforming = this.transformObjectSystem.transforming && this.transformObjectSystem.target.el === this.el;
+    if (!transforming && this.wasTransforming) {
+      this._fireEventsAndAnimate(this.data, true);
```
netpro2k: are there cases where you were transforming the object and holding it,  and stop doing so at the same time? If so we would end up firing the events twice.. not sure if that's an issue.

Issue #1740, opened 2019-09-06
Adjust move speed with mouse wheel


johnshaughnessy:
Adjust move speed with mouse wheel (like the feature in Blender)

PR #1762, opened 2019-09-18
Adjust move speed with mouse wheel

johnshaughnessy:
Allows you to adjust your move speed with the mouse wheel when you add the `wheelspeed` (or `wheelSpeed` or `ws`) URL query string parameter.

Resolves https://github.com/mozilla/hubs/issues/1740

### DIFF
--- a/src/components/character-controller.js
+++ b/src/components/character-controller.js
@@ -1,5 +1,7 @@
 import { paths } from "../systems/userinput/paths";
 import { SOUND_SNAP_ROTATE } from "../systems/sound-effects-system";
+import qsTruthy from "../utils/qs_truthy";
+const enableWheelSpeed = qsTruthy("wheelSpeed") || qsTruthy("wheelspeed") || qsTruthy("ws");
 const CLAMP_VELOCITY = 0.01;
 const MAX_DELTA = 0.2;
 const EPS = 10e-6;
@@ -24,6 +26,7 @@ AFRAME.registerComponent("character-controller", {
   },
 
   init: function() {
+    this.charSpeed = 1;
     this.navGroup = null;
     this.navNode = null;
     this.velocity = new THREE.Vector3(0, 0, 0);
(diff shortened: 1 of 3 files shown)


johnshaughnessy:
This is an enhancement @j-conrad wanted to try out, inspired by the first person mode of blender.

src/components/character-controller.js
```diff
@@ -137,6 +140,10 @@ AFRAME.registerComponent("character-controller", {
       if (userinputAngularVelocity !== null && userinputAngularVelocity !== undefined) {
         this.angularVelocity = userinputAngularVelocity;
       }
+      const dCharSpeed = userinput.get(paths.actions.dCharSpeed) || 0;
+      if (enableWheelSpeed && dCharSpeed) {
```
netpro2k: Minor but I would move the check for the `enableWheelSpeed` to surround this whole new block. Then I am not sure the aCharSpeed check is needed, as it would be handled by the bounds check.

src/components/character-controller.js
```diff
@@ -137,6 +140,10 @@ AFRAME.registerComponent("character-controller", {
...
+        this.charSpeed = Math.min(Math.max(this.charSpeed + this.charSpeed * dCharSpeed, 0.1), 5);
```
netpro2k: `THREE.Math.clamp` is a bit more readable than `min(max())`

src/systems/userinput/bindings/keyboard-mouse-user.js
```diff
@@ -68,6 +69,16 @@ export const keyboardMouseUserBindings = addSetsToBindings({
       dest: { value: paths.actions.characterAcceleration },
       xform: xforms.normalize_vec2
     },
+    {
+      src: { value: paths.device.mouse.wheel },
+      dest: { value: dCharSpeed },
```
netpro2k: Couldn't this just write directly to `paths.actions.dCharSpeed` instead of using an intermediate variable path?

PR #1764, opened 2019-09-20
Enable Object List

johnshaughnessy:
You can now view a list of the media items that have been shared to the room.
image
image

Clicking an item in the list opens a dialog box where you can show/hide the background and pin/unpin/delete the item. 

image

image


src/systems/camera-system.js
```diff
@@ -35,7 +35,7 @@ const calculateViewingDistance = (function() {
     const l2 = Math.abs((halfXExtents * margin) / Math.tan(halfHorFOV));
     const l3 = Math.abs(box.max.z - center.z) + Math.max(l1, l2);
     const l = object.el.sceneEl.is("vr-mode") ? Math.max(0.25, l3) : l3;
-    return l;
+    return l || 1.25;
```
InfiniteLee: wut

PR #1765, opened 2019-09-20
Fix issue with UV tiling not working on mobile

### LINKED ISSUE #1763: Mobile Standard Shader is not supporting tiling textures properly
The same **.glb** looks different on desktop vs. mobile/low quality. (attached)
[TiledUVs.zip](https://github.com/mozilla/hubs/files/3633186/TiledUVs.zip)

I've included a Hubs room to illustrate the difference.
https://hubs.mozilla.com/JejN3su/uv-tiling-issue/
image
image
image

netpro2k:
fixes #1763 

Note this only updates the uniforms once when the Material is created. We still need to do more work to support updating these properties at runtime, which we currently don't do, but might like to someday


gfodor:
@netpro2k going through old PRs, should this be reviewed + pushed?

PR #1769, opened 2019-09-24
Adaptive resolution

johnshaughnessy:
This feature is experimental, and is enabled via the `ar` query string param.

When adaptive resolution is enabled, the following changes take effect:
 - While the player is moving, image quality is reduced to achieve higher framerate/responsiveness.
 - While the player is not moving, image quality is increased at the expense of framerate/responsiveness.
 - If the frame rate is low and the player has not moved for a long time, image quality is allowed to degrade significantly until the player starts to interact again. 
 - The `u` and `i` keys on the keyboard will decrease and increase resolution respectively, and disable automatic adjustments for a fixed amount of time (`USER_COMMAND_TIMEOUT_DURATION`).

In the video below, the resolution is adjusted via keyboard commands:

adaptive-resolution-gif


johnshaughnessy:
Definitely not ready for prime time. Closing for now.

PR #1770, opened 2019-09-24
Press f to fly

johnshaughnessy:
Add keybindings for easy toggling of fly mode for keyboard


gfodor:
On PTO, but def worried about furthering sim sickness due to smooth loco by
making this a hardware button — maybe we should gate this on the
preferences dialog which will allow users to opt-in to all stick locomotion?

johnshaughnessy:
Noted. I don't know how best to expose "flying" to Quest users, so I changed this PR to only include the keybinding for the keyboard.

davecraige:
This would be super helpful and would be in parity with Crypto Voxels commands.

PR #1771, opened 2019-09-24
Add push-to-talk support

johnshaughnessy:
Add push-to-talk support
TODO: 
- [ ] bug/incomplete: If PTT is enabled but the user is not using keyboard/mouse, they won't be able to toggle their mic.
- [ ] How to configure button?
- [ ] How to do on other controllers?


johnshaughnessy:
This ended up being more work than I anticipated. Closing for now in case anyone else wants to prioritize this in the meantime.

PR #1777, opened 2019-10-01
doc grammar fix

Siddharth1698:


johnshaughnessy:
I initially wrote this as a single word to keep it consistent with its use in the codebase. See its definition here: https://github.com/mozilla/hubs/blob/master/src/systems/userinput/userinput.js#L190  I'm open to changing it if it is clearer as two words. 

Relatedly, it's probably more correct for it to be camel case (`userInput`) or hyphenated (`user-input`) in the code, but I wrote `userinput` because `aframe` warns on capitalized component/system names (see https://github.com/MozillaReality/aframe/blob/hubs/master/src/core/component.js#L574) and I wanted to be able to write `scene.systems.userinput` instead of `scene.systems["user-input"]`.

gfodor:
thanks for this, but i think current copy is ok!

PR #1780, opened 2019-10-01
Revamp admin stuff

mqp:
**WIP, but I will merge soon because it's a big change that would suck to get out of date.**

I really packed it into this one:

1. Moved admin console into separate Node package in the `admin` directory. You should now run it with `npm ci` and `npm start` in that directory. It still depends on a few pieces of code in the Hubs package; most of them are basically reasonable dependencies that we should extract into shared library, except for the avatar preview thing which is a mess. (But after talking to team it sounds like everyone would be OK if the avatar preview went away, also.)
2. Reworked admin UI to support sections besides the content admin area, e.g. the new service configuration panels (not functional on our production -- for self-hosted stuff in future.)
3. Made a lot of small fixes to admin console based on @emclaren feedback.


gfodor:
nice one! this stuff is pseudo-dark, so won't dig into doing a huge review, will probably have more feedback once I start adding stuff to it. also it seems that we need to do a bunch of work around build + deploying this now

mqp:
Yeah, it's all just dark low-risk admin console stuff and will also change a lot still in the next few days, so not worth thinking too hard about details yet.

Issue #1787, opened 2019-10-04
"Open Link" goes haywire on Oculus Browser


gfodor:
**Description**
When clicking a link in oculus browser in VR (via Open Link) the browser opens the link in another tab, causing hubs to crash/draw slowly.

**To Reproduce**
Go in on Quest, spawn a photo, open the menu, click "open link"

**Expected behavior**
User should be exited from VR gracefully so they can see the link, and then return to their session when finished (currently, I think their session will time out)

**Hardware** 
Quest

PR #1790, opened 2019-10-04
Properly handle opening links

gfodor:
Fixes https://github.com/mozilla/hubs/issues/1787

### DIFF
--- a/src/utils/vr-interstitial.js
+++ b/src/utils/vr-interstitial.js
@@ -14,7 +14,7 @@ const getUserGesture = () => {
   });
 };
 
-export async function handleExitTo2DInterstitial(isLower, exitAction) {
+export async function handleExitTo2DInterstitial(isLower, exitAction, nonFullscreen) {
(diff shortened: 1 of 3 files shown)


src/utils/vr-interstitial.js
```diff
@@ -14,7 +14,7 @@ const getUserGesture = () => {
   });
 };
 
-export async function handleExitTo2DInterstitial(isLower, exitAction) {
+export async function handleExitTo2DInterstitial(isLower, exitAction, nonFullscreen) {
```
johnshaughnessy: Thoughts on skipping "Continue" when entering the media browser? Iirc you added it so that you had an opportunity to "straighten out your head" after selecting an item -- perhaps we want to only have the "continue" screen on the way back into VR mode.

gfodor: the continue is there because we need a user gesture to transition to full screen mode, so not needed in this case

PR #1792, opened 2019-10-04
fix bindings for movement for vive focus plus

InfiniteLee:
We are currently using the same input bindings for the Vive Focus Plus as the Vive and Index, however it has inverted y touchpad axis coordinates. This fixes that. 

There is a separate issue where the pressing on the left touchpad on either controller causes the trigger to also fire (in Hubs this is treated as a teleport)- this appears to be a bug at the gamepad api level, not in Hubs.

### DIFF
--- a/src/systems/userinput/userinput.js
+++ b/src/systems/userinput/userinput.js
@@ -281,7 +281,6 @@ AFRAME.registerSystem("userinput", {
           mapping && this.registeredMappings.add(mapping);
 
           if (activeDevice instanceof ViveControllerDevice && activeDevice.gamepad) {
-            console.log(activeDevice);
             if (activeDevice.gamepad.id === "HTC Vive Focus Plus Controller") {
               //HTC Vive Focus Plus Controller
               this.registeredMappings.add(viveFocusPlusUserBindings);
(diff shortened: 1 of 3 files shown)


src/systems/userinput/userinput.js
```diff
@@ -274,8 +279,19 @@ AFRAME.registerSystem("userinput", {
           const activeDevice = this.activeDevices.items[i];
           const mapping = vrGamepadMappings.get(activeDevice.constructor);
           mapping && this.registeredMappings.add(mapping);
-          if (activeDevice instanceof ViveControllerDevice && activeDevice.isViveWand) {
-            this.registeredMappings.add(viveWandUserBindings);
+
+          if (activeDevice instanceof ViveControllerDevice && activeDevice.gamepad) {
+            console.log(activeDevice);
```
johnshaughnessy: drop

src/systems/userinput/userinput.js
```diff
@@ -274,8 +279,19 @@ AFRAME.registerSystem("userinput", {
...
+            if (activeDevice.gamepad.id === "HTC Vive Focus Plus Controller") {
+              //HTC Vive Focus Plus Controller
+              this.registeredMappings.add(viveFocusPlusUserBindings);
+            } else if (activeDevice.gamepad.axes.length === 4) {
```
johnshaughnessy: Do wands and index share a string id and should we match on it here? (Or just default all other controllers to one of these since it's out best guess?)

PR #1795, opened 2019-10-07
update cursor rotation for vive focus plus to a more comfortable angle

InfiniteLee:


InfiniteLee:
Actually that was a typo, didn't mean to change that. Good catch

PR #1796, opened 2019-10-08
Object list tweaks

johnshaughnessy:
This PR updates the object list in the following ways:
- A user preference `show-background-while-inspecting` is saved to the browser's local storage and persisted across sessions.
- Lights are added to the inspect layer since they should still be visible from inspect mode. 
- Cancel button is now clickable text, and there is a close button in the corner.
- "Pin" button is now the red "action" color.
- Icon for 3D models changed to match Spoke (boxes -> cubes)
- Mouse zoom while inspecting is given some acceleration
- Pan up and down while inspecting with middle mouse + drag


johnshaughnessy:
Ignoring failed automated check:
```
/home/circleci/repo/src/hub.js
  2:8  error  'configs' is defined but never used  no-unused-vars
```
I think it's expected after https://github.com/mozilla/hubs/pull/1817

PR #1815, opened 2019-10-13
Bug/more dom initialization issues

gfodor:
This fixes up the areas in our code where we are performing query selectors or lookups into the dom by id during `init()`, which is known to break in certain conditions.


gfodor:
believe failing build is unrelated to PR

PR #1825, opened 2019-10-16
Wait for DOMContentLoaded

johnshaughnessy:
Fix for the issue seen in the following screenshot:

image

The first commit should (indirectly) fix this particular issue. The second should fix any potential issues with the interaction system. If we revisit initialization and find a better way to handle this class of issues, we can remove the workaround. For now, I added a guard to the sites that make use of those entities:
```
8 candidates:
./src/systems/constraints-system.js:103:      interaction.options.leftHand.entity.id,
./src/systems/two-point-stretching-system.js:34:        ? interaction.options.leftHand.entity.object3D
./src/systems/interactions.js:182:      this.options.leftHand.entity = document.getElementById("player-left-controller");
./src/systems/interactions.js:230:    if (this.options.leftHand.entity.object3D.visible && !this.state.leftRemote.held) {
./src/systems/super-spawner-system.js:65:      interaction.options.leftHand.entity,
./src/components/super-spawner.js:122:      ? this.el.sceneEl.systems.interaction.options.leftHand.entity.object3D
./src/components/hoverable-visuals.js:60:      interactorOne = interaction.options.leftHand.entity.object3D;
./src/systems/render-manager/hubs-batch-raw-uniform-group.js:93:            interactorOne = interaction.options.leftHand.entity.object3D;
```

### DIFF
--- a/src/components/hoverable-visuals.js
+++ b/src/components/hoverable-visuals.js
@@ -56,6 +56,7 @@ AFRAME.registerComponent("hoverable-visuals", {
 
     let interactorOne, interactorTwo;
     const interaction = AFRAME.scenes[0].systems.interaction;
+    if (!interaction.options.leftHand.entity) return; //DOMContentReady workaround
     if (interaction.state.leftHand.hovered === this.el && !interaction.state.leftHand.held) {
       interactorOne = interaction.options.leftHand.entity.object3D;
     }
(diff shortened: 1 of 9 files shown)


src/components/hoverable-visuals.js
```diff
@@ -56,6 +56,7 @@ AFRAME.registerComponent("hoverable-visuals", {
 
     let interactorOne, interactorTwo;
     const interaction = AFRAME.scenes[0].systems.interaction;
+    if (!interaction.options.leftHand.entity) return; //DOMContentReady workaround
```
gfodor: since this is reaching pretty far into interaction system, i'd suggest maybe exposing this as a simple `ready` flag?

gfodor: (only partially convinced that's a better idea, but wanted to surface)

PR #1828, opened 2019-10-17
Add ability to find "similar" avatars in the media browser

netpro2k:
Adds "find similar" buttons to avatar listing results in the media browsers. Currently this is implemented in the backend to just find avatars that share the same parent avatar, but this can be updated later to actually use a more interesting metric of similarity.

Requires backend changes here: https://github.com/mozilla/reticulum/pull/249

### DIFF
--- a/src/react-components/media-browser.js
+++ b/src/react-components/media-browser.js
@@ -125,11 +125,10 @@ class MediaBrowser extends Component {
     const searchParams = new URLSearchParams(props.history.location.search);
     const result = props.mediaSearchStore.result;
 
-    const newState = { result, query: this.state.query || searchParams.get("q") || "" };
+    const newState = { result, query: searchParams.get("q") || "" };
     const urlSource = this.getUrlSource(searchParams);
     newState.showNav = !!(searchParams.get("media_nav") !== "false");
     newState.selectAction = searchParams.get("selectAction") || "spawn";
-    newState.query = searchParams.get("q") || "";
 
     if (result && result.suggestions && result.suggestions.length > 0) {
       newState.facets = result.suggestions.map(s => {
@@ -411,7 +410,10 @@ class MediaBrowser extends Component {
                 ))}
               {activeFilter === "similar" && (
                 <a className={classNames(styles.facet, "selected")}>
-                  Similar to: &quote;{searchParams.get("similar_name")}&quote;
+                  <FormattedMessage
+                    id="media-browser.similar-to-facet"
+                    values={{ name: searchParams.get("similar_name") }}
+                  />
                 </a>
               )}
             </div>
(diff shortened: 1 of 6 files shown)


src/react-components/media-browser.js
```diff
@@ -129,6 +129,7 @@ class MediaBrowser extends Component {
     const urlSource = this.getUrlSource(searchParams);
     newState.showNav = !!(searchParams.get("media_nav") !== "false");
     newState.selectAction = searchParams.get("selectAction") || "spawn";
+    newState.query = searchParams.get("q") || "";
```
gfodor: it looks like this should be reflected on line 128. do we want to discard the value in `state` here in all cases? i haven't looked at this code in a while

src/react-components/media-browser.js
```diff
@@ -265,7 +272,8 @@ class MediaBrowser extends Component {
       }
     };
 
-    const activeFilter = searchParams.get("filter") || (!searchParams.get("q") && "");
+    const activeFilter =
+      searchParams.get("filter") || (searchParams.get("similar_to") && "similar") || (!searchParams.get("q") && "");
```
gfodor: whoa I found this line confusing. i guess a bunch of `if`s is worse tho.

src/react-components/media-browser.js
```diff
@@ -401,8 +409,13 @@ class MediaBrowser extends Component {
                     {s.text}
                   </a>
                 ))}
-              </div>
-            )}
+              {activeFilter === "similar" && (
+                <a className={classNames(styles.facet, "selected")}>
+                  Similar to: &quote;{searchParams.get("similar_name")}&quote;
```
gfodor: should be moved into strings file

src/react-components/media-tiles.js
```diff
@@ -172,28 +174,41 @@ class MediaTiles extends Component {
...
+              title="Edit"
             >
-              <FontAwesomeIcon icon={faClone} />
+              <FontAwesomeIcon icon={faPencilAlt} />
             </StateLink>
           )}
+          {entry.type === "avatar_listing" && (
+            <StateLink
+              onClick={e => {
+                e.preventDefault();
```
gfodor: i don't think you need to do this w `StateLink`

src/react-components/media-browser.js
netpro2k: From what I can tell `getStoreAndHistoryState` is called only on page load and whenever the store has been updated, in both cases it seems like we want to overwrite the state. I will combine this with the line above though.

netpro2k: From what I can tell `getStoreAndHistoryState` is called only on page load and whenever the store has been updated, in both cases it seems like we want to overwrite the state. I will combine this with the line above though.

PR #1832, opened 2019-10-17
Feature/vive cosmos support

InfiniteLee:
Adds support for Vive Cosmos.

Also updates other vive devices so we have mappings for all their buttons. (We don't currently use all the buttons, but now we can).

### DIFF
--- a/src/systems/userinput/devices/vive-controller.js
+++ b/src/systems/userinput/devices/vive-controller.js
@@ -17,12 +17,37 @@ export class ViveControllerDevice {
     navigator.getVRDisplays();
     this.gamepad = gamepad;
 
-    if (this.gamepad.id === "HTC Vive Focus Plus Controller") {
+    if (this.gamepad.id === "OpenVR Cosmos") {
+      RAY_ROTATION.setFromAxisAngle(new THREE.Vector3(1, 0, 0), -Math.PI / 12);
+      //TODO: verify this button order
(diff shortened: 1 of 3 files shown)


src/systems/userinput/devices/vive-controller.js
```diff
@@ -17,28 +17,55 @@ export class ViveControllerDevice {
...
         { name: "grip", buttonId: 2 }
       ];
+      this.axisMap = [{ name: "touchX", axisId: 0 }, { name: "touchY", axisId: 1 }];
+    } else if (this.gamepad.axes.length === 4) {
+      RAY_ROTATION.setFromAxisAngle(new THREE.Vector3(1, 0, 0), -Math.PI / 12);
+      this.buttonMap = [
+        { name: "touchpad", buttonId: 0 },
+        { name: "trigger", buttonId: 1 },
+        { name: "grip", buttonId: 2 },
+        { name: "primary", buttonId: 3 },
```
johnshaughnessy: A bit unfortunate to have to remember which button is "primary" on the different controllers, but doesn't seem like a big deal

InfiniteLee:
There's an argument to be made to do that, but I figured just keeping all the vive devices together makes more sense. I also think we're pretty well setup to extend the `vive-user` bindings right now, whereas we're less so with the `oculus-touch-user` bindings.

PR #1835, opened 2019-10-19
Set gain value directly, skipping setVolume

johnshaughnessy:
(Maybe) fix issue we saw in standup this week where the audio falloff of other avatars was delayed. I'm not sure how to repro the issue so this is a blind attempt.

In light of the audio issues we were having in the meetup today, I'm not so sure we should expect this change to fix anything after all, since I don't know how this change could have been responsible for the audio degradation we saw in standup.


gfodor:
lgtm

PR #1858, opened 2019-10-24
Bug/set matrix world

johnshaughnessy:
The helper for setting the `matrixWorld` of an `object3D` was not correct with our matrix optimizations. This makes sure the children of an updated object are also updated.

### DIFF
--- a/src/utils/three-utils.js
+++ b/src/utils/three-utils.js
@@ -80,31 +80,6 @@ export const setMatrixWorld = (() => {
   };
 })();
 
-//export const setMatrixWorld = (() => {
-//  const IDENTITY = new THREE.Matrix4().identity();
-//  const helperMatrix = new THREE.Matrix4();
-//  const inverseParentWorld = new THREE.Matrix4();
-//  return function setMatrixWorld(object3D, m) {
-//    object3D.parent.updateMatrices();
-//    object3D.updateMatrices();
-//    if (!object3D.parent.matrixIsModified) {
-//      object3D.parent.applyMatrix(IDENTITY);
-//    }
-//    object3D.parent.updateMatrices();
-//    object3D.updateMatrices();
-//    if (!object3D.matrixIsModified) {
-//      object3D.applyMatrix(IDENTITY);
-//    }
-//    object3D.updateMatrices();
-//
-//    inverseParentWorld.getInverse(object3D.parent.matrixWorld);
-//    helperMatrix.multiplyMatrices(inverseParentWorld, m);
-//    helperMatrix.decompose(object3D.position, object3D.quaternion, object3D.scale);
-//    object3D.matrixNeedsUpdate = true;
-//    object3D.updateMatrices();
-//  };
-//})();
-
 // Modified version of Don McCurdy's AnimationUtils.clone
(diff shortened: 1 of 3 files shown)


src/utils/three-utils.js
```diff
@@ -51,15 +51,34 @@ export function disposeNode(node) {
...
+
+    for (let i = 0; i < object3D.children.length; i++) {
+      setMatrixWorld(
+        object3D.children[i],
+        helperMatrix.multiplyMatrices(object3D.matrixWorld, object3D.children[i].matrix),
+        false
+      );
+    }
+  };
+})();
```
johnshaughnessy: This method is disconnected from the main "happy path" of updating transforms with our matrix optimizations. The "happy path" in my mind is to set the position, rotation, and scale components of the object3D to their desired values, then setting `matrixNeedsUpdate = true`. This would ideally update this object's matrix and matrixWorld, as well as the matrixWorlds of all of its children. However, I was unable to find a bug in this code snippet that attempts to do that. If someone else can get that approach working correctly, I will gladly use it. Here is my (failed) attempt, which did not work as intended:

```
export const setMatrixWorld = (() => {
 const IDENTITY = new THREE.Matrix4().identity();
 const helperMatrix = new THREE.Matrix4();
 const inverseParentWorld = new THREE.Matrix4();
 return function setMatrixWorld(object3D, m) {
   object3D.parent.updateMatrices();
   object3D.updateMatrices();
   if (!object3D.parent.matrixIsModified) {
     object3D.parent.applyMatrix(IDENTITY);
   }
   object3D.parent.updateMatrices();
   object3D.updateMatrices();
   if (!object3D.matrixIsModified) {
     object3D.applyMatrix(IDENTITY);
   }
   object3D.updateMatrices();

   inverseParentWorld.getInverse(object3D.parent.matrixWorld);
   helperMatrix.multiplyMatrices(inverseParentWorld, m);
   helperMatrix.decompose(object3D.position, object3D.quaternion, object3D.scale);
   object3D.matrixNeedsUpdate = true;
   object3D.updateMatrices();
 };
})();
```

johnshaughnessy: I was wrong. A simpler version of the above code DOES work correctly. Updating this PR accordingly.

gfodor:
Can you write some more notes on this PR? it seems important to me that this be documented for posterity, since this code is pretty hairy.

johnshaughnessy:
> Can you write some more notes on this PR? it seems important to me that this be documented for posterity, since this code is pretty hairy.
Yes, but my initial findings were wrong. I had some changes locally that were not reflected in this PR. I will re-open this when I figure out how to make things work correctly.

johnshaughnessy:
I forgot that the `matrix-auto-update` component was on some of the objects I was working with, which led to some very surprising results.

johnshaughnessy:
Need to test on mobile and VR platforms to make sure it's safe to remove the `matrix-auto-update` component from the viewing rig and camera. Ideally these objects will not have special semantics around when/how they update their matrices, since it complicates the code that uses them and leads to some surprises (like I had today).

johnshaughnessy:
not ready yet

PR #1870, opened 2019-10-30
Feature/css overhaul

gfodor:
Enables dynamic styling of the app. Main high level points:

- Uses CSS vars for all definitions now. We should not be adding any hand-written colors to the styling from this point forward.
- Adds a light and dark mode that can be toggled by applying a body CSS class.
- Re-colorizes sprites using a worker to match the theme. Splits the spritesheet in two, one for the 'action' color and one for the 'notice' color.
- Switches a bunch of icons to inline SVGs so they can be colorized with CSS.


src/systems/sprites.js
```diff
@@ -211,81 +222,104 @@ export class SpriteSystem {
...
-          const s = hoverableVisuals.sweepParams[0];
-          const t = hoverableVisuals.sweepParams[1];
-          aSweepParams.setXY(4 * i + 0, s, t);
-          aSweepParams.setXY(4 * i + 1, s, t);
-          aSweepParams.setXY(4 * i + 2, s, t);
-          aSweepParams.setXY(4 * i + 3, s, t);
-          aSweepParams.needsUpdate = true;
+    if (!this.meshes.action || !this.meshes.notice) return;
+
+    for (const [sheetType, mesh] of Object.entries(this.meshes)) {
```
brianpeiris: `Object.entries` allocates a new array. Probably shouldn't do it in `tick` or `raycast`

src/assets/stylesheets/2d-hud.scss
```diff
@@ -61,7 +61,7 @@
       height: 0;
       border-left: 5px solid transparent;
       border-right: 5px solid transparent;
-      border-bottom: 5px solid $bright-blue;
+      border-bottom: 5px solid $notice-background-color;
```
brianpeiris: Why are some of these `var` and others scss variables?

gfodor: because the `var` ones are set by the light/dark theme completely (ie, there are no defaults if unspecified) then it seemed more DRY to avoid an unnecessary macro layer. the other ones have default values so to DRY up references to those defaults I made them macros.

PR #1873, opened 2019-10-31
Unregister extra mappings added to ViveControllerDevice based gamepads when exiting immersive mode.

MortimerGoro:
Currently there is a infinite JavaScript loop in `user_input.js` `dependencySort()` function when exiting immersive mode in Vive devices. The problem is that some extra mappins were added in f7908e69c222c650ed3f9e60f86358f83570066f but they are not deleted when the activeDevice is removed. This causes some dependencies never be resolved in `dependencySort()`

fixes https://github.com/MozillaReality/FirefoxReality/issues/1970


gfodor:
:( so sorry this ended up being our bug!!! thanks so much for fixing it! i hope it was at least educational to set the hubs stack up to be able to debug this 🦆

PR #1881, opened 2019-11-04
Fix spawner scaling issues

InfiniteLee:
Add functionality to set custom mesh scaling on `media-loader` to fix scaling issues with `super-spawners`. The issue is that local clients must update the `mesh` scale of an object to match that of the spawner scale when an object is spawned, but we currently do not network that and as such the spawned entity's scale on remote clients is incorrect. This PR allows for setting of that scale value on `media-loader` (which _is_ networked) so that remote clients can apply the appropriate scale.

Also updates the `media-loaded` event to not leak memory on unused callbacks and updated `super-spawners` to use this event so that other file types can be supported in spawners.

### DIFF
--- a/src/utils/media-utils.js
+++ b/src/utils/media-utils.js
@@ -98,7 +98,9 @@ export const addMedia = (
   contentSubtype = null,
   resolve = false,
   resize = false,
-  animate = true
+  animate = true,
+  useCustomMeshScale = false,
+  customMeshScale = null
 ) => {
   const scene = AFRAME.scenes[0];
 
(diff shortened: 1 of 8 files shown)


src/utils/media-utils.js
```diff
@@ -122,15 +126,18 @@ export const addMedia = (
...
-        clearTimeout(fireLoadingTimeout);
-        scene.emit("media-loaded", { src: src });
-      },
-      { once: true }
-    );
+  const eventNames = ["model-loaded", "video-loaded", "image-loaded", "pdf-loaded"];
+
+  const cb = async () => {
+    clearTimeout(fireLoadingTimeout);
+    entity.emit("media-loaded", { src: src });
```
johnshaughnessy: `{ src }`

PR #1887, opened 2019-11-07
Improved support for audio media

brianpeiris:
Fixes audio media so that they are actually usable. This enables support for audio media from Spoke as well.

image

Audio is essentially treated as a type of video, except in the way it is displayed. The audio is still loaded by the `media-video` media view, but its "texture" and visuals are modified.

However, an `audio` component is exposed in the GLTF mappings which takes all the existing video media properties except for `projection`.


src/components/media-views.js
```diff
@@ -563,10 +565,14 @@ AFRAME.registerComponent("media-video", {
       this.el.setObject3D("mesh", this.mesh);
     }
 
-    this.mesh.material.map = texture;
+    if (this.data.contentType === "audio/mpeg") {
```
netpro2k: Any reason not to just look for anything starting with `audio/`, the browser can probably handle a number of different formats.

src/components/media-views.js
```diff
@@ -586,11 +592,19 @@ AFRAME.registerComponent("media-video", {
...
 
-      const texture = new THREE.VideoTexture(videoEl);
-      texture.minFilter = THREE.LinearFilter;
-      texture.encoding = THREE.sRGBEncoding;
-      const isReady = () =>
-        (texture.image.videoHeight || texture.image.height) && (texture.image.videoWidth || texture.image.width);
+      let texture, isReady;
+      if (contentType === "audio/mpeg") {
+        // We want to treat audio almost exactly like video, so we mock a video texture with an image property.
+        texture = new THREE.Texture();
```
netpro2k: We only create this texture so we can pull the video element off of it, and then later assign it to the map of the material. In the audio case we skip the second part, so it seems we could skip this whole thing by pulling `createVideoEl()` out into the calling code, and then only calling `createVideoTexture` for videos (passing in the videoEl)

src/components/media-views.js
robertlong: I kinda like only supporting media that all browsers can load. It'll make it easier to track down bugs. However, I also like being able to support better formats in most browsers 🤷‍♂

robertlong:
This audio component should still work if you mark it as invisible in Spoke right? I'm trying to figure out if we will be able to put audio emitters in the scene where you don't want to see this black square. So long as you can mark it invisible I think this should work.

src/components/media-views.js
brianpeiris: :+1: Changed it to allow all `audio/` mime types. This accounts for wav and flac, but not for ogg, since ogg uses `application/ogg`. I'm going to punt on ogg for now.

src/components/media-views.js
brianpeiris: Eh. I see what you mean but it seems like it could get hairy since we use texture as a container for things everywhere. It also contains this `hls` property and I was thinking it ought to work with audio as well in the future. So I'm going to punt on the restructuring for now.

brianpeiris:
Invisibility works for videos already, so I imagine it should just work for audio media.

PR #1888, opened 2019-11-08
Feature/assets to cloudflare

gfodor:
Fixes up the cloudflare worker situation, so it actually works now and also adds assets to the cloudflare CDN.


admin/src/react-components/data-transfer.js
```diff
@@ -30,29 +30,47 @@ const styles = withCommonStyles(() => ({
...
-      targetUrl = proxyUrl.protocol + "//" + targetUrl;
+  
+    const isCorsProxy = request.url.indexOf(CORS_PROXY_HOST) === 0;
+    const proxyUrl = new URL(isCorsProxy ? CORS_PROXY_HOST : PROXY_HOST);
+    const targetPath = request.url.substring((isCorsProxy ? CORS_PROXY_HOST : PROXY_HOST).length + 1);
+    let targetUrl;
+  
+    if (targetPath.indexOf("files/") === 0) {
+      targetUrl = \`\${STORAGE_HOST}/\${targetPath}\`;
+    } else if (targetPath.indexOf("hubs/") === 0 || targetPath.indexOf("spoke/") === 0 || targetPath.indexOf("admin/") === 0) {
```
brianpeiris: nit: why not use `startsWith` here?

admin/src/react-components/data-transfer.js
```diff
@@ -30,29 +30,47 @@ const styles = withCommonStyles(() => ({
...
+      targetUrl = \`\${ASSETS_HOST}/\${targetPath}\`;
+    } else {
+      if (!isCorsProxy) {
+        // Do not allow cors proxying from main domain, always require cors-proxy. subdomain to ensure CSP stays sane.
+        return;
+      }
+      targetUrl = targetPath.replace(/^http(s?):\/([^/])/, "http$1://$2");
```
brianpeiris: Maybe add a comment explaining why this substitution is necessary

PR #1889, opened 2019-11-09
Self hosting defaults

brianpeiris:
- Hide Twitter UI and media search tabs based on meta information injected from reticulum
- Force `enable_spoke` if user is an admin

Goes with https://github.com/mozilla/reticulum/pull/259


src/schema.toml
```diff
@@ -8,6 +8,7 @@ translations.en."app-short-domain" = { category = "translations", type = "string
 translations.en."contact-email" = { category = "translations", type = "string", name = "Contact email", description = "General contact email." }
 translations.en."share-hashtag" = { category = "translations", type = "string", name = "Share Hashtag", description = "Hashtag used in share messages." }
 
+features.permissive_rooms = { category = "features", type = "boolean", name = "Permissive Rooms", description = "Use permissive room settings for new rooms." }
```
gfodor: I'd change this to `Enable full content creation permissions on new rooms` or something similar?

brianpeiris: Gotcha. Would be nice if you could do a pass on the copy for the configs when the dust settles.

PR #1896, opened 2019-11-14
Scene Remixing

netpro2k:
Adds a "Copy to my scenes" button an scene listings in the media browser (similar to avatars) as well as a "Remix in Spoke" button on landing pages for scenes ans scene listings. If the user is logged in and viewing one of their own scenes that already has a project associated with it they will instead get an "Edit in Spoke" button. Note that this will be no scenes to start, since linking projects to scenes in the DB is a new feature/

These must be merged before pushing:
- [x]  backend changes: https://github.com/mozilla/reticulum/pull/260
- [x]  spoke change: https://github.com/mozilla/Spoke/pull/846


src/react-components/scene-ui.js
```diff
@@ -59,6 +64,16 @@ class SceneUI extends Component {
     createAndRedirectToNewHub(this.state.customRoomName, this.props.sceneId);
   };
 
+  remixScene = async e => {
+    e.preventDefault();
+    const { scenes } = await fetchReticulumAuthenticated("/api/v1/scenes", "POST", {
+      parent_scene_id: this.props.sceneId
+    });
+    const projectUrl = getReticulumFetchUrl(`/spoke/projects/${scenes[0].project_id}`);
+    console.log(projectUrl);
```
robertlong: You probably want to remove this

src/react-components/scene-ui.js
```diff
@@ -161,6 +178,23 @@ class SceneUI extends Component {
                   </button>
                 </WithHoverSound>
               </div>
+              {isOwner && sceneProjectId ? (
+                <a
+                  target="_blank"
```
robertlong: You need to add `rel="noopener noreferrer"` so that linting passes.

src/react-components/scene-ui.js
```diff
@@ -161,6 +178,23 @@ class SceneUI extends Component {
...
+                  href={getReticulumFetchUrl(`/spoke/projects/${sceneProjectId}`)}
+                  className={styles.spokeButton}
+                >
+                  <FontAwesomeIcon icon={faPencilAlt} />
+                  <FormattedMessage id="scene.edit_button" />
+                </a>
+              ) : (
```
robertlong: I'm still getting my stack up to test everything. What happens when you are not logged in and click the remix scene button?

netpro2k: You won't get the remix button if you are not logged in since we wouldn't detect that you own the scene. That said we do need to handle that for the "Copy to My Scenes" button. I think I will just have the button hidden for now, good catch.

PR #1900, opened 2019-11-15
Feature/waypoints

johnshaughnessy:
Scene authors can now add waypoints to their scenes. Waypoints allow a scene author to specify spawn points, points of interest, places to teleport, and places to sit.

The following options can be configured on a waypoint:
```
    canBeSpawnPoint: { default: false }, // user may appear here when they enter the room
    canBeOccupied: { default: false }, // only one user may appear here at a time
    canBeClicked: { default: false }, // will create a clickable object for the user can interact with
    willDisableMotion: { default: false }, // will temporarily keep the user in place
    snapToNavMesh: { default: false }, // will keep the user connected to the floor plane
    willMaintainInitialOrientation: { default: false }, // will not reorient the user, only translate and scale
```

Spoke : https://github.com/mozilla/Spoke/pull/843

Example scenes (dev) :
https://dev.reticulum.io/scenes/cKobWYp
https://dev.reticulum.io/scenes/P6QZQAM


johnshaughnessy:
The following options can be configured on a waypoint:
```
    canBeSpawnPoint: { default: false }, // user may appear here when they enter the room
    canBeOccupied: { default: false }, // only one user may appear here at a time
    canBeClicked: { default: false }, // will create a clickable object for the user can interact with
    willDisableMotion: { default: false }, // will temporarily keep the user in place
    snapToNavMesh: { default: false }, // will keep the user connected to the floor plane
    willMaintainInitialOrientation: { default: false }, // will not reorient the user, only translate and scale
```

Spoke : https://github.com/mozilla/Spoke/pull/843

Example scenes (dev) :
https://dev.reticulum.io/scenes/cKobWYp
https://dev.reticulum.io/scenes/P6QZQAM

EgleXR:
I will add an improvement for this functionality. I think it would be good to have a predetermined Asset, which would be a small, cylinder-shaped area, to teleport from one place to another by having contact with it. It would work similarly to the waypoints, with the difference that they are operable by contact with the area of the object. I will attach some guide image.  

Portalpoints

PR #1915, opened 2019-11-19
Fix coloring of several links in overlay panels

### LINKED ISSUE #1907: Text to upload custom GLB in avatar editor is white text on white background
This means that you can't see the option unless you know it's there and click on, at which point it gets highlighted 

image

image

netpro2k:
Fixed text color for a few things in the overlay panels like the avatar editor, media browser, and scene info panel. Also fixed the close button color in these panels so it is consistent.

Fixes #1907

### DIFF
--- a/src/assets/stylesheets/avatar-editor.scss
+++ b/src/assets/stylesheets/avatar-editor.scss
@@ -107,7 +107,7 @@
       }
       a {
         margin-left: 10px;
-        color: $action-label-color;
+        color: var(--panel-subtext-link-color);
       }
     }
 
@@ -199,6 +199,10 @@
 
     .license {
       font-size: 8pt;
+      color: var(--panel-subtext-color);
+      a {
+        color: var(--panel-subtext-link-color)
+      }
     }
 
     hr {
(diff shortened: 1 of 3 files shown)


src/assets/stylesheets/avatar-editor.scss
```diff
@@ -107,7 +107,7 @@
       }
       a {
         margin-left: 10px;
-        color: $action-label-color;
```
netpro2k: Why is this variable is resolved at build time instead of runtime? (Seems to be done this way for all usages of `action-label-color` so guessing its intentional.)

gfodor: the `$`'s are used because those macros expand into vars that have default values. so the alternative would have been to paste the default values at every site, which isn't DRY

yuriaru:
This might be related, but some text on the overlay does not contrast with the background color. I am using the dark mode theme. My color variables under the theme settings do not include any black color for the text (they are gray or white).

Screen Shot 2020-06-15 at 11 44 11 AM

PR #1916, opened 2019-11-19
Unbranded client

brianpeiris:
- Remove branded logos from client. Logos need to be set in reticulum instead. In local development, without a local ret server, we use unbranded placeholder assets instead.
- Removed or replaced mentions of "Hubs" and "Mozilla" where possible. Remaining instances require server meta injection in order to be removed.


admin/src/admin.html
```diff
@@ -8,7 +8,7 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no">
 
     <link rel="shortcut icon" type="image/png" href="/favicon.ico">
-    <title>Admin | Hubs by Mozilla</title>
+    <title>Admin | App by Company</title>
```
brianpeiris: We'll have to add meta templates for the html pages that don't have them already, in order to override the titles. I believe there's also code that expects the title to contain a `|`, so I just replaced the name here instead of removing it, so that local development continues to work.

PR #1917, opened 2019-11-19
Add warning when quota exceeded

gfodor:
Goes with https://github.com/mozilla/reticulum/pull/265

Adds a warning to the admin console when the storage quota is met.


admin/src/react-components/system-editor.js
```diff
@@ -41,6 +41,7 @@ class SystemEditorComponent extends Component {
   render() {
     const needsAvatars = this.state.reticulumMeta.repo && !this.state.reticulumMeta.repo.avatar_listings.any;
     const needsScenes = this.state.reticulumMeta.repo && !this.state.reticulumMeta.repo.scene_listings.any;
+    const exceededStorageQuota = this.state.reticulumMeta.repo && this.state.reticulumMeta.repo.storage.in_quota;
```
brianpeiris: `in_quota` should be negated here, no?

gfodor:
Oops yep nice catch — flipped it to debug

PR #1919, opened 2019-11-20
Emoji Spawners

InfiniteLee:
Replaces the avatar emoji system with new pause-menu that lets you spawn grabbable emojis that emit particles when shaken.

emojis2


src/components/emoji-hud.js
```diff
@@ -1,45 +1,102 @@
...
-      emoji.object3D.addEventListener("interact", () => {
-        const newEmoji = this.currentEmoji === emojiType ? "empty" : emojiType;
-        playerRig.setAttribute("player-info", { emojiType: newEmoji });
-      });
-    }
+      const cylinder = document.createElement("a-cylinder");
+      cylinder.setAttribute("visibility-while-frozen", { requireHoverOnNonMobile: false });
+      cylinder.setAttribute("material", { opacity: 0.2, color: "#2f7fee" });
+      cylinder.setAttribute("segment-height", 1);
+      cylinder.setAttribute("segment-radial", 8);
```
brianpeiris: It's `segments` not `segment`.

src/components/particle-emitter.js
```diff
@@ -6,25 +6,28 @@ import { proxiedUrlFor } from "../utils/media-url-utils";
 AFRAME.registerComponent("particle-emitter", {
   schema: {
     resolve: { type: "boolean", default: true },
-    src: { type: "string" },
+    src: {
+      type: "string",
+      default: "https://assets-prod.reticulum.io/spoke/assets/images/dot-75db99b125fe4e9afbe58696320bea73.png"
```
brianpeiris: We can't have assets being pulled from reticulum with the new self hosting stuff. You should replace this with a static client asset that you import here.

InfiniteLee: This was mostly just for testing purposes, but it would be nice to have a default url so that if you drop in a particle-emitter component in hubs it "just works". That said, if we use a static client asset, the previous schema prop "resolve" needs to be false, and right now it defaults to true. Unsure if defaulting it to false will wind up breaking any spoke scenes that use emitters right now.

src/hub.html
```diff
@@ -364,6 +364,51 @@
...
+                    destroy-at-extreme-distances
+                    scalable-when-grabbed
+                    floaty-object="modifyGravityOnRelease: true; autoLockOnLoad: true;"
+                    set-yxz-order
+                    matrix-auto-update
+                    position-at-box-shape-border__freeze="target:.freeze-menu;"
+                    hoverable-visuals
+                >
+                    <a-entity
+                        id="particle-emitter"
```
johnshaughnessy: May want `class="particle-emitter"` since `id`'s "must be unique" (source https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id)

src/network-schemas.js
```diff
@@ -123,6 +123,31 @@ function registerNetworkSchemas() {
...
+      },
+      {
+        component: "emoji",
+        property: "emitEndTime"
+      },
+      "scale",
+      "media-loader",
+      "pinnable",
+      {
+        selector: "#particle-emitter",
```
johnshaughnessy: May want `.particle-emitter` as per comment about `id` above

src/components/emoji-hud.js
```diff
@@ -1,45 +1,102 @@
...
+const emojis = [emoji_0, emoji_1, emoji_2, emoji_3, emoji_4, emoji_5];
+
+AFRAME.registerComponent("emoji-hud", {
+  schema: {
+    spawnerScale: { default: 0.0625 },
+    spawnedScale: { default: 0.25 }
+  },
+  init() {
+    const width = this.data.spawnerScale;
+    const spacing = width / 5;
```
johnshaughnessy: Will this always be `width / (emojis.length- 1)`?

src/components/hoverable-visuals.js
```diff
@@ -87,7 +87,11 @@ AFRAME.registerComponent("hoverable-visuals", {
 
     if (interactorOne || interactorTwo || isFrozen) {
       const worldY = this.el.object3D.matrixWorld.elements[13];
-      const scaledRadius = this.el.object3D.scale.y * this.boundingSphere.radius;
+      const ms1 = this.el.object3D.matrixWorld.elements[4];
+      const ms2 = this.el.object3D.matrixWorld.elements[5];
+      const ms3 = this.el.object3D.matrixWorld.elements[6];
+      const worldScale = Math.sqrt(ms1 * ms1 + ms2 * ms2 + ms3 * ms3);
```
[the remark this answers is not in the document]
InfiniteLee: @brianpeiris?

PR #1920, opened 2019-11-20
Configurable meta

brianpeiris:
- Removed branding from page titles, etc. Reticulum will override these instead.
- Added new configs for app tagline, full app name, and twitter handle
- Configurable editor name and logo, defaulting to "scene editor"
- Reticulum now injects app name into hub.service.js
- Removed static app icons and thumbnail. These must now be configured in reticulum.
- Deleted old spoke landing page

Goes with https://github.com/mozilla/reticulum/pull/266


src/react-components/change-scene-dialog.js
```diff
@@ -50,7 +50,7 @@ export default class ChangeSceneDialog extends Component {
               Paste a URL to a{" "}
               <IfFeature name="enable_spoke">
                 <a href="/spoke" target="_blank" rel="noopener noreferrer">
```
gfodor: i'd just change this to be a link to the phrase "custom scene" to keep it simple

PR #1922, opened 2019-11-21
HEAD url origins to determine if they are hubs servers

brianpeiris:


src/components/gltf-model-plus.js
```diff
@@ -204,15 +204,19 @@ const inflateEntities = function(indexToEntityMap, node, templates, isRoot, mode
...
+    }
     const entityComponents = getHubsComponents(object3D);
     const el = object3D.el;
 
     if (entityComponents && el) {
       for (const prop in entityComponents) {
         if (entityComponents.hasOwnProperty(prop) && AFRAME.GLTFModelPlus.components.hasOwnProperty(prop)) {
           const { componentName, inflator } = AFRAME.GLTFModelPlus.components[prop];
-          inflator(el, componentName, entityComponents[prop], entityComponents, indexToEntityMap);
+          previousInflation = inflator(el, componentName, entityComponents[prop], entityComponents, indexToEntityMap);
```
gfodor: what raised this issue?

brianpeiris: I didn't run into an issue, but I didn't want to break the contract where inflation is currently sequential, given we often have surprises with component ordering and instantiation

brianpeiris: Actually, this code is probably wrong, now that I look at it. I'll do more proper testing later, but I do still want to maintain that contract.

PR #1926, opened 2019-11-22
Fix touchscreen issues on mobile apple devices

johnshaughnessy:
Fix https://github.com/mozilla/hubs/issues/1813
See explanation in https://github.com/mozilla/hubs/issues/1813#issuecomment-557731148


src/utils/is-mobile.js
```diff
@@ -1,3 +1,5 @@
+import { hackyMobileSafariTest } from "./detect-touchscreen";
+
 // lifted from https://github.com/aframevr/aframe/blob/master/src/utils/device.js to ensure consistency
 function isIOS() {
   return /iPad|iPhone|iPod/.test(window.navigator.platform);
```
johnshaughnessy: Should I include the hack here?

PR #1927, opened 2019-11-23
Add idle activity detector

brianpeiris:
Client:
- Idle timeout shows auto exit timer dialog. Timeout is currently hard coded to 2 hours.
- auto exit dialog will dismiss itself on activity
- Idle timeout is disabled if you're using the lobby broadcast "streamer mode", if you're in local dev (unless you use the `idle_timeout` query param), or you use the `allow_idle` query param
- Activity includes clicks, touches, character movement and in-vr interaction

Admin:
- Added an idle detector and dialog. Timeout is hard coded to 1 hour
- Perm token refresh and ret channel are disconnected on session end


src/systems/idle-detector.js
```diff
@@ -0,0 +1,52 @@
...
+    window.dispatchEvent(new CustomEvent("idle_detected"));
+  },
+  tick(time) {
+    if (time - this.lastInputCheck < INPUT_CHECK_INTERVAL_MS) return;
+
+    const userinput = this.el.systems.userinput;
+    const characterAcceleration = userinput.get(paths.actions.characterAcceleration);
+
+    const active =
+      userinput.get(paths.actions.startGazeTeleport) ||
```
gfodor: seems fine, but might want to consider sticking these paths into a constant at the top to make it more clear and easy to modify

PR #1930, opened 2019-11-25
Adds landing page for self-hosted Hubs

misslivirose:
Still some more copy, styling, and content to put in on the page. Will need to update the links to the marketplace when it's up and get the video in place, but here's the start!


netpro2k:
Wonder if we should have this link to a landing page on our own servers rather than having each instance host it's own version. Seems like the kind of thing we would want ownership over. Maybe that still means making it a client page, but the link from the homepage maybe should be to https://hubs.mozilla.com/cloud

Issue #1801, opened 2019-10-09
Audio Continues Playing After Video Deleted


emclaren:
**Description**
Audio continued playing for one person in the room after a video was deleted during load.

A video was taking an unusually long time to load so I deleted the "loading object". From what I could tell, the video never existed in the room. Another user was in the lobby when I deleted the object. He entered the room and told me he could still hear the video audio playing.

I left and re-entered the room, but the audio continued for the other user.

**To Reproduce**
Have not been able to reproduce myself. 

**Expected behavior**
I expect that after a video is deleted, the audio does not play for anyone in the room. 

**Screenshots**
NA

**Hardware** 
The person who tried to load the object 
- Device: Laptop 
 - OS: Mac OS 10.14.6
 - Browser Firefox Nightly 71.0a1 (2019-09-24)

The person hearing the audio
Device: Oculus Go

**Additional context**
I emailed the user to get more information on their system

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-262)

misslivirose:
I would be inclined to make this a P1 if we can get a consistent repro on it - was this fixed by refreshing the room? Or did the audio just play forever?

emclaren:
I refreshed the room and it didn't end the audio. The other person did not attempt to refresh, we got caught up in conversation about something (and he reported the audio was quiet). I just sent a message to this user to see if he could tell us if it ever stopped.

This was in the #office-hours room, so its possible another Oculus Go in this room might still hear the audio playing.

emclaren:
There was reports from someone in last Friday's meetup that they were hearing audio from a video after the video had been removed from the room.

They reported they were on the following system
Lenovo IdeaPad 520 151KB
RAM: 8GB
Graphics: NVIDIA 940MX 2GB
Processor: Intel i5 7th Gen
OS: Lubuntu 18.04
Browser: Firefox Quantum 64 (latest stable)

robin-k-wilson:
I have a repro: 
Mac, Chrome

1. In a hubs.mozilla.com room
2. Load this vimeo video (~16m long): https://vimeo.com/342333493
3. Don't delete the loading cube yet! -- will not load
4. Wait for the audio to start
5. Delete the cube
6. Still hear the audio
7. Reload page in order to stop the audio

matthewbcool:
unable to repro

PR #1942, opened 2019-11-26
Fix removal of hoverable entities

brianpeiris:
This was causing scene removal to break, which would sometimes leave videos playing in a scene after you've left it.


misslivirose:
Hi @brianpeiris ! Would this also be likely to fix #1801 ?

PR #1960, opened 2019-12-06
Rework text bubbles to always use 1024 textures

netpro2k:
Rework scaling of text bubbles to be a bit more sane. Text bubbles will now always fit into a 1024x1024 texture (with the longer dimension being 1024 and the shorter side preserving aspect ratio). This seems to strike a reasonable compromise between text clarity for the common case of fairly short labels, while also batching neatly and not blowing up texture memory.

Completely unrelated, bumps node-sass dependency since I was having an issue building on my system with a newer version of node.


src/react-components/chat-message.js
```diff
@@ -110,34 +112,26 @@ function renderChatMessage(body, from, allowEmojiRender, lowResolution) {
...
-      img.onload = async () => {
+      img.onload = () => {
+        context.clearRect(0, 0, messageCanvas.width, messageCanvas.height);
         context.drawImage(img, 0, 0);
-        const blob = await new Promise(resolve => messageCanvas.toBlob(resolve));
         ReactDOM.unmountComponentAtNode(el);
         el.remove();
-        resolve(blob);
+        img.onload = null;
+        img.src = "";
```
netpro2k: Not sure this is needed but I got a bit paranoid about leaking these data urls

gfodor:
fyi linter

gfodor:
lgtm, please test chat bubbles in VR

PR #1977, opened 2019-12-17
Exit if scene fails to load

brianpeiris:
We ran into this with a particularly large scene, where both Chrome and Firefox would throw an error on Android while uploading textures from the scene after reaching the device's memory limit. This fix at least exits the room instead of spinning on the loading screen forever.

### DIFF
--- a/src/assets/translations.data.json
+++ b/src/assets/translations.data.json
@@ -172,6 +172,7 @@
     "exit.subtitle.disconnected": "You have disconnected from the room. Refresh the page to try to reconnect.",
     "exit.subtitle.left": "You have left the room.",
     "exit.subtitle.full": "This room is full, please try again later.",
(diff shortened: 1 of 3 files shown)


gfodor:
way to close our a 2020 item in 2019 :)

brianpeiris:
For future reference. The hitting the memory limit during texture upload would surface as "An attempt was made to use an object that is not, or is no longer, usable" in Firefox on Android and "Failed to load glTF model DOMException: The ImageBitmap could not be allocated." in Chrome on Android.

Firefox will also experience a hard crash if you reloaded the page after the first failure (or sometimes on the first load), depending on overall system memory pressure. Though I think this is a firefox memory allocation bug, that I will report to bugzilla separately.

Issue #1877, opened 2019-11-01
Changing scene to Hubs Room URL instead of Scene URL breaks room


emclaren:
**Description**
When you change the scene using the "/scene" command, if you use a URL that links to a Hubs Room instead of a Scene by mistake, the room breaks and will no longer load. 

It throws an error that says "SyntaxError: JSON.parse: unexpected character at line 1 column 1 of the JSON data"

**To Reproduce**
1. Go to Hubs room
2. type /scene <link to a Hubs room URL>
3. Refresh the page, it will no longer load

**Expected behaviour**
I expect the scene to not change, and when I refresh the scene it will remain as the last-used scene, or just a big empty room. I do not expect it to make the room URL break. 

**Additional context**
This error happened to Elgin while testing out scenes, but it has also been flagged by the community

misslivirose:
Given that this fairly catastrophically impacts the room, I think it would be good to handle this in a way that doesn't break the room.

emclaren:
Just realized as well - there is no way of removing a room that was broken like this from the list of favourites (since you need to open the room to favourite/unfavourite). My list is stuck looking like this
<img width="1587" alt="Screen Shot 2019-11-18 at 7 57 54 AM" src="https://user-images.githubusercontent.com/6835324/69068204-3a5dcb00-09d9-11ea-8732-29c650a1cc12.png">

misslivirose:
@emclaren just opened a new issue for removing favorites (#1912 )

misslivirose:
When we fix this, we should incorporate the suggestions in #811 as well in terms of general error messaging back to the user when a URL isn't going to work.

PR #1985, opened 2019-12-18
Validate scene url before updating scene

brianpeiris:
Fixes https://github.com/mozilla/hubs/issues/1877

- Does a client-side check on the URL, checking if it's a Hubs scene, using our existing isHubsSceneUrl method, or a GLB by downloading the first chunk of the URL and checking its header
- Prevents submission on the Change Scene dialog using html form validation.

### DIFF
--- a/src/utils/scene-url-utils.js
+++ b/src/utils/scene-url-utils.js
@@ -1,12 +1,19 @@
 import { isHubsSceneUrl, proxiedUrlFor } from "../utils/media-url-utils";
 
 export async function isValidGLB(url) {
-  return fetch(url).then(r => {
+  return fetch(url).then(async r => {
     const reader = r.body.getReader();
-    return reader.read().then(result => {
-      reader.cancel();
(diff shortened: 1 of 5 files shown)


src/utils/scene-url-utils.js
```diff
@@ -0,0 +1,21 @@
+import { isHubsSceneUrl, proxiedUrlFor } from "../utils/media-url-utils";
+
+export async function isValidGLB(url) {
+  return fetch(url).then(r => {
+    const reader = r.body.getReader();
+    return reader.read().then(result => {
+      reader.cancel();
+      return String.fromCharCode.apply(null, result.value.slice(0, 4)) === "glTF";
```
gfodor: this is a little weird, but i guess i can't think of a better way to do it. is there any chance we may get less than 4 characters in the first `read()`?

brianpeiris: Seems really unlikely that a valid GLB will have a first chunk that is less than 4 bytes, but I'll change it to iterate through the stream, just in case.

PR #1988, opened 2019-12-19
Feature/admin import table

gfodor:
This completely updates the content import tool in the admin console:

- Specify multiple URLs separated by commas
- Allow use of "asset pack" files which are just text files ending in `.pack` that are a list of URLs. See examples: https://github.com/mozilla/hubs-cloud/tree/master/asset-packs
- Give the user a visual preview of what is to be imported, allow them to de-select items, and select flags.
- Show visual feedback of each import.

Also fixes the link colors on the admin pages.

### DIFF
--- a/admin/src/react-components/import-content.js
+++ b/admin/src/react-components/import-content.js
@@ -265,16 +265,7 @@ class ImportContentComponent extends Component {
 
       await exec(() => {
         const d = approve(objectRes.data[0]);
-
-        if (!d.payload.data.tags) {
-          d.payload.data.tags = { tags };
-        } else {
-          for (const t of tags) {
-            if (d.payload.data.tags.tags.indexOf(t) < 0) {
-              d.payload.data.tags.tags.push(t);
-            }
-          }
-        }
+        d.payload.data.tags = { tags };
 
         return d;
       });
--- a/admin/src/react-components/import-content.js
+++ b/admin/src/react-components/import-content.js
@@ -31,13 +31,12 @@ import Button from "@material-ui/core/Button";
 import { fetchReticulumAuthenticated } from "hubs/src/utils/phoenix-utils";
 import clsx from "classnames";
 import { GET_MANY_REFERENCE } from "react-admin";
-import { sceneApproveNew, sceneReviewed } from "./scene-actions";
-import { avatarApproveNew, avatarReviewed } from "./avatar-actions";
+import { sceneApproveNew, sceneApproveExisting, sceneReviewed } from "./scene-actions";
+import { avatarApproveNew, avatarApproveExisting, avatarReviewed } from "./avatar-actions";
 import withCommonStyles from "../utils/with-common-styles";
 import configs from "../utils/configs";
 
 const RESULTS = {
-  selecting: "selecting",
   pending: "pending",
   importing: "importing",
   new_listing: "new_listing",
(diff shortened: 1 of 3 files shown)


admin/src/react-components/import-content.js
```diff
@@ -73,129 +80,353 @@ class ImportContentComponent extends Component {
...
+      }
     }
 
-    const { url, isScene } = apiInfo;
-    const type = isScene ? "scenes" : "avatars";
-    const columnPrefix = isScene ? "scene" : "avatar";
-    this.setState({ importing: true });
+    for (let i = 0; i < importableUrls.length; i++) {
+      const url = importableUrls[i];
+      console.log(url);
```
brianpeiris: remove console.log

admin/src/react-components/import-content.js
```diff
@@ -73,129 +80,353 @@ class ImportContentComponent extends Component {
...
-    const tags = [];
+      if (!importUrl) continue;
 
-    if (this.state.addBaseTag) {
-      tags.push("base");
+      const res = await fetch(`https://${configs.CORS_PROXY_SERVER}/${importUrl}`);
+      const type = isScene ? "scenes" : "avatars";
+      const asset = (await res.json())[type][0];
+      const isDefault = (isScene && needsDefaultScene) || (isAvatar && needsDefaultAvatar);
+      const isBase = isAvatar && needsBaseAvatar && i === 0;
```
brianpeiris: Bit odd to condition this on `i === 0` if we might `continue` on some urls above.

admin/src/react-components/import-content.js
```diff
@@ -73,129 +80,353 @@ class ImportContentComponent extends Component {
...
-    if (listingRes.data.length === 0) {
+      const objectRes = await dataProvider(GET_MANY_REFERENCE, type, {
+        sort: { field: "id", order: "desc" },
+        target: `${columnPrefix}_sid`,
+        id: sid
+      });
+      const objId = objectRes.data[0].id;
+      const listingRes = await dataProvider(GET_MANY_REFERENCE, `${columnPrefix}_listings`, {
+        sort: { field: "id", order: "desc" },
+        target: `_${columnPrefix}_id`,
```
brianpeiris: Is the leading underscore a typo?

admin/src/react-components/import-content.js
```diff
@@ -73,129 +80,353 @@ class ImportContentComponent extends Component {
...
+      isEnabled: true
+    });
+    this.setState({ imports });
+  }
+
+  setImportResult(url, result) {
+    this.setImportFields(url, i => (i.result = result));
+
+    if (result === RESULTS.new_listing || result === RESULTS.existing_listing || result === RESULTS.failed) {
+      this.setImportFields(url, i => (i.isImported = true));
```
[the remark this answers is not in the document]
gfodor: Yeah, the `isImported` field is basically a way to mark the row as no longer selectable. It ended up not being necessary, but my original idea was to make it so if you imported two URLs in the same page load it would append to the table.

admin/src/react-components/import-content.js
gfodor: no, this is the hacky postgrest thing we do where the `_` fields are the text-ified integer columns (this code didn't change just got moved around)

admin/src/react-components/import-content.js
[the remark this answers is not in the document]
gfodor: (whereas the final version here ends up just clearing the table each time)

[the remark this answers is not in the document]
gfodor: Oh I should correct that, this flag is also used as a way to determine if there are any non-completed rows as well.

PR #1994, opened 2019-12-20
Remove connection status from loader

johnshaughnessy:
Rather than always expose connection state, we'd prefer to only show connection info if we catch an error. This reverts adding the connection info to the loader (but does not yet replace it with showing something to the user if they fail to connect to janus).


netpro2k:
I probably missed the context that led to this PR, but seems worth just doing the part where we pass through connection errors while you are at it.

johnshaughnessy:
Closed in favor of https://github.com/mozilla/hubs/pull/2008

PR #1998, opened 2019-12-20
Fix mis-named var given toml file having the wrong name

gfodor:
There was a mismatch between `schema.toml` for the primary action subtitle text color and what we were using in the code. Given that the toml file may have caused that name to be stored in the db, this PR updates the client to use the name from the toml file. 

Note this color is currently 'dark' and not used anywhere, though it makes sense to offer it since it has a place in the styling hierarchy if we ever add action buttons with subtitles.


gfodor:
Name here: https://github.com/mozilla/hubs/blob/master/src/schema.toml#L50

Issue #1950, opened 2019-12-02
Failed to execute 'shaderSource' on 'WebGL2RenderingContext'


daved:
**Description**
Hanging at "Entering Lobby..."

**To Reproduce**
1. Go to a room
2. See "Entering Lobby..."

**Expected behavior**
Enter the room.

**Screenshots**
N/A

**Hardware** 
 - Device: Laptop (HP Pavilion dv6t-7000)
 - OS: Ubuntu 16.04 **(64-bit)**
 - Browser: Chromium Version 78.0.3904.108 (Official Build) Built on Ubuntu

**Additional context**
Console error:
```
raven.js:381 Uncaught TypeError: Failed to execute 'shaderSource' on 'WebGL2RenderingContext': parameter 1 is not of type 'WebGLShader'.
    at ji (three.module.js:17648)
    at new to (three.module.js:18311)
    at no.acquireProgram (three.module.js:18755)
    at Ee (three.module.js:25427)
    at Me (three.module.js:25568)
    at three.module.js:24852
    at Dr.traverse (three.module.js:8857)
    at Fo.traverse (three.module.js:8863)
    at ko.compileAndUploadMaterials (three.module.js:24836)
    at ui-root.js:257
    at a (raven.js:377)
ji @ three.module.js:17648
to @ three.module.js:18311
acquireProgram @ three.module.js:18755
Ee @ three.module.js:25427
Me @ three.module.js:25568
(anonymous) @ three.module.js:24852
traverse @ three.module.js:8857
traverse @ three.module.js:8863
compileAndUploadMaterials @ three.module.js:24836
(anonymous) @ ui-root.js:257
a @ raven.js:377
setTimeout (async)
(anonymous) @ raven.js:1134
(anonymous) @ ui-root.js:255
a @ raven.js:377
requestAnimationFrame (async)
(anonymous) @ raven.js:1253
componentDidUpdate @ ui-root.js:254
Qo @ react-dom.production.min.js:220
Yo @ react-dom.production.min.js:212
Vo @ react-dom.production.min.js:211
Wo @ react-dom.production.min.js:211
zo @ react-dom.production.min.js:209
yo @ react-dom.production.min.js:207
enqueueSetState @ react-dom.production.min.js:148
v.setState @ react.production.min.js:12
(anonymous) @ ui-root.js:415
(anonymous) @ loader.js:77
a @ raven.js:377
setTimeout (async)
(anonymous) @ raven.js:1134
[...]

PR #2008, opened 2020-01-06
Catch more errors as scene load time

johnshaughnessy:
This removes the `"connecting"` and `"connected"` strings from the loading screen.

This increases robustness against certain categories of failure at scene load time:
- Listens for `model-error` during scene changes if you are still in the lobby
- Wraps `compileAndUploadMaterial` in a try/catch block in case browser is incapable of rendering the scene (as in #1950 ) 
- Adds a 30s timeout to the connection outside of the normal error handling to catch rare cases where connecting to the networked scene fails but the `catch` block does not run. (I have not been able to diagnose why this happens, but do occasionally see this.)

### DIFF
--- a/src/hub.js
+++ b/src/hub.js
@@ -558,11 +558,21 @@ function handleHubChannelJoined(entryManager, hubChannel, messageDispatch, data)
 
     const loadEnvironmentAndConnect = () => {
       updateEnvironmentForHub(hub, entryManager);
+      function onConnectionError() {
+        console.error("Unknown error occurred while attempting to connect to networked scene.");
+        remountUI({ roomUnavailableReason: "connect_error" });
+        entryManager.exitScene();
+      }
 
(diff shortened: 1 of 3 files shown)


src/hub.js
```diff
@@ -414,6 +415,7 @@ async function updateEnvironmentForHub(hub, entryManager) {
       },
       { once: true }
     );
+    environmentEl.addEventListener("model-error", sceneErrorHandler, { once: true });
```
brianpeiris: I decided against doing this for scene changes in https://github.com/mozilla/hubs/pull/1981

johnshaughnessy: Hmm. I don't think exiting is a very graceful outcome, but it seems important to let the user know about it. How about I only add this event listener if you have not yet entered the scene? I am concerned about the case of having a scene change right when you try to load the page.

brianpeiris: Ah, yeah, makes sense to catch this error if you haven't entered the scene

PR #2012, opened 2020-01-07
Change immersive mode exist behavior to show settings

gfodor:
Currently when using an immersive browser on mobile VR, clicking the menu button exits immersive mode and lands you on the 2d mode version of hubs, with an "Enter in VR" button to return. This PR changes that behavior so the menu button actually shows the settings menu, and closing the settings menu returns to VR mode. There doesn't seem to be much value in showing the 2d viewport (which no longer is visible because the menu is shown), and it seems better to not force the user to fish out the menu in the top left corner if they actually want to access settings.

One challenge is vertical real estate is at a premium, and we also want to leave room for the chat and HUD, so this PR truncates some of the less-relevant menu items in this context, such as the various help links and actions which are unlikely to be needed in an immersive context.

image

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -110,7 +110,7 @@ module.exports = (env, argv) => ({
   devtool: argv.mode === "production" ? "source-map" : "inline-source-map",
   devServer: {
     https: createHTTPSConfig(),
-    host: process.env.HOST_IP || "0.0.0.0",
+    host: "0.0.0.0",
     public: `${host}:8080`,
     useLocalIp: true,
     allowedHosts: [host],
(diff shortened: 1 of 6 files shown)


webpack.config.js
```diff
@@ -110,7 +110,7 @@ module.exports = (env, argv) => ({
   devtool: argv.mode === "production" ? "source-map" : "inline-source-map",
   devServer: {
     https: createHTTPSConfig(),
-    host: process.env.HOST_IP || "0.0.0.0",
```
gfodor: this seemed wrong, in my VM environment my host IP != my VM IP, so this failed to bind.

Issue #1975, opened 2019-12-16
Alternative ways to show multi-page PDF's


blairmacintyre:
**Is your feature request related to a problem? Please describe.**
Only showing 1 page of a PDF at a time is too limiting.  The metaphor is clean and simple, but it doesn't allow the full space to be used.

**Describe the solution you'd like**
When a PDF is imported, alternative modes of display could be offered.  For example, one might want to tile all the pages in a grid / row / column.   

Alternatively, when flipping through the pages of a PDF, a button could be added to "yank out" a copy of the page, and place it nearby.

Alternatively, the PDF viewer could have it's shown page could controlled;  defaults to 1, if you increase to 2 it shows 2 consecutive pages, and so on, next to each other.

Alternatively, a user could set up a PDF ahead of time by marking "key" pages which would generate copies of that page, which the person could position nearby before the room is used. For a presentation of a set of pages, this could serve as an index/summary/etc.
 
**Describe alternatives you've considered**
This suggest was triggered by https://github.com/mozilla/hubs/issues/1856, and experience trying to present slides everywhere (conferences, video, hubs).  An way to see all, or a few "near the current slide", or showing a slide, would be great.

gfodor:
We have a little screenshot icon on videos that lets you take a snapshot of the video at that point in time — would that mechanic work? Seems the most versatile, and would minimize the need for additional DoF like tiling options, etc

misslivirose:
See @gfodor 's comments in: https://github.com/mozilla/hubs/issues/1806

blairmacintyre:
That mechanic would be a good start;  the other part of this would be (perhaps) allowing those resulting images to index back into the PDF (e.g., I click on them, they jump the PDF to that point).  However, that's not as critical as being able to simply pull out images.

A secondary issue, though, is the placement of the resulting images;  it isn't useful to do this during a presentation, as manually repositioning those images will be awkward.  If we can do some sort of reasonable auto-positioning (e.g., lay them out in the plane of the PDF object), smaller and using "simple space-filling" around the PDF, that would be great.  (I think the video snapshot tries to do something like this)

gfodor:
a basic spawning feature is live. the layout issue is something that I hope we can improve through extending the proposed 'media targets' functionality in spoke. for example you could tag media targets as pre-defined destination regions for these snapshots, and have rules around discard. (for example, if you click the snapshot button, then it would automatically move + pin the new snapshot into a target region, discarding any existing page, using round-robin)

[...]

PR #2015, opened 2020-01-08
Spawn PDF pages

gfodor:
You can now spawn individual PDF pages into their own objects.

XZbKDY8U91

This PR also:

- Tightens up z-depth of button labels, to reduce the parallax effect we see over buttons since their icons have been raised above them up until now
- Bumps the resolution on the video player control images, which were grainy up close
- Removes an apparently unused sprite sheet definition
- Fixes a bug that was causing the pager to not be properly removed when you removed a PDF, causing it to appear floating at the origin

In response to https://github.com/mozilla/hubs/issues/1975


src/components/media-loader.js
```diff
@@ -485,6 +485,10 @@ AFRAME.registerComponent("media-pager", {
     this.toolbar = null;
     this.onNext = this.onNext.bind(this);
     this.onPrev = this.onPrev.bind(this);
+    this.onSnap = this.onSnap.bind(this);
+    this.localSnapCount = 0;
```
netpro2k: This doesn't seem to be used.

src/components/media-loader.js
```diff
@@ -485,6 +485,10 @@ AFRAME.registerComponent("media-pager", {
...
+    this.isSnapping = false;
+    this.onSnapImageLoaded = () => (this.isSnapping = false);
```
netpro2k: This never seems to be called and is the only thing mucking with `isSnapping` so they can both probably go

src/components/media-loader.js
```diff
@@ -533,12 +539,6 @@ AFRAME.registerComponent("media-pager", {
     }
   },
 
-  remove() {
```
netpro2k: Why did this get removed?

src/components/media-loader.js
gfodor: oops, thats what i get for copy pasting

src/components/media-loader.js
gfodor: `this.toolbar` is a reference to the toolbar subelement nested a few deep under `this.el` (as per the query selector), which will be removed automatically when the pdf is removed. this routine also caused a bug for some reason, where a toolbar remained floating at the origin after a pdf was removed. (my only guess is that its because we're removing a nested subelement manually, during a remove, which seems unwise since the outer loop is recursively walking the tree presumably)

Issue #1948, opened 2019-11-27
Scene page fails to load 3d scene with media


gfodor:
**Description**
The scene page tries to load and show the scene in a 3d viewport. This fails if the scene has media in it.

**To Reproduce**
1. Go to https://hubs.mozilla.com/scenes/sZoPEtk
2. There's an error in the log during media loader that prevents the scene from loading.

**Expected behavior**
Scene should load

misslivirose:
This is what I see when I view that scene (Firefox stable, desktop) 
![image](https://user-images.githubusercontent.com/4313320/69970287-fcd66300-14eb-11ea-98e2-e5c9bc0e89a9.png)

It loaded slowly, but ti did seem to load. Does this repro 100% for you? Is it a specific device/browser conversation?

PR #2016, opened 2020-01-08
Fix issues with scene landing page not loading scenes

gfodor:
Fixes https://github.com/mozilla/hubs/issues/1948

- `media-loader`'s `init` assumed NAF was accessible, but we load media on the scene landing page where networking is disabled. I couldn't think of a better fix to this -- unclear if having a more clear "Offline" flag somewhere would be better.

- The lobby camera behavior was previously used on both the lobby and scene landing page. Once it was converted to a system, this change was not propagated to the scene page. This PR creates a new systems loop for the scene pages, and adds just the lobby camera.

- Fixes an additional edge case in `floaty-object` that causes a crash on the scene page if the scene has media objects.

### DIFF
--- a/src/components/media-loader.js
+++ b/src/components/media-loader.js
@@ -60,12 +60,16 @@ AFRAME.registerComponent("media-loader", {
     this.onMediaLoaded = this.onMediaLoaded.bind(this);
     this.animating = false;
 
-    NAF.utils
-      .getNetworkedEntity(this.el)
-      .then(networkedEl => {
-        this.networkedEl = networkedEl;
-      })
-      .catch(() => {}); //ignore exception, entity might not be networked
+    try {
+      NAF.utils
+        .getNetworkedEntity(this.el)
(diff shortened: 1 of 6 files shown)


src/components/media-loader.js
```diff
@@ -60,12 +60,16 @@ AFRAME.registerComponent("media-loader", {
...
-      .catch(() => {}); //ignore exception, entity might not be networked
+    try {
+      NAF.utils
+        .getNetworkedEntity(this.el)
+        .then(networkedEl => {
+          this.networkedEl = networkedEl;
+        })
+        .catch(() => {}); //ignore exception, entity might not be networked
+    } catch (e) {
+      // NAF may not exist on scene landing page
```
johnshaughnessy: Would `if (NAF)` work here in place of `try catch`?

gfodor: that's what I tried first, it threw with the same error `NAF is undefined`

PR #2017, opened 2020-01-08
Fix scaling button behavior

johnshaughnessy:
The "scale" button in the menu reacted to the user's head position in VR.
This PR includes a simple rewrite for the scale button that does not have this problem (and removes it from the unrelated capabilities of the the transform-object-button, making it easier to read and understand).


src/components/transform-object-button.js
```diff
@@ -31,6 +27,131 @@ function qAlmostEquals(a, b) {
...
+        this.leftRaycaster = document.getElementById("left-cursor-controller").components[
+          "cursor-controller"
+        ].raycaster;
+        this.rightRaycaster = document.getElementById("right-cursor-controller").components[
+          "cursor-controller"
+        ].raycaster;
+        this.viewingCamera = document.getElementById("viewing-camera").object3D;
+      }
+      setMatrixWorld(this.plane, calculatePlaneMatrix(this.viewingCamera, this.el.object3D));
+      this.planeRotation.extractRotation(this.plane.matrixWorld);
```
gfodor: this seems like it can be extracted into another function

gfodor: (and used in tick)

gfodor: sorry meant line 104+

johnshaughnessy: Good point! :+1:

src/components/transform-object-button.js
```diff
@@ -245,11 +236,13 @@ AFRAME.registerSystem("transform-selected-object", {
     } = this.puppet;
     this.hand.getWorldQuaternion(currentControllerOrientation);
     controllerOrientationDelta.copy(initialControllerOrientation_inverse).premultiply(currentControllerOrientation);
-    this.target.quaternion.copy(initialObjectOrientation).premultiply(controllerOrientationDelta);
+    this.target.quaternion
+      .copy(initialObjectOrientation)
+      .premultiply(controllerOrientationDelta)
+      .premultiply(controllerOrientationDelta);
```
johnshaughnessy: Unrelated change (by request) -- doubles the sensitivity of rotation via the puppet-mode rotate button. I agree that this feels.

gfodor:
this lgtm, would like to talk through the math in person at some point, but the new behavior feels great! tested on quest and 2d.

src/components/transform-object-button.js
gfodor: i agree, it does feel.

johnshaughnessy:
@gfodor I'd be more than happy to talk thru the math in person! It's one of my favorite subjects. And it feels.

PR #2021, opened 2020-01-09
Feature/screenshot refresh

gfodor:
Goes with https://github.com/mozilla/reticulum/pull/300/files

Adds a "refresh" button to page links, so the underlying screenshot can be re-taken. This ended up being pretty involved.

The original approach I took was to add a `force` parameter to the media fetch APIs which would, when "refresh" was clicked, cache bust everything and force a new fetch from everyone. This ended up being really tricky to get working on the server side under race conditions, and also led to a lot of mutation and side effects. It also meant that if multiple copies of the same link were pasted into a room, and each refreshed at different points for some reason, someone joining late would actually render them both as the 'latest' version, leading to a divergence between clients.

Instead, now both the client and server have the concept of a `version` for media. When you paste media, it starts out at version `1`. Currently, it can only be bumped to a new version when you click the "refresh" button -- this sets the version to be the current timestamp as a number.

The net effect of this is that when you fetch the new "version" from the server, it forces everything to roll over to new cache keys. (So the screenshot is re-fetched and all the intermediate cached values are passed over.) Version numbers are also burned into pins. So this means that if you pin two separate links and have refreshed them at different points in time, they will remain in the right state upon re-entry (we will have two separate copies of the screenshot, one for each version.) The only edge case is that after 72 hours *all* of the underlying cached data is removed, so ultimately pinning links will converge on showing the latest version of a site. (Which seems fine.)

Some other downstream effects of this:
- The texture cache and in-flight texture lookup for image loading now needs to be keyed on <url, version>
- When pasting a new media object via `addMedia`, to get the right behavior, we check for any existing entities in the scene with the given `src` and use the latest seen `version` for the new element. For example, if you paste a page, refresh, and then paste it again, we want to use the latest version (the refreshed version), not version 1, which would be a old screenshot.
- Pinned `media` components have a `version`, which is somewhat unfortunate to add durable state like this but it seems correct.


gfodor:
Note that we could also add a refresh button to other media to get the same behavior. This could be particularly useful on sketchfab assets a person is working on: our cache aggressively holds them for 72 hours, so if you were to publish to sketchfab there's no way to get your latest updates unless we had this refresh button.

src/components/media-loader.js
```diff
@@ -257,13 +259,38 @@ AFRAME.registerComponent("media-loader", {
     }
   },
 
+  refresh() {
+    if (this.networkedEl && !NAF.utils.isMine(this.networkedEl) && !NAF.utils.takeOwnership(this.networkedEl)) return;
```
netpro2k: Hmm, if you hit refresh before `networkEl` existed I think you would end up not exiting early and would bump the version. May be fine but a bit weird.

src/components/media-loader.js
```diff
@@ -257,13 +259,38 @@ AFRAME.registerComponent("media-loader", {
...
+
+    if (!src) return;
+    const srcChanged = oldData.src !== src;
+    const versionChanged = !!(oldData.version && oldData.version !== version);
+
+    if (versionChanged) {
+      this.el.emit("media_refreshing");
+
+      // Don't animate if its a refresh.
+      this.data.animate = false;
```
netpro2k: Technically shouldn't be mutating the data object directly like this... but we are so far off the normal aframe path these days not sure if it would be better to do this or setAttribute.

src/components/media-loader.js
```diff
@@ -257,13 +259,38 @@ AFRAME.registerComponent("media-loader", {
...
+
+      // Play the sound effect on a refresh only if we are the owner
+      this.data.playSoundEffect = NAF.utils.isMine(this.networkedEl);
+    }
+
     try {
-      const { src, contentSubtype } = this.data;
+      const { src, version, contentSubtype } = this.data;
```
netpro2k: src and version are already declared in this scope.

src/components/media-loader.js
```diff
@@ -257,13 +259,38 @@ AFRAME.registerComponent("media-loader", {
...
 
       if (!src) return;
```
netpro2k: This is already checked above.

src/components/media-views.js
```diff
@@ -828,34 +833,38 @@ AFRAME.registerComponent("media-image", {
...
+      if (textureCache.has(src, version)) {
         if (this.currentSrcIsRetained) {
-          cacheItem = textureCache.get(src);
+          cacheItem = textureCache.get(src, version);
         } else {
-          cacheItem = textureCache.retain(src);
+          cacheItem = textureCache.retain(src, version);
         }
       } else {
+        const inflightKey = `${src}_${version}`;
```
netpro2k:
```suggestion
        const inflightKey = textureCache.key(src, version);
```

src/components/media-views.js
```diff
@@ -890,6 +899,12 @@ AFRAME.registerComponent("media-image", {
 
     const projection = this.data.projection;
 
+    if (this.mesh && this.data.batch) {
+      // This is a no-op if the mesh was just created.
+      // Otherwise we want to ensure the texture gets updated.
+      batchManagerSystem.removeObject(this.mesh);
```
netpro2k: Can't you just do this above where `textureCache.release(src, version);` is being called.

src/utils/media-utils.js
```diff
@@ -91,6 +91,21 @@ function getOrientation(file, callback) {
   reader.readAsArrayBuffer(file);
 }
 
+function getLatestMediaVersionOfSrc(src) {
```
netpro2k: This feels a bit hairy. Also as written you would start back at version 1 after deleting all instances of a given src. Seems like this should be part of the resolver cache (or better yet part of resolving media on the backend, though that might be overkil)

src/components/media-loader.js
gfodor: this exits early if you call `refresh()` before `networkedEl`, no?

src/utils/media-utils.js
gfodor: the problem here is the resolver cache is node-local, and non durable. so we have no durable, cross-node information about what the last-requested version is for a URL. without adding the durability, it means that information is lost when the last instance of an object is removed.

this behavior I wrote here seems better than *always* starting at version 1, since it's often the case you want to do a simple operation like clone and have it do what you expect, instead of handing you an old version.

so if we wanted to fix this, we'd need to start persisting the resolver cache to the database, which seems like a bad idea. it also would be a different cache than what we have now, which is `<url, version> -> resolved url`, and this cache is `url -> last version`

src/components/media-loader.js
gfodor: ah whoops, had extracted this to a function and reverted but left this bit.

src/components/media-loader.js
gfodor: oh right. i think i'll leave it, because i'm not sure what the behavior is if we do that within `update()`

src/components/media-views.js
gfodor: I don't think so, that release i think is handling the specific case where one of the promises is still resolving since the top of the same function call, where the local `src` and `version` were bound. in this case, we want to un-bind any existing texture even it was bound in a previous call site.

the other place to do it was at the very beginning when we check for the src/version to have changed, which I tried, but the problem there is that it results in the texture blinking for a second, as this function yields to several promises between here and there. by doing the remove later (here), the swap happens (ideally) intra-frame.

PR #2023, opened 2020-01-09
Hotkey to hide the UI

gfodor:
You can now hit the tilde key (`~`) to hide the UI and get a better view.

5f7kBBwun7

### DIFF
--- a/src/react-components/ui-root.js
+++ b/src/react-components/ui-root.js
@@ -309,7 +309,7 @@ class UIRoot extends Component {
     this.props.scene.addEventListener("share_video_failed", this.onShareVideoFailed);
     this.props.scene.addEventListener("exit", this.exitEventHandler);
     this.props.scene.addEventListener("action_exit_watch", () => this.setState({ watching: false, hide: false }));
-    this.props.scene.addEventListener("action_advance_lobby", () => this.setState({ watching: false, hide: false }));
+    this.props.scene.addEventListener("action_toggle_ui", () => this.setState({ hide: !this.state.hide }));
 
     const scene = this.props.scene;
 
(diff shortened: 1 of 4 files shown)


src/react-components/ui-root.js
```diff
@@ -309,7 +309,7 @@ class UIRoot extends Component {
     this.props.scene.addEventListener("share_video_failed", this.onShareVideoFailed);
     this.props.scene.addEventListener("exit", this.exitEventHandler);
     this.props.scene.addEventListener("action_exit_watch", () => this.setState({ watching: false, hide: false }));
-    this.props.scene.addEventListener("action_advance_lobby", () => this.setState({ watching: false, hide: false }));
```
gfodor: note this old event was mistakenly committed in a prior PR

src/systems/ui-hotkeys.js
```diff
@@ -48,6 +48,12 @@ AFRAME.registerSystem("ui-hotkeys", {
     if (this.userinput.get(paths.actions.toggleCamera)) {
       this.el.emit("action_toggle_camera");
     }
+
+    if (this.userinput.get(paths.actions.toggleUI)) {
+      if (this.el.sceneEl.is("entered")) {
```
johnshaughnessy: May be useful to hide the UI from the lobby, too

gfodor: Yeah, I had considered that originally, but it felt a bit weird since that UI is more of a 'landing page' than "HUD" chrome

EgleXR: On the Spanish keyboard, what would the key be?

gfodor: it's wired right now to the backtick character. colloquially this is usually referred to as the tilde key in US

PR #869, opened 2019-02-03
[WIP] Fixes #307: Added HashRouter in index.js

suhailsinghbains:
Fixes #307: Added HashRouter in index.js
I was hoping if I could do partial work for now and do rest after a review :)

PR #2028, opened 2020-01-10
Use HashRouter for home page dialogs

brianpeiris:
An updated version of https://github.com/mozilla/hubs/pull/869

Uses react-router's HashRouter for the Join Us and Report dialogs. The report dialog was previously handled by `/?report`. 

One tiny downside to this is that HashRouter will always append `/#/` to the root URL.


src/index.js
```diff
@@ -42,11 +45,25 @@ const remountUI = function() {
...
+
+  function returnToRoot() {
+    location.href = "/#/";
+  }
+
+  const router = (
+    <HashRouter>
+      <>
+        {root}
+        <Route path="/joinus" render={() => <JoinUsDialog onClose={returnToRoot} />} />
```
gfodor: should on close just perform the back button?

src/react-components/home-root.js
```diff
@@ -283,29 +273,19 @@ class HomeRoot extends Component {
               <div className={styles.links}>
                 <div className={styles.top}>
                   <IfFeature name="show_join_us_dialog">
-                    <a
-                      className={styles.link}
-                      rel="noopener noreferrer"
-                      href="#"
-                      onClick={this.onLinkClicked(this.showJoinUsDialog)}
-                    >
+                    <a className={styles.link} rel="noopener noreferrer" href="/#/joinus">
```
gfodor: nit, i'd make this `join-us`

src/index.js
brianpeiris: It would be a bit surprising for users who happen to arrive to these dialogs from another to click on a close button and have them go back.

PR #2029, opened 2020-01-10
Fade between scenes

gfodor:
Scenes will now fade out between switches.

### DIFF
--- a/src/components/fader.js
+++ b/src/components/fader.js
@@ -1,0 +1,69 @@
+/**
+ * Creates a box around the element (assumed to be the camera's PoV) which can be used for fade-to-black.
+ */
+
+const FADE_DURATION_MS = 750;
+
+AFRAME.registerComponent("fader", {
+  schema: {
+    direction: { type: "string", default: "none" } // "in", "out", or "none"
+  },
+
+  init() {
+    const mesh = new THREE.Mesh(
+      new THREE.BoxGeometry(),
+      new THREE.MeshBasicMaterial({ color: 0x0, side: THREE.BackSide, opacity: 0, transparent: true })
+    );
+    mesh.scale.x = mesh.scale.y = mesh.scale.z = 0.05;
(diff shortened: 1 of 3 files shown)


src/components/fader.js
```diff
@@ -0,0 +1,69 @@
...
+
+  init() {
+    const mesh = new THREE.Mesh(
+      new THREE.BoxGeometry(),
+      new THREE.MeshBasicMaterial({ color: 0x0, side: THREE.BackSide, opacity: 0, transparent: true })
+    );
+    mesh.scale.x = mesh.scale.y = mesh.scale.z = 0.05;
+    mesh.matrixNeedsUpdate = true;
+    this.el.object3D.add(mesh);
+    this.mesh = mesh;
```
netpro2k: Was going to suggest starting the mesh invisible, but looks like it would be hidden in the first tick anyway.

src/components/fader.js
```diff
@@ -0,0 +1,69 @@
...
+  },
+
+  async fadeOut() {
+    this.el.setAttribute("fader", { direction: "out" });
+
+    return new Promise(res => {
+      if (this.mesh.material.opacity === 1) {
+        res();
+      } else {
+        this._resolveFinish = res;
```
netpro2k: I can already see the bug happening where you call fade in/out before a previous promise has resolved and having the previous one never resolve... Maybe keep an array of pending resolves? or immediately reject a previous one if it is interrupted...

gfodor: ah yeah, i was going to do the array thing which seemed overly complex, but i like the idea of just rejecting if you try to overwrite the existing promise.

src/components/fader.js
gfodor: yeah, that's how I had it originally, but it resulted in it never being drawn. i'm not sure why, probably something having to do with the matrix update loop or something.

PR #2038, opened 2020-01-14
Object List Navigation

gfodor:
You can now quickly move between objects in the object list via navigation buttons, and also remove all the objects in the room one at a time.

BV16M7dewt

This PR also:

- tightens up a few design things like increasing the size of the close button and shortening the title so that we don't use excessive vertical real estate on mobile.

- Re-writes the code we use to track media objects to no longer use DOM mutation events, which were firing in many cases when the actual object list did not change.

- Moves `pinned` out of the component state for the object dialog, and pushes it up to a prop, since it was getting difficult to manage as the user navigated through objects.


src/systems/listed-media.js
```diff
@@ -0,0 +1,25 @@
+AFRAME.registerSystem("listed-media", {
```
johnshaughnessy: Good call. I like this much better than using the mutation observer.

Issue #2041, opened 2020-01-14
Support Pico devices


jvonitter:
**Is your feature request related to a problem? Please describe.**
Currently when visiting Hubs on a Pico device, it looks like I'm using a Quest instead.

**Describe the solution you'd like**
Controller models and strings should show Pico versions.

**Additional context**
FxR support for Pico devices (G2, G3 and Neo 2) is planned for Feb 21 release.

bluemarvin:
Currently we use the Oculus strings for controllers so the PicoVR version will work in hubs. The string we were planning to use were for Neo 2: `Pico Neo 2 (Right)` `Pico Neo 2 (LEFT)`
And for G2: `Pico G2 Controller`

johnshaughnessy:
We can add support for those device ID's in hubs and interpret their inputs the same way we do for other controllers (quest controllers or 3DOF controllers, depending on what they are capable of). We will be able to add better support when we get access to those devices, so if we can get some in mountain view or at our upcoming meeting in Berlin that would help a lot!

johnshaughnessy:
Nevermind - they work the same as the existing quest and go controllers, which means we just needed to look for the new device names.

PR #2042, opened 2020-01-14
Add support for pico devices

johnshaughnessy:
Fixes https://github.com/mozilla/hubs/issues/2041

Adds checks for the ID's for pico controllers: https://github.com/MozillaReality/FirefoxReality/blob/experimental/api/app/src/picovr/cpp/DeviceDelegatePicoVR.cpp#L299


gfodor:
Maybe just change this to `Pico `?  I suspect there will be more and more models

gfodor:
o/w lgtm

johnshaughnessy:
We need to at least differentiate between the 3DOF devices and the 6DOF devices (which both start with `Pico `)

PR #2043, opened 2020-01-15
Add an experimental "fix origin" feature for oculus touch controllers

johnshaughnessy:
This PR adds an (experimental) feature that allows people to synchronize the virtual environment with their real environment. 
This feature is enabled with the `reo` query string parameter.
This feature only works with oculus touch controllers.
To perform the synchronization, two people put their right controllers next to each other on the floor in the same orientation, with the controllers resting on the "ring" part. Then they press the "b" button. If the operation succeeds, then their right in-game hands will have been teleported to the TARGET_CONTROLLER_POSITION (the origin). They will still be one-controller's-width out of sync unless the second person slides their controller along the ground to occupy the exact same position (in real life) that the first person's controller did after the first person presses the button.

### DIFF
--- a/src/hub.js
+++ b/src/hub.js
@@ -111,6 +111,7 @@ import "./components/periodic-full-syncs";
 import "./components/inspect-button";
 import "./components/set-max-resolution";
 import { sets as userinputSets } from "./systems/userinput/sets";
+import "./utils/pretty-print";
 
 import ReactDOM from "react-dom";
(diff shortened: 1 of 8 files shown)


gfodor:
nice!

johnshaughnessy:
Closing for newer PR: https://github.com/mozilla/hubs/pull/2107

Issue #1880, opened 2019-11-02
360 Videos from Youtube are Low Resolution


emclaren:
**Is your feature request related to a problem? Please describe.**
360 Youtube videos are pixelated when you use them as scenes in Hubs

For example 
360 Hubs scene: https://hubs.mozilla.com/scenes/Lcb9dNa/360-music-video-large
vs
360 Youtube video original: https://www.youtube.com/watch?v=6M1oopfKnLw

**Describe the solution you'd like**
It would be nice if you could use 360 degree videos from Youtube without losing as much quality 

**Additional context**
This issue that has been raised by several community members

PR #2044, opened 2020-01-15
Improve Quality of 360 Youtube Videos

gfodor:
Goes with: https://github.com/mozilla/reticulum/pull/303

Fixes https://github.com/mozilla/hubs/issues/1880

Utilizes the new audio track resolution to greatly improve 360 video quality from youtube.

### DIFF
--- a/src/components/media-views.js
+++ b/src/components/media-views.js
@@ -108,7 +108,7 @@ async function createGIFTexture(url) {
  * @param {string} src - Url to a video file.
  * @returns {Element} Video element.
  */
-async function createVideoOrAudioEl(type) {
+function createVideoOrAudioEl(type) {
   const el = document.createElement(type);
   el.setAttribute("playsinline", "");
   el.setAttribute("webkit-playsinline", "");
@@ -481,9 +481,9 @@ AFRAME.registerComponent("media-video", {
       this.mesh.material.needsUpdate = true;
     }
 
-    let texture, audioSource;
+    let texture, audioSourceEl;
     try {
-      ({ texture, audioSource } = await this.createVideoTextureAndAudioSource());
+      ({ texture, audioSourceEl } = await this.createVideoTextureAudioSourceEl());
 
       // No way to cancel promises, so if src has changed while we were creating the texture just throw it away.
       if (this.data.src !== src) {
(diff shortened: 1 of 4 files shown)


src/components/media-views.js
```diff
@@ -108,19 +108,19 @@ async function createGIFTexture(url) {
  * @param {string} src - Url to a video file.
  * @returns {Element} Video element.
  */
-async function createVideoEl() {
-  const videoEl = document.createElement("video");
-  videoEl.setAttribute("playsinline", "");
-  videoEl.setAttribute("webkit-playsinline", "");
+async function createVideoOrAudioEl(type) {
```
netpro2k: I know this is not new.. but why is this function async? Doesn't seem to return a promise or do anything async.

src/components/media-views.js
```diff
@@ -700,13 +706,33 @@ AFRAME.registerComponent("media-video", {
...
+
+          this._audioSyncInterval = setInterval(() => {
+            if (Math.abs(audioEl.currentTime - videoEl.currentTime) >= 0.33) {
+              // In Chrome, drift of a few frames seems persistent
+              audioEl.currentTime = videoEl.currentTime;
+            }
+
+            if (videoEl.paused !== audioEl.paused) {
+              videoEl.paused ? audioEl.pause() : audioEl.play();
+              audioEl.currentTime = videoEl.currentTime;
```
netpro2k: Why do this again here? You would have just synced it above it they had diverged at all.

src/components/media-views.js
```diff
@@ -480,9 +481,9 @@ AFRAME.registerComponent("media-video", {
       this.mesh.material.needsUpdate = true;
     }
 
-    let texture;
+    let texture, audioSource;
     try {
-      texture = await this.createVideoTexture();
+      ({ texture, audioSource } = await this.createVideoTextureAndAudioSource());
```
netpro2k: Was at first a bit confused by `audioSource` since its not an audio source but infact a video or audio element... Might be cleaner to have the audio source get created in that function, but this is super minor.

src/components/media-views.js
gfodor: no idea, changing it.

src/components/media-views.js
gfodor: I did it here because it gives us an opportunity to correct any drift that's accrued but still below the amount we require to do a re-sync.

Issue #1903, opened 2019-11-15
Custom logo squished in loading panel


antpb:
**Is your feature request related to a problem? Please describe.**
I noticed that the width and height of the loading panel logo are defined. This causes issues when folks with, for instance, square logos use a custom logo. 

Here's where that is defined: https://github.com/mozilla/hubs/blob/fc21aedb5b1393aaf3aa6910df8dcf7b4255e88f/src/assets/stylesheets/loader.scss#L38-L43

**Describe the solution you'd like**
Maybe a `max-height:` here instead so various dimensions are respected.

If there is a max-height that can be agreed in this issue, I'd be happy to make a PR. :)

Here's what I'm doing it in my fork deployed at https://hubs.broken.place : 

https://github.com/antpb/hubs/blob/94778ae7fb51d5b1c84c030dce9f202cd749e4b6/src/assets/stylesheets/loader.scss#L38-L41

antpb:
Also, not sure if this is feature request or bug so siding on feature request until I know for sure if this was intended. :)

brianpeiris:
I'm working on related things right now and I'll include this fix. Thanks for the report!

brianpeiris:
Fixed in #1916

PR #2046, opened 2020-01-15
Removes defined `height` to account for images with different aspect ratios

antpb:
At <hubs-domain>/link the logo has a defined height that causes it to be squished. 
I opened an issue similar to this at: https://github.com/mozilla/hubs/issues/1903#event-2812760998

This PR removes the height of the logo displayed on the `/link` page to allow for other image aspect ratios in custom instances of Hubs.

Here's what it looks like in mine without patch:
 
Screen Shot 2020-01-15 at 5 41 38 PM

### DIFF
--- a/src/assets/stylesheets/link.scss
+++ b/src/assets/stylesheets/link.scss
@@ -50,7 +50,6 @@ a {
 
   img {
     width: 247px;
-    height: 57px;
   }
 
   @media(max-height: 600px) {


gfodor:
this seems like a good change to me, @brianpeiris might want to provide some input on the overall approach to images in hubs cloud -- I'm wondering if we should just standardize on fixed-width, variable-height, or something else?

brianpeiris:
Fixed with, variable height seems right to me, but we should note somewhere in the docs or configuration page that tall images will probably break the UI.
Thanks for the PR!

brianpeiris:
Sorry it took so long to merge this!

PR #2053, opened 2020-01-17
Object menu improvements

johnshaughnessy:
This PR improves the menu placement algorithm.

object-menus

 - Menus are oriented towards the user so they are easier to read.
 - Menus are positioned such that they are "pushed back" against the objects' meshes via Ammo.js's `convexSweepTest`.
 - Meshes compute bounding volume information only once, instead of every time the menus open.
 - Menu opening animations no longer "jump" after completion, and scaling is smooth in general while menus are open.


johnshaughnessy:
Two important bugs have not yet been figured out. 
- The first bug can reproduced by scaling an object down, cloning it, then opening the menu on the clone. The clone's menu then either intersects the clone's mesh. Pinning the object, refreshing the room, then opening the (cloned) object's menu will show it at the correct position.
- The second bug results in a menu positioned off-center incorrectly. I do not know how to reproduce it. It may coincide with a bug where the object jumps from a surprising location (the origin?) to its correct location after what seems like a single frame. It is difficult to tell if I am seeing it correctly.

johnshaughnessy:
>The first bug can reproduced by scaling an object down, cloning it, then opening the menu on the clone. The clone's menu then either intersects the clone's mesh. Pinning the object, refreshing the room, then opening the (cloned) object's menu will show it at the correct position.

This issue seems to be related to the physics bodies/shapes that are generated for the cloned objects:
image
On the left is a model and its hull at default size.
In the middle is a clone of the model on the left, which was scaled down after cloning.
On the right is a clone of the model in the middle, which has an incorrectly scaled physics shape.

src/systems/menu-placement-system.js
```diff
@@ -0,0 +1,494 @@
...
+})();
+
+function matchRootEl(rootEl) {
+  return function match(tuple) {
+    return tuple.rootEl === rootEl;
+  };
+}
+
+// TODO: Computing the boundaries of skinned/animated meshes does not work correctly.
+// TODO: Cloning small objects sometimes results in the menu placement failing by
```
johnshaughnessy: Can remove this `TODO` https://github.com/mozilla/hubs/issues/2063

johnshaughnessy:
Another bug happens if I scale the loading cube before the model finishes loading. In that case, the physics shape for the model appears to be calculated correctly, and yet the menu is still positioned incorrectly. 

image

I will try to figure out what is going wrong here...

johnshaughnessy:
Limitations, known bugs, and performance improvement recommendations are documented as `TODO`s. Performance testing suggests that the current implementation does not become a hotspot until several menus are displayed at once: ~0.16ms/frame for 1 open menu, ~0.50ms/frame for 10 open menus. (This was on a laptop. I still need to test mobile devices.) Reducing the frequency of calling `convexSweepTest` (as well as trying some of the other recommendations) should mitigate any performance issues considerably.

gfodor:
One thing worth considering (that I believe we did in the old algorithm) was to schedule work so that you process at most N menus per frame, to ensure there's a fixed maximum time budget spent on menus. Eg you could run your convexSweepTest once per frame and as more menus are added you just introduce latency for snapping as the user moves around.

We have a utility for this: https://github.com/mozilla/hubs/blob/master/src/systems/frame-scheduler.js

johnshaughnessy:
Closing in favor of https://github.com/mozilla/hubs/pull/2159

PR #2059, opened 2020-01-17
Public Rooms

netpro2k:
Adds ability to make rooms public, which means they will show up on the homepage. By default only admins can do this, but it can be enabled for all room owners through a config flag. Room privacy is configured via the Rooms Settings screen. 

The homepage will now also show participant counts for favorited (and public) rooms.

todo:
- [x] Hide privacy settings if you do not have permission to set them
- [x] Merge reticulum changed https://github.com/mozilla/reticulum/pull/304


src/index.js
```diff
@@ -76,6 +76,28 @@ window.addEventListener("beforeinstallprompt", e => {
   }
 });
 
+async function fetchFeaturedRooms() {
+  // Fetch favorite rooms
```
gfodor: comment probably should say "fetch favorite + public rooms and merge"

src/index.js
```diff
@@ -76,6 +76,28 @@ window.addEventListener("beforeinstallprompt", e => {
...
+  const [favoriteRoomsResult, publicRoomsResult] = await Promise.all([
+    authChannel.signedIn
+      ? fetchReticulumAuthenticated(
+          `/api/v1/media/search?source=favorites&type=hubs&user=${store.credentialsAccountId}`
```
gfodor: we probably should update this API now to be `type=rooms`

src/index.js
```diff
@@ -76,6 +76,28 @@ window.addEventListener("beforeinstallprompt", e => {
...
+        )
+      : Promise.resolve({ entries: [] }),
+    fetchReticulumAuthenticated("/api/v1/media/search?source=public_rooms")
+  ]);
+
+  if (favoriteRoomsResult && publicRoomsResult) {
```
gfodor: when would this `if` be false? just in error condition?

src/index.js
```diff
@@ -76,6 +76,28 @@ window.addEventListener("beforeinstallprompt", e => {
...
+    publicRoomsResult.entries.push(...favoriteRoomsResult.entries);
+    const ids = publicRoomsResult.entries.map(h => h.id);
+    publicRoomsResult.entries = publicRoomsResult.entries
+      .filter((h, i) => ids.lastIndexOf(h.id) === i)
+      .sort((a, b) => b.participant_count - a.participant_count);
+    featuredRoomsResult = publicRoomsResult;
```
gfodor: this is a little weird, because the result we're passing to the UI is presumed to be a single API result, whereas this is not really an API result but a weird data structure that is two API results mashed together. I'd refactor the prop to just be the entries and update the value here to be the result of the `filter.sort`. if there are other things being accessed in `meta` for example in the subcomponent (I doubt it) then we should lift those to props as well.

src/index.js
```diff
@@ -110,4 +126,7 @@ window.addEventListener("beforeinstallprompt", e => {
 
   hideHero = false;
   remountUI();
+
+  await fetchFeaturedRooms();
+  setInterval(fetchFeaturedRooms, 30000);
```
gfodor: oh, this is a problem for hubs cloud now that I think about it. we gotta be really careful here, because if we have anything like this where a page, when left open, continually accesses the db, it means a single open page can keep their aurora instance hot. we now have an idle timeout in hubs and the admin console for this reason.

instead, we should either skip the polling, or stop polling after some reasonable interval like 15 minutes. i think just skipping it is probably the right tradeoff. note that the 'right' solution here would be to open up a new channel to fetch presence data that has no db access in its read path, so the database could shut down *and* we could get real time updates.

netpro2k: Just removed the polling for now.

src/assets/translations.data.json
```diff
@@ -51,6 +51,7 @@
     "room-settings.spawn_camera": "Create cameras",
     "room-settings.spawn_drawing": "Create drawings",
     "room-settings.pin_objects": "Pin objects",
+    "room-settings.room_privacy": "Room privacy",
```
gfodor: I think I'd call this `Room Access` from user's perspective

src/hub.js
```diff
@@ -266,10 +266,10 @@ let routerBaseName = document.location.pathname
   .join("/");
 
 if (document.location.pathname.includes("hub.html")) {
-  routerBaseName = "";
+  routerBaseName = "/";
 }
 
-const history = createBrowserHistory({ basename: routerBaseName });
+const history = routerBaseName === "/" ? createMemoryHistory() : createBrowserHistory({ basename: routerBaseName });
```
gfodor: comment probably worthwhile here, this could be confusing for people who forget that we offer the default room now

src/index.js
```diff
@@ -76,6 +76,25 @@ window.addEventListener("beforeinstallprompt", e => {
   }
 });
 
+// Fetch favorite + public rooms and maerge, sorting by participant count
```
gfodor: `s/maerge/merge`

src/index.js
```diff
@@ -76,6 +76,25 @@ window.addEventListener("beforeinstallprompt", e => {
...
+  const [favoriteRoomsResult, publicRoomsResult] = await Promise.all([
+    authChannel.signedIn
+      ? fetchReticulumAuthenticated(
+          `/api/v1/media/search?source=favorites&type=rooms&user=${store.credentialsAccountId}`
+        )
+      : Promise.resolve({ entries: [] }),
+    fetchReticulumAuthenticated("/api/v1/media/search?source=public_rooms")
+  ]);
+
+  publicRoomsResult.entries.push(...favoriteRoomsResult.entries);
```
gfodor: nit, but I think down the road pushing these here could lead to bugs. i'd create a new list to concat them into with a proper name. (eg, if anyone comes along here and adds code below this point assuming that list contains only public rooms, would introduce bugs)

src/react-components/home-root.js
```diff
@@ -218,11 +219,8 @@ class HomeRoot extends Component {
...
             <div className={styles.heroContent} style={{ backgroundImage: configs.image("home_background", true) }}>
               {!this.props.hideHero &&
-                (this.props.favoriteHubsResult &&
-                this.props.favoriteHubsResult.entries &&
-                this.props.favoriteHubsResult.entries.length > 0 &&
-                this.state.signedIn
-                  ? this.renderFavoriteHero()
+                (this.props.featuredRooms && this.props.featuredRooms.length > 0
+                  ? this.renderFeaturedRoomsHero()
                   : this.renderNonFavoriteHero())}
```
gfodor: this function seems worth renaming now

src/react-components/home-root.js
```diff
@@ -44,7 +44,8 @@ class HomeRoot extends Component {
     installEvent: PropTypes.object,
     hideHero: PropTypes.bool,
     showAdmin: PropTypes.bool,
-    favoriteHubsResult: PropTypes.object,
+    featuredRooms: PropTypes.array,
+    publicRoomsResult: PropTypes.object,
```
gfodor: this prop is dead code i think

src/react-components/home-root.js
```diff
@@ -379,13 +377,25 @@ class HomeRoot extends Component {
...
           <div className={classNames([mediaBrowserStyles.box, mediaBrowserStyles.darkened])}>
-            <MediaTiles result={this.props.favoriteHubsResult} urlSource="favorites" />
+            <MediaTiles entries={this.props.featuredRooms} urlSource="favorites" />
           </div>
         </div>
       </div>
     ];
   }
 
+  renderPublicRoomsHero() {
```
gfodor: dead code i think

src/react-components/media-browser.js
```diff
@@ -282,6 +282,13 @@ class MediaBrowser extends Component {
     const activeFilter =
       searchParams.get("filter") || (searchParams.get("similar_to") && "similar") || (!searchParams.get("q") && "");
 
+    const meta = this.state.result && this.state.result.meta;
```
gfodor: might want to be careful here, if `result` is set but `meta` is not, then `meta === this.state.result`, so any keys on `result` that align with `meta` (now or ever) would cause bugs. i'd suggest ensuring `meta` is set to null in the case this expression is falsey.

src/react-components/media-browser.js
```diff
@@ -282,6 +282,13 @@ class MediaBrowser extends Component {
...
+    const hasNext = !!(meta && meta.next_cursor);
+    const hasPrevious = searchParams.get("cursor");
+    const page = (meta && meta.page) || 0;
+    const apiSource = (meta && meta.source) || null;
+    const isVariableWidth = ["bing_images", "tenor"].includes(apiSource);
```
gfodor: i'd suggest pushing api source as a prop and moving the variable width concern back into the media tiles component, since it's a rendering concern

src/react-components/media-tiles.js
```diff
@@ -243,7 +243,20 @@ class MediaTiles extends Component {
                 <FontAwesomeIcon icon={faPencilAlt} />
               </a>
             )}
+          {entry.type === "room" && (
+            <div className={styles.info}>
+              <FontAwesomeIcon icon={faUsers} />
+              <span>{entry.participant_count}</span>
```
gfodor: probably worth the `|| 0` for backwards compatibility, with old reticulum servers. not usually something to worry about (only affects up-to-date forked clients running old ret servers), but it's so trivial here seems worthwhile.

src/react-components/room-settings-dialog.js
```diff
@@ -26,6 +27,31 @@ export default class RoomSettingsDialog extends Component {
...
+          value={this.state.privacy}
+          onChange={
+            onChange ||
+            (e =>
+              this.setState({
+                privacy: e.target.value
+              }))
+          }
+        >
+          <option value="public">Public (Listed on the homepage)</option>
```
gfodor: two points here:
- need to push these into translation strings
- I didn't load the design, but I would update it so that these are laid out so "Public" and "Private" are top level radios with label color (black or white depending on theme), and then the "Visible on homepage" and "Only those with the link can join" are subtitle font/color underneath each label

src/schema.toml
```diff
@@ -25,6 +25,9 @@ features.show_avatar_pipelines_link = { category = "features", type = "boolean",
 features.show_model_collection_link = { category = "features", type = "boolean", name = "Show Model Collection Link", description = "Show link to a collection of models." }
 features.hide_powered_by = { category = "features", type = "boolean", name = "Hide Powered By Hubs", description = "Hides the 'Powered by Hubs Cloud' notice on the home page and loading screens." }
 
+features.public_rooms = { category = "features", type = "boolean", name = "Public Rooms", description = "Allow non-admin users to create public rooms that will be listed on the homepage." }
```
gfodor: I'd change this to `Allow visitors to set rooms to public to feature them on the homepage.`

src/schema.toml
```diff
@@ -25,6 +25,9 @@ features.show_avatar_pipelines_link = { category = "features", type = "boolean",
...
+features.default_room_id = { category = "features", type = "string", name = "Homepage Room", description = "Use the specified room in place of the homepage."}
```
gfodor: probably need to say "room ID" here?

src/utils/configs.js
```diff
@@ -46,18 +46,23 @@ if (window.APP_CONFIG) {
...
-  const features = configs.APP_CONFIG && configs.APP_CONFIG.features;
-
-  const forceEnableSpoke = featureName === "enable_spoke" && isAdmin;
-
-  return forceEnableSpoke || enableAll || (features && features[featureName]);
+  const value = configs.APP_CONFIG.features[featureName];
+  if (typeof value === "boolean") {
+    const enableAll =
+      isLocalDevelopment && !process.env.USE_FEATURE_CONFIG && !enableAllBlacklist.includes(featureName);
+    const forceEnableSpoke = featureName === "enable_spoke" && isAdmin;
```
gfodor: see above: the check around `public_rooms` likely should be made a permission check, and logic surrounding it in the client should be removed.

in general, it seems like a anti-pattern for us to be checking `isAdmin` anywhere in the client insofar as that flag confers some rights on the server. (it still should be sane to check a users admin status in the client, when that admin status can't or isn't involved in privileges but "UI convention")

src/react-components/ui-root.js
```diff
@@ -1671,7 +1672,12 @@ class UIRoot extends Component {
               history={this.props.history}
               render={() =>
                 this.renderDialog(RoomSettingsDialog, {
-                  initialSettings: { name: this.props.hubName, member_permissions: this.props.hubMemberPermissions },
+                  showPrivacySettings: configs.isAdmin() || configs.feature("public_rooms"),
```
gfodor: we should revisit this -- this probably needs to be a permission check and not config driven. the reticulum server should carve out and return a permission for being able to make rooms public.

gfodor:
Oh also, note that lint is failing

gfodor:
Ah you already have most of what you need on the server:
- You should probably rename `update_hub_privacy` to `update_hub_privacy_to_public`, since we will eventually have other privacy mode changes that we allow regardless (like `invite_only`)
- Expose the permission in the client's JWT, and check it accordingly

Issue #2052, opened 2020-01-16
Performance regression in character controller / pathfinding


johnshaughnessy:
I introduced a perf regression when changing the character controller. It recomputes the group and closest node every frame here: https://github.com/mozilla/hubs//blob/master/src/systems/character-controller-system.js#L330
Which ends up doing some significant work (and garbage generation)  if the nav mesh has a lot of vertices, e.g. it calls this function :
https://github.com/MozillaReality/three-pathfinding/blob/master/src/Utils.js#L218

I think the fix is for us to be more judicious about when we need to recompute the nav mesh group / closest node (I assume we have to do it when we teleport or when we exit fly mode but besides that I don't know when). We can also patch three-pathfinding to not recreate that polygonVertices array every time. That would reduce garbage when we do recompute the group.

When I made this change, I wanted to recompute the group because there were cases when the auto-generated nav mesh from spoke was putting nearby surfaces into different groups (when I would expect them to end up in the same group). However this is more of a nav mesh generation bug than something the client should try to make up for, since sometimes groups that are close together should _not_ be traversable (e.g. when there is a wall separating the groups).

johnshaughnessy:
Moving this to the top of my list because this is causing some environments to be completely unusable on CPU/memory-constrained devices.

johnshaughnessy:
Not that it matters much at this point, but P0 was a bit of an extreme categorization.

PR #2071, opened 2020-01-22
Fix regression in nav mesh performance

johnshaughnessy:
Fixes https://github.com/mozilla/hubs/issues/2052 by saving the `navGroup` and `navNode` of the previous frame, and only recomputes them from scratch if the user teleports, travels by waypoint, stops flying, or `shouldLandWhenPossible`.

Also fixes a bug where if you happen to be on the zeroth node in the nav mesh, you don't actually travel along the nav mesh. (`this.navNode === 0` is a valid node, which we need to treat differently than `this.navNode === null`)


gfodor:
lgtm! i am going to push this

PR #2072, opened 2020-01-22
Focus on Media

gfodor:
You can now right click-and-hold or hold the F key on desktop or use secondary controller buttons in VR to quickly focus on objects.

I3YSlruSEp

--

Updates the control scheme to have hardware buttons for Focus (The Mode Formerly Known As Inspect):

- On keyboard/mouse, hold the RMB or F key
- On touch, press and hold B or Y
- On vive, press and hold a primary button (untested, will need a volunteer to test)

Also, updates the controls on touch:
- Left stick press toggles experimental camera mode rotation
- Right stick press toggles fly mode

Also fixes a weird issue I saw with exiting a room where the scene node wasn't under the document body -- not sure why, but the new code is clearly more correct.


gfodor:
I feel pretty strongly we should use F for focus, but there's a chance a user may accidentally go into Fly mode if they hit F if not hovering on an object. We could move fly to another key or to Ctrl-F. Not sure, open to ideas.

johnshaughnessy:
I suggest we move toggle-fly-mode to G or V or some other letter. I don't think it matters what key we use, as long as we update the docs. I think it is better for it to be on a different key than focus and since it's modal it doesn't seem important to be on as convenient a key as focus is.

gfodor:
moved it to `g`

src/systems/camera-system.js
```diff
@@ -334,14 +343,27 @@ export class CameraSystem {
       this.viewingCameraRotator.on = true;
 
       this.userinput = this.userinput || scene.systems.userinput;
-      if (
+      this.interaction = this.interaction || scene.systems.interaction;
+
+      if (this.userinput.get(paths.actions.startInspecting) && this.mode !== CAMERA_MODE_INSPECT) {
+        const hoverEl = this.interaction.state.rightRemote.hovered || this.interaction.state.leftRemote.hovered;
```
johnshaughnessy: This will be a bit awkward if ever we have two cursors active at once, but should be fine now since there's only one. I would also double-check that the `this.interaction.state.rightRemote.hovered` state gets cleared when the active remote is switched to the leftRemote. An edge case here that I'm not sure about is if you grab an object with the right remote, then activate your left remote. At that moment I wonder if the right and left remotes are both active and potentially hovering on different elements.

johnshaughnessy: Ah, nevermind -- I see now in the `CursorTogglingSystem` that when the left remote becomes active, the hover state of the right remote gets cleared and vice versa.

PR #2075, opened 2020-01-23
Skip snapping to nav mesh if not moving

gfodor:
- Skip doing a bunch of work in character controller system if not moving
- Skip doing safari check every frame

### DIFF
--- a/src/systems/character-controller-system.js
+++ b/src/systems/character-controller-system.js
@@ -263,12 +263,19 @@ export class CharacterControllerSystem {
         .multiply(snapRotatedPOV);
       const shouldRecomputeNavGroupAndNavNode =
         (didStopFlying || this.shouldLandWhenPossible) && !this.isMotionDisabled;
-      this.findPOVPositionAboveNavMesh(
-        startPOVPosition.setFromMatrixPosition(this.avatarPOV.object3D.matrixWorld),
-        desiredPOVPosition.setFromMatrixPosition(newPOV),
-        navMeshSnappedPOVPosition,
-        shouldRecomputeNavGroupAndNavNode
-      );
+      const triedToMove = displacementToDesiredPOV.lengthSq() > 0.000001;
+      const shouldResnapToNavMesh = shouldRecomputeNavGroupAndNavNode || triedToMove;
+
+      if (shouldResnapToNavMesh) {
+        this.findPOVPositionAboveNavMesh(
+          startPOVPosition.setFromMatrixPosition(this.avatarPOV.object3D.matrixWorld),
+          desiredPOVPosition.setFromMatrixPosition(newPOV),
+          navMeshSnappedPOVPosition,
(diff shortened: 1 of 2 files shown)


src/systems/character-controller-system.js
```diff
@@ -263,12 +263,19 @@ export class CharacterControllerSystem {
         .multiply(snapRotatedPOV);
       const shouldRecomputeNavGroupAndNavNode =
         (didStopFlying || this.shouldLandWhenPossible) && !this.isMotionDisabled;
-      this.findPOVPositionAboveNavMesh(
-        startPOVPosition.setFromMatrixPosition(this.avatarPOV.object3D.matrixWorld),
-        desiredPOVPosition.setFromMatrixPosition(newPOV),
-        navMeshSnappedPOVPosition,
-        shouldRecomputeNavGroupAndNavNode
-      );
+      const triedToMove = displacementToDesiredPOV.lengthSq() > 0.000001;
```
gfodor: note this epsilon was too big before, since we're squaring the values

src/systems/character-controller-system.js
```diff
@@ -250,25 +250,37 @@ export class CharacterControllerSystem {
...
+        calculateDisplacementToDesiredPOV(
+          snapRotatedPOV,
+          this.fly,
+          this.relativeMotion.multiplyScalar(
+            ((userinput.get(paths.actions.boost) ? 2 : 1) * BASE_SPEED * Math.sqrt(playerScale) * dt) / 1000
+          ),
+          displacementToDesiredPOV
+        );
+      } else {
+        displacementToDesiredPOV.set(0, 0, 0);
```
netpro2k: Rather than setting this to 0 and then checking the length below, why not just move all the work into the if, and if you don't have a large enough relative motion, just do nothing.

src/systems/character-controller-system.js
```diff
@@ -250,25 +250,37 @@ export class CharacterControllerSystem {
...
+
+      if (shouldResnapToNavMesh) {
+        this.findPOVPositionAboveNavMesh(
+          startPOVPosition.setFromMatrixPosition(this.avatarPOV.object3D.matrixWorld),
+          desiredPOVPosition.setFromMatrixPosition(newPOV),
+          navMeshSnappedPOVPosition,
+          shouldRecomputeNavGroupAndNavNode
+        );
+      } else {
+        navMeshSnappedPOVPosition.setFromMatrixPosition(newPOV);
```
netpro2k: Shouldn't we just do nothing here? Just having a move that we don't clamp to the navmesh seems wrong. The point of skipping the work to check is that you were effectively standing still.

src/systems/character-controller-system.js
gfodor: we still need to snap to nav mesh if you changed in/out of fly mode, or the land flag is set from a previous frame

Issue #2024, opened 2020-01-10
Cannot record video with sound


johnshaughnessy:
**Description**
When I try to record a video with the in-game camera tool, it gets stuck with 1 second left:
![image](https://user-images.githubusercontent.com/4072106/72114606-04620680-32f9-11ea-9bdf-4901ae369697.png)

I get the following error:
```
NotSupportedError: An audio track cannot be recorded: video/webm; codecs=vp8 indicates an unsupported codec
```
And when I click the "x" button, my client freezes on the next error:
```
TypeError: this.videoRecorder._free is not a functioncamera-tool.js:475:13
    stopRecording camera-tool.js:475
    cancelSnapping camera-tool.js:320
    init camera-tool.js:223
    dispatchEvent three.module.js:172
    tick button-systems.js:21
    tick hubs-systems.js:66
    Aframe 2
    value self-hosted:869
    ko three.module.js:24920
    r three.module.js:7642
    a raven.js:377
```

**To Reproduce**
1. Spawn camera
2. Toggle the sound recording "on" (hover over the camera while in freeze mode to show the option).
3. Try to take a video.

**Expected behavior**
Create a video with sound

**Screenshots**

**Hardware** 
Error found in Firefox desktop on Linux.
Chrome works ok.

emclaren:
Happens for me in Firefox nightly on desktop. Works fine in Chrome.

evmorov:
For FF >= 71 you should use `video/webm;codecs=vp8,opus`

PR #2077, opened 2020-01-23
Fix FF audio camera issues

gfodor:
This fixes two issues with firefox with the camera tool:

- It seems like firefox fails to load mp4 files without audio tracks upon their first load after being created with the camera tool. The browser gives the error: `Media resource https://uploads-prod.reticulum.io/files/2b196e65-4564-4341-8965-5011c05dc9eb.mp4?token=4247fa8700733f31f0bf3865635d9e0b could not be decoded`. As best I could tell, this looked like a browser bug. What's strange is the videos loaded after the second time, or after a fresh page refresh -- the videos only failed to load after being created with the camera. I added a delay to see if it was some server-side race condition, but that did not resolve the problem. Any ideas?

- With FF 71, we need to set a new codec to include audio properly.

Fixes https://github.com/mozilla/hubs/issues/2024

### DIFF
--- a/src/components/camera-tool.js
+++ b/src/components/camera-tool.js
@@ -32,7 +32,7 @@ const isMobileVR = AFRAME.utils.device.isMobileVR();
 const VIEWFINDER_FPS = 6;
 const VIDEO_FPS = 25;
 // Prefer h264 if available due to faster decoding speec on most platforms
-const videoCodec = ["h264", "vp9", "vp8"].find(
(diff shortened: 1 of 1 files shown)


gfodor:
Also the specific error is `NS_ERROR_DOM_MEDIA_METADATA_ERR` and I tested in nightly. Pretty sure this worked previously.

gfodor:
Yeah, I started to, but my reservation is that it seems very quirky and might be Hubs-specific. I didn't want to send them down a wild goose chase -- the fact I can't repro it on a page refresh with a clear cache is really mysterious.

Issue #2061, opened 2020-01-18
Spawning into some scenes cause you to float high in the air mysteriously.


johnshaughnessy:
*EDIT: The initial description of this bug suggests that this is a problem related to spawn points, but this does not seem to be the case. See comment below for updated info! https://github.com/mozilla/hubs/issues/2061#issuecomment-577489805*

**Description**
When you spawn into certain scenes, you start very high above the floor and continue to float upwards. 
I think what's happening is that there is no spawn point and we do not put your character in a reasonable place
Why you float upwards after that is somewhat a mystery to me, but I suspect it's related
I suspect this fix is as simple as, "if there's no spawn points in the scene, spawn somewhere near the origin"

**To Reproduce**
The bug can be experienced in this scene: https://hubs.mozilla.com/scenes/KGMJzEN

**Expected behavior**
We should choose a "reasonable-seeming" spawn location (Either some position on the nav mesh or some place near the origin), and we should ensure your character controller state is set appropriately (e.g. put you in fly mode if there is no nav mesh).

**Screenshots**
![image](https://user-images.githubusercontent.com/4072106/72657633-7242a980-395b-11ea-917e-6161315327d2.png)

**Hardware** 
Any, as far as I can tell.

**Additional context**
It's definitely possible the bug is more complicated, but I'm pretty sure I caused this regression and we need to add a default spawn routine for when there are no spawn points in the scene.

johnshaughnessy:
In this room:
https://hubs.mozilla.com/mkcW2H5/spirited-comfortable-plaza
there is 1 spawn point but the same bug occurs unless I turn disable `snapToNavMesh` on the waypoint. Now I think that this bug will manifest if:
- `snapToNavMesh` is enabled on the waypoint chosen as your spawn point
- there are no spawn points in the scene
- there are only occupiable spawn points in the scene, and they are all occupied.

johnshaughnessy:
OK - I had the wrong idea about this above ^^

The problem does not have to do with waypoints/spawn points (although I still want to double-check that scenes with NO spawn points behave OK).

[...]

PR #2080, opened 2020-01-23
Fix bugs in scenes with no nav mesh.

johnshaughnessy:
This fixes two bugs:
- Fixes https://github.com/mozilla/hubs/issues/2061 
- When switching to a scene without a nav mesh, the client did not remove the previous nav mesh (which would be the one from the loading scene). This explains why one time in standup we switched to a cathedral scene and everyone who was there before could walk on an invisible ground plane (leftover from the loading scene), but the newcomers experienced the bug described in #2061.


src/systems/character-controller-system.js
```diff
@@ -251,14 +261,15 @@ export class CharacterControllerSystem {
...
+      const navMeshExists = NAV_ZONE in this.scene.systems.nav.pathfinder.zones;
       if (!this.isMotionDisabled) {
         const playerScale = v.setFromMatrixColumn(this.avatarPOV.object3D.matrixWorld, 1).length();
         const triedToMove = this.relativeMotion.lengthSq() > 0.000001;
 
         if (triedToMove) {
           calculateDisplacementToDesiredPOV(
             snapRotatedPOV,
-            this.fly,
+            this.fly || !navMeshExists,
```
johnshaughnessy: This makes it so that if there is no nav mesh, the client moves the same way it does when it is in fly mode. However, I chose not to actually enable fly mode (and show a "fly mode enabled" message to the user), so that the user would not have to manually exit fly mode when switching scenes. 

I wonder if we should go even further than this, and disable fly mode toggling when there is no nav mesh. We could show a message like, "cannot toggle fly mode in this scene because there is no floor plan!".

src/systems/nav.js
```diff
@@ -17,5 +35,17 @@ AFRAME.registerSystem("nav", {
     geometry.applyMatrix(mesh.matrixWorld);
     this.pathfinder.setZoneData(zone, Pathfinding.createZone(geometry));
     this.el.sceneEl.emit("nav-mesh-loaded");
+
+    if (this.environmentScene) {
```
johnshaughnessy: I can't imagine `this.environmentScene` here not being set, but guard against it because it set in a promise chain after `waitForDOMContentLoaded`. Should I assume this already exists, or maybe throw an error / warning if it doesn't? Or maybe write a flag like `navMeshLoadedEarly`? I would not expect that flag to ever get set, but am not sure when exactly the DOMContentLoaded callback/chained promise will execute. Advice?

PR #2081, opened 2020-01-23
Inspect Media

gfodor:
You can now inspect media by clicking on the magnifying glass in the menu. Makes it easy to give a presentation with slides or a shared screen, zoom in on shared notes, or watch a videos together!

l88ZxNZjSC

This PR merges in the relevant changes from https://github.com/mozilla/hubs/pull/2072 and https://github.com/mozilla/hubs/pull/2069

This PR has a number of significant changes to support this feature:

- Changes the UX for PDF pagination to use a hover menu instead of the pause menu
- Introduces the concept of "linked media." You can create a 'derivative' media object that is linked to a parent. When you do this:
  - There is a bi-directional sync between the attributes of the media
  - Removing the parent will remove the derivative
- Adds the "mirrored media" concept, which is the internal name for the new inspect feature. (The old inspect feature is still referred to as inspect interally for now, is externally called focus and is now available in the menu an on a hardware button.)
- Adds support for videos to re-use the texture and audio sources from a `linked` attributes passed in. When using inspect with a video, we re-use the existing video element + texture. For the other formats, we make a copy.
- Moves the focus button to be in place of the drop button for 2d media (leaving the drop button now just for 3d media)
- Fixes up the 'refresh' button on links to handle permissions + pinning properly
- Updates the `addMedia` APIs to include nicer abilities to clone + link media, and specify a parent to create the new media under


src/hub.html
```diff
@@ -453,6 +470,24 @@
                 </a-entity>
             </template>
 
+            <template id="linked-media">
+                <a-entity
+                    class="interactable"
+                    body-helper="type: dynamic; mass: 1; collisionFilterGroup: 1; collisionFilterMask: 15;"
+                    is-remote-hover-target
+                    tags="offersRemoteConstraint: true;"
```
johnshaughnessy: Are you allowed to create a physics constraint between the remote and this object? I thought it was going to have a component attached that set its transform to followed your camera / avatar head around.

src/hub.html
```diff
@@ -453,6 +470,24 @@
...
+                    set-unowned-body-kinematic
+                    destroy-at-extreme-distances
+                    scalable-when-grabbed
```
johnshaughnessy: Same as above -- are you supposed to be able to grab and scale this object?

gfodor: no, you shouldn't be able to at this time (tho eventually I'd like to let you do that, and then when you let go, it returns to its normal position)

src/hub.html
gfodor: i'm not familiar with this tag - you should only need to use the cursor with these objects. can this be removed?

johnshaughnessy: Yes, I think you should be able to remove this tag. The `offersRemoteConstraint` tag is meant to signal to the `constraints-system` that it should try to create a physics constraint between this object and the cursor. 
The only other place this tag is used is in choosing which action sets should be active on the next frame. If you need the `right/leftCursorHoveringOnInteractable` action set to be active, then you should give this object a new tag and add it to the check in `resolveActionSets`.

src/hub.html
johnshaughnessy: For now you can remove `scalable-when-grabbed` (since this object won't be grabbed). That'll be a nice UX for the future though.

PR #2084, opened 2020-01-24
Movement preferences

brianpeiris:
Choose your preferred movement style with these additional movement preferences.

image

Also added a "Return to VR" link to the setting menu overlay:

image


johnshaughnessy:
The on-screen joysticks for touch screen will also use this function for motion.
https://github.com/mozilla/hubs/blob/6eb02bf34a55a14eb87da51159c1e9380a03769a/src/systems/character-controller-system.js#L69

 We should be checking user preferences here too.

Issue #1843, opened 2019-10-21
Users on Quest disappear before personal space bubble distance


ducile:
**Description**
When an avatar moves towards you or you moving towards them they will disappear after reaching a certain "close" distance. Happens on the Quest.

**To Reproduce**
1. Go to Hubs with another person 
2. Move towards that person or make the person move toward you.

**Expected behavior**
The avatars should be visible up to the point where they intersect with you and go through you.

**Screenshots**
No screenshots available from the Quest.

**Hardware** 
 - Device: Oculus Quest
 - OS: -
 - Browser: Oculus Browser

misslivirose:
Hey @ducile - are you referring to the personal space bubble? It's intentional that avatars you're too close to disappear, and I'd like to know if this is the regular instantiation of that or if it's happening in a different context.

ducile:
Ah that makes sense to me. I was talking to Blair and we thought it might be a bug.

blairmacintyre:
No, I don't think this is quite it:  I was dissappearing to @ducile before he did for me.  The distance was quite a bit further than the space bubble.

Perhaps the bug is that the distance is wrong on Quest.  Someone should just test.

misslivirose:
I saw that with your specific avatar @blairmacintyre - I didn't see that with other robots. Your specific avatar had different parts that disappeared out of existence when I approached you.

blairmacintyre:
I swtiched to the robot and I thought @ducile said it was also happening with the robot

misslivirose:
Ah okay, good to know

ducile:
It was happening in a similar distance, but I'm not sure if it is part of the personal space bubble.

misslivirose:
According to @j-conrad the issue with the first case (custom meshes) is usually a result of the meshes on the avatars not being one single mesh and instead having separate pieces (it's happening on my own custom avatar mesh too)

misslivirose:
The second one - we have some cases for the personal space bubble to improve the mechanics (Issues #1791 and #328)

misslivirose:
Renaming this again because I realize now that there definitely are issues with Quest users specifically triggering the personal space bubble mechanic or disappearing before they should be. It's not happening with other users in VR, just Quest. Might be related to play space.

gfodor:
I addressed a lot of issues here: https://github.com/mozilla/hubs/pull/2085

Not sure if this covers everything, so leaving this open in case we hit more issues.

Issue #1791, opened 2019-10-04
Space bubble makes avatars disappear instead of going transparent


gfodor:
**Description**
In the smoke test, if you approach and avatar, it no longer goes transparent when close to it, but fully disappears.

**Hardware** 
Chrome desktop

gfodor:
arguably this is a feature not a bug. i've often found it slightly questionable that we should show the avatar when the space bubble is violated, since the whole point is that it should prevent bad actors being visible in your face.

gfodor:
I determined the root cause of this in https://github.com/mozilla/hubs/pull/2085 and I think hiding them is appropriate given the need to protect users. Closing.

PR #2085, opened 2020-01-24
Bug/disappearing avatars

gfodor:
Fixes a wide variety of issues causing avatars to disappear or appear choppy.

Fixes https://github.com/mozilla/hubs/issues/1843

- Previously, we were using vertex skinning to position the hips (updated via `ik-controller`.) The net effect of this is that the displacement made to the hip (for example, the height delta needed to move the head up) was being done on GPU, so was not visible to three.js for accurate frustum culling. Three frustum culling uses the geometry's boundingSphere, and so our boundingSphere for avatars was very far off since the root hip transform was not baked into the matrixWorld of the root avatar geometry. To fix this, `ik-controller` now transforms the `AvatarRoot` entity instead of the hip element with the computed hip transform needed to properly orient the head. The hip joint transform is now fixed at 0,0,0 in local space. I kept the local references to `hips` in the naming, to make it clear that we are still doing the IK math to determine logical hip transform, but are applying it to the root avatar entity.

- Previously, we were not taking into account the camera viewfinder frustum for optimizations which disabled billboarding and IK updates on avatars that were 'not visible'. This led to visible choppiness on avatars, nametags, and other elements when taking a video recording that were in the camera viewfinder frustum but not in the player frustum (eg behind the player.) This PR updates things so that those optimizations will be disabled on entities which are in the view frustum of the currently owned camera when that camera's viewfinder is active. (Note that on mobile the viewfinder is only active when you are holding the camera, so this change will not affect overall frame performance except for when the camera is being held.)

- Previously, space bubble effects were seen in the camera viewfinder. So if you were near someone and taking a picture, they would be invisible in your camera. This updates the camera renderer so that the space bubble behavior is temporarily disabled as we render the viewfinder/recording frame.

- Fixed a bunch of logic in `camera-tools` system that was causing it to misinform callers of the currently owned camera.

Along the way, adds two new facilities to the invader system, the ability to `enable`/`disable` the invader handling explicitly, as well as the ability for an external caller to inform the invader that the element it is managing the visibility for should always be invisible.

Also, I finally tracked down why avatars stopped going translucent during invasion. I think the behavior we have now where they are hidden is better safety-wise so I didn't try to restore translucency, but the problem was that our lookup by name for the avatar mesh started failing for some reason (likely as we worked on the avatar system) and so the invader logic fell back on using visibility instead of material opacity as the invasion response.

### DIFF
--- a/src/components/camera-tool.js
+++ b/src/components/camera-tool.js
@@ -611,12 +611,6 @@ AFRAME.registerComponent("camera-tool", {
           playerHead.updateMatrixWorld(true, true);
         }
 
-        const bubbleSystem = this.el.sceneEl.systems["personal-space-bubble"];
-
-        for (const invader of bubbleSystem.invaders) {
-          invader.disable();
-        }
-
         let playerHudWasVisible = false;
 
         if (this.playerHud) {
@@ -626,6 +620,14 @@ AFRAME.registerComponent("camera-tool", {
             for (const mesh of Object.values(this.el.sceneEl.systems["hubs-systems"].spriteSystem.meshes)) {
               mesh.visible = false;
             }
+          }
+        }
+
+        const bubbleSystem = this.el.sceneEl.systems["personal-space-bubble"];
+
+        if (bubbleSystem) {
+          for (const invader of bubbleSystem.invaders) {
+            invader.disable();
           }
         }
 
@@ -672,8 +674,10 @@ AFRAME.registerComponent("camera-tool", {
           }
         }
 
-        for (const invader of bubbleSystem.invaders) {
-          invader.enable();
+        if (bubbleSystem) {
+          for (const invader of bubbleSystem.invaders) {
+            invader.enable();
+          }
         }
 
         this.lastUpdate = now;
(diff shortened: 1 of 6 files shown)


brianpeiris:
You have a typo in the issue description -- it mentions that this PR fixes #2085. Did you mean #1791?

src/components/camera-tool.js
```diff
@@ -623,6 +623,14 @@ AFRAME.registerComponent("camera-tool", {
           }
         }
 
+        const bubbleSystem = this.el.sceneEl.systems["personal-space-bubble"];
+
+        if (bubbleSystem) {
+          for (const invader of bubbleSystem.invaders) {
+            invader.disable();
+          }
```
brianpeiris: Just wanted to note that you're using a `for...of` loop in a `tock` here. Perhaps it's not a concern if the short-lived allocations we've observed can indeed be ignored.

gfodor: good point i'll refactor

gfodor:
Ah no bug I meant was https://github.com/mozilla/hubs/issues/1843.

I'll set up a two-VR headset test, good call. I think from a math point of view there's minimal difference here since we're just moving the local transform up in the scene graph by a level, but agree its hard to be sure without fully testing other avatars.

gfodor:
Did some testing on quest + rift simultaneously switching between minecraft, robot, and santa avatars. Seemed to work, except santa avatar revealed this bug (which existed previously) https://github.com/mozilla/hubs/issues/2097

Issue #1809, opened 2019-10-11
Teleporter gets stuck


johnshaughnessy:
Repro:
With 6DOF controller, Start teleporting. Put your hand in a pen. Release teleport button. Your teleport ray will remain on.

PR #2090, opened 2020-01-24
Fix issue with teleport getting stuck when over interactable

gfodor:
Partially fixes https://github.com/mozilla/hubs/issues/1809

Need to test on vive. Skipped windows mixed reality controllers for now, since the binding changes looked non-trivial. So we should leave the issue open and update it properly to note it is fixed on oculus + vive.

### DIFF
--- a/src/systems/userinput/bindings/oculus-touch-user.js
+++ b/src/systems/userinput/bindings/oculus-touch-user.js
@@ -793,11 +793,6 @@ export const oculusTouchUserBindings = addSetsToBindings({
     {
       src: { value: leftGripPressed2 },
       dest: { value: leftGripRisingGrab },
-      xform: xforms.rising
-    },
-    {
-      src: { value: leftTriggerPressed2 },
-      dest: { value: leftTriggerRisingGrab },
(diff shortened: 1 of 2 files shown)


src/systems/userinput/bindings/oculus-touch-user.js
```diff
@@ -793,11 +793,6 @@ export const oculusTouchUserBindings = addSetsToBindings({
     {
       src: { value: leftGripPressed2 },
       dest: { value: leftGripRisingGrab },
-      xform: xforms.rising
-    },
-    {
-      src: { value: leftTriggerPressed2 },
-      dest: { value: leftTriggerRisingGrab },
```
gfodor: these targets didn't seem consumed, any idea on if they are actually needed?

johnshaughnessy: Seems safe to remove

johnshaughnessy:
Will push this next time I take the push channel

gfodor:
Note this was never tested on vive

PR #2096, opened 2020-01-28
Fix issue where resuming suspended mobile VR headset causes spinner

gfodor:
Fixes a long standing bug where if you suspend a mobile headset while in hubs and then un-suspend, you get stuck on a spinner since we were re-exiting VR mode.

Also fixes a bug where if you let the headset remain on standby mode long enough, triggering the exit-on-blur behavior, we fail to exit VR mode properly before tearing down the scene so you also get stuck with a spinner when putting the headset on.

Also bumps the blur timeout to 30s, which seems like a better balance between being too fast and mis-triggering, and too long to preserve user privacy.

### DIFF
--- a/src/hub.js
+++ b/src/hub.js
@@ -918,7 +918,6 @@ document.addEventListener("DOMContentLoaded", async () => {
   });
 
   scene.addEventListener("leave_room_requested", () => {
-    scene.exitVR();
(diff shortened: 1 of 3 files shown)


gfodor:
I'm just going to push this, since it might help with demos in all-hands.

PR #2107, opened 2020-02-04
Add sync IRL feature behind qs param

johnshaughnessy:
This PR adds a `SyncIRLSystem` that provides a way to synchronize the virtual space with the real space. It is intended to be used for demos. 

Instructions:
- Mark two spots on the floor : one for the right controller and one for the left.
- Add the `irl` query string parameter to the URL. 
- (optional) Disable motion / teleporting / fly mode via user preferences.
- Enter the room.
- Place 6DOF controllers on the floor at the chosen locations.
- Press the `syncIRL` button and wait for the countdown to complete.


johnshaughnessy:
Missed deadline for demo. Can revive this PR if necessary

PR #2116, opened 2020-02-07
Add support for new account feature flags

gfodor:
Goes with https://github.com/mozilla/reticulum/pull/312, https://github.com/mozilla/reticulum/pull/313, and https://github.com/mozilla/reticulum/pull/314

- Adds proper handling of removal of create room buttons CTAs if permission to create rooms is denied (note there are still some UI elements that show up, but not CTAs. can work these in later)

- Properly handles case when rooms require accounts to join

- Hides the remix button on the scene landing page if spoke is disabled

### DIFF
--- a/src/hub.js
+++ b/src/hub.js
@@ -582,7 +582,7 @@ function handleHubChannelJoined(entryManager, hubChannel, messageDispatch, data)
         entryManager.exitScene();
       }
 
-      const connectionErrorTimeout = setTimeout(onConnectionError, 30000);
+      const connectionErrorTimeout = setTimeout(onConnectionError, 90000);
       scene.components["networked-scene"]
         .connect()
         .then(() => {


src/index.js
```diff
@@ -102,23 +105,29 @@ async function fetchFeaturedRooms() {
...
+    retPhxChannel.push("refresh_perms_token").receive("ok", ({ perms_token }) => {
+      const perms = jwtDecode(perms_token);
+      configs.setIsAdmin(perms.postgrest_role === "ret_admin");
+
+      if (perms.postgrest_role === "ret_admin") {
+        showAdmin = true;
+      }
+
+      showCreate = !!perms.create_hub;
+      remountUI();
```
brianpeiris: I wonder if we should wait for perms before the initial UI mount, since it may be a bit odd to have the primary CTA show up after some delay.

gfodor: Yeah I adjusted the CSS to reduce the visual effect of it appearing late, seems a tough call. A full white screen is probably bad, but maybe having the content pop in (eg background + header/footer shown) makes more sense. Also the recent changes w favorites loading may mitigate this since Dom is going to add an additional pre-load mount.

gfodor:
lol, apparently i totally screwed up this PR via a git push

PR #2123, opened 2020-02-10
Fix mixup between hands and remote constraint checks

johnshaughnessy:
Fixes an issue where the `right/leftHandHoveringOnInteractable` action sets were checking for a `remote` tag instead of a `hand` tag.
Also moves the check for the `togglesHoveredActionSet` tag to activate the `right/leftCursorHoveringOnInteractable` action sets (since I think this was gfodor's intention when adding it).


gfodor:
oops

PR #2130, opened 2020-02-11
Disable aframe's default lights

johnshaughnessy:
AFrame registers its `light` system whether you put it on your `a-scene` or not. This explicitly adds the system to the `a-scene` in order to prevent it from from adding default lights: https://github.com/MozillaReality/aframe/blob/hubs/master/src/systems/light.js#L18

### DIFF
--- a/src/hub.html
+++ b/src/hub.html
@@ -53,6 +53,7 @@
         rotate-selected-object
         environment-map
         set-max-resolution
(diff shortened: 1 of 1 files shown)


robertlong:
This will only affect an extremely tiny subset of old scenes without any lights so I think this is good to merge.

PR #2132, opened 2020-02-12
Add support for the scene-audio-settings component

robertlong:


src/gltf-component-mappings.js
```diff
@@ -378,3 +378,11 @@ AFRAME.GLTFModelPlus.registerComponent("trimesh", "trimesh", el => {
 AFRAME.GLTFModelPlus.registerComponent("particle-emitter", "particle-emitter");
 
 AFRAME.GLTFModelPlus.registerComponent("networked-drawing-buffer", "networked-drawing-buffer");
+
+AFRAME.GLTFModelPlus.registerComponent(
+  "scene-audio-settings",
```
gfodor: perhaps just `audio-settings`? could imagine this being used at lower level nodes at some point (eg the regions we talked about)

src/components/media-views.js
```diff
@@ -517,15 +517,24 @@ AFRAME.registerComponent("media-video", {
             linkedMediaElementAudioSource ||
             this.el.sceneEl.audioListener.context.createMediaElementSource(audioSourceEl);
 
+          if (this.el.components["listed-media"]) {
```
gfodor: this is probably the wrong check here. listed-media semantically are "things we want showing up in the object list", but we have media not in the list whose audio settings need to be set. for example, static scene owned media do not show up there. i'm actually surprised your PR worked, since you are testing scene owned media, so I'm not sure if I'm missing something.

gfodor: My suggestion would be to:
- Create a new component/system pair for registering overridable audio sources and their type (similar in structure to `listed-media`), maybe `audio-source`
- Replace the `avatar-audio-source` component on avatars with it, and add the new component to all the media templates

src/gltf-component-mappings.js
robertlong: I wouldn't use the same component for regions. They would have different functionality so I'd use another name, but yeah we can remove scene from the name. @netpro2k what do you think?

src/components/media-views.js
robertlong: I believe static scene media will be explicitly set in the Spoke scene. At least that's how I have it designed. Having these settings apply to nodes placed in Spoke introduced problems with nodes with stereo audio sources and sources with different distance models. For example falloff is in the range of 0-1 for linear, but 0-Infinity for inverse/exponential. I think these are good reasons to avoid having `scene-audio-setting` affect media added in Spoke.

So based on that design the `listed-media` component does correctly identify all media added to the Hubs scene, but I agree, that `audio-source` component and specifying a type so that we can use the correct implementation would also probably work. It does have the downside of setting audio properties twice and having to wait for the `video-loaded` event. Based on @netpro2k's concerns with setting properties twice I decided on this design.

robertlong: I'm going to go ahead with renaming `scene-audio-settings` to `audio-settings` and implementing this `audio-source` component.

That doesn't address the points @gfodor made about static scene owned media. I don't think `audio-settings` should affect those, but that's purely for UX reasons in Spoke. If we can figure that out, I'm open to it.

PR #2134, opened 2020-02-12
Feature/identity admin console

gfodor:
Goes with https://github.com/mozilla/reticulum/pull/317, and adds the admin tools for dealing with the new identity concept. Also, updates the client to use the new identity name from presence properly.

This incidentally includes a bunch of stuff from an already-reviewed PR due to a git snafu.


admin/src/react-components/accounts.js
```diff
@@ -10,32 +12,62 @@ import {
...
-    </SimpleForm>
-  </Edit>
-);
+  return (
+    <List {...props} filters={<AccountFilter />}>
+      <Datagrid>
+        <TextField source="id" />
+        <DateField source="inserted_at" />
+        <DateField source="updated_at" />
+        <ReferenceManyField label="Identity" target="_account_id" reference="identities">
```
brianpeiris: Why not just a `ReferenceField`?

gfodor: So this is actually a little confusing in the naming, ReferenceField lets you fetch a record via a foreign key on the current record, whereas ReferenceManyField lets you fetch N records in the other direction (where a foreign key in another table references this record.) (There's no equivalent of "has_one", so you just list one record in this case)

PR #2156, opened 2020-02-21
Add per-room caps to room settings and entry

netpro2k:
This moves the hardcoded 24 person room count to be configurable per room by owners. It is still only enforced clientside. Also fixes some cases where the room cap was able to get out of sync (likely unnoticed because we don't often hit the room cap ourselves). Still uses a room cap of 24 if the room does not specify a cap (it is set to 0). Also limits the maximum room cap to 64. We likely want both the maximum and default room caps to be configurable by admin console, just need to discuss how we want this exposed.

- [x] Decide if/how we want to expose default/maximum room cap to admin console (maximum should probably line up with actual Janus room caps
- [x] Merge reticulum PR


src/utils/hub-channel.js
```diff
@@ -54,6 +58,21 @@ export default class HubChannel extends EventTarget {
...
+    if (!hub) return false;
+    if (this.canOrWillIfCreator("update_hub")) return true;
+
+    const roomEntrySlotCount = Object.values(this.presence.state).reduce((acc, { metas }) => {
+      const meta = metas[metas.length - 1];
+      const usingSlot = meta.presence === "room" || (meta.context && meta.context.entering);
+      return acc + (usingSlot ? 1 : 0);
+    }, 0);
+
+    console.log(roomEntrySlotCount, hub.member_cap);
```
gfodor: log

src/utils/hub-channel.js
```diff
@@ -54,6 +58,21 @@ export default class HubChannel extends EventTarget {
...
+
+    return roomEntrySlotCount < (hub.member_cap || DEFAULT_MAX_OCCUPIED_ROOM_ENTRY_SLOTS);
```
gfodor: if `member_cap` is zero its falsey so won't handle this properly

src/utils/hub-channel.js
```diff
@@ -54,6 +58,21 @@ export default class HubChannel extends EventTarget {
     return this.can(permission);
   }
 
+  canEnterRoom(hub) {
+    if (!hub) return false;
+    if (this.canOrWillIfCreator("update_hub")) return true;
+
+    const roomEntrySlotCount = Object.values(this.presence.state).reduce((acc, { metas }) => {
```
gfodor: see the reticulum PR, i'd refactor 'room entry slots' to be 'room member counts'

src/react-components/ui-root.js
```diff
@@ -128,18 +128,12 @@ class UIRoot extends Component {
...
     hubChannel: PropTypes.object,
     linkChannel: PropTypes.object,
-    hubEntryCode: PropTypes.number,
+    hub: PropTypes.object,
     availableVREntryTypes: PropTypes.object,
     environmentSceneLoaded: PropTypes.bool,
     entryDisallowed: PropTypes.bool,
     roomUnavailableReason: PropTypes.string,
     platformUnsupportedReason: PropTypes.string,
-    hubId: PropTypes.string,
```
gfodor: +1

src/schema.toml
```diff
@@ -27,10 +24,15 @@ features.show_avatar_editor_link = { category = "features", type = "boolean", na
...
-
 features.public_rooms = { category = "features", type = "boolean", name = "Public Rooms", description = "Allow visitors to set rooms to public to feature them on the homepage." }
 features.default_room_id = { category = "features", type = "string", name = "Homepage Room ID", description = "Use the specified room in place of the homepage."}
 
+features.permissive_rooms = { category = "rooms", type = "boolean", name = "Permissive Rooms", description = "Enable full content creation permissions on new rooms." }
+features.disable_room_creation = { category = "rooms", type = "boolean", name = "Disable room creation", description = "Disable creating rooms for non-administrators." }
+features.require_account_for_join = { category = "rooms", type = "boolean", name = "Require accounts for room access", description = "Require accounts for accessing rooms." }
+features.default_room_size = { category = "rooms", type = "number", name = "Default room size", description = "Default room size for new rooms. This does not include users in the lobby." }
+features.max_room_size = { category = "rooms", type = "number", name = "Maximum room size", description = "Maximum room size visitors can set." }
```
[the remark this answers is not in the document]
netpro2k: Yep. And we use the term "visitors" in the admin console to mean visitors to the site.

Issue #1187, opened 2019-04-04
Object menu buttons can appear upside down


brianpeiris:
The object menu rotates with the object, so if the object is upside down, so is the menu.

![2019-04-04_14-55-24](https://user-images.githubusercontent.com/79419/55591373-37cc9080-56ea-11e9-8fe2-6663bfcf79e5.png)

misslivirose:
See also: #1460

PR #2159, opened 2020-02-25
Keep menus upright

johnshaughnessy:
Fixes https://github.com/mozilla/hubs/issues/1187


johnshaughnessy:
Closing for now because I'm still quite unhappy with this solution. Will revisit this when I can. :(

PR #2160, opened 2020-02-26
Fix spawner position

johnshaughnessy:
Normally, when we load a model with `media-loader`, we reposition the visible mesh so that the center of its AABB is at its parent's position. This way, setting the parent's position to location X feels like putting the center of the mesh at location X. However, when scene authors create spawners, they position parent nodes such that their child mesh nodes (which are not centered) end up in the right position. If we reposition the mesh in this case, it will end up in the wrong place.

This PR fixes the bug by adding a flag on `media-loader` called `moveTheParentNotTheMesh` that will make sure the position of parent node is at the center of the mesh's AABB, but will keep the mesh fixed in place while it does so.

### DIFF
--- a/src/systems/super-spawner-system.js
+++ b/src/systems/super-spawner-system.js
@@ -1,9 +1,10 @@
 import { addMedia } from "../utils/media-utils";
+import { computeObjectAABB } from "../utils/auto-box-collider";
 import { ObjectContentOrigins } from "../object-types";
 
 // WARNING: This system mutates interaction system state!
 export class SuperSpawnerSystem {
-  maybeSpawn(state, entity, grabPath) {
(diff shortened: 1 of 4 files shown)


src/systems/super-spawner-system.js
```diff
@@ -3,7 +3,7 @@ import { ObjectContentOrigins } from "../object-types";
 
 // WARNING: This system mutates interaction system state!
 export class SuperSpawnerSystem {
-  maybeSpawn(state, entity, grabPath) {
+  maybeSpawn(state, grabPath) {
```
johnshaughnessy: Nothing changed in this file except that `entity` was being passed around these functions for no reason, so I stopped doing that.

PR #2167, opened 2020-02-27
Update to use new workerized physics

InfiniteLee:
This updates Hubs to use upgraded `three-ammo`, `three-to-ammo` and `ammo-debug-drawer` which have all been updated to be able to run in a worker thread.
`three-to-ammo` supports this via both using **SharedArrayBuffers** and **Transferable** **ArrayBuffer** via the **PostMessage** API. This update uses the **Transferable** with **ArrayBuffer** approach, as SharedArrayBuffers aren't fully available in Firefox or Android Chrome.

The overview of how this works is that there is an ArrayBuffer that is created by the main thread in Hubs. This ArrayBuffer contains 26 * 4 bytes (a 64bit float or int) of data per physics object that is added to the world. 
The first 16 * 4 bytes of data represents the 16 float values of world transform (Matrix4) for an object. 
The next 2 * 4 bytes of data are for Linear and Angular velocities (floats) of the object. 
The last 8 * 4 bytes are integers (index values) that are used for tracking what objects this object is colliding with.
This schema can be expanded in the future if additional data is needed.

The ArrayBuffer is passed back and forth as a transferable from main thread to worker thread via PostMessage. On the worker thread's (which is running the physics simulation) next tick after it receives the ArrayBuffer, it updates the transforms for all static/kinematic physics objects based on what is currently written to the matrices for those objects in the buffer. It then steps the physics simulation, then writes back out to the ArrayBuffer the transforms for all the dynamic objects in the world. Also, linear and angular velocity are calculated and any collisions with other objects are updated accordingly in the ArrayBuffer. Finally, the worker thread PostMessages the ArrayBuffer back to the main thread.
On the main thread's next tick after receiving the ArrayBuffer, it copies out the transforms for all dynamic objects and applies those transforms to the associated Object3D. Then, the transforms for any static/kinematic objects are copies from their Object3D's into the ArrayBuffer. A local cache of the velocities and collisions are updated as well for use of the main thread at any time (Note that when using transferables, the contents of an ArrayBuffer are _only_ available _until_ it has been PostMessage'd to another thread. That thread cannot access that ArrayBuffer until it has been PostMessage'd back to it). Finally the main thread PostMessages the ArrayBuffer back to the worker, and the cycle repeats. 

All other events, such as initialization of the world, creation of physics bodies, shapes, and constraints, are all done via PostMessage. 

Please see the following pull requests for each dependency:
https://github.com/InfiniteLee/three-ammo/pull/2
https://github.com/InfiniteLee/three-to-ammo/pull/6
https://github.com/InfiniteLee/ammo-debug-drawer/pull/3

Note: `ammo-debug-drawer` currently only works with browsers that support SharedArrayBuffers. This is enabled by default for Chrome for desktop and can be enabled via a flag for Chrome for Android (e.g. on Quest).

### DIFF
--- a/src/components/body-helper.js
+++ b/src/components/body-helper.js
@@ -1,5 +1,6 @@
-import { Body } from "three-ammo";
-import { ACTIVATION_STATE, TYPE } from "three-ammo/constants";
+import { CONSTANTS } from "three-ammo";
+const ACTIVATION_STATE = CONSTANTS.ACTIVATION_STATE,
+  TYPE = CONSTANTS.TYPE;
 
 const ACTIVATION_STATES = [
   ACTIVATION_STATE.ACTIVE_TAG,
@@ -34,25 +35,24 @@ AFRAME.registerComponent("body-helper", {
   init: function() {
     this.system = this.el.sceneEl.systems["hubs-systems"].physicsSystem;
     this.alive = true;
+    this.uuid = -1;
     this.system.registerBodyHelper(this);
   },
 
   init2: function() {
-    this.body = new Body(this.data, this.el.object3D, this.system.world);
-    this.system.addBody(this.body);
+    this.el.object3D.updateMatrices();
+    this.uuid = this.system.addBody(this.el.object3D, this.data);
   },
 
   update: function(prevData) {
-    if (prevData !== null && this.body) {
-      this.body.update(this.data);
+    if (prevData !== null && this.uuid !== -1) {
+      this.system.updateBody(this.uuid, this.data);
     }
   },
 
   remove: function() {
-    if (this.body) {
-      this.system.removeBody(this.body);
-      this.body.destroy();
-      this.body = null;
+    if (this.uuid !== -1) {
+      this.system.removeBody(this.uuid);
     }
     this.alive = false;
   }
--- a/src/components/shape-helper.js
+++ b/src/components/shape-helper.js
@@ -61,7 +61,7 @@ AFRAME.registerComponent("shape-helper", {
         return;
       }
       this.mesh = this.el.object3DMap.mesh;
-      this.mesh.updateMatrices(true, true);
+      this.mesh.updateMatrices();
     }
 
     this.uuid = this.system.addShapes(this.bodyHelper.uuid, this.mesh, this.data);
--- a/src/components/shape-helper.js
+++ b/src/components/shape-helper.js
@@ -1,5 +1,3 @@
-/* global Ammo */
-import * as threeToAmmo from "three-to-ammo";
 import { SHAPE, FIT } from "three-ammo/constants";
 
 AFRAME.registerComponent("shape-helper", {
@@ -38,6 +36,7 @@ AFRAME.registerComponent("shape-helper", {
   init: function() {
     this.system = this.el.sceneEl.systems["hubs-systems"].physicsSystem;
     this.alive = true;
+    this.uuid = -1;
     this.system.registerShapeHelper(this);
   },
 
@@ -52,36 +51,26 @@ AFRAME.registerComponent("shape-helper", {
         this.bodyHelper = bodyEl.components["body-helper"];
       }
     }
-    if (!this.bodyHelper || !this.bodyHelper.body) {
+    if (!this.bodyHelper || this.bodyHelper.uuid === null || this.bodyHelper.uuid === undefined) {
       console.warn("body not found");
       return;
     }
-    if (this.data.fit !== FIT.MANUAL) {
+    if (this.data.fit === FIT.ALL) {
       if (!this.el.object3DMap.mesh) {
         console.error("Cannot use FIT.ALL without object3DMap.mesh");
         return;
       }
       this.mesh = this.el.object3DMap.mesh;
-      this.mesh.updateMatrices();
+      this.mesh.updateMatrices(true, true);
     }
 
-    this.shapes = threeToAmmo.createCollisionShapes(this.mesh, this.data);
-    for (let i = 0; i < this.shapes.length; i++) {
-      this.bodyHelper.body.addShape(this.shapes[i]);
-    }
+    this.uuid = this.system.addShapes(this.bodyHelper.uuid, this.mesh, this.data);
   },
 
   remove: function() {
-    if (this.shapes) {
-      for (let i = 0; i < this.shapes.length; i++) {
-        if (this.bodyHelper.body) {
-          this.bodyHelper.body.removeShape(this.shapes[i]);
-        }
-        this.shapes[i].destroy();
-        Ammo.destroy(this.shapes[i].localTransform);
-      }
+    if (this.uuid !== -1) {
+      this.system.removeShapes(this.bodyHelper.uuid, this.uuid);
     }
-    this.shapes = null;
     this.alive = false;
   }
 });
--- a/src/systems/physics-system.js
+++ b/src/systems/physics-system.js
@@ -19,22 +19,19 @@ export class PhysicsSystem {
 
     this.bodyHelpers = [];
     this.shapeHelpers = [];
-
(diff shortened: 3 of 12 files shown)


src/systems/physics-system.js
```diff
@@ -1,103 +1,311 @@
...
+              transform.multiplyMatrices(inverse, matrix);
+              transform.decompose(object3D.position, object3D.quaternion, scale);
+            }
+
+            object3D.updateMatrices();
+            this.objectMatricesFloatArray.set(object3D.matrixWorld.elements, index * BUFFER_CONFIG.BODY_DATA_SIZE);
+
+            if (this.bodyLinearVelocities.hasOwnProperty(uuid)) {
+              this.bodyLinearVelocities[uuid] = this.objectMatricesFloatArray[
+                index * BUFFER_CONFIG.BODY_DATA_SIZE + 16
```
johnshaughnessy: Consider having named constants for these offsets (+16, +17, +18)

src/systems/physics-system.js
```diff
@@ -1,103 +1,311 @@
...
-  removeBody(body) {
-    const idx = this.bodies.indexOf(body);
+  removeBody(uuid) {
+    delete this.indexToUuid[this.uuidToIndex[uuid]];
+    delete this.uuidToIndex[uuid];
+    delete this.object3Ds[uuid];
+    delete this.bodyOptions[uuid];
+    delete this.collisions[uuid];
+    delete this.bodyLinearVelocities[uuid];
+    delete this.bodyAngularVelocities[uuid];
```
johnshaughnessy: Many similar data structures here to keep in-sync: Should this data live in the same map?

src/systems/physics-system.js
```diff
@@ -1,103 +1,311 @@
...
+    const idx = this.bodyUuids.indexOf(uuid);
     if (idx !== -1) {
-      this.bodies.splice(idx, 1);
+      this.bodyUuids.splice(idx, 1);
```
johnshaughnessy: Is it a bug if the idx is not found? If so, a log here might help us track it down

src/systems/physics-system.js
```diff
@@ -1,103 +1,311 @@
...
-    if (this.world) {
-      if (this.debugRequested !== this.debugEnabled) {
-        this.debugEnabled = this.debugRequested;
-        if (this.debugEnabled) {
-          this.world.getDebugDrawer(this.scene).enable();
-        } else {
-          this.world.getDebugDrawer(this.scene).disable();
+  enableDebug() {
+    if (!window.SharedArrayBuffer) {
+      console.warn("Physics debug rendering only available in browsers that support SharedArrayBuffers.");
```
johnshaughnessy: What would it take for us to support debug rendering via transferrables api?

src/systems/physics-system.js
```diff
@@ -1,103 +1,311 @@
...
 
-    this.bodies = [];
     this.bodyHelpers = [];
     this.shapeHelpers = [];
 
+    this.bodyUuids = [];
+    this.uuidToIndex = {};
+    this.indexToUuid = {};
+    this.object3Ds = {};
+    window.object3Ds = this.object3Ds;
```
johnshaughnessy: writing to `window`

src/components/shape-helper.js
```diff
@@ -52,36 +51,26 @@ AFRAME.registerComponent("shape-helper", {
...
     }
-    if (this.data.fit !== FIT.MANUAL) {
+    if (this.data.fit === FIT.ALL) {
       if (!this.el.object3DMap.mesh) {
         console.error("Cannot use FIT.ALL without object3DMap.mesh");
         return;
       }
       this.mesh = this.el.object3DMap.mesh;
-      this.mesh.updateMatrices();
+      this.mesh.updateMatrices(true, true);
```
johnshaughnessy: Why are the `force` arguments required here?

src/components/body-helper.js
```diff
@@ -34,25 +35,24 @@ AFRAME.registerComponent("body-helper", {
...
     this.alive = true;
+    this.uuid = -1;
     this.system.registerBodyHelper(this);
   },
 
   init2: function() {
-    this.body = new Body(this.data, this.el.object3D, this.system.world);
-    this.system.addBody(this.body);
+    this.el.object3D.updateMatrices();
+    this.uuid = this.system.addBody(this.el.object3D, this.data);
```
johnshaughnessy: This `uuid` is used as a key in several objects holding info associated with this body -- any reason not to use this (already unique) `body-helper` as the key in `Map` objects (which, unlike objects, allow keys of any type)? It seems like everywhere that uses the `uuid` gets it by first getting a reference to the `body-helper`

src/systems/physics-system.js
```diff
@@ -1,103 +1,311 @@
...
+        this.shapes[bodyUuid].push(shapesUuid);
+      } else if (event.data.type === MESSAGE_TYPES.TRANSFER_DATA) {
+        this.objectMatricesFloatArray = event.data.objectMatricesFloatArray;
+        this.objectMatricesIntArray = new Int32Array(this.objectMatricesFloatArray.buffer);
+        this.stepDuration = event.data.stepDuration;
       }
-      this.shapeHelpers.length = 0;
-      this.bodyHelpers.length = 0;
-    });
+    };
```
johnshaughnessy: Regarding the `onmessage` handler on the other side -- Do we expect messages to be coming in out-of-order, and if not should we be logging this as a bug? Surprised to see a strategy where we stick a message back into the queue if there wasn't a body it's associated with.
https://github.com/InfiniteLee/three-ammo/pull/2/files#diff-3db89a7579b555b025615250a93bb695R326

src/systems/physics-system.js
```diff
@@ -1,103 +1,311 @@
...
+    }
+    this.workerHelpers.removeBody(uuid);
+  }
+
+  addShapes(bodyUuid, mesh, options) {
+    if (mesh) {
+      const scale = new THREE.Vector3();
+      scale.setFromMatrixScale(mesh.matrixWorld);
+    }
+    this.workerHelpers.addShapes(bodyUuid, this.nextShapeUuid, mesh, options);
```
johnshaughnessy: Any concern over cloning all their vertices before posting them to the worker? If we could avoid this, that seems good. https://github.com/InfiniteLee/three-ammo/pull/2/files#diff-168726dbe96b3ce427e7fedce31bb0bcR43

src/systems/physics-system.js
```diff
@@ -1,103 +1,321 @@
...
+    const arrayBuffer = new ArrayBuffer(4 * BUFFER_CONFIG.BODY_DATA_SIZE * BUFFER_CONFIG.MAX_BODIES);
+    this.objectMatricesFloatArray = new Float32Array(arrayBuffer);
+    this.objectMatricesIntArray = new Int32Array(arrayBuffer);
+
+    this.ammoWorker.postMessage(
+      {
+        type: MESSAGE_TYPES.INIT,
+        worldConfig: WORLD_CONFIG,
+        arrayBuffer,
+        wasmUrl: new URL(ammoWasmUrl, configs.BASE_ASSETS_PATH || window.location).href
```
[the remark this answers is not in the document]
InfiniteLee: Thanks for letting us know about this. Unfortunately, `configs.BASE_ASSETS_PATH` is required for Hubs Cloud instances. I don't believe we were aware of any issues with serving custom clients via GitHub page, but I don't think we've tried to do it before either. If you could file an issue, that will let us escalate appropriately and determine what the best resolution is to support your use case.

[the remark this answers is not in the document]
speigg: Okay, perhaps I misinterpreted how configs.BASE_ASSETS_PATH should be used. If it ever resolved to “/“, then the line I highlighted will fail, as “/“ is not a valid base URL (the second parameter to the URL constructor).

PR #2170, opened 2020-02-29
Add JSON snapshotting of input debugging

gfodor:
Eliminates issues where the debug output of the user input system does not reflect the actual state at the time the L key was pressed. Note that the `devices` field has circular references so cannot be serialized.


johnshaughnessy:
Will deploy this next time I take the push channel

PR #2172, opened 2020-02-29
Add logs to diagnose scene loading stalls

robertlong:


src/react-components/ui-root.js
```diff
@@ -1393,6 +1394,18 @@ class UIRoot extends Component {
       (!this.state.hideLoader || !this.state.didConnectToNetworkedScene) &&
       !(this.props.showSafariDialog || this.props.showWebAssemblyDialog);
 
+    if (isLoading) {
+      console.log("Loading screen shown", {
```
johnshaughnessy: I suggest wrapping this in `JSON.parse( JSON.stringify ( ) )` so that you capture the values as they are at this time and don't get fooled in the debugger which will otherwise show what the values are at the time you click to expand the object.

johnshaughnessy: See https://github.com/mozilla/hubs/pull/2170

robertlong:
I don't really want this anymore. I think we've got a handle on the issue that caused me to write this.

Issue #350, opened 2018-05-02
consider showing list of players in footer nav (2D)


cvan:
the user count is useful to have, but sometimes I can't tell if (or which of) the other participants/my friends are in the room. if there are no walls or objects obstructing us, I can eventually find them. it'd be nice to be able to expose a list of participants' user names in a hub space.

<img width="1270" alt="screen shot 2018-05-01 at 9 09 32 pm" src="https://user-images.githubusercontent.com/203725/39504995-743febf4-4d84-11e8-9ee7-71cda4781ed5.png">
<img width="868" alt="screen shot 2018-05-01 at 9 10 58 pm" src="https://user-images.githubusercontent.com/203725/39505001-7ff6a7d0-4d84-11e8-87d1-b494833a0d8f.png">
<img width="868" alt="screen shot 2018-05-01 at 9 11 13 pm" src="https://user-images.githubusercontent.com/203725/39505008-8d548302-4d84-11e8-8858-9fc9933c0c65.png">

cvan:
also came to mind: in standalone VR headsets (e.g., Oculus Go), you won't have a non-VR list to refer to. so having a 2D list accessible _someway_ would be ideal. admittedly, this does open up a can of worms: add scrolling, pagination, and so forth?

should I file a separate issue for `being able to block a user without having to freeze the experience and find them`?

robertlong:
The 2D UI has now been implemented. We still need to implement the UI in VR.

PR #2174, opened 2020-03-02
Fly around in the lobby (if the server mods allow it)

johnshaughnessy:
Adds the ability to fly around in the lobby.
~TODO: Add "enter in vr" button to watch from immersive vr mode.~
~TODO: Only allow this option if the server moderator enables it.~
TODO: All in-world menus are currently turned off. We should enable volume controls, "open link" and inspect/focus.
TODO: Make sure I didn't break scene.html (How do I test this locally?)

### DIFF
--- a/src/components/scene-preview-camera.js
+++ b/src/components/scene-preview-camera.js
@@ -56,10 +56,7 @@ AFRAME.registerComponent("scene-preview-camera", {
     const hubsSystems = this.el.sceneEl.systems["hubs-systems"];
 
     uiRoot = uiRoot || document.getElementById("ui-root");
-    const isGhost =
-      uiRoot &&
-      uiRoot.firstChild &&
-      (uiRoot.firstChild.classList.contains("watching") || uiRoot.firstChild.classList.contains("hide"));
+    const isGhost = uiRoot && uiRoot.firstChild && uiRoot.firstChild.classList.contains("isGhost");
     if (isGhost) return;
 
     let streamerCamera;
--- a/src/components/scene-preview-camera.js
+++ b/src/components/scene-preview-camera.js
@@ -16,6 +16,7 @@ function lerp(start, end, t) {
 
 const newRot = new THREE.Quaternion();
 
+let uiRoot;
 /**
  * Nicely pans the camera for previewing a scene. There's some weirdness with this right now
  * since it ends up panning in a direction dependent upon the start camera orientation,
@@ -53,6 +54,10 @@ AFRAME.registerComponent("scene-preview-camera", {
 
   tick2: function() {
     const hubsSystems = this.el.sceneEl.systems["hubs-systems"];
+
+    uiRoot = uiRoot || document.getElementById("ui-root");
+    const watching = uiRoot && uiRoot.firstChild && uiRoot.firstChild.classList.contains("watching");
+    if (watching) return;
 
     let streamerCamera;
     if (hubsSystems) {
--- a/src/components/scene-preview-camera.js
+++ b/src/components/scene-preview-camera.js
@@ -55,10 +55,6 @@ AFRAME.registerComponent("scene-preview-camera", {
   tick2: function() {
     const hubsSystems = this.el.sceneEl.systems["hubs-systems"];
 
-    uiRoot = uiRoot || document.getElementById("ui-root");
-    const isGhost = uiRoot && uiRoot.firstChild && uiRoot.firstChild.classList.contains("isGhost");
-    if (isGhost) return;
-
     let streamerCamera;
     if (hubsSystems) {
       hubsSystems.cameraSystem.mode = CAMERA_MODE_SCENE_PREVIEW;
--- a/src/components/scene-preview-camera.js
+++ b/src/components/scene-preview-camera.js
@@ -16,7 +16,6 @@ function lerp(start, end, t) {
 
 const newRot = new THREE.Quaternion();
 
-let uiRoot;
 /**
  * Nicely pans the camera for previewing a scene. There's some weirdness with this right now
  * since it ends up panning in a direction dependent upon the start camera orientation,
--- a/src/components/scene-preview-camera.js
+++ b/src/components/scene-preview-camera.js
@@ -56,8 +56,11 @@ AFRAME.registerComponent("scene-preview-camera", {
     const hubsSystems = this.el.sceneEl.systems["hubs-systems"];
 
     uiRoot = uiRoot || document.getElementById("ui-root");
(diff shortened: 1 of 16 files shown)


src/systems/scene-preview-camera-system.js
```diff
@@ -14,10 +17,18 @@ export class ScenePreviewCameraSystem {
   }
 
   tick() {
+    viewingCamera = viewingCamera || document.getElementById("viewing-camera");
+    uiRoot = uiRoot || document.getElementById("ui-root");
+    const isGhost = uiRoot && uiRoot.firstChild && uiRoot.firstChild.classList.contains("isGhost");
+    const entered = viewingCamera && viewingCamera.sceneEl.is("entered");
     for (const el of this.entities) {
```
robertlong: This should probably be a regular for loop. `for of` allocates.

src/components/scene-preview-camera.js
```diff
@@ -54,6 +55,10 @@ AFRAME.registerComponent("scene-preview-camera", {
   tick2: function() {
     const hubsSystems = this.el.sceneEl.systems["hubs-systems"];
 
+    uiRoot = uiRoot || document.getElementById("ui-root");
+    const isGhost = uiRoot && uiRoot.firstChild && uiRoot.firstChild.classList.contains("isGhost");
```
robertlong: There's a lot of checking for this "isGhost" class. I'm a little worried about the perf of the number of calls to `contains` every frame.

gfodor:
fyi, there's already a facility for the "inspect, volume controls, etc"
menus icymi - we do that if the user does not have the relevant
permissions, so we might want to treat this lobby mode as a non-privileged
session basically

src/systems/scene-preview-camera-system.js
johnshaughnessy: fixed

src/components/scene-preview-camera.js
johnshaughnessy: will only check for `isGhost` if the scene is not entered yet

misslivirose:
This is just crossing my radar with the addition of the What's New tag since I was out when this was discussed - if we don't already, we need to put a warning on rooms where this is enabled. I've already had people express discomfort with this idea and IMO even on properties that we're not maintaining, we should have UI that indicates this mode.

misslivirose:
I'll think about a design where that can be made clear, but I think this a pattern where we should err on the side of being overly cautious about the ways that someone could be burned by not knowing there's an invisible avatar listening over their shoulder.

misslivirose:
Some questions: 

1. If I enter a room where this feature is disabled, and an administrator changes it while I'm in the room, what happens? Is there any way that I am informed that the "ghost" feature is now active?

2. How does this interact with future decisions that we make about fall-off and audio attenuation? Many people use the visual metaphor within a room to understand who in the group is present. This feature changes the dynamic of those spaces significantly to a greater extent, because people now drop the visual metaphor of knowing the range of their audio reach. 

Some thoughts: 

From what I recall (and since #350 is still open), we still don't have a mechanic for VR users to even understand who is in the lobby. That feels like something that should be surfaced in VR _especially_ given that there is now a feature where some deployments of Hubs are now going to allow lobby viewers to move freely and break the spatial metaphor. 

I suspect that most people aren't going to realize that the fact that the additional button on this page means that others will be able to freely move about with no indicators to them in the room. Given the sensitivity of again, breaking that metaphor that people who are near you will be made known, this should be more explicitly communicated when this feature is enabled. 

This feels especially important given that HMC _doesn't_ have this enabled by default, and people may be bringing the mental model of proximity / general lobby viewing from one instance to another.

Personally, I'd feel very uncomfortable using Hubs instances that had this feature enabled, given that I wouldn't have any visual way of knowing if someone was invisibly in the lobby up close listening to me and recording me. Without this setting, doing that is a lot harder but that's still a privacy vector to consider, IMO.

misslivirose:
Not a complete design, but at a minimum I'd like to see a subtitle on the 'Enter Room' button in this design that says something explicit like 'This room has enabled lobby ghosts' or maybe even more directly, 'Other participants may be invisible and inaudible'.

johnshaughnessy:
Liv brings up some good points. We might want to pull the lobby ghosts feature from hubs clouds instances until we have a better story around surfacing it to users so that they know what to expect. I think it's OK for other instances to have their own rules / expectations but we can do better by most users with good defaults.

Adding a few related concerns here for when we revisit this topic: 
- Traversing the room from the lobby is possible via the object list
- The lobby camera view is not visible from within the room
- The avatar volume controls allows easy listening at a distance.
- A modded client can access any of the audio streams and play them at any volume
- A modded client can move the camera to anywhere in the room

I support us making it easy to be a good actor and harder to be a bad actor, as long as we don't accidentally give people a false sense of security in the process. 

Related: "Disappearing messages", "preventing screenshots", and DRM are implemented differently in different apps. It's important that people recognize that while these make it slightly more inconvenient for a bad actor to break the social contract, there are no guarantees here.

gfodor:
In terms of the users understanding this capability, isn’t the extra button
on the landing page the “notification” of the feature, in addition to the
actual mechanism to use it?

I agree that awareness of presence is a big problem in VR. We have the
count, but the lack of a UI toolkit is basically the root cause of this and
other poor UX in immersive mode around information not being displayed.

gfodor:
If the current button is not serving as a sufficient notice (I don’t have
the copy handy) perhaps we can try to adjust that accordingly so it solves
both problems?

gfodor:
(Both problems being “users have informed consent” and “there is a way to
enter lobby fly mode” - the user list problem I have no good solution for)

misslivirose:
> In terms of the users understanding this capability, isn’t the extra button
on the landing page the “notification” of the feature, in addition to the
actual mechanism to use it?

I think that the issue with this is that the action button here is designed for the people taking the action of moving in the lobby; it isn't the flow that the user looking for the "Enter Room" is going to take, so the likelihood of them noticing that there's an extra button and reading the fine print on there feels high to me.

mohoman:
From our perspective, we do not really need the concept of a "Lobby" but more of "Spectators"
What is most important for us is to get more people in the room simply to listen to a speaker or watch a video. So in that case, the "Spectator does not need to be able to speak or have an avatar. But we would want other avatars to be able to see them, even if it's just a 2D ghost icon or something.
Things the "Spectator" could do would be: move around the room, chat with people, listen to conversations, watch videos, and sit in chairs.
If the lobby feature could be optimized to maximize the number of "Spectators" in a room, that is what we would like to have.

PR #2179, opened 2020-03-05
Inline basis worker to get around CSP/CDN issues

netpro2k:
Bit of a shame to have to inline this, but much of it was being inlined anyway with the weird bay ThreeJS was generating the worker anyway, and this is the happy path to how all of our other workers are setup.

### DIFF
--- a/src/loaders/HubsBasisTextureLoader.js
+++ b/src/loaders/HubsBasisTextureLoader.js
@@ -1,7 +1,7 @@
 /* eslint-disable */
 
 import transcoderWasmFile from "./basis_transcoder.wasm";
-import transcoderJSFile from "file-loader!./basis_transcoder.js";
+import BasisWorker from "./basis_transcoder.worker.js";
 
 /**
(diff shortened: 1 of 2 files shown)


src/loaders/HubsBasisTextureLoader.js
```diff
@@ -201,17 +193,7 @@ THREE.BasisTextureLoader.prototype = Object.assign(Object.create(THREE.Loader.pr
...
       });
 
-      this.transcoderPending = Promise.all([jsContent, binaryContent]).then(([jsContent, binaryContent]) => {
-        var fn = THREE.BasisTextureLoader.BasisWorker.toString();
-
-        var body = [
-          "/* basis_transcoder.js */",
-          jsContent,
-          "/* worker */",
-          fn.substring(fn.indexOf("{") + 1, fn.lastIndexOf("}"))
```
netpro2k: wow

robertlong: Lol, it's not that bad.

PR #2183, opened 2020-03-05
Check to make sure batching is enabled

johnshaughnessy:
Prevents a crash when batching is disabled


johnshaughnessy:
Fixed in https://github.com/mozilla/hubs/pull/2186

PR #2184, opened 2020-03-06
Fix error that occurs if a body is added and removed in the same frame

InfiniteLee:
Normally this won't happen, but if your tab running Hubs is backgrounded, actions to add/remove bodies are queued up and won't run until Hubs regains focus (this is due to `tick` not running the `physics-system` when backgrounded). This adds guard clauses to ensure that if this happens, a body doesn't remain (incorrectly) partially initialized after it has been removed.

This also refactors the code to use a Map to contain all the various state associated with a body uuid to make it easier to read/understand.

### DIFF
--- a/src/systems/physics-system.js
+++ b/src/systems/physics-system.js
@@ -20,18 +20,8 @@ export class PhysicsSystem {
     this.bodyHelpers = [];
     this.shapeHelpers = [];
     this.bodyUuids = [];
-
     this.indexToUuid = {};
-
-    this.uuidToIndex = {};
-    this.object3Ds = {};
-    this.bodyOptions = {};
-    this.bodyLinearVelocities = {};
-    this.bodyAngularVelocities = {};
-
-    this.shapes = {};
-
-    this.collisions = {};
+    this.bodies = new Map();
 
     this.debugRequested = false;
     this.debugEnabled = false;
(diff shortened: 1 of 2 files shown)


src/systems/physics-system.js
```diff
@@ -155,9 +153,10 @@ export class PhysicsSystem {
         if (this.objectMatricesFloatArray.buffer.byteLength !== 0) {
           for (let i = 0; i < this.bodyUuids.length; i++) {
             const uuid = this.bodyUuids[i];
-            const index = this.uuidToIndex[uuid];
-            const type = this.bodyOptions[uuid].type ? this.bodyOptions[uuid].type : TYPE.DYNAMIC;
-            const object3D = this.object3Ds[uuid];
+            const body = this.bodies.get(uuid);
```
johnshaughnessy: maybe `bodyInfo` or `infoForBody` or something that indicates that this is a collection of various stuff associated with this body uuid

src/systems/physics-system.js
```diff
@@ -217,35 +213,39 @@ export class PhysicsSystem {
 
   addBody(object3D, options) {
     this.workerHelpers.addBody(this.nextBodyUuid, object3D, options);
-    this.object3Ds[this.nextBodyUuid] = object3D;
-    this.bodyOptions[this.nextBodyUuid] = options;
-    this.collisions[this.nextBodyUuid] = [];
-    this.bodyLinearVelocities[this.nextBodyUuid] = 0;
-    this.bodyAngularVelocities[this.nextBodyUuid] = 0;
+
+    this.bodies.set(this.nextBodyUuid, {
```
johnshaughnessy: Same as above -- maybe a name like `bodyInfos` or `bodyInfoForUuid` to indicate that it's a collection of stuff related to this body uuid

PR #2186, opened 2020-03-06
Disable batching by default

netpro2k:
Since it still seems to be causing stability and rendering issues on various platforms, the negatives seem to outweigh the potential benefits. We should likely still work on getting this back enabled by default, but it makes sense to disable it for now.

Also discovered that we were crashing the render loop in when exiting inspect mode if batching was disabled (which was already the case on platforms that don't support WebGL2 like iOS).

Batching can still be enabled via `forceMeshBatching` or `forceImageBatching`.

### DIFF
--- a/src/components/media-loader.js
+++ b/src/components/media-loader.js
@@ -519,9 +519,7 @@ AFRAME.registerComponent("media-loader", {
           { once: true }
         );
         this.el.addEventListener("model-error", this.onError, { once: true });
-        let batch =
-          !disableBatching &&
-          (forceMeshBatching || (AFRAME.utils.device.isMobile() && window.APP && window.APP.quality === "low"));
+        let batch = !disableBatching && forceMeshBatching;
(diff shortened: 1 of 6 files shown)


src/components/media-loader.js
```diff
@@ -457,7 +458,7 @@ AFRAME.registerComponent("media-loader", {
           { once: true }
         );
         this.el.setAttribute("floaty-object", { reduceAngularFloat: true, releaseGravity: -1 });
-        let batch = !disableBatching;
+        let batch = !disableBatching && forceImageBatching;
```
johnshaughnessy: Maybe defer to the `batchManagerSystem.batchingEnabled` instead of `disableBatching` (but keep the check for `forceImageBatching`)

netpro2k: Yeah was going to do that but then became slightly afraid of system initialization bugs. Will probably do that when enabling batching again though

Issue #2192, opened 2020-03-07
Right-click to inspect sometimes toggles


johnshaughnessy:
**Description**
Sometimes, when I am hovering on an object and I right click very quickly, I remain inspecting the object instead of needing to hold down my right mouse button. When I add some logging to the userinput system, it seems that it happens if, while I'm hovering on the inspectable, the path `"/device/mouse/buttonRight"` is `true`, `true`, and then `false`. That is, the path has `true` for only two generations (frames). It also fails when the right mouse button is `true` for one frame and then `false` on the next frame. 

Users have reported having this problems on trackpads, which I am guessing is doing a very rapid "click" that our bindings are not handling properly. I suspect the way our "xform state" is carrying over has a bug in it, in that the "falling" transform in the `"stopInspecting"` `binding` that becomes active when the `"inspecting"` action set becomes active _does not know about the right mouse button being true in the previous frame_ when it runs, so that it is incorrectly missing a "falling" happening. 

This seems plausible when the right mouse button is `true` on gen 1 and then `false` on gen 2, but I don't know why this would happen if it went `true` on gen 1, `true` on gen 2, and then `false` on gen 3. Here is a log showing some state in the first failure case:
```
active sets changed: ["global","leftHandHoveringOnNothing","leftCursorHoveringOnNothing","leftCursorHoldingNothing","rightCursorHoldingNothing","rightCursorHoveringOnInteractable"]
userinput.js:216 button right on gen 752 true
camera-system.js:369 inspect
userinput.js:479 active sets changed: ["global","leftHandHoveringOnNothing","leftCursorHoveringOnNothing","leftCursorHoldingNothing","rightCursorHoldingNothing","rightCursorHoveringOnInteractable","inspecting"]
userinput.js:216 button right on gen 753 false
userinput.js:479 active sets changed: ["global","leftHandHoveringOnNothing","leftCursorHoveringOnNothing","leftCursorHoldingNothing","rightCursorHoldingNothing","inspecting","rightHandHoveringOnNothing","rightCursorHoveringOnNothing"]
```
Here is a log showing the second failure case:
```
active sets changed: ["global","leftHandHoveringOnNothing","leftCursorHoveringOnNothing","leftCursorHoldingNothing","rightCursorHoldingNothing","rightCursorHoveringOnInteractable"]
userinput.js:216 button right on gen 37899 true
camera-system.js:369 inspect
[...]

PR #2193, opened 2020-03-07
Fix bug where a rapid click gets stuck in inspect mode

johnshaughnessy:
Fix https://github.com/mozilla/hubs/issues/2192


johnshaughnessy:
Actually, this does not fix the bug in cases where the right mouse button is down for only 1 or 2 frames. See issue for details.

Issue #2065, opened 2020-01-21
Cannot spawn emoji if permissions are locked down


gfodor:
**Description**
If you disable media spawning, this also prevents emoji spawning. It seems like we should add a distinct permission for emoji spawning.

PR #2207, opened 2020-03-11
Add permission checks for spawning emojis

### LINKED ISSUE #2154: Emoji permissions have conflicting UX
**Description**
When a user is not permitted to spawn objects, they cannot spawn emoji, but they still see the emoji menu.

**Expected behavior**
Either:
 - Spawning emojis should always be permitted OR
 - The emoji menu should either not show up or be grayed out (like the other UI elements that are blocked by permissions).

InfiniteLee:
Also adds a check in the `emoji` component that the `media-loader`'s `src` being used matches one of the emoji urls as defined in `emoji-hud`. 

See: https://github.com/mozilla/reticulum/pull/335

fixes #2154 and #2065

### DIFF
--- a/src/components/emoji.js
+++ b/src/components/emoji.js
@@ -17,9 +17,17 @@ AFRAME.registerComponent("emoji", {
   init() {
     this.data.emitEndTime = performance.now() + this.data.emitDecayTime * 1000;
     this.physicsSystem = this.el.sceneEl.systems["hubs-systems"].physicsSystem;
+
+    this.emojiHud = document.querySelector("[emoji-hud]").components["emoji-hud"];
   },
 
   play() {
+    const mediaLoader = this.el.components["media-loader"];
+    if (this.emojiHud.emojiUrls.indexOf(mediaLoader.data.src) === -1) {
+      this.el.parentNode.removeChild(this.el);
+      return;
+    }
+
     const uuid = this.el.components["body-helper"].uuid;
     this.lastLinearVelocity = this.physicsSystem.getLinearVelocity(uuid);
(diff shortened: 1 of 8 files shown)


src/components/emoji.js
```diff
@@ -17,9 +17,17 @@ AFRAME.registerComponent("emoji", {
   init() {
     this.data.emitEndTime = performance.now() + this.data.emitDecayTime * 1000;
     this.physicsSystem = this.el.sceneEl.systems["hubs-systems"].physicsSystem;
+
+    this.emojiHud = document.querySelector("[emoji-hud]").components["emoji-hud"];
   },
 
   play() {
+    const mediaLoader = this.el.components["media-loader"];
+    if (this.emojiHud.emojiUrls.indexOf(mediaLoader.data.src) === -1) {
```
johnshaughnessy: When does this happen / what does this guard against?

InfiniteLee: One of the concerns with moving emojis to their own permission instead of the "Create Media" permission is the potential room for exploit via setting the emoji's `media-loader` `src` property to any arbitrary media url. This guards against stuff other than emojis from being spawned after this permission check.

Emojis are spawned from `super-spawner-system` which create an `interactable-emoji` entity via `addMedia` https://github.com/mozilla/hubs/blob/fd28c27f7d23b9d69c0ffa5dbcc1f0a5dcbfb5ad/src/systems/super-spawner-system.js#L23-L32 (which adds the `media-loader` component. The `interactable-emoji` template contains this `emoji` component. When an emoji is spawned, this `play` will fire after the `media-loader` has been initialized, and validates the `src` set on the `media-loader`.

PR #2208, opened 2020-03-11
Add local refresh button for specific media

johnshaughnessy:
Adds a local-only "refresh" button to media loaded from twitch domains. Does not appear on non-live videos.

### DIFF
--- a/src/hub.html
+++ b/src/hub.html
@@ -230,10 +230,9 @@
                             is-remote-hover-target
                             tags="singleActionButton: true; isHoverMenuChild: true;"
                             local-refresh-media-button
-                            scale="0.8 0.8 0.8"
                             position="0 -0.6 0.001"
                         >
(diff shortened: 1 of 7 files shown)


src/hub.html
```diff
@@ -226,6 +226,17 @@
                     listed-media
                 >
                     <a-entity class="freeze-menu" visibility-while-frozen="withinDistance: 100;">
+                        <a-entity mixin="rounded-text-button ui"
```
johnshaughnessy: Weirdest part of this is that we want to have a networked refresh for admins and a local refresh for non-admins when there's a media-image that results from a twitch.tv url (e.g. the "stream offline" case). We don't give non-admins the networked refresh button because they aren't permitted to send NAF traffic for the object. This is fine, but means that admins have both buttons - networked refresh on hover and local refresh in the freeze menu.

PR #2210, opened 2020-03-12
Support for morph based animations and multiple looping animations

netpro2k:
- Fixes morph based animations not playing back correctly when inflated
- Adds support for playing multiple clips to `loop-animation` component. It is only added to the "deprecated" string based `clip` property for now since this is what we use in the blender addon. We should expand this into the numeric one as well, but making a numeric array property is a bit of a pain and trying to keep this as low risk as possible.

### DIFF
--- a/src/components/loop-animation.js
+++ b/src/components/loop-animation.js
@@ -14,13 +14,12 @@ AFRAME.registerComponent("loop-animation", {
 
   init() {
     this.mixerEl = findAncestorWithComponent(this.el, "animation-mixer");
+    this.currentActions = [];
 
     if (!this.mixerEl) {
       console.warn("loop-animation component could not find an animation-mixer in its ancestors.");
       return;
     }
-
-    this.updateClip();
   },
 
   update(oldData) {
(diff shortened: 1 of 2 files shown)


src/components/loop-animation.js
```diff
@@ -14,13 +14,12 @@ AFRAME.registerComponent("loop-animation", {
...
   init() {
     this.mixerEl = findAncestorWithComponent(this.el, "animation-mixer");
+    this.currentActions = [];
 
     if (!this.mixerEl) {
       console.warn("loop-animation component could not find an animation-mixer in its ancestors.");
       return;
     }
-
-    this.updateClip();
```
netpro2k: This is actually not needed since `update` is always called after `init`, so its just doing extra work

src/components/loop-animation.js
```diff
@@ -43,28 +44,30 @@ AFRAME.registerComponent("loop-animation", {
...
 
-    let clip;
-
+    let clips;
     if (clipName !== "") {
-      clip = animations.find(({ name }) => name === clipName);
+      clips = clipName.split(",").map(n => animations.find(({ name }) => name === n));
     } else {
-      clip = animations[activeClipIndex];
+      clips = [animations[activeClipIndex]];
```
netpro2k: We should decide if/how we want to support this for the index based property. Just avoided doing it for now since the happy path in the blender addon right now is string based, and having numeric array properties in aframe is a bit of a pain.

robertlong: I think we'll want it for use in Spoke. We can talk more about it when we want to add this feature to Spoke.

Issue #2141, opened 2020-02-14
Provide user pref to disable/override the auto-pixel-ratio system


johnshaughnessy:
**Is your feature request related to a problem? Please describe.**
Elgin uses a mac computer and often wants to take a high-resolution screenshot. The auto-pixel-ratio system automatically blurs her screen in an attempt to improve performance, but in this case she would prefer to have a low framerate with a crisp image.

**Describe the solution you'd like**
User pref to disable (temporarily or permanently) the auto-pixel-ratio system.

**Describe alternatives you've considered**
Lower the threshold for when this system kicks in, or try to improve performance in other ways. (What if we only render 1 object per frame? ;) )

brianpeiris:
@emclaren Would you prefer a setting in the preferences dialog, or would a query string parameter suffice?

emclaren:
thanks @johnshaughnessy for writing this up!

a query string would be just fine if its easier for you, however, long run I would prefer to have it in preferences so I could just toggle it to be always off and not have to worry about remembering to flag it every time I attend an event in Hubs. 

I also suspect this would be useful for other non-expert hubs users, because it may be confusing for them when the resolution drops suddenly mid-way through using hubs.

PR #2211, opened 2020-03-12
Lower the auto-pixel-ratio threshold. Allow it to increase again.

johnshaughnessy:
- Fixes https://github.com/mozilla/hubs/issues/2141 
- Adds a user preference to disable the auto-pixel-ratio system. (Also separates this into a new section called "advanced" in the preference list.)
- Reduces the "low FPS" threshold at which pixel ratio will decrease. 
- Adds a "high FPS" threshold at which pixel ratio will increase.
- Flip flopping between two pixel ratios repeatedly is prevented: If the system decreases the pixel ratio from `N` to `N-1` more than 3 times (`NUM_TIMES_DECREASED_BEFORE_CHANGING_MAX_PIXEL_RATIO`), then it sets `N-1` as the maximum pixel ratio and will not attempt `N` again.

### DIFF
--- a/src/assets/translations.data.json
+++ b/src/assets/translations.data.json
@@ -280,6 +280,7 @@
     "preferences.disableBackwardsMovement": "Disable backwards movement",
     "preferences.disableStrafing": "Disable strafing",
     "preferences.disableTeleporter": "Disable teleporter",
+    "preferences.disableAutoPixelRatio": "Disable automatic pixel ratio adjustments",
     "settings.return-to-vr": "Return to VR",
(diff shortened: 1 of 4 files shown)


johnshaughnessy:
Yes, that's correct @netpro2k 
In testing I found that it's totally possible that a certain pixel ratio value destroys framerate, and dropping it down one lower will cause it to recover completely. Since this is the case, I am going to adjust the algorithm such that in cases where it is flip-flopping between two values consistently (definition of "consistently" tbd), it drops it to the lower value and makes that the new "max pixel ratio" (preventing it from coming back up at all).

PR #2219, opened 2020-03-13
fix spawn_emoji permission if spawn_and_move_media permission is not enabled

InfiniteLee:
There are a handful of issues that occur if the `spawn_emoji` permission is enabled but the `spawn_and_move_media` permission is not. This fixes them so that the permission works by itself as intended.

### DIFF
--- a/src/utils/permissions-utils.js
+++ b/src/utils/permissions-utils.js
@@ -4,8 +4,11 @@ export function showHoverEffect(el) {
   const isFrozen = el.sceneEl.is("frozen");
   const isPinned = el.components.pinnable && el.components.pinnable.data.pinned;
   const isSpawner = !!el.components["super-spawner"];
+  const isEmoji =
+    el.components["super-spawner"] && el.components["super-spawner"].data.template === "#interactable-emoji";
   const canMove =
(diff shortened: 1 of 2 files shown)


src/utils/permissions-utils.js
```diff
@@ -14,15 +17,16 @@ export function canMove(entity) {
   const networkedTemplate = entity && entity.components.networked && entity.components.networked.data.template;
   const isCamera = networkedTemplate === "#interactable-camera";
   const isPen = networkedTemplate === "#interactable-pen";
-  const isEmoji = networkedTemplate === "#interactable-emoji";
+  const spawnerTemplate =
+    entity && entity.components["super-spawner"] && entity.components["super-spawner"].data.template;
+  const isEmoji = spawnerTemplate === "#interactable-emoji";
```
brianpeiris: Should this be `networkedTemplate === "#interactable-emoji" ||  spawnerTemplate === "#interactable-emoji"`? Because I think this change implies that you can't move other peoples' emoji

InfiniteLee: That was indeed the case, though I just saw this comment and just pushed an update to address that :)

PR #2224, opened 2020-03-14
add permission for enabling flying

InfiniteLee:
see: https://github.com/mozilla/reticulum/pull/343

### DIFF
--- a/src/message-dispatch.js
+++ b/src/message-dispatch.js
@@ -52,11 +52,12 @@ export default class MessageDispatch {
     switch (command) {
       case "fly":
         if (this.scene.systems["hubs-systems"].characterController.fly) {
-          this.scene.systems["hubs-systems"].characterController.fly = false;
+          this.scene.systems["hubs-systems"].characterController.enableFly(false);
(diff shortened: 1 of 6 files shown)


src/message-dispatch.js
```diff
@@ -52,11 +52,12 @@ export default class MessageDispatch {
     switch (command) {
       case "fly":
```
johnshaughnessy: If flying is disabled globally, this command should be ignored and should not appear in the help text when you type "/"

InfiniteLee: I'm wondering now if maybe if you try to enable fly when it's disabled, we message to the user that fly is disabled on the room at the moment.

InfiniteLee:
`May need a coherent strategy for handling / explaining the other cases where fly mode is entered besides the one when the user explicitly presses the toggle fly key: (Entering a scene with no nav-mesh, activating a waypoint that is far above the nav-mesh that does not have "snapToNavMesh" enabled, toggling streamer mode, being a ghost)`
Right now I'm leaving those cases as-is. I figure in those situations, your scene is basically not setup correctly, and so we should default to something sane that doesn't look/feel terrible (fly mode).

Issue #2189, opened 2020-03-06
Spawning on quest at wrong height


johnshaughnessy:
**Description**
Sometimes when we spawn on quest, the user is not at the correct height. Possibly also not in the correct place. (I ended up underground unexpectedly in the default "river" scene.)

**To Reproduce**
Unreliably - enter a scene in quest.

**Expected behavior**
Depending on the flags for the waypoint, we should either be matching the eye-height exactly or snapping to the floor underneath the waypoint using the user's reported sittingToStandingMatrix.

**Screenshots**

**Hardware** 
- Quest. Oculus Browser. (Possibly FxR - have not tested.)

saitogroup:
Same. I'm getting spawned below the ground, even though my spawn point is at the origin.

saitogroup:
Works fine in the broswer.

Issue #2225, opened 2020-03-14
Player height at spawn time is surprisingly tall


johnshaughnessy:
**Description**
When spawning into the room on quest, `getCurrentPlayerHeight` reports something like 2.6 meters even though my vr devices `sittingToStandingMatrix` reports something much less than this. This means that the avatar rig and the avatar pov are unusually far apart or out of sync for reasons unknown.

**To Reproduce**
Spawn into a room on quest. 
Print `getCurrentPlayerHeight` from inside the waypoint travel code.

**Expected behavior**
The distance between the avatar rig and avatar pov node should be something like your sittingToStandingMatrix (larger in world space if your scale is not 1). 

Quest

**Additional context**
~Contributed to this bug: https://github.com/mozilla/hubs/issues/2189~
~The reason https://github.com/mozilla/hubs/pull/2226 fixes the bug above is because if the player height is at the (surprising) value of 2.6 meters, we have to use that surprising value. However, it should not be this far apart so we need to figure out why that is happening.~

Edit: This issue and https://github.com/mozilla/hubs/issues/2189 are both fixed by #2226 . We were spawning into the spawn point without taking the vr display data into account. Delaying `moveToSpawnPoint` fixes the issue.

johnshaughnessy:
The unusually tall height seems to be related to moving the `viewing-camera` in the `scene-preview-camera-system` without also moving the `viewing-rig`. If I make sure that the `viewing-rig` always moves when the `viewing-camera` moves (as in https://github.com/mozilla/hubs/commit/120234bbf2dd0c277b55fc9e8ef5ef767d06bb20 ), then the reported height at the time we travelByWaypoint is `0`, which makes sense because there is no height offset in `hub.html` between `#viewing-rig` and `#viewing-camera`. 
This is also wrong, because by the time we are spawning we want to be using the `sittingToStandingTransform` from the vr display in some way. We need to `moveToSpawnPoint` _after_ the vr display takes over the viewing camera.

johnshaughnessy:
Keeping the `viewing-rig` in sync with the `viewing-camera` in the `scene-preview-camera-system` is unnecessary. Delaying `moveToSpawnPoint` until after vr is enabled on the renderer seems to fix the issue altogether. Fixed in this commit: https://github.com/mozilla/hubs/pull/2226/commits/e5f6342819548435fc8998593771dd415f071b90

brianpeiris:
Can we call this closed?

[...]

PR #2226, opened 2020-03-14
Fix spawn height

johnshaughnessy:
Fixes https://github.com/mozilla/hubs/issues/2189

This PR addresses two problems. The first was that when we spawned at a waypoint that was not snapped to the nav mesh, we used a default height of 1.6 meters. If the current player height was different than that, we would end up above or below the floor. The second problem is that we are using the delta between `avatar-rig` and `avatar-pov-node` to calculate player height, so we need to delay spawning at the waypoint until after these reflect the reported height of the player.

### DIFF
--- a/src/systems/character-controller-system.js
+++ b/src/systems/character-controller-system.js
@@ -124,10 +124,8 @@ export class CharacterControllerSystem {
         inPosition.setFromMatrixPosition(inMat4Copy);
         this.findPositionOnNavMesh(inPosition, inPosition, outPosition, true);
         finalPOV.setPosition(outPosition);
-        translation.makeTranslation(0, getCurrentPlayerHeight(), -0.15);
-      } else {
-        translation.makeTranslation(0, 1.6, -0.15);
       }
+      translation.makeTranslation(0, getCurrentPlayerHeight(), -0.15);
       finalPOV.multiply(translation);
       if (willMaintainInitialOrientation) {
         initialOrientation.extractRotation(this.avatarPOV.object3D.matrixWorld);
--- a/src/utils/get-current-player-height.js
+++ b/src/utils/get-current-player-height.js
@@ -4,8 +4,8 @@ export const getCurrentPlayerHeight = (function() {
   return function getCurrentPlayerHeight(world) {
     avatarPOV = avatarPOV || document.getElementById("avatar-pov-node");
     avatarRig = avatarRig || document.getElementById("avatar-rig");
-    avatarPOV.object3D.updateMatrices();
     avatarRig.object3D.updateMatrices();
+    avatarPOV.object3D.updateMatrices();
     if (world) {
       return avatarPOV.object3D.matrixWorld.elements[13] - avatarRig.object3D.matrixWorld.elements[13];
     }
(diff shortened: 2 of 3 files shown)


src/utils/get-current-player-height.js
```diff
@@ -4,8 +4,8 @@ export const getCurrentPlayerHeight = (function() {
   return function getCurrentPlayerHeight(world) {
     avatarPOV = avatarPOV || document.getElementById("avatar-pov-node");
     avatarRig = avatarRig || document.getElementById("avatar-rig");
-    avatarPOV.object3D.updateMatrices();
     avatarRig.object3D.updateMatrices();
+    avatarPOV.object3D.updateMatrices();
```
johnshaughnessy: This reordering is unrelated to the bug, but it just seems strange to me to update child then parent instead of parent then child.

src/systems/character-controller-system.js
```diff
@@ -124,10 +124,8 @@ export class CharacterControllerSystem {
         inPosition.setFromMatrixPosition(inMat4Copy);
         this.findPositionOnNavMesh(inPosition, inPosition, outPosition, true);
         finalPOV.setPosition(outPosition);
-        translation.makeTranslation(0, getCurrentPlayerHeight(), -0.15);
-      } else {
-        translation.makeTranslation(0, 1.6, -0.15);
       }
+      translation.makeTranslation(0, getCurrentPlayerHeight(), -0.15);
```
johnshaughnessy: Player height at spawn point is too big, but if we don't use it we spawn under the floor. Investigating the underlying issue here: https://github.com/mozilla/hubs/issues/2225

johnshaughnessy: Underlying issue resolved by moving to spawn point after entering vr. https://github.com/mozilla/hubs/pull/2226/commits/e5f6342819548435fc8998593771dd415f071b90

johnshaughnessy:
As I was trying to understand the problem, I created another issue with discussion as I figured stuff out. This is probably not important but I will link it for posterity: https://github.com/mozilla/hubs/issues/2225

johnshaughnessy:
The relationship between the locations of the viewing camera/rig, the avatar pov/rig, the height of the avatar, and the reported height of the VR device is made complicated by our use of the THREE.VRManager. Our reliance on this makes it so that updating the camera transform is that last thing that happens before rendering the frame, when in fact we wish it was (one of) the first things. I will track this tech debt in a separate issue. (https://github.com/mozilla/hubs/issues/2229)

One last thing that we should be aware of is that the scene-preview-camera-system moves the `viewing-camera` without moving the `viewing-rig`, so at the time when you click "enter room", the delta between them could be arbitrarily big. It seemed to be the case that this delta ended up being equal to the currentPlayerHeight (the delta between `avatar-rig` and `avatar-pov-node`) because I believe we copy one to the other. When we were spawning underground, it seemed to be because this delta was ~2.8 meters whereas most peoples' actual height is typically less than 2 meters ( and regardless, we had hardcoded the spawning code to use 1.6 meters ). We need to be especially careful during this transition from lobby to room (which unfortunately does not happen within the control flow of a regular update loop, and instead is in a click handler that awaits async steps while entering, which makes it challenging to know _when_ things will happen relative to the nearest frame).

PR #2255, opened 2020-03-20
Avoid unnecessary CORS proxying via pre-emptive checks

gfodor:
WIP, do not merge.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-831)


robertlong:
Should probably revisit this at a later date.

emstanley:
Adding to needs review.

netpro2k:
We still probably want to do this but if we do it will be in a new codepath anyway. Closing this.

PR #2274, opened 2020-03-23
Add links to docs

robertlong:
Adds links to the documentation to the home page, help modal in room, and settings menu.


src/react-components/home-root.js
```diff
@@ -203,6 +203,9 @@ class HomeRoot extends Component {
                       <FormattedMessage id="editor-name" />
                     </a>
                   </IfFeature>
+                  <a href="https://hubs.mozilla.com/docs" rel="noreferrer noopener">
+                    <FormattedMessage id="home.docs_link" />
+                  </a>
```
brianpeiris: We should probably make this configurable for cloud users. Either disable the link by default and let them turn it on, or allow them to configure the link. You just need to add the corresponding configs to `src/schema.toml`

brianpeiris: See the existing `features.show_community_link` and `links.community` configs as examples

johnshaughnessy: Not sure if this is the same thing (having a difficult time tracking down where this configuration lives), but this may be how the `HelpDialog` is configured: https://github.com/mozilla/hubs/blob/master/src/react-components/help-dialog.js#L20

brianpeiris: `configs.link("features", messages["help.docs_url"])` means "take the configuration from the admin panel for the `links.features` config, or if that doesn't exist, use the `help.docs_url` string from the `translations.data.json` file"

robertlong:
@brianpeiris thanks for the review. Could you check it again? I think I did things right, but Idk how to test it locally.

robertlong:
I still need to make the run-local-reticulum script parameterized to enable USE_FEATURE_CONFIG

robertlong:
Nvm, I'm gonna split out the rest of the DX work into another PR. I've reverted defaulting the USE_FEATURE_CONFIG variable to true for now.

HeadClot:
I am super excited by this :D

PR #2277, opened 2020-03-24
Add hidden slash command to disable/enable positional audio

InfiniteLee:
This adds `/audiomode` which will disable/renable positional audio at runtime (persists only for the session). This is to be used to help debug and diagnose clients where WebRTC noise cancellation is not working. When positional audio is disabled, avatar and video audio will attenuate volume by distance based on inverse square formula.

### DIFF
--- a/src/message-dispatch.js
+++ b/src/message-dispatch.js
@@ -140,6 +140,14 @@ export default class MessageDispatch {
           }
         }
         break;
+      case "audiomode":
+        if (window.APP.store.state.preferences.audioOutputMode !== "audio") {
+          window.APP.store.state.preferences.audioOutputMode = "audio";
+          this.log("Positional Audio disabled.");
+        } else {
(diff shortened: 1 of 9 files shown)


src/message-dispatch.js
```diff
@@ -140,6 +140,14 @@ export default class MessageDispatch {
           }
         }
         break;
+      case "audiomode":
```
johnshaughnessy: Thoughts on adding this to the user preference screen?

InfiniteLee: Originally was going to be, but held off for now because we don't actually know that this will help noise cancellation problems. Can easily re-enable it in the preferences if we prove that this helps.

arpu:
should noise cancellation not be set with noiseSuppression: true in getUserMedia ?

InfiniteLee:
@arpu we have a theory that audio spatialization causes noise cancellation to not work properly in some cases- this is to help debug those situations.

PR #2279, opened 2020-03-25
Improve UI/UX of "Watch from Lobby" button.

johnshaughnessy:
Reorders the buttons and makes sure the size is correct by using `<button>`, not `<a>`.
Also I confirmed that this works on my mobile device without overflowing.
image


johnshaughnessy:
Also reduces my 
shame
shame.

PR #2286, opened 2020-03-26
Fix race condition in tar step of deploy

robertlong:
This bug seems to only happen on Mac. Making the tar step synchronous seems to fix it. The absolute path may not be necessary, but is a good practice anyway.


antpb:
AH! This has been bugging me for months! I've been working on my mac and deploying from my Windows machine. Just wanted to chime in and say this PR works for Mac. May need a Windows confirmation still.

PR #2315, opened 2020-04-01
Try fixing worker fork deploy

gfodor:


robertlong:
@gfodor could you provide more info on what this was intended to fix?

netpro2k:
From what I can tell this no longer applies. Closing.

PR #2321, opened 2020-04-03
Add sign-up form for newsletter to home page

johnshaughnessy:
- Adds support for our email mailing list. 
image
(Final styling TBD and can be edited in mailchimp.)
 
- Adds a link to the footer to subscribe to the newsletter. 
image

 - Updates the "Join Us" modal copy and adds the subscribe to newsletter link.
image

 
 - Adds links to the `Sign in` (in progress) modal and the `Sign in` (success) modals.
image

 - Adds an unbranded hero background image for use in local development.
image

image


src/utils/configs.js
```diff
@@ -74,7 +75,8 @@ if (isLocalDevelopment) {
   localDevImages = {
     logo: appLogo,
     company_logo: companyLogo,
-    editor_logo: sceneEditorLogo
+    editor_logo: sceneEditorLogo,
+    home_background: homeHeroBackground
```
brianpeiris, line 79: nit: these local dev images were meant to be unbranded placeholders, so if we do need to use home-hero-background here, I would replace it with a generic image instead of one that says "mozilla" in it. Not a huge deal, but I've generally been trying to remove all branded assets from the client so that we don't worry about having to enforce users accidentally showing them on forks going forward.

src/assets/stylesheets/index.scss
```diff
@@ -356,3 +356,45 @@ body {
...
+  height: 40px;
+}
+
+:local(.email-input) {
+  height: 40px;
+  border-radius: 24px;
+  margin: 0px 16px 0px 0px;
+  width: 24em;
+  text-align: center;
+}
```
brianpeiris, line 396: Is there a reason this input deviates from our existing design for inputs? I would try to use our existing classes to maintain design consistency or at least try to match the style.
In particular, I'm talking about the shape of the input, the border around it and the font family, font alignment and font size of the input text:
image
image

src/assets/stylesheets/index.scss
```diff
@@ -356,3 +356,45 @@ body {
...
+
+:local(.subscribeBar) {
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  justify-content: center;
+}
+
+:local(.subscribe-text) {
+  font-size: 14pt;
```
brianpeiris, line 380: nit: this makes the sign up text the biggest text on the page of its style. It breaks the visual hierarchy a bit, unless it is our intention to direct the user's eye to the sign up form, I would omit the font-size here and let it use the default page font size.

src/react-components/home-root.js
```diff
@@ -275,7 +275,55 @@ class HomeRoot extends Component {
...
+                            type="email"
+                            name="EMAIL"
+                            className={styles.emailInput}
+                            id="mce-EMAIL"
+                          />
+                        </div>
+                        <div id="mce-responses" className="clear">
+                          <div className="response" id="mce-error-response" style={{ display: "none" }} />
+                          <div className="response" id="mce-success-response" style={{ display: "none" }} />
+                        </div>
```
brianpeiris, line 307: There look like they will never be used? Would be nice to clean up this form to remove the unused elements and `id` attributes.

src/assets/stylesheets/index.scss
```diff
@@ -356,3 +356,45 @@ body {
   }
 }
 
+#mc_embed_signup {
```
brianpeiris, line 359: nit: I would continue using `:local(.class-names)` where possible. I don't see a reason for this to be an id.

src/react-components/home-root.js
```diff
@@ -275,7 +275,55 @@ class HomeRoot extends Component {
...
                 </div>
               )}
+              <IfFeature name="show_newsletter_signup">
+                <div id="mc_embed_signup" className={styles.footerContent}>
+                  <form
+                    action="https://mozilla.us19.list-manage.com/subscribe/post?u=a97038e0879516f7b8300d67f&amp;id=ebf17d04e4"
+                    method="post"
+                    id="mc-embedded-subscribe-form"
+                    name="mc-embedded-subscribe-form"
+                    className="validate"
```
brianpeiris, line 285: I don't think we have a `validate` class. Same for the `clear` classNames below.

src/react-components/home-root.js
```diff
@@ -275,7 +275,55 @@ class HomeRoot extends Component {
...
+                    target="_blank"
+                    noValidate
```
brianpeiris, line 287: Why wouldn't we want to validate that the email input is an email?

src/react-components/home-root.js
```diff
@@ -275,7 +275,55 @@ class HomeRoot extends Component {
...
+                  >
+                    <div id="mc_embed_signup_scroll">
+                      <div className={styles.subscribeText}>
+                        <FormattedMessage id="home.subscribe_to_mailing_list" />
+                      </div>
+                      <div className={styles.subscribeBar}>
+                        <div className="mc-field-group">
+                          {/* <label htmlFor="mce-EMAIL">Email Address </label> */}
```
brianpeiris, line 295: Just delete this.

src/react-components/home-root.js
```diff
@@ -275,7 +275,55 @@ class HomeRoot extends Component {
...
+                            id="mce-EMAIL"
+                          />
+                        </div>
+                        <div id="mce-responses" className="clear">
+                          <div className="response" id="mce-error-response" style={{ display: "none" }} />
+                          <div className="response" id="mce-success-response" style={{ display: "none" }} />
+                        </div>
+                        <div style={{ position: "absolute", left: "-5000px" }} aria-hidden="true">
+                          <input type="text" name="b_a97038e0879516f7b8300d67f_ebf17d04e4" tabIndex="-1" value="" />
+                        </div>
```
brianpeiris, line 310: What is this for? Do we need it?

src/assets/stylesheets/index.scss
```diff
@@ -356,3 +356,45 @@ body {
...
+
+:local(.email-input) {
+  height: 40px;
+  border-radius: 24px;
+  margin: 0px 16px 0px 0px;
+  width: 24em;
+  text-align: center;
+}
+
+.mc_field-group {
```
brianpeiris, line 398: Does this have to be a bare classname instead of a `:local` class?

src/assets/stylesheets/index.scss
```diff
@@ -356,3 +356,45 @@ body {
   }
 }
 
+#mc_embed_signup {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+}
```
brianpeiris, line 364: This form is cut off on mobile screens:
image

src/utils/configs.js
johnshaughnessy: Replaced

src/assets/stylesheets/index.scss
johnshaughnessy: Good point. I started to `@extend %input-field` but then removed the in-page sign up entirely.

src/react-components/home-root.js
johnshaughnessy: My mistake -- Was copying the form as given by mailchimp. Not sure why they suggested noValidate here.

src/react-components/home-root.js
johnshaughnessy: This was to guard against Robots, but I'm not sure how. May only work if you add their javascript to the page, too.

src/assets/stylesheets/index.scss
johnshaughnessy: No, good point.

src/react-components/home-root.js
```diff
@@ -56,9 +56,7 @@ class HomeRoot extends Component {
 
   state = {
     dialog: null,
-    signedIn: null,
-    mailingListEmail: "",
```
johnshaughnessy, line 60: Ghosts of mailing lists past

PR #2326, opened 2020-04-03
Improve menu placement

johnshaughnessy:
This change makes sure the menus on 3D objects are upright and in a pretty good position most of the time.

3D media have menus placed at their bounding sphere (when you are 2+ meters outside the bounding sphere), else the menus are placed close to the cursor intersection point or, if there is no intersection point (which will happen with touchscreen controls), a fraction of the way from the camera to the mesh center.  
Flat media is treated the same as before.

The physics system is not used and no `convexSweepTest` is performed. Render order and depth test are unchanged. (Commits that use these techniques are still in the history, but I can rebase before deploying if desired.) 

Fixes https://github.com/mozilla/hubs/issues/1187


misslivirose:
I gave a little test to this yesterday and really loved how the menus were usable even when scaling up models.

src/systems/menu-animation-system.js
```diff
@@ -0,0 +1,73 @@
...
+    const cameraPosition = new THREE.Vector3();
+    return function tick(t) {
+      if (!this.viewingCamera) {
+        return;
+      }
+      this.viewingCamera.updateMatrices();
+      cameraPosition.setFromMatrixPosition(this.viewingCamera.matrixWorld);
+
+      for (let i = 0; i < this.components.length; i++) {
+        const datum = this.data.get(this.components[i]);
```
netpro2k, line 42: Seems like the only time we ever get the component data is here when iterating, so we can probably ditch the array + Map and just use an array with the data. You will need to iterate to find the item again for `unregister` but you are already calling `indexOf` on the array anyway, so it already is.

src/utils/auto-box-collider.js
```diff
@@ -1,13 +1,70 @@
+function isVisibleUpToRoot(node, root) {
+  let child = node;
+  while (child) {
+    if (child === root) return true;
+    if (!child.visible) return false;
+    child = child.parent;
+  }
+  console.error(`Root ${root} is not in the hierarchy of node ${node}`);
```
netpro2k, line 8: "in the hierarchy of" to me reads like it should be a child even though I think its technically correct. I might be more specific and say "is not an ancestor of"

src/utils/auto-box-collider.js
```diff
@@ -1,13 +1,70 @@
...
+export const computeLocalBoundingBox = (function() {
+  const vertex = new THREE.Vector3();
+  const rootInverse = new THREE.Matrix4();
+  const toRootSpace = new THREE.Matrix4();
+  return function computeLocalBoundingBox(root, box, excludeInvisible) {
+    box.makeEmpty();
+    root.updateMatrices();
+    rootInverse.getInverse(root.matrixWorld);
+    root.traverse(node => {
+      if (excludeInvisible && !isVisibleUpToRoot(node, root)) {
```
netpro2k, line 21: Hmm slightly scary here that we are calling `isVisibleUpToRoot` here on every node in the subtree, lots of duplicate work happening.. Might make sense to do the traverse manually and roll in the visibility check to cull out duplicating work. That or we come up with a similar solution to matrixNeedsUpdate but for visibility, which sounds painful.

src/utils/easing.js
```diff
@@ -1,3 +1,24 @@
...
+  amplitude = 1,
+  period = 0.3;
+export const elasticOut = (function custom(a, p) {
+  const s = Math.asin(1 / (a = Math.max(1, a))) * (p /= tau);
+
+  function elasticOut(t) {
+    return 1 - a * Math.pow(2, -10 * (t = +t)) * Math.sin((t + s) / p);
+  }
+
+  elasticOut.amplitude = function(a) {
```
netpro2k, line 16: Confused what these are for? Same goes for the self executing function, seems like you can just do:
```javascript
const tau = 2 * Math.PI,
  a = 1,
  p = 0.3,
  s = Math.asin(1 / (a = Math.max(1, a))) * (p /= tau); 
export  function elasticOut(t) {
    return 1 - a * Math.pow(2, -10 * (t = +t)) * Math.sin((t + s) / p);
}
```
(though pick better variable names or surround with {})

src/utils/scene-graph.js
```diff
@@ -4,3 +4,21 @@ export function findAncestorWithComponent(entity, componentName) {
   }
   return entity;
 }
+
+export function findComponentsInNearestAncestor(entity, componentName) {
+  const components = [];
+  while (entity) {
+    if (entity.components) {
+      for (const c in entity.components) {
+        if (c.startsWith(componentName)) {
```
netpro2k, line 13: Would not have guessed in the code above using this that it was using a prefix search, though I guess that is the only way it would return multiple components from a single entity given a single component name... But then I am pretty confused by usage to find "position-at-border"

netpro2k: Ah.. I forgot how aframe handles multiple instances of the same component with a prefix... I think there is a component property that actually has the component name, so you can avoid the `startsWith` check which I am sure will bite us at some point.

src/systems/hubs-systems.js
```diff
@@ -80,10 +82,11 @@ AFRAME.registerSystem("hubs-systems", {
     this.soundEffectsSystem.tick();
     this.scenePreviewCameraSystem.tick();
     this.physicsSystem.tick(dt);
-    this.spriteSystem.tick(t, dt);
     this.batchManagerSystem.tick(t);
     this.cameraSystem.tick(this.el, dt);
     this.waypointSystem.tick(t, dt);
+    this.menuAnimationSystem.tick(t);
+    this.spriteSystem.tick(t, dt);
```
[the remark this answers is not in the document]
johnshaughnessy: Yes -- We want the sprite system to update _after_ I animate the menus and move the camera

src/components/visibility-while-frozen.js
```diff
@@ -111,7 +112,8 @@ AFRAME.registerComponent("visibility-while-frozen", {
       allowed &&
       ((isFrozen && this.data.visible) || (!isFrozen && !this.data.visible)) &&
       isWithinDistance &&
-      !isTransforming;
+      !isTransforming &&
+      !isHoldingAnything;
```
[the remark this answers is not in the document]
johnshaughnessy: You have to display the menus in order to move a pinned object, but once you pick it up the menus should disappear. This is akin to having the menus disappear while transforming it.

src/utils/auto-box-collider.js
johnshaughnessy: You're right, but since this is called very infrequently I'm not very concerned. (Once per mesh, and once per opened menu each time a camera is toggled, but the menus do not have deep hierarchies.)

PR #2340, opened 2020-04-07
Improved Developer Experience: Part 1

robertlong:
This PR addresses the developer experience of the npm run scripts in Hubs.

## Starting Hubs
For the majority of developers you will want to use `npm start` just as you would before. However, this script now fetches the app config and environment variables from the server you are logged into. Meaning, you'll now see the page as will look when you deploy it. The features, images, etc. will all be loaded and injected into the page. If you are developing for Hubs Cloud, you will no longer need to use the `use-hubs-cloud-stack.sh` script.

You will now need to run `npm login` before running `npm start`. For Hubs Cloud developers, this has not changed, but for hubs.mozilla.com developers or community members developing against `dev.reticulum.io` this will be new to you. You will need to be an admin on `dev.reticulum.io` to log in. This can likely change as we don't expose any secrets in `/api/v1/app_configs`.

For users who wish to develop against dev.reticulum.io, but don't have an admin account. Or if you just want to use the default app config, there is now `npm run dev` which connects to `dev.reticulum.io` but loads the default app config with all features enabled.

If you are running reticulum locally, you used to run hubs with `./scripts/run-local-reticulum.sh`. You can now use `npm run local`. This uses the default app config, but can be configured to load the app config from the local reticulum instance with the `LOAD_APP_CONFIG` environment variable.

## Script Guide:

 - `npm run login` Log into your Hubs Cloud stack or dev.reticulum.io
 - `npm start` Run Hubs against the server you are logged into
 - `npm run dev` Run Hubs against dev.reticulum.io with all features enabled
 - `npm run local` Run Hubs against a local reticulum server with all features enabled
 - `npm run deploy` Deploy Hubs Cloud to the stack you are logged into.
 - `npm run logout` Log out of the current Hubs Cloud stack

### Environment Variables
 - `LOAD_APP_CONFIG=true` Load the app config from the configured server.


InfiniteLee:
what happens if you run `npm start` without running `npm login`? does it default to the old behavior e.g. what `npm run dev` does now?

robertlong:
@InfiniteLee it doesn't, it gives you an error message telling you to login. The reason being that `npm start` is what most node developers reach for first and `hubs-cloud` will be the common use case. Hubs devs can still use `npm start` you just need to run `npm run login` and configure it to use `dev.reticulum.io` first. Personally, I think I'll be using it that way a majority of the time. I'll pull out `npm run dev` when I need all the features enabled.

src/utils/configs.js
```diff
@@ -34,6 +34,8 @@ let isAdmin = false;
 
 configs.AVAILABLE_INTEGRATIONS = window.AVAILABLE_INTEGRATIONS || {};
 
+console.log(process.env.APP_CONFIG, window.APP_CONFIG);
```
netpro2k, line 37: console log

PR #2343, opened 2020-04-08
Use a larger key size when generating a self-signed cert.

jdm:
This allows browsers with more restrictive SSL settings (such as Servo) to reach the local dev server.


brianpeiris:
Thanks. Looks good to me. I'll ship this in my next push.

PR #2353, opened 2020-04-09
Tweak object list button

johnshaughnessy:
- Convert object list div into a button. 
- Fix dimensions in widescreen and mobile views to be a known, stable size across browsers. 
- Fix drop-down location so that it does not block the object list button (as it did in the small-screen view).

Work is ongoing, but this commit can be deployed as-is, if desired.


johnshaughnessy:
WIP - I want to make sure both buttons in the top right match each other.

PR #2356, opened 2020-04-09
Minor UI improvements

johnshaughnessy:
- Converts the object list and presence list `div`s into `button`s to assist screen readers.
- Applies uniform styling to both so that they are the same size as each other
- Fixes object list placement on small screens so that it does not cover the object list button when open.
- Shows (very minimal) info when the object list is empty and the button is clicked, instead of doing nothing.
- Tweaks the `top` values so that all the upper buttons are the same distance from the top of the page.
- Move choose scene button closer to share button on small screens to make better use of limited screen space.

### DIFF
--- a/src/react-components/object-list.js
+++ b/src/react-components/object-list.js
@@ -1,4 +1,5 @@
 import React, { Component } from "react";
+import { FormattedMessage } from "react-intl";
 import PropTypes from "prop-types";
 import classNames from "classnames";
 import rootStyles from "../assets/stylesheets/ui-root.scss";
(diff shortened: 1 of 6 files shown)


src/react-components/object-list.js
```diff
@@ -142,9 +142,17 @@ export default class ObjectList extends Component {
...
-      <div className={styles.presenceList}>
+      <div className={rootStyles.objectList}>
         <div className={styles.contents}>
-          <div className={styles.rows}>{this.state.mediaEntities.map(this.domForEntity.bind(this))}</div>
+          <div className={styles.rows}>
+            {this.state.mediaEntities.length ? (
+              this.state.mediaEntities.map(this.domForEntity.bind(this))
+            ) : (
+              <div className={styles.listItem}>
+                <span>There is no media in the scene yet.</span>
```
brianpeiris, line 152: Use a `<FormattedMessage />` instead with the text in translations.data.json.

brianpeiris: nit: I would remove the period at the end of the text and center the text.

brianpeiris: The message is cut off on mobile screens: 
image

PR #2361, opened 2020-04-10
Enable echo cancellation (AEC) on Chrome

InfiniteLee:
This patch adds a workaround that enables AEC on desktop Chrome/Chromium. It reroutes all audio post-WebAudio processing (e.g. spatialization and gain adjustments) into a loopback RTCPeerConnection that allows Chromium's AudioMixer to run AEC properly. This is a known issue in Chromium that is tracked here: https://bugs.chromium.org/p/chromium/issues/detail?id=687574

Thanks to @padenot for these drawings that explain why this workaround is needed:
firefox
chromium
chromium-workaround

Note: Chrome on Android is supposed to use hardware AEC, but that might not be the case on Oculus Browser; more testing needed.


gfodor:
This is huge - nice work!

src/systems/audio-system.js
```diff
@@ -0,0 +1,78 @@
...
+  constructor(sceneEl) {
+    sceneEl.audioListener = sceneEl.audioListener || new THREE.AudioListener();
+    if (sceneEl.camera) {
+      sceneEl.camera.add(sceneEl.audioListener);
+    }
+    sceneEl.addEventListener("camera-set-active", evt => {
+      evt.detail.cameraEl.getObject3D("camera").add(sceneEl.audioListener);
+    });
+
+    if (!AFRAME.utils.device.isMobile() && /chrome/i.test(navigator.userAgent)) {
```
brianpeiris, line 74: Might want to explicitly exclude Oculus Browser here, since they also include "Chrome" in their UA and future versions will default to desktop mode.

src/systems/audio-system.js
```diff
@@ -0,0 +1,78 @@
...
+          .createAnswer()
+          .then(answer => {
+            inboundPeerConnection.setLocalDescription(answer).catch(onError);
+
+            outboundPeerConnection.setRemoteDescription(answer).catch(onError);
+          })
+          .catch(onError);
+      })
+      .catch(onError);
+  });
```
brianpeiris, line 61: nit: could make this more readable with `awaits`.

src/systems/audio-system.js
arpu: is this not needed on OculusBrowser?

brianpeiris: I don't think we've repro'd echo cancellation issues on Oculus Browser.

InfiniteLee: I've heard anecdotally that there have been echo issues on Oculus Browser, so I plan on testing that a bit today and seeing if this PR helps with that or not. But otherwise, doesn't checking `!AFRAME.utils.device.isMobile()` effectively exclude oculus browser?

brianpeiris: `isMobile` might stop working on Quest if they switch their browser to desktop mode by default, which they are currently rolling out in trials.

InfiniteLee: After doing some testing, it doesn't look like `isMobile` returns true in Oculus Browser right now- but it also looks like hardware AEC doesn't work on Quest. So we want to enable this feature on Oculus Browser anyways (but keep it disabled on true mobile browsers e.g. android chrome because they should be using hardware AEC).

InfiniteLee: Well, turns out this workaround doesn't seem to work on Oculus Browser, so going to change to 
```
    if (/chrome/i.test(navigator.userAgent) && !/OculusBrowser/i.test(window.navigator.userAgent)) {
```

Will come back to this in the future and figure out why it's not working.

arpu: would be interessting if AEC works in OCulus go ( different Hardware) maybe @Artyom17 has more infos on this?

arpu:
looks like the audio-context-fix for user gesture does not work anymore with this 
https://github.com/mozilla/hubs/blob/master/src/utils/audio-context-fix.js

after user click and webaudio Context is resumed and audio play() again not can be 
heard

InfiniteLee:
> 
> 
> looks like the audio-context-fix for user gesture does not work anymore with this
> https://github.com/mozilla/hubs/blob/master/src/utils/audio-context-fix.js
> 
> after user click and webaudio Context is resumed and audio play() again not can be
> heard

Ah I totally forgot about this hack. I'm not sure why I wasn't running into this on desktop when testing, but it looks like this is what was preventing this workaround from working on Quest as well! Thanks!

arpu:
for reference  i tested it with localhost and this chrome start flags ```-disable-features=PreloadMediaEngagementData,AutoplayIgnoreWebAudio,MediaEngagementBypassAutoplayPolicies``` 

works now perfect!

trcoffman:
@InfiniteLee, When will this be deployed to production hubs.mozilla.com? I am planning on hosting a party on Hubs next week and having echo cancellation working on chromium would be amazing for that.

gfodor:
We push master, so this is live

oveddan:
It seems like with this fix, there is no stereo on chrome, and as a result no positional audio on chrome.  Is that correct?

PR #2365, opened 2020-04-10
Update static-media and static-controlled-media collision behavior

InfiniteLee:
Updates static-media and static-controlled-media so that they only collide with interactables, and not the environment and avatar hands.
This prevents physics from being calculated between the environment mesh and `static-media` in scenes that export both.


johnshaughnessy:
:+1:

Issue #2204, opened 2020-03-10
Close Button not labeled for screen readers in information message


emclaren:
**Description**
Button not labelled in information message pop up

![image(7)](https://user-images.githubusercontent.com/6835324/76282042-ced38e00-6253-11ea-925f-3d5659e1b626.png)

**Additional context**
Flagged by the a11y meetup group using Hubs

Issue #2203, opened 2020-03-10
Labels for Object List and People List dropdowns are insufficient for screen reader


emclaren:
**Description**
Labels for Object List and People List are insufficient for screen reader user
Right now screen reader only hears "6" and "3" to describe the UI elements. We should prepend better labels in front of each item such as "Objects 6" and "People 3"

![image(5)](https://user-images.githubusercontent.com/6835324/76281725-2c1b0f80-6253-11ea-90fe-94d66a76ace2.png)

**Additional context**
Flagged by the a11y community using Hubs

Issue #2202, opened 2020-03-10
Object List items should have a button or link role for screen readers


emclaren:
**Description**
The items in the object list should read as a button or link role type because they have the functionality that they are interactive. 

![image(4)](https://user-images.githubusercontent.com/6835324/76281502-7bad0b80-6252-11ea-9fc4-c7d803177781.png)

**Additional context**
This was flagged by the a11y community using Hubs

Issue #2201, opened 2020-03-10
Object Menu Dialog need better aria-label information for screen readers


emclaren:
**Description**
All the unlabeled controls should get simple text labels like "Close", "Previous Object", "Next Object" etc 

![image(3)](https://user-images.githubusercontent.com/6835324/76281412-312b8f00-6252-11ea-9ce8-6124a55c6db6.png)

**Additional context**
Flagged by the a11y meetup group using Hubs

PR #2374, opened 2020-04-15
Object list accessibility

johnshaughnessy:
- Adds `aria-label` attributes to icon-based `button`s
- Converts object list row-item `div`s into `button`s to improve accessibility for screen readers and keyboard use. 

Closes https://github.com/mozilla/hubs/issues/2203 
Closes https://github.com/mozilla/hubs/issues/2204
Closes https://github.com/mozilla/hubs/issues/2202
Closes https://github.com/mozilla/hubs/issues/2201


src/react-components/object-list.js
```diff
@@ -112,9 +113,10 @@ export default class ObjectList extends Component {
 
   domForEntity(el, i) {
     return (
-      <div
+      <button
+        aria-label="Show Object Info Modal"
```
brianpeiris, line 117: I would probably use "Panel" instead of "Modal". I don't know if the technical difference matters here, and some users might not be familiar with the concept of a modal.

src/assets/stylesheets/object-list-styles.scss
```diff
@@ -0,0 +1,39 @@
...
+  }
+}
+
+:local(.icon) {
+  svg {
+    vertical-align: 0px !important;
+    width: 20px !important;
+    height: 20px !important;
+  }
+}
```
brianpeiris, line 39: Maybe add a right margin on the icons? The longer object names are a bit too close:
image

src/react-components/ui-root.js
```diff
@@ -1067,13 +1067,18 @@ class UIRoot extends Component {
           ) : (
             <span>{this.props.hub.name}</span>
           )}
-          <button onClick={() => this.setState({ watching: true })} className={entryStyles.collapseButton}>
+          <button
+            aria-label="Close room entry modal"
```
brianpeiris, line 1071: Would this be more descriptive as "Close room entry panel and enter lobby"?

src/react-components/ui-root.js
```diff
@@ -2136,6 +2141,7 @@ class UIRoot extends Component {
 
                 {!streaming && (
                   <button
+                    aria-label="Favorite"
```
brianpeiris, line 2144: Should this be "Toggle Favorited" as well?

johnshaughnessy: Yes

src/react-components/ui-root.js
johnshaughnessy: Yes

src/react-components/object-list.js
johnshaughnessy: Good call!

src/assets/stylesheets/object-list-styles.scss
johnshaughnessy: Nice catch, will fix. I also want to revisit this list at some point because I don't think the text labels are adding enough value and suspect an "icons-only" or "generated-thumbnails" view could be better.

Issue #2241, opened 2020-03-18
Color Contrast on Landing Page


mgifford:
**Description**
There isn't sufficient color contrast footer links
https://contrast-ratio.com/#rgb%28128%2C%20128%2C%20128%29-on-white

**To Reproduce**
1. Go to https://hubs.mozilla.com/#/
2. Run WAVE Toolbar
3. See error

<img width="978" alt="Screen Shot of home page" src="https://user-images.githubusercontent.com/116832/76965275-67c07400-68fa-11ea-8aa0-dab09c11bbf8.png">

**Expected behavior**
It should have sufficient contrast.

harshil1712:
I would like to work on this. 
How about using #676767 9https://contrast-ratio.com/#%23676767-on-white), it solves the issues and is a shade of grey.

brianpeiris:
Sounds good to me @harshil1712

brianpeiris:
Closed in #2386. Thanks @harshil1712!

PR #2386, opened 2020-04-17
changed footer link color fixing #2241

harshil1712:
Changed the footer links color to #676767.


brianpeiris:
This seems like an acceptable solution to me. Looking at the list of existing colors in `colors.scss` and the rest of the vars in `root-vars.scss`, I'm not sure if it makes sense to make this another core color, and it doesn't seem to make sense to change `dark-grey` to #676767 since that variable is used in varying cases, sometimes with darker backgrounds.

I'll let @gfodor comment as well, since he's more familiar with how we setup our color themes and variables.

PR #2400, opened 2020-04-21
instruct to run the package script 'login' rather than the npm command

wmurphyrd:
Adds the login command to the instructions and clarifies the command to run in the error message


wmurphyrd:
actually this only got me to `Error fetching Hubs Cloud config "Internal Server Error"`, when trying to start after logging in with hubs.mozilla.com. Is it no longer possible to run dev instances off of the Hubs network servers?

Issue #1893, opened 2019-11-12
Avatars stay in room after user leaves


emclaren:
**Description**
Sometimes when users leave a Hubs room, their avatar stays behind. If the user re-joins the room, their old avatar is not visible to them, however, for everyone else in the space they see two of that user

**To Reproduce**
Have not figured out how to reproduce, but occurs fairly often

**Expected behavior**
Expect avatar to leave when user leaves

**Screenshots**
This user left the room but their avatar stayed. There were two when they returned.
<img width="647" alt="Screen Shot 2019-11-08 at 11 55 17 AM" src="https://user-images.githubusercontent.com/6835324/68681969-37fefb00-0519-11ea-9c03-f97f64fad280.png">

**Hardware** 
It happened once with one user when they refreshed the page using Chrome browser on PC, not sure if it has happened on other systems/devices

**Additional context**
We've seen this issue at Hubs meetups, and it was also flagged by community members.

misslivirose:
Just saw this today when @robertlong 's browser crashed. The original avatar did "leave" shortly afterwards ('Crested-Shelduck' is the original display name that Robert was using)  

![image](https://user-images.githubusercontent.com/4313320/68798315-0b2b1080-060b-11ea-8692-1bed0f3d0e22.png)

gfodor:
I did some preliminary investigation into this and could not reproduce. *However*, it is a known "by-design" issue where if someone's client terminates unexpectedly, the various connection timeouts need to be triggered before their avatar will disappear. (30 seconds) -- so we should confirm when we see this happening if there are instances where the avatar stays permanently visible.

emclaren:
I can confirm that many of the times I've seen this happen it has been longer than 30 seconds. I often end up kicking them from the room manually because the avatar seems to be permanently visible.

gfodor:
OK thanks. Not sure what that is then. Hmm. If we have a solid way to repro
that'd be the main thing now. I did a bunch of attempts to repro by
shutting computers off, pulling network cables, etc. All the avatars got
cleaned up after 30s.

emclaren:
I have also tried to repro this by crashing tabs, closing browsers with no success. I will make a point of trying to gather more information for room visitors when this happens (browser/OS/circumstance)

gfodor:
I think I finally have a repro for this - it seems to happen with bots if you open a second browser window and then alt-tab to it while the bots disconnect. If you alt tab back, the avatars are stuck.

[...]

PR #2402, opened 2020-04-22
Use Reticulum presence list to determine which users to subscribe to on Janus

InfiniteLee:
This update is the Hubs-side change needed to support https://github.com/mozilla/naf-janus-adapter/pull/100. This allows Hubs to use the presence list to control what users to connect to on Janus.

This partially addresses https://github.com/mozilla/hubs/issues/1893, but there seem to be additional cases where stuck avatars can occur.


src/hub.js
```diff
@@ -1282,7 +1284,8 @@ document.addEventListener("DOMContentLoaded", async () => {
             entryDisallowed: !hubChannel.canEnterRoom(uiProps.hub)
           });
 
-          const occupantCount = Object.entries(presence.state).length;
+          const userIds = Object.getOwnPropertyNames(presence.state);
```
gfodor, line 1287: i'd refer to these as `sessionIds`, since in phoenix presence context, they are the `SessionSocket` identifiers.

src/hub.js
```diff
@@ -1273,6 +1273,8 @@ document.addEventListener("DOMContentLoaded", async () => {
       if (isInitialJoin) {
         store.addEventListener("profilechanged", hubChannel.sendProfileUpdate.bind(hubChannel));
 
+        const requestedOccupants = [];
```
gfodor, line 1276: is this just to avoid allocations? on the fence if we should care here given the rarity of this event, vs just doing a `.filter.map` on the `presence.state`

InfiniteLee: presence.state is an object, so not sure if that makes this any easier to do

PR #2410, opened 2020-04-23
implement useThisCamera preference

### LINKED ISSUE #2406: Hubs client does not allow video camera selection
**Description**
Hubs client does not allow video camera selection.
This may be due to provided video constraints such as facingMode: user,
which not all cameras support.

**To Reproduce**
On laptop, configure Chrome to use a different camera by default than the first in the list 
(e.g. one that will not report facingMode as user).
On mobile, imagine that you want to share your environment camera.
Use Hubs, and share your camera.
Notice that the wrong camera is used.

**Expected behavior**
If Hubs is not going to offer a camera selector, it should use the default camera set in browser (laptop).  Since this does not work for mobile, I would not recommend this option.
If it is going to specifically select a camera device, it should provide a list for the user to override selection, or a control to toggle through cameras (catching errors so that a broken camera does not stop the selection cycle).

**Screenshots**
N/A

**Hardware** 
 - Device: Laptop, also Samsung S20Ultra
 - OS: Win10, Android 10
 - Browser: Chrome, Samsung Internet

**Additional context**

machenmusik:
Fixes #2406 by implementing a user preference
- default: same behavior as now
- user: ask for user-facing camera
- environment: ask for environment-facing camera
- specific deviceId: pick a specific camera (the preference list has friendly names for them)

### DIFF
--- a/src/assets/translations.data.json
+++ b/src/assets/translations.data.json
@@ -261,6 +261,7 @@
     "commands.capture": "Capture a 15 second video.",
     "commands.debug": "Toggle physics debug rendering.",
     "commands.vrstats": "Toggle stats in VR.",
+    "preferences.useThisCamera": "Use this camera",
     "preferences.muteMicOnEntry": "Mute microphone on entry",
     "preferences.enableOnScreenJoystickLeft": "Enable left on-screen joystick for moving around",
     "preferences.enableOnScreenJoystickRight": "Enable right on-screen joystick for looking around",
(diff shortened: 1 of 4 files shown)


machenmusik:
(This was built against the `hubs-cloud` branch originally.)

netpro2k:
Thanks for putting this together, made some minor tweaks but looks good! Rolling this out now.

machenmusik:
Thanks for doing the finishing touches!

PR #2411, opened 2020-04-23
External Landing Page and Admin Projects

robertlong:
This PR is to be merged with the rest of the PRs in Reticulum, Ita, and elsewhere that makes the landing page and admin projects deployable as separate ita targets.

To Do:
- [ ] Move landing page and other HMC pages to a new HMC landing page project


robertlong:
Closing this, that's not how we're going to do this.

PR #2422, opened 2020-04-25
Fix bounding box generation for scaled objects (many avatar previews)

netpro2k:
Patches `updateWorldMatrix()` (which is different from `updateMatrixWorld()`) to be aware of our matrix update flags. An update in three started using this function in `Box3.expandToObject` which caused some avatar previews to start rendering incorrectly. It likely manifested in other ways as well like broken frustum culling and incorrect physics shape generation for some objects.

### DIFF
--- a/src/utils/threejs-world-update.js
+++ b/src/utils/threejs-world-update.js
@@ -92,6 +92,19 @@ THREE.Object3D.prototype.applyMatrix = function() {
   handleMatrixModification(this);
 };
 
+// Updates this function to use updateMatrices(). In general our code should prefer calling updateMatrices() directly,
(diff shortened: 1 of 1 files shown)


src/utils/threejs-world-update.js
```diff
@@ -92,6 +92,19 @@ THREE.Object3D.prototype.applyMatrix = function() {
   handleMatrixModification(this);
 };
 
+// Updates this function to use updateMatrices(). In general our code should prefer calling updateMatrices() directly,
+// patching this for compatability upstream, namely with Box3.expandToObject and Object3D.attach
```
vincentfretin, line 96: typo compatability => compatibility

PR #2427, opened 2020-04-27
Remove duplicate translation `tips.mobile.object_pin`

antpb:
Was digging around in my editor and got a warning about this duplicate translation. Might get confusing for folks so here's a PR that removes it. :)


johnshaughnessy:
Nice catch! I'll deploy this next time I update the client

PR #2435, opened 2020-04-28
Improve Webpack Code Splitting

robertlong:
I rewrote how we handle splitting chunks in webpack.config.js to reduce the bundle size of all of our pages, particularly the pages that don't need Three.js or AFrame.

Before:
Screen Shot 2020-04-22 at 10 18 06 PM
After:
Screen Shot 2020-04-28 at 2 21 36 PM


arpu:
maybe interessting in this https://web.dev/granular-chunking-nextjs/  granular chunking

robertlong:
That is really interesting @arpu, maybe next iteration I'll use more chunks. I kinda figured you'd want fewer because you'd have fewer requests, but with HTTP/2 that's less important. Thanks for sharing that!

PR #2438, opened 2020-04-29
Add mobile-friendly object info panel

johnshaughnessy:
The new mobile-friendly object info panel makes better use of limited screen space and adds better support for touch-based interaction. 

image

image


src/react-components/ui-root.js
```diff
@@ -1912,93 +1922,91 @@ class UIRoot extends Component {
...
-                          [inviteStyles.hideSmallScreens]: this.occupantCount() > 1 && entered,
-                          [inviteStyles.inviteButtonLowered]: hasTopTip
-                        })}
-                        onClick={() => this.toggleShareDialog()}
-                      >
-                        <FormattedMessage id="entry.share-button" />
-                      </button>
-                    )}
-                  {showChooseSceneButton && (
+            {showInviteButton && (
```
johnshaughnessy, line 1925: The diffs below are formatting changes as a result of using the `showFoo` constants defined above.

johnshaughnessy:
Notes from reviewing this with Dom. May not get to these right away, but want to capture them here for later:

// TODO: handle resize event
// TODO: Remove setTimeout for listed-media
// TODO: format Message
// TODO: Call onItemSelected in wheel
// TODO: renderSmallScreen could be just a SmallScreen component
// TODO: avoid re-creating so many `onClick` functions every render if perf is an issue

Issue #2298, opened 2020-03-30
Scale not sticking


gfodor:
It looks like if you scale an object in a room with multiple people (using mouse or scale button) the object can be reset to the previous scale. Perhaps due to ownership transfer failing.

johnshaughnessy:
#2442 is a partial fix for this. (I forgot to take ownership of the object when the scale button was clicked.)

However, there is still an issue where scale will not stick if object ownership is under contention, which can come as a surprise to users especially because network updates are paused during freeze mode. ( We stop processing network updates while the scene is frozen. ) One thing we could do is to repeatedly call `takeOwnership` so that whoever was the most recent person to be manipulating the object "wins", even if both parties are in freeze mode at the beginning of the interaction. That way the "loser" is more likely to at least see that the object is being manipulated by the ownership "winner". This would require changing `takeOwnership` so that it updates ownership info (like `lastOwnerTime`) even if the object `isMine()` already. I don't recommend we do this because it complicates the semantics quite a bit including for things like taking an object out of someone's hand.

johnshaughnessy:
I suggest that after deploying #2442 , we lower the priority to P2 or P3 or even replace it with a generic "ownership contention" issue (which I think is the cause of scale not sticking when using the mouse wheel).

joshmarinacci:
bump. This is likely fixed by now, right?

PR #2442, opened 2020-04-30
Take ownership of the object you are scaling

johnshaughnessy:
Partial fix for https://github.com/mozilla/hubs/issues/2298 for the scale button.

### DIFF
--- a/src/components/scale-button.js
+++ b/src/components/scale-button.js
@@ -64,6 +64,7 @@ AFRAME.registerComponent("scale-button", {
     this.dragVector = new THREE.Vector3();
     this.currentObjectScale = new THREE.Vector3();
     NAF.utils.getNetworkedEntity(this.el).then(networkedEl => {
+      this.networkedEl = networkedEl;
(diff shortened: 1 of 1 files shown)


src/components/scale-button.js
```diff
@@ -73,6 +74,11 @@ AFRAME.registerComponent("scale-button", {
       if (this.isScaling || !this.objectToScale) {
         return;
       }
+
+      if (!(NAF.utils.isMine(this.networkedEl) || NAF.utils.takeOwnership(this.networkedEl))) {
```
InfiniteLee, line 78: might need to check for this.networkedEl in case the getNetworkedEntity promise hasn't resolved yet

johnshaughnessy: The check above for `this.objectToScale` will early out in this case, since they're both given values in the `getNetworkedEntity` callback https://github.com/mozilla/hubs/pull/2442/files#diff-425a09cd704e35df9abb0e6b291f4685R74
I could add another check for `networkedEl` though if that reads clearer.

Issue #2436, opened 2020-04-28
Ability to turn off system audio (e.g. chat sounds, object interactions)


misslivirose:

johnshaughnessy:
@misslivirose Is #2443 what you have in mind? It adds a pref that will disable sound effects (including chat sounds and object interactions), but audio tracks from other people and from media in the room are left unaffected.

misslivirose:
Yes!

PR #2443, opened 2020-04-30
Add preference to disable sound effects

johnshaughnessy:
Closes https://github.com/mozilla/hubs/issues/2436

### DIFF
--- a/src/systems/sound-effects-system.js
+++ b/src/systems/sound-effects-system.js
@@ -107,9 +107,21 @@ export class SoundEffectsSystem {
         this.sounds.set(sound, audioBuffer);
       });
     });
+
+    this.isDisabled = window.APP.store.state.preferences.disableSoundEffects;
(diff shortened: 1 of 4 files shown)


src/systems/sound-effects-system.js
```diff
@@ -203,7 +217,24 @@ export class SoundEffectsSystem {
     );
   }
 
+  stopAllPositionalAudios() {
```
InfiniteLee, line 220: Does this get used?

johnshaughnessy: Yes, when the pref flips on: https://github.com/mozilla/hubs/pull/2443/files#diff-42eb0478cd8f412d222db2e8229724daR115

InfiniteLee: Oh wow sorry, somehow when I did a search just now it didnt find anything, bizarre

PR #2444, opened 2020-04-30
Properly switch ret server for hubs cloud

gfodor:
It seems that npm run start was not talking to my hubs cloud server. Kind of surprised this was missed given the refactoring was focused on enabling this, so if there was a gap in testing this reveals we should chase down any other potential issues, or maybe I'm doing something wrong.


robertlong:
Closing this since we merged another PR that fixed it.

PR #2448, opened 2020-04-30
Basic support for DASH video streams

netpro2k:
You can now play back DASH video streams in Hubs! This will automatically be used in some cases on video providers like Vimeo, but also provides additional options for those hosting their own video streaming setups. This is just the most basic integration so do report any issues you run into.

Just the bare minimum needed to support dash video streams. Seems to work fine with both VOD and live content as is. It's likely we may find some things that need more configuration/tuning but seems like the quickest way to find them is just to add support and find out what breaks. 

What prompted this is that Vimeo started resolving to DASH videos sometimes. Once we are convinced this works well we can update our youtube-dl queries to explicitly look for DASH (and maybe prefer it over HLS if we decide its better). 

Not sure at what stage we should mention it in our FAQ about supported video/streaming formats.

### DIFF
--- a/src/components/media-loader.js
+++ b/src/components/media-loader.js
@@ -396,6 +396,11 @@ AFRAME.registerComponent("media-loader", {
       // we don't think we can infer it from the extension, we need to make a HEAD request to find it out
       contentType = contentType || guessContentType(canonicalUrl) || (await fetchContentType(accessibleUrl));
 
+      // TODO we should probably just never return "application/octet-stream" as expectedContentType, since its not really useful
+      if (contentType === "application/octet-stream") {
+        contentType = guessContentType(canonicalUrl) || contentType;
+      }
+
       // Some servers treat m3u8 playlists as "audio/x-mpegurl", we always want to treat them as HLS videos
       if (contentType === "audio/x-mpegurl") {
         contentType = "application/vnd.apple.mpegurl";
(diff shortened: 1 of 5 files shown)


src/components/media-loader.js
```diff
@@ -396,6 +396,11 @@ AFRAME.registerComponent("media-loader", {
       // we don't think we can infer it from the extension, we need to make a HEAD request to find it out
       contentType = contentType || guessContentType(canonicalUrl) || (await fetchContentType(accessibleUrl));
 
+      // TODO we should probably just never return "application/octet-stream" as expectedContentType, since its not really useful
+      if (contentType === "application/octet-stream") {
+        contentType = guessContentType(canonicalUrl) || contentType;
```
johnshaughnessy, line 401: Not sure what the expectation is for getting to this line. If we got `"application/octet-stream"` and then `guessContentType(canonicalUrl)` failed, should we `await fetchContentType(accessibleUrl)` like line https://github.com/mozilla/hubs/pull/2448/files#diff-50f3f477609a941f9eafc236322b8160R397 above?

src/components/media-views.js
```diff
@@ -687,6 +698,25 @@ AFRAME.registerComponent("media-video", {
...
+        dashPlayer.addUTCTimingSource(
+          "urn:mpeg:dash:utc:http-xsdate:2014",
+          proxiedUrlFor("https://time.akamai.com/?iso")
+        );
+        // We can also use our own HEAD request method like we use to sync NAF
+        // dashPlayer.addUTCTimingSource("urn:mpeg:dash:utc:http-head:2014", location.href);
+
+        texture.dash = dashPlayer;
+
+        // window.d = window.d || [];
```
johnshaughnessy, line 718: comment

src/components/media-loader.js
netpro2k: As noted in the TODO, best way to handle this would be to have the server just not return `application/octet-stream`, until we do that I just wanted to make the least risky change possible, and didn't want to accidentally trigger a second head request nor muck with the logic too much, so just falling back to guessing content type for known filenames as a last resort.

Issue #2229, opened 2020-03-16
Apply VRDisplay pose data to the camera at the beginning of the frame.


johnshaughnessy:
We rely on `THREE.VRManager` to query the `VRDisplay` for pose data and update the transform with a camera attached to it. The problem we have is that this happens just before render (after everything else in our frame), whereas we want it to happen near the beginning of the frame. Everything that refers to the camera transform in our update loop is a frame behind.

johnshaughnessy:
See also https://github.com/mozilla/hubs/pull/2226

PR #2450, opened 2020-04-30
Fix WebXR support for Oculus Browser

brianpeiris:
Oculus Browser has removed support for WebVR. This PR fixes our WebXR support and enables it for Oculus Browser specifically.
- Three.js' WebXRManager updates the scene's camera directly, instead of using the previous concept of a "poseTarget", so we need to grab headset transforms directly from the camera instead of the parent object3D
- Added a generic WebXRControllerDevice for WebXR gampads. The new webxr gamepad spec has different button and axis mappings, so we can't use existing controller devices, but the spec should now be much more standardized so the bindings should work for most devices when we expand webxr support. The bindings are a copy of our existing Oculus Touch bindings, with keyboard bindings removed and some cleanup. See https://immersive-web.github.io/webxr-gamepads-module/#xr-standard-gamepad-mapping
- Forced a max distance of 1000m on the skybox. Not sure why the camera's far field is different when entering webxr immersive mode. Perhaps something to investigate later.
- Updated anime.js to use `setTimeout` for animations instead of `window.requestAnimationFrame` since the global rAF is paused when in webxr immersive mode.
- Adds a `no_force_webvr` flag so that we can test webxr in other browsers.

Goes with:
- https://github.com/MozillaReality/aframe/pull/27
- https://github.com/MozillaReality/anime/pull/1

### DIFF
--- a/src/components/audio-feedback.js
+++ b/src/components/audio-feedback.js
@@ -169,7 +169,7 @@ AFRAME.registerComponent("scale-audio-feedback", {
 
   async init() {
     await waitForDOMContentLoaded();
-    this.camera = document.getElementById("viewing-camera").object3D;
+    this.camera = document.getElementById("viewing-camera").object3DMap.camera;
   },
 
   tick() {
--- a/src/systems/camera-system.js
+++ b/src/systems/camera-system.js
@@ -314,6 +314,10 @@ export class CameraSystem {
     const translation = new THREE.Matrix4();
     let uiRoot;
     return function tick(scene, dt) {
+      this.viewingCamera.object3DMap.camera.matrixNeedsUpdate = true;
+      this.viewingCamera.object3DMap.camera.updateMatrix();
+      this.viewingCamera.object3DMap.camera.updateMatrixWorld();
+
       const entered = scene.is("entered");
       uiRoot = uiRoot || document.getElementById("ui-root");
       const isGhost = !entered && uiRoot && uiRoot.firstChild && uiRoot.firstChild.classList.contains("isGhost");
@@ -322,7 +326,6 @@ export class CameraSystem {
         const position = new THREE.Vector3();
         const quat = new THREE.Quaternion();
         const scale = new THREE.Vector3();
-        this.viewingCamera.object3DMap.camera.updateMatrices();
         this.viewingRig.object3D.updateMatrices();
         this.viewingRig.object3D.matrixWorld.decompose(position, quat, scale);
         position.setFromMatrixPosition(this.viewingCamera.object3DMap.camera.matrixWorld);
--- a/src/systems/camera-system.js
+++ b/src/systems/camera-system.js
@@ -211,11 +211,6 @@ export class CameraSystem {
     if (this.mode === CAMERA_MODE_SCENE_PREVIEW) return;
 
     this.mode = NEXT_MODES[this.mode] || 0;
-    if (this.mode === CAMERA_MODE_FIRST_PERSON) {
-      AFRAME.scenes[0].renderer.vr.setPoseTarget(this.avatarPOV.object3D);
-    } else if (this.mode === CAMERA_MODE_THIRD_PERSON_NEAR || this.mode === CAMERA_MODE_THIRD_PERSON_FAR) {
-      AFRAME.scenes[0].renderer.vr.setPoseTarget(this.viewingCamera.object3D);
-    }
   }
 
   inspect(o, distanceMod, temporarilyDisableRegularExit) {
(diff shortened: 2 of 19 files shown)


src/systems/waypoint-system.js
```diff
@@ -365,10 +365,14 @@ export class WaypointSystem {
...
         elementFromTemplate.object3D.visible = this.scene.is("frozen");
         if (elementFromTemplate.object3D.visible) {
-          this.viewingCamera = this.viewingCamera || document.getElementById("viewing-camera").object3D;
+          this.viewingCamera = this.viewingCamera || document.getElementById("viewing-camera").object3DMap.camera;
           setMatrixWorld(
             elementFromTemplate.object3D,
-            calculateIconTransform(waypointComponent.el.object3D, this.viewingCamera, this.helperMat4)
+            calculateIconTransform(
+              waypointComponent.el.object3D,
+              this.viewingCamera.object3DMap.camera,
```
johnshaughnessy, line 373: This `object3DMap.camera` seems redundant with the `object3DMap.camera` above

src/systems/userinput/devices/webxr-controller.js
```diff
@@ -0,0 +1,75 @@
+import { paths } from "../paths";
+import { Pose } from "../pose";
+
+const ONES = new THREE.Vector3(1, 1, 1);
+const LEFT_HAND_OFFSET = new THREE.Matrix4().makeTranslation(-0.025, -0.03, 0.12);
```
johnshaughnessy, line 5: Noting this for later -- these offsets should probably depend on the specific controller being used.

src/components/audio-feedback.js
```diff
@@ -169,7 +169,7 @@ AFRAME.registerComponent("scale-audio-feedback", {
 
   async init() {
     await waitForDOMContentLoaded();
-    this.camera = document.getElementById("viewing-camera").object3D;
```
johnshaughnessy, line 172: What was the motivation / reason for changing all these spots to use the camera object in the `object3DMap`?

johnshaughnessy: Oops, nevermind I see you explained it in the summary!

johnshaughnessy:
> Updated anime.js to use setTimeout for animations instead of window.requestAnimationFrame since the global rAF is paused when in webxr immersive mode.

This is alarming, because presumably the `timeout`s are more expensive than using a raf loop. What is the looping mechanism that replaces `window.requestAnimationFrame` for immersive webxr?

johnshaughnessy:
> Three.js' WebXRManager updates the scene's camera directly, instead of using the previous concept of a "poseTarget", so we need to grab headset transforms directly from the camera instead of the parent object3D

Note for later: I'd like for us to control this ourselves because I think want to get the updated camera transform at the _beginning_ of the frame https://github.com/mozilla/hubs/issues/2229

src/systems/camera-system.js
```diff
@@ -393,11 +388,11 @@ export class CameraSystem {
         this.avatarRig.object3D.updateMatrices();
         setMatrixWorld(this.viewingRig.object3D, this.avatarRig.object3D.matrixWorld);
         if (scene.is("vr-mode")) {
-          this.viewingCamera.object3D.updateMatrices();
-          setMatrixWorld(this.avatarPOV.object3D, this.viewingCamera.object3D.matrixWorld);
+          this.viewingCamera.object3DMap.camera.updateMatrices();
+          setMatrixWorld(this.avatarPOV.object3D, this.viewingCamera.object3DMap.camera.matrixWorld);
         } else {
           this.avatarPOV.object3D.updateMatrices();
-          setMatrixWorld(this.viewingCamera.object3D, this.avatarPOV.object3D.matrixWorld);
```
johnshaughnessy, line 400: I forget why there was this complexity around having the `viewingCamera` follow the `avatarPOV` node or vice versa, but I thought it was related to the `setPoseTarget` stuff and how it interacted with the-mode-formerly-known-as-inspect (aka focus mode). I'd just want to double check that focus and object list navigation works in 2D, touchscreen, and in VR.

src/systems/userinput/devices/webxr-controller.js
brianpeiris: Ah, good point. I'll add a comment to the code.

src/systems/waypoint-system.js
brianpeiris: :+1: nice catch. I'll make sure to test the features I've touched as well.

brianpeiris:
> This is alarming, because presumably the `timeout`s are more expensive than using a raf loop. What is the looping mechanism that replaces `window.requestAnimationFrame` for immersive webxr?

WebXR provides a requestAnimationFrame on the XRSession itself, which three.js and aframe use, but it would be hard to get anime.js to use it without some refactoring of the library. My understanding is that setTimeout isn't any more expensive -- it just uses the usual javascript event loop. Though I will try adding fixed timeout duration to the setTimeout, instead of the implicit 0ms timeout I'm using right now.

PR #2451, opened 2020-05-01
Updated Admin styles so black background expands to bottom on small windows. 

robin-k-wilson:
Before:
image

After:
image


robin-k-wilson:
Already integrated with another PR

Issue #2398, opened 2020-04-21
Videos always spatialized


j-conrad:
When setting a video audio to be **stereo**, the audio still seems to be spatialized.

Here is a Hubs scene you can try to reproduce the bug:
https://hubs.mozilla.com/scenes/SpDVYWJ

The scene is remixable so you can confirm how it's set in Spoke.

Just move around the small floor and you'll hear the audio pan from ear to ear.

j-conrad:
Edited this bug to be ALL video audio. (It used to specify 360 video, but I've since confirmed it's all video.)

PR #2456, opened 2020-05-02
Audio bug fixes and refactor

johnshaughnessy:
This PR:
- Fixes bug causing `PositionalAudio`s to get created instead of `Audio`s, even if the `media-video`'s `audioType` was not set to `"pannernode"`. Closes https://github.com/mozilla/hubs/issues/2398 
- Fixes a bug where `audio-source` component attempts to access the `networked-audio-source`'s `sound` before it exists. Closes https://github.com/mozilla/hubs/issues/2466
- Uses the `update` method and `statechanged` events when interfacing with the preference `store`. 
- Updates `media-video` such that it will recreate their `Audio` or `PositionalAudio`s when necessary. 
- Adds `avatar-audio-source`, which is a re-write of `networked-audio-source` that will recreate its `Audio` or `PositionalAudio` when its `positional` component data changes. (It also uses `async` and `await` as a quality of life improvement, which we can do because it's not in the NAF repo which doesn't use those language features.)
- Updates `avatar-volume-controls` so that they continue to work if an `Audio` or `PositionalAudio`s get recreated.

### DIFF
--- a/src/components/avatar-audio-source.js
+++ b/src/components/avatar-audio-source.js
@@ -58,6 +58,7 @@ AFRAME.registerComponent("avatar-audio-source", {
   createAudio: async function() {
     this.isCreatingAudio = true;
     const stream = await getMediaStream(this.el);
+    this.isCreatingAudio = false;
     const isRemoved = !this.el.parentNode;
     if (!stream || isRemoved) return;
 
@@ -75,7 +76,6 @@ AFRAME.registerComponent("avatar-audio-source", {
     audio.setNodeSource(mediaStreamSource);
     this.el.setObject3D(this.attrName, audio);
     this.el.emit("sound-source-set", { soundSource: mediaStreamSource });
-    this.isCreatingAudio = false;
   },
 
   destroyAudio() {
--- a/src/components/avatar-audio-source.js
+++ b/src/components/avatar-audio-source.js
@@ -1,4 +1,4 @@
-const INFO_INIT_FAILED = "Failed to initialize networked-audio-source.";
+const INFO_INIT_FAILED = "Failed to initialize avatar-audio-source.";
 const INFO_NO_NETWORKED_EL = "Could not find networked el.";
 const INFO_NO_OWNER = "Networked component has no owner.";
 
@@ -50,7 +50,7 @@ async function getMediaStream(el) {
   return stream;
 }
 
-AFRAME.registerComponent("networked-audio-source", {
+AFRAME.registerComponent("avatar-audio-source", {
   schema: {
     positional: { default: true },
     distanceModel: {
--- a/src/components/avatar-audio-source.js
+++ b/src/components/avatar-audio-source.js
@@ -15,20 +15,6 @@ function createSilentAudioEl(stream) {
   return audioEl;
 }
 
-function getOrCreateAudioListenerForScene(scene) {
-  if (!scene.audioListener) {
-    // Would prefer not to have to do this, and am not sure if it's required,
-    // so adding a warning for now and hoping to remove this later.
-    console.warn("Scene did not have an audio listener. Creating one now.");
-    scene.audioListener = new THREE.AudioListener();
-    scene.camera && scene.camera.add(scene.audioListener);
-    scene.addEventListener("camera-set-active", function(evt) {
-      evt.detail.cameraEl.getObject3D("camera").add(scene.audioListener);
-    });
-  }
-  return scene.audioListener;
-}
-
 async function getMediaStream(el) {
   const networkedEl = await NAF.utils.getNetworkedEntity(el).catch(e => {
     console.error(INFO_INIT_FAILED, INFO_NO_NETWORKED_EL, e);
@@ -50,6 +36,13 @@ async function getMediaStream(el) {
   return stream;
 }
 
+function setPositionalAudioProperties(audio, settings) {
+  audio.setDistanceModel(settings.distanceModel);
+  audio.setMaxDistance(settings.maxDistance);
+  audio.setRefDistance(settings.refDistance);
+  audio.setRolloffFactor(settings.rolloffFactor);
+}
+
 AFRAME.registerComponent("avatar-audio-source", {
   schema: {
     positional: { default: true },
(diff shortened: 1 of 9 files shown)


src/systems/audio-settings-system.js
```diff
@@ -99,23 +102,37 @@ AFRAME.registerComponent("audio-source", {
...
-        networkedAudioSource.data.positional = audioOutputMode === "panner" ? true : false;
-        networkedAudioSource.sound.disconnect();
-        networkedAudioSource.setupSound();
-        const soundSource = networkedAudioSource.sound.context.createMediaStreamSource(networkedAudioSource.stream);
-        networkedAudioSource.sound.setNodeSource(soundSource);
-        networkedAudioSource.el.emit("sound-source-set", { soundSource });
-      }
+    if (networkedAudioSource && this.networkedAudioSource !== networkedAudioSource) {
+      this.networkedAudioSource = networkedAudioSource;
+      this.originalValueOfPositional = networkedAudioSource.data.positional;
```
johnshaughnessy, line 134: This line seems wrong to me. What is the correct way to determine whether a networked audio source should be `positional`?

src/components/avatar-volume-controls.js
```diff
@@ -41,20 +41,28 @@ AFRAME.registerComponent("avatar-volume-controls", {
...
+      }
+
+      const { audioOutputMode, globalVoiceVolume } = window.APP.store.state.preferences;
+      const volumeModifier = (globalVoiceVolume !== undefined ? globalVoiceVolume : 100) / 100;
+      let gain = volumeModifier * this.data.volume;
+      if (audioOutputMode === "audio") {
+        this.avatarAudioSource.el.object3D.getWorldPosition(positionA);
+        this.el.sceneEl.camera.getWorldPosition(positionB);
+        const squaredDistance = positionA.distanceToSquared(positionB);
+        gain = gain * Math.min(1, 10 / Math.max(1, squaredDistance));
```
InfiniteLee, line 60: This shouldn't be in `update` - it was intentionally originally added to `tick` otherwise the volume attenuation would only happen `update` happens to be called (very noticeably less frequently).

InfiniteLee: Oh nvm I see you're calling update on every `tick` now

src/components/avatar-audio-source.js
```diff
@@ -0,0 +1,115 @@
+const INFO_INIT_FAILED = "Failed to initialize avatar-audio-source.";
```
johnshaughnessy, line 1: For reference, here is `networked-audio-source` : https://github.com/MozillaReality/networked-aframe/blob/8d0dc7355c1c72ba0bb49a20e5920da0a0a92719/src/components/networked-audio-source.js#L1

I also put a diff up in this gist, but they're pretty different so it's probably easier to just read this new version. https://gist.github.com/johnshaughnessy/33517261b58f36d393ca740829f40e1e/revisions

src/components/avatar-audio-source.js
```diff
@@ -0,0 +1,115 @@
...
+function createSilentAudioEl(stream) {
+  const audioEl = new Audio();
+  audioEl.setAttribute("autoplay", "autoplay");
+  audioEl.setAttribute("playsinline", "playsinline");
+  audioEl.srcObject = stream;
+  audioEl.volume = 0; // we don't actually want to hear audio from this element
+  return audioEl;
+}
+
+function getOrCreateAudioListenerForScene(scene) {
```
InfiniteLee, line 18: You should be able to nuke this since `audio-system` should ensure the AudioListener is created.

src/components/avatar-audio-source.js
```diff
@@ -0,0 +1,115 @@
...
+      oneOf: ["linear", "inverse", "exponential"]
+    },
+    maxDistance: { default: 10000 },
+    refDistance: { default: 1 },
+    rolloffFactor: { default: 1 }
+  },
+
+  init: async function() {
+    this.el.sceneEl.systems["hubs-systems"].audioSettingsSystem.registerAvatarAudioSource(this);
+    const stream = await getMediaStream(this.el);
```
InfiniteLee, line 67: as I undestand it, while we can make the init async, it won't actually be treated asnyc by a-frame, that is to say, anything checking whether or not this component is initialized won't know that the `init` has actually finished or not. I don't know if that matters or not, but it seems like we should use a promise here instead of using async + await

src/components/avatar-audio-source.js
```diff
@@ -0,0 +1,115 @@
...
+    const isRemoved = !this.el.parentNode;
+    if (!stream || isRemoved) return;
+
+    const audioListener = getOrCreateAudioListenerForScene(this.el.sceneEl);
+    const audio = this.data.positional ? new THREE.PositionalAudio(audioListener) : new THREE.Audio(audioListener);
+    if (this.data.positional) {
+      audio.setDistanceModel(this.data.distanceModel);
+      audio.setMaxDistance(this.data.maxDistance);
+      audio.setRefDistance(this.data.refDistance);
+      audio.setRolloffFactor(this.data.rolloffFactor);
```
InfiniteLee, line 77: This should be abstracted into a function since it's performed down below as well

src/components/avatar-audio-source.js
```diff
@@ -0,0 +1,115 @@
...
+  },
+
+  update(oldData) {
+    const audio = this.el.getObject3D(this.attrName);
+    if (!audio) return;
+
+    const shouldRecreateAudio = oldData.positional !== this.data.positional;
+    if (shouldRecreateAudio) {
+      audio.disconnect();
+      this.el.removeObject3D(this.attrName);
```
InfiniteLee, line 97: The overlap between this and what `remove` does should also maybe be abstracted

src/components/avatar-audio-source.js
johnshaughnessy: I think it doesn't matter because I just want to _start_ the initialization (not wait on it), but I separated an `async createAudio` function that is called from `init` and `update` to make this as clear as possible.

src/components/avatar-audio-source.js
johnshaughnessy: :+1:

src/components/avatar-audio-source.js
johnshaughnessy: :+1:

src/components/avatar-audio-source.js
johnshaughnessy: :+1:

PR #2460, opened 2020-05-05
React Code Refactor: Part 1

robertlong:
Hubs has been using non-standard patterns for its ReactJS code. As we prepare for page extensibility, we need a good foundation to build our UI code on. This PR updates the landing page to make it more understandable for React devs.

It also splits the signin and verify dialogs out into separate pages. This is done so that swapping out the landing page with your own is simpler and so that the signin flow can be utilized from an external site. For example a sign in link in an app on myhubs.com links to hubs.myhubs.com/signin and then redirects back to myhubs.com when you are finished. This will require a corresponding Reticulum PR: https://github.com/mozilla/reticulum/pull/384


gfodor:
At 30k feet I get nervous about large, sweeping refactoring PRs touching so many files, esp since we do not have any integration testing of the website. I realize a lot of this is pretty mindless, but I'd advise breaking this up into smaller releases if possible or adding tests. Not a blocking request, but just raising the flag that this kind of thing often will result in some breakages, with minimal user upside, so usually ideal to take smaller steps.

robertlong:
Yeah, I did break it up a little bit. As soon as I can resolve these merge conflicts the changes should go down a bit.

src/react-components/home/HomePage.js
```diff
@@ -0,0 +1,100 @@
...
+
+export function HomePage() {
+  const featuredRooms = useFeaturedRooms();
+  const auth = useContext(AuthContext);
+
+  useEffect(() => {
+    const qs = new URLSearchParams(location.search);
+
+    // Support legacy sign in urls.
+    if (qs.has("sign_in")) {
```
netpro2k, line 28: What actually uses this sign in url?

src/react-components/home/HomePage.js
```diff
@@ -0,0 +1,100 @@
...
+      const redirectUrl = new URL("/signin", window.location);
+      redirectUrl.search = location.search;
+      window.location = redirectUrl;
+    } else if (qs.has("auth_topic")) {
+      const redirectUrl = new URL("/verify", window.location);
```
netpro2k, line 33: We will need to figure out the migration path for the emails we send out. Either we have to make a breaking change in reticulum (and make sure to notify everyone with a good lead up time so they can make sure their clients have been updated with the new endpoints) or just always support this version.

src/react-components/home/HomePage.js
```diff
@@ -0,0 +1,100 @@
...
+addLocaleData([...en]);
+
+export function HomePage() {
+  const featuredRooms = useFeaturedRooms();
+  const auth = useContext(AuthContext);
+
+  useEffect(() => {
+    const qs = new URLSearchParams(location.search);
+
+    // Support legacy sign in urls.
```
netpro2k, line 27: We need to make sure this is still handled when using a "default room", since it wont actually hit this page. Also thinking ahead a bit, we are probably going to need to move this up a level since we don't want "plugin" versions of the page to have to remember to support the sign in flow (that was why we wanted to split these off into their own pages in the first place)

src/react-components/home/useFeaturedRooms.js
```diff
@@ -0,0 +1,35 @@
...
+        const featuredRooms = entries
+          .filter((h, i) => ids.lastIndexOf(h.id) === i)
+          .sort((a, b) => b.member_count - a.member_count);
+
+        setFeaturedRooms(featuredRooms);
+      };
+
+      initAsync();
+    },
+    [auth.isSignedIn, auth.userId]
```
netpro2k, line 31: This seems to be the main argument I can see for the auth state/functions being a `useContext`

src/react-components/auth/AuthContext.js
```diff
@@ -0,0 +1,96 @@
...
+            retPhxChannel.leave();
+            resolve(perms);
+          });
+        });
+      });
+
+      configs.setIsAdmin(perms.postgrest_role === "ret_admin");
+      store.update({ credentials: { isAdmin: perms.postgrest_role === "ret_admin" } });
+    },
+    [store]
```
netpro2k, line 37: Just to confirm my understanding, we don't actually ever expect store to change correct?

src/react-components/auth/AuthContext.js
```diff
@@ -0,0 +1,96 @@
...
+  const signOut = useCallback(
+    async () => {
+      store.update({ credentials: { token: null, email: null, isAdmin: false } });
+      await store.resetToRandomDefaultAvatar();
+    },
+    [store]
+  );
+
+  const [context, setContext] = useState({
+    isSignedIn: !!store.state.credentials.token,
```
netpro2k, line 59: I think we talked about this before but since these are all just properties of `store` maybe they would just be part of a `useHubsStore` hook. Though I can see an argument for breaking up the different parts of store like credentials, preferences, and tutorial flow.

src/react-components/auth/AuthContext.js
```diff
@@ -0,0 +1,96 @@
...
+    isAdmin: store.state.credentials.isAdmin,
+    email: store.state.credentials.email,
+    userId: store.credentialsAccountId,
+    signIn,
```
netpro2k, line 63: Are these 3 functions just exposed on a context so that its easier to get at them? They seem to only rely on the store so they could also probably just be some sort of globals right?

src/react-components/auth/AuthContext.js
```diff
@@ -0,0 +1,96 @@
...
+import jwtDecode from "jwt-decode";
+import AuthChannel from "../../utils/auth-channel";
+import { connectToReticulum } from "../../utils/phoenix-utils";
+
+export const AuthContext = createContext();
+
+export function AuthContextProvider({ children, store }) {
+  const signIn = useCallback(
+    async email => {
+      const authChannel = new AuthChannel(store);
```
johnshaughnessy, line 15: Seems like we repeat this auth flow a few times throughout the code. I know this part is not new, and I would not suggest refactoring it in this (already large) PR, but it seems we can DRY some of this up.

src/react-components/auth/AuthContext.js
```diff
@@ -0,0 +1,96 @@
...
+  });
+
+  // Trigger re-renders when the store updates
+  useEffect(
+    () => {
+      const onStoreChanged = () => {
+        setContext(state => ({
+          ...state,
+          isSignedIn: !!store.credentials,
+          isAdmin: store.isAdmin,
```
netpro2k, line 75: Should this be `store.credentials.isAdmin`? Right now I bet any store update would cause the admin button to get hidden since this is probably `undefined`

src/cloud.js
```diff
@@ -3,198 +3,100 @@ import React from "react";
...
-          </IfFeature>
-          <IfFeature name="show_docs_link">
-            <li>
-              <a href={configs.link("docs", "https://hubs.mozilla.com/docs")}>
-                <FormattedMessage id="home.docs_link" />
+    <Page>
+      <div className={styles.hero}>
+        <section className={styles.colLg}>
+          <div className={classNames(styles.hideLgUp, styles.centerLg)}>
+            <h1>Introducing Hubs Cloud</h1>
```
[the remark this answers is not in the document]
robertlong: Yes, but I'd like to do that in another PR since there's already a lot going into this. This is the same content from the original cloud.js

src/react-components/auth/AuthContext.js
```diff
@@ -0,0 +1,96 @@
...
+          retPhxChannel.push("refresh_perms_token").receive("ok", ({ perms_token }) => {
+            const perms = jwtDecode(perms_token);
+            retPhxChannel.leave();
+            resolve(perms);
+          });
+        });
+      });
+
+      configs.setIsAdmin(perms.postgrest_role === "ret_admin");
+      store.update({ credentials: { isAdmin: perms.postgrest_role === "ret_admin" } });
```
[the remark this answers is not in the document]
robertlong: I don't think we should block the page load to check if the user is an admin especially since a majority of users of Hubs are not admins. Instead we should get this isAdmin flag into the JWT session. We should make JWT sessions expire relatively quickly and use [refresh tokens](https://auth0.com/blog/refresh-tokens-what-are-they-and-when-to-use-them/) in order to ensure that changes to profiles take effect in a reasonable period of time. Then, since we will be storing this JWT access token in localStorage, you will be able to check whether the user is an admin without blocking the page load.

That's out of scope for this PR though.

I will revert to blocking the page for now. However, the `Hubs.store.isAdmin()` method that I use in the Hubs API PR makes use of this and will need to be changed. I would like to resolve this before merging the API PR.

src/react-components/home/HomePage.js
robertlong: I don't actually know, I just copy pasted that.

src/react-components/home/HomePage.js
robertlong: I think we will need to make the breaking change in Reticulum, but keeping the old emails and this current redirect method for one release cycle seems like a good upgrade path.

src/react-components/home/HomePage.js
robertlong: There's currently a hook that will support this sign in flow in the HomePage API PR, but yes I would like to get rid of that so I'll move it into index.js now. My current thoughts are that we will deprecate the old sign-in urls, ship this to all Hubs Cloud instances, ask users to update their clients, and update the email template to use the new urls. At that point we can safely remove the redirect from the index.js page.

How does the default room get handled now? I haven't tested that yet.

src/react-components/home/useFeaturedRooms.js
robertlong: Also, eventually, pagination, and additional filtering.

src/react-components/auth/AuthContext.js
robertlong: No, but the rules of hooks expect it to be passed in as a dependency. eslint will complain otherwise.

src/react-components/auth/AuthContext.js
robertlong: Yeah, I think we can rethink this in the API PR. I already have changed it a bit.

src/react-components/auth/AuthContext.js
robertlong: Yes, and they are in the API PR.

src/react-components/auth/AuthContext.js
robertlong: I plan on it in the API PR.

src/react-components/auth/AuthContext.js
robertlong: Pretty certain it should be.

src/react-components/auth/AuthContext.js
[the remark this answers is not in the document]
netpro2k: I agree we should not block page loading, you just don't show the admin button right away.

PR #2467, opened 2020-05-05
Fix transparency on some images, notably avatar previews

### LINKED ISSUE #2463: Tranparency of avatar thumbnails (possibly other images) breaks on scene transition
**Description**
When switching scenes the transparent part of avatar thumbnails breaks. Its possible this happens with other images as well but so far only seeing it with avatar thumbnails

**To Reproduce**
1. Paste in a link to an avatar
2. Change the scene
3. Note that avatar transparency is now broken 

**Expected behavior**
Avatar transparency should be preserved on scene transition

**Screenshots**
image

**Hardware**
 - Device: Desktop
 - OS: Linux
 - Firefox

**Additional context**
We have seen this a bunch during meetups

netpro2k:
Transparency on some images, particularly web previews, most notably avatar previews, was sometimes being set incorrectly. This manifested in some weird issues, particularly on scene transitions.

Fixes this by:
1. correctly checking batching flags for web thumbnails (they were missed when batching was flipped to disabled by default) 
2. preventing `transparent` from being `undefined`, this was happening for cases where `image` or `image.hasAlpha` was undefined.

fixes #2463

### DIFF
--- a/src/components/media-loader.js
+++ b/src/components/media-loader.js
@@ -589,7 +589,7 @@ AFRAME.registerComponent("media-loader", {
           { once: true }
         );
         this.el.setAttribute("floaty-object", { reduceAngularFloat: true, releaseGravity: -1 });
-        let batch = !disableBatching;
+        let batch = !disableBatching && forceImageBatching;
         if (this.data.mediaOptions.hasOwnProperty("batch") && !this.data.mediaOptions.batch) {
           batch = false;
         }
--- a/src/components/media-views.js
+++ b/src/components/media-views.js
@@ -1104,7 +1104,7 @@ AFRAME.registerComponent("media-image", {
       !this.data.batch ||
       texture == errorTexture ||
       this.data.contentType.includes("image/gif") ||
-      (texture.image && texture.image.hasAlpha);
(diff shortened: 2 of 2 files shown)


src/components/media-loader.js
```diff
@@ -589,7 +589,7 @@ AFRAME.registerComponent("media-loader", {
           { once: true }
         );
         this.el.setAttribute("floaty-object", { reduceAngularFloat: true, releaseGravity: -1 });
-        let batch = !disableBatching;
+        let batch = !disableBatching && forceImageBatching;
```
johnshaughnessy, line 592: nit : These flags are awkward. Batch if it's not disabled and it's forced on... but forceImageBatching can be disabled...

netpro2k: Yep, agreed super awkward... The way it works right now is batching is not enabled by default, it becomes enabled if you either `forceImageBatching` or `forceMeshBatching`, the `disableBatching` flag no longer really makes too much sense since its off by default.. The intent was to leave things in the state they would be in for if/when we enable batching again by default... It could be awhile so maybe warrants revisiting that..

PR #2470, opened 2020-05-05
Disable audio input processing on FxR

InfiniteLee:
Disable audio input processing (echoCancellation, noiseSuppression and autoGainControl) on FxR by default as a workaround for https://bugzilla.mozilla.org/show_bug.cgi?id=1626081.

Also expose preferences to enable/disable audio input processing for ease of future debugging.

image


src/react-components/ui-root.js
```diff
@@ -623,6 +624,28 @@ class UIRoot extends Component {
...
+    constraints.audio.autoGainControl = true;
+
+    if (isFirefoxReality) {
+      //workaround for https://bugzilla.mozilla.org/show_bug.cgi?id=1626081
+      constraints.audio.echoCancellation =
+        window.APP.store.state.preferences.disableEchoCancellation === false ? true : false;
+      constraints.audio.noiseSuppression =
+        window.APP.store.state.preferences.disableNoiseSuppression === false ? true : false;
+      constraints.audio.autoGainControl =
+        window.APP.store.state.preferences.disableAutoGainControl === false ? true : false;
```
johnshaughnessy, line 638: If the user does not set a preference for this, it will be `undefined` (not `false`), which will mean these are all disabled by default. Is this the intended result?

(The `defaultBool` in the preference screen only determines whether the checkbox is checked, but does not actually result in anything being put in the store for that preference. The thinking behind this is to make it so that values in the store can be thought of as being explicitly chosen by the user :
https://github.com/mozilla/hubs/blob/master/src/react-components/preference-list-item.js#L119

We should probably have a simple method for getting the default value for a given preference, so that there's less of a chance of a mismatch between what the list item shows and what the code does.)

src/react-components/ui-root.js
```diff
@@ -623,6 +624,28 @@ class UIRoot extends Component {
       this.state.audioTrackClone.stop();
     }
 
+    constraints.audio.echoCancellation = true;
+    constraints.audio.noiseSuppression = true;
+    constraints.audio.autoGainControl = true;
```
johnshaughnessy, line 629: Should the preferences be used here, or are they only available for FxR?

johnshaughnessy: If I understand the `MediaTrackConstraints` correctly, the `getUserMedia` call below will only succeed if an audio track can be found with these constraints. Do we need a fallback in case _some_ audio track is available, but not one that is capable of `echoCancellation`, `noiseSuppression`, and `autoGainControl`?

src/assets/translations.data.json
```diff
@@ -290,6 +290,9 @@
     "preferences.disableTeleporter": "Disable teleporter",
     "preferences.movementSpeedModifier": "Movement speed modifier",
     "preferences.disableAutoPixelRatio": "Disable automatic pixel ratio adjustments",
+    "preferences.disableEchoCancellation": "Disable microphone echo cancellation (requires restart)",
```
johnshaughnessy, line 293: If these are only available on FxR, perhaps add that into the description or put it into a separate heading (like the touchscreen preferences)

src/react-components/ui-root.js
InfiniteLee: yes, they are "disable" flags so they should be false by default. Only on FxR will they default to true, which means they are disabled.

src/assets/translations.data.json
InfiniteLee: These are available on any platform. We only will set them to true (disable those features) in the case of FxR.

src/react-components/ui-root.js
InfiniteLee: these flags are a no-op in the case that somehow they are not available, though I don't ever expect that to be the case. Strictly speaking, setting these 3 lines are not necessary as they default to true anyways, but I wanted to be explicit since it's not obvious that is what is happening.

src/react-components/ui-root.js
InfiniteLee: oh wait i have a typo here

src/assets/translations.data.json
johnshaughnessy: I only see them being used in a `isFirefoxReality` block. Am I missing something?

src/react-components/ui-root.js
InfiniteLee: I realize this is confusing, the intent is this (for FxR):
e.g. for echo cancellation 
if `disableEchoCancellation` is undefined or TRUE, the constraint should default to false.
if `disableEchoCancellation` is explicitly FALSE, its assumed that the user explicitly wants to enable it anyways, despite being default to off in FxR, so the constraint is set to true.

src/assets/translations.data.json
InfiniteLee: Oh you're right, im sorry. I need to be reading these preferences where i do: 
```
    constraints.audio.echoCancellation = true;
    constraints.audio.noiseSuppression = true;
    constraints.audio.autoGainControl = true;
```

PR #2481, opened 2020-05-09
Admin panel: Create accounts using email or create accounts in a batch with emails separated by ;

robin-k-wilson:


robin-k-wilson:
^^ many commits because I could only see the admin console when deployed on hubs-cloud

admin/src/react-components/accounts.js
```diff
@@ -40,29 +40,143 @@ const AccountFilter = props => (
...
     }
     render() {
       const { classes } = this.props;
-
       return (
         <>
+          <Card classes={{ root: classes.searchCard }}>
+            <CardContent>
+              <Typography component="h2">
+                Create one account or mutliple: separate emails with &quot;;&quot;)
```
robertlong, line 141: Is that a hanging `)`? I can't tell.

robin-k-wilson:
image

image

admin/src/react-components/accounts.js
```diff
@@ -40,29 +40,143 @@ const AccountFilter = props => (
...
       }).then(r => r.json());
       if (result && result.data) {
         window.location = `#/accounts/${result.data[0].id}`;
       } else {
-        this.setState({ searching: false, accountSearchStatus: "Account not found" });
+        this.setState({ searching: false, searchStatus: "Account not found" });
       }
+      // Quickfix snackbar component does not always close
+      // Setting snackbar message to empty string closes
+      setTimeout(() => this.setState({ searchStatus: "" }), 5000);
```
robertlong, line 71: It's possible this gets called after the component has unmounted. Maybe clear the timeout in `componentWillUnmount`?

robin-k-wilson:
To see the accounts in the list, you need to refresh the page.

robertlong:
> To see the accounts in the list, you need to refresh the page.

Maybe we should refresh the page? Also is there any loading indicator?

PR #2489, opened 2020-05-11
Remove unused duck component

netpro2k:
This code is no longer being used and is using old physics system code so basically needs to be rewritten if we did want to use it again.


robertlong:
LGTM 👍

PR #2494, opened 2020-05-12
Fixes duck & quack

InfiniteLee:
Update `duck` to use the new `physics-system`. Update `quack` to use the `interaction-system`. Update  DuckyMesh.glb to use the `duck` component.

(The quack component was used on spawners in old pre-spoke scenes but it doesn't seem to work anymore, so updating it just to randomly play quack when the duck is grabbed). 

Closes #2489


src/components/quack.js
```diff
@@ -1,24 +1,38 @@
...
+    let isInteracting = false;
+    for (let i = 0; i < INTERACTORS.length; i++) {
+      if (interaction.state[INTERACTORS[i]].held === entity) {
+        isInteracting = true;
+        if (!this.wasInteracting) {
+          this.quack();
+        }
+        break;
+      }
+    }
```
johnshaughnessy, line 30: There's an `isHeld` helper you could use:
```
const isInteracting = interaction.isHeld(entity);
if (isInteracting && !this.wasInteracting) {
  this.quack();
}
this.wasInteracting = isInteracting;
```

src/components/duck.js
```diff
@@ -11,35 +10,24 @@ AFRAME.registerComponent("duck", {
...
       const force = ratio * this.data.maxForce;
       if (force > 0) {
         const angle = Math.random() * Math.PI * 2;
         const x = Math.cos(angle);
         const z = Math.sin(angle);
-        this.force.set(x, force, z);
-        this.position.set(x * 0.01, 0, z * 0.01);
-        this.el.object3D.matrixNeedsUpdate = true;
-        this.el.body.applyForce(this.force, this.position);
+        this.networkedEntity.setAttribute("body-helper", { gravity: { x: x, y: force, z: z } });
```
johnshaughnessy, line 30: could write `{x, y: force, z}` if you prefer

src/gltf-component-mappings.js
```diff
@@ -22,7 +22,10 @@ function registerRootSceneComponent(componentName) {
 registerRootSceneComponent("fog");
 registerRootSceneComponent("background");
 
-AFRAME.GLTFModelPlus.registerComponent("duck", "duck");
+AFRAME.GLTFModelPlus.registerComponent("duck", "duck", el => {
+  el.setAttribute("duck", "");
+  el.setAttribute("quack", { quackPercentage: 0.1 });
```
johnshaughnessy, line 27: Is this here just to avoid having to modify the gltf we have?

src/components/duck.js
```diff
@@ -11,35 +10,24 @@ AFRAME.registerComponent("duck", {
...
+      this.initialScale = this.networkedEntity.object3D.scale.x;
+      this.maxScale = this.data.maxScale * this.initialScale;
+    });
   },
 
-  beforeStep: function() {
-    if (this.el.body && NAF.utils.isMine(this.el)) {
-      const currentScale = this.el.object3D.scale.x;
+  tick: function() {
+    if (this.networkedEntity && NAF.utils.isMine(this.networkedEntity)) {
```
johnshaughnessy, line 22: If this component is not under a networked entity, can we use `this.el` and skip checking `isMine`?

src/gltf-component-mappings.js
```diff
@@ -22,7 +22,10 @@ function registerRootSceneComponent(componentName) {
 registerRootSceneComponent("fog");
 registerRootSceneComponent("background");
 
-AFRAME.GLTFModelPlus.registerComponent("duck", "duck");
+AFRAME.GLTFModelPlus.registerComponent("duck", "duck", el => {
```
johnshaughnessy, line 25: Might rename this component to something like `"float-away-when-scaled"`

src/components/quack.js
InfiniteLee: ah yea that's much better, thanks

src/gltf-component-mappings.js
InfiniteLee: makes it easier to change the quackPercentage rather than having to unbundle the glb, update the gltf, then rebundle it.

src/components/duck.js
InfiniteLee: yep

Issue #2474, opened 2020-05-06
Logo occludes hubs cloud access code with HMD & FF Reality


ayman:
**Description**
In Firefox Reality on an Oculus entering standalone mode, the user supplied logo overlaps the pin pad entry on an HMD.  Clicking the occluded pin buttons hits the logo and takes you to the hubs's root url (like hubs.mozilla.com).  

The workaround is to tell FireFox Reality to Request Desktop site and then there is no logo and the pin pad works fine.

**To Reproduce**
1. In a hubs cloud instance, set all the image logs in the app setting to the same image (this case App Logo, App Icon, App Thumbnail, Company Logo was all the same 200px x 165px image). 
2. Get a room and code to enter.
3. On an Oculus Quest, enter a hubs.link url into the nav bar.
4. Type the room pin/code.
5. If the code overlaps the logo, you'll be take home when you try to hit that button.

**Expected behavior**
Images shouldn't overlap with the pinpad so buttons can work.

Alternatively, the images should be checked for size (or the description can hint to a good target size).

**Screenshots**
In this example, the occluded buttons and even the 3 doesn't work. 

![96082682_10103944636813025_2223064459545411584_o](https://user-images.githubusercontent.com/1846/81228736-74308700-8fa3-11ea-9215-e1cbd57b31ca.jpeg)

**Hardware** 
 - Server: Hubs Cloud with Logos set in admin pane as described above.
 - Device: Oculus Quest (Headset version 602380.24900.0)
 - OS: Oculus
 - Browser (Tested): Firefox reality 9.1 3-31-2020

piercechueng:
The same on my Oculus Quest with Firefox reality. 
2 screenshots: in windows size 1.0 and 2.0.
the 2.0 one looks like the same issue.

![org mozilla vrbrowser-20200512-151654-1](https://user-images.githubusercontent.com/61266081/81752707-da674f00-9466-11ea-89c6-4ccb46878422.jpg)
![org mozilla vrbrowser-20200512-151626-1](https://user-images.githubusercontent.com/61266081/81752712-e2bf8a00-9466-11ea-8dd1-1c68887a1276.jpg)

piercechueng:
oh, and also on iPad with Safari:

![WIN_20200511_15_28_03_Pro](https://user-images.githubusercontent.com/61266081/81753285-4c8c6380-9468-11ea-84b4-aed0a967800e.jpg)

antpb:
[...]

PR #2498, opened 2020-05-13
Fix 2474 - Set relative logo positioning

antpb:
I've noticed the issue outlined in https://github.com/mozilla/hubs/issues/2474 where the logo has the ability to overlap with the dial pad for room login. This is due to the absolute positioning of the logo used on that page. This PR may have unintended consequences so testing probably should be done before merge. 

here's what the logo looks like with relative positioning set. 

ezgif-4-a00f2833aa2e


antpb:
yikes. Commits and a dirt package-lock are in this PR. Cleaning that up.

PR #2511, opened 2020-05-16
Allow mixing of multiple outbound mediaStreams' audio

InfiniteLee:
This update refactors how outbound audio is handled. Instead of using the `mediaStream` returned from `getUserMedia` directly in NAF/naf-janus-adapter, the stream is passed into the web audio API via a new `AudioContext` to allow for audio from multiple `mediaStreams` to be "mixed" together. The output of this new `AudioContext` is a new `mediaStream` that is then sent over the network. This allows for mixing of audio from additional sources such as screensharing on Chrome, as well as generally simplifying some logic around how the `audio-feedback` component works as well as how we handle when your microphone stream has to be recreated/reconnected (e.g. sometimes Quest will kill the microphone stream so it has to restarted).

This also has the side effect that every outbound audio stream can be individually processed via the Web Audio API (so you can control gain for your microphone, for example). Exposing this ability in preferences will come in a later update. This should also allow for re-selection of your microphone device (without restarting hubs) to be much easier to do in a future update as well.

### DIFF
--- a/src/systems/audio-system.js
+++ b/src/systems/audio-system.js
@@ -61,19 +61,32 @@ export class AudioSystem {
       evt.detail.cameraEl.getObject3D("camera").add(sceneEl.audioListener);
     });
 
+    const ctx = THREE.AudioContext.getContext();
+    this.outboundContext = new (window.AudioContext || window.webkitAudioContext)();
+    this.audioNodes = new Map();
+
     /**
      * Chrome and Safari will start Audio contexts in a "suspended" state.
      * A user interaction (touch/mouse event) is needed in order to resume the AudioContext.
      */
-    const ctx = THREE.AudioContext.getContext();
     const resume = () => {
       ctx.resume();
+      this.outboundContext.resume();
 
-      setTimeout(function() {
-        if (ctx.state === "running") {
+      setTimeout(() => {
+        if (ctx.state === "running" && this.outboundContext.state === "running") {
           if (!AFRAME.utils.device.isMobile() && /chrome/i.test(navigator.userAgent)) {
             enableChromeAEC(sceneEl.audioListener.gain);
           }
+
+          this.mediaStreamDestinationNode = this.outboundContext.createMediaStreamDestination();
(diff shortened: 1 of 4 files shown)


src/systems/audio-system.js
```diff
@@ -61,20 +61,33 @@ export class AudioSystem {
...
 
+          this.mediaStreamDestinationNode = this.outboundContext.createMediaStreamDestination();
+          this.outboundStream = this.mediaStreamDestinationNode.stream;
+          this.outboundGainNode = this.outboundContext.createGain();
+          this.outboundAnalyser = this.outboundContext.createAnalyser();
+          this.outboundAnalyser.fftSize = 32;
+          this.analyserLevels = new Uint8Array(this.outboundAnalyser.fftSize);
+          this.outboundGainNode.connect(this.outboundAnalyser);
+          this.outboundAnalyser.connect(this.mediaStreamDestinationNode);
+
```
johnshaughnessy, line 90: Not related to this commit but we avoid the call to `removeEventListener` below by passing `once:true` as `options` to `addEventListener`:
```
document.body.addEventListener("touchend", resume, {capture: false, once: true});
```
https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#Parameters

src/systems/audio-system.js
```diff
@@ -84,4 +97,25 @@ export class AudioSystem {
...
+  removeStreamFromOutboundAudio(id) {
+    if (this.audioNodes.has(id)) {
+      const nodes = this.audioNodes.get(id);
+      nodes.sourceNode.disconnect();
+      nodes.gainNode.disconnect();
+      this.audioNodes.delete(id);
+    }
+  }
 }
```
johnshaughnessy, line 121: Sorry, another comment unrelated to the lines changed in this PR... 

Here we call `setLocalDescription` and `setRemoteDescription`, which both return promises. Should we be `await`ing on them (or using `.then`)?

https://github.com/mozilla/hubs/blob/1477448a4576eb693b31c9004fbdd4eb39a30c1d/src/systems/audio-system.js#L46

PR #2513, opened 2020-05-18
Map xbox gamepad right joystick horizontal axis to camera yaw. Fixes #2449

### LINKED ISSUE #2449: Gamepad doesn't register right joystick X axis input for desktop gaze orientation
**Description**
When using an XBox gamepad with Hubs, the position joystick (left) input works, and the orientation joystick (right) registers Y axis input but not X axis input. I've tested this on multiple computers with Firefox, Edge (Chromium), Chome, and the Windows Hubs app, with the same issue recurring. 

**To Reproduce**
1. Using a gamepad, attempt to orient yourself with the joystick along the X axis. 

**Expected behavior**
The joystick should allow for X-axis input

**Screenshots**

**Hardware** 
 - Device: Desktop
 - OS: Windows
 - Browser: Chrome, Firefox, Edge (Chromium), Hubs App (Windows)

luser:
Previously this axis was being mapped to `angularVelocity` which doesn't seem to be correct. With this change I can move the camera as expected with an Xbox controller.

### DIFF
--- a/src/systems/userinput/bindings/xbox-controller-user.js
+++ b/src/systems/userinput/bindings/xbox-controller-user.js
@@ -7,6 +7,7 @@ const button = paths.device.xbox.button;
 const axis = paths.device.xbox.axis;
 const scaledRightJoystickVertical = paths.device.xbox.v("scaledRightJoystickVertical");
 const scaledLeftJoystickCursorDelta = paths.device.xbox.v("scaledLeftJoystickCursorDelta");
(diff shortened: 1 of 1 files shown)


luser:
> 'm noticing that the right joystick is extremely sensitive. Did you find the same thing and do you think we should decrease the sensitivity?

I did, which is why I knocked that "horizontal look speed modifier" scale factor down from 0.6 to 0.125 (which matches the vertical scale factor, it also needed the sign inverted). That felt OK to me but you should certainly feel free to adjust it further if desired.

PR #2520, opened 2020-05-21
Fix non-detached heads for avatars

netpro2k:
This PR fixes avatars with non detached heads stretching out the neck verts to the last position the avatar was updated (spawn or when camera tool is open). It does this by continuing to update the head bone's matrix even when it is hidden. 

Fixing this uncovered another issue where animations were still playing on the invisible head bone. This has been fixed by moving animation-mixer and bone-visibility into systems and ensuring they tick in the correct order such that bone-visibility has the final say.

### DIFF
--- a/src/components/animation-mixer.js
+++ b/src/components/animation-mixer.js
@@ -2,16 +2,29 @@
  * Instantiates and updates a THREE.AnimationMixer on an entity.
  * @component animation-mixer
  */
+
+const components = [];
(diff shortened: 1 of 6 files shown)


mattrossman:
I was banging my head all morning trying to figure out what was wrong with my model. Glad to see there is an application-side fix for this. Hoping to see this merged soon!

PR #2521, opened 2020-05-24
Added a new preference: invertTouchscreenCameraMove

MegaMotion:
Added a new preference: invertTouchscreenCameraMove to define whether or not user wants camera motion to be inverted.

### DIFF
--- a/src/systems/userinput/devices/app-aware-touchscreen.js
+++ b/src/systems/userinput/devices/app-aware-touchscreen.js
@@ -176,8 +176,13 @@ export class AppAwareTouchscreenDevice {
         );
         break;
       case MOVE_CAMERA_JOB:
-        assignment.delta[0] += touch.clientX - assignment.clientX;
-        assignment.delta[1] += touch.clientY - assignment.clientY;
+		if (window.APP.store.state.preferences.invertTouchscreenCameraMove) {
+        	assignment.delta[0] += touch.clientX - assignment.clientX;
+        	assignment.delta[1] += touch.clientY - assignment.clientY;
+		} else {
+            assignment.delta[0] -= touch.clientX - assignment.clientX;
+            assignment.delta[1] -= touch.clientY - assignment.clientY;
+		}
         assignment.clientX = touch.clientX;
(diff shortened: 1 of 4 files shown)


src/systems/userinput/devices/app-aware-touchscreen.js
```diff
@@ -176,8 +176,13 @@ export class AppAwareTouchscreenDevice {
         );
         break;
       case MOVE_CAMERA_JOB:
-        assignment.delta[0] += touch.clientX - assignment.clientX;
-        assignment.delta[1] += touch.clientY - assignment.clientY;
+	if (window.APP.store.state.preferences.invertTouchscreenCameraMove) {
```
johnshaughnessy, line 179: I'm going to move this check closer to where the camera rotation is applied so so that other places where these `delta` values are read still get the non-inverted values.

MegaMotion: Awesome, great news! Thanks!

johnshaughnessy:
Closing this in favor of https://github.com/mozilla/hubs/pull/2555 . Thanks again!

MegaMotion:
Thank you, John!! I'm glad to be able to contribute to the project. Great
work, hubs is awesome!

I'm really glad you fixed the VR camera looking bug, too, that was next on
my list to try to figure out.

Do you have any input into the issue of touchscreen navigation? I see that
there are preferences for a joystick on the left or right, but in my build
that doesn't do anything, and you are forced to pinch your way around the
world, which I find extremely awkward. My next task after I get your
changes merged (and get back to working on my client) will be to try to
enable those or make my own system for driving the camera around.

Anyway, thanks again,  have a great day!
Chris Calef
Co-Coordinator, the M8rix (Virtual Oregon Country Fair)

vincentfretin:
I found this funny. On an A-Frame project I'm working on for a customer, they wanted the opposite, that the desktop behavior matches the touchscreen behavior, I set `look-controls="reverseMouseDrag: true"` for them.

PR #2524, opened 2020-05-25
Merge cn-zh translation

mofadashi:


johnshaughnessy:
Thank you @mofadashi !  

There are some places where we accidentally didn't use the translations file, so I've been meaning to add those keys as well. 

I do not know what to do about missing translations. When we need to add a new item to the translations file -- what is a good way for us to handle that new item in the languages we cannot translate to ourselves? (We will need help from community members like you.)

If we can resolve (or come up with a strategy to resolve) this issue, I'd like to merge this PR!

mofadashi:
Have your thought of using weblate?

Issue #2515, opened 2020-05-18
Preferences Menu Refresh


misslivirose:
**Is your feature request related to a problem? Please describe.**
Our preferences menu has gotten long and full of levers to pull and buttons to poke for a custom Hubs experience. We should improve this so that the menu is easy to navigate so that a user can find what they're looking for. 

**Describe the solution you'd like**
The preferences menu should be broken into different categories of settings, grouped by similar properties. For this pass, I would recommend: 
* General 
* Controls
* Audio (_not present in mockups below_)
* Advanced

| **Category** | Setting           | Description |
| ------------- |-------------| -----|
| General    | Mute microphone on entry  | The default setting for whether your mic is muted when you enter a Hubs room |
| General     | Only show nametags while frozen      | If true, nametags will be hidden unless menus are open |
| General | Incoming voice volume     |   Volume of other avatars |
| General | Media volume | Volume of video or audio media in the environment | 
| General | Disable sound effects | Do not play chat bubble, spawn sounds, or other sound effects |
| Controls | Rotation per snap (in degrees) | How far your avatar turns in a single snap turn | 
| Controls | Disable movement | Prevent your avatar from moving with the joystick or WASD keys |
| Controls | Disable backwards movement | Prevent your avatar from moving backwards | 
| Controls | Disable strafing | Prevent your avatar from strafing | 
| Controls | Disable teleporter | Prevent your avatar from moving with the teleport capability | 
| Controls | Movement speed modifier | How fast you move | 
| Controls | Enable left on-screen joystick for moving around | Displays a virtual left joystick on touch screen devices when in landscape mode, used to move around | 
| Controls | Enable right on-screen joystick for looking around | Displays a virtual right joystick on touch screen devices when in landscape mode, used to look around |
| Audio | Disable microphone echo cancellation (requires restart) | Turns off the echo cancellation audio processing | 
| Audio | Disable microphone noise suppression (requires restart) | Turns off the microphone noise suppression audio processing | 
| Audio | Disable microphone automatic gain control (requires restart) | Turns off the automatic gain control audio processing | 
| Advanced | Max Resolution (width x height in pixels) | Sets the max resolution for the Hubs canvas | 
| Advanced | Material quality (requires restart) | Overrides the default material quality for a scene | 
| Advanced | Disable automatic pixel ration adjustments | ??? | 
[...]

Issue #2056, opened 2020-01-17
Preferences page should show a refresh button after changing preferences


robertlong:
**Is your feature request related to a problem? Please describe.**
Some preferences on preferences page require a "restart" to take effect. Users may not know this means they need to refresh their browser.

**Describe the solution you'd like**
We should show a refresh button when those settings are changed.

**Describe alternatives you've considered**
We could also change the copy to say "requires page refresh" instead of "requires restart"

Issue #2050, opened 2020-01-16
Preferences text fields won't accept more than one input value


j-conrad:
**Description**
In the Hubs Preferences page, the text fields will not accept more than one input value. Typing a value exits the field before you can type another. So, for example, trying to type '30' into the Rotation Snap field only lets you type '3'. The only way to get a '30' in there is to paste it.

**To Reproduce**
1. Open Hubs Preferences page
2. Click on the field for Rotation Snap, for example
3. Try to type '30'
4. You can only ever type the first character and you cannot 'insert' your cursor anywhere.

**Expected behavior**
I should be able to type normally in those fields.

**Screenshots**
![image](https://user-images.githubusercontent.com/31968496/72484034-d9bff400-37b7-11ea-993c-ac139c6c4447.png)

**Hardware** 
 - Device: Desktop or HMD
 - OS: Windows
 - Browser: Firefox or Firefox Reality

**Additional context**
The sliders probably don't need the precision of multiple decimal places, which make selecting integer values nearly impossible. Fixing that would ostensibly eliminate this bug.

gfodor:
@johnshaughnessy mind taking this one?

PR #2527, opened 2020-05-28
Improve Preference Screen

johnshaughnessy:
- Adds navigation bar and subcategories to help users find what they are looking for
- Adds notification that prompts the user to refresh the page when specific preferences are changed
- Updates styles for a more uniform look and better legibility on small screens
- Fixes a bug causing numerical inputs to break after a single character is entered
- Adds `step` and `digits` properties to number-with-range-slider inputs for improved interaction and display

image

image

Closes https://github.com/mozilla/hubs/issues/2515
Closes https://github.com/mozilla/hubs/issues/2056
Closes https://github.com/mozilla/hubs/issues/2050


johnshaughnessy:
TODO:
- [ ] Add descriptions / help text for non-obvious preferences.
- [ ] Add the ability to scroll the navigation bar left/right on small screens. (This also unlocks the ability to give the top-level categories longer names like "Miscellaneous", which wouldn't otherwise fit on small screens.)

PR #2554, opened 2020-06-08
Add uv-scroll component to animate UV offsets

netpro2k:
The `uv-scroll` component will animate the offsets for maps associated with the material of the `mesh` of an entity. This can be expanded to become more feature and target other meshes, multiple materials, etc, but this has high utility and low impact so it felt silly not to write. It assumes the material has at least a `base` map.

image

### DIFF
--- a/src/components/uv-scroll.js
+++ b/src/components/uv-scroll.js
@@ -4,29 +4,11 @@
  */
 AFRAME.registerComponent("uv-scroll", {
   schema: {
-    speed: { type: "vec2", default: { x: 0, y: 0 } },
-    animateBase: { type: "boolean", default: true },
-    animateEmissive: { type: "boolean", default: true },
-    animateNormal: { type: "boolean", default: false },
-    animateAO: { type: "boolean", default: false },
-    animateRoughness: { type: "boolean", default: false },
-    animateMetalness: { type: "boolean", default: false },
-    animateLightmap: { type: "boolean", default: false }
-  },
-  init() {
-    this.offset = new THREE.Vector2();
+    speed: { type: "vec2", default: { x: 0, y: 0 } }
   },
   tick(_t, dt) {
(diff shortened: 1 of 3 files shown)


src/components/uv-scroll.js
```diff
@@ -0,0 +1,14 @@
+/**
+ * Animate the UV offset of a mesh's material
+ * @component uv-scroll
+ */
+AFRAME.registerComponent("uv-scroll", {
+  schema: {
+    speed: { type: "vec2", default: { x: 0, y: 0 } }
```
johnshaughnessy, line 7: nit: May call this `velocity` because it has direction and magnitude

netpro2k: Was matching similar properties we already have on particle systems, and also didn't want it to incorrectly be associated with physics.

markvandenbergen:
Hi this is great. 
Do I need AWS Hubs Cloud for this to work, or will it work in a default Mozilla hubs scene?

andybak:
Where are the docs on actually using this?

vincentfretin:
I think https://www.twitch.tv/videos/689725972 is related to this.

andybak:
> I think https://www.twitch.tv/videos/689725972 is related to this.

So for other people that loathe video for documentation as much as me I was hoping to summarise what that video shows you need to do but I'll admit I'm baffled. 

It looks like the Hubs UI (rather than Spoke) but I can't work out how to bring up the component UI.

It's a couple of months since I've used Hubs in any depth so I might have forgotten something obvious.

Can anyone enlighten me?

vincentfretin:
@andybak I watched the full video some weeks ago, not only the highlight. As far I understand it, the component is meant to be used via a gltf model created with blender with https://github.com/MozillaReality/hubs-blender-exporter

vincentfretin:
Ah I found it again on youtube, this video: https://www.youtube.com/watch?v=zOzTGby8Iag

PR #2563, opened 2020-06-12
Add ability for room owners to mute users in a room

InfiniteLee:
This updates the presence list to show microphone state for all users (muted, unmuted, and talking) and allows room owners to mute other users. By clicking "Mute All", all non-owners will be muted. By clicking on an individual volume icon, a specific user can be muted. Non-owners will not see the "Mute All" button and cannot click on individual users to mute them. This is currently not a toggled mute- any users muted this way can unmute themselves normally. 

mute-all
mute-individual


src/systems/microphone-presence-system.js
```diff
@@ -0,0 +1,58 @@
...
+          if (
+            microphonePresences.hasOwnProperty(networkedAudioAnalyser.playerSessionId) &&
+            networkedAudioAnalyser.volume > VOLUME_THRESHOLD
+          ) {
+            microphonePresences[networkedAudioAnalyser.playerSessionId].talking = true;
+          }
+        }
+      }
+
+      if (this.remountUI) this.remountUI({ microphonePresences });
```
robertlong, line 55: I don't think we should use this `remountUI` method. I'd just emit an event when the muted/talking state changes and listen for it in the user list component. Otherwise you're diffing the whole app's UI every frame.

src/react-components/presence-list.js
```diff
@@ -128,10 +171,21 @@ export default class PresenceList extends Component {
   }
 
   renderExpandedList() {
+    const meta = this.props.presences[this.props.sessionId].metas[0];
```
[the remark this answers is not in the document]
InfiniteLee: I believe so, since you will always be in the array yourself- but perhaps worth adding a check anyways.

src/systems/microphone-presence-system.js
InfiniteLee: this only is updated once per second- is it better to fire an event every second or call remountUI every second? My understanding is the cost of calling remountUI is negligible in the case that the component isn't being rendered (in this case, it's the presence list which is only visible when you have it open) but I'm not familiar enough with react to know for sure. Also in my experience you typically don't want to fire events on repeat in a loop.

InfiniteLee: Or is calling remountUI bad in this case because it's going through ui-root, so it's causing ui-root to update everything? I thought react was smart and knew to only update components that actually had properties that changed.

src/react-components/presence-list.js
```diff
@@ -76,13 +108,24 @@ export default class PresenceList extends Component {
         &#x2605;
       </span>
     );
+    const microphonePresence = this.props.microphonePresences[sessionId];
+    const micState =
+      microphonePresence && meta.presence === "room" ? getMicrophonePresenceIcon(microphonePresence) : "";
+    const micIconStyles = [microphonePresence && microphonePresence.muted ? styles.iconRed : styles.icon];
```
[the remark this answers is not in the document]
InfiniteLee: Thanks, didn't know about this!

src/react-components/presence-list.js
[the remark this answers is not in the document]
InfiniteLee: oh wait you meant the `metas` array- yes there will always be *only* one entry in there, which is a side effect of the way the presence api from reticulum returns data.

src/systems/microphone-presence-system.js
robertlong: So it's not re-mounting things but it is calling `render()` on the root of the tree. That will require more reconciliation and rerendering than if you listened for the change in the presence list. See this part of the React docs for more details:
https://reactjs.org/docs/reconciliation.html#tradeoffs

I also just don't think that passing this `remountUI` method around is good practice. Here's what I'd do instead:

```js
export default class PresenceList extends Component {
  updateMicrophoneState = () => {
    const microphonePresences = AFRAME.scenes[0].systems["hubs-systems"].microphonePresenceSystem.microphonePresences;
    this.setState({ microphonePresences });
    this.timeout = setTimeout(this.updateMicrophoneState, 1000);
  };

  componentDidMount() {
    this.updateMicrophoneState();
  }

  componentWillUnmount() {
    clearTimeout(this.timeout);
  }
}
```

Alternately:
```js
export default class PresenceList extends Component {
  updateMicrophoneState = () => {
    const microphonePresences = AFRAME.scenes[0].systems["hubs-systems"].microphonePresenceSystem.microphonePresences;
    this.setState({ microphonePresences });
  };

  componentDidMount() {
    document.body.addEventListener("microphone-presence-update", this.updateMicrophoneState);
  }

  componentWillUnmount() {
    document.body.removeEventListener("microphone-presence-update", this.updateMicrophoneState);
  }
}
```

Both of these would only update the part of the UI that needs to be re-rendered.

InfiniteLee: Refactored so that it is no longer in a hubs-system, but rather just a util function (`getMicrophonePresences`). Also updated it so that it will only call this function whenever the presence list is expanded.

PR #2573, opened 2020-06-16
fix broken pdf when the pdf is not on the first page and a participant join

vincentfretin:
Fix broken pdf when a pdf which is not on the first page is in the scene and a new participant joins the room.

You can reproduce the issue like this:
With Participant A:
- create a room
- spawn a pdf
- go to page 2 of the pdf

Then participant B joins and sees a broken pdf (the media-error.gif texture).
The console error is:
```
TypeError: "this.pdf is undefined"
    update media-views.js:1219
    asyncGeneratorStep hub-424d731663df443df67f.js:10178
    _next hub-424d731663df443df67f.js:10180
    _asyncToGenerator hub-424d731663df443df67f.js:10180
    _asyncToGenerator hub-424d731663df443df67f.js:10180
    update media-views.js:1195
```

### DIFF
--- a/src/components/media-views.js
+++ b/src/components/media-views.js
@@ -1216,8 +1216,15 @@ AFRAME.registerComponent("media-pdf", {
         this.el.setAttribute("media-pager", { maxIndex: this.pdf.numPages - 1 });
       }
 
-      const page = await this.pdf.getPage(index + 1);
-      if (src !== this.data.src || index !== this.data.index) return;
+      if (!this.pdf) {
+        // The pdf may not have been loaded yet, this can happen with initial
+        // render followed with a networked update with index > 0
+        // (not the first page of the pdf)
+        return;
+      }
+      // Use here this.data.index and not index to have the networked update.
+      const page = await this.pdf.getPage(this.data.index + 1);
+      if (src !== this.data.src) return;
 
       const viewport = page.getViewport({ scale: 3 });
       const pw = viewport.width;
@@ -1234,7 +1241,7 @@ AFRAME.registerComponent("media-pdf", {
 
       this.renderTask = null;
 
-      if (src !== this.data.src || index !== this.data.index) return;
+      if (src !== this.data.src) return;
     } catch (e) {
       console.error("Error loading PDF", this.data.src, e);
       texture = errorTexture;


src/components/media-views.js
```diff
@@ -1216,8 +1216,15 @@ AFRAME.registerComponent("media-pdf", {
...
-      if (src !== this.data.src || index !== this.data.index) return;
+      if (!this.pdf) {
+        // The pdf may not have been loaded yet, this can happen with initial
+        // render followed with a networked update with index > 0
+        // (not the first page of the pdf)
+        return;
+      }
+      // Use here this.data.index and not index to have the networked update.
+      const page = await this.pdf.getPage(this.data.index + 1);
+      if (src !== this.data.src) return;
```
vincentfretin, line 1227: I removed the "index !== this.data.index" check. We want to render the pdf completely.
The code flows like this:
- first call with `src: test.pdf; index:0`
- pdf is loading (first call)
- second call with `src: test.pdf; index:1` , `(src !== oldData.src) is false` and `this.pdf` is undefined, don't run further
- on the first call, pdf is loaded, `index` equals `0`, `this.data.index` equals `1`, do the rendering with updated `this.data.index`

netpro2k: I believe the `index !== this.data.index` check is still needed since `this.pdf.getPage` is async. If you are paging through quickly you could have a race condition where a previous page ends up finishing after the current page, and you do not want to use that result

netpro2k:
Hmm, I am unable to reproduce this but not surprised there are some race conditions here. The use of async/await makes it pretty hard to tell you maybe have multiple instances of this update function going "at once", and during each `await` your state could have been invalidated and you should stop executing the rest. I think we should be doing a `if (src !== this.data.src || index !== this.data.index) return;` after each await, and then I don't think you would end up in the situation where `this.pdf` is undefined.

vincentfretin:
There are really two issues here.
The first "TypeError: "this.pdf is undefined" happens when `src === this.data.src && index !== this.data.index`
The second issue it that it will render the `index` page, and not the latest `this.data.index` page, so wrongly render once the first page and not render the second page of the pdf that is really at the second page in the scene.

brianpeiris:
Notes from QA's reproduction:

I was able to reproduce this issue following the steps described, in multiple attempts. Tested on Windows 10 using Chrome and Firefox, and also using cross-platform between Windows 10 and mobile.

[Notes]:
- The issue is only reproducible if the file's pages are turned to the second page and up.
- The issue is dismissed if the pages of the PDF are toggled with, after reproducing the issue.
- The issue is not reproducible if the PDF file is pinned.
- Here is a screen recording: https://drive.google.com/file/d/1qCtH4wXGsdiD5TqUSJwZU0uhumc5329A/view?usp=sharing

PR #2580, opened 2020-06-22
Fix messaging for unsupported and legacy browsers

### LINKED ISSUE #2057: Browser incompatibility error messages no longer display, breaking room entry flow on unsupported devices
**Description**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Hardware** 
 - Device: [e.g. Desktop, phone, VR headset]
 - OS: [e.g. Windows, iOS, Linux]
 - Browser [e.g. Firefox]

**Additional context**

brianpeiris:
The previous approach for legacy and unsupported browsers was simply to display a message or dialog as part of the existing UI. This approach broke as we imported new libraries and used newer JS language features and newer browser APIs. 
This PR aims to workaround those issues by separating platform support checks into its own webpack entry point so that it can be self-contained and independent of the rest of the codebase.

Fixes #2057


src/support.js
```diff
@@ -0,0 +1,124 @@
...
+    const detectedOS = detectOS(navigator.userAgent);
+
+    return (
+      <div className={styles.supportMain}>
+        <div className={styles.supportContent}>
+          <div>
+            <img className={styles.logo} src={configs.image("logo")} />
+          </div>
+          <p>
+            Your browser is missing required features.<br />
```
johnshaughnessy, line 88: Can `messages` from `utils/i18n.js` be used here, or does that run the risk of having the page break? (I suspect it breaks if its dependency, `utils/configs.js`, uses an unsupported language feature)

src/assets/stylesheets/support.scss
```diff
@@ -0,0 +1,47 @@
+:local(.support-main) {
```
johnshaughnessy, line 1: Do you want to use any shared styles or variables here, or will that fail in some browsers?

webpack.config.js
```diff
@@ -226,13 +226,14 @@ module.exports = async (env, argv) => {
...
   const host = process.env.HOST_IP || env.localDev || env.remoteDev ? "hubs.local" : "localhost";
 
-  // Remove comments from .babelrc
-  const babelConfig = JSON.parse(
-    fs
-      .readFileSync(path.resolve(__dirname, ".babelrc"))
-      .toString()
-      .replace(/\/\/.+/g, "")
-  );
+  const legacyBabelConfig = {
```
johnshaughnessy, line 229: Do you want to delete `.babelrc` and are there some settings missing? (e.g. the `"react-intl"` plugin and some of the properties of `babel/env`)

src/assets/stylesheets/support.scss
brianpeiris: Yeah, just trying to minimize the surface area that could cause this to break again.

webpack.config.js
brianpeiris: `.babelrc` is still used for all other source files. The env options here are the crucial difference. Specifying a target of IE 11 gives us the lowest possible baseline for the browsers we need to support and we only want to apply that to support.js

src/support.js
brianpeiris: Good point, I'll see if it's possible to use i18n here.

webpack.config.js
```diff
@@ -337,11 +339,9 @@ module.exports = async (env, argv) => {
           }
         },
         {
-          // We reference the sources of some libraries directly, and they use async/await,
-          // so we have to run it through babel in order to support the Samsung browser on Oculus Go.
-          test: [path.resolve(__dirname, "node_modules/naf-janus-adapter")],
```
johnshaughnessy, line 342: Does this comment and test for `naf-janus-adapter` no longer apply?

brianpeiris: Yup, support.js and its direct dependencies are the only files that need to be loaded with this legacy babel config now, and support.js does not require anything related to three.js, aframe or naf.

PR #2597, opened 2020-06-30
Fix uv-scroll on shared materials, add increment

netpro2k:
Fixes `uv-scroll` to not combine offsets when multiple objects share a material. When multiple instances exist, only the values from the first one will have any effect. 

Also adds support for `increment` which when non 0 will make the uv scrolling increment in steps. This is useful for doing "sprite animation" style effects.

The hack for `MobileStandardMaterial` got a bit more ugly here... All the code that cares about Materials could go away if fixed. I did look into doing it the "right way" but it requires a ThreeJS update to add the hook we would need, and that feels a bit heavy handed for this "quick" feature.

### DIFF
--- a/src/components/uv-scroll.js
+++ b/src/components/uv-scroll.js
@@ -2,7 +2,7 @@ const textureToData = new Map();
 const registeredTextures = [];
 
 export class UVScrollSystem {
-  tick(_t, dt) {
+  tick(dt) {
     for (let i = 0; i < registeredTextures.length; i++) {
       const map = registeredTextures[i];
       const { offset, instances } = textureToData.get(map);
@@ -43,7 +43,6 @@ AFRAME.registerComponent("uv-scroll", {
     const mesh = this.el.getObject3D("mesh") || this.el.getObject3D("skinnedmesh");
     const material = mesh && mesh.material;
     if (material) {
-      const offset = new THREE.Vector2();
       // We store mesh here instead of the material directly because we end up swapping out the material in injectCustomShaderChunks.
       // We need material in the first place because of MobileStandardMaterial
       const instance = { component: this, mesh };
@@ -53,7 +52,7 @@ AFRAME.registerComponent("uv-scroll", {
 
       if (!textureToData.has(material.map)) {
         textureToData.set(material.map, {
-          offset,
+          offset: new THREE.Vector2(),
           instances: [instance]
         });
         registeredTextures.push(material.map);
@@ -64,16 +63,17 @@ AFRAME.registerComponent("uv-scroll", {
         textureToData.get(material.map).instances.push(instance);
       }
     }
-    // }, 0);
   },
 
   pause() {
-    const instances = textureToData.get(this.map).instances;
-    instances.splice(instances.indexOf(this.instance), 1);
-    // If this was the last uv-scroll component for a given texture
-    if (!instances.length) {
-      textureToData.delete(this.map);
-      registeredTextures.splice(registeredTextures.indexOf(this.map), 1);
+    if (this.map) {
+      const instances = textureToData.get(this.map).instances;
+      instances.splice(instances.indexOf(this.instance), 1);
+      // If this was the last uv-scroll component for a given texture
+      if (!instances.length) {
+        textureToData.delete(this.map);
+        registeredTextures.splice(registeredTextures.indexOf(this.map), 1);
+      }
     }
   }
 });
--- a/src/components/uv-scroll.js
+++ b/src/components/uv-scroll.js
@@ -1,20 +1,79 @@
+const textureToData = new Map();
+const registeredTextures = [];
+
+export class UVScrollSystem {
+  tick(_t, dt) {
+    for (let i = 0; i < registeredTextures.length; i++) {
+      const map = registeredTextures[i];
+      const { offset, instances } = textureToData.get(map);
+      const { component } = instances[0];
+
+      offset.addScaledVector(component.data.speed, dt / 1000);
+
+      const increment = component.data.increment;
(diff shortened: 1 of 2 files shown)


src/components/uv-scroll.js
```diff
@@ -1,20 +1,79 @@
...
+      }
+    }
+    // }, 0);
+  },
 
-    // TODO this should be handled by MobileStandardMaterial itself
-    if (material.isMobileStandardMaterial) {
-      material.refreshUniforms();
+  pause() {
+    const instances = textureToData.get(this.map).instances;
```
johnshaughnessy, line 71: Do we need a guard here like the `if (material)` guard in `play`? https://github.com/mozilla/hubs/pull/2597/files#diff-5aff2dd73415c1d2a28e0d360f143652R45

src/components/uv-scroll.js
```diff
@@ -1,20 +1,79 @@
...
+        });
+        registeredTextures.push(material.map);
+      } else {
+        console.warn(
+          "Multiple uv-scroll instances added to objects sharing a texture, only the speed/increment from the first one will have any effect"
+        );
+        textureToData.get(material.map).instances.push(instance);
+      }
+    }
+    // }, 0);
```
johnshaughnessy, line 67: Can remove commented snippet here

src/components/uv-scroll.js
```diff
@@ -1,20 +1,79 @@
...
+    increment: { type: "vec2", default: { x: 0, y: 0 } }
   },
-  tick(_t, dt) {
+  play() {
     const mesh = this.el.getObject3D("mesh") || this.el.getObject3D("skinnedmesh");
     const material = mesh && mesh.material;
-    if (!material || !material.map) return;
-    material.map.offset.addScaledVector(this.data.speed, dt / 1000);
+    if (material) {
+      const offset = new THREE.Vector2();
```
johnshaughnessy, line 46: Move this down to the one line where it's used https://github.com/mozilla/hubs/pull/2597/files#diff-5aff2dd73415c1d2a28e0d360f143652R45

src/components/uv-scroll.js
```diff
@@ -1,20 +1,79 @@
+const textureToData = new Map();
+const registeredTextures = [];
+
+export class UVScrollSystem {
```
johnshaughnessy, line 4: Not important to address here (and I know we have the same thing written elsewhere because I've been doing this too) but I think we should consider ending our practice of wrapping `tick` functions in a `class`es when there is no `constructor`, local data, or other `class`-construct helping us in some way. Our preference is something like, "Perform all the transformations related to a particular subsystem at a known point during the frame, and batch like-transformations instead of doing it at a per-entity level." We can stick to this preference without the "`class` with `tick` function" pattern. 

One thing we are missing is a uniform way of storing/accessing entity data, though this doesn't seem to cause us trouble very often in practice. Here (and elsewhere) we work around the problem by initializing/registering some entity data in an aframe component lifecycle method while making sure the the component-data stores `textureToData` and `registeredTextures` are available to the (singular) "system" `tick`.

src/components/uv-scroll.js
```diff
@@ -1,20 +1,79 @@
...
+  tick(_t, dt) {
```
johnshaughnessy, line 5: Remove unused `_t`. 

Not sure if this was intentional (the underscore makes me think it was). If it was I'd argue we don't need/want to pass unused values here (or anywhere)

src/components/uv-scroll.js
```diff
@@ -1,20 +1,79 @@
...
+    for (let i = 0; i < registeredTextures.length; i++) {
+      const map = registeredTextures[i];
+      const { offset, instances } = textureToData.get(map);
+      const { component } = instances[0];
+
+      offset.addScaledVector(component.data.speed, dt / 1000);
+
+      const increment = component.data.increment;
+      map.offset.x = increment.x ? offset.x - (offset.x % increment.x) : offset.x;
```
johnshaughnessy, line 14: Today I learned that `%` works on all numbers, not just integers. (I guess this makes sense, given that javascript's "integers" are just those `Number`s that are not `NaN`, `+∞`, or `-∞`, and satisfies `floor(abs(argument)) == abs(argument)` https://tc39.es/ecma262/#sec-isinteger )

src/components/uv-scroll.js
```diff
@@ -1,20 +1,79 @@
...
+      map.offset.y = increment.y ? offset.y - (offset.y % increment.y) : offset.y;
+
+      // TODO this should be handled by MobileStandardMaterial itself
```
johnshaughnessy, line 17: A bit confused why this is here -- is this the only thing that has ever changed a material's `uvTransform`? Why does `map.updateMatrix`only need to be called for `MobileStandardMaterial`?

src/systems/hubs-systems.js
```diff
@@ -101,6 +103,7 @@ AFRAME.registerSystem("hubs-systems", {
     this.menuAnimationSystem.tick(t);
     this.spriteSystem.tick(t, dt);
     this.enterVRButtonSystem.tick();
+    this.uvScrollSystem.tick(t, dt);
```
johnshaughnessy, line 106: `t` is unused

src/components/uv-scroll.js
netpro2k: It's called by ThreeJS for the built in ThreeJS materials

src/components/uv-scroll.js
netpro2k: Yep I agree, there is not much practical benefit to this except that when you look in HubsSystems you dont have to wonder "Hey why is only the uv scroll system done like that?". I would rather just convert over the ones that make sense all at once. But then also that feels like a waste of time... So we continue doing the silly pattern.

PR #2600, opened 2020-06-30
ensure that occupantsChanged is called when occupants list is mutated

InfiniteLee:
This ensures that `connectedClients` in NAF is populated correctly.

### DIFF
--- a/src/naf-dialog-adapter.js
+++ b/src/naf-dialog-adapter.js
@@ -181,6 +181,10 @@ export default class DialogAdapter {
           this._onOccupantConnected(peer.id);
           this.occupants.push(peer.id);
 
+          if (this._onOccupantsChanged) {
+            this._onOccupantsChanged(this.occupants);
+          }
+
           break;
         }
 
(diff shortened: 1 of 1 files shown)


src/naf-dialog-adapter.js
```diff
@@ -566,13 +578,18 @@ export default class DialogAdapter {
...
-      const peerId = this.occupants[i];
+    const occupantIds = Object.keys(this.occupants);
+    for (let i = 0; i < occupantIds.length; i++) {
+      const peerId = occupantIds[i];
       if (peerId === this._clientId) continue;
       this._onOccupantDisconnected(peerId);
     }
 
-    this.occupants = [];
+    this.occupants = {};
```
johnshaughnessy, line 588: I don't think we have to worry about this but since we since there's a comment that this is a public field it got me thinking -- whatever is looking at `occupants` should _not_ keep a local reference to it because we overwrite it here. If we wanted to mitigate the risk that someone keeps an out-of-date reference to the old object, we'd have to clear it with something like 
```js
  const keys = Object.keys(this.occupants);
  for (const key of keys) {
    delete this.occupants[key];
  }
}
```
instead of setting it to a new empty object

johnshaughnessy: On second thought I see how this is being used in `NAF` and I don't think this will cause any issues

PR #2613, opened 2020-07-08
Remove delete buttons: Accounts, Avatars, Scenes

robin-k-wilson:
Removes unimplemented delete button as an option for editing accounts, scenes, avatars.
Fixed some errors in console about styling.

### DIFF
--- a/admin/src/admin.js
+++ b/admin/src/admin.js
@@ -214,7 +214,10 @@ const HiddenAppBar = withStyles({
       display: "none"
     }
   }
-})(props => <AppBar {...props} className={props.classes.hideOnDesktop} />);
+})(props => {
+  const { classes, ...other } = props;
+  return <AppBar {...other} className={classes.hideOnDesktop} />;
+});
 
 document.addEventListener("DOMContentLoaded", async () => {
   const socket = await connectToReticulum();
--- a/admin/src/react-components/accounts.js
+++ b/admin/src/react-components/accounts.js
@@ -169,6 +169,8 @@ export const AccountList = withStyles(styles)(
         }, 6000);
       }
       render() {
+        // refreshView() is only needed in onCreateAccounts()
+        // eslint-disable-next-line no-unused-vars
         const { classes, refreshView, ...other } = this.props;
         return (
           <>
--- a/admin/src/react-components/accounts.js
+++ b/admin/src/react-components/accounts.js
@@ -10,6 +10,7 @@ import Typography from "@material-ui/core/Typography";
 import CircularProgress from "@material-ui/core/CircularProgress";
 import Snackbar from "@material-ui/core/Snackbar";
 import SnackbarContent from "@material-ui/core/SnackbarContent";
+import { ToolbarWithoutDelete } from "./toolbar-without-delete";
 
 import {
   BooleanField,
@@ -168,10 +169,10 @@ export const AccountList = withStyles(styles)(
         }, 6000);
       }
       render() {
-        const { classes } = this.props;
+        const { classes, refreshView, ...other } = this.props;
         return (
           <>
-            <Card classes={{ root: classes.searchCard }}>
+            <Card className={classes.searchCard}>
               <CardContent>
                 <Typography component="h2">
                   <b>Create one or multiple accounts with (optional) identities</b>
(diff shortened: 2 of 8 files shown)


admin/src/react-components/toolbar-without-delete.js
```diff
@@ -0,0 +1,10 @@
+import React from "react";
+import { Toolbar, SaveButton } from "react-admin";
+
+export const ToolbarWithoutDelete = props => {
```
netpro2k, line 4: Clearly we should have done this a long time ago heh, nice and simple solution. I might have called this `SaveButtonToolbar` or `SaveOnlyToolbar`, but that's personal preference.

admin/src/react-components/accounts.js
```diff
@@ -257,11 +258,11 @@ export const AccountList = withStyles(styles)(
 );
 
 export const AccountEdit = withStyles(styles)(props => {
-  const { classes } = props;
+  const { classes, ...other } = props;
 
   return (
-    <Edit {...props}>
-      <SimpleForm>
+    <Edit {...other}>
```
netpro2k, line 264: Was passing `classes` through breaking something?

admin/src/react-components/accounts.js
```diff
@@ -168,10 +169,10 @@ export const AccountList = withStyles(styles)(
         }, 6000);
       }
       render() {
-        const { classes } = this.props;
+        const { classes, refreshView, ...other } = this.props;
```
netpro2k, line 172: we are no longer passing through `refreshView` here. Is that intended?

admin/src/react-components/accounts.js
```diff
@@ -168,10 +169,10 @@ export const AccountList = withStyles(styles)(
...
         return (
           <>
-            <Card classes={{ root: classes.searchCard }}>
+            <Card className={classes.searchCard}>
```
netpro2k, line 175: Hmm why does this change from `classes` to `className`? Quick search shows that some react-admin stuff expects `classes`, not sure if this is one of them. Possible it was just wrong before. Did you notice it breaking something?

admin/src/admin.js
```diff
@@ -214,7 +214,10 @@ const HiddenAppBar = withStyles({
       display: "none"
     }
   }
-})(props => <AppBar {...props} className={props.classes.hideOnDesktop} />);
+})(props => {
+  const { classes, ...other } = props;
+  return <AppBar {...other} className={classes.hideOnDesktop} />;
```
netpro2k, line 219: Why do we no longer want to pass classes down?

admin/src/react-components/accounts.js
robin-k-wilson: Yes, was throwing an error in my console. Material ui was not liking it.

admin/src/react-components/accounts.js
robin-k-wilson: Yes, thanks for checking, it's being used where it needs to be in the function above outside of render().

admin/src/react-components/accounts.js
netpro2k: Ah right, you mentioned that "Fixed some errors in console about styling".

admin/src/react-components/accounts.js
robin-k-wilson: Added comment to make this clearer: // Do not pass refreshView to other components, only onCreateAccounts()

Was throwing an error getting passed in.

admin/src/react-components/accounts.js
robin-k-wilson: Double checked, both ways work. I've worked with material UI before and this is the simpler preferred way - no errors in console, style is applied!

admin/src/react-components/toolbar-without-delete.js
robin-k-wilson: Open to changing it -- I don't feel strongly -- though I think the name is clear for its purpose, I'd add a comment if I changed it. When we integrate delete functionality we only need to search for "Delete" where it's referenced.

admin/src/admin.js
robin-k-wilson: Throwing the error in console from Material UI, styles still work as intended with the new syntax.

admin/src/react-components/accounts.js
robin-k-wilson: Added eslint ignore line as well.

admin/src/react-components/toolbar-without-delete.js
netpro2k: Yep I don't have strong preference either, just wanted to note.

PR #2619, opened 2020-07-09
Add common Button component

robertlong:
I added a common Button component to encapsulate the button styles. I also moved the home page specific `CreateRoomButton`, `PWAButton`, and `useInstallPWA` files to the home page directory.


src/react-components/input/Button.js
```diff
@@ -0,0 +1,25 @@
+import React from "react";
+import PropTypes from "prop-types";
+import classNames from "classnames";
+import styles from "./Button.scss";
+
+export function Button({ primary, secondary, cta, children, ...rest }) {
```
netpro2k, line 6: Not a super strong opinion here but might be nice to instead have something like `type="primary"`, since they are mutually exclusive and then proptypes could enforce you are not making invalid states.

robertlong: My reasoning behind this was that `primary` and `cta` were being used together. Also `type` should be reserved for `type="submit"` etc.

Issue #2254, opened 2020-03-20
Allow additional characters in usernames


broehl:
Right now, usernames are restricted to alphanumerics plus hyphen. Can that restriction be lifted? Underscores and tildes would be useful.

I'm hosting an event where everyone in attendance has an account on a separate discussion forum. It would be handy to be able to say to people "just use your forum nickname in Hubs, so people can reach you in the forums to follow up on things". However, many of them have underscores and/or tildes.

PR #2623, opened 2020-07-15
Updates to allow _~ in avatar names

lordbron:
One fix is for the regex, while the other is for the message. Not sure if there are other languages that need updating.

Fixes mozilla/hubs#2254

### DIFF
--- a/src/assets/translations.data.json
+++ b/src/assets/translations.data.json
@@ -109,7 +109,7 @@
     "entry.notify_me": "Notify me when others arrive",
     "entry.mute-on-entry": "Mute my microphone",
     "profile.save": "Accept",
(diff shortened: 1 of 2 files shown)


robertlong:
The translations need to be updated as they changed in the redesign. Also we need to check if backend changes are needed in Reticulum. @netpro2k do you want to take a look?

netpro2k:
Verified nothing on the backend seems to care about this. I had to modify things for the updated translations so merged it in manually https://github.com/mozilla/hubs/commit/8a77824ef4813556fff882a155df2583cdb90d6f closing out this PR. Thanks for the fix @lordbron!

rawnsley:
@netpro2k while this is fresh in your mind perhaps you could comment on https://github.com/mozilla/hubs/issues/3934 ?

PR #2675, opened 2020-07-21
Move from node-sass to sass and upgrade webpack

### LINKED ISSUE #1913: Move from node-sass to dart-sass
I think we could improve developer experience by moving from `node-sass` to `dart-sass`. `node-sass` works fine, but uses a native module using`node-gyp` that requires Python 2.7 to be installed. `node-dart` is a drop in replacement and doesn't have this dependency.

robertlong:
The `sass` package removes the python dependency for building hubs and avoids us having to rebuild node-sass.

Fixes #1913
Closes #2609


habitat/plan.sh
```diff
@@ -22,9 +22,6 @@ do_build() {
 
   # main client
   npm ci --verbose --no-progress
-  npm rebuild node-sass # HACK sometimes node-sass build fails
```
netpro2k, line 25: oh god.. well this should cut down our build times quite a bit heh

src/components/virtual-gamepad-controls.css
```diff
@@ -47,9 +47,10 @@
   border-top-right-radius: 50%;
   border-bottom-right-radius: 50%;
   border-bottom-left-radius: 50%;
-  &__hidden {
-    background-color: transparent;
-  }
+}
+
+:local(.hidden) {
```
netpro2k, line 52: Was this just an unrelated fix or is something about this syntax not supported in `sass`?

robertlong: Something about the syntax not being supported in sass. Although I think it's the css-loader not the sass-loader, so I'm not sure what exactly was erroring but this fixed it.

PR #2676, opened 2020-07-21
Update shadows on scene load

### LINKED ISSUE #2650: Turn off Dynamic Shadow Map Rendering on Desktop
Right now shadows only render for the environment, every frame we are recalculating the directional light camera frustum and updating the shadow map. We can likely only recompute these on the first frame after a scene has loaded. Perhaps we also add an option to force enable realtime shadows. Maybe realtime shadows are disabled on medium quality and enabled on high quality? The main downside to this is that animated objects in the environment (ex. the fan in the atrium will not have dynamic shadows).

robertlong:
This PR only recalculates the shadow map and resizes the directional light's shadow map camera frustum when the environment loads. This significantly improves CPU and GPU performance when rendering in high quality.

Fixes #2650


src/systems/shadow-system.js
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    const environmentObject3D = this.environmentObject3D;
+
+    if (!environmentObject3D) {
+      return;
+    }
+
+    if (!this.dynamicShadowsEnabled) {
+      // Recompute the shadowmap when the environment changes.
```
netpro2k, line 53: Is this comment correct? Seems like this block of code is just disabling shadows on animated objects. Assuming so, this code is pretty self describing, so I would just ditch the comment.

src/utils/resizeShadowCameraFrustum.js
```diff
@@ -53,7 +53,7 @@ export default function resizeShadowCameraFrustum(light, scene) {
   verts[6].set(max.x, max.y, min.z);
   verts[7].set(max.x, max.y, max.z);
 
-  inverseLightMatrixWorld.getInverse(light.shadow.camera.matrixWorld);
+  inverseLightMatrixWorld.getInverse(light.matrixWorld);
```
netpro2k, line 56: You are reading an objects matrixWorld here (and above), so you should call `updateMatrices` (as per https://github.com/mozilla/hubs/blob/master/src/utils/threejs-world-update.js#L20). Since you are updating the matrices at the calling location it will be a noop (since the dirty flags wont be set). Alternatively we can note somehow that this function is pushing this responsibility to the caller (maybe we need a naming convention for this)

src/systems/shadow-system.js
```diff
@@ -0,0 +1,73 @@
...
+      traverseAnimationTargets(environmentObject3D, environmentObject3D.animations, animatedNode => {
+        animatedNode.traverse(child => {
+          child.castShadow = false;
+          child.receiveShadow = false;
+        });
+      });
+    }
+
+    environmentObject3D.updateMatrixWorld(true, true);
```
netpro2k, line 62: Do you actually need to forceWorldUpdate and includeInvisible? Also its a bit confusing to be calling this here because this function is not reading any matricies. I try to put it as close as possible to the site where I am reading the matrix otherwise people reading that code will think the matrix is mistakenly out of date. If we move the bounding box calculation out to here like I suggested then it probably makes sense to keep this here with that.

src/utils/resizeShadowCameraFrustum.js
```diff
@@ -53,7 +53,7 @@ export default function resizeShadowCameraFrustum(light, scene) {
   verts[6].set(max.x, max.y, min.z);
```
netpro2k, line 53: maybe not a huge deal now that we are only doing this on scene load, but we compute the scene bounding box again for each light. Might be better to calculate that once and pass it into `resizeShadowCameraFrustum`

src/systems/shadow-system.js
```diff
@@ -0,0 +1,73 @@
...
+export class ShadowSystem {
+  constructor(sceneEl) {
+    this.needsUpdate = false;
+    this.dynamicShadowsEnabled = window.APP.store.state.preferences.enableDynamicShadows;
+    this.sceneEl = sceneEl;
+    this.onEnvironmentSceneLoaded = this.onEnvironmentSceneLoaded.bind(this);
+    this.sceneEl.addEventListener("environment-scene-loaded", this.onEnvironmentSceneLoaded);
+  }
+
+  onEnvironmentSceneLoaded({ detail: scene }) {
```
netpro2k, line 31: Super minor nit, but I would call this `environment` or `environmentObject3D` so as not to confuse it with the aframe scene.

src/systems/shadow-system.js
```diff
@@ -0,0 +1,73 @@
+import resizeShadowCameraFrustum from "../utils/resizeShadowCameraFrustum";
+
+function traverseAnimationTargets(rootObject, animations, callback) {
```
netpro2k, line 3: This seems like a generally useful function, maybe move it to three-utils.js?

src/systems/shadow-system.js
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  onEnvironmentSceneLoaded({ detail: scene }) {
+    this.environmentObject3D = scene;
+    this.needsUpdate = true;
+    this.sceneEl.renderer.shadowMap.autoUpdate = this.dynamicShadowsEnabled;
+  }
+
+  tick() {
+    if (!this.needsUpdate) {
```
netpro2k, line 38: I would combine these 3 preconditions together, but definitely just personal style.

src/systems/shadow-system.js
```diff
@@ -0,0 +1,73 @@
...
+    this.dynamicShadowsEnabled = window.APP.store.state.preferences.enableDynamicShadows;
+    this.sceneEl = sceneEl;
+    this.onEnvironmentSceneLoaded = this.onEnvironmentSceneLoaded.bind(this);
+    this.sceneEl.addEventListener("environment-scene-loaded", this.onEnvironmentSceneLoaded);
+  }
+
+  onEnvironmentSceneLoaded({ detail: scene }) {
+    this.environmentObject3D = scene;
+    this.needsUpdate = true;
+    this.sceneEl.renderer.shadowMap.autoUpdate = this.dynamicShadowsEnabled;
```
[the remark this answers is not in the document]
robertlong: I'm going to opt for the prompt for refresh since it's simple.

Issue #2104, opened 2020-01-31
stop and refresh buttons on video streams


blairmacintyre:
**Is your feature request related to a problem? Please describe.**
Sometimes video streams break, such as on poor wifi, and the only solution is to refresh the room.  Sometimes you aren't watching a stream, or the wifi is poor and you want to chat but not have the stream consume bandwidth, and want to stop it 

**Describe the solution you'd like**
A refresh button, like on hubs' web pages.  It shuts down the stream and tries to restart it at the "live" position (i.e., refreshing doesn't try to pick up where you were, it just goes to "live")
A stop button, to just stop the stream, which just shuts down the stream.

**Describe alternatives you've considered**
none

**Additional context**
We don't want to remember when in the stream you are.  My mental model is that all live streams are "live", there is not pause/back/forward.  So, stop just kills a stream, and refresh restarts it at the "live" point.

gfodor:
One q about this is that it sounds like this is meant to rectify local issues on an individual's machine. If so, this is somewhat different than the existing 'refresh' facility on thumbnails, which is a networked, global event. I think a general "reload" facility on all media objects (separate from the existing thumbnail re-fetch concept) would fit in here, that just basically re-runs the resolution of the media on your local machine, akin to if you reloaded the room.

gfodor:
UX-wise, its a bit hard to know where such a thing could fit in. We could ofc throw it on the menu, but it could get easily confused with the rotate button if we rely upon iconography. adding a "refresh" (labelled) button in the lower row could work, but it's getting a bit crowded!

blairmacintyre:
If we just do it for streams (the critical use case, since they are the things that are fragile), we could just put it in the overlay you get on the video when you mouse over it).  So above the volume slider, beside the magic "grab an image" icon?  Add both a "stop" (which essentially just "breaks" the stream) and refresh (which does, as you describe, the same thing as when you start the room, tries to reload it from afresh)

gfodor:
ah, good point. if we want to focus exclusively on live streams for this issue (which seems relatively OK), we actually don't even show the play/pause/seek controls on those, since they're non-nonsensical, so we can just slap a refresh button in lieu of the play button and call it a day. does that work?

the other use case that came to mind was 3d model refresh (for model authors who upload changes to sketchfab), but in that case, its the web thumbnail semantics, not this one.

blairmacintyre:
[...]

PR #2733, opened 2020-07-29
Bunch of HLS fixes

### LINKED ISSUE #2572: HLS stream FxR hack makes buffering unpredictable
**Description**
While using HLS m3u8 streams (with and without adaptive bitrate indexes) users may experience stuttering / choppy video stream. My best guess is that the video player is not able to buffer properly and is forced to stay on edge of video stream (making buffering unused) because of the FxR hack in media-views.js:697

`this.video.currentTime = this.video.duration - 0.01`

Setting the value to for example 4 instead of 0.01 makes the video transitions between HLS adaptive streams much smoother as well as overall HLS performance better because I guess the player is able get some buffer time.

**To Reproduce**
1. Set a m3u8 source for a video element in Spoke
2. Some users experience stuttering
3. Change value from 0.01 to some number (depends on your parameters for HLS stream and how many fragments etc. you are buffering, setting it too high (essentially setting the video time back further than the video itself) will actually refresh the whole website from my experience)
4. Get smoother HLS

**Expected behavior**
HLS Streams in a-frame should buffer and be predictable like a normal video element.

**Hardware** 
Tested on various users on Mac OS, Windows, Linux, performance reported as better. No extensive / structured testing performed. Might need to verify root cause. We are not using VR gear.

**Additional context**

Also talked to @netpro2k about this in Discord on May 10th, in #dev-general. (Just search for "adaptive HLS seems much smoother now compared" in #dev-general discord channel to find the conversation.)

netpro2k:
A round of fixes for HLS videos. A lot of this is just removing hacks we have in place and letting HLS.js do its thing. I have still run into cases where I can get the player to stall indefinitely, but adding a manual refresh button should hopefully alleviate most of these. Probably not the last round of these but seems worth cutting a PR for.

- Update HLS.js to 0.14.6 (from 0.13.2), contains a whole bunch of bugfixes which people may or may not be hitting. Given the complexity of HLS and wide userbase of HLS.js it seems worth keeping this as up to date as we can.
- Fix handling of audio-only streams (fixes #2521)
- The XHR override we were using was running in sync mode, this was likely slowing requests down and stalling the main thread. Fixed to be async as is the default in HLS.js
- Enable refresh buttons in the pause menu for all objects. We were previously limiting this just to live twitch streams, but it feels generally useful and is a better option than page refresh. When the media is an HLS stream, uses `recoverMediaError()` in hls.js instead of a hard resolve-level refresh.
- Not in this PR but linking here since its related to this round of fixes, and should enable the client to actually do adaptive bitrate streaming when available in the source playlist https://github.com/mozilla/reticulum/pull/397
- Remove HLS timeout hacks. HLS.js has a bunch of internal logic for this already and we are just messing with it. In cases HLS.js stalls out (still can make it happen) the manual refresh above seems preferable and less likely to break what would have otherwise been a working stream. This is especially important with the fixes to adaptive bitrate streaming, since it should automatically fallback to lower bitrate streams if things are taking too long to load.
- Remove hacks we had in place for seeking ahead on live videos. This is handled internally already in HLS.js and us doing it messes with the buffering (fixes #2572)
- added `hlsDebug` query param to turn on HLS.js debugging to help diagnose future issues

### DIFF
--- a/src/components/media-loader.js
+++ b/src/components/media-loader.js
@@ -402,7 +402,7 @@ AFRAME.registerComponent("media-loader", {
       }
 
       // Some servers treat m3u8 playlists as "audio/x-mpegurl", we always want to treat them as HLS videos
-      if (contentType === "audio/x-mpegurl") {
+      if (contentType === "audio/x-mpegurl" || contentType === "audio/mpegurl") {
         contentType = "application/vnd.apple.mpegurl";
       }
 
--- a/src/components/refresh-media-button.js
+++ b/src/components/refresh-media-button.js
@@ -9,7 +9,7 @@ AFRAME.registerComponent("local-refresh-media-button", {
         if (c["media-video"] && c["media-video"].videoTexture && c["media-video"].videoTexture.hls) {
           c["media-video"].videoTexture.hls.recoverMediaError();
         } else if (c["media-loader"]) {
-          // Otherwise fall back to hard resolve-level refresh which should work for any meida
+          // Otherwise fall back to hard resolve-level refresh which should work for any media
           c["media-loader"].update(c["media-loader"].data, true);
         }
       }
--- a/src/components/refresh-media-button.js
+++ b/src/components/refresh-media-button.js
@@ -4,36 +4,23 @@ AFRAME.registerComponent("local-refresh-media-button", {
   init() {
     this.onClick = async () => {
       if (this.targetEl) {
-        this.targetEl.components["media-loader"] &&
-          this.targetEl.components["media-loader"].update(this.targetEl.components["media-loader"].data, true);
+        const c = this.targetEl.components;
+        // If an HLS stream is loaded, refresh directly with HLS.js
+        if (c["media-video"] && c["media-video"].videoTexture && c["media-video"].videoTexture.hls) {
+          c["media-video"].videoTexture.hls.recoverMediaError();
+        } else if (c["media-loader"]) {
+          // Otherwise fall back to hard resolve-level refresh which should work for any meida
+          c["media-loader"].update(c["media-loader"].data, true);
+        }
       }
     };
 
     NAF.utils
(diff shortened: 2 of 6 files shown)


src/components/media-loader.js
```diff
@@ -402,7 +402,7 @@ AFRAME.registerComponent("media-loader", {
       }
 
       // Some servers treat m3u8 playlists as "audio/x-mpegurl", we always want to treat them as HLS videos
-      if (contentType === "audio/x-mpegurl") {
+      if (contentType === "audio/x-mpegurl" || contentType === "audio/mpegurl") {
```
johnshaughnessy, line 405: Not sure, but we might want to check `"application/x-mpegurl"`, `"application/mpegurl"`, and `"application/vnd.apple.mpegurl.audio"` as per https://en.wikipedia.org/wiki/M3U#Internet_media_types

src/components/refresh-media-button.js
```diff
@@ -4,36 +4,23 @@ AFRAME.registerComponent("local-refresh-media-button", {
...
     this.onClick = async () => {
       if (this.targetEl) {
-        this.targetEl.components["media-loader"] &&
-          this.targetEl.components["media-loader"].update(this.targetEl.components["media-loader"].data, true);
+        const c = this.targetEl.components;
+        // If an HLS stream is loaded, refresh directly with HLS.js
+        if (c["media-video"] && c["media-video"].videoTexture && c["media-video"].videoTexture.hls) {
+          c["media-video"].videoTexture.hls.recoverMediaError();
+        } else if (c["media-loader"]) {
+          // Otherwise fall back to hard resolve-level refresh which should work for any meida
```
johnshaughnessy, line 12: typo "...  which should work for any ~meida~ media"

camelgod:
FYI Related issue could be referenced:
https://github.com/mozilla/hubs/issues/2104

Issue #2729, opened 2020-07-28
When user enters room, last_active_at or max_occupant_count for room is not added to database


robin-k-wilson:
**Description**
The websocket channel that handles updating the last_active_at and max_occupant_count values is not applying when users enter a room. 

**To Reproduce**
1. Run local hubs + local reticulum + database
2. Create a room 
3. Look up room in db "hubs" table
4. See that last_active_at = null and max_occupant_count = 0 ; not updated to relevant values for the hub room.
5. In another browser window, go to the same room

**Expected behavior**
Those values should be updated when a user enters a room.

robin-k-wilson:
Found the root of the problem: when we integrated dialogue (github.com/mozilla/dialog) we didn't include an update to "initialOccupants" variable that was configured on NAF janus extension. 

Used on enter event in **hubs-channel.js**: https://github.com/mozilla/hubs/blob/a0ad9529b8f149d1df51055bdbead775b95a0ff7/src/utils/hub-channel.js#L144
Deprecated naf-janus-adapter **initialOccupants**: https://github.com/mozilla/naf-janus-adapter/blob/e09ffae1afa1e798e7b1f79ac58d8436d3feb1f6/src/index.js#L251
And incremented occupants: https://github.com/mozilla/naf-janus-adapter/blob/e09ffae1afa1e798e7b1f79ac58d8436d3feb1f6/src/index.js#L322

InfiniteLee:
https://github.com/mozilla/hubs/blob/5ed0b5c27477bf7a93d9c09158fcb72fd5a217f2/src/utils/hub-channel.js#L147
should be
`initialOccupantCount = Object.keys(NAF.connection.adapter.occupants).length;`

PR #2742, opened 2020-07-30
Fix missing occupant data

robin-k-wilson:
With the update from janus to media-soup the initial occupant count was referenced incorrectly. Rather than .publisher we look at the occupants object.

Occupants is defined in naf-dialog-adapter.js.


robin-k-wilson:
Fixes:
https://github.com/mozilla/hubs/issues/2729

PR #2760, opened 2020-07-31
Medium Quality Mode

### LINKED ISSUE #2541: Avatar badly metallic in preview window
**Description**
Avatar badly metallic in preview window
UnLit, cartoon style avatar, same on mobile.

In the hubs scene, it works fine but in the preview window, it has completely different look. 
It seems a bug in the preview window but it can also be fixed by documentation or good example.
Tell me how to fix it!

My discussion and testing
https://discordapp.com/channels/498741086295031808/715710184575598655/717589055138955307
Same issue by own avatar
https://discordapp.com/channels/498741086295031808/715710184575598655/716447343742025759

The preview is badly metallic! The shader on Blender is 
Texture {Linear, Flat, Repeat, Single Image} -> Background {Color} -> Material Output {Surface}
It is similar shader for UnLit characters.

Here is my test blend and GLB with texture files. Thanks to find any solutions!
https://www.dropbox.com/s/5flnzn4n7wzf9tn/TestAvater20200603.zip?dl=0"

**To Reproduce**
1. Go to 'Set Name & Avatar'
2. Click on 'Browse Avatar'-> 'Create Avatar'
3. Upload own GLB (here is my testing file https://www.dropbox.com/s/5flnzn4n7wzf9tn/TestAvater20200603.zip?dl=0 )
4. See metallic avatar (but the unlit shader works fine in the room)

**Expected behavior**
Preview window must be same with shader in the room.

**Screenshots**
Issue-Metallic-Preview3
Issue-Metallic-Preview
Issue-Metallic-Preview2

**Hardware** 
(tested by myself)
 - Device: 　Desktop, iPhone
 - OS: Windows, iOS
 - Browser: Chrome on Windows, Safari on iOS

**Additional context**
This model works fine on iOS and Desktop. We need to know how to realize cartoon style in pipeline through Blender->Export->GLB+Texture->Hubs.

Issue-Metallic-Preview-correct

Thanks in advance! Akihiko

robertlong:
This PR introduces another material quality mode to Hubs targeted at budget desktops/laptops and mobile devices.  The medium quality material mode uses a non-PBR Blinn-Phong lighting model. This is quite a bit cheaper than the physically based standard material that we use on high quality mode which should help on integrated graphics cards and mobile GPUs.

This MeshPhongMaterial has been modified so that it can be used for existing glTF PBR assets and the materials are converted from PBR to Phong at runtime. This modification allows the original MeshPhongMaterial to support glTF's [tangent space normal maps](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#materialnormaltexture) instead of the default, object space normal maps.

This PR also moves the MobileStandardMaterial used in the low quality mode over to the MeshBasicMaterial with a small modification to add lighting from the emissive map.

Here is a list of supported features for each material:

### High Quality
- ThreeJS [MeshStandardMaterial](https://threejs.org/docs/index.html#api/en/materials/MeshStandardMaterial)
- All of the PBR features from the glTF Spec including the following textures:
  - Base Color Map
  - Normal Map
  - Ambient Occlusion Map
  - Metallic Roughness Map
  - Emissive Map
  - Light Map
- Image based environment map generated at runtime from the skybox
- Realtime per-pixel lighting

### Medium Quality
- ThreeJS [MeshPhongMaterial](https://threejs.org/docs/index.html#api/en/materials/MeshPhongMaterial)
- The following textures:
  - Base Color Map
  - Normal Map
  - Ambient Occlusion Map
  - Emissive Map
  - Light Map
- The material has a constant reflectivity value of 0.5 and does not use the metallic roughness map for specularity
- Spherical Harmonics based global illumination generated at runtime from the skybox. This uses a single light probe placed at the origin and only samples from the skybox. This achieves a similar effect to the environment map generated from the skybox. Given reflectivity is constant, the environment map looked strange as it was reflected by every surface. Using the spherical harmonics from the light probe gives us a much more subtle effect for global illumination.
- Realtime per-pixel lighting

### Low Quality
- ThreeJS [MeshBasicMaterial](https://threejs.org/docs/index.html#api/en/materials/MeshBasicMaterial)
- The following textures:
  - Base Color Map
  - Ambient Occlusion Map
  - Emissive Map
  - Light Map
- This material is unlit and does not use any lighting from the environment.

## Further optimizations and improvements:

While the medium quality mode does save some GPU resources and looks somewhat similar to the high quality material, we still have some quality improvements that could be made. We could generate a specular map at runtime using the metallic roughness map and then use the image based environment map.

For the low quality mode our only option for increased quality is light baking. We can do this offline, as is currently possible with our MOZ_lightmap glTF extension and Blender workflow, we could do light bake in Spoke at publish time, or we could bake lights at runtime as is done by [PlayCanvas](https://developer.playcanvas.com/en/user-manual/graphics/lighting/runtime-lightmaps/).

Currently both medium and high quality mode render all of the lights in the scene individually. This is very computationally expensive. We could limit users to a single light, implement forward clustered rendering, or bake our lighting data. Forward clustered rendering would reduce the per-light cost of rendering and allow us to support more realtime lights as seen [here](https://github.com/AmanSachan1/WebGL-Clustered-Deferred-Forward-Plus). I think this makes a lot of sense for future improvements to our high quality mode. However, for medium and low quality even forward clustered lighting may be too expensive. Instead, I think we should look into baking our lighting data. We can do this with and/or without baked specularity, perhaps the difference is enough to make baked specularity the new medium quality mode and low quality baked lighting without specularity. We will have to do more research into these solutions.

Fixes #2541


joshmarinacci:
exciting!

src/components/uv-scroll.js
```diff
@@ -13,19 +13,6 @@ export class UVScrollSystem {
       const increment = component.data.increment;
       map.offset.x = increment.x ? offset.x - (offset.x % increment.x) : offset.x;
       map.offset.y = increment.y ? offset.y - (offset.y % increment.y) : offset.y;
-
```
netpro2k, line 16: Yay, this was super ugly, glad to see it go away. Worth just quickly double checking it still works in medium and low quality modes (it should since these are now all just using default ThreeJS materials which is great).

src/react-components/avatar-preview.js
```diff
@@ -218,8 +218,7 @@ class AvatarPreview extends Component {
   loadPreviewAvatar = async avatarGltfUrl => {
     let gltf;
     try {
-      const technique = window.APP.quality === "low" ? "KHR_materials_unlit" : "pbrMetallicRoughness";
-      gltf = await loadGLTF(avatarGltfUrl, "model/gltf", technique, null, ensureAvatarMaterial);
+      gltf = await loadGLTF(avatarGltfUrl, "model/gltf", null, ensureAvatarMaterial);
```
netpro2k, line 221: Hmm is this going to result in the avatar preview always using "high quality" materials?

src/utils/material-utils.js
```diff
@@ -17,3 +17,144 @@ export function mapMaterials(object3D, fn) {
...
+
+    material.emissive.copy(source.emissive);
+    material.emissiveMap = source.emissiveMap;
+    material.emissiveIntensity = source.emissiveIntensity;
+
+    material.normalMapType = source.normalMapType;
+    material.normalMap = source.normalMap;
+    material.normalScale.copy(source.normalScale);
+
+    material.onBeforeCompile = shader => {
```
netpro2k, line 49: A quick comment explaining why we are patching the shader would be helpful here I think.

src/utils/material-utils.js
```diff
@@ -17,3 +17,144 @@ export function mapMaterials(object3D, fn) {
...
+
+    return material;
+  } else if (quality === "low") {
+    const material = new THREE.MeshBasicMaterial();
+
+    THREE.Material.prototype.copy.call(material, source);
+
+    material.color.copy(source.color);
+
+    material.onBeforeCompile = shader => {
```
netpro2k, line 115: Same here, though this one is more obvious.

src/utils/media-utils.js
```diff
@@ -284,9 +284,13 @@ export function injectCustomShaderChunks(obj) {
 
       const newMaterial = material.clone();
       // This will not run if the object is never rendered unbatched, since its unbatched shader will never be compiled
-      newMaterial.onBeforeCompile = shader => {
+      newMaterial.onBeforeCompile = (shader, renderer) => {
         if (!vertexRegex.test(shader.vertexShader)) return;
 
+        if (material.onBeforeCompile) {
+          material.onBeforeCompile(shader, renderer);
```
netpro2k, line 291: Ah good, was wondering how the double shader patching was going to work, this is pretty clean.

robertlong:
So I had to move back to separate classes for our patch to MeshBasicMaterial. We change the emissive map in the avatar preview page and you couldn't change it any other way. To keep things consistent I also did the same for MeshPhongMaterial. IMO, this is slightly better anyway because `onBeforeCompile` now no longer needs to be manually copied if you were to clone these materials.

Also I fixed up a bunch of bugs in the avatar preview page when viewing it in low/medium quality. The other materials are sometimes missing environment, normal, metalness, and roughness maps so I added checks for quality modes.

kearwood:
This all seems reasonable; however, it seems that even the high quality mode should be performant on mobile or integrated gpus.

Do we have specific hardware targets for reference?  Have we measured the performance of each setting to see if they spread far enough?

I suspect that the performance savings are less likely due to the cost of fragment shader performance, but rather due to the sampler cache misses. Mip-mapping and texture compression may allow you to keep more channels enabled.

kearwood:
I will do some GPU-level profiling and see what is happening.

robertlong:
That sounds awesome @kearwood! Yes, I think texture compression could help a lot. I wasn't able to devise a test for sampler cache misses. Is there a straightforward way of doing that? I guess just test the perf difference with compressed textures? @netpro2k and I couldn't see much benefit in our tests with Basis a while back. We saved a bunch of VRAM, but framerate stayed the same. I probably didn't devise a good enough benchmark for this though. Perhaps we can create a scene that will let us benchmark fragment cost vs sampler cache misses? And we can always adjust resolution to determine vertex vs fragment bottlenecks.

Issue #2738, opened 2020-07-30
Single group meeting audio settings


takahirox:
This issue is top of #2662 and #2737.

If I'm right, I think Hubs is primary designed for 1. multiple groups are in a room and 2. having conversation in each group simultaneously. Audio setting is based on it, then it has positional audio and distance based audio attenuation because one doesn't want to hear other group's or far people's conversation.

Especially recent tough situations, people are more interested in remote conversation and the use cases of Hubs would be getting wider.

One of them would be Single group meeting, as we Hubs teem weekly does, where basically only one group is in a room, basically only one person speaks, and everyone wants to hear the speaker. Hubs can be used for it (as Zoom or other tools alternate).

In Single group meeting, we may need other audio settings. They are

1. No positional audio
2. No distance based audio attenuation
3. Auto members' voice volume normalization (keeping their volumes in a certain range)

For 1, we already have `/audiomode` command to enable/disable positional audio.

For 2, currently we have no way to disable distance based audio attenuation (even with disabled positional audio) #2737

For 3, currently we can manually control each person's voice volume but can't automatically do #2662

**Describe the solution you'd like**
I'd like to have options which satisfy 2 and 3 for Single group meeting. They may be an option disabling distance based audio attenuation and an option automatically normalizing all members' voice volumes. (The latter could cover the former?)

**Additional Context**
* Any other audio setting we should have for Single group meeting?
* Let me know if you think Single group meeting should not be one of our targets

deanmax:
Maybe this is not 100% related, but I wonder if, as a room owner, that I want to announce something to the whole room, would there be some flag/mode available that I can make everyone in the room hear me?

robertlong:
I think we should really think a lot about the design of these things. The first pass at #2 with the `/rolloff` command seems like a ok interim solution, but that and the `/audiomode` command do not strike me as good user-oriented design.

Here are some ideas of things we could do:

### 1. Temporary User Audio Settings
Adjusting audio settings to disable/adjust audio spatialization in a single room is somewhat heavy-handed. If a user turns off audio spatialization or turns up audio rolloff for one room, they may forget about it in their next visit to Hubs and not know what's happening. Perhaps there's some additional UI that we could add to the HUD, separate from the preference pane that had these temporary audio settings. I could imagine a checkbox for disabling positional audio and a slider for adjusting audio rolloff with visual feedback in the room to show you where you will stop hearing media and users. The next time you visit Hubs these settings will be reset.

[...]

PR #2766, opened 2020-08-01
Introduce audioRolloff command to chat box to set global audio rolloff factor

### LINKED ISSUE #2737: A member's voice volume can be bigger/smaller if I move closer to/farther from the one even if I disable positional audio
**Description**
A member's voice volume can be bigger/smaller if I move closer to/farther from the one even if I disable positional audio with `/audiomode` command.

**To Reproduce**
1. Start local server with `npm run dev`
2. Create and enter a room `https://localhost:8080`
3. `cd scripts/bot && node run-bot.js -r room_id -a bot-recording.mp3`
4. Move closer to/farther from the bot avatar and confirm that the bot's audio volume can be bigger/smaller

**Expected behavior**
If positional audio is disabled, we shouldn't see audio attenuation effect (correct? /cc @InfiniteLee)

**Hardware** 
 - Device: Desktop
 - OS: Windows
 - Browser: Chrome

Maybe non platform specific issue

**Additional context**
I speculate the root issue is here

https://github.com/mozilla/hubs/blob/master/src/components/avatar-volume-controls.js#L56-L61

```javascript
if (audioOutputMode === "audio") {
  this.avatarAudioSource.el.object3D.getWorldPosition(positionA);
  this.el.sceneEl.camera.getWorldPosition(positionB);
  const squaredDistance = positionA.distanceToSquared(positionB);
  gain = gain * Math.min(1, 10 / Math.max(1, squaredDistance));
}
```

It doesn't seem to check if positional audio is enabled to apply audio attenuation.

I want to try to fix this issue.

takahirox:
Resolves #2737

This PR introduces `/audioRolloff <factor>` command to chat box which controls rolloff/attenuation of audio. This is a function which is sort of similar to `/audiomode` command, the both have an effect to position based audio. So I added it as command too, not as preference menu.

`factor` is stored as `globalRolloffFactor` and applied to the `rolloffFactor` attributes of `media-audio` and `avatar-audio-source` as `rolloffFactor: settings.[media|avatar]RolloffFactor * globalRolloffFactor` for positional audio.

And it's also applied to non-positional audio as `finalAttenuation = globalRolloffFactor * distanceBasedAttenuation + (1.0 - globalRolloffFactor)`, assuming both `globalRolloffFactor` and `distanceBasedAttenuation` are in [0.0, 1.0] range.

The use case and motivation of introducing this command is in #2738

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-829)


src/components/media-views.js
```diff
@@ -943,12 +943,15 @@ AFRAME.registerComponent("media-video", {
           this.el.object3D.getWorldPosition(positionA);
           this.el.sceneEl.camera.getWorldPosition(positionB);
           const distance = positionA.distanceTo(positionB);
+          const globalRolloffFactor = window.APP.store.state.preferences.globalRolloffFactor;
```
fernandojsg, line 946: Instead of assuming that `globalRolloffFactor` is between 1 and 0, I'd just do a clamp on it

robertlong:
Maybe this should be `/audiorolloff` to provide more context to the user?

src/components/media-views.js
takahirox: Thanks, I did it.

takahirox:
I like that, renamed to `/audioRolloff`.

takahirox:
How do you Hubs folks test and write test when you add a new feature? I checked `test` directory but there is only one command-mapping test. And smoking test may not be for testing a new feature.

I manually tested that `/audioRolloff` command is accepted and confirmed rolloff factor seems to work with a bot speaking by moving closer to/farther from it.

robertlong:
Yup, manual tests are all we've got for this sort of stuff :man_shrugging:

takahirox:
I see, thanks. So I think it would be ready to go if looks ok.

src/components/media-views.js
johnshaughnessy: Preferences can be `undefined` if a user hasn't specified a pref. You should set a default in this case with something like:
```js
const pref = window.APP.store.state.preferences.globalRolloffFactor
const globalRolloffFactor = pref === undefined ? 1 : pref;
```
I see below that you are setting the preference when the audio system starts up, but it would be bad to rely on initialization order to prevent this bug. If we elevate this into the preference menu (instead of having it be a slash command) then we will also remove the code that resets it on startup. (https://github.com/mozilla/hubs/pull/2766/files#diff-02e07dc2b8c1d9775319e5e3806b1aaeR55) 

We do this for example here: https://github.com/mozilla/hubs/blob/24d7e9327483531b22883f88206459614207cd55/src/components/media-views.js#L474

src/components/avatar-volume-controls.js
```diff
@@ -50,14 +50,17 @@ AFRAME.registerComponent("avatar-volume-controls", {
...
       const volumeModifier = (globalVoiceVolume !== undefined ? globalVoiceVolume : 100) / 100;
       let gain = volumeModifier * this.data.volume;
       if (audioOutputMode === "audio") {
         this.avatarAudioSource.el.object3D.getWorldPosition(positionA);
         this.el.sceneEl.camera.getWorldPosition(positionB);
         const squaredDistance = positionA.distanceToSquared(positionB);
-        gain = gain * Math.min(1, 10 / Math.max(1, squaredDistance));
+        const distanceBasedAttenuation = Math.min(1, 10 / Math.max(1, squaredDistance));
+        // Clamp globalRolloffFactor to [0.0, 1.0]
+        gain = gain * (1.0 + Math.max(0.0, Math.min(1.0, globalRolloffFactor)) *
```
johnshaughnessy, line 62: Need to guard against `globalRolloffFactor` being `undefined`. If it is, `Math.min(1.0, undefined)` is `NaN`

src/message-dispatch.js
```diff
@@ -150,6 +150,21 @@ export default class MessageDispatch {
...
+          if (!isNaN(factor)) {
+            window.APP.store.update({
+              // Factor must be equal to or greater than 0.0
+              preferences: { globalRolloffFactor: Math.max(0.0, factor) }
+            });
+          } else {
+            this.log("audioRolloff command needs a valid number parameter.");
+          }
+        } else {
+          this.log("audioRolloff command needs one parameter.");
```
johnshaughnessy, line 165: Might want to specify the range - (`"audioRolloff command needs a number between 0 [no rolloff] and 1 [normal rolloff]"`)

src/components/media-views.js
johnshaughnessy: Sorry to un-resolve the above conversation with a different concern related to this same line. I was just worried you wouldn't have seen it otherwise.

takahirox:
@johnshaughnessy Thanks for the review! Yes, I wrote the code relying on default value is set at https://github.com/mozilla/hubs/pull/2766/files#diff-02e07dc2b8c1d9775319e5e3806b1aaeR55 but I understand we shouldn't rely on it. I'll update.

takahirox:
Updated

takahirox:
Implementation itself is already done. But before merging I want to

* wait for new US design
* discuss how to expose (Behind chat command or preference? Should the area where other people can hear be visible?)
* have a meeting to discuss entire Hubs Audio problems

PR #2782, opened 2020-08-04
Fix signin dialog bug

takahirox:
This PR fixes signin dialog bug that the input email is not displayed but `{email}` is displayed. The root issue is `<FormattedMessage>` takes `values` but wrongly `value` (dropping s) is set.

image


takahirox:
Thanks for the review!

PR #2801, opened 2020-08-05
Add volume option to run-bot.js

### LINKED ISSUE #2793: Add volume option to scripts/bot/run_bot.js for audio volume control

I'm working on #2662 and want to test with bots speaking. `scripts/bot/run_bot.js` supports `audio` option which lets a bot loads audio file and speaks.

I want multiple bots in a room and want some of them to speak loudly and the others to speak softly. But `scripts/bot/run_bot.js` doesn't provide a way to control audio volume.

I'd like to suggest to add `-v --volume` option to `scripts/bot/run_bot.js` to control audio volume. With the new option you can place speaking bots in a room with different audio volumes.

```sh
$ node run-bot.js -r roomid -a bot-recording.mp3 -v 0.2
$ node run-bot.js -r roomid -a bot-recording.mp3 -v 0.5
$ node run-bot.js -r roomid -a bot-recording.mp3 -v 1.0
```

**Describe alternatives you've considered**

Having different volume level audio files may be another option to satisfy my demand but it would be a bit too annoying. And letting `scripts/bot/run_bot.js` support `volume` option would be more useful and it can be even used for other audio related tests.

takahirox:
This PR resolves #2793 by adding `-v --volume` option to `run-bot.js` which controls audio volume level of bot.

```
$ node run-bot.js -r roomid -a bot-recording.mp3 -v 0.2
```

Some notes:
* I insert GainNode to control the volume instead of setting `audio` element volume because "the audio level or volume of the media element does not affect the volume of captured audio"

https://www.w3.org/TR/mediacapture-fromelement/#html-media-element-media-capture-extensions

* I pass `audio_volume` parameter to `scene-entry-manager.js` by adding `audio_volume` in URL like `https://localhost:8080/hub.html?allow_multi=true&audio_volume=2.0&bot=true`. Let me know if there is a better way
* `-v --volume` option has an effect only if `-a --audio` option is set

### DIFF
--- a/scripts/bot/run-bot.js
+++ b/scripts/bot/run-bot.js
@@ -3,12 +3,13 @@ const doc = `
 Usage:
     ./run-bot.js [options]
 Options:
-    -h --help         Show this screen
-    -u --url=<url>    URL
-    -o --host=<host>  Hubs host if URL is not specified [default: localhost:8080]
-    -r --room=<room>  Room id
-    -a --audio=<file> File to replay for the bot's outgoing audio
-    -d --data=<file>  File to replay for the bot's data channel
+    -h --help            Show this screen
+    -u --url=<url>       URL
+    -o --host=<host>     Hubs host if URL is not specified [default: localhost:8080]
+    -r --room=<room>     Room id
+    -a --audio=<file>    File to replay for the bot's outgoing audio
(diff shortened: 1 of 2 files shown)


brianpeiris (approved):
LGTM!

PR #2837, opened 2020-08-10
Add support to specify a spawn point name on the url with an anchor hash

### LINKED ISSUE #2833: Specify anchor link for spawn points
It could be nice to be able to define a preferred spawn point to start for multiple reasons, one of my current use cases is to be able to run benchmarks and testing consistently so the user appears always at the same spot.

I have a working solution by implementing a `occupySpawnPoint(waypoint)` to be called before the default `tryToOcuppyAnyOf` (https://github.com/mozilla/hubs/blob/master/src/systems/waypoint-system.js#L336) if we provide a hash tag and it can be dynamically detected so we can change it at runtime too.
If the spawn point name is not found, it will just follow the standard flow and will assign a random point.

Here is a video testing it:
https://streamable.com/taxasv

Partially discussed too in https://github.com/mozilla/hubs/issues/2667

fernandojsg:
It adds support to define the spawn point where you want to start, but it also lets you change it dynamically without refreshing the page.
The format is `url#[spawn_point_name]` for example in the default room you could try:
```
https://localhost:8080/hub.html?hub_id=XXX#Spawn_Point
https://localhost:8080/hub.html?hub_id=XXX#Spawn_Point_1
https://localhost:8080/hub.html?hub_id=XXX#Spawn_Point_2
https://localhost:8080/hub.html?hub_id=XXX#Spawn_Point_3
```

If the name of the spawn point doesn't exist it will just follow the standard flow and it will spawn you at a random point.

Fixes #2833


src/systems/waypoint-system.js
```diff
@@ -323,7 +329,24 @@ export class WaypointSystem {
       this.waitOneTick = false;
       return;
     }
+
+    const hashUpdated = window.location.hash !== "" && this.previousWaypointHash !== window.location.hash;
+
+    if (hashUpdated && this.initialSpawnHappened) {
```
fernandojsg, line 335: I introduced `initialSpawnHappened` as once you enter the scene, async, `moveToSpawnPoint` will get called and it will override the hash location that will be set otherwise on the first tick after detecting the new hash

src/systems/waypoint-system.js
```diff
@@ -304,17 +307,20 @@ export class WaypointSystem {
     }
     return this.nextMoveToSpawn;
   }
+  occupySpawnPoint(waypointComponent) {
```
johnshaughnessy, line 310: A name like `moveToWaypoint` matches the functionality better than `occupySpawnPoint`. The `waypointComponent` does not become occupied nor is anything checking `canBeSpawnPoint`.

johnshaughnessy: Also would be good to call this new function from here instead of having duplicate lines of code:
https://github.com/mozilla/hubs/blob/40173b2e6a90ed7825bcfa0e6f038b384fbf07c7/src/systems/waypoint-system.js#L149

This spot is very similar, except that in this case we do not want to `releaseAnyOccupiedWaypoints` (because we actually want to occupy one of them:
https://github.com/mozilla/hubs/blob/40173b2e6a90ed7825bcfa0e6f038b384fbf07c7/src/systems/waypoint-system.js#L164
So, 3 of the 4 lines are the same. Might be a good idea to have `releaseAnyOccupiedWaypoints` happen at the call site and then use `moveToWaypoint` in all three places (and not have `moveToWaypoint` release anything).

src/systems/waypoint-system.js
```diff
@@ -323,7 +329,24 @@ export class WaypointSystem {
       this.waitOneTick = false;
       return;
     }
+
+    const hashUpdated = window.location.hash !== "" && this.previousWaypointHash !== window.location.hash;
+
+    if (hashUpdated && this.initialSpawnHappened) {
+      this.previousWaypointHash = window.location.hash;
+      const spawnPointName = window.location.hash.replace("#", "");
+      const spawnPoint = this.ready.find(c => c.el.className === spawnPointName);
```
johnshaughnessy, line 338: `spawnPoint`/`spawnPointName` -> `waypoint`/`waypointName`

src/systems/waypoint-system.js
```diff
@@ -323,7 +329,24 @@ export class WaypointSystem {
...
+      if (spawnPoint) {
+        this.occupySpawnPoint(spawnPoint);
+        this.currentMoveToSpawn = null;
+        this.currentMoveToSpawnResolve = null;
+        this.nextMoveToSpawn = null;
+        this.nextMoveToSpawnResolve = null;
```
johnshaughnessy, line 344: It used to be that we had code that looked like,
```js
waypointSystem.moveToSpawnPoint().then(()=> { // do something after spawning })
```
We don't do that anymore (https://github.com/mozilla/hubs//blob/master/src/hub.js#L440), but still I'm nervous that `null`ifying these book-keeping variables may cause unintended consequences down the road if ever that changes. I don't think you actually need to modify these. 

If you want to make sure to avoid conflicts with a spawn-flow-in-progress (which happens over several frames), maybe add `this.currentMoveToSpawn === null && this.nextMoveToSpawn === null` to the check above: 
```js
if (hashUpdated && this.initialSpawnHappened && this.currentMoveToSpawn === null && this.nextMoveToSpawn === null)
```

src/systems/waypoint-system.js
```diff
@@ -323,7 +329,24 @@ export class WaypointSystem {
...
+      }
+    }
+
     if (!this.currentMoveToSpawn && this.nextMoveToSpawn) {
+      this.initialSpawnHappened = true;
```
johnshaughnessy, line 349: This is when the initial spawn begins, but it might finish on a later frame. Here might be a more appropriate place: https://github.com/mozilla/hubs/blob/40173b2e6a90ed7825bcfa0e6f038b384fbf07c7/src/systems/waypoint-system.js#L375

src/systems/waypoint-system.js
johnshaughnessy: I wonder whether there's any difference between `c.el.object3D.name` and `c.el.className`. I think the `object3D.name` matches what's in spoke and the gltf -- I didn't know/realize that the entity's `className` matched this as well. Not sure if that's always the case (e.g. if something else modifies an entity's `classList`)

src/systems/waypoint-system.js
fernandojsg: Addressed that and refactored both functions as they are pretty much the same just changing the instant, that now is a parameter to `moveToWaypoint`

src/systems/waypoint-system.js
fernandojsg: > `spawnPoint`/`spawnPointName` -> `waypoint`/`waypointName`

Renamed, makes more sense indeed, thanks!

> I wonder whether there's any difference between `c.el.object3D.name` and `c.el.className`. I think the `object3D.name` matches what's in spoke and the gltf -- I didn't know/realize that the entity's `className` matched this as well. Not sure if that's always the case (e.g. if something else modifies an entity's `classList`)

Yeah I was surprised also that it matches so I thought it was something specific from hubs. In any case I changed it to use `object3D.name` which makes feels more safe.

src/systems/waypoint-system.js
fernandojsg: With the current flow it won't conflict as the initial spawn will happen and I'll just resolve the hashtag later. I have removed all the nulls anyway as they were not affecting it

src/systems/waypoint-system.js
fernandojsg: Done! thanks

fernandojsg:
@johnshaughnessy thanks for the comments, addressed the issues, ready for another shot :)

Issue #1821, opened 2019-10-15
Improve Moderation Tools with Room Lock / Password Feature 


misslivirose:
### **Problem Statement**
The existing Hubs infrastructure has two different design mechanics related to securing rooms. At the core, room access is restricted to users who have a permalink to a room. If that room is bridged to a secondary authentication provider (currently, the only option for this is through Discord), then the user must also have appropriate access to the secondary provider in addition to having the room link. 

**Example Use Cases**
There is an active community of Hubs users who communicate with the team via Twitter, but do not use Discord as a communication tool. These users are unable to attend meetups hosted by the Hubs team because they do not want to create an additional account and sign up for a new service. The meetups use the secondary authentication provider after open publishing of the links resulted in users behaving inappropriately and harassing others. 

A meetup group posts links to their Hubs room on Twitter. They do not require any kind of secondary authentication to participate. They frequently have larger crowds of users than come to the Hubs meetup despite an overall smaller community size, but have no mechanic available to them to effectively remove a bad actor while keeping the meetup going for the welcomed community members. 

### **“Lock” Feature**
The lock feature would allow a room owner or moderator to ‘freeze’ a room’s attendees and place a password onto the room after kicking out a user. When new users attempt to return to the room via their knowledge of the link they previously used, instead of being granted access immediately, they would be prompted with a password screen instead of the lobby. Users inside the room would be able to see the current password and the room owner could send this alongside the link to allow new users in while preventing others from distributing the link. 

**MVP:**
A room owner can set a password on their room that must be used in conjunction with the room link. This can be changed as part of the ‘Room Settings’. The password can be removed and a new one can be set. 

- **Enhancement 1:** When kicking a user from the room, the room owner is informed of the ability to set a password on the room. There is some kind of affordance that they can select to not show the message again. 

- **Enhancement 2:** The room owner can choose whether or not the password is shared with people who are currently in the room. If they enable password sharing, all participants in the room can view the password. 

**Considerations**
We should think about how this would interact with the contract that is currently managed by bridging with Discord-bound rooms. It may be more desirable to prevent the locking feature from being used with Discord altogether. 

### **Password Feature** 
When a room is “locked”, the room owner will be able to specify if users who have the link to the room can “knock” to enter the room through this feature. This will give the room owner and moderators a queue or peek into who is requesting access to a locked room. The room owner can then choose to allow access to the room based on the information that is given to them by the user. 

**Considerations**
[...]

PR #2838, opened 2020-08-10
Basic Hub Invites

brianpeiris:
You can now set a room to "invite only", which generates a unique invite link that can be revoked to prevent future access.

image

Implements a basic version of room invite links. A room can now be set to "Invite only", which then generates an invite link that allows room access. Invite links look like `https://hubs.mozilla.com/BJtpjmt/hopeful-realm/?hub_invite_id=pKuResV`. The invite link can be revoked, or the room can be put back into "Private" mode. In this initial version, users must use the invite link to access the room. No state is stored against a user's account when they use the link. If a user attempts to enter a room without an invite id, they are denied access and shown our typical black room entry denial screen.

For #1821 

Goes with https://github.com/mozilla/reticulum/pull/406


src/components/in-world-hud.js
```diff
@@ -53,12 +53,20 @@ AFRAME.registerComponent("in-world-hud", {
     this.onInviteClick = () => {
       this.el.emit("action_invite");
     };
+
+    this.onHubUpdated = () => {
+      if (window.APP.hub) {
```
netpro2k, line 58: TIL we had a 3d invite button. Also I get why but this line reads as backwards if you don't realize we are now mixing what "invite" means.

src/hub.js
```diff
@@ -665,7 +665,9 @@ function handleHubChannelJoined(entryManager, hubChannel, messageDispatch, data)
         });
     };
 
+    window.APP.hub = hub;
     updateUIForHub(hub, hubChannel);
+    scene.emit("hub_updated");
```
netpro2k, line 670: Any reason not to just pass the hub here instead of storing the state globally? Alternatively just roll this as another thing that happens in `updateUIForHub` and skip the event alltogether.

src/react-components/ui-root.js
```diff
@@ -1730,15 +1726,18 @@ class UIRoot extends Component {
...
                     description: this.props.hub.description,
                     member_permissions: this.props.hub.member_permissions,
                     room_size: this.props.hub.room_size,
-                    allow_promotion: this.props.hub.allow_promotion
+                    allow_promotion: this.props.hub.allow_promotion,
+                    entry_mode: this.props.hub.entry_mode
                   },
-                  onChange: settings => this.props.hubChannel.updateHub(settings)
+                  onChange: settings => this.props.hubChannel.updateHub(settings),
+                  onFetchInvite: () => this.props.hubChannel.fetchInvite(),
```
netpro2k, line 1739: I might just pass in the hubChannel instead and then just call these directly. As implemented when I was reading the component code I was a bit confused since I thought these were callbacks that would trigger when these things happened.

src/utils/hub-channel.js
```diff
@@ -223,6 +223,18 @@ export default class HubChannel extends EventTarget {
     this.channel.push("update_hub", settings);
   };
 
+  fetchInvite = () => {
+    if (!this._permissions.update_hub) return "unauthorized";
```
netpro2k, line 227: Looks like we are doing it in all the other calls too but it was a bit surprising that we do this check locally. Is this just an optimization? 

Also extra strange that as a result these functions either return a string or a Promise. I suspect most of the places we are calling these would actually blow up on the string case. Maybe we should start changing these to `Promise.reject("unauthorized")` instead?

src/hub.js
brianpeiris: I primarily wanted to be able to access the hub object in the `tips` system, which of course is an a-frame system, not a react component. I'm pretty sure this is the first time we've had to access `hub` in a-frame code. All the rest of the react code passes `hub` data down through props.
If you have a better suggestion for how to get it to the a-frame system, I'd like to use it. I agree the global is not great.

src/react-components/ui-root.js
brianpeiris: I don't like the way we name "callbacks" in react either. We've made it a convention to do things like `onChange` as well. Are you suggestion we reverse that convention?

src/utils/hub-channel.js
brianpeiris: Yeah, mostly just cargo-culting here. I think I'll just remove the early return, since the server does actually respond appropriately if the user doesn't have permissions.

vjFred:
@brianpeiris hello is this already implemented on hubs cloud?

src/components/in-world-hud.js
brianpeiris: Decided not to refactor the existing code to rename the existing feature to "share", since it does look like we are going to rename things to "invite" in the new redesign. We'll have to revisit how these things work together at that time.

src/hub.js
brianpeiris: Had to leave the global in for the `tips` system, but I did pass the hub along with the event instead.

PR #2876, opened 2020-08-17
Room UI Redesign: Storybook JS and Button

robertlong:
We announced recently that we're starting a redesign of Hubs. Starting with the in-room UI. In order to ease this transition we will be using Storybook JS to gradually implement our new design system.

This PR implements the first component, `Button` and its variants. I moved the existing button component used on the home page to the `react-components/home` folder to make way for this new Button component. In the process of setting up Storybook I also added a css-reset that is based off of the one used in Tailwind CSS. This lives in `react-components/styles.global.scss`. Our new color theme lives in `react-components/styles/theme.scss`, before we roll this out to everyone, I will ensure that this color theme can be configured in the admin panel and injected via CSS variables.

With this new design, I'm putting a focus on designing things with accessibility up front. We do an ok job at accessibility right now, but there's a lot of room for improvement. With this Button component, I wanted to ensure that keyboard users will get the correct focused element outline when using tab. I also wanted to hide that outline for mouse users. I did some research and found that a common workaround is to disable outlines in the page until the tab button is pressed. As seen in this [codepen](https://codepen.io/stowball/pen/LNZyrK).

To start Storybook, run `npm run storybook` this will start the Storybook dev server. You can define stories by creating a file with the `.stories.js` extension. More info on writing stories can be found [here](https://storybook.js.org/docs/react/writing-stories/introduction).

You'll notice I added:
```
Basic.parameters = {
  design: {
    type: "figma",
    url: "https://www.figma.com/file/Xag5qaEgYs3KzXvoxx5m8y/Hubs-Redesign?node-id=70%3A2186"
  }
};
```

That design parameter can be used to show a Figma design alongside the component in Storybook. This is really helpful when working on your design and also for design verification. To get a url for a node, you can select a frame/component in Figma, click the share button in the top right corner, and then click copy link.

Some build system specific things were needed to get Storybook running in our project. The first was installing core-js. Storybook uses it's own babel config, and while it does inherit our own, it is opinionated about using core-js vs the polyfill in babel-env. It also uses `loose` mode for some JS syntax features so I needed to also use it for `@babel/proposal-class-properties`. This should be just fine for us and we may even see some performance benefits. See [here](https://babeljs.io/docs/en/babel-plugin-proposal-class-properties#loose) for more details on that.


src/react-components/input/useAccessibleOutlineStyle.js
```diff
@@ -0,0 +1,29 @@
+import { useEffect, useCallback } from "react";
+
+// The page is styled by default not to have any outlines on selected input elements.
+// When the user presses the tab key the keyboardUserClass is added to the body and all outlines are re-enabled.
+// This lets us provide a clean look for mouse users, while still maintaining accessibility for keyboard users.
+export function useAccessibleOutlineStyle(keyboardUserClass = "keyboard-user") {
```
robertlong, line 6: Just realized I accidentally removed the `keyboard-user` class

PR #2879, opened 2020-08-17
Improved i18n

InfiniteLee:
- Refactors i18n code to support dynamically loading locale files (so all locales don't need to be bundled at build time)
- Use `navigator.languages` correctly if available (don't just grab the first one in the list)
- Splits up locales into different locale files
- Add short README for adding new locales
- Updates `react-intl` library to latest


src/utils/i18n.js
```diff
@@ -1,35 +1,86 @@
...
 
-const _messages = localeData[lang] || localeData.en;
-
-if (configs.APP_CONFIG && configs.APP_CONFIG.translations && configs.APP_CONFIG.translations[lang]) {
-  const configTranslations = configs.APP_CONFIG.translations[lang];
-  for (const messageKey in configTranslations) {
-    if (!configTranslations.hasOwnProperty(messageKey)) continue;
-    if (!configTranslations[messageKey]) continue;
-    _messages[messageKey] = configTranslations[messageKey];
+function getLocales() {
```
johnshaughnessy, line 8: Doesn't need to happen in this PR but it would be good to make available all the languages we have translations for and a language pref in the preferences screen for cases when `navigator.language[s]` doesn't match the user preference.

src/verify.js
```diff
@@ -16,7 +16,7 @@ window.APP = { store };
 
 function Root() {
   return (
-    <IntlProvider locale={lang} messages={messages}>
+    <IntlProvider locale={getLocale()} messages={getMessages()}>
```
johnshaughnessy, line 19: I noticed you moved the IntlProvider from Root to the HomePage https://github.com/mozilla/hubs/pull/2879/files#diff-1fdf421c05c1140f6d71444ea2b27638R17 but didn't make the same refactor to the Roots containing SignInPage or VerifyPage . Was the point of the refactor just to move the IntlProvider closer to the code using it? Also -- a question related to React Contexts - is it a bug if we accidentally had an IntlProvider nested within an IntlProvider or is that OK?

src/assets/locales/README.md
```diff
@@ -0,0 +1,10 @@
+# Adding Locales
```
johnshaughnessy, line 1: Might be good to add a link to this document to the main `README.md` and/or `CONTRIBUTING.md` under a heading about adding translations/locales.

src/assets/locales/README.md
```diff
@@ -0,0 +1,10 @@
...
+
+1. Copy an existing locale (e.g. [en.json](en.json)) to a new file using appropriate locale code as filename. (e.g. [zh.json](zh.json))
+   * For locales that have duplicate codes (e.g. `zh` and `zh-ch`), edit [locale_config.js](locale_config.js) to define the fallback locale so that the locale file doesn't need to be duplicated.
+2. Edit your new locale file with your translations.
+
+# Adding to existing Locales
+
+1. Add your new key and translation to [en.json](en.json).
+2. Add the key and translation (either the same english translation from step 1, or the correct language equivalent if available, to the other locale files).
```
johnshaughnessy, line 10: Would be good for us to write a test that verifies all the keys in a canonical locale are present in all the others (and possibly no extra). I wouldn't be surprised if we've already missed a few items in `zh.json`

src/react-components/create-object-dialog.js
```diff
@@ -23,7 +23,7 @@ const attributionHostnames = {
 
 const isMobile = AFRAME.utils.device.isMobile() || AFRAME.utils.device.isMobileVR();
 const instructions = "Paste a URL to an image, video, model, scene, or upload.";
-const desktopTips = `Tip: You can paste URLs directly into ${messages["app-name"]} with Ctrl+V`;
+const desktopTips = `Tip: You can paste URLs directly into ${getMessages()["app-name"]} with Ctrl+V`;
```
johnshaughnessy, line 26: Oops, this is one of several remaining places where we should be using `getMessages()` for the whole text. Can be addressed later

InfiniteLee:
> lgtm! Glad to have a simple way to add / edit locales. Eventually we might want to lazy load them but we only have two now (not including `APP_CONFIG`s) so we don't need to do that work yet. It's really good that we have a clear and simple way for people to contribute new locales. Even without a sophisticated system of submission/reviewing correctness, this lets people get started.

Locales are lazy-loaded, via dynamic imports:
https://github.com/mozilla/hubs/blob/7d1b6299688c2a5fd7f64767b449fa8b963af358/src/utils/i18n.js#L32

src/assets/locales/README.md
InfiniteLee: Added to `CONTRIBUTING.md`

src/assets/locales/README.md
InfiniteLee: Made a node script `sync_locales.js` that will sync `en.json` to any other locale files present in this directory.

PR #2896, opened 2020-08-20
Improve accessibility for main UI

### LINKED ISSUE #2198: Menu of Options and Toolbar not available to screen reader
**Description**
Screen readers can't access the information in the menu of options and toolbar because of aira-hidden=true

image

Any time aria-hidden="true" set on an element a screen reader won't be able to find it

unnamed

**Additional context**
Flagged by the a11y meetup group using Hubs

### LINKED ISSUE #2199: Text chat entry field is not adequately described for screen readers
**Description**
The chat box doesn't provide enough information about what you are supposed to do with it. As users become more familiar with your interface a shorter description might be preferred but for now the information is not sufficient to understand what is happening when the field is encountered.

image(1)

**Additional context**
Flagged by the a11y meetup group using hubs

brianpeiris:
Adds ARIA roles and labels to several buttons and elements in the room UI. Tested with NVDA screen reader.

Closes #2198, closes #2199

### DIFF
--- a/src/react-components/2d-hud.js
+++ b/src/react-components/2d-hud.js
@@ -190,6 +190,7 @@ class TopHUD extends Component {
             ? "Stop sharing"
             : `Share ${capitalize(primaryVideoShareType)}${this.state.mediaDisabled ? " Disabled" : ""}`
         }
+        role="button"
         onClick={this.state.mediaDisabled ? noop : maybeHandlePrimaryShare}
         onMouseOver={this.state.mediaDisabled ? noop : showExtrasOnHover}
       >
@@ -231,7 +232,7 @@ class TopHUD extends Component {
     const tipDivForType = (type, cancelFunc) => (
       <div className={cx(styles.topTip)}>
         {cancelFunc && (
-          <button className={styles.tipCancel} onClick={cancelFunc}>
+          <button className={styles.tipCancel} aria-label="close" onClick={cancelFunc}>
             <i>
               <FontAwesomeIcon icon={faTimes} />
             </i>
@@ -247,7 +248,7 @@ class TopHUD extends Component {
     if (this.props.watching) {
       tip = (
         <div className={cx([styles.topTip, styles.topTipNoHud])}>
-          <button className={styles.tipCancel} onClick={() => this.props.onWatchEnded()}>
+          <button className={styles.tipCancel} aria-label="close" onClick={() => this.props.onWatchEnded()}>
             <i>
               <FontAwesomeIcon icon={faTimes} />
             </i>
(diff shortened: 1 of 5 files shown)


src/react-components/2d-hud.js
```diff
@@ -231,7 +232,7 @@ class TopHUD extends Component {
     const tipDivForType = (type, cancelFunc) => (
       <div className={cx(styles.topTip)}>
         {cancelFunc && (
-          <button className={styles.tipCancel} onClick={cancelFunc}>
+          <button className={styles.tipCancel} aria-label="close" onClick={cancelFunc}>
```
johnshaughnessy, line 235: I think at some point we'll want to go back and get the `aria-labels` (and as much of everything else as we can) from a translations file.

src/react-components/settings-menu.js
```diff
@@ -430,14 +438,17 @@ export default class SettingsMenu extends Component {
...
+            role="button"
+            aria-label="settings menu"
             onClick={() => this.setState({ expanded: !this.state.expanded })}
             className={classNames({
               [rootStyles.cornerButton]: true,
               [rootStyles.cornerButtonSelected]: this.state.expanded
             })}
-          />
+          >
+            <FontAwesomeIcon icon={faBars} />
```
johnshaughnessy, line 450: Does the change to nesting affect the styling at all? In general a best practice to nest the `FontAwesomeIcon`s instead of styling/making them clickable directly?

src/react-components/settings-menu.js
```diff
@@ -430,14 +438,17 @@ export default class SettingsMenu extends Component {
     return (
       <div>
         {!this.props.showAsOverlay && (
-          <FontAwesomeIcon
-            icon={faBars}
+          <div
+            role="button"
```
johnshaughnessy, line 442: Should this be a `<button>` instead of a `<div>`? (This may affect styling due to default browser styles)

src/react-components/2d-hud.js
brianpeiris: Yeah, that's fair. I think it makes sense to do so when we do the larger redesign, since a lot of these buttons ought to be abstracted away behind common components.

src/react-components/settings-menu.js
brianpeiris: I think the styling just happened to work out in this case, I do think it's bad practice to make the icon clickable, especially since these FontAwesomeIcons don't accept ARIA attributes, and they are directly rendered as `<svg>` tags.

src/react-components/settings-menu.js
brianpeiris: Yeah, we definitely ought to use `<button>` or `<a>` tags for clickable elements, instead of `<div>`s. I think I would have had to spend a lot of time fiddling with CSS if I made that change here, and I wanted to avoid that because of the upcoming redesign.

brianpeiris:
Thanks for the review @johnshaughnessy!

Issue #2898, opened 2020-08-21
Support waypoint anchor short-hand links with 'Go To' button


misslivirose:
**Is your feature request related to a problem? Please describe.**
It's possible to use the new waypoint anchor feature (#2837) to create a teleporter link that can jump you into a new waypoint without clicking on it as seen in the attached gif. However, right now it uses the full qualified URL for the room, so the button on the link image says "Visit Room", which implies navigating to a new room entirely. 

**Describe the solution you'd like**
This is a two-part feature request, but the first part would still help significantly though I'm not entirely sure how feasible it is. It would be awesome to have a different string on the button when it is a teleport link instead of a new room (for example, saying something like 'Go To' or 'Teleport' or 'Go' or even just dropping to 'Visit' instead of 'Visit Room') so that when a user hovered, it would be more clear that it wouldn't necessarily open a new room. 

The second part is presumably more complex, but it would be great to just be able to specify the desired anchor point (e.g. `#WayPointName` ) in a link element (or other custom element) that had the correct button associated with it. I'm not sure it's entirely necessary at this point, but it would great to have this for scenes so the teleport functionality would work even without having to change it for individual rooms. This way, if a link element (or new, custom element) detected that the link was just the waypoint anchor, it would automatically navigate to the waypoint for that specific room if it existed. 

**Additional context**
![Teleport-Small](https://user-images.githubusercontent.com/4313320/90837788-e71dfd80-e307-11ea-9b88-93e13fc4952c.gif)

Charlesc22:
Love this idea Liv!
I just made a room with anchors from your video. Worked great. This opens up lots of possibilities!

PR #2918, opened 2020-08-24
Improved room links w/ anchor hash

InfiniteLee:
As requested in https://github.com/mozilla/hubs/issues/2898, this replaces the button text in anchor hashed room links with "Go To". This also adds a check such that if an anchor hash is pasted by itself (e.g. "#Spawn_Point_1"), the rest of the current room url will be prepended so that it resolves correctly as a room link.

waypoints3

Closes https://github.com/mozilla/hubs/issues/2898

### DIFF
--- a/src/components/media-loader.js
+++ b/src/components/media-loader.js
@@ -359,7 +359,7 @@ AFRAME.registerComponent("media-loader", {
 
       //check if url is an anchor hash e.g. #Spawn_Point_1
       if (src.charAt(0) === "#") {
-        src = this.data.src = `${window.location.origin}${window.location.pathname}${src}`;
+        src = this.data.src = `${window.location.origin}${window.location.pathname}${window.location.search}${src}`;
       }
 
       let canonicalUrl = src;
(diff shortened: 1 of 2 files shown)


src/components/media-loader.js
```diff
@@ -356,6 +357,11 @@ AFRAME.registerComponent("media-loader", {
         this.showLoaderTimeout = setTimeout(this.showLoader, 100);
       }
 
+      //check if url is an anchor hash e.g. #Spawn_Point_1
+      if (src.charAt(0) === "#") {
+        src = this.data.src = `${window.location.origin}${window.location.pathname}${src}`;
```
johnshaughnessy, line 362: Might want to include `${window.location.search}` after `pathname` (these are the query string parameters, if any)

johnshaughnessy, line 362: Odd to be writing to `this.data.src` because it's AFRAME component state, but we wouldn't really want to `setAttribute` either because that would cause another call to `update`. :thinking:

InfiniteLee: Yea, this was unfortunately necessary because other components specifically refer to this component's `data.src`

PR #2919, opened 2020-08-25
add preference to override browser locale

InfiniteLee:
locale

### DIFF
--- a/src/assets/locales/locale_config.js
+++ b/src/assets/locales/locale_config.js
@@ -2,6 +2,11 @@
 // For non "en" (since it's the default), use this mapping to define a fallback,
 // which will only be used if the specified locale file doesn't exist.
 
+export const AVAILABLE_LOCALES = {
+  en: "English",
+  zh: "简体中文"
+};
+
 export const FALLBACK_LOCALES = {
   "zh-cn": "zh",
   "zh-hans": "zh",
--- a/src/react-components/preferences-screen.js
+++ b/src/react-components/preferences-screen.js
@@ -9,6 +9,7 @@ import { FormattedMessage } from "react-intl";
 import { WrappedIntlProvider } from "./wrapped-intl-provider";
 import styles from "../assets/stylesheets/preferences-screen.scss";
 import { getMessages } from "../utils/i18n";
+import { AVAILABLE_LOCALES } from "../assets/locales/locale_config";
 
 const isMobile = AFRAME.utils.device.isMobile() || AFRAME.utils.device.isMobileVR();
 
@@ -500,6 +501,11 @@ const preferredCamera = {
   defaultString: "default"
 };
 
+const availableLocales = [{ value: "browser", text: getMessages()["preferences.browserDefault"] }];
+for (const locale in AVAILABLE_LOCALES) {
+  availableLocales.push({ value: locale, text: AVAILABLE_LOCALES[locale] });
+}
+
 const DEFINITIONS = new Map([
   [
     CATEGORY_TOUCHSCREEN,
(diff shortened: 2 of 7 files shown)


src/assets/locales/locale_config.js
```diff
@@ -2,6 +2,11 @@
 // For non "en" (since it's the default), use this mapping to define a fallback,
 // which will only be used if the specified locale file doesn't exist.
 
+export const AVAILABLE_LOCALES = {
+  en: "English",
+  zh: "简体中文"
+};
```
johnshaughnessy, line 8: Haha I didn't think about this before but I guess these don't need to be in translation files, do they :laughing: 
(I guess they _could_ be, as in "Spanish" vs "Español", but having them written in their own language ("Español", not "Spanish") seems better.)

johnshaughnessy: I was about to suggest adding flags, but this is sometimes discouraged: https://wplang.org/never-use-flags-language-selection/
This article _does_ recommend including _both_ the native language AND the translated name. I don't think it's necessary for right now (especially because the formatting of the dropdown isn't especially good), but something for us to consider.

src/react-components/preferences-screen.js
```diff
@@ -582,6 +588,13 @@ const DEFINITIONS = new Map([
   [
     CATEGORY_MISC,
     [
+      {
+        key: "locale",
+        prefType: PREFERENCE_LIST_ITEM_TYPE.SELECT,
+        options: availableLocales,
+        defaultString: "browser",
+        promptForRefresh: true
```
johnshaughnessy, line 596: It would be nice not to have to refresh (in part because the refresh text may be in a language the user does not read), but simplest thing is probably best for now. Actually I'm surprised it's necessary, because I thought that enabling this was the goal of `WrappedIntlProvider`. Does that not work the way I thought it would?

src/assets/locales/locale_config.js
johnshaughnessy: One nice thing about the flag is that lots of sites use something like a flag to indicate a language selection. I don't think it applies here because the preference screen is already "hidden" in the menu from the perspective of someone who doesn't read English (or w/e the default locale of the browser that we also support happens to be).
Makes me think we may want to elevate language selection to somewhere more easily noticed than the preferences screen (like something in the "settings menu" that "preferences" is in).

vincentfretin: You already use react-intl in the code base, you may want to use Language Display Names, see
https://formatjs.io/docs/intl#formatdisplayname
and https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DisplayNames
Chrome implements it. For Firefox you need the polyfill.

src/react-components/preferences-screen.js
vincentfretin: Don't you just need to specify like `key={locale}` on the `WrappedIntlProvider` component to be sure react rerender the whole tree?

InfiniteLee: The main issue is that `i18n.js` is a really "dumb" chunk of code that is included in a bunch of other files- but it has the duty of loading the locale files via dynamic imports. It somehow needs to be notified that the locale has been changed via preferences in order to load the newly requested locale. This is typically done elsewhere by listening for `statechanged` on the parameter store, however since `i18n.js` is loaded so early, there isn't a reference it can use to the parameter store in order to listen for those events yet. 

Assuming the above is solved, the next step is re-rendering the various `WrappedIntlProvider` trees- but that should be handled automatically as `locale-updated` will be fired, and that will cause the state on `WrappedIntlProvider` to get updated, causing a re-render. I'm not sure that `key={locale}` will do anything more to help with that (maybe that is something from the older versions of `react-intl`?).

vincentfretin: For the `key={locale}` yeah I read it somewhere, but can't find it in the current documentation. Anyway I saw you're using setState in your component, so this already force a rerender as far as I know.

PR #2967, opened 2020-09-02
Media Frames

netpro2k:
Adds support for "Media Frames" that you can snap different types of media into to position them precisely in space. Items will be scaled to fill the media frame, while preserving their original aspect ratio. When an item is removed from a frame it will return to its original scale. Media frames can be created from the Elements panel in Spoke.

image

The visuals are also very crude and could use some design love, but again, doesn't feel worth holding back this useful feature for.

Associated Spoke PR: https://github.com/mozilla/Spoke/pull/1019

TODO:
- [x] merge https://github.com/MozillaReality/networked-aframe/pull/42
- [x] update package-lock.json to point at updated networked-aframe

Known Issues:
- Pinned objects will not be re-captured into a frame if everyone leaves the room. They will remain in their position and scale as they naturally would, but the media frame will be considered empty. This is unfortunately kind of tricky to handle and will need some fixes to our networking code. (The pinned object is unowned if nobody interacts with it, and media frame snapping is handled by each client for objects it owns)
- When grabbing an object that you don't own that is snapped into a media frame, it will immediately un-snap from the frame rather than waiting to be dragged out of the frame. This is likely something to do with changing collision flags while objects are already intersecting. The same thing also seems to happen for objects that were tossed into a media zone instead of placed... this one is more puzzling.


src/systems/super-spawner-system.js
```diff
@@ -38,6 +38,13 @@ export class SuperSpawnerSystem {
       data.mediaOptions
     ).entity;
 
+    // TODO HACK objects become held mid frame here, so if something runs between here and
+    // when the networked component is initialized isMine will throw since data doesnt exist
+    // should probably fix isMine in networked aframe instead of merging this
+    if (spawnedEntity.components.networked && !spawnedEntity.components.networked.data) {
+      spawnedEntity.components.networked.data = {};
+    }
```
johnshaughnessy, line 46: note to myself: ask Dom what this is about

src/systems/media-frames.js
```diff
@@ -0,0 +1,219 @@
...
+      const bodyUUID = zone.el.components["body-helper"].uuid;
+      // TODO would be nice to fix the timing so we don't need to check this
+      if (!this.physicsSystem.bodyInitialized(bodyUUID)) continue;
+
+      const objectToCapture = getCapturableObjectCollidingWithBody(this.physicsSystem, zone.data.mediaType, bodyUUID);
+
+      const previewMesh = zone.el.getObject3D("preview");
+      const guideMesh = zone.el.getObject3D("guide");
+
+      guideMesh.visible = this.interactionSystem.isHoldingAnything(isCapturableByType[zone.data.mediaType]);
```
johnshaughnessy, line 72: If we add networked state, might be nice if non-empty zones showed up differently.
Also might want guides to show up when hovering on an object in a zone? (Maybe only when in freeze mode?)

src/systems/media-frames.js
```diff
@@ -0,0 +1,219 @@
...
+
+      if (zone.state === TARGET_ZONE_STATE.EMPTY) {
+        // TODO not sure if we want to use tags for this
+        if (objectToCapture && !isTagged(objectToCapture.el, "isHoveringMediaTargetZone")) {
+          setTag(objectToCapture.el, "isHoveringMediaTargetZone");
+
+          zone.state = TARGET_ZONE_STATE.HOVERING;
+          zone.target = objectToCapture;
+
+          const imageObject3D = objectToCapture.el.getObject3D("mesh");
```
johnshaughnessy, line 82: `imageObject3D`

src/systems/media-frames.js
```diff
@@ -0,0 +1,219 @@
...
+
+          // TODO this feels a bit hacky and error prone, also needs support for previewing 3D objects
+          if (isCapturableByType[MediaType.ALL_2D]) {
```
johnshaughnessy, line 85: This seems to be checking the wrong thing. Isn't this a function?

src/systems/media-frames.js
```diff
@@ -0,0 +1,219 @@
...
+            // Preview mesh UVs are set to accomidate textureLoader default, but video textures don't match this
+            previewMesh.scale.y *= TEXTURES_FLIP_Y !== previewMesh.material.map.flipY ? -1 : 1;
+            previewMesh.matrixNeedsUpdate = true;
+            previewMesh.visible = true;
+          }
+
+          guideMesh.material.uniforms.color.value.set(0x6fc0fd);
+        }
+      } else {
+        // TODO currently this works because the object will continue to collide, we may not want this
```
johnshaughnessy, line 101: Does this get easier with networked state?

src/systems/media-frames.js
```diff
@@ -0,0 +1,219 @@
...
+          previewMesh.material.map = null;
+          previewMesh.material.needsUpdate = true;
+          previewMesh.visible = false;
+
+          guideMesh.material.uniforms.color.value.set(0x2f80ed);
+        }
+
+        if (zone.state === TARGET_ZONE_STATE.FULL && zone.data.dynamic) {
+          zone.target.position.copy(zone.el.object3D.position);
+          zone.target.rotation.copy(zone.el.object3D.rotation);
```
johnshaughnessy, line 124: I expect zone targets to be at root level of the scene graph, but zones may be deeper in the hierarchy (e.g. when added in blender). Might want `worldPosition/Rotation`.

src/systems/media-frames.js
```diff
@@ -0,0 +1,219 @@
...
+  EMPTY: "empty",
+  HOVERING: "hovering",
+  FULL: "full"
+};
+
+function scaleForAspectFit(containerSize, itemSize) {
+  return Math.min(containerSize.x / itemSize.x, containerSize.y / itemSize.y);
+}
+
+const isCapturableByType = {
```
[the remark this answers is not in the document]
netpro2k: We currently don't have any non-networked objects that aren't static parts of the scene, but if we do eventually add some this is easy to change.

src/systems/media-frames.js
netpro2k: They currently do show a different color when full, though like you said its not networked.

src/systems/media-frames.js
```diff
@@ -0,0 +1,295 @@
...
+      if (!this.physicsSystem.bodyInitialized(bodyUUID)) continue;
+
+      const holdingSomethingCapturable = this.interactionSystem.isHoldingAnything(
+        isCapturableByType[frame.data.mediaType]
+      );
+
+      const guideMesh = frame.el.getObject3D("guide");
+      guideMesh.visible = holdingSomethingCapturable;
+
+      if (frame.data.targetId === "empty") {
```
netpro2k, line 57: I would like to clean up the branching logic here a bit, but would like to do it as a followup PR since its tricky to get right and I don't want to restart the testing process now.

src/network-schemas.js
```diff
@@ -161,6 +161,22 @@ function registerNetworkSchemas() {
     ]
   });
 
+  NAF.schemas.add({
+    template: "#interactable-media-frame",
```
johnshaughnessy, line 165: Later may need to add to https://github.com/mozilla/reticulum/blob/master/lib/ret_web/channels/hub_channel.ex#L695 to allow users to spawn them. (I don't think this is necessary for objects in the scene tho).

src/systems/media-frames.js
```diff
@@ -0,0 +1,297 @@
...
+    if (this.data.debug) {
+      console.log(
+        `${this.el.id}: ${oldData.targetId}(${JSON.stringify(oldData.originalTargetScale)}) -> ${
+          this.data.targetId
+        }(${JSON.stringify(this.data.originalTargetScale)})`
+      );
+    }
+
+    // Ownership race, whoever owns the old object needs to take care of "ejecting" it
+    if (oldData.targetId !== "empty" && oldData.tergetId !== this.data.targetId) {
```
johnshaughnessy, line 203: typo `oldData.tergetId` -> `oldData.targetId`

jessicasymons:
Media Frames is a great idea. However I cannot work out how to put my displayed image/file into the Media Frame using a mouse.

PR #2969, opened 2020-09-02
Fixing exception and black screen on Quest, when Quest is in Hands tracking mode.

Artyom17:
In the hands mode, the controllers are emulated by the hands. They are represented with the gamepad object with a single button and no axis. Current code expects to have two buttons and two sets of axis, which causes an exception and malfunction.


johnshaughnessy (approved):
Thanks @Artyom17 !

PR #2992, opened 2020-09-09
Refactor Pen Tool to use triangles instead of triangle strips for rendering

InfiniteLee:
Three.js as of r112 has removed `drawMode` from meshes, thus deprecating support for triangle-strip rendering. This updates `networked-drawing` and `sharedbuffergeometry` to use indexed triangles instead. 

This also moves the serialize drawing button on the drawing hover menu so that it doesn't overlap with the new media refresh button.


src/utils/sharedbuffergeometry.js
```diff
@@ -10,29 +9,21 @@ export default class SharedBufferGeometry {
...
-        const componentSize = 3;
+        const componentSize = key === "index" ? 1 : 3;
         let pos = prevIdx[key] * componentSize;
         const start = (idx[key] + 1) * componentSize;
         const end = this.idx[key] * componentSize;
         for (let i = start; i < end; i++) {
-          this.current.attributes[key].array[pos++] = this.current.attributes[key].array[i];
+          if (key === "index") {
+            //index needs to be handled specially
+            this.current.index.array[pos++] = this.current.index.array[i] - idx.position - 1;
```
johnshaughnessy, line 23: What does this line do / why are indices handled differently? I'm having a hard time keeping track of `idx`, `prevIdx`, `this.idx` and all of their `.index` attributes

src/components/tools/networked-drawing.js
```diff
@@ -707,29 +656,36 @@ AFRAME.registerComponent("networked-drawing", {
...
+      const b = bufferVertexCount - (this.segments - i);
+      const c = i + 1 < this.segments ? b + 1 : bufferVertexCount - this.segments;
+      this._addIndex(a, c, b);
     }
   },
 
-  _restartPrimitive() {
-    this.sharedBuffer.restartPrimitive();
-    ++this.vertexCount;
+  _drawEndCap(point, segments, normal) {
```
johnshaughnessy, line 679: Seems like `segments` are unused

johnshaughnessy (approved):
I'm not as familiar with this code as I'd like to be - but these changes lgtm

src/utils/sharedbuffergeometry.js
InfiniteLee: unlike `position`, `normals` or `colors`, indexes are not accessible on the `geometry.attributes` property, they're instead accessed via `geometry.index` directly, or via `getIndex()` or `setIndex()`

Issue #2999, opened 2020-09-10
Stop non-admin accounts from seeing admin panel at all


robin-k-wilson:
**Is your feature request related to a problem? Please describe.**
Any user can see the admin panel, but NOT the data if they aren't an admin. 

It's confusing for users that believe they are an admin and see the panel, but are not. 

**Describe the solution you'd like**
If you are not an admin, the admin panel does not load at all. Maybe with a redirect back to the homepage or you can't access page shows up. 

Keywords: admin panel, admin console

Not an admin:
![image](https://user-images.githubusercontent.com/29560735/92756053-4be5da00-f341-11ea-9eda-8b8cbc5c9193.png)

robin-k-wilson:
Live on HMC

PR #3027, opened 2020-09-14
non-admins see a "sorry not an admin" page

robin-k-wilson:
Added Unauthorized page (see screenshot)

If not an admin, admin panel renders briefly, then you are redirected to the unauthorized page. (added via this PR + reticulum PR (https://github.com/mozilla/reticulum/pull/415/files))
If an admin, you see the admin panel. (obviously, already implemented)
If not logged in at all and try to access admin panel, you're redirected to login (already implemented)

image

### DIFF
--- a/admin/src/admin.js
+++ b/admin/src/admin.js
@@ -84,7 +84,6 @@ class AdminUI extends Component {
     if (process.env.NODE_ENV !== "development" || qs.get("idle_timeout")) detectIdle();
     window.addEventListener("idle_detected", this.onIdleDetected);
     window.addEventListener("activity_detected", this.onActivityDetected);
-    console.log("inside did mount");
     const adminInfo = await getAdminInfo();
     // Unauthorized account
     if (adminInfo.status === 401) this.setState({ isAdmin: false });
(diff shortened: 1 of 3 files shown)


robin-k-wilson:
Fixes: https://github.com/mozilla/hubs/issues/2999

ocularvr:
Brilliant!!!! Thank you. Looks so much more professional like this.

admin/src/admin.js
```diff
@@ -74,13 +76,17 @@ class AdminUI extends Component {
...
   };
 
-  componentDidMount() {
+  async componentDidMount() {
     if (process.env.NODE_ENV !== "development" || qs.get("idle_timeout")) detectIdle();
     window.addEventListener("idle_detected", this.onIdleDetected);
     window.addEventListener("activity_detected", this.onActivityDetected);
+    const adminInfo = await getAdminInfo();
+    // Unauthorized account
+    if (adminInfo.status === 401) this.setState({ isAdmin: false });
```
johnshaughnessy, line 89: A little bit awkward to have `getAdminInfo` return the `Response` object in the error case (so that we can check the `status`) and the parsed json body in the success case. I don't have a better suggestion for folding in the status check/error handling but noting it because it stuck out to me.

johnshaughnessy (approved):
So glad to see this

PR #3061, opened 2020-09-21
Update loop-animation struct

keianhzo:
Update loop-animation struct to match the one introduced in https://github.com/mozilla/Spoke/pull/1029


robertlong (commented):
Looks good! I'd add `activeClipIndices: { type: "array" }` to the schema just to be consistent.

keianhzo:
@robertlong Updated

PR #3067, opened 2020-09-24
Add ability to inspect avatars

johnshaughnessy:
Adds the ability to inspect avatars (including your own) like you can for other media objects.


johnshaughnessy (commented):
notes from talking with dom

misslivirose:
What is the experience like with this feature enabled in a crowded room? My specific concern is how it will interact with teleportation.

misslivirose:
I'm also concerned a bit with the optics of this one - there's not a lot of detail that I can see here so apologies if this is obvious by the feature and I'm missing it - does this have any impact on audio, and being able to hear some one from far away? And is the inspection just grabbing a copy of the 3D model, or is it the actual avatar duplicated to what's inspected here? (E.g. if I'm talking in VR and moving around, will that be shown when it's being inspected, or is it just a static copy of the avatar I've chosen to wear?

johnshaughnessy:
> What is the experience like with this feature enabled in a crowded room? My specific concern is how it will interact with teleportation.

I aimed to keep the hoverable bounding box for avatar inspection small so as to have a minimal impact on object interactions and teleportation. It is not currently based on the avatar mesh itself, so large avatars or poorly-calculated bounding box cannot affect its size. Below is a screenshot of the box visualized:

image
image



> does this have any impact on audio, and being able to hear some one from far away?

As it's written, yes -- The audio of the person you are inspecting sounds as if it's coming from right in front of you. (It move the source, not the listener, so voices around the inspected party remain as quiet as they were before inspecting.) This sounds useful to me, but if it's undesirable I will change it so that inspecting does not affect audio.

> is the inspection just grabbing a copy of the 3D model, or is it the actual avatar duplicated to what's inspected here? (E.g. if I'm talking in VR and moving around, will that be shown when it's being inspected, or is it just a static copy of the avatar I've chosen to wear?

It is not a copy of the 3D model, but acts like one. Inspecting uses the same mechanic as the freeze mode; once Alice starts inspecting Bob, Alice will not see Bob's movements until Alice stops inspecting.

misslivirose:
Thanks for the info! 

I think that if we are changing how people can experience audio this way, we should run this by the DPX team and get a sense of how it may have an impact on how people feel comfortable in a space. The implications of spatialized audio is still new to a lot of people, and I can potentially see that being controversial. 

The other two concerns seem fine.

misslivirose:
I would be fine with us shipping this as a V1 where the audio listener stayed relative to the physical location of the avatar inspecting. I think that amplifying a user's voice is pretty important but want to make sure we do so in a way that feels comfortable and understandable by everyone.

johnshaughnessy:
Ok, for now I've made it so that the audio of the inspected person remains unchanged to the inspecting person

PR #3068, opened 2020-09-25
Add ability to inspect yourself

johnshaughnessy:
Press the `i` key to inspect your avatar like you can inspect other media objects

### DIFF
--- a/src/hub.html
+++ b/src/hub.html
@@ -1171,6 +1171,7 @@
                 set-yxz-order
                 position="0 1.6 0"
             >
+                <a-entity id="inspect-yourself"></a-entity>
                 <a-entity
                     id="gaze-teleport"
                     position = "0.15 0 0"
--- a/src/systems/camera-system.js
+++ b/src/systems/camera-system.js
@@ -76,6 +76,7 @@ const moveRigSoCameraLooksAtObject = (function() {
   const cwp = new THREE.Vector3();
   const oForw = new THREE.Vector3();
   const center = new THREE.Vector3();
+  const boxFix = new THREE.Vector3(0.3, 0.3, 0.3);
   const target = new THREE.Object3D();
   return function moveRigSoCameraLooksAtObject(rig, camera, object, distanceMod) {
     if (!target.parent) {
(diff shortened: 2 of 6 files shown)


src/hub.html
```diff
@@ -1171,6 +1171,7 @@
                 set-yxz-order
                 position="0 1.6 0"
             >
+                <a-entity id="inspect-yourself"></a-entity>
```
johnshaughnessy, line 1174: self-inspection-target

src/systems/camera-system.js
```diff
@@ -76,6 +76,7 @@ const moveRigSoCameraLooksAtObject = (function() {
   const cwp = new THREE.Vector3();
   const oForw = new THREE.Vector3();
   const center = new THREE.Vector3();
+  const boxFix = new THREE.Vector3(0.3, 0.3, 0.3);
```
johnshaughnessy, line 79: `default-box-size`

src/systems/camera-system.js
```diff
@@ -213,7 +219,7 @@ export class CameraSystem {
     this.mode = NEXT_MODES[this.mode] || 0;
   }
 
-  inspect(o, distanceMod, temporarilyDisableRegularExit) {
+  inspect(o, distanceMod, temporarilyDisableRegularExit, optionalAlternativeToNotHide) {
```
johnshaughnessy, line 222: Probably better to replace `optionalAlternativeToNotHide` with an offset instead. Did not do it in this PR because it was slightly more work, but might followup and do it later or leave a comment about it in the code.

src/systems/userinput/bindings/keyboard-mouse-user.js
```diff
@@ -908,7 +915,17 @@ export const keyboardMouseUserBindings = addSetsToBindings({
       priority: 101
     },
     {
-      src: [stopInspectingViaMouse, stopInspectingViaKeyboardEscape, stopInspectingViaKeyboardFocusRelease],
+      src: { value: paths.device.keyboard.key("i") },
+      dest: { value: stopInspectingSelf },
+      xform: xforms.rising
```
johnshaughnessy, line 920: For consistency, it might be better to use `falling` but personally I don't like the idea of having to hold the `i` key while orbiting.

src/systems/camera-system.js
johnshaughnessy, line 222: Instead of adding `optionalAlternativeToNotHide` -- add `alternatePivot` or something to indicate rotating around something else.

PR #3087, opened 2020-09-28
pt-br translation

tttigo:
add brazilian portuguese support to hubs

@peusodre made some corrections


src/assets/locales/locale_config.js
```diff
@@ -4,11 +4,13 @@
 
 export const AVAILABLE_LOCALES = {
   en: "English",
-  zh: "简体中文"
+  zh: "简体中文",
+  pt: "Portugês"
```
keianhzo, line 8: The way this is configured right now it's not finding the locale correctly as it willl look for **pt.json** localization file. For the fallback mechanism to correctly work when the browser language is set to **Português (Brasil)** we need the available locale to be set to **pt-br**:

```
export const AVAILABLE_LOCALES = {
  en: "English",
  zh: "简体中文",
  "pt-br": "Portugês (Brasil)"
};
```

And fallback to **Português** until we have a **Português** localization:
```
export const FALLBACK_LOCALES = {
  "zh-cn": "zh",
  "zh-hans": "zh",
  "zh-hans-cn": "zh",
  pt: "pt-br"
};
```

keianhzo (changes requested):

src/assets/locales/locale_config.js
keianhzo: @johnshaughnessy Can you please confirm this?

keianhzo:
@tttigo Looks good! You just need to fix the locale configuration.

src/assets/locales/locale_config.js
johnshaughnessy: Yes that's correct and the subsequent commit fixed this.

(Oops, I missed this ping. My github notifications are all messed up / disabled at the moment. I'll try to fix them so I don't miss things but also feel free to dm me on discord)

johnshaughnessy (approved):
looks fixed as per @keianhzo 's review

PR #3088, opened 2020-09-29
Fix scene transitions when src url does not change

johnshaughnessy:
If the `src` of the environment's `gltf-model-plus` doesn't change (which might happen if you are going to/from the loading scene url on purpose, or if two scene transitions occur in rapid succession), then `update` is never called and you can get stuck with a black screen because the transition never finishes and the `fader` never fades back in.

### DIFF
--- a/src/hub.js
+++ b/src/hub.js
@@ -445,6 +445,10 @@ async function updateEnvironmentForHub(hub, entryManager) {
         );
 
         sceneEl.emit("leaving_loading_environment");
+        if (environmentEl.components["gltf-model-plus"].data.src === sceneUrl) {
+          console.warn("Updating environment to the same url.");
+          environmentEl.setAttribute("gltf-model-plus", { src: "" });
+        }
         environmentEl.setAttribute("gltf-model-plus", { src: sceneUrl });
       },
       { once: true }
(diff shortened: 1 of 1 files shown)


src/hub.js
```diff
@@ -441,6 +441,10 @@ async function updateEnvironmentForHub(hub, entryManager) {
         );
 
         sceneEl.emit("leaving_loading_environment");
+        if (environmentEl.components["gltf-model-plus"].data.src === sceneUrl) {
+          console.warn("Updating environment to the same url.");
+          environmentEl.setAttribute("gltf-model-plus", { src: "" });
+        }
```
johnshaughnessy, line 447: This won't happen in "normal conditions" -- maybe solve this by adding a qs parameter to `loadingEnvironment`. Also we should just audit the fader code/usage because it's suspicious.

Won't merge this.

johnshaughnessy:
I initially thought that this would only happen in a development environment, but that's not correct. This will occur on Hubs Cloud instances on their first attempted scene transition in cases when they have not replaced the default scene. While I'm happy if someone wants to replace this fix with an alternative (adding a qs parameter to the default loading environment url so that the `src` doesn't match or something like that), for now I think this fix should be adopted because it works / has already been tested.

PR #3089, opened 2020-09-29
Allow hub_refresh without session id

johnshaughnessy:
The upcoming graphql API will broadcast `hub_refresh` on the hub channel, but will not have a `session_id`. In this case, we want the same behavior, except that the message to users will not have another user's display name and instead will say, "API".

### DIFF
--- a/src/hub.js
+++ b/src/hub.js
@@ -1553,6 +1553,57 @@ document.addEventListener("DOMContentLoaded", async () => {
     addToPresenceLog(incomingMessage);
   });
 
+  hubPhxChannel.on("hub_refresh_by_api", ({ hubs, stale_fields }) => {
+    const hub = hubs[0];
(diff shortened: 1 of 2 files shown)


src/hub.js
```diff
@@ -1556,6 +1556,7 @@ document.addEventListener("DOMContentLoaded", async () => {
   hubPhxChannel.on("hub_refresh", ({ session_id, hubs, stale_fields }) => {
     const hub = hubs[0];
     const userInfo = hubChannel.presence.state[session_id];
+    const displayName = (userInfo && userInfo.metas[0].profile.displayName) || "API";
```
johnshaughnessy, line 1559: Add comment about this being the first (of possibly many) places where we do something differently based on whether the message came from the API vs from another user on the channel.

Issue #3081, opened 2020-09-28
Zooming in an object without spatialized audio has an effect to audio


takahirox:
**Description**

If I disable spatialized audio with `/audiomode` command and zoom in an asset by pressing f, voice of people whose avatars are close to the asset will be louder.

If I don't disable spatialized audio, this doesn't happen. (More precisely, the voice can become louder for a single frame and then back to normal? But I don't investigate it here in this issue so far. Update: I opened an issue #3098. I tested more and I felt it may be an audio noise rather than louder for a single frame(?))

Even zooming in, avatar position and audio listener position don't seem to move (Thanks @johnshaughnessy), then the behavior with spatialized audio seems correct (and more intuitive, I think).

(Note: If the asset has audio source, the audio source is close to the audio listener if you zoom in. But it doesn't affect the audio of people around the asset.)

**To Reproduce**

1. Enter a room
2. Run a bot with audio `node run-bot.js -a bot-recording.mp3 -r roomId`
3. Disable spatialized audio with `/audiomode`
4. Place an asset like a picture close to the bot avatar
5. Move far from the avatar
6. Press f on the asset and zoom in
7. Confirm the bot voice becomes louder

**Expected behavior**

The behavior without spatialized audio should be same as the one with spatialized audio.

**Hardware** 
 - Device: [e.g. Desktop, phone, VR headset] Desktop
 - OS: [e.g. Windows, iOS, Linux] Windows
 - Browser [e.g. Firefox] Chrome and Firefox

**Additional context**

Here seems the root issue (Thanks again @johnshaughnessy) that it calculates audio rolloff from the distance between avatar and camera, not audio listener, in non-spatialized audio mode.

https://github.com/mozilla/hubs/blob/master/src/components/avatar-volume-controls.js#L58

PR #3090, opened 2020-09-29
Use audio listener for falloff, not camera

johnshaughnessy:
Fix https://github.com/mozilla/hubs/issues/3081


takahirox (approved):
I confirmed this change resolves the problem. This change looks nice to me unless there is any special case where we want to calculate the distance from camera.

PR #3093, opened 2020-09-29
Fallback to default local model when there is an avatar load error

### LINKED ISSUE #3004: Avatar audio should be decoupled from the loading of the GLTF model / head node
**Description**
If an issue causes an avatar not to load on the page, the user cannot be heard. We should decouple the audio from being dependent on the avatar loading, so that the client is more resilient to server loading 502/504 errors.

keianhzo:
Fixes #3004 This PR ensures that when there is any avatar load error we gracefully fallback to a local default avatar (which ideally would never fail, at least due to networking issues). Collaterally it guarantees that the avatar-audio-source component is always present in the avatar element which fixes the linked issue.

This only updates the avatar locally and the avatar update will not be network propagated.

@misslivirose I have used the current default local duck avatar, that should work for the moment. The avatar update can be a follow-up PR depending on how long that takes.


brianpeiris (approved):
Looks good to me. Nice one!

One relatively minor note is that the DefaultAvatar.glb does not work well with our personal space bubble system -- the duck will disappear entirely if it gets close to another user since we assume avatars have a "SkinnedMesh" in `personal-space-bubble.js`. Not a big deal, but something to keep in mind until we get a newer default avatar model.

misslivirose:
Should we use a robot as the DefaultAvatar.glb model? I think that the duck has other issues IIRC. Maybe @j-conrad has an idea of one that we could drop in relatively easily?

brianpeiris:
Yeah, I don't see a problem with grabbing one of the standard bots from our default set. You'd have to pack it as a GLB though.
Also, now that you mention it, we'll have to keep in mind that this fallback will also be used on Hubs Cloud instances and it would require a custom client deploy to change it, so best to use something that is unbranded as possible.

j-conrad:
I'd probably go with one of the standard 'color' bots since they have hands (in case you're using 6-DOF controllers).

keianhzo:
@misslivirose @j-conrad If we go for one of the default ones I guess we need to choose between male/female or just pack both and randomize it. Any other gender neutral alternatives?

misslivirose:
We could use the box bot avatar and then just do a gray texture, maybe? That reads to me like a 'placeholder', non-gendered avatar, but would love other thoughts. 

image

misslivirose:
@keianhzo - Here's a zip version of the gray box bot as a .glb (I couldn't upload the GLB directly to GitHub) with the head animation defaults. I can send you the .glb directly another way if you need it! 

[Default_Bot.zip](https://github.com/mozilla/hubs/files/5319556/Default_Bot.zip)

keianhzo:
@misslivirose Updated, I've added eyes and blinking to it to make it a bit more human.

PR #3104, opened 2020-09-30
Fix the last used input device id

keianhzo:
As part of my research on audio issues, I have stumbled upon this one. When storing the `lastUsedMicDeviceId` we are trying to get the deviceId using the mediaStream destination node, but the label in that mediaStream is not an input device label. We already store the current audioTrack in the state so we can use that one to get the correct input device label otherwise `lastUsedMicDeviceId` is always undefined.

### DIFF
--- a/src/react-components/ui-root.js
+++ b/src/react-components/ui-root.js
@@ -596,6 +596,8 @@ class UIRoot extends Component {
     let hasAudio = false;
     const { lastUsedMicDeviceId } = this.props.store.state.settings;
 
+    console.log(`Last used input device: ${lastUsedMicDeviceId}`);
+
     // Try to fetch last used mic, if there was one.
     if (lastUsedMicDeviceId) {
       hasAudio = await this.fetchAudioTrack({ audio: { deviceId: { ideal: lastUsedMicDeviceId } } });
@@ -684,12 +686,15 @@ class UIRoot extends Component {
     await this.fetchMicDevices();
 
     // we should definitely have an audioTrack at this point unless they denied mic access
-    if (this.state.mediaStream) {
-      const micDeviceId = this.micDeviceIdForMicLabel(this.micLabelForMediaStream(this.state.mediaStream));
+    if (this.state.audioTrack) {
+      const micDeviceId = this.micDeviceIdForMicLabel(this.micLabelForAudioTrack(this.state.audioTrack));
       if (micDeviceId) {
         this.props.store.update({ settings: { lastUsedMicDeviceId: micDeviceId } });
+        console.log(`Selected input device: ${this.micLabelForDeviceId(micDeviceId)}`);
       }
       this.props.scene.emit("local-media-stream-created");
+    } else {
+      console.log("No available audio tracks");
     }
   };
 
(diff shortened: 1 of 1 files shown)


src/react-components/ui-root.js
```diff
@@ -596,6 +596,8 @@ class UIRoot extends Component {
     let hasAudio = false;
     const { lastUsedMicDeviceId } = this.props.store.state.settings;
 
+    console.log(`Last used input device: ${lastUsedMicDeviceId}`);
```
johnshaughnessy, line 599: Can remove these console logs

src/react-components/ui-root.js
```diff
@@ -718,13 +723,16 @@ class UIRoot extends Component {
...
+        const inputDevices = mediaDevices
+          .filter(d => d.kind === "audioinput")
+          .map(d => ({ deviceId: d.deviceId, label: d.label }));
+        inputDevices.forEach(d => console.log(`\t${d.label} (${d.deviceId})`));
         this.setState(
           {
-            micDevices: mediaDevices
-              .filter(d => d.kind === "audioinput")
-              .map(d => ({ deviceId: d.deviceId, label: d.label }))
+            micDevices: inputDevices
```
johnshaughnessy, line 735: Can drop `inputDevices` inline the `filter`+`map` as it was before since you no longer need the `console.log`s above

src/react-components/ui-root.js
```diff
@@ -786,8 +798,8 @@ class UIRoot extends Component {
     const mediaStream = this.state.mediaStream;
 
     if (mediaStream) {
-      if (mediaStream.getAudioTracks().length > 0) {
-        console.log(`Using microphone: ${mediaStream.getAudioTracks()[0].label}`);
+      if (this.state.audioTrack) {
+        console.log(`Using microphone: ${this.state.audioTrack.label}`);
```
johnshaughnessy, line 802: Should we keep this console log since it was here before and is logging the "final state" of your microphone selection?

johnshaughnessy (approved):
lgtm! Sorry for the delay -- I missed this review request.

Going to share this PR with Robert. I think he'll be happy to see this bug fixed because we lost some time to it in the UI redesign work

johnshaughnessy:
Now that you have repo access -- could you use this repo as your main remote rather than your fork? It will be easier for people to pull down your branches locally than having them add your fork to their remotes.

src/react-components/ui-root.js
keianhzo: Removed.

src/react-components/ui-root.js
keianhzo: Updated

src/react-components/ui-root.js
keianhzo: Removed

src/react-components/ui-root.js
keianhzo: Not sure if I get this one, the log line is still there, do you mean removing that log?

keianhzo:
@johnshaughnessy I've updated removing some logs. I just wanted to make it as verbose as possible in case we need to debug audio issues.

PR #3118, opened 2020-10-03
Use audio listener for distance based attenuation

johnshaughnessy:
Same fix as https://github.com/mozilla/hubs/pull/3090 but for other media objects in the room.


takahirox (approved):
This change looks fine to me as #3090.

PR #3109, opened 2020-10-01
Spanish localization

keianhzo:
Spanish localization

PR #3146, opened 2020-10-09
Spanish localization from user

robin-k-wilson:
Spanish localization! To be looked over! 

https://github.com/mozilla/hubs/discussions/3108


robin-k-wilson:
Localized via this PR https://github.com/mozilla/hubs/pull/3109

PR #3148, opened 2020-10-09
Room UI Redesign: People Sidebar

robertlong:
This PR adds the people sidebar and cleans up the styling for editing your own profile and viewing other users.


src/react-components/room/PeopleSidebar.js
```diff
@@ -0,0 +1,141 @@
...
+            <ButtonListItem
+              className={styles.person}
+              key={person.id}
+              type="button"
+              aria-label={getLabel(person)}
+              onClick={e => onSelectPerson(person, e)}
+            >
+              {<DeviceIcon title={getDeviceLabel(person.context)} />}
+              {!person.context.discord && VoiceIcon && <VoiceIcon title={getVoiceLabel(person.micPresence)} />}
+              <p>{getPersonName(person)}</p>
```
johnshaughnessy, line 117: Would like to be able to select/copy this text

src/react-components/room/PeopleSidebarContainer.js
```diff
@@ -0,0 +1,142 @@
...
+}
+
+function PeopleListContainer({ hubChannel, presences, mySessionId, onSelectPerson, onClose }) {
+  const people = usePeopleList(presences, mySessionId);
+
+  const onMuteAll = useCallback(
+    () => {
+      for (const person of people) {
+        if (person.presence === "room" && person.permissions && !person.permissions.mute_users) {
+          hubChannel.mute(person.id);
```
johnshaughnessy, line 52: Not sure if I'm doing something wrong but I don't seem to be reaching the line. (Mute all doesn't seem to be working)

src/react-components/room/PeopleSidebarContainer.js
```diff
@@ -0,0 +1,142 @@
...
+  );
+}
+
+PeopleSidebarContainer.propTypes = {
+  displayNameOverride: PropTypes.string,
+  store: PropTypes.object.isRequired,
+  mediaSearchStore: PropTypes.object.isRequired,
+  hubChannel: PropTypes.object.isRequired,
+  history: PropTypes.object.isRequired,
+  onOpenAvatarSettings: PropTypes.func.isRequired,
```
johnshaughnessy, line 136: `onOpenAvatarSettings` appears to be unused / undefined

johnshaughnessy (approved):
lgtm!

A note that's unrelated to these changes but affects the new UI -- It seems like tabbing through the more menu popover is broken. I can tab through it if I hold tab, but as soon as I let go it seems to lose focus.

src/react-components/room/PeopleSidebarContainer.js
robertlong: Can you expand on this? Were you logged in? Was there another user in the room? Were they not muted?

src/react-components/room/PeopleSidebar.js
robertlong: Hmm, I've made these buttons for accessibility reasons. Buttons have a default behavior which is to not allow text selection. IMO this is correct. If copying a user's name is important we should figure out the best place to put a copy user name button. This could be a right click menu or in the user's profile.

robertlong:
Can you give me some more reproduction steps for the more menu bug? I can't figure out how to trigger that.

src/react-components/room/PeopleSidebarContainer.js
johnshaughnessy: Took me a bit to figure out -- the case I was running into was that I was logged in as the same user twice, so the check for `person.permissions.mute_users` prevented the mute from going thru. Is this behavior correct? (Is "mute all" not supposed to apply to admins / owners?)

johnshaughnessy:
Quick fixes for `onOpenAvatarSettings` and an issue preventing promotion / demotion in the people sidebar: https://github.com/mozilla/hubs/pull/3171/files

src/react-components/room/PeopleSidebar.js
johnshaughnessy: Having this row be a button seems good to me -- it makes keyboard navigation easier and allows screen readers to parse it correctly. Does making the text in the button selectable with a `user-select: text;` css rule impact its accessibility? Keyboard navigation / button behavior seems to remain in tact in [this example](https://stackoverflow.com/questions/47090073/how-will-make-the-button-text-selectable).

robertlong: It lets you select the text but it still triggers the button press. So you can't really copy the text. I think this is a relatively minor thing that we can tackle later.

src/react-components/room/PeopleSidebarContainer.js
robertlong: It doesn't currently check for that: https://github.com/mozilla/hubs/blob/master/src/react-components/presence-list.js#L195

PR #3151, opened 2020-10-10
Added custom email + auth tab + multiline

robin-k-wilson:
Added ability to customize magic link email subject + body.

- App Settings Auth tab
- Multi-line field

### DIFF
--- a/admin/src/utils/ita.js
+++ b/admin/src/utils/ita.js
@@ -10,7 +10,8 @@ const schemaCategories = [
   "rooms",
   "images",
   "theme",
-  "links"
+  "links",
+  "auth"
 ];
 const serviceNames = configs.CONFIGURABLE_SERVICES.split(",");
 let currentAuthToken = null;
@@ -41,6 +42,8 @@ function getCategoryDisplayName(category) {
       return "Theme";
     case "links":
       return "Links";
+    case "auth":
+      return "Auth";
     default:
       return null;
   }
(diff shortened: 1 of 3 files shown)


robin-k-wilson:
image

image

src/schema.toml
```diff
@@ -76,3 +76,6 @@ links.issue_report = { category = "links", type = "string", name = "Issue Report
 links.promotion = { category = "links", type = "string", name = "Promotion Info", description = "Link to info about promotion info and licensing."}
 links.remixing = { category = "links", type = "string", name = "Remixing Info", description = "Link to info about remixing info and licensing."}
 links.model_collection = { category = "links", type = "string", name = "Model Collection", description = "Link to a collection of recommended models."}
+
+auth.login_subject = { category = "auth", type="string", name="Magic Link Email Subject", description="Customize the email subject line for users logging in" }
+auth.login_body = { category = "auth", type="longstring", name="Magic Link Email Body", description="Customize message. Add '{{ link }}' to insert the magic link, otherwise it will be appended at the end." }
```
netpro2k, line 81: Not sure what all parses this, we need to make sure they are also able to handle this new "type"...

admin/src/utils/ita.js
```diff
@@ -70,6 +73,8 @@ function getCategoryDescription(category, provider) {
       return "Replace colors in the app.";
     case "links":
       return "Replace links in the app.";
+    case "auth":
```
netpro2k, line 76: I would have first thought to put this in email settings. Looks like you thought so too with the email settings hints. Seems find to do this for now but eventually we might want to rework admin panel so that storage location is not tied to where we are showing it in the UI.

netpro2k (commented):
lgtm solong as `longstring` wont confuse anything else current reading the appconfig.

Issue #2937, opened 2020-08-27
Set quality setting with query string parameters (again)


johnshaughnessy:
**Is your feature request related to a problem? Please describe.**
We used to set the quality setting with a query string parameter. I removed this feature when adding the preference screen, thinking that it was best left as an end-user feature, and didn't want there to be confusion on the part of the end user in cases where the query string parameter did not match the user pref.

**Describe the solution you'd like**
Re-enable the ability to set quality via query string parameter. If end user has a specific preference saved, I think the user pref should take priority, but if the user does not have any preferences, the query string parameter should take precedence.

**Describe alternatives you've considered**

**Additional context**
Crimson's team ran into this problem because they wanted people's default quality setting to be "low" instead of "high" but couldn't do this with the query string parameter.

Also update the docs when doing this : https://hubs.mozilla.com/docs/hubs-query-string-parameters.html

lue:
Hi, @johnshaughnessy

Sorry, I am a bit confused. Should ?default_material_quality=low string parameter work right now on hubs.mozilla.com ? I tried and it does not seem to force low settings.

PR #3179, opened 2020-10-15
Set quality setting with query string parameters (again) 

johnshaughnessy:
Fix https://github.com/mozilla/hubs/issues/2937 after a failed attempt that needed to be reverted (https://github.com/mozilla/hubs/pull/2959)

### DIFF
--- a/src/storage/store.js
+++ b/src/storage/store.js
@@ -2,6 +2,7 @@ import { Validator } from "jsonschema";
 import merge from "deepmerge";
 import Cookies from "js-cookie";
 import jwtDecode from "jwt-decode";
+import { qsGet } from "../utils/qs_truthy.js";
 
 const LOCAL_STORE_KEY = "___hubs_store";
 const STORE_STATE_CACHE_KEY = Symbol();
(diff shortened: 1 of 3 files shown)


src/storage/store.js
```diff
@@ -10,6 +11,34 @@ const validator = new Validator();
...
 import { fetchRandomDefaultAvatarId, generateRandomName } from "../utils/identity.js";
 
+export const defaultMaterialQualitySetting = (function() {
+  const MATERIAL_QUALITY_OPTIONS = ["low", "medium", "high"];
+  let isMobile;
+  try {
+    // HACK: AFRAME is not available on all pages, so we catch the ReferenceError.
+    // We could move AFRAME's device utils into a separate package (or into this repo)
+    // if we wanted to use these checks without having to import all of AFRAME.
+    isMobile = AFRAME.utils.device.isMobile() || AFRAME.utils.device.isMobileVR();
```
robertlong, line 21: Instead of wrapping this in a try/catch you could just check for `window.AFRAME`.

robertlong (commented):
Looks good, I'm happy we're adding this back in!

PR #3193, opened 2020-10-19
Room UI Redesign: Object List and Object Menu

robertlong:
This PR adds the object list and object menu functionality. I've tried to preserve all existing functionality and adapt it to the new UI. The only feature that I've removed is the ability to toggle the "lights" on and off while inspecting an object, but we could easily find a way to fit that into the design. IMO, there's still some work to be done on the UX. I'm interested in getting feedback on this feature.


src/react-components/room/useObjectList.js
```diff
@@ -0,0 +1,242 @@
...
+
+  useEffect(
+    () => {
+      function updateMediaEntities() {
+        const objects = scene.systems["listed-media"].els.sort(mediaSort).map(el => ({
+          id: el.object3D.id,
+          name: getDisplayString(el),
+          type: getMediaType(el),
+          el,
+          enableLights: false
```
brianpeiris, line 63: As far as I can tell, none of the codepaths change this to true. Previously, there was UI to toggle lights, which would set the "show-background-while-inspecting" flag in local storage, but this PR removes that as well.

In other words, there is no mechanism for a user to enable lights while inspecting an object through the object list. If it's not an omission in the UX, maybe we should remove this state from the `objects` array, and remove references to the local storage flag as well.

src/react-components/room/useObjectList.js
```diff
@@ -0,0 +1,242 @@
...
+    focusObject,
+    unfocusObject,
+    selectObject,
+    deselectObject,
+    selectPrevObject,
+    selectNextObject
+  };
+
+  // Note: If we move ui-root to a functional component and use hooks,
+  // we can use the useObjectList hook instead of cloneElement.
```
brianpeiris, line 226: Should this remain in the code, or is it a personal TODO for later?

src/react-components/room/object-hooks.js
```diff
@@ -0,0 +1,145 @@
...
+  return el.components.pinnable && el.components.pinnable.data.pinned;
+}
+
+function hasIsStaticTag(el) {
+  return el.components.tags && el.components.tags.data.isStatic;
+}
+
+export function getPlayerInfo(object) {
+  return object.el.components["player-info"] && object.el.components["player-info"].data;
+}
```
brianpeiris, line 16: This is only used to determine if an object is a player. Maybe just call it `isPlayer` and have it strictly return a boolean.

src/react-components/room/object-hooks.js
```diff
@@ -0,0 +1,145 @@
...
+import { removeNetworkedObject } from "../../utils/removeNetworkedObject";
+import { rotateInPlaceAroundWorldUp, affixToWorldUp } from "../../utils/three-utils";
+import { getPromotionTokenForFile } from "../../utils/media-utils";
+
+function getPinnedState(el) {
+  return el.components.pinnable && el.components.pinnable.data.pinned;
+}
+
+function hasIsStaticTag(el) {
+  return el.components.tags && el.components.tags.data.isStatic;
```
brianpeiris, line 11:
```suggestion
  return !!(el.components.tags && el.components.tags.data.isStatic);
```

src/react-components/room/object-hooks.js
```diff
@@ -0,0 +1,145 @@
+import { useEffect, useState, useCallback, useMemo } from "react";
+import { removeNetworkedObject } from "../../utils/removeNetworkedObject";
+import { rotateInPlaceAroundWorldUp, affixToWorldUp } from "../../utils/three-utils";
+import { getPromotionTokenForFile } from "../../utils/media-utils";
+
+function getPinnedState(el) {
+  return el.components.pinnable && el.components.pinnable.data.pinned;
```
brianpeiris, line 7:
```suggestion
  return !!(el.components.pinnable && el.components.pinnable.data.pinned);
```

src/react-components/room/ObjectsSidebarContainer.js
```diff
@@ -0,0 +1,42 @@
...
+            key={object.id}
+            onClick={() => selectObject(object)}
+            onMouseOver={() => focusObject(object)}
+            onFocus={() => focusObject(object)}
+            onMouseOut={onUnfocusListItem}
+            onBlur={onUnfocusListItem}
+          />
+        ))}
+      </List>
+    </ObjectsSidebar>
```
brianpeiris, line 36: Design nit: It would be nice if the design made it clear that the objects list is empty. I'm thinking something like a large centered message, maybe with iconography, that says "There are no objects in the room". Right now you're just presented with a blank sidebar except for the title and the close button and close button at the top. It's especially important on mobile, where the blank space of the sidebar takes up the entire screen.

brianpeiris (commented):
No major comments or blockers, so looks good to me otherwise. Nice work!

src/systems/userinput/devices/mouse.js
```diff
@@ -62,8 +62,10 @@ export class MouseDevice {
     const left = event.button === 0;
     const middle = event.button === 1;
     const right = event.button === 2;
-    this.coords[0] = (event.clientX / window.innerWidth) * 2 - 1;
-    this.coords[1] = -(event.clientY / window.innerHeight) * 2 + 1;
+    // Note: This assumes the canvas always starts in the top left.
+    // This works with the current sidebar and toolbar layout.
+    this.coords[0] = (event.clientX / this.canvas.clientWidth) * 2 - 1;
+    this.coords[1] = -(event.clientY / this.canvas.clientHeight) * 2 + 1;
```
johnshaughnessy, line 68: May need to update similar lines for touchscreens:
```
./src/systems/userinput/devices/app-aware-touchscreen.js:46:      x: (touch.clientX / window.innerWidth) * 2 - 1,
./src/systems/userinput/devices/app-aware-touchscreen.js:174:          (touch.clientX / window.innerWidth) * 2 - 1,
./src/systems/userinput/devices/app-aware-touchscreen.js:234:        (touch.clientX / window.innerWidth) * 2 - 1,
./src/systems/userinput/devices/app-aware-touchscreen.js:355:      frame.setVector2(path.touchCameraDelta, delta[0] / window.innerWidth, delta[1] / window.innerHeight);
```
It probably won't affect phones because the panels are full-screen takeovers but for bigger touchscreens like ipads, not updating it may cause issues.

johnshaughnessy (approved):
lgtm!

I left one comment about the "enable lights" preference. (I can't find it now -- I think it's outdated by more recent changes). Personally, I almost always prefer to have the "lights on" (keep the background shown) and only toggle them off only when the object I'm looking at is hard to see otherwise. I think it's good functionality for us to keep, but I'm not sure how best to slot it in because 1) it's not really an action you perform _on_ a particular object and 2) currently there's only one preference that controls the behavior for objects, avatars, and whether you focus on those from the menu or in-world.

Now that the menus appear when holding right click / f to inspect objects/avatars in-world, and we have on-screen ui to close the menu, it may be good to change the interaction to a toggle rather than requiring you to hold the button (on mouse/keyboard/touchscreen only, since the interface will not be there in vr). We could toggle it off either by pressing the x or repeating input that toggled it on. (Press f or right click a second time)

src/react-components/room/useObjectList.js
robertlong: If you inspect an object using the in-world ui, the lights are enabled, if you inspect an object from the object list then they are not enabled. I didn't want to change that behavior so this is how I implemented that.

src/react-components/room/useObjectList.js
robertlong: Yeah, I'm working towards this, but it could take a while and may even happen after we ship the redesign. I'd rather keep this in code so people know why I'm using `cloneElement` here in case they run across it before things change.

src/systems/userinput/devices/mouse.js
robertlong: Ah, thanks. I don't actually have a way to test on my iPad due to the secure websocket connection and locally signed cert. I'll make the change and I guess we can test on the staging server?

robertlong:
Idk if I want the "lights" on or off when focusing on an object, but overall I find the focusing on objects UX very confusing. I don't think we want the 3D orbiting behavior for 2D media, only 3D media. Having a way to full screen 2D media like videos would be nice. I'd like to have a larger design discussion and I'll add this to our discussion topics with DPX.

src/react-components/input/IconButton.js
```diff
@@ -5,10 +5,15 @@ import styles from "./IconButton.scss";
...
 export const IconButton = memo(
-  forwardRef(({ className, as: ButtonComponent, children, ...rest }, ref) => {
+  forwardRef(({ className, as: ButtonComponent, compactSm, children, ...rest }, ref) => {
     return (
       <ButtonComponent
-        className={classNames(styles.iconButton, textInputStyles.iconButton, className)}
+        className={classNames(
+          styles.iconButton,
+          textInputStyles.iconButton,
+          { [styles.compactSm]: compactSm },
```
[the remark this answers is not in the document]
robertlong: I added a comment to the prop types.

src/react-components/room/ObjectsSidebar.js
```diff
@@ -0,0 +1,84 @@
...
+  }
+}
+
+const objectTypeNames = {
+  video: "Video",
+  audio: "Audio",
+  image: "Image",
+  pdf: "PDF",
+  model: "Model",
+  default: "Object"
```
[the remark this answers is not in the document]
robertlong: Yeah, I will in a future i18n focused PR for the redesign.

src/react-components/room/ObjectMenuContainer.js
```diff
@@ -0,0 +1,78 @@
...
+      objectCount={objects.length}
+      onClose={deselectObject}
+      onBack={deselectObject}
+      onNextObject={selectNextObject}
+      onPrevObject={selectPrevObject}
+    >
+      {canPin && (
+        <ObjectMenuButton onClick={togglePinned}>
+          <PinIcon />
+          <span>{isPinned ? "Unpin" : "Pin"}</span>
```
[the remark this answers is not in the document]
robertlong: I will make sure to do that in the i18n PR

src/systems/userinput/devices/mouse.js
robertlong: I made the change. Idk if it works or not though. I don't have a way to test it.

src/react-components/room/ObjectMenuContainer.js
```diff
@@ -0,0 +1,78 @@
...
+        >
+          <DeleteIcon />
+          <span>Delete</span>
+        </ObjectMenuButton>
+      )}
+      {isPlayer && (
+        <ObjectMenuButton onClick={onOpenProfile}>
+          <AvatarIcon />
+          <span>Edit Avatar</span>
+        </ObjectMenuButton>
```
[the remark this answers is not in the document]
brianpeiris: I think you might be mistaken. It is possible to inspect other avatars at the moment, you just right-click on them. The best way is probably to add a `isLocalPlayer` function that uses the `isLocalPlayerInfo` property from the `player-info` component.

src/react-components/room/useObjectList.js
brianpeiris: Ah, I get it now.

PR #3206, opened 2020-10-22
Room UI Redesign: Toolbar Buttons / Menus

robertlong:
This PR includes the functionality for the share popover, place popover, reaction/emoji menu, and leave button.

I'm not a huge fan of the way the reaction/emoji menu turned out. The original plan for phase one of the redesign was to use the existing menu but this puts you into freeze mode. I think we should go with this concept instead:

image

When you click one of the emojis, it'll spawn in the world and you can still grab it.


src/react-components/room/PlacePopover.js
```diff
@@ -5,19 +5,20 @@ import { Popover } from "../popover/Popover";
 import { ToolbarButton } from "../input/ToolbarButton";
 import { ReactComponent as ObjectIcon } from "../icons/Object.svg";
 
-export function PlacePopoverButton({ items, onSelect }) {
+export function PlacePopoverButton({ items }) {
```
johnshaughnessy, line 8: I've mentioned this before but at some point I hope we find a clearer word for this menu than "Place". I suspect "Place" will give people the wrong impression of it being about "this place we're in" (aka the scene), not "place a pen in your hand or an object in the scene". Some alternatives to consider: 
- Add
- Create
- Load
- New
- Spawn
- Generate

Similarly, I think "Share" is overloaded in the mobile world and we might want to consider something like, "Broadcast" or "Stream"

src/systems/pen-tools.js
```diff
@@ -0,0 +1,43 @@
+import { waitForDOMContentLoaded } from "../utils/async-utils";
+
+// Used for tracking and managing pen tools in the scene
+AFRAME.registerSystem("pen-tools", {
```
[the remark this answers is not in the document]
johnshaughnessy: You may also need to listen to alternative event for knowing when to update the place UI: 
https://github.com/mozilla/hubs/blob/redesign/toolbar-buttons/src/react-components/room/PlacePopoverContainer.js#L100

src/react-components/room/ReactionButtonContainer.js
```diff
@@ -0,0 +1,43 @@
...
+
+      return () => {
+        scene.removeEventListener("stateadded", onSceneStateChange);
+        scene.removeEventListener("stateremoved", onSceneStateChange);
+      };
+    },
+    [scene]
+  );
+
+  // TODO: We probably shouldn't use freeze mode for users spawning emojis on a 2d device.
```
johnshaughnessy, line 28: +1 - Good idea to either implement a 2D emoji menu or separate the logic that ties their visibility to the frozen state. https://github.com/mozilla/hubs/blob/7159d67ccc84d6dfccf7bf3b1590275f04092367/src/components/emoji-hud.js#L88-L91

src/react-components/room/useMicrophone.js
```diff
@@ -1,17 +1,17 @@
-import { useState, useEffect, useRef } from "react";
+import { useState, useEffect, useRef, useCallback } from "react";
 import MovingAverage from "moving-average";
 
-const MOVING_AVG_TIMESPAN = 100;
 const UPDATE_RATE = 50;
 
-export function useMicrophoneVolume(scene) {
+export function useMicrophone(scene, updateRate = UPDATE_RATE) {
```
johnshaughnessy, line 6: Should this line below use `updateRate` (instead of `UPDATE_RATE`)?
https://github.com/mozilla/hubs/blob/7159d67ccc84d6dfccf7bf3b1590275f04092367/src/react-components/room/useMicrophone.js#L29

Seems like you can remove `UPDATE_RATE` and write `updateRate = 50`.

src/react-components/room/ReactionButtonContainer.js
```diff
@@ -0,0 +1,43 @@
+import React, { useEffect, useState, useCallback } from "react";
+import PropTypes from "prop-types";
+import { ReactComponent as ReactionIcon } from "../icons/Reaction.svg";
+import { ToolbarButton } from "../input/ToolbarButton";
+
+export function ReactionButtonContainer({ scene }) {
```
johnshaughnessy, line 6: We'll need to do some thinking around the UX for "send to 3D world" vs "send to chat" vs "send to both" . It is surprising to me that there's a "react" (emoji) menu in the toolbar rather than in the chat sidebar or the "Place" popover menu

johnshaughnessy (approved):
lgtm. Left some suggestions but didn't find any major issues

src/react-components/room/ReactionButtonContainer.js
johnshaughnessy: I like the suggested design of the emoji popover menu as a first (or second) pass. I think eventually we'll want something integrated into the chat sidebar and easily extended / searchable like slack or discord's emoji menu:
image
image

johnshaughnessy: I especially like that when Discord's emoji button is hovered, it will change the emoji each time. I would like to do the same for our emoji button. Maybe it would fit for our "Place" menu too.

src/components/mute-mic.js
```diff
@@ -55,7 +53,6 @@ AFRAME.registerComponent("mute-mic", {
     if (!NAF.connection.adapter) return;
     if (!this.el.sceneEl.is("entered")) return;
 
-    this.el.sceneEl.systems["hubs-systems"].soundEffectsSystem.playSoundOneShot(SOUND_TOGGLE_MIC);
```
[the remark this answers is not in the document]
robertlong: Can we play this toggle sound just in VR? IMO it doesn't make sense in the 2D UI. None of the other buttons make noise and other 2D apps don't have sounds bound to their 2D UI buttons.

src/systems/pen-tools.js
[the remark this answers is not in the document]
robertlong: I like the current implementation because it allows for removing a pen that you own but have dropped. Also this code is a direct copy of the `camera-tools` system.

src/react-components/room/PlacePopover.js
robertlong: I'll bring this up with DPX. Thanks for flagging this :+1:

robertlong:
I want to cover the reactions/emojis issues in detail in another PR. But I agree with all of these concerns.

I've created two issues to cover the Reactions menu and emojis in chat. I think these are two separate issues and deserve two separate spots in the UI. However, we need to discuss how each of them will work.

https://github.com/mozilla/hubs/issues/3232
https://github.com/mozilla/hubs/issues/3233

src/components/mute-mic.js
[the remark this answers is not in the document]
johnshaughnessy: Discord makes a sound when you mute/unmute your mic (by clicking the button or using a hotkey). I think it's helpful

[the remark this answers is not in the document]
robertlong: Hmm, good point. I'll add it back.

PR #3209, opened 2020-10-23
Landing page panels

brianpeiris:
Adds panels to the landing page describing Hubs' core features.

The panels are behind an internal feature flag and disabled by default. The images in the panels must also be configured using internal app settings (accessed via `/admin#/app-settings?show_internal_configs=true`). The panels are hidden if there would otherwise be user-favorited rooms or public rooms on the landing page.

firefox_CSNigwOd2a


.eslintignore
```diff
@@ -1,3 +1,4 @@
 !.eslintrc.js
 src/vendor/*
+src/loaders/basis_transcoder.worker.js
```
brianpeiris, line 3: Snuck this in here. This worker is a large, minified vendor file. It adds 40% more time to the eslint check, so it's best to ignore it.

src/assets/locales/en.json
```diff
@@ -240,6 +240,12 @@
   "home.terms_of_use": "Terms of Use",
   "home.made_with_love": "made with 🦆 by ",
   "home.environment_author_by": " by ",
+  "home.rooms-title": "Instantly create rooms",
```
robin-k-wilson, line 243: I think these should use underscores "_" rather than dashes "-".

src/assets/locales/es.json
```diff
@@ -240,6 +240,12 @@
   "home.terms_of_use": "Condiciones de uso",
   "home.made_with_love": "hecho con 🦆 por ",
   "home.environment_author_by": " por ",
+  "home.rooms-title": "Instantly create rooms",
+  "home.rooms-blurb": "Share virtual spaces with your friends, co-workers, and communities. When you create a room with Hubs, you’ll have a private virtual meeting space that you can instantly share - no downloads or VR headset necessary.",
```
robin-k-wilson, line 244: "_" -> "-"

src/assets/locales/jp.json
```diff
@@ -240,6 +240,12 @@
   "home.terms_of_use": "利用規約",
   "home.made_with_love": "madewith🦆by",
   "home.environment_author_by": "by",
+  "home.rooms-title": "Instantly create rooms",
+  "home.rooms-blurb": "Share virtual spaces with your friends, co-workers, and communities. When you create a room with Hubs, you’ll have a private virtual meeting space that you can instantly share - no downloads or VR headset necessary.",
```
robin-k-wilson, line 244: "_" -> "-"

src/assets/locales/pt-br.json
```diff
@@ -240,6 +240,12 @@
   "home.terms_of_use": "Termos de uso",
   "home.made_with_love": "feito com 🦆 por ",
   "home.environment_author_by": " por ",
+  "home.rooms-title": "Instantly create rooms",
+  "home.rooms-blurb": "Share virtual spaces with your friends, co-workers, and communities. When you create a room with Hubs, you’ll have a private virtual meeting space that you can instantly share - no downloads or VR headset necessary.",
```
robin-k-wilson, line 244: "_" -> "-"

src/assets/locales/zh.json
```diff
@@ -240,6 +240,12 @@
   "home.terms_of_use": "使用条款",
   "home.made_with_love": "由 🦆 制作 ",
   "home.environment_author_by": " 通过 ",
+  "home.rooms-title": "Instantly create rooms",
```
robin-k-wilson, line 243: "_" -> "-"

src/assets/locales/en.json
```diff
@@ -240,6 +240,12 @@
...
+  "home.rooms-blurb": "Share virtual spaces with your friends, co-workers, and communities. When you create a room with Hubs, you’ll have a private virtual meeting space that you can instantly share - no downloads or VR headset necessary.",
+  "home.communicate-title": "Communicate naturally",
+  "home.communicate-blurb": "Choose an avatar to represent you, put on your headphones, and jump right in. Hubs makes it easy to stay connected with voice and text chat to other people in your private room.",
```
johnshaughnessy, line 246: I'm a bit concerned that the "put on your headset" part will turn non-VR users away, but it may be OK because `rooms-blurb` says "no downloads or VR headset necessary"

src/schema.toml
```diff
@@ -36,6 +36,7 @@ features.require_account_for_join = { category = "rooms", type = "boolean", name
 features.default_room_size = { category = "rooms", type = "number", name = "Default room size", description = "Default room size for new rooms. This does not include users in the lobby." }
 features.max_room_size = { category = "rooms", type = "number", name = "Maximum room size", description = "Maximum room size visitors can set." }
 
+features.show_feature_panels = { category = "features", type = "boolean", internal = "true" }
```
robin-k-wilson, line 39: Thanks for making this a toggle in the Admin panel! Was my first thought looking at the feature.

johnshaughnessy (approved):
lgtm! Thanks for doing this

src/schema.toml
```diff
@@ -53,6 +54,9 @@ images.company_logo = { category = "images", type = "file", name = "Company Logo
 images.home_background = { category = "images", type = "file", name = "Home background", description = "Background image on the home page." }
 
 images.editor_logo = { category = "images", type = "file", internal = "true" }
+images.landing_rooms_thumb = { category = "images", type = "file", internal = "true" }
+images.landing_communicate_thumb = { category = "images", type = "file", internal = "true" }
+images.landing_media_thumb = { category = "images", type = "file", internal = "true" }
```
robin-k-wilson, line 59: Should people be able to customize the title/text as well? We now have a "longstring" type. 

If not in this PR, I'd create an issue for community members to vote if they want to customize the text/titles on HC for the 3 panel homepage.

src/assets/locales/en.json
brianpeiris: Oh, nice catch.

src/assets/locales/en.json
johnshaughnessy: lol, yep. :facepalm:

src/schema.toml
brianpeiris: It's a good thought, but this particular PR is specifically meant for HMC, which is why it's disabled by default and the flags are internal.
I think it's a good idea to consider enabling something like this for HC, but that's a bit of a bigger discussion around the whole redesign and general site customization that we're intentionally side-stepping in this PR.

src/assets/locales/en.json
misslivirose: John are you referring to the "put on your headphones" part? The 'No VR headset necessary' phrase was to try and alleviate concerns about needing a headset when we use the terms 'social VR', so we definitely want to make that clear.

Issue #3220, opened 2020-10-24
UV scroll with normal map causes graphics corruption over time


zach-capalbo:
**Description**

When an object with a `uv-scroll` component and a normal map is left playing for a while, little black boxes begin to appear on the object. The frequency of the appearance of black boxes increases with time, and increases the closer the viewer is to the object.

If you call `el.components['uv-scroll'].pause(); el.components['uv-scroll'].play();` on the corresponding entity, the black boxes go away. They will however return if the `uv-scroll` is left running for any considerable amount of time.

Here's an example of this, `increment: {x: 0.1666, y: 0.0}, speed: {x: 2.666, y: 0.0}`

Upon entering scene: (12:42 PM)

![image](https://user-images.githubusercontent.com/1325621/97087208-53fa9000-15f6-11eb-9260-18ed3a45f2b8.png)

After 1 minute has elapsed: (12:43 PM)

![image](https://user-images.githubusercontent.com/1325621/97087226-860bf200-15f6-11eb-9a73-ec329bf726a9.png)

After 3 minutes has elapsed: (12:45 PM)

![image](https://user-images.githubusercontent.com/1325621/97087256-d1260500-15f6-11eb-8358-79e719c28fae.png)

After using browser developer console to call: (12:47 PM) `$('*[uv-scroll]').components["uv-scroll"].pause(); $('*[uv-scroll]').components["uv-scroll"].play()`

![image](https://user-images.githubusercontent.com/1325621/97087268-ec911000-15f6-11eb-9673-b59becd96bcf.png)

After 1 minute after play and pause:

![image](https://user-images.githubusercontent.com/1325621/97087307-2e21bb00-15f7-11eb-88b3-2d2b0b12910d.png)

**To Reproduce**
1. Import glb file with mesh with `uv-scroll` and a normal map
2. Wait

**Expected behavior**
1. Should not have black boxes

**Screenshots**
See Description

**Hardware** 
 - Device: Desktop and VR
 - OS: Confirmed on Windows
 - Browser Chromium and Firefox

**Additional context**
I'm working on a possible fix for this and will submit a PR if I find one.

j-conrad:
Thanks for reporting this one. We’ve noted this before— it’s due to floating point errors as the uv coordinates get bigger and bigger and lose accuracy. We need to change the code to have it ‘loop back’ to the starting point instead of scrolling to infinity. It’s fascinating yet annoying. 😆

brianpeiris:
[...]

PR #3221, opened 2020-10-24
Fix uv-scroll graphical corrution

zach-capalbo:
Fix issue where large offset cause graphical corruption.

Fixes issue #3220


netpro2k:
Thanks for the detailed bug report and this fix! Will get it deployed this week.

PR #3238, opened 2020-10-28
Reparenting audioListener immediately after camera mode is updated

### LINKED ISSUE #3098: Sometimes audio noise can happen at the moment when zooming in an object
**Description**

Sometimes, not always, audio noise can happen at the moment when zooming in an object if someone in the room is speaking.

**To Reproduce**

* Enter a room
* Run a bot with audio `node run-bot.js -a bot-recording.mp3 -r roomId`
* Place an asset like a picture anywhere in the room
* Press f on the asset and zoom in
* Confirm that an audio noise can happen. If it doesn't, repeatedly press until it happens

If the audio noise doesn't happen, move close to or far from the bot and try again. It seems this problem very depends on the distance from the speaker or audio volume.

**Expected behavior**

Audio noise doesn't happen.

**Hardware** 
 - Device: [e.g. Desktop, phone, VR headset] Desktop
 - OS: [e.g. Windows, iOS, Linux] Windows
 - Browser [e.g. Firefox] Firefox and Chrome

**Additional context**

* In my environment, it happens both on Firefox and Chrome but I feel it happens more easily on Chrome? No confidence tho.
* As I wrote in "To reproduce", this problem seems very depending on the distance from the speaker or audio volume. In my test, it may happen more easily if I am closer to the bot?
* It doesn't seem to happen at the moment when you are back to normal from zoom in mode.

takahirox:
Resolves #3098. Refer to https://github.com/mozilla/hubs/issues/3098#issuecomment-703917282 for the detail.

### DIFF
--- a/src/systems/camera-system.js
+++ b/src/systems/camera-system.js
@@ -411,6 +411,19 @@ export class CameraSystem {
         this.nextMode();
       }
 
+      if (scene.audioListener && this.avatarPOV) {
+        if (this.mode === CAMERA_MODE_INSPECT && scene.audioListener.parent !== this.avatarPOV.object3D) {
+          this.avatarPOV.object3D.add(scene.audioListener);
+        } else if (
+          (this.mode === CAMERA_MODE_FIRST_PERSON ||
+            this.mode === CAMERA_MODE_THIRD_PERSON_NEAR ||
+            this.mode === CAMERA_MODE_THIRD_PERSON_FAR) &&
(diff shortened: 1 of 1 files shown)


takahirox:
I didn't notice that `inspect` and `uninsect` are called from out of camera system tick. This current change is not good enough. I'll update soon.

takahirox:
Let me close this PR for now because I noticed that the root issue may be different. I'll make another PR or reopen if I figure out.

PR #3239, opened 2020-10-28
Room UI Redesign: Room Settings Sidebar

robertlong:
This PR migrates the Room Settings Dialog to a sidebar and uses the new design system. This PR also adds the NumericInput and RadioInput components which are used in the sidebar.


src/react-components/input/ToggleInput.js
```diff
@@ -4,10 +4,15 @@ import classNames from "classnames";
 import styles from "./ToggleInput.scss";
 
 export const ToggleInput = memo(
-  forwardRef(({ className, label, value, disabled, ...rest }, ref) => {
+  forwardRef(({ className, label, description, value, disabled, ...rest }, ref) => {
     return (
       <label className={classNames(styles.toggleInput, { [styles.disabled]: disabled }, className)}>
-        {label && <span className={styles.label}>{label}</span>}
+        {label && (
```
johnshaughnessy, line 10: May be nice to vertically align the toggle inputs. Right now it looks a bit awkward to have each at its own distance from the right hand side:
image

src/react-components/input/NumericInput.js
```diff
@@ -0,0 +1,42 @@
...
+import styles from "./NumericInput.scss";
+import { TextInput } from "./TextInput";
+import { ReactComponent as CaretUpIcon } from "../icons/CaretUp.svg";
+import { ReactComponent as CaretDownIcon } from "../icons/CaretDown.svg";
+
+export const NumericInput = memo(
+  forwardRef(({ className, afterInput, disabled, value, onChange, ...rest }, ref) => {
+    return (
+      <TextInput
+        inputmode="numeric"
```
johnshaughnessy, line 13: `inputmode` -> `inputMode` since react prefers camelCase DOM attributes (https://reactjs.org/docs/dom-elements.html#all-supported-html-attributes)

johnshaughnessy (approved):
lgtm! What's your opinion on combining `room info` and `room settings` into a single panel?

src/react-components/input/NumericInput.js
robertlong: Thanks!

src/react-components/input/ToggleInput.js
robertlong: I flipped the toggle input layout so that the buttons are all aligned and the text is now to the right.

robertlong:
Thanks for the review! I think I want to combine them and add a change scene button so we can remove a couple menu items from the more menu. I'll do that with another pass of refactoring in another PR.

PR #3243, opened 2020-10-28
Room UI Redesign: Room Info Sidebar

robertlong:
This PR migrates the room info dialog to the new redesign.


src/react-components/room-info-dialog.js
```diff
@@ -1,118 +1,95 @@
...
+
+// TODO: Move to /room folder
+export function RoomInfoDialog({ scene, store, hubName, hubDescription, onClose }) {
+  const hasScene = !!scene;
+  const showSceneLink = hasScene && allowDisplayOfSceneLink(scene, store);
+  const attributions = (scene && scene.attributions && scene.attributions.content) || [];
+  const creator = scene && scene.attributions && scene.attributions.creator;
+
+  return (
+    <Sidebar title={<FormattedMessage id="room-info.title" />} beforeTitle={<CloseButton onClick={onClose} />}>
```
johnshaughnessy, line 48: Here I think you want to use the `intl.formatMessage` function to get a string instead of the object created by putting `FormatMessage` in jsx. https://formatjs.io/docs/react-intl/api/#usage

src/react-components/room-info-dialog.js
```diff
@@ -1,118 +1,95 @@
...
+      <div className={styles.roomInfoDialog}>
+        <h2 className={styles.sectionTitle}>Room</h2>
+        {hubName && <InputField label="Name">{hubName}</InputField>}
+        {hubDescription && <InputField label="Description">{hubDescription}</InputField>}
```
johnshaughnessy, line 52: The text in this panel seems to be inheriting a value for `user-select` that isn't `text` -- or something else is preventing text highlighting/selection

johnshaughnessy (approved):
lgtm!

src/react-components/room-info-dialog.js
robertlong: You can actually pass react elements in as props. I changed the `title` propType to `PropTypes.node` so that it stops warning on this. I think we should probably use the `<FormattedMessage />` unless you explicitly require a string or if you need to use `defineMessages`. Perhaps you ware confused by the `title` prop though? It's not setting the browser's `title` prop. It's actually the title of the sidebar.

src/react-components/room-info-dialog.js
robertlong: I found the culprit in the old `.ui` class and removed it. This makes a lot more things than I would like selectable though. I think we need to introduce styles/components for non-selectable labels, headings, images, and links. I think it's probably safer to opt-out of text selection than opt-in though. I'll follow up this PR with one to address this.

PR #3246, opened 2020-10-29
Room UI Redesign: Add inspect other player behavior

robertlong:
@brianpeiris [brought up that you can inspect other players by right clicking](https://github.com/mozilla/hubs/pull/3193#discussion_r512844982). I was unaware of that at the time. This PR fixes that behavior and adds a hide button when inspecting other players.

### DIFF
--- a/src/react-components/room/ObjectMenuContainer.js
+++ b/src/react-components/room/ObjectMenuContainer.js
@@ -2,14 +2,23 @@ import React from "react";
 import PropTypes from "prop-types";
 import { ObjectMenu, ObjectMenuButton } from "./ObjectMenu";
 import { useObjectList } from "./useObjectList";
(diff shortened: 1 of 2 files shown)


johnshaughnessy (commented):
I see the hide avatar button in the menu while inspecting an avatar, but it's not really possibly to interact with it -- should a button press toggle inspect instead of the button to be held down?

It seems like the inspect mode should be triggered when you click an avatar in the menu (so that the behavior is the same between the object list and avatar list). Is the idea to move the buttons from the sidebar into the menu in the main part of the screen?

robertlong:
@johnshaughnessy Idk. I agree it's a weird UX. I think we need to decide on one place to put the details. Inspect mode seems like a better option to me. Personally I'd like to merge this so we can show this behavior to DPX on staging and come up with a better solution.

PR #3249, opened 2020-10-30
Room UI Redesign: Auto Exit Modal and Exited Screen

robertlong:
This PR migrates the Auto Exit Modal and Exited Screen to the new design. It is dependent on https://github.com/mozilla/hubs/pull/3245


src/react-components/ui-root.js
```diff
@@ -295,7 +295,7 @@ class UIRoot extends Component {
 
   onConcurrentLoad = () => {
     if (qsTruthy("allow_multi") || this.props.store.state.preferences["allowMultipleHubsInstances"]) return;
-    this.startAutoExitTimer("autoexit.concurrent_subtitle");
+    this.startAutoExitTimer("concurrent-session");
```
johnshaughnessy, line 298: Looks like this should be `"concurrentSession"`

https://github.com/mozilla/hubs/blob/redesign/exit-modals/src/react-components/room/AutoExitWarningModal.js#L9

Any thoughts on how we might be able to catch these (especially if they're infrequently used) at build time, given that the message id's will not change?

src/react-components/room/ExitedRoomScreen.js
```diff
@@ -1,75 +1,110 @@
...
+  },
+  closed: {
+    id: "exit.subtitle.closed",
+    defaultMessage: "This room is no longer available."
+  },
+  denied: {
+    id: "exit.subtitle.denied",
+    defaultMessage: "You are not permitted to join this room. Please request permission from the room creator."
+  },
+  disconnected: {
```
johnshaughnessy, line 19: Would be good to distinguish "you disconnected" from "you've been kicked", even if the ability to rejoin the room is the same in both cases. (This isn't a new change with this PR)

johnshaughnessy (approved):
lgtm besides the one issue I mentioned in a comment

src/react-components/ui-root.js
robertlong: Types :sweat_smile: But I think a string enum object will do the trick here.

src/react-components/room/ExitedRoomScreen.js
robertlong: Agreed. There's also a comment in hub.js that mentions this: https://github.com/mozilla/hubs/blob/master/src/hub.js#L1134

PR #3251, opened 2020-10-30
Room UI Redesign: Remove Help and Support Dialogs

robertlong:
This PR removes the Help and Support Dialogs. The help links are now in the more menu and they don't require an in-app modal. And the Support Dialog was used to invite a team member to the room. This has been defunct for some time now so I think it's safe to remove.


johnshaughnessy (approved):
Nice, I love to see the balance in the red! :laughing: 
image

And wow, I forgot all about the `invite-team-dialog`...

PR #3255, opened 2020-10-31
Room UI Redesign: Leave Room Modal

robertlong:
This PR migrates the Leave Room Modal to the new design. It is dependent on https://github.com/mozilla/hubs/pull/3245


src/react-components/ui-root.js
```diff
@@ -1306,7 +1306,7 @@ class UIRoot extends Component {
             label: "Create Room",
             icon: HomeIcon,
             onClick: () =>
-              this.showNonHistoriedDialog(LeaveRoomDialog, {
+              this.showNonHistoriedDialog(LeaveRoomModal, {
```
johnshaughnessy, line 1309: Might also be good to flow thru this modal as a confirmation step for visiting rooms via a room link. I've accidentally clicked one before and it would have been nice to back out of it.
image

src/react-components/room/LeaveRoomModal.js
```diff
@@ -0,0 +1,45 @@
...
+});
+
+export function LeaveRoomModal({ messageType, destinationUrl, onClose }) {
+  const intl = useIntl();
+
+  return (
+    <Modal title="Leave Room" beforeTitle={<CloseButton onClick={onClose} />} contentClassName={styles.leaveRoomModal}>
+      <p>{intl.formatMessage(messages[`${messageType}-message`])}</p>
+      <Button as="a" preset="cancel" href={destinationUrl} rel="noopener noreferrer">
+        {intl.formatMessage(messages[`${messageType}-confirm`])}
```
johnshaughnessy, line 35: The hover colors seem a bit off for this button. The button remains the same color and the text turns gray. Should instead the button change color and the text stay the same (like for other buttons)?

src/react-components/room/LeaveRoomModal.js
```diff
@@ -0,0 +1,45 @@
...
+    id: "leave-room-dialog.join-room.confirm",
+    defaultMessage: "Join Room"
+  },
+  "create-room-message": {
+    id: "leave-room-dialog.create-room.message",
+    defaultMessage: "Creating a new room will leave this one. Are you sure?"
+  },
+  "create-room-confirm": {
+    id: "leave-room-dialog.create-room.confirm",
+    defaultMessage: "Leave Room"
```
johnshaughnessy, line 24: The title and confirmation button text for this case are awkward. I know this is taken from what we already have, but it seems like the button text or title should be something like "Create Room" or possibly "Leave and Create New Room"

johnshaughnessy (approved):

src/react-components/ui-root.js
robertlong: Hmm, that would break the flow of things like Chamberlain's lighthouse experience though. I don't know that I agree with this, but I'd be happy to follow up discussion in another issue.

src/react-components/room/LeaveRoomModal.js
robertlong: On my end, the button hover color turns a little lighter and the text does stay the same. Are you seeing something else?

johnshaughnessy: I can't reproduce this -- Maybe I messed something up locally :thinking:

PR #3264, opened 2020-11-02
Room UI Redesign: Promote Client Modal

robertlong:
Migrate Promote Client Dialog to Promote Client Modal. Dependent on https://github.com/mozilla/hubs/pull/3245


src/react-components/room/PromoteClientModal.js
```diff
@@ -0,0 +1,30 @@
...
+import { CloseButton } from "../input/CloseButton";
+
+export function PromoteClientModal({ onClose, onConfirm, displayName }) {
+  return (
+    <Modal title="Promote User" beforeTitle={<CloseButton onClick={onClose} />} contentClassName={styles.modalContent}>
+      <p>
+        <FormattedMessage id="promote.message" />
+      </p>
+      <Button preset="accept" onClick={onConfirm}>
+        <FormattedMessage id="promote.confirm-prefix" />
```
johnshaughnessy, line 16: Something odd is happening for me with the whitespace here. The `displayName` `span` is pushed right up against, `"Yes, Promote"`,
image
but then if I add a `&nbsp;`, the whitespace between `"Promote"` and the username is too big:
image
```jsx
        <FormattedMessage id="promote.confirm-prefix" />&nbsp;
```

src/react-components/client-info-dialog.js
```diff
@@ -55,15 +56,18 @@ export default class ClientInfoDialog extends Component {
...
     performConditionalSignIn(
       () => hubChannel.can("update_roles"),
       async () => {
-        this.props.showNonHistoriedDialog(PromoteClientDialog, {
+        this.props.showNonHistoriedDialog(PromoteClientModal, {
           displayName: profile.displayName,
-          onConfirm: () => hubChannel.addOwner(user.id)
+          onConfirm: () => {
+            hubChannel.addOwner(user.id);
+            onCloseDialog();
```
johnshaughnessy, line 69: The PeopleSidebar may need a re-render if it's open. (Maybe not a direct call to re-render here, but instead could be downstream of updates from hub-channel.)

johnshaughnessy (approved):

src/react-components/client-info-dialog.js
robertlong: I forgot that setting a selected person would mean that it would store a stale object when that person is updated. Now I just store the id of the person.

PR #3280, opened 2020-11-05
Room UI Redesign: Reaction Menu

robertlong:
This PR adds the newest iteration of the reaction menu. Rather than toggling freeze mode, it shows a popover with the reaction thumbnails. When you click one, it spawns in front of you where you can grab it or let it despawn.

This was a bit of a refactor of the emoji system in general. I moved a lot of logic out of the emoji-hud component and into an emoji-system and the emoji component itself. The emoji-hud is now only shown in VR mode. The emoji-system is responsible for showing/hiding the emoji-hud entity. The system is also responsible for the emoji hotkeys since the emoji-hud component is no longer always shown. Another difference is that the emoji component contains the references to the emoji assets and a common `spawnEmojiInFrontOfUser` function which allows us to depend on these exports in the reaction menu, emoji-system, and emoji-hud. 

The biggest change is that the emoji particles are no longer activated by velocity, but by grab state. I made this decision because it simplified the process of displaying particles when spawning emojis via hotkeys or the reaction menu. I think we should also be shifting the focus of emoji behavior away from a flinging, or shaking behavior. I think emojis/reactions should largely take a passive role. You should spawn them (perhaps more than once), and then they shortly fade away. This minimizes abuse, accidental negative consequences of flinging an emoji, and provides consistent interactions across platforms.


src/components/emoji-hud.js
```diff
@@ -180,24 +125,6 @@ AFRAME.registerComponent("emoji-hud", {
     this.el.sceneEl.removeEventListener("stateremoved", this._onThaw);
   },
 
-  tick() {
```
johnshaughnessy, line 183: >  I think we should also be shifting the focus of emoji behavior away from a flinging, or shaking behavior. I think emojis/reactions should largely take a passive role. You should spawn them (perhaps more than once), and then they shortly fade away. This minimizes abuse, accidental negative consequences of flinging an emoji, and provides consistent interactions across platforms.

I don't feel the same way. I think having the emojis be these juicy, silly objects we toss around has been a delightful addition and removing them will be our loss. I could definitely get behind something that addressed specific concerns about negative consequences (e.g. turn emojis transparent as they fly close to you so that it doesn't seem like they're hitting you in the head). Flinging them often indicates directional sentiment, though I'd like to see this expressed another way because I agree that it feels rude to fling emojis at someone (especially when they're unfamiliar with Hubs) in the current implementation.

src/systems/emoji-system.js
```diff
@@ -0,0 +1,55 @@
...
+  paths.actions.spawnEmoji0,
+  paths.actions.spawnEmoji1,
+  paths.actions.spawnEmoji2,
+  paths.actions.spawnEmoji3,
+  paths.actions.spawnEmoji4,
+  paths.actions.spawnEmoji5,
+  paths.actions.spawnEmoji6
+];
+
+const EMOJI_RATE_LIMIT = 1000;
```
johnshaughnessy, line 14: I'd very much appreciate it if we removed or reduced this limit. Rapid emoting is definitely something I want to be able to do. Pre-empting a couple concerns:
- If performance is an issue here, it seems easy to address with our batching system.
- If abuse is an issue here, this seems like the wrong cut-point for addressing it.

src/systems/emoji-system.js
```diff
@@ -0,0 +1,55 @@
...
+
+    for (let i = 0; i < actions.length; i++) {
+      if (
+        userinput.get(actions[i]) &&
+        i < emojis.length &&
+        window.APP.hubChannel.can("spawn_emoji") &&
+        time - this.lastSpawnedEmoji > EMOJI_RATE_LIMIT
+      ) {
+        this.lastSpawnedEmoji = time;
+        spawnEmojiInFrontOfUser(emojis[i]);
```
johnshaughnessy, line 51: Would love to see a slight force added to the emoji upon being spawned so that it has a bit of movement / life to it.

johnshaughnessy (approved):
Love the new menu. Thanks for doing some refactoring, too!

As mentioned in an earlier PR, I'd love to see us expand the emoji menu like other chat applications and also improve ui / ux concerns about spawning-to-world vs spawning-to-chat

robertlong:
I've created an issue containing those design decisions here: https://github.com/mozilla/hubs/issues/3329

PR #3304, opened 2020-11-10
Room UI Redesign: Refactor Room Sidebar

robertlong:
This PR refactors the room sidebar. The primary goals of the refactor were to reduce menu items in the more menu and make the sidebars able to be tested in storybook.

The room settings sidebar and the change scene button can now be accessed by clicking the edit button in the top right of the room sidebar. This removed two items from the more menu. I think we could likely move "Close Room" into this sidebar as well.

In the process of doing this, I cleaned up all of the form handling logic for the room settings sidebar. I pulled in the [react-hook-form](https://react-hook-form.com) library to handle form validation and form state. The library primarily focuses on uncontrolled inputs. Our NumericInput and ToggleInputs were controlled only components (using onChange and value/checked), but were pretty simple to refactor to uncontrolled inputs. The NumericInput's custom styled increment/decrement buttons had to go as a result and I moved back to the native ones. I don't think this will be a huge issue, we could even hide them if we wanted. It's better for accessibility too. There were a lot of small changes needed to make this library work well, but I think it will be a really useful tool going forward as we design forms.

Also I fixed the server message when changing scenes in this PR and I'll cross it off in the room-ui-redesign PR description when it's merged.

Overall, I'm still not totally satisfied with the design. I don't know that the invite url being in this menu makes the most sense. At least not how it is. I also think the change scene UI and UX is still a little clunky.

I think it's better than it was before and it's time for me to move on to something else though.


src/react-components/room/RoomSettingsSidebar.js
```diff
@@ -0,0 +1,228 @@
...
+    },
+    [spawnAndMoveMedia, setValue]
+  );
+
+  return (
+    <Sidebar
+      title="Room Settings"
+      beforeTitle={showBackButton ? <BackButton onClick={onClose} /> : <CloseButton onClick={onClose} />}
+    >
+      <form onSubmit={handleSubmit(onSubmit)} className={styles.roomSettingsForm}>
```
johnshaughnessy, line 71: Have any thoughts on us saving changes immediately instead of requiring a submit button?

johnshaughnessy (approved):
Love the combined modal!

> I think we could likely move "Close Room" into this sidebar as well
That sounds like a good idea to me. "Favorite Room" might belong there as well.

At some point we may want to revisit this and expose more details in the room info panel so that:
- moderators know that the edit button will show more options than room name and scene and
- other users know what permissions are granted in the room at a given time

src/react-components/room/RoomSettingsSidebar.js
robertlong: I think this method is slightly better. Being able to bulk apply changes is probably better and mimics the API a bit better.

Issue #2643, opened 2020-07-16
Investigate Issues With Users Audio Not Working


robertlong:
Umbrella issue for looking into recent reports of users not being able to hear others. This is intended to capture only bug reports made after we switched to the MediaSoup SFU.

emclaren:
the hubs email address received this message related to audio issues on July 29 


"Hubs is my absolute favorite video chat and conferencing tool. However, a couple of friends and I tested out his room where we had some audio issues. We are all very tech savvy, so none of believe it was a permissions or settings issue on our devices. I used the latest version of Firefox on a MacBook Pro while my friend used Chrome on his Windows PC. He could hear me, but I could not hear him nor could our other friend hear him. The other friend was using his iPhone and could hear me, but I could not hear him either.

Unless there is an obscure setting or permission we need to change that we haven’t thought of being related (e.g. accessibility features), I believe this is a programming bug."

zach-vz:
@emclaren @robertlong   We are also experiencing this issue in our Hubs Cloud deployment. It only happens in some instances, so does not appear to be driven by a setting or permission. We have found no way to fix yet except refreshing on the user's end. If there are additional logs we can provide, of course, we'll be in touch.

JacobErvin:
Also encountered this. Was in a group, and one member (Person A) was unable to see or hear another member (person B). All other attendees could see and hear person B correctly. Person A re-loaded and it resolved the issue, but it caused significant confusion.

vjFred:
Have this same issue, and also people sometimes don´t see each other, this on Hubs Cloud.

zach-vz:
Additional context (@vjFred @JacobErvin -- would love for you to comment on whether this is your experience as well). One pattern we're seeing is that avatars are not loading up for users, and if they don't load up, that often seems to exclude them from AV in the space (not to mention they are invisible). I think these things could be very closely linked. Most of my tests have been in Hubs Cloud, but we've experienced this in Hubs as well, with both the default set of avatars as well as custom avatars we have imported.

emclaren:
Hey folks - we have created a form to help us see if we can spot a pattern of when these issues are occuring

If you are experiencing issues with WebRTC (e.g. voice audio, webcam, and desktop streaming) could you complete [this form](https://docs.google.com/forms/d/e/1FAIpQLSd_dw8uQN2VjxLcxXmXFzzaJVU8IWXuUAnRJd1kjoB_TohqsA/viewform)

[...]

PR #3305, opened 2020-11-10
RTC Debugging panel

keianhzo:
This PR adds a support for RTC debugging tools. It exposes internal RTC information related to the signaling protocol, WebRTC (Mediasoup and raw WebRTC) and logs for most of the signaling and WebRTC events.

It's still a WIP as I still would like to optimize some of the stats generation code as there is too much processing being done right now per update.

First step towards https://github.com/mozilla/hubs/issues/2643

Although is not necessary this should ideally be tested against this Dialog PR https://github.com/mozilla/dialog/pull/9


src/naf-dialog-adapter.js
```diff
@@ -99,6 +102,62 @@ export default class DialogAdapter {
     this._onOccupantMessage = messageListener;
   }
 
+  async getServerStats() {
+    const result = {};
+    try {
+      if (this._sendTransport && !this._sendTransport._closed && this._sendTransport._connectionState == "connected") {
```
robin-k-wilson, line 108: Since these if statements are relatively complicated with multiple checks, I'd like to see a short comment of what each one is doing.

src/naf-dialog-adapter.js
```diff
@@ -99,6 +102,62 @@ export default class DialogAdapter {
...
+        });
+        result[this._sendTransport.id]["producers"] = {};
+        for (const producer of this._sendTransport._producers) {
+          const id = producer[0];
+          result[this._sendTransport.id]["producers"][id] = await this._protoo.request("getProducerStats", {
+            producerId: id
+          });
+        }
+      }
+      if (this._recvTransport && !this._recvTransport._closed && this._recvTransport._connectionState == "connected") {
```
robin-k-wilson, line 122: Comment

src/naf-dialog-adapter.js
```diff
@@ -99,6 +102,62 @@ export default class DialogAdapter {
...
+        level: "error",
+        tag: "Adapter",
+        msg: `Error getting the server status: ${e}`
+      });
+      return { error: `Error getting the server status: ${e}` };
+    }
+  }
+
+  async restartSendICE() {
+    if (this._sendTransport !== null && !this._sendTransport._closed) {
```
robin-k-wilson, line 148: comment

src/react-components/rtc-debug-panel.js
```diff
@@ -0,0 +1,1011 @@
...
+    return (
+      <p className={classNames(styles.rtcTileText)}>
+        {`${this.props.propKey}: `}
+        <span className={classNames(styles.rtcValueText)}>{`${this.props.propValue}`}</span>
+      </p>
+    );
+  }
+}
+
+export class Button extends Component {
```
robin-k-wilson, line 43: I don't think you need to define a Button component in this file since the Button component in react-components/input/Button.js has what you need or can be modified to what you need.
If you don't think so, then I'd make a new button component in that folder.


```javascript
// import Button from ect. ect.

// the className prop should override OR you update the Button component if props.className exists, use props.className
<Button disabled={true/false} onClick={func} className={classNames(styles.rtcPanelButton)} >
       <FormattedMessage id={id} />
</Button>
```

src/react-components/rtc-debug-panel.js
```diff
@@ -0,0 +1,1011 @@
+import React, { Component } from "react";
```
robin-k-wilson, line 1: I can see a few places where we can refactor the react into smaller components. Happy to work with you on this!

robin-k-wilson (changes requested):

src/naf-dialog-adapter.js
```diff
@@ -596,12 +782,13 @@ export default class DialogAdapter {
     // Close protoo Peer, though may already be closed if this is happening due to websocket breakdown
     if (this._protoo && this._protoo.connected) {
       this._protoo.close();
+      this.scene.emit("rtc_event", { level: "info", tag: "Signaling", msg: "[close]" });
     }
 
     // Close mediasoup Transports.
-    if (this._sendTransport) this._sendTransport.close();
+    if (this._sendTransport && !this._sendTransport._closed) this._sendTransport.close();
```
netpro2k, line 789: Was this causing some sort of error?

package.json
```diff
@@ -159,7 +159,7 @@
     "htmlhint": "^0.11.0",
     "jsdom": "^15.1.1",
     "localstorage-memory": "^1.0.3",
-    "mediasoup-client": "github:versatica/mediasoup-client#v3",
+    "mediasoup-client": "^3.6.21",
```
netpro2k, line 162: Nice call updating this, we should definitely try to stay on top of bugfixes upstream and keep it up to date (though its hard to do so with lots of confidence we arent breaking things). This debug panel should help, as should QA process.

src/assets/locales/es.json
```diff
@@ -284,6 +284,7 @@
   "commands.debug": "Alternar renderizado de depuración física.",
   "commands.vrstats": "Alternar estadísticas en realidad virtual",
   "commands.audiomode": "(experimental) Alternar audio posicional.",
+  "commands.audioNormalization": "(experimental) Change audio normalization factor.",
```
netpro2k, line 287: This looks like it may have snuck in here.

src/react-components/rtc-debug-panel.js
```diff
@@ -0,0 +1,1011 @@
...
+  }
+
+  render() {
+    const rootClassName = this.props.border ? classNames(styles.rtpTile) : classNames(styles.rtcStatusPanel);
+    const flow = { flexFlow: (this.props.row ? "row " : "column ") + (this.props.wrap ? "wrap" : "nowrap") };
+    return (
+      <div className={rootClassName}>
+        <div style={{ display: "flex", flexFlow: "row nowrap", justifyContent: "center" }}>
+          {this.props.title && (
+            <button className={classNames(styles.collapsibleButton)} onClick={this.collapse.bind(this)}>
```
netpro2k, line 599: You can avoid having to bind these in every render by either binding them in the constructor, or defining them using the public class field syntax: `collapse = () => {`

src/react-components/rtc-debug-panel.js
```diff
@@ -0,0 +1,1011 @@
...
+  OUTBOUND_RTP: "outbound-rtp",
+  LOCAL_CANDIDATE: "local-candidate",
+  REMOTE_CANDIDATE: "remote-candidate",
+  CANDIDATE_PAIR: "candidate-pair"
+};
+
+const STATS_REFRESH_TIME = 500;
+const PRODUCERS_KEY = "producers";
+const CONSUMERS_KEY = "consumers";
+const MEDIASOUP_DOC_BASE_URL = "https://mediasoup.org/documentation/v3/libmediasoupclient/api/";
```
netpro2k, line 24: Nice touch with the built in documentation links

netpro2k (commented):
This looks great! This should hopefully help us start to narrow down issues and target them more specifically. Since this is just debug UI I only skimmed the CSS and particulars of the React code, but it looks good anyway. We might consider dropping the log panel and just using the web inspector log, but I could see how having a webrtc specific log could be useful for debugging, so I don't have a strong opinion.

src/assets/locales/es.json
brianpeiris: This was probably inserted automatically from the locale script we have. Probably just missed doing that in the audioNormalization PR. I think it's okay to include it in this one.

src/naf-dialog-adapter.js
brianpeiris: Not to be pedantic, but I'm curious about what type of comment you would expect here. The code seems pretty self-documenting to me. I just read this as "if we have a _sendTransport, and the transport is not closed, and the transport's connection state is connected". 
Perhaps instead of adding a comment, you'd be satisfied with something like this, where the variable name serves as a summary of the condition?
```js
const sendTransportAvailable = (
  this._sendTransport && 
  !this._sendTransport._closed && 
  this._sendTransport._connectionState == "connected"
);
if (sendTransportAvailable) {
...
```
Generally I think we should make the code as clear as possible (which I think is already the case here, IMO) and reserve comments for "why" explanations where necessary, as opposed to "what" explanations. It would feel redundant to me to just have a comment say what the code is doing when you can make the code readable in the first place, but perhaps it's a matter of preference.

src/assets/locales/jp.json
```diff
@@ -322,6 +323,7 @@
   "preferences.allowMultipleHubsInstances": "複数のHubsインスタンスが開いているときに自動終了を無効にする",
   "preferences.disableIdleDetection": "アイドル時またはバックグラウンド時の自動終了を無効にする",
   "preferences.preferMobileObjectInfoPanel": "モバイルオブジェクト情報パネルを優先する",
+  "preferences.showRtcDebugPanel": "Show RTC Panel",
```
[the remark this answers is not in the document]
keianhzo: This is generated automatically when you run the `npm sync_locales.js` in the `src/assets/locales` folder. I tend to do that after I update the English local as that's what the README in that folder says.

src/naf-dialog-adapter.js
```diff
@@ -109,17 +168,29 @@ export default class DialogAdapter {
 
     await new Promise(res => {
       this._protoo.on("open", async () => {
+        this.scene.emit("rtc_event", { level: "info", tag: "Signaling", msg: "Open" });
```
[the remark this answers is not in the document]
keianhzo: This is there just to forward events to the Log panel. Do you mean that I should also log to console any time I forward a log entry?

src/naf-dialog-adapter.js
```diff
@@ -99,6 +102,62 @@ export default class DialogAdapter {
...
 
+  async getServerStats() {
+    const result = {};
+    try {
+      if (this._sendTransport && !this._sendTransport._closed && this._sendTransport._connectionState == "connected") {
+        result[this._sendTransport.id] = {};
+        result[this._sendTransport.id]["name"] = "Send";
+        result[this._sendTransport.id]["stats"] = await this._protoo.request("getTransportStats", {
+          transportId: this._sendTransport.id
+        });
```
brianpeiris, line 113: Just a suggestion for cleaner syntax:
```js
const sendTransport = result[this._sendTransport.id] = {};
sendTransport.name = "Send";
sendTransport.stats = await this._protoo.request("getTransportStats", {
  transportId: this._sendTransport.id
});
```

src/naf-dialog-adapter.js
keianhzo: I think I got some exception at some point but I can't reproduce it anymore so I'm removing it.

package.json
keianhzo: Yep, the version we were using was pretty old and this seems like a good moment to bump it.

src/assets/locales/es.json
keianhzo: This is also generated by `npm sync_locales.js` same as: https://github.com/mozilla/hubs/pull/3305#discussion_r521385117

src/react-components/rtc-debug-panel.js
keianhzo: Fixed, thanks for that.

src/react-components/rtc-debug-panel.js
keianhzo: At least will save some links for most users.

src/naf-dialog-adapter.js
keianhzo: Actually that can really be simplified. I'll update.

src/naf-dialog-adapter.js
keianhzo: Same as above

src/naf-dialog-adapter.js
```diff
@@ -173,6 +249,11 @@ export default class DialogAdapter {
     });
 
     this._protoo.on("notification", notification => {
+      this.scene.emit("rtc_event", {
+        level: "debug",
+        tag: "Signaling",
+        msg: `Notification [${notification.method}]: ${JSON.stringify(notification.data)}`
+      });
```
brianpeiris, line 256: I don't have a good sense for how frequently all of these events will be emitted, but if they happen constantly, I'd be worried about the performance impact and garbage collection due to all of these objects, strings and the events themselves. 

Ideally there would be no additional impact on performance or memory usage when the debug panel is closed. 

Perhaps these emits should be encapsulated in a helper function that can be turned off with a global flag?
Something like:
```js
function emitRTCEvent(level, tag, msgFunc) {
  if (!window.APP.store.state.preferences.showRtcDebugPanel) return;
  this.scene.emit("rtc_event", { level, tag, msg: msgFunc() });
} 

...

emitRTCEvent(
  "debug", 
  "Signaling",
  () => `Notification [${notification.method}]: ${JSON.stringify(notification.data)}`
);
```

src/naf-dialog-adapter.js
keianhzo: This can also be simplified. Updated and comment added.

src/react-components/rtc-debug-panel.js
keianhzo: Refactored after our talk.

src/naf-dialog-adapter.js
keianhzo: Thanks for that, updated.

src/naf-dialog-adapter.js
keianhzo: That sounds about right, I'll update with that.

package.json
```diff
@@ -187,7 +187,8 @@
     "webpack-bundle-analyzer": "^3.3.2",
     "webpack-cli": "^3.2.3",
     "webpack-dev-server": "^3.1.14",
-    "worker-loader": "^2.0.0"
+    "worker-loader": "^2.0.0",
+    "recharts": "^1.8.5"
```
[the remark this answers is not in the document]
keianhzo: Yep, I'll look for a leaner alternative.

keianhzo:
@netpro2k @brianpeiris @robin-k-wilson I've updated with a refactor and the comments above addressed.

package.json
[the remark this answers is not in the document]
keianhzo: I've removed and just implemented a basic graph using canvas.

brianpeiris (changes requested):
Just one major comment. Looks good to me otherwise!

src/naf-dialog-adapter.js
```diff
@@ -99,6 +102,67 @@ export default class DialogAdapter {
     this._onOccupantMessage = messageListener;
   }
 
+  /**
+   * Gets transport/consumer/producer stats on the server side.
+   */
+  async getServerStats() {
+    const result = {};
+    try {
+      if (!this._sendTransport?._closed) {
```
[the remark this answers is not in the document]
keianhzo: Updated

keianhzo:
@brian updated with the babel plugin update, some CSS fixes for Safari and the CI case related fix.

src/react-components/debug-panel/RtcDebugPanel.js
```diff
@@ -0,0 +1,699 @@
...
+  }
+
+  componentWillUnmount() {
+    this.props.scene.removeEventListener("rtc_event", this.logEvent);
+    this.props.store.removeEventListener("statechanged", this.storeUpdated);
+    this.runStats(false);
+  }
+
+  logEvent = e => {
+    const currentLog = this.state?.log?.slice(-200) || [];
```
robinkwilson, line 172: With `this.state` values specified, we know what type the data is (so we can handle an empty case) and that `this.state.log` is defined.

We also *expect* the data to have empty cases on the initial load because the data hasn't been fetched yet.

src/react-components/debug-panel/RtcDebugPanel.js
```diff
@@ -0,0 +1,699 @@
...
+    intl: PropTypes.object,
+    presences: PropTypes.object,
+    history: PropTypes.object,
+    sessionId: PropTypes.string
+  };
+
+  constructor() {
+    super();
+
+    this.state = {};
```
robinkwilson, line 156: In react, you want to specify default values for this.state so it has defaults to update. 

This line would say:
```javascript
this.state = {
    showPanel: false, // default option
    log: [],
    // other values
}
```

src/react-components/debug-panel/RtcDebugPanel.js
```diff
@@ -0,0 +1,699 @@
...
+            data={candidates[StatsType.REMOTE_CANDIDATE]}
+            title={`Remote candidate`}
+          />
+        );
+    }
+    return components;
+  }
+
+  createProducers(producers, stats) {
+    const components = [];
```
robinkwilson, line 512: Could rewrite this function as:

```javascript
// Will return [] if producers is undefined or empty array
createProducers(producers = [], stats) {
   return producers.map((producer) => (
      <TrackStatsPanel
          key={producer.id}
          title={"producer"}
          data={producer}
          xAxis={"timeDelta"}
          yAxis={"bytesSent"}
          stats={stats[producer.id]}
      />)
   )
}
```

src/react-components/debug-panel/RtcDebugPanel.js
```diff
@@ -0,0 +1,699 @@
+import React, { Component } from "react";
```
robinkwilson, line 1: Anyway we can break this RtcDebugPanel into Left, Right, Middle components?

src/react-components/debug-panel/DataPanel.js
```diff
@@ -0,0 +1,51 @@
+import React from "react";
```
robinkwilson, line 1: This is beautiful! 🎉

src/react-components/debug-panel/DataGraph.js
```diff
@@ -0,0 +1,74 @@
+import React from "react";
```
robinkwilson, line 1: import React, { useEffect } from "react";

src/react-components/debug-panel/DataGraph.js
```diff
@@ -0,0 +1,74 @@
...
+import PropTypes from "prop-types";
+
+const useCanvas = (draw, color, xData, yData, options = {}) => {
+  const canvasRef = React.useRef(null);
+
+  React.useEffect(
```
robinkwilson, line 7: Then just "useEffect"

src/react-components/debug-panel/CollapsiblePanel.js
```diff
@@ -0,0 +1,54 @@
+import React from "react";
+import PropTypes from "prop-types";
+import classNames from "classnames";
+import { CreatePropsFromData } from "./Prop.js";
+import styles from "./CollapsiblePanel.scss";
+
+function collapse(evt) {
```
robinkwilson, line 7: You could turn `isCollapsed` variable managed by the CollapsiblePanel state.

src/react-components/debug-panel/RtcDebugPanel.js
```diff
@@ -0,0 +1,699 @@
...
+        onClick={onRestart}
+        disabled={!isButtonEnabled}
+        primary
+      />
+      {candidates?.length > 0 && (
+        <CollapsiblePanel key={`Candidates`} title={`Candidates`} row url={`${MDN_DOC_BASE_URL}RTCIceCandidate`} border>
+          {candidates}
+        </CollapsiblePanel>
+      )}
+      {producers?.length > 0 && (
```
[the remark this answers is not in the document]
keianhzo: Fixed!

src/react-components/debug-panel/RtcDebugPanel.js
keianhzo: I guess initializing will save the overload of doing this check every time a log event is received. Updated.

src/react-components/debug-panel/RtcDebugPanel.js
keianhzo: Cool! updated.

src/react-components/debug-panel/RtcDebugPanel.js
keianhzo: I guess we could, do you think is worth refactoring that even if there are many changes that it's reused?

src/react-components/debug-panel/DataGraph.js
keianhzo: Updated

PR #3308, opened 2020-11-10
Add preference to toggle gyroscope control

johnshaughnessy:
Adds a preference to prevent the gyroscope from controlling the camera view


cvanmeurs:
Thanks for adding this.  

Might also be nice if it were disabled by default in landscape mode, when the joysticks are enabled, because they fight with each other.

src/assets/locales/en.json
```diff
@@ -311,6 +311,7 @@
   "preferences.movementSpeedModifier": "Movement speed modifier",
   "preferences.enableOnScreenJoystickLeft": "Enable left on-screen joystick for moving around",
   "preferences.enableOnScreenJoystickRight": "Enable right on-screen joystick for looking around",
+  "preferences.enableGyro": "Enable gyroscope (when supported by browser/device",
```
takahirox, line 314: Lack of the closing bracket.

johnshaughnessy: oops! fixed.

takahirox (approved):
The change looks good to me. And I confirmed on my Samsung S8+ that this change works fine.

PR #3320, opened 2020-11-12
Room UI Redesign: Fix testing mobile devices against a Hubs Cloud server

robertlong:
This PR turns off Webpack live reload and fixes the cors proxy settings so that it works with non-localhost addresses. This allows you to test devices against a Hubs Cloud server with the CORS settings set to allow local addresses or `*` and a webpack-dev-server serving the client on your local network.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -171,7 +171,7 @@ async function fetchAppConfigAndEnvironmentVars() {
 
   const { shortlink_domain, thumbnail_server } = hubsConfigs.general;
 
-  const localIp = await internalIp.v4();
+  const localIp = (await internalIp.v4()) || "localhost";
 
   process.env.RETICULUM_SERVER = host;
   process.env.SHORTLINK_DOMAIN = shortlink_domain;
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -171,7 +171,7 @@ async function fetchAppConfigAndEnvironmentVars() {
 
   const { shortlink_domain, thumbnail_server } = hubsConfigs.general;
 
-  const localIp = (await internalIp.v4()) || "localhost";
+  const localIp = process.env.HOST_IP || (await internalIp.v4()) || "localhost";
 
   process.env.RETICULUM_SERVER = host;
   process.env.SHORTLINK_DOMAIN = shortlink_domain;
@@ -231,6 +231,8 @@ module.exports = async (env, argv) => {
 
   const host = process.env.HOST_IP || env.localDev || env.remoteDev ? "hubs.local" : "localhost";
 
+  const liveReload = !!process.env.LIVE_RELOAD || false;
+
   const legacyBabelConfig = {
     presets: ["@babel/react", ["@babel/env", { targets: { ie: 11 } }]],
     plugins: [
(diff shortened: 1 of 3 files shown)


webpack.config.js
```diff
@@ -271,7 +274,8 @@ module.exports = async (env, argv) => {
       headers: {
         "Access-Control-Allow-Origin": "*"
       },
```
johnshaughnessy, line 276: I know when that

webpack.config.js
```diff
@@ -271,7 +274,8 @@ module.exports = async (env, argv) => {
...
-      inline: !env.bundleAnalyzer,
+      hot: false,
+      inline: false,
```
johnshaughnessy, line 278: bling
hotline
That can only mean one thing
hotline2

webpack.config.js
```diff
@@ -170,11 +171,13 @@ async function fetchAppConfigAndEnvironmentVars() {
 
   const { shortlink_domain, thumbnail_server } = hubsConfigs.general;
 
+  const localIp = await internalIp.v4();
+
   process.env.RETICULUM_SERVER = host;
   process.env.SHORTLINK_DOMAIN = shortlink_domain;
-  process.env.CORS_PROXY_SERVER = "localhost:8080/cors-proxy";
+  process.env.CORS_PROXY_SERVER = `${localIp}:8080/cors-proxy`;
```
johnshaughnessy, line 178: May want to include or fallback to `localhost` in case `internalIp.v4()` returns `undefined` (e.g. if `ip` is not installed https://github.com/sindresorhus/internal-ip#usage)

johnshaughnessy (commented):
This looks good to me but I don't know how to determine whether this impacts security. Would feel more comfortable if Dom or Brian looked it over instead of me

webpack.config.js
```diff
@@ -271,7 +274,8 @@ module.exports = async (env, argv) => {
       headers: {
         "Access-Control-Allow-Origin": "*"
       },
-      inline: !env.bundleAnalyzer,
+      hot: false,
```
netpro2k, line 277: I generally do like having hot reloading most of the time. Maybe expose as an env var? I also dont really mind just editing my webpack config locally though, so not a very strong opinion.

webpack.config.js
```diff
@@ -170,11 +171,13 @@ async function fetchAppConfigAndEnvironmentVars() {
 
   const { shortlink_domain, thumbnail_server } = hubsConfigs.general;
 
+  const localIp = (await internalIp.v4()) || "localhost";
```
netpro2k, line 174: The concept of a "local" ip is a bit muddy, you might have multiple "local" interfaces. The lib seems to resolve it based on the default gateway, which is slightly weird conceptually but will probably be the right thing in most cases. All that to say, maybe we want to expose an env var for this to be configured, though I doubt most people would need it.

netpro2k (commented):
lgtm, made some suggestions about more options to maybe expose, but they are not very important.

webpack.config.js
robertlong: Cool, added it.

webpack.config.js
robertlong: Yup, reused an existing one.

PR #3337, opened 2020-11-13
Room UI Redesign: Reduce Custom CSS

robertlong:
This PR removes a bunch of unnecessary CSS styles by reusing common components and semantic html elements.


robertlong:
I talked with @johnshaughnessy about this. I'm going to merge since there isn't much to review here.

PR #3351, opened 2020-11-16
Use primary language subtag locale strings

johnshaughnessy:
If we do not have a translation entry for a user's full language tag, we should check the first subtag which indicates the primary language.

A language tag carries more information than we need:
```
[primary language subtag]
-[up to three extended language subtags]
-[optional script subtag]
-[optional region subtag]
-[optional variant subtag]
-[optional extension subtag]
-[optional private-use subtag]
```

See https://en.wikipedia.org/wiki/IETF_language_tag#Syntax_of_language_tags


robin-k-wilson (approved):

vincentfretin:
You should check first the full code, then the subtag.
Your code won't work if somebody provides a new translation for zh-HK or zh-TW that is different from zh-CN.

johnshaughnessy:
> You should check first the full code, then the subtag.
> Your code won't work if somebody provides a new translation for zh-HK or zh-TW that is different from zh-CN.

Good call @vincentfretin 
Updated in https://github.com/mozilla/hubs/pull/3351/commits/da6af109e29cf13057582cd4436c5a7b0ee9cb57

PR #3363, opened 2020-11-17
Room UI Redesign: Migrate Modals

robertlong:
This PR affects the following modals:

- Migrated AvatarUrlModal
- Migrated SceneUrlModal
- Migrated ObjectUrlModal
- ClientInfoDialog moved to UserProfileSidebarContainer
- Migrated CloseRoomModal
- Migrated WebVRUnsupportedModal
- Removed WebRTCScreenShareUnsupported dialog
- Migrated Tweet/TwitterOAuth Modals
  - Twitter OAuth now done in a popup instead of using a redirect. This allows us to preserve the UI state and avoid having to reload the whole room.
- Migrated OAuthDialog to OAuthScreen

To Do:
- [ ] Twitter OAuth Modal needs to refresh the perms token so that subsequent tweets don't require connecting to Twitter again. This needs a backend change: https://github.com/mozilla/reticulum/issues/432 but we don't need to block on this.
- [x] Thoroughly test the OAuthScreen


package.json
```diff
@@ -118,7 +118,7 @@
     "react-emoji-render": "^0.4.6",
     "react-hook-form": "^6.10.1",
     "react-infinite-scroller": "^1.2.2",
-    "react-intl": "^5.4.5",
+    "react-intl": "^5.10.1",
```
robertlong, line 121: Note: this was probably reverted in a merge from the master branch. So I'm updating it again in this PR. It's needed for the `useIntl` hook.

src/hub.js
```diff
@@ -1229,6 +1197,8 @@ document.addEventListener("DOMContentLoaded", async () => {
     }
   });
 
+  console.log(oauthFlowPermsToken);
```
johnshaughnessy, line 1200: console log

src/react-components/room/ObjectUrlModal.js
```diff
@@ -0,0 +1,112 @@
...
+              <IconButton as="label" className={classNames({ [styles.hidden]: showCloseButton })} htmlFor="file">
+                <AttachIcon />
+                <input id="file" className={styles.hidden} name="file" type="file" ref={register} />
+              </IconButton>
+            </>
+          }
+          description="Accepts glb, png, jpg, gif, mp4, and mp3 files"
+        />
+        <Button type="submit" preset="accept">
+          <FormattedMessage id="avatar-url-dialog.apply" />
```
johnshaughnessy, line 99: "Apply" feels awkward here. Maybe "add" or "create" or "load"?
image

johnshaughnessy (approved):
I was not able to test the OAuth flows but otherwise this seems good to me!

Issue #3314, opened 2020-11-11
Ability to change from "spectate" to "join"


heaversm:
**Is your feature request related to a problem? Please describe.**

In the new UI, if I choose "Spectate", it doesn't appear that I can change to "Join Room" afterward unless I refresh my browser.

**Describe the solution you'd like**

Ability to change from the spectator of a room to a participant in the room.

**Additional context**

![spectate](https://user-images.githubusercontent.com/520541/98844856-c98d9b00-2401-11eb-96c1-b15a5f52bfa4.gif)

@mbransn

heaversm:
Closing to surface within DPX first.

misslivirose:
We should add visual UI for this - you can return to the main lobby panel by hitting the 'Escape' key when spectating

mbransn:
@misslivirose I don't have a way to add labels, can you add the dpx label for zenhub integration and we'll work to resource this UI?

fernzharp:
This has been solved by the Hubs team
![Screenshot (446) 1.png](https://images.zenhubusercontent.com/5f691a47ee469c25e860480a/7fe3cfbf-ed60-45fe-bec2-a4b4fa382556)

PR #3366, opened 2020-11-18
Room UI Redesign: Tips / Spectator Mode / Streamer Mode

robertlong:
This PR migrates the tips over to a new design and significantly reduces the number of onboarding tips in Hubs> All of the object spawning / manipulation tips were removed due to relocation of the spawn menu and the quality of the UX for those actions. Ideally we will revisit onboarding in detail in the future after design research.

Spectator mode now has dedicated enter room and enter vr buttons.

Streamer mode now has a red outline around the viewport when streaming. The exit streamer mode button has been relocated to the more menu to avoid conflict with the more menu button in the top left on mobile.

Fixes https://github.com/mozilla/hubs/issues/3369
Fixes https://github.com/mozilla/hubs/issues/3314
Fixes https://github.com/mozilla/hubs/issues/3370

Depends on https://github.com/mozilla/hubs/pull/3363


src/assets/locales/en.json
```diff
@@ -346,59 +346,21 @@
   "settings.row-profile": "You",
   "settings.row-room": "Room",
   "settings.row-tools": "Tools",
-  "tips.mobile.video_share_mode": "Tap to stop sharing video.",
```
johnshaughnessy, line 349: We'll need to remove these tips from the alternative translations too. Perhaps the easiest way is to update sync_locales.js so that it removes extra entries. (Currently it only adds missing entries). 
https://github.com/mozilla/hubs/blob/redesign/tips/src/assets/locales/sync_locales.js

src/systems/tips.js
```diff
@@ -82,263 +37,62 @@ const tipPlatform = () => {
...
-    if (
-      scene.is("frozen") &&
-      (userinput.activeSets.includes(sets.rightCursorHoveringOnInteractable) ||
-        userinput.activeSets.includes(sets.leftCursorHoveringOnInteractable))
-    )
-      return FINISH;
-    return VALID;
-  },
-  invite: function(userinput, scene, _mediaCounter, _store, hub) {
+  invite: function(_userinput, scene, hub) {
```
johnshaughnessy, line 64: I think the FINISH step for this should gated on pressing the invite button - not on becoming copresent.

src/systems/tips.js
```diff
@@ -82,263 +37,62 @@ const tipPlatform = () => {
...
+    localStorage.setItem(LOCAL_STORAGE_KEY, JSON.stringify({}));
+    localStorageCache = null;
+    window.APP.store.resetTipActivityFlags();
+    window.APP.store.resetConfirmedBroadcastedRooms();
+    this.activeTip = null;
+    finished = false;
+  },
+
+  skipTips: function() {
+    for (let i = 0; i < platformTips.length; i++) {
```
johnshaughnessy, line 90: Seems like the skip button should only skip a single tip - not dismiss all of them.

johnshaughnessy (approved):

johnshaughnessy:
Nice job -- the tips looks much better

johnshaughnessy (commented):
Woops -- I accidentally submitted the previous review before finishing

src/systems/tips.js
robertlong: I'm pretty much keeping the existing behavior here. Skipping all the tips is a pretty common action on these onboarding flows. I think we'll look into onboarding in detail with the design team soon.

src/systems/tips.js
robertlong: I'm probably not going to change this just now since it's new functionality. But I tend to agree.

PR #3423, opened 2020-11-24
move hard coded string in js to locales json.

hATrayflood:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-825)


brianpeiris:
Thanks for the pull request. Given that we are currently working on a significant redesign, I think it's best to wait until that work is done and then see if we need to pull in some of these changes if they still apply after the redesign.

robertlong:
@hATrayflood we recently shipped the redesign which significantly changed the translations. Would you mind updating this PR to use those new ids? You can find more info here https://github.com/mozilla/hubs/tree/master/src/assets/locales

Also feel free to reach out if you need any more help!

johnshaughnessy:
Closing this PR because there doesn't seem to be any movement on updating it after January's redesign. Feel free to re-open this if that changes!

Issue #2060, opened 2020-01-17
Set a room as unlisted


HeadClot:
**Is your feature request related to a problem? Please describe.**
The ability to set rooms as unlisted. [Example](https://www.youtube.com/watch?v=BGmbI0zS8Pc)

**Describe the solution you'd like**
The ability to set a room as unlisted: Not be able See or search for a room but still be able to access it. When you hit something like a trigger or perform a certain action.

**Describe alternatives you've considered**
N/A

**Additional context**
This should be an option in Spoke as well as hubs. This would be great for secret areas in the thing I am making.

gfodor:
I'm not sure what you mean exactly here -- all rooms in hubs are currently unlisted, meaning there is no way to discover the URL to a room unless you've been given that URL directly by someone, or indirectly by a link saved to a scene or another room.

HeadClot:
So I would like the ability to set the ability to explicitly set a room as unlisted. So for example I set a room as private I want the ability to set a room back to unlisted.

gfodor:
I think this feature can be described as the ability to close and re-open a room. Right now rooms can be closed but not re-opened. As part of our access control updates, we could potentially make it possible to re-open the room (perhaps only via an API)

emclaren:
I hear from community members reports of closing rooms by mistake (it has even happened to our meetup room by a team member). It would be nice if instead of having to create a brand new room, we could reopen old rooms so the URL would still work. We received an email about a mistakenly closed room this week. 

On a related note - as an event organizer in Hubs, I would love to be able to close Hubs rooms when they aren't in use and have custom text displayed on the landing page. 

Use case:  We cancelled last week's meetup but users could still enter into the room. No one was around to moderate the space during this time, so things could have been unpleasant for users if a bad-actor attended. Further, anyone new to Hubs may have come to the space, and not known the controls well enough to see the multiple "Meetup cancelled this week" on the walls. It would be great if this could be customized when you close the room (e.g. see screenshot)
<img width="1694" alt="Screen Shot 2020-06-23 at 9 36 44 AM" src="https://user-images.githubusercontent.com/6835324/85433592-80fc4080-b539-11ea-9ac2-36bc45780c81.png">

For me, this would also be helpful for displaying information about office hours times in case people dropped in when they were closed.

emclaren:
Right after I wrote my previous note, we received a message to the hubs email address from a community member asking for the same thing ("We would like to have the site only open specific hours when it can be staffed.")

PR #3450, opened 2020-12-02
Switch colors of the close room dialogue buttons so "Cancel" is highlighted

emclaren:
We occasionally get emails from people who have accidentally closed their rooms (see https://github.com/mozilla/hubs/issues/2060) - this update makes the "cancel" button brighter so people will hopefully be less likely to quickly click through and shut their room down by mistake.


emclaren:
Closing because #3510 makes this redundant

PR #3462, opened 2020-12-02
Room UI Redesign: Standalone VR Fixes

robertlong:
This PR uses the small breakpoint for the short standalone VR viewport size. It also fixes some issues with landscape orientation on mobile phones.


keianhzo:
@robertlong Sorry I'm late for the review. I've tested it today and it looks good now. The only thing I've seen is that in the avatar settings screen you still need to scroll to see the Accept button, but it's not terrible. The rest looks good.

PR #3469, opened 2020-12-03
Room UI Redesign: Accessibility Polish

robertlong:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-824)


netpro2k:
Unfortunately things have diverged enough that merging this PR as is is probably not realistic. Will keep the original issue open which links to this PR for reference if/when we get to this again.

Issue #3459, opened 2020-12-02
Invite UI should change when room creator has chosen "Invite only" mode


brianpeiris:
**Description**
If a room creator sets room access to "Invite only", the "Invite" button in the bottom left corner shows a popup with unusable info.

**To Reproduce**
1. In the Room Settings panel, set the room to "Invite only"
2. The "Invite" popup in the bottom left now contains info that is unusable

**Expected behavior**
In the previous design, we hid the invite dialog entirely when a room was put into "invite only" mode, since the room URL, Room Code and Embed Code there are not usable in that mode. We should either remove the invite button and popup, or change it to hide the Room Code and Embed sections, and have the "Room Link" include the "hub_invite_id" query param.

**Hardware** 
 - Device: Desktop
 - OS: Windows
 - Browser: Chrome

robertlong:
@brianpeiris we want users in the room to be able to invite other users when it's in invite mode right? So I'm thinking maybe we should show the invite link with the hub_invite_id param.

PR #3503, opened 2020-12-07
Room UI Redesign: Invite Link Popover and Fixes

robertlong:
Fixes: https://github.com/mozilla/hubs/issues/3459

The invite popover shows the redesign link when it is required and if you have permission to retrieve it via the Hub Channel. If it is required but you don't have permission, the invite popover button will be hidden. I made it so we can reuse the `InviteLinkInputField` and `useInviteUrl` code between the `RoomSettingsSidebar` and `InvitePopover` components.

In the process I noticed that the "Edit" button at the top of the `RoomSidebar` component was visible when you didn't have permission to edit the room. I've fixed that bug. I also noticed that the revoke invite link button closed the popover component because it changed the content of the popover before the `click` event fired. So I moved the event to use `mousedown`. This exposed a bug with the `MoreMenuSidebar` where clicking on items wouldn't close the popover so I added an explicit call to `closePopover`.

### DIFF
--- a/src/react-components/room/InviteLinkInputField.js
+++ b/src/react-components/room/InviteLinkInputField.js
@@ -1,0 +1,61 @@
+import React, { useState, useCallback } from "react";
+import PropTypes from "prop-types";
+import styles from "./RoomSettingsSidebar.scss";
+import { IconButton } from "../input/IconButton";
+import { FormattedMessage } from "react-intl";
+import { CopyableTextInputField } from "../input/CopyableTextInputField";
+
+export function InviteLinkInputField({ fetchingInvite, inviteUrl, onRevokeInvite }) {
+  const [showRevokeConfirmation, setShowRevokeConfirmation] = useState(false);
+
+  const revokeInvite = useCallback(() => {
+    setShowRevokeConfirmation(true);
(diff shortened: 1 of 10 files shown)


src/react-components/room/InviteLinkInputField.js
```diff
@@ -0,0 +1,61 @@
...
+      setShowRevokeConfirmation(false);
+    },
+    [onRevokeInvite]
+  );
+
+  return (
+    <CopyableTextInputField
+      label="Invite link"
+      disabled={fetchingInvite}
+      value={fetchingInvite ? "Generating invite..." : inviteUrl}
```
brianpeiris, line 31: Reminder to add i18n if you haven't gotten to that yet.

brianpeiris (approved):
LGTM

src/react-components/room/InviteLinkInputField.js
robertlong: Yeah, currently putting that on pause so I can unblock some of the QA testing, but I'll come back to it soon.

robertlong:
Thanks for the review!

Issue #2625, opened 2020-07-16
Improve room deletion flow


misslivirose:
**Is your feature request related to a problem? Please describe.**
We have gotten reports that our current room deletion flow makes it too easy to accidentally close rooms, which cannot be undone. (Related: #2060) Given the destructive nature of this action, we should make it harder to accidentally close a room. 

**Describe the solution you'd like**
GitHub requires you to type the name of a repository when you take a delete action. We should have an additional confirmation that the user needs to take before the room is closed (either typing the room name or DELETE or something similar). Additionally, we might want to incorporate #2303 as part of this flow.

PR #3510, opened 2020-12-08
Type room name to confirm close room

robinkwilson:
- Updated fsevents in package.json + package-lock.json because the old version is not hosted where npm expects -- likely deleted
- Textbox is difficult to see but matches the global styles
- If you type in a different input rather than the room name, a reddish border surrounds the textbox. 

closes https://github.com/mozilla/hubs/issues/2625

image

image

image

image


robinkwilson:
The text box is really hard to see but matches the global styles.

src/assets/locales/en.json
```diff
@@ -72,6 +72,7 @@
   "close-room.message": "Closing this room will remove yourself and others from the room, shutting it down permanently.\n\nAre you sure? This action cannot be undone.",
   "close-room.confirm": "Yes, Close Room",
   "close-room.cancel": "Cancel",
+  "close-room.type-to-confirm": "Type room name to confirm:",
```
brianpeiris, line 75: You'll need to run the `sync-locales.js` script to update the other locales with this new string.

src/react-components/close-room-dialog.js
```diff
@@ -1,27 +1,54 @@
...
+        <input
+          className={classNames(
+            styles.formFieldText,
+            !isMatch && this.state.showIsNotMatchError ? styles.errorBorder : ""
+          )}
+          value={this.state.confirmText}
+          onChange={e => this.setState({ confirmText: e.target.value })}
+          required
+          spellCheck="false"
+          ref={inp => (this.nameInput = inp)}
```
brianpeiris, line 41: This `nameInput` ref is unused

brianpeiris (commented):
Couple of comments. LGTM otherwise.

brianpeiris:
Do we want to merge Elgin's change into this as well? #3450

Issue #3524, opened 2020-12-09
Viewport is resized when sidebar is open and pause menu is opened


robertlong:
**Description**
This bug was reported by @robinkwilson 

**To Reproduce**
1. Open the chat sidebar
2. Open pause menu
3. The canvas viewport shrinks

**Expected behavior**
The canvas should always be resized to the size of the viewport

brianpeiris:
Probably the same bug, but I also see this happen when the sidebar is closed.

brianpeiris:
To add more detail, I see this happen in Firefox on Windows. If I resize my browser window, the rendered viewport squeezes horizontally, and then when I press the spacebar, it stretches back horizontally.

Softvision-GeluHaiduc:
I verified this issue using Firefox and Chrome on Windows 10 and Ubuntu 20.04.1
The main issue reported is fixed but unfortunately, the scenario depicted by Brian in the above comment is still reproducible on Chrome on both Windows and Ubuntu, the issue is not reproducible on Firefox.
[Here](https://drive.google.com/file/d/1sdJ3NYx19wdzEP6Qg_ocbiyiDNBoJJtE/view?usp=sharing) is a video of the issue.

robertlong:
@Softvision-GeluHaiduc I think this should be fixed in all browsers now

Softvision-GeluHaiduc:
I have verified this issue using Firefox and Chrome on Windows 10 and Ubuntu 20.04.1
Now, the viewport is not resized when the sidebar is open and the pause menu is opened, the scenario from Brian's comment is also fixed.

PR #3533, opened 2020-12-10
Room UI Redesign: Add useResizeViewport hook

robertlong:
This new hook manages the canvas size and viewport resolution. It overrides the internal AFrame resize handler and replaces our set-max-resolution component. By doing so we can ensure that the maximum resolution is always respected and resizing the viewport via the sidebar opening/closing, etc. is handled correctly.

Fixes https://github.com/mozilla/hubs/issues/3524


src/react-components/ui-root.js
```diff
@@ -1484,10 +1485,11 @@ class UIRoot extends Component {
                   scene={this.props.scene}
                 />
               )}
-            <RoomLayout
+            <RoomLayoutContainer
+              scene={this.props.scene}
+              store={this.props.store}
               objectFocused={!!this.props.selectedObject}
               streaming={streaming}
-              onResizeViewport={this.onResizeViewport}
```
brianpeiris, line 1490: You removed the reference to `this.onResizeViewport` here, but the method still exists and is now unused.

brianpeiris (approved):
Just minor comment. LGTM otherwise.

PR #3543, opened 2020-12-11
Try to recreate audio track if the mic was temporarily disconnected

vincentfretin:
This fixes the first issue in my comment https://github.com/mozilla/hubs/issues/2643#issuecomment-742750908
If the USB mic is temporarily disconnected, try recreating the audio track right away. If the mic is not found, try in 10s, retries 3 times.

Here is the screenshot where I disconnected the mic, and plugged it again, it successfully created back the audio track and the mic is still working. I disconnected again the mic, and it gave up and switched to default internal mic at the end.

Capture d’écran de 2020-12-11 17-17-35

Feel free to modify this PR and adapt it to the new UI Redisign so you can alert the user that the mic was disconnected and let the user choose to select another mic or the usb mic again if it's again available.

### DIFF
--- a/src/react-components/ui-root.js
+++ b/src/react-components/ui-root.js
@@ -649,28 +649,69 @@ class UIRoot extends Component {
 
       this.setState({ audioTrack, mediaStream });
 
-      if (/Oculus/.test(navigator.userAgent)) {
-        // HACK Oculus Browser 6 seems to randomly end the microphone audio stream. This re-creates it.
-        // Note the ended event will only fire if some external event ends the stream, not if we call stop().
-        const recreateAudioStream = async () => {
+      // Chrome ends the microphone audio stream if the USB is disconnected a few sceonds. This re-creates it.
+      // Note the ended event will only fire if some external event ends the stream, not if we call stop().
+      const recreateAudioStream = async (showWarning=true) => {
+        if (showWarning) {
           console.warn(
-            "Oculus Browser 6 bug hit: Audio stream track ended without calling stop. Recreating audio stream."
+            "Audio stream track ended without calling stop. Recreating audio stream."
           );
+        }
 
-          const newStream = await navigator.mediaDevices.getUserMedia(constraints);
-          const audioTrack = newStream.getAudioTracks()[0];
+        const { lastUsedMicDeviceId } = this.props.store.state.settings;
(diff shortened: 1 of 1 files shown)


vincentfretin:
I changed, now with only 2 retries with 5s, so total 10s before it switches to internal mic.

vincentfretin:
Please note that this "ended" behavior triggers only with Chrome. Firefox doesn't trigger the behavior.

vincentfretin:
I confirm this code is still executed on latest Oculus browser on the Oculus Quest, when I take off the headset, I get the ended event on the mic audio track. So the code is executed indefinitely until you put back the headset.

robertlong:
@keianhzo now that we're out of the redesign launch, we can probably fix the conflicts and merge this. Thanks again @vincentfretin!

keianhzo:
@vincentfretin I've addressed this in this PR: https://github.com/mozilla/hubs/pull/3965 but I have avoided the retry mainly because a `trackend` event can be caused not only by a random mic disconnect but also an explicit user permissions revoke for example. In that case we don't want to retry.
That PR disconnects the microphone in case we get a `trackend` event. I think the right think to do in this case is probably notify that the mic has been disconnected and provide a more discoverable way of switching mic inputs from the microphone button in the bottom menu instead of having to go to the preferences.

vincentfretin:
I didn't think about the revoke permission, hum.
Be able to easily switch mic is a must-have yes. Now for the use case of the ended event that can sometimes happen if you slightly touch the usb cable, this will not be a good user experience if I need to select another mic (If I'm lucky to have one, I usually switch to USB mic that is no working mic detected because I'm connected to a dock) and switch back to the previous mic that is the internal mic. This what I need to do in Google Meet by the way when somebody tell me they can't hear me... because I touched the usb-c cable to my dock at one point and the mic ended. :)
So as a user I won't understand why it says my mic was disconnected and I need to switch the mic.

keianhzo:
@vincentfretin I agree that we need better UX for notifying the user that there was an issue with the mic or it was disconnected somehow. I''ll raise this with the UX team to come with the right way of flagging this. For the moment I'm' going to close this PR  in favor of this fix in https://github.com/mozilla/hubs/pull/3965. Thanks a lot for your feedback!

PR #3561, opened 2020-12-16
Enable promotion to staging channel in jenkins build

brianpeiris:
We want to be able to promote habitat builds directly to a staging channel in order to test Hubs Cloud builds. This allows a jenkins job to be configured to skip S3 uploads (which are only used for hubs.mozilla.com) and promote to a staging channel instead.


johnshaughnessy (approved):
lgtm
Might be nice to include more details in the slack message (like my comment in the ret PR).

Would be good to DRY our Jenkinsfiles up someday, but we don't need to address this in this PR.

PR #3599, opened 2020-12-18
Update community Discord link so it opens the #welcome channel

emclaren:
The old link to Discord (https://discord.gg/wHmY4nd) defaulted to opening the dev channels. Now that these are behind an emoji-reaction Discord role, when new users enter the server they first see a screen like this
discord screen

This updated link should take the user to the #welcome channel where they can review the server rules.


src/react-components/layout/Header.js
```diff
@@ -50,7 +50,7 @@ export function Header() {
             </IfFeature>
             <IfFeature name="show_community_link">
               <li>
-                <a href={configs.link("community", "https://discord.gg/wHmY4nd")}>
+                <a href={configs.link("community", "https://discord.gg/dFJncWwHun")}>
```
brianpeiris, line 53: This might be a good opportunity to just remove this URL here, and also in settings-menu.js below, since it is already configurable (via `configs.link()`). I believe just removing the second parameter should be sufficient.

Not a big deal though, I'd be okay with pushing this PR either way.

brianpeiris (approved):
Minor comment. LGTM otherwise.

src/react-components/join-us-dialog.js
```diff
@@ -9,7 +9,7 @@ export default class JoinUsDialog extends Component {
         <span>
           <p>
             Join us in the{" "}
-            <a href="https://discord.gg/wHmY4nd" target="_blank" rel="noopener noreferrer">
+            <a href="https://discord.gg/dFJncWwHun" target="_blank" rel="noopener noreferrer">
```
brianpeiris, line 12: If you're feeling adventurous, you could even remove this entire file. Looks like we don't use it any more.

PR #3604, opened 2020-12-19
reexecute enableChromeAEC if network changed

vincentfretin:
When the network change from wired to Wi-Fi, the RTCPeerConnection loopback stays disconnected, recreate it after 10s if it didn't get back to connected.

Capture d’écran de 2020-12-19 14-40-31

Note : this is only a partial fix. There is still an issue with the websocket and Send Transport that doesn't reconnect. It just stays disconnected on Chrome.

### DIFF
--- a/src/systems/audio-system.js
+++ b/src/systems/audio-system.js
@@ -115,15 +115,12 @@ export class AudioSystem {
     this.analyserLevels = new Uint8Array(this.outboundAnalyser.fftSize);
     this.outboundGainNode.connect(this.outboundAnalyser);
     this.outboundAnalyser.connect(this.mediaStreamDestinationNode);
-    let alreadyExecuting = false;
 
     /**
      * Chrome and Safari will start Audio contexts in a "suspended" state.
      * A user interaction (touch/mouse event) is needed in order to resume the AudioContext.
      */
     const resume = () => {
-      if (alreadyExecuting) return;
-      alreadyExecuting = true;
       this.audioContext.resume();
 
       setTimeout(() => {
--- a/src/systems/audio-system.js
+++ b/src/systems/audio-system.js
@@ -1,3 +1,16 @@
+let delayedReconnectTimeout = null;
+function performDelayedReconnect(gainNode) {
+  if (delayedReconnectTimeout) {
+    clearTimeout(delayedReconnectTimeout);
+  }
+
+  delayedReconnectTimeout = setTimeout(() => {
+    delayedReconnectTimeout = null;
+    console.warn("enableChromeAEC: recreate RTCPeerConnection loopback because the local connection was disconnected for 10s");
+    enableChromeAEC(gainNode);
+  }, 10000);
+}
+
 async function enableChromeAEC(gainNode) {
   /**
    *  workaround for: https://bugs.chromium.org/p/chromium/issues/detail?id=687574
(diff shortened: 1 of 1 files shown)


src/systems/audio-system.js
```diff
@@ -76,12 +115,15 @@ export class AudioSystem {
...
+    let alreadyExecuting = false;
 
     /**
      * Chrome and Safari will start Audio contexts in a "suspended" state.
      * A user interaction (touch/mouse event) is needed in order to resume the AudioContext.
      */
     const resume = () => {
+      if (alreadyExecuting) return;
+      alreadyExecuting = true;
       this.audioContext.resume();
```
vincentfretin, line 127: In my tests, sometimes it executed this code twice, so I added a boolean to execute only once.

vincentfretin:
Even with this fix in my app, a user reported to me that on Chrome macOS he saw the other participant avatar talking (so the audio-analyzer is working) but couldn't hear anything. I don't have much details about it... I'm wondering if the Audio object created in enableChromeAEC can be paused by the browser (when taking a Skype or Teams call like it's the case for Safari iOS) or any other event...
I added the following code in my app, but I don't know if it helps or not, I don't have sentry integrated yet.

```
  const audioEl = new Audio();
  audioEl.setAttribute("autoplay", "autoplay");
  audioEl.setAttribute("playsinline", "playsinline");
  window.app.audioEl = audioEl;
  // https://developer.mozilla.org/en-US/docs/Web/Guide/Audio_and_video_delivery/Cross-browser_audio_basics
  audioEl.addEventListener("pause", (e) => {
    console.error("enableChromeAEC: audioEl pause", e);
    performDelayedReconnect(gainNode);
  });
  audioEl.addEventListener("ended", (e) => {
    console.error("enableChromeAEC: audioEl ended", e);
    performDelayedReconnect(gainNode);
  });
  audioEl.addEventListener("suspend", (e) => {
    console.error("enableChromeAEC: audioEl suspend", e);
    performDelayedReconnect(gainNode);
  });
  audioEl.addEventListener("abort", (e) => {
    console.error("enableChromeAEC: audioEl abort", e);
    performDelayedReconnect(gainNode);
  });
  audioEl.addEventListener("error", (e) => {
    console.error("enableChromeAEC: audioEl error", e);
    performDelayedReconnect(gainNode);
  });
  audioEl.addEventListener("emptied", (e) => {
    console.error("enableChromeAEC: audioEl emptied", e);
    performDelayedReconnect(gainNode);
  });
  audioEl.addEventListener("stalled", (e) => {
    console.error("enableChromeAEC: audioEl stalled", e);
    performDelayedReconnect(gainNode);
  });
```

vincentfretin:
Please forget my last comment. The issue I was having is actually the enableChromeAEC function that is sometime not called because of my alreadyExecuting boolean. :D Sometimes the setTimeout 0 to execute the callback on next tick is not enough, the audioContext is not in the running state yet.

src/systems/audio-system.js
```diff
@@ -76,12 +115,15 @@ export class AudioSystem {
...
     this.outboundAnalyser.connect(this.mediaStreamDestinationNode);
+    let alreadyExecuting = false;
 
     /**
      * Chrome and Safari will start Audio contexts in a "suspended" state.
      * A user interaction (touch/mouse event) is needed in order to resume the AudioContext.
      */
     const resume = () => {
+      if (alreadyExecuting) return;
+      alreadyExecuting = true;
```
vincentfretin, line 126: Remove this. This causes a regression that sometimes the enableChromeAEC is not executed because the audioContext is still not in the running state when the callback from the setTimeout is executed.

robertlong:
@keianhzo I think we're ready to work towards merging this now pending a review. Thanks again @vincentfretin!

src/systems/audio-system.js
```diff
@@ -1,3 +1,16 @@
+let delayedReconnectTimeout = null;
+function performDelayedReconnect(gainNode) {
+  if (delayedReconnectTimeout) {
+    clearTimeout(delayedReconnectTimeout);
+  }
+
+  delayedReconnectTimeout = setTimeout(() => {
```
keianhzo, line 7: Now we do retries to reconnect for both signaling and ICE after disconnects/failures so to keep this aligned with that behavior maybe it would make sense to do a `setInterval` here instead of a `setTimeout` here too so we keep on retrying until it reconnects? Also keeping the retry time aligned with the other retry times of 2secs might be good so all services go up at the same time.

keianhzo (changes requested):

src/systems/audio-system.js
keianhzo: Can you also make sure  that the linter is working so the code structure is aligned with the rest of the app?

keianhzo:
Looks great, thanks a lot for this @vincentfretin! Now that we are doing retries for other RTC relates services it's great to have this aligned. I've just pointed out a couple of things. Let me know what your thoughts are.

src/systems/audio-system.js
vincentfretin: It's actually a local RTCPeerConnection and from my tests it doesn't seem to be an issue if you lose the connection to the server. But it goes into disconnected state if network changed. In my tests I power off the wired network to force the switch to Wi-Fi so probably the wired network interface is gone. So the setTimeout seems enough in this particular case.
You can test this further if you want, but I'm not gonna spend any more time on it.

keianhzo (approved):

PR #3610, opened 2020-12-21
RTC panel fixes

keianhzo:
A few RTC panel updates:
- Change consumer/producer background to red when there is no traffic to outline a possible error better.
- Always show the candidates status and set bg to red in case they couldn't be gathered.
- Fix the 00 shown when there are no local/remote consumers.
-


keianhzo:
Closing in favor of https://github.com/mozilla/hubs/pull/3662 that is built on top of the new UI

Issue #2397, opened 2020-04-21
WASD-navigation on non-qwerty keyboards (e.g. Azerty)


eaganj:
**Description**
The wasd/qe navigation keys are awkward to use on non-qwerty keyboards.  For example, on a French keyboard layout, the Q and A keys are swapped as are the W and Z keys.  This makes for unintuitive navigation on such keyboards.

**To Reproduce**
1. Use an AZERTY keyboard (or re-map to a virtual azerty keyboard layout).
2. Try to navigate using the wasd/qe keys.
3. See https://en.wikipedia.org/wiki/AZERTY#/media/File:KB_France.svg for why this problematic.

**Expected behavior**
Use zqsd/qe keys instead when on an Azerty keyboard (probably through a config option).

**Screenshots**
See step 3 above.

**Hardware** 
 - Device: Desktop
 - OS: yes
 - Browser: yes

**Additional context**
Azerty is the standard keyboard layout in France.  There are probably other keyboard layouts with similar problems (although QWERTZ does not seem to be one of them).

eaganj:
As a temporary, albeit awkward and uncomfortable, workaround: it is possible to navigate with the arrow keys and rotate with the q-e keys (in the a-e position of a qwerty keyboard).

vincentfretin:
In a simple scene created with A-Frame 1.0.4 with wasd-controls component, zqsd on a French azerty keyboard works great, as well as wasd on a qwerty keyboard . I did a quick search in hubs repo, I found `wasd-controls="enabled: false"` in `src/scene.html` so it doesn't seem to use `wasd-controls` but its own component, I guess it's in `src/systems/userinput/bindings/keyboard-mouse-user.js`.

`wasd-controls` is using key codes:
```
    KEYCODE_TO_CODE: {                                                               
    '38': 'ArrowUp',                                                               
    '37': 'ArrowLeft',                                                             
    '40': 'ArrowDown',                                                             
    '39': 'ArrowRight',                                                            
    '87': 'KeyW',                                                                  
    '65': 'KeyA',                                                                  
    '83': 'KeyS',                                                                  
    '68': 'KeyD'                                                                   
    }
```

`src/systems/userinput/bindings/keyboard-mouse-user.js` is using for example
` paths.device.keyboard.key("w")` so hard coding the letter "w" instead of using the key code.
So that's the difference it seems.

vincentfretin:
On a French keyboard, on keydown, you get
event.code, event.keyCode, event.key
[...]

PR #3617, opened 2020-12-23
Use event.code and not event.key for wasdqe controls so this works for non QWERTY keyboards

vincentfretin:
This fixes https://github.com/mozilla/hubs/issues/2397

### DIFF
--- a/src/systems/userinput/devices/keyboard.js
+++ b/src/systems/userinput/devices/keyboard.js
@@ -1,5 +1,15 @@
 import { paths } from "../paths";
 import { ArrayBackedSet } from "../array-backed-set";
+
+const CODE_TO_KEY = {
+  "KeyW": "w",
+  "KeyA": "a",
+  "KeyS": "s",
+  "KeyD": "d",
+  "KeyQ": "q",
+  "KeyE": "e",
+};
+
 export class KeyboardDevice {
   constructor() {
     this.seenKeys = new ArrayBackedSet();
@@ -45,7 +55,10 @@ export class KeyboardDevice {
         this.keys = {};
         this.seenKeys.clear();
       } else {
-        const key = event.key.toLowerCase();
+        let key = event.key.toLowerCase();
+        // Use event.code and not event.key for wasdqe controls so this works
+        // for non QWERTY keyboards.
+        key = (event.code && CODE_TO_KEY[event.code]) || key;
         this.keys[key] = event.type === "keydown";
         this.seenKeys.add(key);
       }


src/systems/userinput/devices/keyboard.js
```diff
@@ -45,7 +55,10 @@ export class KeyboardDevice {
         this.keys = {};
         this.seenKeys.clear();
       } else {
-        const key = event.key.toLowerCase();
+        let key = event.key.toLowerCase();
+        // Use event.code and not event.key for wasdqe controls so this works
+        // for non QWERTY keyboards.
+        key = (event.code && CODE_TO_KEY[event.code]) || key;
```
Slayug, line 61:
```suggestion
        // Use event.code and not event.key for wasdqe controls so this works
        // for non QWERTY keyboards.
        const key = CODE_TO_KEY[event.code] ?? event.key.toLowerCase();
```
`event.code` always truthy in modern browser, so `key = CODE_TO_KEY[event.code] || event.key.toLowerCase();`
`event.code` always falsy in old browser, so `key = event.key.toLowerCase();`

You can support backward compatibility with:
```
export const KEYCODE_TO_KEY = {
  65: "q",
  68: "d",
  69: "e",
  81: "a",
  83: "s",
  90: "w"
};
// https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code#browser_compatibility
// https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode#browser_compatibility
const key = CODE_TO_KEY[event.code] ?? KEYCODE_TO_KEY[event.keycode] ?? event.key.toLowerCase()
```

Slayug (changes requested):
Hi, short code and backward compatibility suggestions

src/systems/userinput/devices/keyboard.js
vincentfretin: You are mixing two different things, code "KeyW", and keycode "90". There is no need for a mapping between keycode to key here, we already have to key in `event.key`.

"event.code always truthy in modern browser, so key = CODE_TO_KEY[event.code] || event.key.toLowerCase();
event.code always falsy in old browser, so key = event.key.toLowerCase();"
Yes this is exactly what I want. For old browsers, they don't have the required thing to know which key is pressed (event.code), then we default to the wrong qwerty behavior, no choice.

vincentfretin: For your other way to write it
`const key = CODE_TO_KEY[event.code] ?? event.key.toLowerCase();`
This should work I guess yes.
I'll leave this decision to Mozilla if they decide to include a fix for this issue. I won't spend time to modify this PR.

Slayug: I've test it works and more clean for the final JS code. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

Yes you are right for the backward thing.

netpro2k:
One thing to note with this solution is that on some layouts some key mappings may be completely masked, where the physical hardcoded keys here overlap with a key mapped by character elsewhere. To fix this we probably need to switch all mappings to be based on physical location rather than character. We also eventually would want to let the users customize these keys and the problem largely goes away.

This also means that for non QWERTY users, any of the tutorial tips and documentation are going to be incorrect. The tutorial tips are fixable using the experimental `navigator.keyboard.getLayoutMap()` API, but this is currently only supported in Chrome. The documentation is currently static, so we would probably need to do some work to get arbitrary javascript onto those pages to dynamically explain the key mappings... Instead we may want to just show an image of the "reference keyboard" so people can just compare it to their physical layout https://www.w3.org/TR/uievents-code/#key-alphanumeric-writing-system. 

I have added a fix for the in client tutorial tips at least https://github.com/mozilla/hubs/commit/b99a0c91463d0ccacfbd2565dc7dc6cd5a51cda9.

Even given these tradeoffs this solution feels like its better than the status queue, and probably provides an overall better experience for users of non QWERTY layouts. Keys that it might mask are certainly less important than the primary movement keys.

I am going to merge it with my patch and a TODO to improve it later. Thanks for the contribution @vincentfretin, sorry again we are so behind on accepting PRs.

Issue #3492, opened 2020-12-04
[New UI] A prompt for generating a join code remains displayed when clicking the Back arrow from the "Enter on Device" prompt if first the Esc key is pressed


Softvision-RemusDranca:
**[Affected Versions]:**
- Firefox Release 83.0
- Chrome 87.0.4280.88

**[Affected Platforms]:**
- Windows 10x64

**[Prerequisites]:**
- Have a room with the new UI created.

**[Steps to reproduce]:**
1. Open the room from prerequisites.
2. Click the "Enter On Device" button.
3. Press the "Esc" key while the "Enter on Device" prompt is displayed.
4. Click the "Back" arrow button from the "Enter on Device" prompt.
5. Observe the behavior.

**[Expected results]:**
- The same prompt when the room is first opened is displayed.

**[Actual results]:**
- A prompt with the message "Generating join code" is displayed.

**[Notes]:**
- The prompt displayed after reproducing the issue can't be dismissed by clicking the "Back" arrow.
- Attached is a screen recording:
![GeneratingCode](https://user-images.githubusercontent.com/40061643/101183644-73a2c080-3658-11eb-9c35-ecef155a9519.gif)

Softvision-RemusDranca:
I have verified this issue and it's no longer reproducible using App version: 1.0.0.20210108173817. The prompt for generating a join code no longer remains displayed when clicking the Back arrow from the "Enter on Device" prompt if first the Esc key is pressed.
Tested on Windows 10x64 using Firefox Release 84.0.2 and Chrome 87.0.4280.141.

Considering this I'll close this issue as verified fixed.

PR #3650, opened 2021-01-05
Bug/escape key in modal

johnshaughnessy:
Fix https://github.com/mozilla/hubs/issues/3492

### DIFF
--- a/src/react-components/profile-entry-panel.js
+++ b/src/react-components/profile-entry-panel.js
@@ -134,12 +134,12 @@ export default class ProfileEntryPanel extends Component {
         this.props.mediaSearchStore.sourceNavigateWithNoNav("avatars", "use");
       },
       onSubmit: this.saveStateAndFinish,
-      showBackButton: this.props.showBackButton,
       onClose: this.props.onClose,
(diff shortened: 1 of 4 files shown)


src/react-components/profile-entry-panel.js
```diff
@@ -134,12 +134,12 @@ export default class ProfileEntryPanel extends Component {
...
       },
       onSubmit: this.saveStateAndFinish,
-      showBackButton: this.props.showBackButton,
       onClose: this.props.onClose,
       onBack: this.props.onBack
     };
 
     if (this.props.containerType === "sidebar") {
+      avatarSettingsProps.showBackButton = this.props.showBackButton;
       return <AvatarSettingsSidebar {...avatarSettingsProps} />;
```
robertlong, line 143: Rather than mutating the object in a render method, you can just pass in the prop:
`<AvatarSettingsSidebar {...avatarSettingsProps} showBackButton={this.props.showBackButton} />`

Generally we should avoid mutation in render methods, even though in this case it seems acceptable.

robertlong (approved):
One minor nit, otherwise lgtm!

Issue #3317, opened 2020-11-11
Right clicking has weird behavior with object menu


heaversm:
**Description**

The video linked below in screenshots illustrates a few potential issues I encountered when adding content via the new UI.

* the old ui for media options can still be seen by clicking spacebar on a placed item.
* it is difficult to right click and hold and then left click over an option to delete/view/link/pin
* hitting "view" places the object in a strange location
* link goes to a dead URL
* sign-in doesn't prompt for a password
* cancel button after sign-in attempt should perhaps just be "close" or "ok?"

**To Reproduce**

See video below

**Expected behavior**

* remove old ui for placed object options
* add ability to reveal a placed object's options without having it disappear after releasing right click / spacebar.
* "view" should place the object directly in your view in a flat, non-skewed position
* "link" should either go to the media link or not be selectable if it is not link-able.
* sign-in should prompt for password, not send an e-mail (?)

**Screenshots**

[Click here for video of issues](https://share.getcloudapp.com/OAuJAmAw)

@mbransn

mbransn:
@robertlong @misslivirose can you let us know if any of the UI mentioned here was downgraded for implementation with the upcoming release?

misslivirose:
We are not changing the in-world UI as part of the refresh. The UI updates only change the 2D client UI. The object menu is not yet changing because it needs significant design work both visually and from an interaction standpoint. We are similarly not making interaction changes to the inspect mode (which is what is triggered holding right click over an object). 

Having the 'link' option appear on a screen/camera share seems to be a bug, as does having the object controls in inspect mode be visible (I just checked on the current UI, which doesn't have a link for camera objects and inspect mode does not have the controls to delete, etc. objects).

robertlong:
We need clarification on the right click behavior. We have three methods of selecting an object right now.

1. Select from object list. This shows the object menu as you would expect.
2. Right click on an object to inspect. This triggers the same object menu because the underlying state changes. But as soon as you release it un-inspects the object again.
3. Press `i` when hovering on an object.

I can see two pretty simple methods of resolving 2:

1. Differentiate between temporary inspecting an object and selecting an object in the object menu or with the `i` hotkey.
2. Make them all use the same behavior by making right click select the object.

robinkwilson:
The sticky view behavior isn't discoverable using **"F" key** OR using the **object menu**.

PR #3655, opened 2021-01-06
Room UI Redesign: Toggle inspect mode and lights

robertlong:
This PR makes the inspect mode toggle on/off by using the `f` key or right clicking while an object is under the cursor. It also brings back the toggle lights button so that you can see the background environment while viewing an object.

Fixes https://github.com/mozilla/hubs/issues/3317


src/systems/camera-system.js
```diff
@@ -252,13 +252,12 @@ export class CameraSystem {
     this.mode = NEXT_MODES[this.mode] || 0;
   }
 
-  inspect(el, distanceMod, temporarilyDisableRegularExit, fireChangeEvent = true) {
```
johnshaughnessy, line 255: Glad to see us get rid of `temporarilyDisableRegularExit`

johnshaughnessy (approved):
lgtm!

PR #3662, opened 2021-01-06
ICE retries and RTC fixes

keianhzo:
- Now we listen to **connectionstatechange** events and retry ICE candidates negotiation
- A watchdog has been added to monitor the state of the ICE connection state and retry candidates negotiation in case it's failed or disconnected.
- New UI design fixes
- Change consumer/producer background to red when there is no traffic to outline a possible error better.
- Always show the candidates status and set background to red in case they couldn't be gathered.
- Fix the 00 shown when there are no local/remote consumers.


keianhzo:
@robertlong This should be ready too.

Issue #3198, opened 2020-10-20
Change "Kick" user text


emclaren:
When you want to remove someone from the room, the text copy is "Kick" which we've received a few reports that it's perhaps a bit harsh. 

Perhaps we could change to "remove user"  or something else which is more clear/less violent 

see this tweet: https://twitter.com/morionart/status/1318517465429741568

also mentioned to us by Megan on the design team

rajreet:
I would like to work on this issue. I am a complete beginner to open source, It would be great if someone guides me.

Mansibisen:
@emclaren  I will like to work on this issue . Can you guide a bit about bug ?

SolKuczala:
hi, I'll take a chance

johnshaughnessy:
It seems like there are two different issues here:
- "Add a send-to-lobby button for moderators." (The tweet reads to me like a feature request for managing who is in the room and who is in the lobby -- not for responding to bad actors.)
- "Change the language around removing bad actors from the room." 

The first issue requires some work to de-spawn your avatar and maybe also release ownership or destroy temporary objects you've added to the room. (This is the same work we need to do for an individuals to be able to click a button and send themselves back to the lobby.)

My only ask for addressing the second issue is to keep the expectation that removing the user is temporary / easily circumvented by the bad actor. Here "kick" is chosen in contrast to "ban" (an option we don't have) so that moderators are not surprised if the same user refreshes the page and joins again. If we switch the action text to something like "remove user", I would encourage us to include further details explaining what this does, and what options people have for locking down the room to prevent a bad actor from joining back in (e.g. tie the room to Discord permissions, switch it to invite-only mode, or manage accounts/entry in a hubs cloud instance.)

johnshaughnessy:
Separately, we may want to work against the redesigned UI because the translations files have changed significantly and it seems good to avoid merge conflicts.

SolKuczala:
ok, so this wasn't only a re-naming problem. Maybe I can't help right now as I'm not having time to focus a lot these weeks. It requires some insight from the project and trying to join discord I couldn't see a channel where I could ask someone about it. Leave open to somebody else to tackle in ;)

PR #3671, opened 2021-01-06
Change all the ocurrences of "kick" variables, references and kick-button.js file

SolKuczala:
fixes bug #3198


johnshaughnessy:
I am concerned that if the only thing we do is change the word `kick` to the word `remove`, moderators will expect room users to be prevented from rejoining the room. https://github.com/mozilla/hubs/issues/3198#issuecomment-763027885

robertlong:
We want to look into this further to make sure that we are properly addressing the original issue and we need to update the translations after the new design. We're going to close and revisit this PR.

PR #3692, opened 2021-01-11
RTC retry improvements

keianhzo:
- Added some more missing retry cases
- Added missing Recv watchdog stop when disconnecting
- Fixed comment typos
- Reduced retry time to 2secs
- RTC Log clear/download
- Pause remote server stats when collapse (start with the remote panel collapsed)

### DIFF
--- a/src/naf-dialog-adapter.js
+++ b/src/naf-dialog-adapter.js
@@ -30,9 +30,13 @@ const PC_PROPRIETARY_CONSTRAINTS = {
   optional: [{ googDscp: true }]
 };
 
+const ICE_RECONNECT_INTERVAL = 2000;
+const INITIAL_ROOM_RECONNECTION_INTERVAL = 2000;
+
+const isFirefox = navigator.userAgent.toLowerCase().indexOf("firefox") > -1;
+
 export default class DialogAdapter {
   constructor() {
-    this._forceTcp = false;
     this._timeOffsets = [];
     this._occupants = {};
     this._micProducer = null;
@@ -50,13 +54,24 @@ export default class DialogAdapter {
     this._blockedClients = new Map();
     this.type = "dialog";
     this.occupants = {}; // This is a public field
-    this._sendIceParameters = {};
-    this._recvIceParameters = {};
+    this._forceTcp = false;
+    this._forceTurn = false;
+    this._iceTransportPolicy = "all";
+    this._iceReconnectSendTaskId = null;
+    this._iceReconnectRecvTaskId = null;
+    this._reconnectionDelay = INITIAL_ROOM_RECONNECTION_INTERVAL;
+    this._reconnectionTimeout = null;
+    this._reconnectionAttempts = 0;
+    this._lastSendConnectionState = null;
+    this._lastRecvConnectionState = null;
+    this._sendTaskId = null;
+    this._recvTaskId = null;
+    this._closed = true;
     this.scene = document.querySelector("a-scene");
   }
 
-  setForceTcp(forceTcp) {
-    this._forceTcp = forceTcp;
+  get serverUrl() {
+    return this._serverUrl;
   }
 
   setServerUrl(url) {
@@ -67,14 +82,43 @@ export default class DialogAdapter {
     this._joinToken = joinToken;
   }
 
-  setPeerConnectionConfig(peerConnectionConfig) {
-    if (peerConnectionConfig.iceServers) {
-      this._iceServers = peerConnectionConfig.iceServers;
+  setTurnConfig(forceTcp, forceTurn) {
+    this._forceTcp = forceTcp;
+    this._forceTurn = forceTurn;
+
+    if (this._forceTurn || this._forceTcp) {
+      this._iceTransportPolicy = "relay";
+    }
+  }
+
+  getIceServers(host, port, turn) {
+    const iceServers = [];
+
+    this._serverUrl = `wss://${host}:${port}`;
+
+    if (turn && turn.enabled) {
+      turn.transports.forEach(ts => {
+        // Try both TURN DTLS and TCP/TLS
+        if (!this._forceTcp) {
+          iceServers.push({
+            urls: `turns:${host}:${ts.port}`,
+            username: turn.username,
+            credential: turn.credential
+          });
+        }
+
+        iceServers.push({
+          urls: `turns:${host}:${ts.port}?transport=tcp`,
+          username: turn.username,
+          credential: turn.credential
+        });
+      });
+      iceServers.push({ urls: "stun:stun1.l.google.com:19302" });
+    } else {
+      iceServers.push({ urls: "stun:stun1.l.google.com:19302" }, { urls: "stun:stun2.l.google.com:19302" });
     }
 
-    if (peerConnectionConfig.iceTransportPolicy) {
-      this._iceTransportPolicy = peerConnectionConfig.iceTransportPolicy;
-    }
+    return iceServers;
   }
 
   setApp() {}
@@ -106,6 +150,11 @@ export default class DialogAdapter {
    * Gets transport/consumer/producer stats on the server side.
    */
   async getServerStats() {
+    if (this.getConnectStatus() === NAF.adapters.NOT_CONNECTED) {
+      // Signaling channel not connected, no reason to get remote RTC stats.
+      return;
+    }
+
     const result = {};
     try {
       if (!this._sendTransport?._closed) {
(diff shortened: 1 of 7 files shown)


src/naf-dialog-adapter.js
```diff
@@ -143,26 +192,159 @@ export default class DialogAdapter {
...
+   * Checks the Send Transport ICE status and restarts it in case is in failed or disconnected states.
+   * This is called by the Send Transport "connectionstatechange" event listener.
+   * @param {boolean} connectionState The transport connnection state (ICE connection state)
+   */
+  checkSendIceStatus(connectionState) {
+    // If the ICE connection state failed we force an ICE negotiation
+    if (connectionState === "failed" || connectionState === "disconnected") {
+      this.restartSendICE(false);
+    } else if (connectionState === "connected") {
+      this._iceReconnectSendTaskId && clearTimeout(this._iceReconnectSendTaskId);
```
robinkwilson, line 246: Is this a new javascript syntax for a simple if statement? That's awesome!

src/naf-dialog-adapter.js
```diff
@@ -143,26 +192,159 @@ export default class DialogAdapter {
...
+  }
+
   /**
    * Restart ICE in the underlying receive peerconnection.
+   * @param {boolean} force Forces the execution of the reconnect.
    */
-  async restartRecvICE() {
+  async restartRecvICE(force = true) {
+    // Renegotiate ICE in case ICE state is failed, disconnected. The last condition cover the cases where the
+    // Send transport has been create but for whatever reason is stuck in new state and doesn't gather candidates.
```
robinkwilson, line 283:
```suggestion
    // send transport has been create but for whatever reason is stuck in new state and doesn't gather candidates.
```
semantic change, let's me know it's in the middle of the sentence, up to you to change.

src/naf-dialog-adapter.js
```diff
@@ -143,26 +192,159 @@ export default class DialogAdapter {
...
     if (!this._recvTransport?._closed) {
-      this.emitRTCEvent("log", "RTC", () => `Restarting receive transport ICE`);
-      this._recvTransport.restartIce({ iceParameters: this._recvIceParameters });
+      if (
+        force ||
+        this._recvTransport.connectionState === "failed" ||
+        (!isFirefox && this._sendTransport.connectionState === "disconnected") ||
+        (this._recvTransport.connectionState === "new" && this._lastRecvConnectionState === "failed")
```
robinkwilson, line 289: Should this block be its own function? It _seems_ like it's the same code twice here and line 214

src/naf-dialog-adapter.js
```diff
@@ -143,26 +192,159 @@ export default class DialogAdapter {
...
-      this._sendTransport.restartIce({ iceParameters: this._sendIceParameters });
+      // Renegotiate ICE in case ICE state is failed, disconnected. The last two conditions cover the cases where the
+      // Send transport has been create but for whatever reason is stuck in new state and doesn't gather candidates.
+      if (
+        force ||
+        this._sendTransport.connectionState === "failed" ||
+        (!isFirefox && this._sendTransport.connectionState === "disconnected") ||
+        (this._sendTransport.connectionState === "new" && this._lastSendConnectionState === "failed")
+      ) {
+        this.emitRTCEvent("log", "RTC", () => `Restarting send transport ICE`);
```
robinkwilson, line 220: See line 289 (since looks like this is the same codeblock)

robinkwilson (commented):
A few comments + small code refactor. I appreciate your fantastic comments that make the functions/code blocks really clear.

src/naf-dialog-adapter.js
```diff
@@ -143,26 +192,159 @@ export default class DialogAdapter {
     }
   }
 
+  /**
+   * This applies to the ICE restart method below.
+   * Why do we need new ICE servers? In case the ICE fails we need to re-negotiate new candidates
+   * but most probably our TURN credentials have expired so we need to request new credentials and
+   * update the TURN servers to be able to gather new candidates.
+   * Firefox doesn't support hot updating ICE servers through mediasoup's updateIceServers (which
```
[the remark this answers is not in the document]
keianhzo: Mediasoup is the underlying WebRTC stack that we are using in Dialog.

src/naf-dialog-adapter.js
```diff
@@ -143,26 +192,159 @@ export default class DialogAdapter {
     }
   }
 
+  /**
+   * This applies to the ICE restart method below.
+   * Why do we need new ICE servers? In case the ICE fails we need to re-negotiate new candidates
+   * but most probably our TURN credentials have expired so we need to request new credentials and
```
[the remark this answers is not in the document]
keianhzo: Fixed

src/naf-dialog-adapter.js
```diff
@@ -143,26 +192,159 @@ export default class DialogAdapter {
...
+
   /**
    * Restart ICE in the underlying send peerconnection.
    */
-  async restartSendICE() {
+  async restartSendICE(force = true) {
     if (!this._sendTransport?._closed) {
-      this.emitRTCEvent("log", "RTC", () => `Restarting send transport ICE`);
-      this._sendTransport.restartIce({ iceParameters: this._sendIceParameters });
+      // Renegotiate ICE in case ICE state is failed, disconnected. The last two conditions cover the cases where the
```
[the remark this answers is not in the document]
keianhzo: Updated

src/naf-dialog-adapter.js
```diff
@@ -143,26 +192,159 @@ export default class DialogAdapter {
...
+      // Send transport has been create but for whatever reason is stuck in new state and doesn't gather candidates.
```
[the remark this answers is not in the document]
keianhzo: Fixed

src/naf-dialog-adapter.js
keianhzo: You can see it that way, it's just a slightly more concise and idiomatic way of doing a condition check and save some lines of code. Same as the safe call operator (?.)

src/naf-dialog-adapter.js
keianhzo: Updated

src/naf-dialog-adapter.js
keianhzo: As they are using different transports that function would end up not saving that much code lines. But I just realized that I was referencing the wrong transport in `restartRecvICE`. I've fixed that.

robinkwilson: Nice catch!

robinkwilson (approved):
LGTM

vincentfretin:
Be careful about the RTCPeerConnection disconnected state. Did you try on iOS? Doing a ICE restart right away when a RCTPeerConnection goes in disconnected state is a bad idea. You should at least wait a few seconds to see if the state doesn't go back to connected by itself. Did you see my recent tests on the janus adapter on https://github.com/mozilla/naf-janus-adapter/pull/107#issuecomment-744591501 ?

keianhzo:
@vincentfretin Thanks for that. For what I have seen the recovery time from disconnects can vary and some times they don't even recover at all and transition to a failed state depending of the type of disconnection. Anyway I've introduced a delay of two secs for ICE starts to give a chance for recovery.

vincentfretin:
Great work @keianhzo You'll see if 2s is enough. Did you test the changes on Safari iOS?

In my app I'm using a modified version of naf-janus-adapter 3.0.20 and finally I don't do the reconnect if a peer connection is in state disconnected, I reverted that. It does a full reconnect after 10s if a peer connection is in state failed or after 40s if the websocket timed out. I could have tried to reconnect if a peer connection is in state disconnected for 10s, but I didn't do it because the janus session is only ended after 40s of inactivity anyway. See the discussion https://github.com/mozilla/janus-plugin-sfu/pull/73

I don't know how it works with mediasoup, does mediasoup detects that the connection is dead and ends it? Can you have several send transport for the same peerId? With janus sfu, we can only have one publisher for a user id, this is why I can't reconnect before 40s.

keianhzo:
The `RTCPeerConnection` state transition does not depend on Mediasoup but on the underlying browser WebRTC implementation and that varies across browsers. From my experience most browsers recover from a disconnect but there are certain cases where they don't or the don't do it fast enough. I could emulate one of this cases in Chrome quickly bringing down and up the TURN server while using TURN. In that case FF seems to recover quickly but Chrome stays in disconnected state. That's why the ICE restart in that case. Even in the case where Chrome recovers from that at some point, we rather not letting the user down for so long.

PR #3707, opened 2021-01-13
Update react-admin version v2 -> v3

robinkwilson:
Followed migration instructions here: https://github.com/marmelab/react-admin/blob/master/UPGRADE.md

Still gets this error on the scene or avatars pages when there are no scenes or no avatars.
We'd need to update the responses to: `{ data: [], total: 0}` to get rid of errors when data is an empty array. In the previous version this wasn't displaying. 

image

With the NaN in upper right hand corner:

image

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-823)


admin/src/admin.html
```diff
@@ -28,7 +28,10 @@
   a {
     color: #FF3464;
   }
-    </style>
+  .no-margin > div {
+    margin: 0px;
+  }
+  </style>
```
robinkwilson, line 34: Not ideal solution but selects the correct element to remove the 48px margin-top it was adding. I could not find a way to do this cleaner.

brianpeiris (changes requested):
LGTM in general, but I found a bug where the admin panel fails to add an identity to an account.

image

netpro2k:
Things have diverged and admin panel will likely be reworked significantly in the future, so closing this for now.

PR #3726, opened 2021-01-19
Fix loading of custom navmeshes

netpro2k:
Navmeshes exported from Blender were not loading correctly. I was able to trace it down to the conversion from BufferGeometry to Geometry. Several issues around this were fixed in upstream THREE, so I backported this function. 

We could put this into our custom THREE branch instead of inline, but Geometry is now deprecated in THREE and going to be dropped in the next version, so we will need to get rid of this eventually anyway.

Also adds a `debugNavmesh` query param that will show a wireframe version of the navmesh (used this to find the issue initially, but seems generally useful).

### DIFF
--- a/src/systems/nav.js
+++ b/src/systems/nav.js
@@ -1,4 +1,114 @@
 const { Pathfinding } = require("three-pathfinding");
+import qsTruthy from "../utils/qs_truthy";
+
+const Vector3 = THREE.Vector3;
+const Vector2 = THREE.Vector2;
+const Face3 = THREE.Face3;
+
+// TODO this is backported from a later version of THREE.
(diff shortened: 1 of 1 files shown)


robertlong (approved):
LGTM, I'm a little concerned about the perf of this on big nav meshes, but I think it'll get resolved when we upgrade Three.

rawnsley:
> LGTM, I'm a little concerned about the perf of this on big nav meshes, but I think it'll get resolved when we upgrade Three and then update the pathfinding library.

@robertlong I've addressed at least one problem with [big nav meshes](https://github.com/MozillaReality/three-pathfinding/pull/1).

Issue #3702, opened 2021-01-13
Chat message do not preserve line breaks (shift + enter)


robinkwilson:
**Description**
Chat messages do not preserve line breaks from text with enters in the chatbox.

![image](https://user-images.githubusercontent.com/29560735/104500101-d14cf580-5592-11eb-8433-0a99bf599737.png)

![image](https://user-images.githubusercontent.com/29560735/104500112-d4e07c80-5592-11eb-8a06-89bbffae021e.png)

**Expected behavior**
Preserve line breaks.

Issue #3694, opened 2021-01-11
Viewport is not correctly updated after a fullscreen in Fenix


keianhzo:
**Description**
When entering fullscreen mode after entering the room, the viewport seems not to be correctly updated.

**To Reproduce**
1. Open a room in Fenix
2. The viewport is correct
2. Enter the room (will trigger a fullcreen)
3. The viewport is not correct

**Expected behavior**
After entering the room the viewport should be correct. It doesn't seem to be updated after entering fullscreen mode.

**Screenshots**
![screen](https://user-images.githubusercontent.com/837184/104215219-c41bf380-5438-11eb-86e2-85a7d68ee289.png)

**Hardware** 
 - Device: Galaxy S8
 - OS: Android 9
 - Browser: Fenix

**Additional context**
This doesn't seem to happen in Android Chrome.

PR #3727, opened 2021-01-19
Room UI Redesign: Miscellaneous Fixes

robertlong:
- Center the hero content on mobile
- Fixes chat message line height
- Fixes the modal width for the verify auth page and all other modals on mobile
- Fixes https://github.com/mozilla/hubs/issues/3702
- Fixes https://github.com/mozilla/hubs/issues/3694


robertlong:
Couldn't test the chat message spawning locally so I tested on staging. There seems to be some issues with the new presence-log styles. Fixing now.

robertlong:
I couldn't find the right SVG foreign element CSS incantation so I fixed the chat message rendering with html2canvas which needs thorough testing, particularly with VR perf. I could have rendered chat bubbles manually, but multi-line text, emojis etc. would have been a lot of work. If perf isn't an issue I think html2canvas is easier to maintain as we can more or less style chat messages as we want. It could also be used for more things in the future.

src/react-components/home/HomePage.scss
```diff
@@ -37,6 +37,7 @@
   display: flex;
   flex-direction: column;
   align-items: center;
+  flex: 1;
```
johnshaughnessy, line 40: :muscle:

src/utils/chat-message.js
```diff
@@ -4,57 +4,33 @@ import { toArray as toEmojis } from "react-emoji-render";
...
 
-  const cleanedBody = (monospace ? body.substring(1, body.length - 1) : body).trim();
+  const multiline = messages.length > 1;
 
-  const bodyWithEmojis = toEmojis(cleanedBody);
+  let emoji = false;
 
-  const formattedBody = <Linkify properties={{ target: "_blank", rel: "noopener referrer" }}>{bodyWithEmojis}</Linkify>;
+  if (messages.length === 1) {
+    const emojiComponents = toEmojis(cleanedBody);
```
johnshaughnessy, line 25: i am a potato
here are some emojis

src/react-components/chat-message.js
```diff
@@ -4,12 +4,10 @@ import PropTypes from "prop-types";
 import styles from "../assets/stylesheets/presence-log.scss";
 import classNames from "classnames";
 import { toArray as toEmojis } from "react-emoji-render";
-import serializeElement from "../utils/serialize-element";
+import html2canvas from "html2canvas";
```
johnshaughnessy, line 6: I don't have a VR headset handy to test perf... (you mentioned in your comments might be a problem)

But it seems good that `html2canvas` only pulls in a single dependency.

johnshaughnessy (approved):

src/utils/chat-message.js
robertlong: :open_hands: :potato:

Issue #3728, opened 2021-01-20
HMC "Whats New" page is blank


emclaren:
The what's new page on hubs.mozilla.com is no longer working properly. None of the content from Github is being populated

https://hubs.mozilla.com/whats-new

We first noticed this problem when testing the new UI last week
<img width="1769" alt="Screen Shot 2021-01-19 at 4 54 02 PM" src="https://user-images.githubusercontent.com/6835324/105112012-f8b22f80-5a76-11eb-959e-48fa2acc53e3.png">

PR #3729, opened 2021-01-20
Update PAT for whats-new page

johnshaughnessy:
Fix https://github.com/mozilla/hubs/issues/3728

The API returned `401 Unauthorized` responses with the old token. I do not know what happened with the old token (was it revoked or rate limited?) but replacing it seems to fix the "What's New" page. 

At some point in the future, we might want to consider hiding this token in a reticulum config and delivering the json response to the client. That way, there is a lower risk that the token can be taken and used maliciously. (The token only has the `public_scope` so it can't be used to do much besides spam the API with requests. )

### DIFF
--- a/src/whats-new.js
+++ b/src/whats-new.js
@@ -30,9 +30,14 @@ class WhatsNew extends Component {
     currentDate: null
   };
   async getNotes(page) {
+    // TODO Move this request to reticulum and return the results to the client
     const endpoint = "https://api.github.com/repos/mozilla/hubs/pulls";
     // Read-only, public access token.
-    const token = "5ce5d912b79609593aa6c1929aa5034f84515a0e";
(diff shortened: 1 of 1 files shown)


johnshaughnessy:
It looks like github notices tokens and revokes them. I received this email after publishing the first commit:

> We noticed that a valid OAuth, GitHub App or Personal Access Token of yours was committed to a GitHub repository. Disclosing a valid access token would allow other people to interact with GitHub on your behalf, potentially altering data, your contact information, and billing data.

> As a precautionary measure, we have revoked the token. You will need to generate a new token for the app to authenticate to GitHub.

I think we need to move this request to reticulum, but this fixes the page in the short term.

brianpeiris (approved):
LGTM. I agree we should move this to reticulum, if we want to keep it around.

Issue #3724, opened 2021-01-18
Functional waypoint scaling 


paulmelis:
**Is your feature request related to a problem? Please describe.**

Currently it seems that setting a non-unit scale on a waypoint does more-or-less work, but has some downsides. The use case we're experimenting with is a historical 3D model of part of a city (basically a city block), where we want to start viewing the model from high above for explaining context, layout, etc. And then use a waypoint to teleport into the city itself, at life-size scale, to get a feel for walking around in this historical model. And then get back to the overview point. The 3D model is actually quite large, around 2M polygons and lots of textures, but so far performance both on desktop and Quest has been fine!

We've tried two approaches so far:

* Setting a scale of 0.01 on the model (which has meters as units) and adding a waypoint in the city also with scale 0.01. The spawn point in the hubs room (scale 1) then starts with the high overview point. Although teleporting to the city waypoint works, it seems the scale is actually off (we're too high over the model). There's also some strange things with mouse interaction no longer fully working (e.g. no mouse cursor visible anymore and can't select anything, this is in both firefox and chrome).

* Setting a scale of 1 on the 3D model, adding a waypoint in the city, also with scale 1. The spawn point has scale 100 and starts high above. This actually works quite well and the scale after teleporting to the city waypoint seems to be correct. We also added a second waypoint close to the spawn point in order to get back to the overview point and scale (as there seems to be no easy way to jump back to the spawn situation). However, this setup highlights interesting behaviour: while it is possible to select the city waypoint when located in the overview waypoint, it is not possible to select the overview waypoint when located in the city waypoint. This apparently has to do with the selection distance used, which gets influenced by the scaling. It's only when flying really close to the overview waypoint that it becomes possible to select it.

**Describe the solution you'd like**

[...]

PR #3741, opened 2021-01-21
Fix waypoint teleport height

johnshaughnessy:
In a previous PR https://github.com/mozilla/hubs/pull/2226 , I changed how the camera height was calculated when traveling to a `waypoint`. I cited this behavior as a bug:

> when we spawned at a waypoint that was not snapped to the nav mesh, we used a default height of 1.6 meters. If the current player height was different than that, we would end up above or below the floor. 

But I believe this diagnosis was incorrect, and only the second thing listed in the PR was a problem:
> we need to delay spawning at the waypoint

My "fix" made it impossible to create vr-enabled `waypoint` "seats" like those pictured below:
image
(Credit Discord user Dreamwriter#8148 . These should be lowered so that the robots are in the car, and `"Snap to floor plan"` should be disabled.)

If a `waypoint` does not have `"Snap to floor plan"` enabled, then:
- fly mode should be enabled (with `shouldLandWhenPossible` enabled too)
- the `avatar POV node` should be aligned with the robot's eyes. (I am referring to the robot that appears when the `waypoint` icon is hovered or appears in Spoke when you place a `waypoint` in a scene.)

This second objective is achieved by using the hard-coded value of 1.6 meters (the height of the robot's eyes above the waypoint transform) when calculating the `finalPOV`. Note that this is wrong for waypoints that are not at unit scale `(1,1,1)`, but I did not want to address scale issues in this PR. There is a separate issue tracking problems with waypoint scaling https://github.com/mozilla/hubs/issues/3724

If a waypoint author wants to align people based on the floor position, they need to use `"Snap to floor plan"`. There is no option to create a waypoint that is not snapped to the floor plan that will align people by their "floor point".

Thanks @j-conrad and Discord user Dreamwriter#8148 for bringing this to my attention.

### DIFF
--- a/src/systems/character-controller-system.js
+++ b/src/systems/character-controller-system.js
@@ -125,8 +125,16 @@ export class CharacterControllerSystem {
         inPosition.setFromMatrixPosition(inMat4Copy);
         this.findPositionOnNavMesh(inPosition, inPosition, outPosition, true);
         finalPOV.setPosition(outPosition);
+        translation.makeTranslation(0, getCurrentPlayerHeight(), -0.15);
+      } else {
+        // If we are not snapping to the nav mesh, align the user's
+        // perspective exactly to the robot eyes as they appear in the
+        // waypoint indicator. (1.6 meters up and 0.15 meters forward)
+        // This does _not_ require taking the player's height into account
+        // on this line because we are only interested in where the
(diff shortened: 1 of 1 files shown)


netpro2k (approved):
lgtm

PR #3753, opened 2021-01-23
Update locale_config.js

cbo04:
add fr "Français"


cbo04:
> @cbo04 I think it could be great to have only on PR instead of these:
> 
>     #3753
>     #3754

@Slayug Do I close Pull request?

robertlong:
Closing and moving to https://github.com/mozilla/hubs/pull/3897

PR #3769, opened 2021-01-26
Add cookie-based authentication

johnshaughnessy:
See https://github.com/mozilla/reticulum/pull/461

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-822)

### DIFF
--- a/src/react-components/auth/AuthContext.js
+++ b/src/react-components/auth/AuthContext.js
@@ -5,7 +5,7 @@ import configs from "../../utils/configs";
 // TODO: We really shouldn't include these dependencies on every page. A dynamic import would work better.
 import jwtDecode from "jwt-decode";
 import AuthChannel from "../../utils/auth-channel";
-import { connectToReticulum } from "../../utils/phoenix-utils";
+import { connectToReticulum, getReticulumFetchUrl } from "../../utils/phoenix-utils";
 
 export const AuthContext = createContext();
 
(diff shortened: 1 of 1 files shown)


src/react-components/auth/AuthContext.js
```diff
@@ -65,6 +65,12 @@ export function AuthContextProvider({ children, store }) {
       const socket = await connectToReticulum();
       authChannel.setSocket(socket);
       await authChannel.verifyAuthentication(authParams.topic, authParams.token, authParams.payload);
+      await fetch(getReticulumFetchUrl("/api/v1/accounts/set_cookie"), {
```
netpro2k, line 68: Minor but I would use `fetchReticulumAuthenticated` here instead.

src/react-components/auth/AuthContext.js
```diff
@@ -65,6 +65,12 @@ export function AuthContextProvider({ children, store }) {
...
+        method: "GET",
```
netpro2k, line 69: This is an issue on the reticulum side, but since I noticed it here. These 2 calls are mutating state, so they should not be GET requests. This isn't just a semantic issue either, it can lead to some nasty hard to debug issues when intermediate parties (rightfully) assume they can call this url as much as they like with no side effects.

takahirox:
I'm not sure if we should use Cookies because, If I remember correctly, Cookies will be deprecated.

johnshaughnessy:
Closing this PR because I abandoned this work a long time ago

PR #3811, opened 2021-01-30
Allow webxr gamepads that are not connected to update

johnshaughnessy:
It seems like devices from the [WebXR API Emulator](https://github.com/mozilla/hubs/compare/bug/webxr-extension?expand=1) do not identify as being `connected`. If we ignore this property, the extension works in hubs.

### DIFF
--- a/src/systems/userinput/devices/webxr-controller.js
+++ b/src/systems/userinput/devices/webxr-controller.js
@@ -22,7 +22,7 @@ export class WebXRControllerDevice {
     this.orientation = new THREE.Quaternion();
   }
   write(frame, scene, referenceSpace) {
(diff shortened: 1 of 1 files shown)


johnshaughnessy:
There is a related issue that has this call stack:
```js
userinput.js:417 Uncaught TypeError: Cannot set property 'isWebXRGamepad' of null
    at retrieveXRGamepads (userinput.js:417)
    at HTMLElement.<anonymous> (userinput.js:432)
    at HTMLElement.emit (a-node.js:263)
    at enterVRSuccess (a-scene.js:332)
    at requestSuccess (a-scene.js:289)
 ```
 
 But it does not cause the application to freeze and still I can control the avatar hands with the extension.

PR #3860, opened 2021-02-04
add the min-line height needed to display letter "g" properly to dropdown

csnktms:
Adding the `min` `line-height` needed, `"g"` character can be rendered properly.

Screenshot 2021-02-04 at 4 17 09


robertlong:
Looks good, thanks!

PR #3904, opened 2021-02-12
Add link behavior for images and videos

robertlong:


src/gltf-component-mappings.js
```diff
@@ -251,18 +258,24 @@ async function mediaInflator(el, componentName, componentData, components) {
...
 
+  if ((componentName === "video" || componentName === "image") && components.link) {
+    mediaOptions.href = sanitizeUrl(components.link.href);
+  }
+
   const src = componentName === "link" ? componentData.href : componentData.src;
 
   el.setAttribute("media-loader", {
-    src,
+    src: sanitizeUrl(src),
```
netpro2k, line 268: Hmm, sanitizing href makes sense since its becoming a clickable link. Not sure we need to do it for src since that is getting loaded via a media element... If we do want to do it we need to make sure it doesnt break things like the `hubs://` urls used for webcam/screen sharing.

netpro2k (commented):
lgtm if we confirm that sanitizing src isnt breaking sharing of webcam/screen.

src/gltf-component-mappings.js
robertlong: `src` is used as the clickable link if href isn't there. So we need to sanitize it here. Good catch though, I'll make sure `hubs://` urls are sanitized correctly.

robertlong: Tested the webcam, camera tool, and screenshare. All look to be working.

Issue #3911, opened 2021-02-15
iframe embedding not working in secure contexts such as incognito mode


rawnsley:
In [this simple example](https://alpha.avncloud.com/dev/persistent/hubs-iframe-bug.html) of a Hubs room embedded in an iframe the room loads fine in normal browser mode, but fails to load when viewed in Chrome's Incognito mode. It has also been reported on normal Chrome sessions, possibly where the user has adopted a more restricted default security model.

It appears to be because [access to the `localStorage` API is blocked](https://www.chromium.org/for-testers/bug-reporting-guidelines/uncaught-securityerror-failed-to-read-the-localstorage-property-from-window-access-is-denied-for-this-document) and an exception is thrown in [store.js](https://github.com/mozilla/hubs/blob/master/src/storage/store.js) like this:

```
store.js:215 Uncaught DOMException: Failed to read the 'localStorage' property from 'Window': Access is denied for this document.
    at new C (https://assets-prod.reticulum.io/hubs/assets/js/hub-596e542bb782311ce205.js:1:1082953)
    at new i (https://assets-prod.reticulum.io/hubs/assets/js/hub-596e542bb782311ce205.js:1:1085762)
    at Module.1666 (https://assets-prod.reticulum.io/hubs/assets/js/hub-596e542bb782311ce205.js:1:924116)
    at o (https://assets-prod.reticulum.io/hubs/assets/js/hub-596e542bb782311ce205.js:1:561)
    at n (https://assets-prod.reticulum.io/hubs/assets/js/hub-596e542bb782311ce205.js:1:430)
    at https://assets-prod.reticulum.io/hubs/assets/js/hub-596e542bb782311ce205.js:1:2534
    at https://assets-prod.reticulum.io/hubs/assets/js/hub-596e542bb782311ce205.js:1:2538
```
A causal search has yielded no obvious way around the problem. There is an [experimental option to prompt the user for access](https://developer.mozilla.org/en-US/docs/Web/API/Storage_Access_API/Using), but it would be hard to provide any meaningful context to the question, not to mention the friction it would create.

One solution would be more graceful fallback behaviour when `localStorage` is not available. Perhaps simply storing the data in an in-memory javascript map?

It does not appear to be a problem in Firefox's Private browsing mode.

Test platform: Chrome Version 89.0.4389.47 (Official Build) beta (x86_64) on MacOS

robertlong:
Hey @rawnsley thanks for the PR. Is this something you've been running into in your own work?

[...]

PR #3912, opened 2021-02-15
Potential workaround for localStorage exceptions

rawnsley:
Suggested fix for https://github.com/mozilla/hubs/issues/3911

You will need something more idiomatic for your codebase. This change on its own breaks some of the other pages like login.


robertlong:
Closing this PR so we can continue with a new solution of adding checks to the store class and anywhere else that uses localstorage directly.

PR #3922, opened 2021-02-17
Save lightsEnabled setting

johnshaughnessy:
This PR restores some functionality we removed during the UI redesign. It saves the `enableLights` state that effects whether the background is rendered when inspecting objects/avatars. This way users don't have to set it each time they join a room.


robertlong:
LGTM :+1:

PR #3925, opened 2021-02-17
Webcam Avatars

robertlong:
You can now share your webcam to your avatar! On supported avatars, a "Avatar Camera" button will appear in the Share menu. Click it to use your webcam with your avatar.

Avatar creators can add this functionality to their own avatars by adding the `webcam-texture-target` component to the material they want to set the webcam to. This will swap out the base color map with a video texture. The other maps and base color factor will still be applied, making it possible to better integrate the webcam into your avatar.

To Do:
- [ ] Finish featured webcam avatars
- [ ] Add `webcam-texture-target` instructions to docs
- [ ] Add screenshot to PR and tag with whats-new
- [ ] Investigate additional bandwidth requirements from increased webcam usage


src/components/webcam-texture-target.js
```diff
@@ -0,0 +1,84 @@
+import { disposeTexture } from "../utils/material-utils";
+
+AFRAME.registerComponent("webcam-texture-target", {
+  schema: {
+    src: { type: "string" }
+  },
+
+  getMaterial() {
```
netpro2k, line 8: Hmm, seems like we don't actually end up taking into account which material to replace based on what materials had the component on it. 

I would say we either want to account for that (probably by keeping track of it somehow at inflation time), or just have this component be something you add to a Mesh instead, and then keep the current behavior of just swapping that mesh's material (getting rid of the extra code around inflating material components).

src/components/webcam-texture-target.js
```diff
@@ -0,0 +1,84 @@
...
+
+      this.originalTexture = material.map;
+
+      NAF.connection.adapter.getMediaStream(streamClientId, "video").then(stream => {
+        if (src !== this.data.src) {
+          // Prevent creating and loading video texture if the src changed while we were fetching the video stream.
+          return;
+        }
+
+        const video = document.createElement("video");
```
netpro2k, line 47: Not a big deal since its just a few lines, but may want to use `createVideoOrAudioEl("video")` from media-views.js here.

src/components/webcam-texture-target.js
```diff
@@ -0,0 +1,84 @@
...
+        video.setAttribute("playsinline", "");
+        video.setAttribute("webkit-playsinline", "");
+        // iOS Safari requires the autoplay attribute, or it won't play the video at all.
+        video.autoplay = true;
+        // iOS Safari will not play videos without user interaction. We mute the video so that it can autoplay and then
+        // allow the user to unmute it with an interaction in the unmute-video-button component.
+        video.muted = AFRAME.utils.device.isIOS();
+        video.preload = "auto";
+        video.crossOrigin = "anonymous";
+        video.srcObject = new MediaStream(stream.getVideoTracks());
```
netpro2k, line 57: The stream already should just be video tracks so I think you can skip wrapping it in a new stream. If it did have audio tracks you arguably might want to play those anyway.

src/components/webcam-texture-target.js
```diff
@@ -0,0 +1,84 @@
...
+
+        const texture = new THREE.VideoTexture(video);
+        texture.flipY = false;
```
netpro2k, line 60: We don't set this on other video textures we create from what I can see. Did you run into something that needed this?

src/react-components/room/SharePopoverContainer.js
```diff
@@ -29,15 +33,18 @@ function useShare(scene, hubChannel) {
             .then(devices => {
               const hasCamera = devices.find(device => device.kind === "videoinput");
               setCanShareCamera(hasCamera);
+              setCanShareCameraToAvatar(hasCamera && hasWebcamTextureTarget); // TODO Check avatar model for component
```
netpro2k, line 36: Seems like you are doing this TODO already. Left in by mistake?

netpro2k (dismissed):
Mostly lgtm. Main thing I think we need to resolve is it looks like we are not using the actual material the component is on to decide what material to render the video on. We should either take that into account or just make this a component you add to a mesh instead of a material.

src/components/gltf-model-plus.js
```diff
@@ -233,6 +244,17 @@ async function inflateComponents(inflatedEntity, indexToEntityMap) {
       }
     }
 
+    const materialComponents = getHubsMaterialComponents(object3D);
+
+    if (materialComponents && el) {
+      for (const prop in materialComponents) {
+        if (materialComponents.hasOwnProperty(prop) && AFRAME.GLTFModelPlus.components.hasOwnProperty(prop)) {
+          const { componentName, inflator } = AFRAME.GLTFModelPlus.components[prop];
+          await inflator(el, componentName, materialComponents[prop], materialComponents, indexToEntityMap);
```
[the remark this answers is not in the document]
robertlong: Kinda confused by what you mean here. The components are added to the entity that is the direct parent of the object3D, not the root of the avatar.

It also shouldn't have any issue with the existing material components we have since they are not registered in our gltf component mappings. This new method requires less component specific code in the gltf-model-plus component. Which we may want to use for things like lightmaps etc. going forward.

src/components/webcam-texture-target.js
robertlong: In this design, the component is not responsible for where it is attached. `gltf-model-plus` was modified such that material components are inflated and placed on an entity that is the direct parent of the mesh object.

src/components/webcam-texture-target.js
robertlong: It was upside down :man_shrugging:

src/components/player-info.js
```diff
@@ -37,7 +37,8 @@ AFRAME.registerComponent("player-info", {
   schema: {
     avatarSrc: { type: "string" },
     avatarType: { type: "string", default: AVATAR_TYPES.SKINNABLE },
-    muted: { default: false }
+    muted: { default: false },
+    sharingAvatarCamera: { default: false }
```
johnshaughnessy, line 41: nit: Variable name could signal that it's a boolean (`isSharingAvatarCamera` or `hasAvatarCamera`, etc). I know we don't always do this (the line above has `muted` instead of `isMuted`)

src/components/webcam-texture-target.js
```diff
@@ -0,0 +1,76 @@
...
+
+  init() {
+    const material = this.getMaterial();
+    this.originalTexture = material && material.map;
+  },
+
+  update(prevData) {
+    const material = this.getMaterial();
+
+    console.log({ src: this.data.src, material, prevSrc: prevData.src });
```
johnshaughnessy, line 25: can remove when finished

src/components/webcam-texture-target.js
```diff
@@ -0,0 +1,76 @@
...
+    const src = this.data.src;
+
+    if (prevData.src === src) {
+      return;
+    }
+
+    if (src && src.startsWith("hubs://")) {
+      const streamClientId = src.substring(7).split("/")[1]; // /clients/<client id>/video is only URL for now
+
+      this.originalTexture = material.map;
```
johnshaughnessy, line 40: Why is `originalTexture` reassigned here?

src/react-components/room/useAvatar.js
```diff
@@ -0,0 +1,24 @@
+import { useState, useEffect } from "react";
+
+export default function useAvatar() {
+  const [state, setState] = useState({ hasWebcamTextureTarget: false });
+
+  useEffect(() => {
+    const avatarModelEl = document.querySelector("#avatar-rig .model");
```
johnshaughnessy, line 7: Does this `querySelector` have to run again when someone changes their avatar?

johnshaughnessy (commented):
lgtm - I couldn't tell just by eyeballing it whether it was resilient to having people change their avatars. (I'm guessing that if you change to an avatar that has a `webcam-texture-target`, your client should either keep streaming or re-stream your video and if it does not have a `webcam-texture-target`, streaming should end.)

src/components/webcam-texture-target.js
robertlong: Yeah, I'm going to remove this. The component should have the precondition that the entity has a mesh with a material prior to being added. I've added a warning and another check in the remove function to guard against that.

src/react-components/room/useAvatar.js
robertlong: No it doesn't. The `model` class is on the entity with the `gltf-model-plus` component which is never removed. So this query will only be ran when the React component mounts.

src/components/gltf-model-plus.js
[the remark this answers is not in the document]
robertlong: I'm going to punt on this. We're going to need to revisit this if we want to use @takahirox's GLTFLoader plugin system and I think we can handle it then.

robertlong:
> (I'm guessing that if you change to an avatar that has a webcam-texture-target, your client should either keep streaming or re-stream your video and if it does not have a webcam-texture-target, streaming should end.)

It ends the stream and you need to restart the stream if you change from one avatar to another with the webcam-texture-target component. I think this is acceptable UX and it simplifies the code a little bit.

robertlong:
I've gone over most of the PR feedback. Thank you both for the reviews! @netpro2k and I discussed revisiting material components as a concept when we move to @takahirox's GLTFLoader plugin system. Right now it doesn't make too much sense to inflate a component that is attached to an entity for every usage of a material. Especially if that material is used more than once. We'll probably want a system to handle the single material instance in that case.

Also this component is pretty close to what is needed for streaming to multiple screens in a scene or this Spoke issue https://github.com/mozilla/Spoke/issues/1099

src/components/webcam-texture-target.js
robertlong: Checked with @j-conrad and we think this is correct as is.

robertlong:
Pushing this out so we can test a little easier. Look out for the featured webcam avatars soon :eyes:

Yajo:
How can this feature be tested? 😻

astrojuanlu:
I'd love to give this a spin as well :) Too bad there are no screenshots here!

johnshaughnessy:
@astrojuanlu @Yajo This feature is already live. Here are some example avatars using webcams surfaces:
https://hubs.mozilla.com/avatars/RATQPDA
https://hubs.mozilla.com/avatars/2BtycVC

TheBricktop:
> 
> 
> @astrojuanlu @Yajo This feature is already live. Here are some example avatars using webcams surfaces:
> https://hubs.mozilla.com/avatars/RATQPDA
> https://hubs.mozilla.com/avatars/2BtycVC
Can You share a source of this avatars? (blend maybe?)

hbaileylxd:
@robertlong and @thrbricktop etc...

Hello! This is so great and we want to use it for our Christmas party,  but when i try this my webcam "on" light lights up, but I get a red dot above my camera share button and my friends cannot see my video, even though i have chosen Webcam astronaut or Webcam Hologram. Have checked, and my browser, Chrome, is set to allow the site access to my camera ok. We can talk, just not see video. Any ideas?

PR #3930, opened 2021-02-18
Fix ICE reconnect flashings

keianhzo:
This PR adds support for better ICE failures recovery. Now instead of rejoining in case of an ICE failure we:
- Restart ICE for browser that support it
- Recreate the transports and recreate producers/consumers for browser that don't support ICE `updateIceServers`

This PR adds a couple of methods in the Dialog side for closing a transport and getting the consumers.

This has to be merged at the same time as: https://github.com/mozilla/dialog/pull/12

### DIFF
--- a/src/components/audio-feedback.js
+++ b/src/components/audio-feedback.js
@@ -68,7 +68,7 @@ AFRAME.registerComponent("networked-audio-analyser", {
         this.levels = new Uint8Array(this.analyser.fftSize);
         event.detail.soundSource.connect(this.analyser);
       },
-      { once: false }
+      { once: true }
     );
 
     this.playerSessionId = findAncestorWithComponent(this.el, "player-info").components["player-info"].playerSessionId;
--- a/src/components/audio-feedback.js
+++ b/src/components/audio-feedback.js
@@ -68,7 +68,7 @@ AFRAME.registerComponent("networked-audio-analyser", {
         this.levels = new Uint8Array(this.analyser.fftSize);
         event.detail.soundSource.connect(this.analyser);
       },
-      { once: true }
+      { once: false }
     );
 
     this.playerSessionId = findAncestorWithComponent(this.el, "player-info").components["player-info"].playerSessionId;
--- a/src/components/avatar-audio-source.js
+++ b/src/components/avatar-audio-source.js
@@ -99,17 +99,29 @@ AFRAME.registerComponent("avatar-audio-source", {
   },
 
   async _onStreamUpdated(peerId, kind) {
+    const audio = this.el.getObject3D(this.attrName);
+    if (!audio) return;
+    const stream = audio.source.mediaStream;
+    if (!stream) return;
+
     getOwnerId(this.el).then(async ownerId => {
       if (ownerId === peerId && kind === "audio") {
         // The audio stream for this peer has been updated
-        const stream = await NAF.connection.adapter.getMediaStream(peerId, "audio").catch(e => {
+        const newStream = await NAF.connection.adapter.getMediaStream(peerId, "audio").catch(e => {
           console.error(INFO_INIT_FAILED, `Error getting media stream for ${peerId}`, e);
         });
-        const audioListener = this.el.sceneEl.audioListener;
-        const audio = this.data.positional ? new THREE.PositionalAudio(audioListener) : new THREE.Audio(audioListener);
-        const mediaStreamSource = audio.context.createMediaStreamSource(stream);
-        audio.setNodeSource(mediaStreamSource);
-        this.el.emit("sound-source-set", { soundSource: mediaStreamSource });
+
+        if (newStream) {
+          const oldAudioTrack = stream.getAudioTracks()[0];
+          const newAudioTrack = newStream.getAudioTracks()[0];
+          if (oldAudioTrack.id !== newAudioTrack.id) {
+            oldAudioTrack.stop();
+            stream.removeTrack(oldAudioTrack);
+            if (audio) {
+              stream.addTrack(newAudioTrack);
+            }
+          }
+        }
       }
     });
   },
--- a/src/components/avatar-audio-source.js
+++ b/src/components/avatar-audio-source.js
@@ -15,20 +15,24 @@ function createSilentAudioEl(stream) {
   return audioEl;
 }
 
-async function getMediaStream(el) {
+async function getOwnerId(el) {
   const networkedEl = await NAF.utils.getNetworkedEntity(el).catch(e => {
     console.error(INFO_INIT_FAILED, INFO_NO_NETWORKED_EL, e);
   });
   if (!networkedEl) {
     return null;
   }
-  const ownerId = networkedEl.components.networked.data.owner;
-  if (!ownerId) {
+  return networkedEl.components.networked.data.owner;
+}
+
+async function getMediaStream(el) {
+  const peerId = await getOwnerId(el);
+  if (!peerId) {
     console.error(INFO_INIT_FAILED, INFO_NO_OWNER);
     return null;
   }
-  const stream = await NAF.connection.adapter.getMediaStream(ownerId).catch(e => {
-    console.error(INFO_INIT_FAILED, `Error getting media stream for ${ownerId}`, e);
+  const stream = await NAF.connection.adapter.getMediaStream(peerId).catch(e => {
+    console.error(INFO_INIT_FAILED, `Error getting media stream for ${peerId}`, e);
   });
   if (!stream) {
     return null;
@@ -88,7 +92,26 @@ AFRAME.registerComponent("avatar-audio-source", {
 
   init() {
     this.el.sceneEl.systems["hubs-systems"].audioSettingsSystem.registerAvatarAudioSource(this);
+    // We subscribe to audio stream notifications for this peer to update the audio source
+    // This could happen in case there is an ICE failure that requires a transport recreation.
+    getOwnerId(this.el).then(peerId => {
+      NAF.connection.adapter.addEventListener(peerId, "audio", this);
+    });
     this.createAudio();
+  },
+
+  async onStreamUpdated(peerId, kind) {
+    if (kind === "audio") {
+      // The audio stream for this peer has been updated
+      const stream = await NAF.connection.adapter.getMediaStream(peerId, "audio").catch(e => {
+        console.error(INFO_INIT_FAILED, `Error getting media stream for ${peerId}`, e);
+      });
+      const audioListener = this.el.sceneEl.audioListener;
+      const audio = this.data.positional ? new THREE.PositionalAudio(audioListener) : new THREE.Audio(audioListener);
+      const mediaStreamSource = audio.context.createMediaStreamSource(stream);
+      audio.setNodeSource(mediaStreamSource);
+      this.el.emit("sound-source-set", { soundSource: mediaStreamSource });
+    }
   },
 
   update(oldData) {
@@ -108,6 +131,9 @@ AFRAME.registerComponent("avatar-audio-source", {
 
   remove: function() {
     this.el.sceneEl.systems["hubs-systems"].audioSettingsSystem.unregisterAvatarAudioSource(this);
+    getOwnerId(this.el).then(peerId => {
+      NAF.connection.adapter.removeEventListener(peerId);
+    });
     this.destroyAudio();
   }
 });
--- a/src/components/avatar-audio-source.js
+++ b/src/components/avatar-audio-source.js
@@ -76,10 +76,13 @@ AFRAME.registerComponent("avatar-audio-source", {
       createSilentAudioEl(stream); // TODO: Do the audio els need to get cleaned up?
     }
 
-    const mediaStreamSource = audio.context.createMediaStreamSource(stream);
-    audio.setNodeSource(mediaStreamSource);
+    this.destination = audio.context.createMediaStreamDestination();
+    this.mediaStreamSource = audio.context.createMediaStreamSource(stream);
+    const destinationSource = audio.context.createMediaStreamSource(this.destination.stream);
+    this.mediaStreamSource.connect(this.destination);
+    audio.setNodeSource(destinationSource);
     this.el.setObject3D(this.attrName, audio);
-    this.el.emit("sound-source-set", { soundSource: mediaStreamSource });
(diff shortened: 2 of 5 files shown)


netpro2k (commented):
I am still reviewing the actual ICE related changes in naf-dialog-adapter but wanted to submit the review comments I already have. Mostly just some minor styling feedback on the events system being used.

Starting to look at naf-dialog-adapter more, at first glance it feels like (even prior to this PR) we are doing a lot of work polling the connection state and trying to reconnect things.. I wonder if we are actually hurting ourselves, since I was under the impression that the ICE agent in the browser should already be handling that for us... I want to read through more of the mediasoup-client library to see if its already doing any sort of reconnect logic that we might be getting in the way of with our own reconnect logic.

I also want to read more about TURN and our coturn implementation in particular to understand why our TURN sessions would be expiring. Surely there is some way this is supposed to be getting refreshed... If we fix that then I don't think we have to worry about special handling for Firefox since updateIceServers should never need to be called, meaning we never have to recreate the transports, which would greatly simplify this PR.

src/components/media-views.js
```diff
@@ -736,6 +736,21 @@ AFRAME.registerComponent("media-video", {
       if (url.startsWith("hubs://")) {
         const streamClientId = url.substring(7).split("/")[1]; // /clients/<client id>/video is only URL for now
         const stream = await NAF.connection.adapter.getMediaStream(streamClientId, "video");
+        // We subscribe to video stream notifications for this peer to update the video element
+        // This could happen in case there is an ICE failure that requires a transport recreation.
+        NAF.connection.adapter.addEventListener(streamClientId, "video", {
```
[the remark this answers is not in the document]
keianhzo: Fixed

src/components/media-views.js
```diff
@@ -736,6 +736,21 @@ AFRAME.registerComponent("media-video", {
...
+          onStreamUpdated: async (peerId, kind) => {
+            if (peerId === streamClientId && kind === "video") {
```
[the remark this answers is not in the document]
keianhzo: That's right, we can safely remove that.

src/naf-dialog-adapter.js
```diff
@@ -146,6 +150,24 @@ export default class DialogAdapter {
     this._onOccupantMessage = messageListener;
   }
 
+  // Add an event listener for a peer and an event type
+  addEventListener(peerId, eventType, listener) {
```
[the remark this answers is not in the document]
keianhzo: No reason at all, just what I came up with at the time. Regarding `EventEmitter` I saw it but I couldn't see it being used anywhere else but I'm happy to change to a well known and tested solution. For the scene events I considered it but just prefer each class to be responsible for it's own events rather than throwing everything inside the scene unless it's a global event that's going to be emitted across the board.

src/components/avatar-audio-source.js
```diff
@@ -88,9 +92,28 @@ AFRAME.registerComponent("avatar-audio-source", {
...
+    // We subscribe to audio stream notifications for this peer to update the audio source
+    // This could happen in case there is an ICE failure that requires a transport recreation.
+    getOwnerId(this.el).then(peerId => {
+      NAF.connection.adapter.addEventListener(peerId, "audio", this);
+    });
     this.createAudio();
   },
 
+  async onStreamUpdated(peerId, kind) {
+    if (kind === "audio") {
```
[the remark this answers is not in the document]
keianhzo: Moved to EventEmitter so I have kept that.

src/naf-dialog-adapter.js
```diff
@@ -146,6 +150,24 @@ export default class DialogAdapter {
...
+      this._listeners.set(peerId, {
+        [eventType]: listener
+      });
+    } else {
+      this._listeners.get(peerId)[eventType] = listener;
+    }
+  }
+
+  // Removes an event listener for a peer and an event type
+  removeEventListener(peerId, eventType) {
```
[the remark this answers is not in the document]
keianhzo: I've moved this to eventemitter.

src/naf-dialog-adapter.js
```diff
@@ -202,36 +224,91 @@ export default class DialogAdapter {
    * https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/setConfiguration
```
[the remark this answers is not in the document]
keianhzo: That's what I understand and what I saw during testing. When using TURN we need to update the ICE servers in case the credentials have expired otherwise we are good with just restarting ICE in both sides. As yous y say, if we are not using proxy candidates we can just restart ICE without updating the servers. 

We can also check here if we are we are using a non proxy candidate to avoid an ICE server update as even when forcing TURN we still add a STUN server. I was assuming here that if TURN is forced is because the client is having ICE issues using STUN.

src/components/avatar-audio-source.js
```diff
@@ -88,9 +92,28 @@ AFRAME.registerComponent("avatar-audio-source", {
...
+      // The audio stream for this peer has been updated
+      const stream = await NAF.connection.adapter.getMediaStream(peerId, "audio").catch(e => {
```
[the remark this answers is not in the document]
keianhzo: I agree, replacing tracks seems to be a better option. Updating.

src/components/audio-feedback.js
```diff
@@ -68,7 +68,7 @@ AFRAME.registerComponent("networked-audio-analyser", {
         this.levels = new Uint8Array(this.analyser.fftSize);
         event.detail.soundSource.connect(this.analyser);
       },
-      { once: true }
+      { once: false }
```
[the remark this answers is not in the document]
keianhzo: Replacing tracks avoid dealing with this so reverting.

keianhzo:
@netpro2k I'm happy to pair anytime and go over this and see if we can simplify it.

src/naf-dialog-adapter.js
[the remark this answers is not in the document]
netpro2k: Even when a candidate selected via STUN fails the STUN server itself should not need to change so I think if we can fix the credentials issue we should not ever need to change the ICE servers (except in the case of a dialog server transfer

keianhzo:
@netpro2k I've updated this PR to stop handling the disconnected state.

src/components/avatar-audio-source.js
```diff
@@ -88,9 +95,34 @@ AFRAME.registerComponent("avatar-audio-source", {
...
+    if (!stream) return;
+
+    getOwnerId(this.el).then(async ownerId => {
+      if (ownerId === peerId && kind === "audio") {
+        // The audio stream for this peer has been updated
+        const newStream = await NAF.connection.adapter.getMediaStream(peerId, "audio").catch(e => {
+          console.error(INFO_INIT_FAILED, `Error getting media stream for ${peerId}`, e);
+        });
+
+        if (newStream) {
```
netpro2k, line 117: Nice this feels much simpler than before. Important that we keep this as simple as possible since I could easily see this being a point of failure for "audio issues"

src/components/media-views.js
```diff
@@ -694,6 +694,20 @@ AFRAME.registerComponent("media-video", {
       if (url.startsWith("hubs://")) {
         const streamClientId = url.substring(7).split("/")[1]; // /clients/<client id>/video is only URL for now
         const stream = await NAF.connection.adapter.getMediaStream(streamClientId, "video");
+        // We subscribe to video stream notifications for this peer to update the video element
+        // This could happen in case there is an ICE failure that requires a transport recreation.
+        this._onStreamUpdated = async (peerId, kind) => {
```
netpro2k, line 699: Don't think this will actually happen in practice since I don't think we ever change the `src` for a video running from a MediaStream, but it might be worth checking if `this._onStreamUpdated` is non null here and removing the listener before  registering the new one.

src/naf-dialog-adapter.js
```diff
@@ -192,161 +188,122 @@ export default class DialogAdapter {
...
+    }
+
+    try {
+      if (!this._sendTransport?._closed) {
+        // Renegotiate ICE in case ICE state is "failed".
+        // We also try to renegotiate in case we got stuck in a new state after an ICE failure.
+        if (
+          force ||
+          this._sendTransport.connectionState === "failed" ||
+          (this._sendTransport.connectionState === "new" && this._lastSendConnectionState === "failed")
```
netpro2k, line 224: Not too sure what state this is checking for

src/naf-dialog-adapter.js
```diff
@@ -192,161 +188,122 @@ export default class DialogAdapter {
...
+  async recreateSendTransport(iceServers) {
+    this.emitRTCEvent("log", "RTC", () => `Recreating send transport ICE`);
+    await this.closeSendTransport();
+    await this.createSendTransport(iceServers);
+  }
 
   /**
    * Restart ICE in the underlying send peerconnection.
    */
   async restartSendICE(force = true) {
```
netpro2k, line 211: Feels slightly weird to have this force argument. I think its in part a holdover from when this was being directly used as part of the watchdogs. Instead I would argue this function should always restart ice if called (or recreate the transport if its been closed I guess, still want to understand what that case is). Then we can handle any other logic about when we actually want to restart ice outside of this (which I think we are already doing in `checkXIceStatus`.

src/naf-dialog-adapter.js
```diff
@@ -519,9 +442,55 @@ export default class DialogAdapter {
...
+
+      if (pendingMediaRequests.video) {
+        pendingMediaRequests.video.resolve(null);
+      }
+
+      this._pendingMediaRequests.delete(peerId);
+    }
+
+    // Resolve initial audio resolver since this person left.
+    const initialAudioResolver = this._initialAudioConsumerResolvers.get(peerId);
```
netpro2k, line 475: Not part of this PR but the whole `_initialAudioConsumerResolvers` thing is something I want to look into soon. This stuff still feels a bit fishy to be and would like to avoid it if we can. I would bet there are cases where we end up stalling on this for whatever reason. Its unclear why we need to wait for this.

src/naf-dialog-adapter.js
```diff
@@ -613,141 +582,197 @@ export default class DialogAdapter {
     // Not implemented
   }
 
-  async _joinRoom() {
-    debug("_joinRoom()");
-
-    try {
-      this._mediasoupDevice = new mediasoupClient.Device({});
+  async createSendTransport(iceServers) {
```
netpro2k, line 585: Didn't comb through the changes here very closely since I think this is just moving code around, let me know if that is not the case.

src/naf-dialog-adapter.js
```diff
@@ -613,141 +582,197 @@ export default class DialogAdapter {
...
-      this._recvTransport = this._mediasoupDevice.createRecvTransport({
-        id: recvTransportInfo.id,
-        iceParameters: recvTransportInfo.iceParameters,
-        iceCandidates: recvTransportInfo.iceCandidates,
-        dtlsParameters: recvTransportInfo.dtlsParameters,
-        sctpParameters: recvTransportInfo.sctpParameters,
-        iceServers,
-        iceTransportPolicy: this._iceTransportPolicy
-      });
+  async createRecvTransport(iceServers) {
```
netpro2k, line 689: Don't think we should do it in this PR but just noting for future cleanup. Seems like we could DRY up creating the send/receive transports. A lot of the code in here is the same for both.

netpro2k (commented):
Pretty happy with where we landed on this, nice work! I think the refactoring of naf-dialog-adapter is a nice step forward in starting to make it easier to reason about, and I think removing the watchdogs and handing stream recreation should hopefully lead to more reliability. I made some suggestions on future cleanup work we can do but I think most of it can be done in future PRs and we don't need to hold this back. I am still a bit confused on some of the stuff in `restartSendICE` and would like to discuss that, but other than that lgtm!

src/naf-dialog-adapter.js
keianhzo: That's just a state I've seen a couple of time happening. I can just remove it and keep it simple until we have more insights about WebebRTC errors.

src/naf-dialog-adapter.js
keianhzo: Yep, that argument doesn't make much sense now, I'm going to remove it too.

src/components/media-views.js
keianhzo: Updated

src/naf-dialog-adapter.js
keianhzo: I guess that covers the case where that peer disconnected before it's audio consumer resolved which is something that we do upon connect to wait until we receive all the other peer's consumers. I agree that can cause some issues. I'm not sure why we do that though, is it really that important that we wait until we can hear all the other peers to connect?

keianhzo:
@netpro2k Updated with some of the change requests above. If you are ok, I'll move forward with landing this.

src/components/media-views.js
```diff
@@ -694,6 +694,23 @@ AFRAME.registerComponent("media-video", {
...
+          NAF.connection.adapter.off("stream_updated", this._onStreamUpdated);
+        }
+        this._onStreamUpdated = async (peerId, kind) => {
+          if (peerId === streamClientId && kind === "video") {
+            // The video stream for this peer has been updated
+            const stream = await NAF.connection.adapter.getMediaStream(peerId, "video").catch(e => {
+              console.error(`Error getting video stream for ${peerId}`, e);
+            });
+            if (stream) {
+              videoEl.srcObject = new MediaStream(stream);
```
gfodor, line 709: Hi! I discovered this doesn't work unless you also call `videoEl.play()`, because if the track dies the video is stopped.

keianhzo: Does this happen in all browsers? Can you provide STRs? I can can't really reproduce it with current master in FF/Chrome.

The `stream_update` event in **media-view.js** and **avatar-audio-source.js** should only be as a consequence of a _receive transport recreation_, otherwise both components are not  yet created so the `stream_update` event is not triggered on them. Maybe we are are missing some case where it's being triggered? I  haven't found any but having STRs would really help.

A _receive transport recreation_ should only happen in case where the transport has been closed but signaling is still opened. The transport disconnected/failed states are handled by the WebRTC stack and we should recover without any consumer recreation or transport updates whatsoever (so no `stream_update` events). 

I couldn't also reproduce the issue mentioned above. How are you triggering it?

PR #3932, opened 2021-02-18
Fix bot script for new UI

keianhzo:
Fix bot script for new UI

### DIFF
--- a/src/scene-entry-manager.js
+++ b/src/scene-entry-manager.js
@@ -565,7 +565,7 @@ export default class SceneEntryManager {
 
     const getAudio = () => {
       audioEl.loop = true;
-      audioEl.muted = true;
+      audioEl.muted = false;
       audioEl.crossorigin = "anonymous";
       audioEl.src = URL.createObjectURL(audioInput.files[0]);
       document.body.appendChild(audioEl);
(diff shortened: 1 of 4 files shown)


src/scene-entry-manager.js
```diff
@@ -581,7 +581,7 @@ export default class SceneEntryManager {
 
     const getAudio = () => {
       audioEl.loop = true;
-      audioEl.muted = true;
+      audioEl.muted = false;
```
robinkwilson, line 584: This makes the bots unmuted when they enter yeah? I wonder why that was switched to true.

src/utils/media-devices-manager.js
```diff
@@ -20,7 +20,7 @@ export default class MediaDevicesManager {
     this._videoDevices = [];
     this._deviceId = null;
     this._audioTrack = null;
-    this._mediaStream = null;
+    this._mediaStream = new MediaStream();
```
robinkwilson, line 23: I think this should be initialized as null since folks may not have any media streams available and gets assigned later.

keianhzo: I've fixed in the next commit https://github.com/mozilla/hubs/pull/3932/commits/64bf34decb577291d4741188497e9ca6e14d7be1

src/scene-entry-manager.js
keianhzo: I'm not sure about that, that line goes back quite a long time. I guess maybe before new UI we were un-muting that? The fact is that if it's that without that it doesn't work.

Issue #2140, opened 2020-02-14
When Closing Room, No link back to main Hubs page.


kwadroke:
**Description**
No Link to return to the Hubs main page (or any other page) when you close a room when logged in. You have to hit Back to go to another room or another page.

**To Reproduce**
Create a new room.
Close the room.
Try to click to go somewhere else to connect to another room.

**Expected behavior**
Have a link going to another page like the main Hubs page

**Hardware** Not hardware specific
 - Device: Desktop
 - OS: Linux (but shouldn't matter)
 - Browser: Firefox

This page just needs a Link to go to another page. Else its a dead end.

mrsVelena:
Hi, I'd like to work on this issue.

Softvision-GeluHaiduc:
I have verified this issue using Firefox Release 85.0.2 and Chrome 88.0.4324.182 on Windows 10.
Now, when closing a room there is a button to go back to the homepage.

PR #3936, opened 2021-02-19
Link to return to the Hubs Home page when you close a room

MuditJain5:
#2140  Added Link to return to the Hubs Home page when you close a room 

fix


robertlong:
Oops, looks like I missed this PR. I've addressed this in #3678 Thanks!

PR #3944, opened 2021-02-22
Recreate the mic stream in iOS in case it's muted/unmuted externally

keianhzo:
PR based on this contribution from @vincentfretin https://github.com/mozilla/hubs/issues/2643#issuecomment-765398974

In iOS devices when another app tab uses the mic, the track transitions to a muted state and never recovers: https://bugs.webkit.org/show_bug.cgi?id=213853

This PR listen to mute state changes on the mic track and recreates the track if it transitions unmuted->muted->unmuted state.

This need testing in Safari mobile as I don't have an iOS device and I couldn't reproduce in any other environments.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-819)

### DIFF
--- a/src/utils/media-devices-manager.js
+++ b/src/utils/media-devices-manager.js
@@ -229,6 +229,7 @@ export default class MediaDevicesManager {
   _muteStateChanged() {
     if (this._audioTrack && this._audioTrack.muted) {
       if (!this.reconnectIfUnmuted) {
+        this.wasMicEnabledBeforeMute = NAF.connection.adapter.micEnabled;
         NAF.connection.adapter.enableMicrophone(false);
       }
       this.reconnectIfUnmuted = true;
@@ -236,7 +237,9 @@ export default class MediaDevicesManager {
       if (this.reconnectIfUnmuted) {
         this.reconnectIfUnmuted = false;
         this.setMediaStreamToDefault();
-        NAF.connection.adapter.enableMicrophone(true);
+        if (this.wasMicEnabledBeforeMute) {
+          NAF.connection.adapter.enableMicrophone(true);
+        }
       }
     }
   }
--- a/src/utils/media-devices-manager.js
+++ b/src/utils/media-devices-manager.js
@@ -1,6 +1,7 @@
 const isMobile = AFRAME.utils.device.isMobile();
 const isMobileVR = AFRAME.utils.device.isMobileVR();
 const isFirefoxReality = isMobileVR && navigator.userAgent.match(/Firefox/);
+import { detectOS, detect } from "detect-browser";
 
 // This is a list of regexes that match the microphone labels of HMDs.
 //
@@ -11,6 +12,10 @@ const isFirefoxReality = isMobileVR && navigator.userAgent.match(/Firefox/);
 // Note that this doesn't have to be exhaustive: if no devices match any regex
 // then we rely upon the user to select the proper mic.
 const HMD_MIC_REGEXES = [/\Wvive\W/i, /\Wrift\W/i];
+
+const IOS_MUTED_HACK_MIN_VERSION = 13;
+const detectedOS = detectOS(navigator.userAgent);
+const browser = detect();
 
 export default class MediaDevicesManager {
   constructor(scene, store, audioSystem) {
@@ -172,6 +177,8 @@ export default class MediaDevicesManager {
         this._scene.emit("action_end_mic_sharing");
       });
 
+      this._iOSTrackMutedHack();
+
       if (/Oculus/.test(navigator.userAgent)) {
         // HACK Oculus Browser 6 seems to randomly end the microphone audio stream. This re-creates it.
         // Note the ended event will only fire if some external event ends the stream, not if we call stop().
@@ -183,7 +190,9 @@ export default class MediaDevicesManager {
           const newStream = await navigator.mediaDevices.getUserMedia(constraints);
           this.audioTrack = newStream.getAudioTracks()[0];
 
-          this.audioSystem.addStreamToOutboundAudio("microphone", newStream);
+          this._iOSTrackMutedHack();
+
+          audioSystem.addStreamToOutboundAudio("microphone", newStream);
 
           this._scene.emit("local-media-stream-created");
 
@@ -199,6 +208,36 @@ export default class MediaDevicesManager {
       console.error("Error during getUserMedia: ", e);
       this.audioTrack = null;
       return false;
+    }
+  }
+
+  /**
+   * Webkit issue on iOS>13: https://bugs.webkit.org/show_bug.cgi?id=213853
+   */
+  _iOSTrackMutedHack() {
+    const iOSVersion = parseInt(browser.version.split(".")[0]);
+    if (
+      ["iOS", "Mac OS"].includes(detectedOS) &&
+      ["safari", "ios"].includes(browser.name) &&
+      iOSVersion >= IOS_MUTED_HACK_MIN_VERSION
+    ) {
+      this.audioTrack.onmute = this._muteStateChanged;
+      this.audioTrack.onunmute = this._muteStateChanged;
+    }
+  }
+
+  _muteStateChanged() {
+    if (this._audioTrack && this._audioTrack.muted) {
+      if (!this.reconnectIfUnmuted) {
+        NAF.connection.adapter.enableMicrophone(false);
+      }
+      this.reconnectIfUnmuted = true;
+    } else {
+      if (this.reconnectIfUnmuted) {
+        this.reconnectIfUnmuted = false;
+        this.setMediaStreamToDefault();
+        NAF.connection.adapter.enableMicrophone(true);
+      }
     }
   }
 
--- a/src/utils/media-devices-manager.js
+++ b/src/utils/media-devices-manager.js
@@ -192,7 +192,7 @@ export default class MediaDevicesManager {
 
           this._iOSTrackMutedHack();
 
-          audioSystem.addStreamToOutboundAudio("microphone", newStream);
+          this.audioSystem.addStreamToOutboundAudio("microphone", newStream);
 
           this._scene.emit("local-media-stream-created");
 
--- a/src/naf-dialog-adapter.js
+++ b/src/naf-dialog-adapter.js
@@ -1044,6 +1044,10 @@ export default class DialogAdapter extends EventEmitter {
     });
   }
 
+  get micEnabled() {
+    return this._micEnabled;
+  }
+
   setWebRtcOptions() {
     // Not implemented
   }


keianhzo:
@vincentfretin this is ready based on your contribution in case you want to help testing it.

src/utils/media-devices-manager.js
```diff
@@ -191,6 +200,36 @@ export default class MediaDevicesManager {
...
+    ) {
+      this.audioTrack.onmute = this._muteStateChanged;
+      this.audioTrack.onunmute = this._muteStateChanged;
+    }
+  }
+
+  _muteStateChanged() {
+    if (this._audioTrack && this._audioTrack.muted) {
+      if (!this.reconnectIfUnmuted) {
+        NAF.connection.adapter.enableMicrophone(false);
```
vincentfretin, line 221: You need to update the react ui as well to indicate the mic is muted and even disabled until it recreate the audio track.

src/utils/media-devices-manager.js
```diff
@@ -191,6 +200,36 @@ export default class MediaDevicesManager {
...
+      }
+      this.reconnectIfUnmuted = true;
+    } else {
+      if (this.reconnectIfUnmuted) {
+        this.reconnectIfUnmuted = false;
+        this.setMediaStreamToDefault();
+        NAF.connection.adapter.enableMicrophone(true);
+      }
```
vincentfretin, line 229: Instead of 
```
this.setMediaStreamToDefault();
NAF.connection.adapter.enableMicrophone(true);
```
you should call 
```
this.recreateAudioStream()
```
so it calls getUserMedia again with the previous deviceId and reexecute `_iOSTrackMutedHack()`
(`setMediaStreamToDefault()` is not calling again `_iOSTrackMutedHack()` here)
and be sure it calls `enableMicrophone(currentMicStatus)`.
If the user was muted, then answered a Skype call, you don't want to enable the mic after he finished his Skype call.

You probably need to merge #3543 first to expose recreateAudioStream on `this`.
You may want to really have and exact match (not the default "ideal" match) in recreateAudioStream too, like I did in my PR https://github.com/mozilla/hubs/pull/3543 see the comments in the code.

vincentfretin (changes requested):
Thanks for doing it. I can try to test it this weekend on my iPad. I need someone to call me on my Skype though to reproduce it, I'll ask a friend. :) Not sure how to test the local hubs on my ubuntu from the iPad though, maybe with ngrok?

src/utils/media-devices-manager.js
keianhzo: That call will update the UI too by a store property update.

src/utils/media-devices-manager.js
keianhzo: That's what `setMediaStreamToDefault` does using the last used mic device, it ends up calling `fetchAudioTrack` which recreates the track and calls `_iOSTrackMutedHack`.

keianhzo: I'm going to update to handle the case when the mic was muted before the mute state change so we don't unmute.

keianhzo:
Thanks for that @vincentfretin! I'm not sure about the testing in a physical phone though, I have managed to successfully launch hubs in Android//iOS using the emulators but I'm not sure about a physical device.

src/utils/media-devices-manager.js
vincentfretin: Ah this is fine then, I'm not familiar with the new code. The enableMicrophone from the janus adapter doesn't do such thing, so I wasn't expecting it to trigger an UI update here.

src/utils/media-devices-manager.js
vincentfretin: You're absolutely right, it calls `fetchAudioTrack`, all good.

vincentfretin:
Sorry I didn't take the time last weekend to test it. I was busy with networked-aframe maintenance. :)
If any of you have a real iPhone or iPad, don't hesitate to test this too!

netpro2k (commented):
Hard to tell by just this diff but will this tirgger a recreation of audio stream even if the user manually clicks mute and then unmute? Might be unavoidable since even if we tried to detect the user manually muting its possible they would trigger this bug while muted and then would need the stream to be recreated on manually unmuting anyway... This lgtm and we will keep an eye on the upstream bug.

vincentfretin:
You mean if the user muted/unmuted from the Safari tab or address bar? Yes it will recreate it.
If you mean the Hubs mic button, no because this doesn't touch the audio track muted property, it touches the enabled property.

vincentfretin:
I did some tests.
PC and iPad are the same network at home.

On PC Ubuntu
connected with Skype test account

Don't add hubs.local to /etc/hosts
I changed my hostname to hubs and restarted avahi
```
sudo hostnamectl set-hostname hubs
sudo systemctl restart avahi-daemon.service
```

```
git checkout master
git pull
npm install
npm run dev
```

Open Chrome on
https://hubs.local:8080
create a room

on iPad 14.3
connect with Skype normal account
open Safari on https://hubs.local:8080 and connect to the room

With master
First scenario (refuse call):
- if PC Skype calls iPad, while it's ringing, Safari mic is muted, if I refuse the call the mic is unmuted (the mic indicator in Safari address bar) but mic volume doesn't work. Also I don't hear others (probably the AudioContext suspended state here, see the other issue #3969).

Second scenario (accept call):
- if PC Skype calls iPad, while it's ringing, Safari mic is muted, if I accept the call and then end it, the mic in Safari address bar is kept muted, no possibility to unmute from there, need to refresh. Also I don't hear others.

Now with ios-safari-mute branch (`git checkout ios-safari-mute && npm run dev`)
First scenario (refuse call):
- if PC Skype calls iPad, while it's ringing, Safari mic is muted, if I refuse the call the mic is unmuted (the mic indicator in Safari address bar) and now the mic volume works. But sometimes I still don't hear others (probably the AudioContext suspended state here).
But sometimes I have this:
- if PC Skype calls iPad, while it's ringing, Safari mic is muted, if I refuse the call the mic is kept muted, the hubs scene is completely frozen (because of an error?)

Second scenario (accept call):
I had several different behavior:
- if PC Skype calls iPad, while it's ringing, Safari mic is muted, if I accept the call and then end it, the mic in Safari address bar is back to unmuted, all is working fine.
- if PC Skype calls iPad, while it's ringing, Safari mic is muted, if I accept the call and then end it, the mic in Safari address bar is back to unmuted but the hubs scene completely frozen, probably an error? (I don't have macOS to connect the device to look at console errors)
- (same as master) if PC Skype calls iPad, while it's ringing, Safari mic is muted, if I accept the call and then end it, the mic in Safari address bar is kept muted, no possibility to unmute from there, need to refresh. Also I don't hear others (probably the AudioContext suspended state here).

So the changes seems to fix the issue sometimes but apparently sometimes hubs completely freeze and you need to refresh. I don't know why. I think someone else with macOS and iPad/iPhone need to test this and look at Safari logs.

vincentfretin:
The hubs scene freeze after rejecting the Skype call is not related to those changes because I just had the same issue with the safari-mobile-audio-context-fix branch. And same with master and accepting the call, after ending the call, hubs froze (not always).

I just updated to iOS 14.4.1, the issue still apply.
So the changes may be good but it's completely useless with the freeze issue.

src/utils/media-devices-manager.js
```diff
@@ -191,6 +200,39 @@ export default class MediaDevicesManager {
...
+      this.audioTrack.onmute = this._muteStateChanged;
+      this.audioTrack.onunmute = this._muteStateChanged;
+    }
+  }
+
+  _muteStateChanged() {
+    if (this._audioTrack && this._audioTrack.muted) {
+      if (!this.reconnectIfUnmuted) {
+        this.wasMicEnabledBeforeMute = NAF.connection.adapter.micEnabled;
+        NAF.connection.adapter.enableMicrophone(false);
```
vincentfretin, line 222: FYI `NAF.connection.adapter.enableMicrophone(false)` indeed triggers the UI update when I run it in the console. But in the case of receiving a Skype call, it doesn't update the UI. I think React doesn't update the UI when the browser is not in foreground (when you have the Skype notification).

src/utils/media-devices-manager.js
```diff
@@ -191,6 +200,39 @@ export default class MediaDevicesManager {
...
+      }
+      this.reconnectIfUnmuted = true;
+    } else {
+      if (this.reconnectIfUnmuted) {
+        this.reconnectIfUnmuted = false;
+        this.setMediaStreamToDefault();
```
vincentfretin, line 228: Just to be sure, you probably want 
`await this.setMediaStreamToDefault();` here and add `async` to `_muteStateChanged`

vincentfretin:
Ah actually after the update to 14.4.1, when hubs doesn't freeze after rejecting a call, the mic is unmuted in the address bar, but the mic volume doesn't work, I can hear others. So the fix doesn't seem to work anymore. :(

vincentfretin:
I have the same behavior on my app (not hubs based, and not using react for the UI), I had the scene and UI freeze one time and had to refresh, another time the hack successfully worked, another time it didn't work and I needed to refresh, another time I really had to restart Safari for the mic to work again. :(

keianhzo:
I can't still test this but I've rebased to keep it aligned with upstream.

vincentfretin:
I may test that again after #3969 is merged and this PR rebased again.
For the logs, I think I can integrate http://www.vorlonjs.io in the html to be able to access the logs.

keianhzo:
Great, thanks @vincentfretin I'll rebase after merging #3969 and ping you so you can test again.

keianhzo:
@vincentfretin I've rebased with latest master including https://github.com/mozilla/hubs/pull/3969

vincentfretin:
iOS 14.4.2
```
git fetch
git reset --hard origin/ios-safari-mute
git checkout package-lock.json
npm ci
npm run dev
```

Sometimes the Safari tab is frozen after hanging up the Skype call. I got the same on master so it is not related to the changes in this PR. This may be a Safari issue on iOS >= 14.4.1.

Note that on the branch and current master, it asks me for the mic but after that it says Microphone Disabled, so the mic is not working anymore at all.

vincentfretin:
If you want to get the console logs with vorlon:
```
npm i vorlon
```
edit node_modules/vorlon/Server/config.json to set
```
    "useSSL": true,  
```
run vorlon:
```
./node_modules/vorlon/bin/vorlon
```

open the port on your firewall (Ubuntu):
```
sudo ufw allow 1337
```

edit src/hub.html to add in head :
```
<script src="https://192.168.1.14:1337/vorlon.js"></script>
```
(modify with your machine ip)

on iOS go to https://192.168.1.14:1337 (hubs.local doesn't work, it reload infinitly) and accept the certificate

go to the room
https://hubs.local:8080/hub.html?hub_id=RoOmId

on your machine, consult the logs at https://192.168.1.14:1337

For the mic issue:
Capture d’écran de 2021-04-17 16-13-22

vincentfretin:
I have no js error when the tab freeze. I got in the logs:
```
AudioContext state changed to interrupted
AudioContext state changed to running
AudioContext state changed to interrupted
AudioContext state changed to running
AudioContext state changed to interrupted
```

So yeah Safari...

vincentfretin:
For the mic disabled, maybe #4151 fixes it? I use iPad internal mic but maybe the device id changed between iOS 14.4.1 and 14.4.2 so this why I have the OverconstrainedError.

takahirox:
I couldn't track the status. And new iOS versions have been released. Is this change still needed?

vincentfretin:
I tested again with a room at hubs.mozilla.com, Skype, Safari on iPadOS 15.5. I don't have the freeze anymore after I refused a Skype call. I had sometimes the issue of mic not working (mic volume indicator not showing at all, others don't hear me) after I refused or accepted a Skype call. To fix it I needed to mute and unmute the mic from the tab, then it worked again.
I suppose you can close this issue if we have this working workaround.

netpro2k:
Seems like the conclusion in the discussion above was that we don't need this workaround. Closing for now as less hacks == better.

PR #3951, opened 2021-02-24
Fix personal space bubble opacity

robertlong:
This PR fixes a bug where the personal space bubble would set the opacity/transparent property of an already transparent object when leaving it's bounds. It also improves the space bubble so that avatars with multiple meshes will turn transparent when entering their bounds. To do this without turning the ui transparent, I added a new tag `ignoreSpaceBubble` that can be applied to the root of objects that should not be affected by the space bubble.

### DIFF
--- a/src/components/tags.js
+++ b/src/components/tags.js
@@ -20,7 +20,7 @@ AFRAME.registerComponent("tags", {
     isStatic: { default: false },
     inspectable: { default: false },
     preventAudioBoost: { default: false },
-    spaceBubbleIgnore: { default: false }
+    ignoreSpaceBubble: { default: false }
   },
   update() {
     if (this.didUpdateOnce) {
--- a/src/components/tags.js
+++ b/src/components/tags.js
@@ -19,7 +19,8 @@ AFRAME.registerComponent("tags", {
     isHoverMenuChild: { default: false },
     isStatic: { default: false },
     inspectable: { default: false },
-    preventAudioBoost: { default: false }
+    preventAudioBoost: { default: false },
+    spaceBubbleIgnore: { default: false }
   },
   update() {
     if (this.didUpdateOnce) {
--- a/src/hub.html
+++ b/src/hub.html
@@ -137,7 +137,7 @@
                         </template>
 
                         <template data-name="Neck">
-                            <a-entity>
+                            <a-entity tags="spaceBubbleIgnore: true;">
                                 <a-entity
                                     billboard
                                     class="nametag"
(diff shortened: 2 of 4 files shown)


src/hub.html
```diff
@@ -173,7 +173,7 @@
                         </template>
 
                         <template data-name="Spine">
-                            <a-entity is-remote-hover-target personal-space-invader="radius: 0.2; useMaterial: true;" bone-visibility>
+                            <a-entity tags="ignoreSpaceBubble: true;" is-remote-hover-target personal-space-invader="radius: 0.2; useMaterial: true;" bone-visibility>
                                 <a-entity class="avatar-inspect-collider" avatar-inspect-collider tags="togglesHoveredActionSet: true;" is-remote-hover-target inspect-button></a-entity>
                                 <a-entity class="chest-image" scale="0.18 0.18 0.18" position="0 -0.025 0.13"></a-entity>
                                 <a-entity class="interactable ui interactable-ui">
```
netpro2k, line 179: Should `ignoreSpaceBubble` be on this instead of Spine? I imagine we would want to fade the `chest-image` in personal space.

src/systems/personal-space-bubble.js
```diff
@@ -234,10 +234,36 @@ AFRAME.registerComponent("personal-space-invader", {
...
+              opacity: material.opacity,
+              transparent: material.transparent
+            };
+          }
+
+          // Note: Sharing materials will cause all objects with the material to turn transparent / opaque
+          // This is generally fine for avatars since a material will only be shared within a glTF model,
+          // not across avatars in the room.
+          material.opacity = invading ? this.data.invadingOpacity : originalProps.opacity;
+          material.transparent = invading ? true : originalProps.transparent;
```
netpro2k, line 265: This could be written as `material.transparent = invading || originalProps.transparent`. It's a tossup which one actually reads more clearly, but a ternary with `true` in it stood out so figured I would mention it.

src/components/tags.js
```diff
@@ -19,7 +19,8 @@ AFRAME.registerComponent("tags", {
     isHoverMenuChild: { default: false },
     isStatic: { default: false },
     inspectable: { default: false },
-    preventAudioBoost: { default: false }
+    preventAudioBoost: { default: false },
+    ignoreSpaceBubble: { default: false }
```
netpro2k, line 23: Probably warants thinking through the product implications, so maybe we don't need it in this PR, but wonder if we should expose this to gltfs like

```js
AFRAME.GLTFModelPlus.registerComponent("ignore-space-bubble", "tags", el => {
  el.setAttribute("tags", { ignoreSpaceBubble: true });
});
```

So people could make things in their avatars that ignore the space bubble effect.

netpro2k (approved):
lgtm. This has been an issue for far too long, thanks for getting a fix in.

src/components/tags.js
robertlong: I think this could be abused and we probably want to keep it internal. Happy to have a separate discussion about this at a later point though.

Issue #3743, opened 2021-01-21
Audio files can still be heard from the tab after permanently closing a room


Softvision-GeluHaiduc:
**[Affected Versions]:**
- Firefox Release 84.0.2

**[Affected Platforms]:**
- Windows 10x64

**[Prerequisites]:**
- Have a new room created and entered.

**[Steps to reproduce]:**
1. Drag and drop a music file in the room or upload the file.
2. Close the room.
3. Observe if the audio file can still be heard.

**[Expected results]:**
- The audio file can no longer be heard after the room is closed.

**[Actual results]:**
- The audio file continues to be heard.

**[Notes]:**
- The issue is reproducible Hubs Cloud as well.
- The audio stops after the page is refreshed.
- Attached a screen recording of the issue below
![audio](https://user-images.githubusercontent.com/51400741/105354680-9170b880-5bf9-11eb-9c8e-9adb5d0b36e2.gif)

johnshaughnessy:
This is important but should not block release.

PR #3958, opened 2021-02-25
Fix closed room cleanup

robertlong:
This PR address multiple places where we are relying on invalid state when exiting the room and removing the scene.

The main culprit is using `AFRAME.scenes[0]` which will be null after it is removed. The pen and camera system also fail to properly deregister objects because dependent elements such as the avatar rig are already removed. Finally, Networked Aframe will not have a network adapter set when joining an already closed room, so I added a guard for that.

Fixes https://github.com/mozilla/hubs/issues/3743

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-818)

### DIFF
--- a/src/components/tools/networked-drawing.js
+++ b/src/components/tools/networked-drawing.js
@@ -111,7 +111,7 @@ AFRAME.registerComponent("networked-drawing", {
   },
 
   play() {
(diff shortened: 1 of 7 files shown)


netpro2k:
Things have diverged since this was done. Most of the changes are related to global AFRAME state and we are actively working to move off of that anyway so its likely we will fix the root issues indirectly anyway.

PR #3965, opened 2021-02-25
Simulcast support

keianhzo:
Related https://github.com/mozilla/hubs/pull/3543

This PR ad adds support for simulcast and refactors some media manager related code required for supporting simulcast. It adds three layers for webcam sharing and two for screen sharing. It also exposes layers and downlink info in the RTC panel.

It also addresses the lack of support for handling sharing permissions at the browser level. Now when stopping sharing mic/video from the browser site permissions we should remove the related media and stop producing for those sources.

### DIFF
--- a/src/naf-dialog-adapter.js
+++ b/src/naf-dialog-adapter.js
@@ -33,6 +33,15 @@ const PC_PROPRIETARY_CONSTRAINTS = {
 
 const INITIAL_ROOM_RECONNECTION_INTERVAL = 2000;
 
+const WEBCAM_SIMULCAST_ENCODINGS = [
+  { scaleResolutionDownBy: 4, maxBitrate: 500000 },
+  { scaleResolutionDownBy: 2, maxBitrate: 1000000 },
+  { scaleResolutionDownBy: 1, maxBitrate: 5000000 }
+];
+
+// Used for simulcast screen sharing.
+const SCREEN_SHARING_SIMULCAST_ENCODINGS = [{ dtx: true, maxBitrate: 1500000 }, { dtx: true, maxBitrate: 6000000 }];
+
 export default class DialogAdapter extends EventEmitter {
   constructor() {
     super();
@@ -40,7 +49,8 @@ export default class DialogAdapter extends EventEmitter {
     this._timeOffsets = [];
     this._occupants = {};
     this._micProducer = null;
-    this._videoProducer = null;
+    this._cameraProducer = null;
+    this._shareProducer = null;
     this._mediaStreams = {};
     this._localMediaStream = null;
     this._consumers = new Map();
@@ -64,6 +74,16 @@ export default class DialogAdapter extends EventEmitter {
     this._lastRecvConnectionState = null;
     this._closed = true;
     this.scene = document.querySelector("a-scene");
+    this._downlinkBwe = null;
+    this._consumerStats = {};
+  }
+
+  get consumerStats() {
+    return this._consumerStats;
+  }
+
+  get downlinkBwe() {
+    return this._downlinkBwe;
   }
 
   get serverUrl() {
@@ -347,6 +367,16 @@ export default class DialogAdapter extends EventEmitter {
               this.removeConsumer(consumer.id);
             });
 
+            if (kind === "video") {
+              const { spatialLayers, temporalLayers } = mediasoupClient.parseScalabilityMode(
+                consumer.rtpParameters.encodings[0].scalabilityMode
+              );
+
+              this._consumerStats[consumer.id] = this._consumerStats[consumer.id] || {};
+              this._consumerStats[consumer.id]["spatialLayers"] = spatialLayers;
+              this._consumerStats[consumer.id]["temporalLayers"] = temporalLayers;
+            }
+
             // We are ready. Answer the protoo request so the server will
             // resume this Consumer (which was paused for now if video).
             accept();
@@ -398,7 +428,10 @@ export default class DialogAdapter extends EventEmitter {
           const { consumerId } = notification.data;
           const consumer = this._consumers.get(consumerId);
 
-          if (!consumer) break;
+          if (!consumer) {
+            info(`consumerClosed event received without related consumer: ${consumerId}`);
+            break;
+          }
 
           consumer.close();
           this.removeConsumer(consumer.id);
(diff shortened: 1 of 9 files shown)


src/scene-entry-manager.js
```diff
@@ -397,52 +397,30 @@ export default class SceneEntryManager {
...
     };
 
-    this.scene.addEventListener("action_share_camera", event => {
+    this.scene.addEventListener("action_share_camera", () => {
+      if (isHandlingVideoShare) return;
+      isHandlingVideoShare = true;
+
       const constraints = {
         video: {
           width: isIOS ? { max: 1280 } : { max: 1280, ideal: 720 },
```
netpro2k, line 426: We may want to revisit these. A bit afraid of encoding multiple high resolution stream on a phone for example.

src/naf-dialog-adapter.js
```diff
@@ -33,14 +33,24 @@ const PC_PROPRIETARY_CONSTRAINTS = {
...
 const INITIAL_ROOM_RECONNECTION_INTERVAL = 2000;
 
+const WEBCAM_SIMULCAST_ENCODINGS = [
+  { scaleResolutionDownBy: 4, maxBitrate: 500000 },
+  { scaleResolutionDownBy: 2, maxBitrate: 1000000 },
+  { scaleResolutionDownBy: 1, maxBitrate: 5000000 }
+];
+
+// Used for simulcast screen sharing.
+const SCREEN_SHARING_SIMULCAST_ENCODINGS = [{ dtx: true, maxBitrate: 1500000 }, { dtx: true, maxBitrate: 6000000 }];
```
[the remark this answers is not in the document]
netpro2k: Also same question here. Is there a reason you chose these bitrates? And no resolution scaling here just bitrates?

src/naf-dialog-adapter.js
```diff
@@ -33,14 +33,24 @@ const PC_PROPRIETARY_CONSTRAINTS = {
...
+
 const isFirefox = navigator.userAgent.toLowerCase().indexOf("firefox") > -1;
 
 export default class DialogAdapter {
   constructor() {
     this._timeOffsets = [];
     this._occupants = {};
     this._micProducer = null;
-    this._videoProducer = null;
+    this._cameraProducer = null;
```
netpro2k, line 52: Is there a reason this was split up? Seems like we (currently) are only allowing producing one or the other. Might simplify the code a bit to keep it as `_vidoeProducer` so you don't have to check both in multiple places.

src/utils/media-devices-manager.js
```diff
@@ -191,34 +202,68 @@ export default class MediaDevicesManager {
...
+          // Ideally we would use track.contentHint but it seems to be read-only in Chrome so we just add a custom property
+          track["_hubs_contentHint"] = isDisplayMedia ? "share" : "camera";
+          track.addEventListener("ended", () => {
+            this._scene.emit("action_end_video_sharing");
+          });
+          this._mediaStream.addTrack(track);
+        });
+
+        if (newStream && newStream.getAudioTracks().length > 0) {
+          this.audioSystem.addStreamToOutboundAudio("screenshare", newStream);
```
netpro2k, line 242: Not related to this PR exactly, but now that I understand MediaSoup better I think we should eventually rework the stuff being done here to mix in screen share audio into the users output audio stream. Its fairly easy to send this as just another audio producer. This gets rid of some of the hoop jumping we are doing here to mix it, and probides a better experience anyway since the desktop audio can actually come from the screen share instead of the avatar. I have some ideas of adding a "group" concept to setLocalMediaStream/getMediaStream. We should discuss it.

netpro2k (commented):
This looks good to me. I would like to understand better the values we are picking and know a bit more about the performance impact on the sending system before we push this, but I think it should definitely help improve the video experience for people, so I am excited to get this out.

src/naf-dialog-adapter.js
```diff
@@ -33,14 +33,24 @@ const PC_PROPRIETARY_CONSTRAINTS = {
 const ICE_RECONNECT_INTERVAL = 2000;
 const INITIAL_ROOM_RECONNECTION_INTERVAL = 2000;
 
+const WEBCAM_SIMULCAST_ENCODINGS = [
+  { scaleResolutionDownBy: 4, maxBitrate: 500000 },
```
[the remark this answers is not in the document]
keianhzo: Not really, I just followed the examples and was hoping to see how was that working for everybody to adjust.

src/naf-dialog-adapter.js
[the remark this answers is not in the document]
keianhzo: Same as above.

src/naf-dialog-adapter.js
```diff
@@ -513,6 +543,40 @@ export default class DialogAdapter {
...
+          this._downlinkBwe = notification.data;
+          break;
+        }
+
+        case "consumerLayersChanged": {
+          const { consumerId, spatialLayer, temporalLayer } = notification.data;
+
+          const consumer = this._consumers.get(consumerId);
+
+          if (!consumer) break;
```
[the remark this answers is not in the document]
keianhzo: Fixed.

src/naf-dialog-adapter.js
keianhzo: I guess now that we are going to add support for webcam avatars it will make sense to separate this?

src/utils/media-devices-manager.js
keianhzo: Screen share audio only works for Chrome right now and only for tab sharing for this might be a limited feature but we can definitely do it if we are willing to.

src/scene-entry-manager.js
keianhzo: I think the next step towards better video sharing in general is to expose options to do HD, SD,... so the user can choose that or even switch between them real time. We can adjust the different setups when implementing that. I'll open an issue for that.

src/naf-dialog-adapter.js
```diff
@@ -68,6 +78,16 @@ export default class DialogAdapter {
     this._recvTaskId = null;
     this._closed = true;
     this.scene = document.querySelector("a-scene");
+    this._downlinkBwe = null;
+    this._consumerStats = {};
```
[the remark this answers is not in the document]
keianhzo: yep, so we can see the download link and spatial/temporal layers in the RTC panel.

src/naf-dialog-adapter.js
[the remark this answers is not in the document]
netpro2k: Ok well that's what we did with the rest of Dialog :) .. Seems like a reasonable starting point.

src/naf-dialog-adapter.js
netpro2k: yeah, this is another thing the "groups" concept I talked about will be helpful for. No strong preference on if we keep the 2 producer variables separate or combine for this PR.

src/utils/media-devices-manager.js
netpro2k: Ah, what I mean is we area already doing this in Chrome, we just mix it into your outgoing audio stream. Once we rework naf-dialog-adapter to support multiple outgoing streams (we likely will do this first for video so that you can use a webcam and screen share at the same time) then we can get rid of this code that mixes your desktop audio with your mic audio and instead send it on its own producer.

keianhzo: Oh yep, get it. Yep makes sense that we keep them separated so we can manipulate them independently on the client side although screen share audio seems to be quite limited right now.

src/naf-dialog-adapter.js
keianhzo: ok, let's keep it like it is and address groups when we implement multiple producers support.

src/scene-entry-manager.js
keianhzo: https://github.com/mozilla/hubs/issues/3989

PR #3970, opened 2021-02-26
2d webcams

robinkwilson:
DO NOT MERGE, work in progress

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-817)


netpro2k:
This was a hackweek experiment and we do not intend to merge it as is. It is likely we will eventually want some form of this, so the branch will remain useful as reference.

PR #4001, opened 2021-03-09
Fix property typo in media-frames.js

speigg:
Fixed property typo. `oldData.tergetId` changed to `oldData.targetId`


netpro2k (approved):
Nice catch! Will get this out tomorrow.

netpro2k:
Better late than never... Sorry it took so long, thanks for the PR!

PR #4045, opened 2021-03-18
Frontend Dev Best Practices Guide

robertlong:
This guide highlights some best practices we follow on the Hubs team and better reflects the Hubs codebase after the redesign.


johnshaughnessy (approved):
Love it, thanks for putting this together.

PR #4047, opened 2021-03-19
Initial support for KHR_texture_basisu

netpro2k:
We will continue to support `MOZ_HUBS_texture_basisu` for now with a deprecation warning since it is being used in the wild, though we mostly have a good line of communication with those that are using it so we can probably work pretty quickly to get them off of it.

Relies on https://github.com/MozillaReality/three.js/pull/51

TODO
- [x] release threejs update and update package.lock to point at the new commit


src/components/gltf-model-plus.js
```diff
@@ -354,7 +366,17 @@ export async function loadGLTF(src, contentType, onProgress, jsonPreprocessor) {
   const loadingManager = new THREE.LoadingManager();
   loadingManager.setURLModifier(getCustomGLTFParserURLResolver(gltfUrl));
   const gltfLoader = new THREE.GLTFLoader(loadingManager);
```
netpro2k, line 368: I don't think we are doing a great job of cleaning up after ourselves as far as the GLTF loader is concerned. I think the only reason we don't just share a single instance is because of the LoadingManager, and the only reason that needs to be unique if I remember correctly is because we don't have a good way of dealing with relative urls. If I am not mistaken this has gotten better with the latest GLTFLoader code with the concept of "handlers", so I would like to revisit this soon as part of updating ThreeJS. I bet @takahirox will be able to help with this when he is back later this month. Since this PR doesn't makes things any worse (we area already likely leaking the Basis transcoders), this seems fine for now. But I just want to note it so we remember to follow up on it.

takahirox: > this has gotten better with the latest GLTFLoader code with the concept of "handlers"

Probably the easiest way is to use `beforeRoot` hook, which is called before starting the parse of entire glTF, and override the resource urls.

```
class GLTFURLModifier {
  constructor(parser) {
    this.parser = parser;
  }
  beforeRoot() {
    const json = this.parser.json;
    const options = this.parser.options;
    // traverse json and override resource URLs
  }
}

const loader = new GLTFLoader();
loader.register(parser => new GLTFURLModifier(parser));
loader.load( ... );
```

takahirox (approved):
I haven't run the test yet but the code looks to me.

src/components/gltf-model-plus.js
netpro2k: Yep that seems like a decent option as well. What's nice about using the loading manager is it theoretically is agnostic of plugins we might add later, since anything could point at external resources not just textures. In practice it might now actually matter much and we can just iterate the few things we know to use external resources like images and buffers.

takahirox: Yes, ideally using loading manager is better. It's simpler, straightforward, and can be used for future extensions without changes.

But unfortunately loading manager doesn't take care of root asset URL (in this case glTF url) then we need to create loading manager per glTF url, as we do now, and it leads to create loader instance per glTF url. I suggested `beforeRoot` hook if we want to reuse the loader instance by using glTF loader plugin system.

PR #4064, opened 2021-03-19
Fix issue that breaks the UI when you sign in after attempting to pin an object

### LINKED ISSUE #3822: The UI is broken if the user signs in while the object menu is displayed
**[Affected Versions]:**
- Firefox Release 85
- Chrome 87.0.4280.88

**[Affected Platforms]:**
- Windows 10x64

**[Prerequisites]:**
- Have a new room created and entered.
- Make sure no user is signed in.
- Make sure to have at least one item created.

**[Steps to reproduce]:**
1. Click the “Objects” button from the top-right of the screen.
2. Click on one of the objects from the list.
3. Click the “Pin” button and sign into an account.
4. Finish the sign-in process and observe the UI.

**[Expected results]:**
- After finishing the sign-in process the object can be pinned.

**[Actual results]:**
- After finishing the sign-in process the buttons disappear and nothing can be clicked.

**[Notes]:**
- The issue is not reproducible if the user signs in from the “More” menu. 
- Attached a screen recording of the issue below
sign in

brianpeiris:
We seem to have removed the concept of "sign-in complete message" with the redesign, but missed that it was still being passed around as a parameter. This happened to work in most cases, since most calls to `performConditionalSignIn` only passed in three parameters, except for the code path triggered by pinning objects, with used the `onFailure` parameter with a callback. https://github.com/mozilla/hubs/blob/master/src/scene-entry-manager.js#L251-L262

This PR removes the `signInCompleteMessage` parameter entirely since it's not actually in use.

I've also fixed null reference errors related to scene info UI, which I ran into while testing this on a fresh dev environment, where we load create rooms without a scene, and fallback to the loadingEnvironment.glb instead.

Fixes #3822

### DIFF
--- a/src/hub.js
+++ b/src/hub.js
@@ -788,7 +788,7 @@ document.addEventListener("DOMContentLoaded", async () => {
   window.APP.mediaDevicesManager = new MediaDevicesManager(scene, store);
   window.APP.hubChannel = hubChannel;
 
-  const performConditionalSignIn = async (predicate, action, signInMessage, signInCompleteMessage, onFailure) => {
+  const performConditionalSignIn = async (predicate, action, signInMessage, onFailure) => {
     if (predicate()) return action();
 
     await handleExitTo2DInterstitial(true, () => remountUI({ showSignInDialog: false }));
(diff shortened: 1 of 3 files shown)


robertlong (approved):
LGTM :+1:

PR #4030, opened 2021-03-16
Hubs Cloud Theme Support

robertlong:

PR #4109, opened 2021-04-01
Add themes customization to admin panel

johnshaughnessy:
Allows pasting an array of themes (like this one: https://github.com/mozilla/hubs/blob/feature/themes-in-admin-panel/themes.json)  into the admin console to specify custom colors for the new UI. 

Has a backend change to correctly handle the JSON. https://github.com/mozilla/reticulum/pull/488


src/react-components/styles/theme.js
```diff
@@ -34,6 +38,9 @@ export function useTheme(themeId) {
 
   useEffect(
     () => {
+      // Make sure themes is properly configured. (It must be an array.)
+      if (!themes || !themes.find || !themes.__proto__ !== Array.prototype) return;
```
robinkwilson, line 42: There is an `Array.isArray()` function in JS to do this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray

`if (!themes || !Array.isArray(themes)) return;`

admin/src/utils/ita.js
```diff
@@ -149,7 +152,9 @@ const DESCRIPTOR_FIELDS = [
   "name",
   "description",
   "internal",
-  "source"
+  "source",
+  "deprecated",
```
robinkwilson, line 156: Where is "deprecated" being used now?

robinkwilson (commented):
A few questions and a few nits!

admin/src/utils/ita.js
```diff
@@ -40,6 +41,8 @@ function getCategoryDisplayName(category) {
       return "Images";
     case "theme":
       return "Theme";
+    case "themes":
+      return "Themes";
```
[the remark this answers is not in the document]
robinkwilson: I might have called this `themes_json` because it seems to do the same thing as `theme` just in a different format.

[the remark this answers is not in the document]
johnshaughnessy: `"theme"` is how the cloud admins used to setup custom colors in the old UI. That's going to be deprecated when I merge https://github.com/mozilla/hubs/pull/4030 

I did not remove `"theme"` for two reasons: 
1) I think it might still be used for determining in-world menu/sprite colors. (I need to check this).
2) Users might not have a copy of their colors elsewhere. If we remove the "theme" menu, they might lose this data. By keeping it in, they have an easier migration path for writing the json blob they need to migrate their old colors to the new UI.

As you said -- `themes` is a list of available themes to select from as you said. Selecting one of these themes puts your preference into local storage but does not affect the `theme` config.

admin/src/utils/ita.js
johnshaughnessy: This is added in the underlying branch. Parsing the config schema fails if this keyword is not added. Here's a comment in context: https://github.com/mozilla/hubs/pull/4030#pullrequestreview-625953237

johnshaughnessy: Nothing is actually using the "deprecated" flag yet though. I need to check if in-world menu/sprite colors are still using the old theme configs. If so, they're still important and should not be marked "deprecated" in a visible way until we change that.

robinkwilson (approved):
Lgtm!

admin/src/utils/ita.js
johnshaughnessy: Update - I now hide the "deprecated" configs unless the `show_deprecated_configs` query string parameter is present.

admin/src/utils/ita.js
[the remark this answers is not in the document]
johnshaughnessy: Update: I updated the in-world menu/sprite colors so that they use the new `themes` config values.

admin/src/react-components/service-editor.js
```diff
@@ -377,23 +410,29 @@ class ConfigurationEditor extends Component {
...
-              [this.props.classes.success]: !this.state.saveError,
-              [this.props.classes.warning]: !!this.state.saveError
+              [this.props.classes.success]: !this.state.saveError && !this.state.warningMessage,
+              [this.props.classes.warning]: !!this.state.saveError || !!this.state.warningMessage
             })}
             message={
               <span id="import-snackbar" className={this.props.classes.message}>
                 <Icon className={clsx(this.props.classes.icon, this.props.classes.iconVariant)} />
-                {this.state.saveError || "Settings saved."}
+                {this.state.saveError || this.state.warningMessage || (this.state.saved && "Settings saved.") || ""}
```
robinkwilson, line 427: Does it need the extra `|| ""`? Does this stay popped up every time you type?

admin/src/react-components/service-editor.js
```diff
@@ -224,14 +225,45 @@ class ConfigurationEditor extends Component {
...
+
     return (
       <TextField
         key={displayPath}
         id={displayPath}
         label={descriptor.name || displayPath}
         inputProps={{ maxLength: 4096 }}
         value={currentValue || ""}
-        onChange={ev => this.onChange(path, ev.target.value)}
+        onChange={ev => {
```
robinkwilson, line 246: Maybe not for this PR, but I dont know if I like that the popup will keep popping up every time you type within an invalid json object. 

How does it feel when you type in the field? -- if it feels fine, then this works.

Maybe having the textfield highlight in red with the initial pop up of invalid json. Which the red will go away afterward. You can do this with a `style={isValidJson ? {} : {border: "2px solid red"}}` (or `styles` I've forgotten). Disregard if you think the "feeling"  of typing in invalid json is fine.

admin/src/react-components/service-editor.js
```diff
@@ -224,14 +225,45 @@ class ConfigurationEditor extends Component {
...
+          if (descriptor.json) {
+            if (!isValidJSON(ev.target.value)) {
+              this.setState({
+                warningMessage: `Invalid JSON for ${descriptor.name || displayPath}. See console for details.`
+              });
+              console.error(`Invalid JSON for ${descriptor.name || displayPath}.`);
+              console.error(ev.target.value);
+            } else {
+              this.setState({ warningMessage: null });
```
robinkwilson, line 255: Can you add a check: `if (this.state.warningMessage !== null) this.setState({ warningMessage: null });` so we don't hammer this.setState unnecessarily.

admin/src/react-components/service-editor.js
robinkwilson: You could also check if the warningMessage is already set with the matching descriptor name/path before setting this.

admin/src/react-components/service-editor.js
```diff
@@ -224,14 +225,45 @@ class ConfigurationEditor extends Component {
...
+            }
+          }
+          this.onChange(path, ev.target.value);
+        }}
+        onBlur={ev => {
+          if (descriptor.json && isValidJSON(ev.target.value)) {
+            // Pretty print json strings
+            const pretty = JSON.stringify(JSON.parse(ev.target.value), null, 2);
+            this.onChange(path, pretty);
```
robinkwilson, line 264: nice.

admin/src/react-components/service-editor.js
```diff
@@ -377,23 +410,29 @@ class ConfigurationEditor extends Component {
...
-                {this.state.saveError || "Settings saved."}
+                {this.state.saveError || this.state.warningMessage || (this.state.saved && "Settings saved.") || ""}
               </span>
             }
             action={[
-              <IconButton key="close" color="inherit" onClick={() => this.setState({ saved: false })}>
+              <IconButton
+                key="close"
+                color="inherit"
+                onClick={() => this.setState({ saved: false, warningMessage: null })}
```
robinkwilson, line 434: I dont know if you want to reset the warning message, since the warning message still is valid. I could be wrong.

src/react-components/styles/theme.js
```diff
@@ -9,9 +9,17 @@ if (!config && process.env.STORYBOOK_APP_CONFIG) {
   config = JSON.parse(process.env.STORYBOOK_APP_CONFIG);
 }
 
+if (!config) {
+  config = window.APP_CONFIG;
+}
+
+if (config && config.theme && config.theme.failed_to_load) {
```
robinkwilson, line 16: Could be written:

```suggestion
if (config?.theme?.failed_to_load) {
```

src/react-components/styles/theme.js
```diff
@@ -9,9 +9,17 @@ if (!config && process.env.STORYBOOK_APP_CONFIG) {
...
+  console.error("Custom themes failed to load. If you are an admin, reconfigure your themes in the admin panel.");
+}
+
 export const defaultTheme = "default";
 
-export const themes = (config && config.themes) || [];
+export const themes = (config && config.theme && config.theme.themes) || [];
```
robinkwilson, line 22:
```suggestion
export const themes = (config?.theme?.themes) || [];
```

If it fails on any of those, it will return `undefined`

src/utils/theme.js
```diff
@@ -18,19 +18,22 @@ const DEFAULT_COLORS = {
...
-  if (configs.APP_CONFIG && configs.APP_CONFIG.theme && configs.APP_CONFIG.theme[name])
-    return configs.APP_CONFIG.theme[name];
+  const themeId = window.APP?.store?.state?.preferences?.theme;
+
+  const theme =
+    themeId &&
+    configs.APP_CONFIG &&
+    configs.APP_CONFIG.theme &&
+    configs.APP_CONFIG.theme.themes &&
+    configs.APP_CONFIG.theme.themes.find(theme => theme.id === themeId);
```
robinkwilson, line 28:
```suggestion
    configs?.APP_CONFIG?.theme?.themes?.find(theme => theme.id === themeId);
```

then remove lines 25 -> 27 definitely my favorite recent JS upgrade :D

src/utils/theme.js
```diff
@@ -18,19 +18,22 @@ const DEFAULT_COLORS = {
...
+  if (theme && theme.variables && theme.variables[name]) {
```
robinkwilson, line 29: Same here.

admin/src/react-components/service-editor.js
johnshaughnessy: After I clear the `warningMessage`, there's a brief animation that clears the toast. But then as that thing is rendering, it changes color (because `warningMessage` is no longer set). I don't care so much that the color changes -- but I don't want people to accidentally read "settings saved" when they have not been saved. This can probably be handled more cleanly but I wanted to rewrite as little as possible about how this worked.

admin/src/react-components/service-editor.js
johnshaughnessy: I need to clear the `warningMessage` here because the user is explicitly closing the toast and the existence of the `warningMessage` is how it decides that it should be `open`. 

A related concern is that I don't actually prevent the user from saving invalid JSON here. The server will fail to parse it in that case and the themes won't be available to the client. I'm hoping that with clear instructions, people won't have too much trouble with this (and we can revisit the UI if they do).

src/schema.toml
```diff
@@ -59,17 +59,18 @@ images.landing_rooms_thumb = { category = "images", type = "file", internal = "t
...
+theme.action-label-color = { category = "theme", type = "color", name = "Primary Label Color", description = "[deprecated] Call to action label color.", deprecated="true" }
+theme.notice-background-color = { category = "theme", type = "color", name = "Secondary Color", description = "[deprecated] Notice and indicator color.", deprecated="true" }
+theme.notice-text-color = { category = "theme", type = "color", name = "Secondary Text Color", description = "[deprecated] Notice and indicator text color.", deprecated="true" }
+theme.notice-widget-color = { category = "theme", type = "color", name = "Secondary Widget Color", description = "[deprecated] Notice and indicator UI widget color.", deprecated="true" }
+theme.favorited-color = { category = "theme", type = "color", name = "Favorited Color", description = "[deprecated] Favorited Color", deprecated = "true" }
+theme.themes = { category = "theme", type = "longstring", name = "Themes", description = "A json array describing custom client themes.", json="true"}
```
brianpeiris, line 73: nit: I wonder if we should be using `type="json"` here instead of `json="true"`.

PR #4112, opened 2021-04-03
Move AudioNormalization configuration from chat command to preference

### LINKED ISSUE #3589: Add dynamic audio normalization option under preference

Currently Dynamic audio normalization feature is behind the chat command `/audioNormalization <factor>`.  The reason why it isn't under the preference or isn't enabled by default was I didn't have confident of the quality of the normalization.

But we did the test in the team and I got the positive feedbacks. And I resolved the background noise problem #3252 found in the test. So I'd like to move this feature next step.

I'd like to move the normalization feature option from the chat command to the preference, and I'd like more users to test it and give us the feedbacks. So "Experiment feature. Give us the feedback (link to GitHub)" label next to the option would be good.

I'd like to enable this feature by default at some point, but first I'd like to test more before doing that. Because I'd like to make this feature more stable and if this feature works incorrectly it may provide very bad experience to users, for example it misestimates the volume and makes the volume too loud.

takahirox:
Resolved #3589

### DIFF
--- a/src/react-components/preferences-screen.js
+++ b/src/react-components/preferences-screen.js
@@ -389,6 +389,10 @@ const preferenceLabels = defineMessages({
     id: "preferences-screen.preference.disable-auto-gain-control",
     defaultMessage: "Disable microphone automatic gain control"
   },
+  audioNormalization: {
+    id: "preferences-screen.preference.audio-normalization",
+    defaultMessage: "Dynamic audio normalization factor (0 is disabled, 4 is recommend)"
+  },
   snapRotationDegrees: {
     id: "preferences-screen.preference.snap-rotation-degrees",
     defaultMessage: "Rotation per snap (in degrees)"
@@ -955,6 +959,15 @@ class PreferencesScreen extends Component {
             prefType: PREFERENCE_LIST_ITEM_TYPE.CHECK_BOX,
             defaultBool: false,
             promptForRefresh: true
+          },
+          {
+            key: "audioNormalization",
+            prefType: PREFERENCE_LIST_ITEM_TYPE.NUMBER_WITH_RANGE,
+            min: 0,
+            max: 10,
+            step: 1,
+            digits: 1,
+            defaultNumber: 0
           }
         ]
       ],
--- a/src/storage/store.js
+++ b/src/storage/store.js
@@ -96,7 +96,6 @@ export const SCHEMA = {
         preferredCamera: { type: "string" },
         muteMicOnEntry: { type: "bool" },
         audioOutputMode: { type: "string" },
-        audioNormalization: { type: "bool" },
         invertTouchscreenCameraMove: { type: "bool" },
         enableOnScreenJoystickLeft: { type: "bool" },
         enableOnScreenJoystickRight: { type: "bool" },
@@ -124,6 +123,7 @@ export const SCHEMA = {
         disableEchoCancellation: { type: "bool" },
         disableNoiseSuppression: { type: "bool" },
         disableAutoGainControl: { type: "bool" },
+        audioNormalization: { type: "number" },
         locale: { type: "string" },
         showRtcDebugPanel: { type: "bool" }
       }
--- a/src/message-dispatch.js
+++ b/src/message-dispatch.js
@@ -160,28 +160,6 @@ export default class MessageDispatch extends EventTarget {
           );
         }
         break;
-      case "audioNormalization":
-        {
-          if (args.length === 1) {
-            const factor = Number(args[0]);
-            if (!isNaN(factor)) {
-              const effectiveFactor = Math.max(0.0, Math.min(255.0, factor));
-              window.APP.store.update({
-                preferences: { audioNormalization: effectiveFactor }
-              });
-              if (factor) {
-                this.log(LogMessageType.setAudioNormalizationFactor, { factor: effectiveFactor });
-              } else {
-                this.log(LogMessageType.audioNormalizationDisabled);
-              }
-            } else {
-              this.log(LogMessageType.audioNormalizationNaN);
-            }
-          } else {
-            this.log(LogMessageType.invalidAudioNormalizationRange);
-          }
-        }
-        break;
     }
   };
 }
--- a/src/react-components/room/ChatSidebar.js
+++ b/src/react-components/room/ChatSidebar.js
@@ -81,11 +81,7 @@ export const LogMessageType = {
   captureAlreadyStopped: "captureAlreadyStopped",
   captureAlreadyRunning: "captureAlreadyRunning",
   positionalAudioEnabled: "positionalAudioEnabled",
-  positionalAudioDisabled: "positionalAudioDisabled",
-  setAudioNormalizationFactor: "setAudioNormalizationFactor",
-  audioNormalizationDisabled: "audioNormalizationDisabled",
-  audioNormalizationNaN: "audioNormalizationNaN",
-  invalidAudioNormalizationRange: "invalidAudioNormalizationRange"
+  positionalAudioDisabled: "positionalAudioDisabled"
 };
 
 const logMessages = defineMessages({
@@ -140,23 +136,6 @@ const logMessages = defineMessages({
   [LogMessageType.positionalAudioDisabled]: {
     id: "chat-sidebar.log-message.positional-audio-disabled",
     defaultMessage: "Positional audio disabled."
-  },
-  [LogMessageType.setAudioNormalizationFactor]: {
-    id: "chat-sidebar.log-message.set-audio-normalization-factor",
-    defaultMessage: "audioNormalization factor is set to {factor}."
-  },
-  [LogMessageType.audioNormalizationDisabled]: {
-    id: "chat-sidebar.log-message.audio-normalization-disabled",
-    defaultMessage: "audioNormalization is disabled."
-  },
-  [LogMessageType.audioNormalizationNaN]: {
-    id: "chat-sidebar.log-message.audio-normalization-nan",
-    defaultMessage: "audioNormalization command needs a valid number parameter."
-  },
-  [LogMessageType.invalidAudioNormalizationRange]: {
-    id: "chat-sidebar.log-message.invalid-audio-normalization-range",
-    defaultMessage:
-      "audioNormalization command needs a base volume number between 0 [no normalization] and 255. Default is 0. The recommended value is 4, if you would like to enable normalization."
   }
 });
 
--- a/src/systems/audio-settings-system.js
+++ b/src/systems/audio-settings-system.js
@@ -49,12 +49,6 @@ export class AudioSettingsSystem {
         preferences: { audioOutputMode: "panner" }
       });
     }
-    if (window.APP.store.state.preferences.audioNormalization !== 0.0) {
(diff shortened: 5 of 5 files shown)


src/react-components/preferences-screen.js
```diff
@@ -389,6 +389,10 @@ const preferenceLabels = defineMessages({
     id: "preferences-screen.preference.disable-auto-gain-control",
     defaultMessage: "Disable microphone automatic gain control"
   },
+  audioNormalization: {
+    id: "preferences-screen.preference.audio-normalization",
+    defaultMessage: "Dynamic audio normalization factor (0 is disabled, 4 is recommend)"
```
takahirox, line 394: Let me know if you have an idea to simplify the message.

image

johnshaughnessy: I agree that people will not know what this means. Adding additional help information (like you suggested in Discord) is a good idea. 

Maybe this should be exposed as a boolean property - "ON" means "4" and "OFF" means "0". I know that 4 is not perfect for all scenarios, but if it is _usually_ better than 0, then maybe we want to set it to 4 by default and later give people the option to either disable it or set a custom factor.

johnshaughnessy (approved):

src/react-components/preferences-screen.js
takahirox: Yeah, I have been thinking of boolean, too, because it's clear to users. But also I still want factor configuration for the test. So what do you think of add normalization configuration to preference as boolean and also keep it in chat command which accept factor 0 to x?

johnshaughnessy: It's up to you. We tested the audio normalization factors a couple months ago and I could not tell the difference. How do you intend to figure out what the best default value is (surveys, analytics, user studies)? If you just want this code to be used and let people on the team tell you over time what their preferred number is, maybe just start with what you have (a slider) and change it to a boolean if you ask in 6 months and everyone likes the same value.

takahirox: Dynamic audio normalization attempts to make master volume into a certain range specified with audio normalization factor. The biggest concern about boolean (fix factor number) is the appropriate factor should very depend on users' environment. For example there could be a chance that 4 could be too quiet for speaker users but too loud for ear pad users. So I'm leaning toward a slider first and then change it to boolean if we will get surveys or analytics and realize most of users like the same factor...

The factor number may not be straight forward to users, I would add a help icon which pops dialog explaining what it is.

netpro2k:
I was going to suggest we call this "automatic gain control" to align with other apps, but actually it looks like we are already using AGC (and even expose a preference to disable it). I wonder if something is not working with our AGC, because shouldn't that solve the same problem? 

https://developer.mozilla.org/en-US/docs/Web/API/MediaTrackSettings/autoGainControl

takahirox:
I first expected `MediaTrackSettings.autoGainControl` could solve the audio gain lavel balance problem but even though probably AGC may be enabled by default on most input devices and most users don't disable it via preference I still need set individual volumes.

According to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/MediaTrackSettings/autoGainControl) AGC seems typically being used on microphones (input devices). I guess input device AGC is for controlling its output signal gain level into a certain range but its actual output gain range very depends on the device, so the gain level ranges of voices coming from the remote peers still varies depending on their input devices. (This is my guess. Please correct me if I'm wrong.)

Hubs allows us to control individual volume offsets for users. But even if they are saved, manually setting them every time a new person joins a room may be annoying. So I still think automatic volume control is useful.

netpro2k:
Not sure about Zoom but Discord I think is just very directly using AGC from libwebrtc and does not do server side mixing. It does allow you to set individual's volume offsets (and saves them), and I have to do this for some people. 

What discord also allows you to do that Hubs does not have is set your "input volume" and they have a nice mic test that shows you (and lets you hear) your own mic levels. This might help users set thing up appropriately for their setup. Though not sure if most users actually bother. 

image

In my experience on Discord this problem is usually solved socially, by someone joining the channel and then people saying "You are very quiet" or "you are super loud", they might fiddle with their audio settings for a bit (both at the Discord level and the OS level), which might fix it. If they don't people usually say "whatever it's fine I just adjusted your volume". To me this seems like a perfectly reasonable solution, but I perhaps have a higher tolerance for this sort of thing than most people and mostly deal with more technically advanced people, so maybe most people find this an unacceptable solution.

To be clear, not against audio normalization per se, just feel like we should get the basics right before reaching for more automation. But again I am weird so maybe most people would significantly benefit from an option like this.

Also based on this discussion, maybe the name "Automatic Output Gain Control" or "Automatic Output Volume Control" would be a good name for this preference.

takahirox:
> just feel like we should get the basics right before reaching for more automation.

I agree with that. But I rarely set individual volumes on Zoom or Discord while I need to often do on Hubs. Fernando also said so, too, so probably this is not only my personal opinion. I think there is a room for audio experience improvement by automatically adjust.

I guess they apply some effects for the better user audio experience or we may need some extra effects because our audio system is positional based. I would investigate how they mix the audio well.

> maybe the name "Automatic Output Gain Control" or "Automatic Output Volume Control" would be a good name for this preference.

Yeah, I'm thinking the same thing, too.

src/react-components/preferences-screen.js
takahirox: Sorry for changing my opinion again but I changed the Automatic Output Gain Control from a slider to a checkbox because it will be much straightforward to users. I renamed Audio Normalization to Audio Output Gain Control and I speculate users can understand what it is from the name. I started to think that making the preference layout complex by adding an additional dialog just for explaining the factor would be bad.

src/react-components/preferences-screen.js
```diff
@@ -484,6 +490,8 @@ const preferenceLabels = defineMessages({
   }
 });
 
+preferenceLabels.automaticOutputGainControl.helpLink = "https://github.com/mozilla/hubs/issues/3214";
```
takahirox, line 493: @brianpeiris As I explained in Discord, I couldn't come up with other ways to extend `preferenceLabels`. Do you know more appropriate way?

src/react-components/preferences-screen.js
```diff
@@ -509,8 +517,20 @@ class PreferenceListItem extends Component {
...
+      <span className={styles.preferenceLabel}>
+        {intl.formatMessage(preferenceLabels[this.props.storeKey])}
+        {preferenceLabels[this.props.storeKey].helpLink !== undefined ? (
+          <IconButton onClick={onHelpIconClick}>
+            <HelpIcon />
+          </IconButton>
+        ) : (
+          <></>
+        )}
+      </span>
```
takahirox, line 533: image

I wanted to put the help icon in the same line of labels but I couldn't avoid the new line. I guess `intl.formatMessage()` adds a new line? Is it possible to remove the new line?

src/react-components/preferences-screen.js
brianpeiris: Ah, I guess `defineMessages` removes extraneous properties. We could consider generalizing it a bit more if we ended up using this pattern more often, but this works for now.
One thing we should consider though is how this would work on Hubs Cloud. If we want Hubs Cloud instances to be completely white-label, it might not make sense for a help link to point to our github repo.
I'd also suggest, to make it more user-friendly, that we create and point to a page under https://hubs.mozilla.com/docs/, but maybe we can do that once we consider this a permanent feature and not an experiment.

src/react-components/preferences-screen.js
brianpeiris: I would suggest something this, where you use React's `as` property to turn this element into a link, instead of using javascript to open a new window, and also a class to style the link properly so that it appears on the same line:

```javascript
<IconButton
  className={styles.helpButton}
  as="a"
  href={preferenceLabels[this.props.storeKey].helpLink}
  target="_blank"
>
```

```scss
/* preferences-screen.scss */
:local(.help-button) {
  display: inline;

  svg {
    display: inline;
    vertical-align: text-bottom;
    margin-left: 5px;
  }
}
```

src/react-components/preferences-screen.js
```diff
@@ -390,6 +392,10 @@ const preferenceLabels = defineMessages({
     id: "preferences-screen.preference.disable-auto-gain-control",
     defaultMessage: "Disable microphone automatic gain control"
   },
+  automaticOutputGainControl: {
+    id: "preferences-screen.preference.audio-automatic-output-gain-control",
+    defaultMessage: "[Experiment] Automatic Output Gain Control"
```
brianpeiris, line 397: Suggestion for wording and capitalization to match the existing preferences:
```suggestion
    defaultMessage: "Automatic incoming voice volume control (experimental)"
```

brianpeiris (changes requested):
Noticed one blocker issue and left some other suggestions.

src/react-components/preferences-screen.js
brianpeiris: For Hubs Cloud, one option might be to use the existing "show_docs_link" feature flag, so if this preference does land on Hubs Cloud, at least users will have some ability to disable it. We could also add a new "auto_gain_control_help_link" config so that Hubs Cloud users can decide if they want to change this URL.

I also wonder what benefit you want users to get from linking to that particular issue. Maybe it should link to a new issue or a new discussion with less technical detail about the feature?

src/storage/store.js
```diff
@@ -96,7 +96,6 @@ export const SCHEMA = {
         preferredCamera: { type: "string" },
         muteMicOnEntry: { type: "bool" },
         audioOutputMode: { type: "string" },
-        audioNormalization: { type: "bool" },
```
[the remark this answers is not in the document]
takahirox: Thanks. I didn't know it blocks. I should have been more careful to add a preference.

src/react-components/preferences-screen.js
takahirox: Looking good to me. Thanks.

src/react-components/preferences-screen.js
takahirox: Updated, thanks!

src/react-components/preferences-screen.js
takahirox: > One thing we should consider though is how this would work on Hubs Cloud. If we want Hubs Cloud instances to be completely white-label, it might not make sense for a help link to point to our github repo.

Good point. I overlooked it.

> For Hubs Cloud, one option might be to use the existing "show_docs_link" feature flag, so if this preference does land on Hubs Cloud, at least users will have some ability to disable it. We could also add a new "auto_gain_control_help_link" config so that Hubs Cloud users can decide if they want to change this URL.

Maybe it would be a good option. But I don't want to make the configuration more complex for this new feature. So I'm thinking of removing the linking.

> I also wonder what benefit you want users to get from linking to that particular issue. Maybe it should link to a new issue or a new discussion with less technical detail about the feature?

I wanted users to know what Dynamic Audio Normalization is from the linking. But it has been renamed to Automatic incoming voice volume control so perhaps they can imagine what it is from the name now. And I want them to send the feedback for this new feature's quality but I'm thinking of an alternate way of the linking for example writing a blog post instead.

takahirox:
Let me switch this PR to draft so far because at least I couldn't come up with any solutions to use case that users intentionally speak quietly.

PR #4118, opened 2021-04-06
Do not wait to resolve consumers to complete the room connection flow

keianhzo:
We have talked about this in the past when going over the Dialog adapter code. We are currently waiting for all the consumers to be resolved to complete the connection flow. I don't find any reason why we want to do this instead of letting them resolve lazily. Unless there is a strong reason why, I think it might be a good idea to remove this gate to avoid potential deadlocks when waiting to resolve consumers that will never resolve for whatever reason.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-813)


keianhzo:
Yep, this has already been addressed. Closing.

PR #4126, opened 2021-04-07
Add support for rendering in world cameras to textures

netpro2k:
This PR builds on the `video-texture-target` component used for video avatars by adding a new `video-texture-source` component. This component is to be placed on a `Camera` entity. The `video-texture-target` can then specify that entity as its `target` to pull "video" from that camera onto its texture. This allows creating things like "jumbotron" screens and mirrors. This should be considered "beta" for now and may change depending on what we find trying it out in some scenes.

image

Blender exporter support for these new components is here: https://github.com/MozillaReality/hubs-blender-exporter/pull/28

Still a few TODOs in the code that can be dealt with later. Ideally we would also rework the camera-tool component/system to use this new component to do its actual rendering to remove any duplicate code.

### DIFF
--- a/src/components/video-texture-target.js
+++ b/src/components/video-texture-target.js
@@ -1,18 +1,86 @@
 import { disposeTexture } from "../utils/material-utils";
 import { createVideoOrAudioEl } from "../utils/media-utils";
+import { findNode } from "../utils/three-utils";
+
+/**
+ * @component video-texture-source
+ * This component is intended to be used on entities with a Camera Object3D as a child.
+ * That camera is used to render the scene to a WebGLRenderTarget of the specified resolution
+ * at a maximum of the specified frame rate. It will only render a frame if something sets
+ * it's textureNeedsUpdate property to true. Currently video-texture-target does this
+ * whenever its material is used during the primary camera render (as in, its in view).
+ */
+AFRAME.registerComponent("video-texture-source", {
+  schema: {
+    resolution: { type: "vec2", default: [1280, 720] },
+    fps: { default: 15 }
+  },
+
+  init() {
+    this.camera = findNode(this.el.object3D, n => n.isCamera);
+    this.camera.aspect = this.data.resolution[0] / this.data.resolution[1];
+
+    // TODO currently if a video-texture-source tries to render itself it will fail with a warning.
+    // If we want to support this we will need 2 render targets to swap back and forth.
+    this.renderTarget = new THREE.WebGLRenderTarget(this.data.resolution[0], this.data.resolution[1], {
+      format: THREE.RGBAFormat,
+      minFilter: THREE.LinearFilter,
+      magFilter: THREE.NearestFilter,
+      encoding: THREE.GammaEncoding,
+      depth: false,
+      stencil: false
+    });
+
+    const texture = this.renderTarget.texture;
+    texture.matrixAutoUpdate = false;
+    texture.matrix.scale(1, -1);
+    texture.matrix.translate(0, 1);
+
+    this.textureNeedsUpdate = false;
+  },
+
+  tock(time) {
+    if (!this.textureNeedsUpdate) return;
+
+    if (time < this.lastRenderTime + 1000 / this.data.fps) return;
+
+    // TODO handle hiding UI and showing first person head
+    // Once thats done also rework camrea-tool to use this to do its actual rendering
+
+    const sceneEl = this.el.sceneEl;
+    const renderer = this.renderer || sceneEl.renderer;
+
+    const tmpVRFlag = renderer.vr.enabled;
+    const tmpOnAfterRender = sceneEl.object3D.onAfterRender;
+    delete sceneEl.object3D.onAfterRender;
+    renderer.vr.enabled = false;
+
+    renderer.setRenderTarget(this.renderTarget);
+    renderer.render(sceneEl.object3D, this.camera);
+    renderer.setRenderTarget(null);
+
+    renderer.vr.enabled = tmpVRFlag;
+    sceneEl.object3D.onAfterRender = tmpOnAfterRender;
+
+    this.lastRenderTime = time;
+    this.textureNeedsUpdate = false;
+  }
+});
 
 /**
  * @component video-texture-target
  * This component is intended to be used on entities with a mesh/skinned mesh Object3D
  * The component swaps the base color map on the mesh's material with a video texture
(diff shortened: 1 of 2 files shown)


src/components/video-texture-target.js
```diff
@@ -64,38 +132,56 @@ AFRAME.registerComponent("video-texture-target", {
...
+        this.applyTexture(texture);
+
+        // Bit of a hack here to only update the renderTarget when the screens are in view
+        material.map.isVideoTexture = true;
+        material.map.update = () => {
+          videoTextureSource.textureNeedsUpdate = true;
+        };
+      } else {
+        if (material.map && material.map !== this.originalMap) {
+          disposeTexture(material.map);
```
johnshaughnessy, line 150: If I'm reading this right, this will `dispose` a `video-texture-source`'s `renderTarget.texture`. It seems like that should be the responsibility of the `video-texture-source`. 

Related, if you have two jumbotron `video-texture-targets` with the same `video-texture-source`, and then you delete or change the `src / target` of one of the jumbotrons, will the second jumbotron get messed up?

src/components/video-texture-target.js
```diff
@@ -1,18 +1,86 @@
 import { disposeTexture } from "../utils/material-utils";
 import { createVideoOrAudioEl } from "../utils/media-utils";
+import { findNode } from "../utils/three-utils";
+
+/**
+ * @component video-texture-source
+ * This component is intended to be used on entities with a Camera Object3D as a child.
+ * That camera is used to render the scene to a WebGLRenderTarget of the specified resolution
+ * at a maximum of the specified frame rate. It will only render a frame if something sets
+ * it's textureNeedsUpdate property to true. Currently video-texture-target does this
```
johnshaughnessy, line 10: it's -> its
its textureNeedsUpdate property

src/components/video-texture-target.js
```diff
@@ -1,18 +1,86 @@
...
+ * whenever its material is used during the primary camera render (as in, its in view).
```
johnshaughnessy, line 11: its -> when it's
when it's in view

johnshaughnessy (approved):
I like that the `video-texture-source` component is simple. It just renders its camera to a texture whenever its flag is set. 

Something about using `indexToEntityMap[src]` to find the `video-texture-source` component's `entity` feels slightly suspicious to me, but I didn't have a good suggestion for how to do it differently. I thought about suggesting putting unique entity ID's into the `src` (similar to what we do with `networked` objects), but that has its own problems. I think I like that ID's are serializable but so is the `src` property that you're using to identify the `target`, so maybe I'm just wrong to be suspicious.

src/components/video-texture-target.js
```diff
@@ -1,18 +1,86 @@
...
+      magFilter: THREE.NearestFilter,
+      encoding: THREE.GammaEncoding,
+      depth: false,
+      stencil: false
+    });
+
+    const texture = this.renderTarget.texture;
+    texture.matrixAutoUpdate = false;
+    texture.matrix.scale(1, -1);
+    texture.matrix.translate(0, 1);
```
johnshaughnessy, line 37: Why the `scale` and `translate`?

src/components/video-texture-target.js
```diff
@@ -1,18 +1,86 @@
...
+    const sceneEl = this.el.sceneEl;
+    const renderer = this.renderer || sceneEl.renderer;
+
+    const tmpVRFlag = renderer.vr.enabled;
+    const tmpOnAfterRender = sceneEl.object3D.onAfterRender;
+    delete sceneEl.object3D.onAfterRender;
+    renderer.vr.enabled = false;
+
+    renderer.setRenderTarget(this.renderTarget);
+    renderer.render(sceneEl.object3D, this.camera);
```
johnshaughnessy, line 59: In order to create a mirror, do you invert the camera's x axis? (scale.x = -scale.x)?

src/components/video-texture-target.js
```diff
@@ -1,18 +1,86 @@
...
+ * a video-texture-src component on it.
  */
 AFRAME.registerComponent("video-texture-target", {
   schema: {
     src: { type: "string" },
     targetBaseColorMap: { type: "boolean", default: true },
-    targetEmissiveMap: { type: "boolean", default: false }
+    targetEmissiveMap: { type: "boolean", default: false },
+    // TODO having both src and target is a bit odd
+    target: { type: "selector" }
```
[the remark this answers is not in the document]
netpro2k: agree, `target` is a very confusing name, I think I just ended up blindly using that since I was looking at the trigger volume stuff as an example of being able to reference one node from another node in a gltf. I knew I was going to rework it anyway so didn't bother thinking of a better name.

My issue with having both `src` and `srcEl` it's pretty easy to create ambiguous states. Like if you set both `src` and `srcEl` which one should be preferred? Maybe `src` should be set to "srcEl" in this case or something... Right now the only actual usage of `src` is for video avatars which is completely internal (as in not in gltfs) and not networked, so it might also be fine to just rework that.

src/gltf-component-mappings.js
```diff
@@ -423,4 +423,26 @@ AFRAME.GLTFModelPlus.registerComponent("audio-settings", "audio-settings", (el,
...
+  "video-texture-target",
+  (el, componentName, componentData, _components, indexToEntityMap) => {
+    const { targetBaseColorMap, targetEmissiveMap, src } = componentData;
+
+    const target = indexToEntityMap[src];
+    if (!target) {
+      console.warn(
+        `Error inflating gltf component "video-texture-target": Couldn't find target entity with index ${src}`
+      );
+      return;
```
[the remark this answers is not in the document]
netpro2k: Nope good catch. I think I would have caught that when testing for regressions in video avatars which I planned to do, but yeah this is obviously incorrect. This might be the right behavior if `target` was specified in the gltf but not found. Because in that case there was some sort of error in the asset.

src/components/video-texture-target.js
netpro2k: Ah yeah, didn't re-review this code yet, just carried it over from the existing code... I think the `video-'texture-source` should be responsible for cleaning this up. Will probably rework the video avatar stuff so that it just ends up passing in a texture so it can be the one managing its own texture as well.

src/components/video-texture-target.js
netpro2k: I think that's right yeah. The user has full control over the camera properties they put into the gltf so this should already be possible if they want to go for that effect.

Issue #3565, opened 2020-12-16
Improve the entering lobby time


takahirox:
**Is your feature request related to a problem? Please describe.**

Entering lobby takes time if there are many objects or avatars in a room especially on low-end device or slow internet because Hubs client fully loads all the necessary assets before entering a lobby.

![image](https://user-images.githubusercontent.com/7637832/102322458-f4ed3380-3f33-11eb-9387-1823a89ed3b1.png)

Long entering time may be a bad experience for users.

**Describe the solution you'd like**

Lazy downloading and background loading would improve the entering lobby time a lot and provide better experience.

A concern is that there may be a chance that "entered a lobby but assets are not ready" might be a bad experience for users. For example, "avatars are not displayed or rendered incorrectly (like lack of textures) but hearing voices" might be stressful. To reduce such stresses, we may need to experiment and check if actually the experience is good.

Probably we need different approaches for different asset types. For example, just lazily downloading background and then starting to render them when they are ready is ok for pictures and videos. But for avatars (and scenes), downloading default light assets before entering a lobby, first rendering them, downloading the real ones background, switching to the real ones when they are ready might be better.

Another concern is this change can make the Hubs client design complex.

**Describe alternatives you've considered**

Optimizing our assets and make them much lighter may be another solution. Pros is we don't need to change the code and can keep the design simple. Cons is no effect to user custom assets.

A mix solution may be letting assets have low resolution data like LOD. First downloading the lower ones and rendering them. Then downloading the real ones background and switching to the real ones when they are ready.

**Additional context**

First of all, we may need to measure the entering time especially on low-end devices and slow internet to know how important solving this problem is.

keianhzo:
A few ideas here:
- LOD info on assets would be good so we can load lowres textureless meshes until highres ones are ready. Also might help in lowend devices
- Show greyed out meshes before textures are loaded. Some translucent grey material or even use a fancy loading shader. Textures are usually the heaviest part of a model so we can show meshes early.
- For avatars maybe load a default avatar until the user's model is ready.

takahirox:
I'm working on LOD. I'm writing a glTF loader plugin https://github.com/takahirox/three-gltf-extensions for glTF LOD extension. It loads all levels simultaneously but the lowest one should be completed first. So we can expect shorter response time.

[...]

PR #4135, opened 2021-04-09
Lazy load media on scene entry

keianhzo:
Related https://github.com/mozilla/hubs/issues/3565

This PR adds initial support for lazy loading media upon scene loading to avoid waiting for all the media to be loaded to enter the scene.


takahirox:
Nice work! Let me test soon later.

One thing I talked with @netpro2k on Discord is there may be a chance that we need to change the loading cube effect because the current one may be annoying to users if there a lot of loading cubes in a viewport especially if media objects are big. This situation can happen on entry with this change.

takahirox:
One more thing. Do you think there is a use case that scene owner wants to disable lazy load on room entry, meaning the owner want users to enter a room when all the assets are ready? (eg. Artistic purpose.) If so we may add an option to disable it in scene settings or somewhere else.

And for the automatic performance test I have been working on, I may want a capability to disable it or an event which is fired on all the assets are ready.

keianhzo:
@takahirox I've updated to enable/disable that through an admin config feature "Lazy load media". You should be able to override that from a script doing a `configs.APP_CONFIG.features["lazy_load_media"] = true/false;`

Regarding the loading cube, if you both agree I'd rather consulting that with DPX and opening another issue to replace it in case they consider that's necessary, otherwise we might be blocking this for too long.

takahirox:
Thanks for the update. Just in case I would like to know if other devs agree `lazy_media_load` configuration (for example, some of them might want to always enable it).

Opening another issue for loading cube sounds good to me.

I don't think I understand the scene entry flow so I would like other devs to review the code more deeply.

(And I would like to check the user experience of the case where a lot of medias are in a room but I couldn't do yet due to lack of time..)

keianhzo:
Now that you mention that, it might be interesting to also allow set it in the room settings and not at an admin level. Maybe @Dom has a better sense of what might be more useful.

keianhzo:
I've rebased this and fixed a couple of issues:
- The loading cube blend mode (was `alphaBlend` for some reason)
- Since the networking refactor we don't get the `didConnectToNetworkedScene` event when the adapter is connected anymore. I've added events to the adapter to handle the connected event correctly as we rely on it for the scene loading UI.

src/react-components/room/useRoomLoadingState.js
```diff
@@ -53,6 +69,7 @@ const messages = defineMessages({
 export function useRoomLoadingState(sceneEl) {
   // Holds the id of the current
   const loadingTimeoutRef = useRef();
+  const lazyLoadMedia = useRef(!!APP.store.state.preferences.lazyLoadSceneMedia);
```
netpro2k, line 72: I don't think using `useRef()` here is buying you anything, the only thing you are passing it to is the `useReducer()` initial state, but you aren't actually passing the reference to it, just the current value (it would be pretty weird to pass a mutable reference to a reducer anyway)

I assume what you were aiming to do with this was to have this re-render if that pref changed, but as written this wont actually do that. The "correct" way to handle this would be listening for `statechanged` event in a `useEffect()` like we do for example here https://github.com/mozilla/hubs/blob/master/src/react-components/room/useResizeViewport.js#L44-L64 and then dispatching an action to update the reducer state. I don't think its worth doing this since you only see this screen on first load anyway, so there is no real way a user is going to be able to change the preference. Even if we did show it multiple times, the user would never change the preferences during a given load. I would just get rid of `useRef()` and directly access the value (be sure to update all the locations accessing `.current`.

src/react-components/room/useRoomLoadingState.js
```diff
@@ -105,35 +130,50 @@ export function useRoomLoadingState(sceneEl) {
...
-        sceneEl.addEventListener("pdf-loading", onObjectLoading);
-        sceneEl.addEventListener("model-loaded", onObjectLoaded);
-        sceneEl.addEventListener("image-loaded", onObjectLoaded);
-        sceneEl.addEventListener("pdf-loaded", onObjectLoaded);
-        sceneEl.addEventListener("model-error", onObjectLoaded);
+        if (lazyLoadMedia.current) {
+          sceneEl.addEventListener("environment-scene-loading", onEnvironmentLoading);
+          lazyLoadMediaRefValue = lazyLoadMedia.current;
+        } else {
+          console.log("Attaching room state event listeners");
```
netpro2k, line 142: console.log

src/react-components/room/useRoomLoadingState.js
```diff
@@ -105,35 +130,50 @@ export function useRoomLoadingState(sceneEl) {
...
+          sceneEl.removeEventListener("pdf-loaded", onObjectLoaded);
+          sceneEl.removeEventListener("video-loaded", onObjectLoaded);
+          sceneEl.removeEventListener("model-error", onObjectLoaded);
+        }
         sceneEl.removeEventListener("environment-scene-loaded", onEnvironmentLoaded);
         sceneEl.removeEventListener("didConnectToNetworkedScene", onNetworkConnected);
       };
     },
-    [sceneEl, loading, onObjectLoaded, onObjectLoading, onEnvironmentLoaded, onNetworkConnected]
+    [sceneEl, loading, onObjectLoaded, onObjectLoading, onEnvironmentLoading, onEnvironmentLoaded, onNetworkConnected]
```
netpro2k, line 176: I think you need `lazyLoadMedia` in here since the effect should re-run if this changes.

src/react-components/room/useRoomLoadingState.js
```diff
@@ -14,18 +14,29 @@ function reducer(state, action) {
         loading: !(state.environmentLoaded && state.networkConnected),
         messageKey: state.environmentLoaded ? "enteringRoom" : "loadingObjects"
       };
+    case "environment-loading":
```
netpro2k, line 17: Been a long time since I looked at this code. Is there a reason this needs its own action vs setting `enviornmentLoaded` to false in the initial reducer state (its falsey either way since its `undefined` but setting it is cleaner)?

src/hub.js
```diff
@@ -654,7 +657,7 @@ function handleHubChannelJoined(entryManager, hubChannel, messageDispatch, data,
       scene.components["networked-scene"]
         .connect()
         .then(() => {
-          scene.emit("didConnectToNetworkedScene");
+          console.log("Networked scene connected");
```
netpro2k, line 660: Pretty sure emitting `didConnectToNetworkedScene` here was correct. Was there a reason you needed to move it? In any case this log message should probably be wherever `didConnectToNetworkedScene` is emitted.

netpro2k (commented):
This mostly lgtm. 

The moving of `didConnectToNetworkedScene` seems pretty fishy, it should be the reticulum connection and NAF adapter that have an opinion about this not dialog. 

Also, it seems like this PR removes LoadingObject_Atom.glb ... That doesn't seem correct. Was there supposed to be a replacement? This would be a good time to grab the updated model from Jim.

Made a few minor comments on the react code, but those are mostly just informational.

src/components/media-video.js
```diff
@@ -684,6 +684,7 @@ AFRAME.registerComponent("media-video", {
         }
       };
 
+      this.el.emit("video-loading");
```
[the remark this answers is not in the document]
keianhzo: Moved

src/react-components/room/useRoomLoadingState.js
keianhzo: Thanks for that, updated.

src/react-components/room/useRoomLoadingState.js
keianhzo: yeah, that's true. Removed

src/hub.js
```diff
@@ -776,6 +779,9 @@ document.addEventListener("DOMContentLoaded", async () => {
   const entryManager = new SceneEntryManager(hubChannel, authChannel, history);
   window.APP.entryManager = entryManager;
 
+  APP.dialog.on(DIALOG_CONNECTION_CONNECTED, () => {
+    scene.emit("didConnectToNetworkedScene");
```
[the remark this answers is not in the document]
keianhzo: So my understanding is that there are two different things here. First is the connection to the networked scene and then there is also the connection to dialog. I think we want to gate the room entry until both are ready. I've updated that in the latest commit.

src/hub.js
keianhzo: That's right, we still need that event. I've separated this in two events, the connection to the networked scene and the connection to dialog. I believe we need both to grant access to the room.

keianhzo:
@netpro2k Thanks for the review, there were a few thing that were wrong or not optimal. I've have updated with a few changes.

PR #4144, opened 2021-04-12
Bump Puppeteer to the latest one

takahirox:
As I explained on Discord, `run-bot.js` no longer works on my Windows10 due to page crash error on Chromium. Somehow bumping Puppeteer to the latest one has resolved the problem.

Brian couldn't reproduce the problem on his Windows so I'm not sure you folks encounter this problem and I couldn't figure out the root issue but I think updating the dependencies is generally good.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-812)

### DIFF
--- a/scripts/bot/package.json
+++ b/scripts/bot/package.json
@@ -4,7 +4,7 @@
   "license": "MPL-2.0",
   "devDependencies": {
     "docopt": "^0.6.2",
-    "puppeteer": "7.0.1",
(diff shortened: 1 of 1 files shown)


brianpeiris:
Done in #5251

takahirox:
Oh, I completely forgot that I had the same problem and suggested the same solution.

PR #4163, opened 2021-04-20
Restore the occupancy count lost in the new UI update

rawnsley:
I'm not sure if this was removed for a good reason or lost in the shuffle, but the number of occupants is the first thing I look at when entering a Hubs room. Same layout as before and the same as when you open the people list:

Screenshot 2021-04-20 at 16 15 34

I based the implementation on the more sophisticated `PeopleSidebarContainer`, which uses a timer for regular updates rather than any event-based subscription, which would otherwise have been the more obvious choice. I'm learning React as I go along so apologies if I've missed something important.


rawnsley:
Closed in favour of https://github.com/mozilla/hubs/pull/4199

rawnsley:
Closed in favour of https://github.com/mozilla/hubs/pull/4199

PR #4177, opened 2021-04-22
Browsers on mobile devices can return null for map.get("KeyQ")

rawnsley:
Chrome 90.0.4430.82 on Android 10 returns null for calls like `map.get("KeyQ")`, presumably because there is no keyboard attached. The error doesn't appear to be fatal and the result is not used by TipContainer, but it leads to an unnecessary and distracting error message in the developer console:

Screenshot 2021-04-22 at 09 29 42

### DIFF
--- a/src/react-components/room/TipContainer.js
+++ b/src/react-components/room/TipContainer.js
@@ -16,8 +16,8 @@ let turnRightKey = "E";
 if (window.navigator.keyboard !== undefined && window.navigator.keyboard.getLayoutMap) {
   window.navigator.keyboard.getLayoutMap().then(function(map) {
     moveKeys = `${map.get("KeyW")} ${map.get("KeyA")} ${map.get("KeyS")} ${map.get("KeyD")}`.toUpperCase();
-    turnLeftKey = map.get("KeyQ").toUpperCase();
-    turnRightKey = map.get("KeyE").toUpperCase();
+    turnLeftKey = map.get("KeyQ")?.toUpperCase();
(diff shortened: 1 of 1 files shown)


src/react-components/room/TipContainer.js
```diff
@@ -16,8 +16,8 @@ let turnRightKey = "E";
 if (window.navigator.keyboard !== undefined && window.navigator.keyboard.getLayoutMap) {
   window.navigator.keyboard.getLayoutMap().then(function(map) {
     moveKeys = `${map.get("KeyW")} ${map.get("KeyA")} ${map.get("KeyS")} ${map.get("KeyD")}`.toUpperCase();
```
johnshaughnessy, line 18: If these return `null` we'll get `moveKeys = "NULL NULL NULL NULL"`, but I guess if we're on mobile, `moveKeys` won't be used in a tip... Still, perhaps this line should read
```js
    moveKeys = `${map.get("KeyW") || "W"} ${map.get("KeyA") || "A"} ${map.get("KeyS") || "S"} ${map.get("KeyD") ||
      "D"}`.toUpperCase();
```

johnshaughnessy (approved):
Thanks @rawnsley

src/react-components/room/TipContainer.js
rawnsley: I've added that in as suggested. I ignore it originally because, as you say, it's not used, but it could be exposed and trip somebody up in the future.

PR #4178, opened 2021-04-22
document.body can be null when setLocale is first called

rawnsley:
A common, but non-deterministic, console error is being thrown in `setLocale` because document.body can be null.

Screenshot 2021-04-22 at 09 31 30

It should be safe to skip raising the event with `?.` in this situation because at best it was throwing an exception before.

A common pattern elsewhere in the codebase is to hang these events off `window` (which is guaranteed to be non-null) rather than `document.body`, and that would appear to be preferable in this case as well, but there may be something special about events on the `document.body` that I'm missing.

### DIFF
--- a/src/react-components/preferences-screen.js
+++ b/src/react-components/preferences-screen.js
@@ -489,11 +489,11 @@ class PreferenceListItem extends Component {
   };
   componentDidMount() {
     this.props.store.addEventListener("statechanged", this.storeUpdated);
-    document.body.addEventListener("locale-updated", this.storeUpdated);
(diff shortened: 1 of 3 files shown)


netpro2k:
Hmm a quick check of how we are using `locale-updated` seems like its not particularly important where the event is coming from. I think your suggestion to use window would work, and I would prefer that over silently failing which might happen with how the PR is currently implemented.

rawnsley:
Reimplemented as suggested by @netpro2k

netpro2k:
Merged. Thanks for making these additional changes!

Issue #4117, opened 2021-04-05
Replace our Three.js glTF loader fork with the official one for Hubs client


takahirox:
This issue is a sub issue of #4108 to discuss how we can replace [our Three.js glTF loader fork](https://github.com/MozillaReality/three.js/blob/hubs/master/examples/js/loaders/GLTFLoader.js) with the official one for Hubs Client.

### Involved files in hubs repo

* [avatar-editor.js](https://github.com/mozilla/hubs/blob/master/src/react-components/avatar-editor.js)
* [gltf-model-plus.js](https://github.com/mozilla/hubs/blob/master/src/components/gltf-model-plus.js)
* [HubsTextureLoader.js](https://github.com/mozilla/hubs/blob/master/src/loaders/HubsTextureLoader.js)
* [GLTFLoader@MozillaReality/three.js](https://github.com/MozillaReality/three.js/blob/hubs/master/examples/js/loaders/GLTFLoader.js)

Please edit if lack of the involved files

### Why do we need our fork and how can we replace?

* We want to handle our custom glTF extensions (TODO: List up the extensions and clarify their specifications). Ideally we may replace the handlers with the ones using Three.js glTF loader extensibility mechanism, called plugin-system.
* We want to use UUID, not name, for specifying animation target node (For what?). IIRC, the official Three.js glTF loader devs seem to think of adding an option using UUID for animation target node. We may wait for it or we may be able to use glTF loader `afterRoot` hook to replace the target names with the target UUIDs after parsing the entire glTF file(?).

Please edit or update if lack of information

### How can we safely replace?

* Rewriting our handlers to the ones using glTF plugin-system may not be so hard (hopefully). But we should carefully confirm that the change won't break Hubs. Especially `glTF-model-plus` is one of the core components in Hubs Client. If it's broken it has a big bad impact. We need to prepare the test sets to check it keeps working.

netpro2k:
> We want to use UUID, not name, for specifying animation target node (For what?).
[...]

Issue #1613, opened 2019-08-06
Certain GLBs load with black textures


brianpeiris:
**Description**
Some GLBs load with missing/black textures. E.g: https://poly.google.com/view/ejF8pZvNHHh
This model loads correctly in all other GLTF viewers, but not in Spoke and Hubs.

**To Reproduce**
1. Load https://poly.google.com/view/ejF8pZvNHHh in Hubs
1. See that the texture is black

**Expected behavior**
GLB should appear as it does in other GLTF viewers.

**Screenshots**
GLB in Hubs:
![image](https://user-images.githubusercontent.com/79419/62565528-d3f2d980-b83b-11e9-846a-9bd82c795c47.png)
GLB in Don McCurdy's GLTF Viewer:
![image](https://user-images.githubusercontent.com/79419/62565609-03a1e180-b83c-11e9-9e44-83eeaaaa51ad.png)

**Hardware** 
 - Device: Desktop
 - OS: Windows
 - Browser: Firefox

joshmarinacci:
Do we support this now?

aelainelong:
Also curious to know whether this has been/is being addressed or will be in the near future?

takahirox:
It seems #4198 resolves this issue. Specular/Glossiness looks rendered correctly on my test server.

takahirox:
@brianpeiris I merged #4198. Would you please test on your end again and close this issue if you confirm it's fixed?

PR #4198, opened 2021-04-27
Bump into the latest glTF loader

takahirox:
Requires: https://github.com/MozillaReality/three.js/pull/57
From: #4117

This PR replaces our Three.js glTF loader fork with the latest official Three.js one and updates `gltf-model-plus.js` and `avatar-editor.js` to follow the latest official Three.js `GLTFLoader` API.

Benefits

* Less cost for maintenance because we will no longer need to maintain our GLTFLoader fork
* We will be easily able to add or manage new and existing glTF related features with the official GLTFLoader plugin API for example LOD

Please refer to https://github.com/mozilla/hubs/issues/4117#issuecomment-818319656 for the changes.

Note that the changes in this PR is just for replacing the glTF loader. There may be a chance that they can be further cleaned up or optimized with the new API. But I would like to do them if needed in another PR because I don't want to make this PR bigger and the review harder.

TODO
- [x] merge https://github.com/MozillaReality/three.js/pull/57
- [x] update package.json to point back at hubs/master for three
- [x] update package-lock.json to point back at hubs/master and the merged commit sha

### DIFF
--- a/package.json
+++ b/package.json
@@ -134,7 +134,7 @@
     "resize-observer-polyfill": "^1.5.1",
     "screenfull": "^4.0.1",
     "semver": "^7.3.2",
-    "three": "github:mozillareality/three.js#hubs/r128devGLTFLoader",
+    "three": "github:mozillareality/three.js#hubs/r128GLTFLoader",
     "three-ammo": "^1.0.12",
     "three-bmfont-text": "github:mozillareality/three-bmfont-text#hubs/master",
     "three-mesh-bvh": "^0.1.2",
--- a/package.json
+++ b/package.json
@@ -134,7 +134,7 @@
     "resize-observer-polyfill": "^1.5.1",
     "screenfull": "^4.0.1",
     "semver": "^7.3.2",
-    "three": "github:mozillareality/three.js#hubs/r128GLTFLoader",
+    "three": "github:mozillareality/three.js#hubs/master",
     "three-ammo": "^1.0.12",
     "three-bmfont-text": "github:mozillareality/three-bmfont-text#hubs/master",
     "three-mesh-bvh": "^0.1.2",
--- a/package.json
+++ b/package.json
@@ -134,7 +134,7 @@
     "resize-observer-polyfill": "^1.5.1",
     "screenfull": "^4.0.1",
     "semver": "^7.3.2",
-    "three": "github:mozillareality/three.js#hubs/master",
+    "three": "github:mozillareality/three.js#hubs/r128devGLTFLoader",
     "three-ammo": "^1.0.12",
     "three-bmfont-text": "github:mozillareality/three-bmfont-text#hubs/master",
     "three-mesh-bvh": "^0.1.2",
--- a/src/components/gltf-model-plus.js
+++ b/src/components/gltf-model-plus.js
@@ -355,6 +355,7 @@ class GLTFHubsPlugin {
     // so we can use ImageBitmapLoader even for Firefox.
     // When we replace our Three.js fork with the latest official one
     // we need to revisit this workaround.
+    // GLTFLoader creates CanvasTexture from ImageBitmap if textureLoader is ImageBItmapLoader.
     if (!parser.textureLoader.isImageBitmapLoader && typeof createImageBitmap !== undefined) {
       parser.textureLoader = new THREE.ImageBitmapLoader(parser.options.manager);
     }
--- a/src/components/gltf-model-plus.js
+++ b/src/components/gltf-model-plus.js
@@ -347,6 +347,18 @@ class GLTFHubsPlugin {
     this.parser = parser;
     this.jsonPreprocessor = jsonPreprocessor;
     this.fileMap = fileMap;
+
+    // The latest GLTFLoader doesn't use ImageBitmapLoader for Firefox
+    // because the latest ImageBitmapLoader passes an option parameter
+    // to createImageBitmap() but Firefox createImageBitmap() fails
+    // if an option parameter is passed (known bug and already reported to bugzilla).
+    // But our r111 based ImageBitmapLoader doesn't pass the option parameter yet
+    // so we can use ImageBitmapLoader even for Firefox.
+    // When we replace our Three.js fork with the latest official one
+    // we need to revisit this workaround.
+    if (!parser.textureLoader.isImageBitmapLoader && typeof createImageBitmap !== undefined) {
+      parser.textureLoader = new THREE.ImageBitmapLoader(parser.options.manager);
+    }
   }
 
   beforeRoot() {
--- a/src/components/gltf-model-plus.js
+++ b/src/components/gltf-model-plus.js
@@ -356,7 +356,7 @@ class GLTFHubsPlugin {
     // When we replace our Three.js fork with the latest official one
     // we need to revisit this workaround.
     // GLTFLoader creates CanvasTexture from ImageBitmap if textureLoader is ImageBItmapLoader.
-    if (!parser.textureLoader.isImageBitmapLoader && typeof createImageBitmap !== undefined) {
+    if (!parser.textureLoader.isImageBitmapLoader && typeof createImageBitmap !== "undefined") {
       parser.textureLoader = new THREE.ImageBitmapLoader(parser.options.manager);
     }
   }
--- a/src/components/gltf-model-plus.js
+++ b/src/components/gltf-model-plus.js
@@ -370,7 +370,7 @@ class GLTFHubsPlugin {
       parser.json = jsonPreprocessor(parser.json);
     }
 
-    //
+    // Ideally Hubs components stuffs should be handled in MozHubsComponents plugin?
     let version = 0;
     if (
       parser.json.extensions &&
--- a/src/components/gltf-model-plus.js
+++ b/src/components/gltf-model-plus.js
@@ -5,6 +5,7 @@ import { getCustomGLTFParserURLResolver } from "../utils/media-url-utils";
 import { promisifyWorker } from "../utils/promisify-worker.js";
 import { MeshBVH, acceleratedRaycast } from "three-mesh-bvh";
 import { disposeNode, cloneObject3D } from "../utils/three-utils";
+import HubsTextureLoader from "../loaders/HubsTextureLoader";
 import { KTX2Loader } from "three/examples/jsm/loaders/KTX2Loader";
 
 THREE.Mesh.prototype.raycast = acceleratedRaycast;
@@ -347,18 +348,21 @@ class GLTFHubsPlugin {
     this.parser = parser;
     this.jsonPreprocessor = jsonPreprocessor;
 
-    // The latest GLTFLoader doesn't use ImageBitmapLoader for Firefox
-    // because the latest ImageBitmapLoader passes an option parameter
-    // to createImageBitmap() but Firefox createImageBitmap() fails
-    // if an option parameter is passed (known bug and already reported to bugzilla).
-    // But our r111 based ImageBitmapLoader doesn't pass the option parameter yet
-    // so we can use ImageBitmapLoader even for Firefox.
-    // When we replace our Three.js fork with the latest official one
-    // we need to revisit this workaround.
-    // GLTFLoader creates CanvasTexture from ImageBitmap if textureLoader is ImageBItmapLoader.
-    if (!parser.textureLoader.isImageBitmapLoader && typeof createImageBitmap !== "undefined") {
-      parser.textureLoader = new THREE.ImageBitmapLoader(parser.options.manager);
-    }
+    // We override glTF parser textureLoader with our HubsTextureLoader for
+    // 1. Clean up the texture image related resources after it is uploaded to WebGL texture
+    // 2. Use ImageBitmapLoader if it is available. The latest official Three.js (r128) glTF loader
+    //    attempts to use ImageBitmapLoader if it is avaiable except for Firefox.
+    //    But we want to use ImageBitmapLoader even for Firefox so we need this overriding hack.
+    // But overriding the textureLoader is hacky and it can cause future potential problems
+    // especially in upstraming Three.js. So ideally we should replace this hack
+    // with the one using more proper APIs like plugin system at some point.
+    // Note: Be careful when replacing our Three.js fork with the official upstraming one that
+    //       the latest ImageBitmapLoader passes the second argument to createImageBitmap()
+    //       but currently createImageBitmap() on Firefox fails if the second argument is defined.
+    //       https://bugzilla.mozilla.org/show_bug.cgi?id=1367251
+    //       and this is the reason why the official glTF loader doesn't use ImageBitmapLoader
+    //       for Firefox. We will need a workaround for that.
+    parser.textureLoader = new HubsTextureLoader(parser.options.manager);
   }
 
   beforeRoot() {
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -134,7 +134,7 @@
     "resize-observer-polyfill": "^1.5.1",
     "screenfull": "^4.0.1",
     "semver": "^7.3.2",
-    "three": "github:mozillareality/three.js#hubs/master",
+    "three": "github:mozillareality/three.js#hubs/r128GLTFLoader",
```
takahirox, line 137: If https://github.com/MozillaReality/three.js/pull/57 is merged, I would update this line back to `master` in this PR or follow up PR. `package-lock.json` is same.

takahirox:
What I have tested so far.
- Enter a room and see an environment is rendered (it means the loader succeeds in loading a glTF scene)
- Upload a glTF object including moz light map extension
- See avatar preview in Avatar settings
- Upload a custom avatar and change the avatar to it

package.json
netpro2k: I usually do it just before merging, and I leave a [ ] todo in the PR description to remind me

src/components/gltf-model-plus.js
```diff
@@ -341,127 +340,221 @@ function runMigration(version, json) {
...
-    parser.json.extensions.MOZ_hubs_components &&
-    parser.json.extensions.MOZ_hubs_components.hasOwnProperty("version")
-  ) {
-    version = parser.json.extensions.MOZ_hubs_components.version;
-  }
-  runMigration(version, parser.json);
+  afterRoot(gltf) {
+    gltf.scene.traverse(object => {
+      // GLTFLoader sets matrixAutoUpdate on animated objects, we want to keep the defaults
+      // @TODO: Should this be fixed in the gltf loader?
```
netpro2k, line 402: I know this was just copied over from our existing code, but it still does seem like the GLTFLoader should be respecting this default.

src/components/gltf-model-plus.js
```diff
@@ -341,127 +340,221 @@ function runMigration(version, json) {
...
+    // The latest GLTFLoader doesn't use ImageBitmapLoader for Firefox
+    // because the latest ImageBitmapLoader passes an option parameter
+    // to createImageBitmap() but Firefox createImageBitmap() fails
+    // if an option parameter is passed (known bug and already reported to bugzilla).
+    // But our r111 based ImageBitmapLoader doesn't pass the option parameter yet
+    // so we can use ImageBitmapLoader even for Firefox.
+    // When we replace our Three.js fork with the latest official one
+    // we need to revisit this workaround.
+    if (!parser.textureLoader.isImageBitmapLoader && typeof createImageBitmap !== undefined) {
+      parser.textureLoader = new THREE.ImageBitmapLoader(parser.options.manager);
```
netpro2k, line 360: ImageBitmapLoader returns an ImageBitmap not a Texture instance I thought. Doesn't this need to be a some kind of texture loader?

src/components/gltf-model-plus.js
```diff
@@ -341,127 +340,221 @@ function runMigration(version, json) {
...
-  if (nodes) {
-    for (let i = 0; i < nodes.length; i++) {
-      const node = nodes[i];
+    //
+    gltf.scene.traverse(object => {
+      if (!object.material) {
+        return;
+      }
+      const materials = Array.isArray(object.material) ? object.material : [object.material];
+      // @TODO: Handle more efficiently
```
netpro2k, line 414: We are already overriding `parser.textureLoader` in the constructor. Seems like we can still use the HubsTextureLoader to take care of this no?

src/components/gltf-model-plus.js
```diff
@@ -341,127 +340,221 @@ function runMigration(version, json) {
...
+            }
+          }
+        }
       }
+    }
 
-      node.extras.gltfIndex = i;
+    //
+    if (this.fileMap) {
+      // The GLTF is now cached as a THREE object, we can get rid of the original blobs
```
netpro2k, line 448: Just a mental note I re-remember while reading through this PR. This is to clean up the root level resources we create (from Sketchfab zip files) to load the GLTF.

Does the official GLTFLoader have any method of cleaning up the resources it creates?

src/components/gltf-model-plus.js
```diff
@@ -341,127 +340,221 @@ function runMigration(version, json) {
...
+    if (jsonPreprocessor) {
+      parser.json = jsonPreprocessor(parser.json);
+    }
 
-  if (ktxLoader) {
-    gltfLoader.setKTX2Loader(ktxLoader);
-  }
+    //
+    let version = 0;
+    if (
```
netpro2k, line 375: Not particularly important but eventually we might want this to be its own MozHubsComponetnsExtension that just handles the components extension and not the other weird hubs specific workarounds. Doesn't really matter too much now since the components stuff is all still very tightly integrated into aframe and the rest of Hubs.

src/components/gltf-model-plus.js
```diff
@@ -341,127 +340,221 @@ function runMigration(version, json) {
...
+              delete prop.image;
+            };
+          }
+        }
+      }
+    });
 
-      if (!node.extras) {
-        node.extras = {};
+    // Replace animation target node name with the node uuid.
```
netpro2k, line 429: As we have discussed before, it still seems a bit weird that this isn't just the default behavior since GLTF does not guarantee unique names... This is definitely preferable to maintaining a fork, but would like to get to the bottom of why this isn't always desirable.

src/components/gltf-model-plus.js
```diff
@@ -341,127 +340,221 @@ function runMigration(version, json) {
...
+  }
+
+  // @TODO: Ideally we should use extendMaterialParams hook.
+  //        But the current official glTF loader doesn't fire extendMaterialParams
+  //        hook for unlit and specular-glossiness materials.
+  //        So using loadMaterial hook as workaround so far.
+  //        Cons is loadMaterial hook is fired as _invokeOne so
+  //        if other plugins defining loadMaterial is registered
+  //        there is a chance that this light map extension handler isn't called.
+  //        The glTF loader should be updated to remove the limitation.
```
netpro2k, line 470: Hmm yeah we should discuss. Figuring out ow multiple extensions handling the same thing should interact is tricky, but I could definitely see cases where there are multiple layers of extensions for loading a single thing.

src/components/gltf-model-plus.js
```diff
@@ -341,127 +340,221 @@ function runMigration(version, json) {
...
+    this.name = "MOZ_lightmap";
+  }
+
+  // @TODO: Ideally we should use extendMaterialParams hook.
+  //        But the current official glTF loader doesn't fire extendMaterialParams
+  //        hook for unlit and specular-glossiness materials.
+  //        So using loadMaterial hook as workaround so far.
+  //        Cons is loadMaterial hook is fired as _invokeOne so
+  //        if other plugins defining loadMaterial is registered
+  //        there is a chance that this light map extension handler isn't called.
```
netpro2k, line 469: Since we have control over the loader this shouldnt really be an issue in practice, but we will need to be careful when doing upgrades that we don't accidentally break it by some plugin being introduced upstream. Hopefully this workaround doesn't last too long anyway.

src/components/gltf-model-plus.js
```diff
@@ -341,127 +340,221 @@ function runMigration(version, json) {
...
+    const parser = this.parser;
+    const json = parser.json;
+    const textureDef = json.textures[textureIndex];
+
+    if (!textureDef.extensions || !textureDef.extensions[this.name]) {
+      return null;
+    }
+
+    if (!parser.options.ktx2Loader) {
+      // @TODO: Display warning (only if the extension is in extensionsRequired)?
```
netpro2k, line 513: In practice we will always be providing a ktx2loader so if this happens its certainly a bug... We probably don't want to fail silently.

src/components/gltf-model-plus.js
```diff
@@ -341,127 +340,221 @@ function runMigration(version, json) {
...
+              break;
+            }
+          }
+        }
       }
+    }
 
-      node.extras.gltfIndex = i;
+    //
+    if (this.fileMap) {
```
netpro2k, line 447: Might be cleaner to just handle this in `loadGLTF` instead of passing `fileMap` through.

src/components/gltf-model-plus.js
```diff
@@ -569,7 +662,7 @@ AFRAME.registerComponent("gltf-model-plus", {
       // If we had inflated something already before, clean that up
       this.disposeLastInflatedEl();
 
-      this.model = gltf.scene || gltf.scenes[0];
+      this.model = gltf.scene;
```
netpro2k, line 665: Is this guaranteed to be set now? I think the reason we had it originally is that not GLTFs are not required to have a default scene (though almost all I have encountered in practice do).

src/react-components/avatar-editor.js
```diff
@@ -47,6 +47,53 @@ const fetchAvatar = async avatarId => {
...
+    this.parser = parser;
+    this.inputFiles = inputFiles;
+  }
+
+  beforeRoot() {
+    const parser = this.parser;
+    const { body } = parser.extensions.KHR_binary_glTF;
+    const content = JSON.stringify(ensureAvatarMaterial(parser.json));
+
+    // Inject hubs components on upload. Used to create base avatar
```
netpro2k, line 62: I think you can just get rid of this commented code now. This predates the Blender extension and was an easy way for us to create some of the initial avatars, but we don't need it anymore.

src/react-components/avatar-editor.js
```diff
@@ -47,6 +47,53 @@ const fetchAvatar = async avatarId => {
...
+      type: "model/gltf"
+    });
+    this.inputFiles.bin = new File([body], "file.bin", {
+      type: "application/octet-stream"
+    });
+
+    // This plugin just wants to split gltf and bin from glb and
+    // doesn't want to start the parse. But glTF loader plugin API
+    // doesn't have an ability to cancel the parse. So overriding
+    // parser.json with very light glTF data as workaround.
```
netpro2k, line 92: hah nice workaround

src/react-components/avatar-editor.js
```diff
@@ -113,42 +160,18 @@ class AvatarEditor extends Component {
...
-      // });
-      // content = JSON.stringify(gltf);
-
-      this.inputFiles.gltf = new File([content], "file.gltf", {
-        type: "model/gltf"
-      });
-      this.inputFiles.bin = new File([body], "file.bin", {
-        type: "application/octet-stream"
+      await new Promise((resolve, reject) => {
+        // This GLTFBinarySplitterPlugin saves gltf and bin in inputFiles.gltf/bin
```
netpro2k, line 170: Something about this feels a bit weird but I don't really have a much better idea. I maybe would have had the plugin just store it on itself rather than have it passed in and then done something like this after load:

```js
this.inputFiles.gltf = binarySplitterPlugin.gltf;
this.inputFiles.bin = binarySplitterPlugin.bin;
```

But not sure that is really much better.

netpro2k (commented):
This looks great! Definitely feels like a lot more of this is now being don the "right way'. Still some hacks in place but I think we have a clear path forward to improve most of them. Nice work!

src/components/gltf-model-plus.js
takahirox: [`GLTFLoader` makes `CanvasTexture` from `ImageBitmap` if `textureLoader` is `ImageBitmapLoader`](https://github.com/mrdoob/three.js/blob/r128/examples/jsm/loaders/GLTFLoader.js#L2578-L2586) so this line should be fine.

But this line may be confusing for people who don't know the inside of `GLTFLoader`. I would add comment.

src/components/gltf-model-plus.js
takahirox: I think `GLTFLoader` should respect the default configuration, too. I'm going to suggest it to Three.js at some point. If it will be merged, we can drop this code.

src/components/gltf-model-plus.js
takahirox: OK, I would add warning here

src/components/gltf-model-plus.js
takahirox: [Yes, it's guaranteed in the latest `GLTFLoader`](https://github.com/mrdoob/three.js/blob/r128/examples/jsm/loaders/GLTFLoader.js#L2042). IIRC the official loader didn't do that before so I think you encountered the problem with the old loader.

src/react-components/avatar-editor.js
takahirox: Good to know that this workaround seems acceptable.

brianpeiris:
I will remove my need to review, since you are both already experts in this code and Dom has done a thorough review.

takahirox:
Thanks for the reviews.

Just in case, I would like to inform you folks that please don't merge this PR until we finish the test with the QA team.

takahirox:
I just noticed (remembered) that the latest `GLTFLoader` has dropped DDS texture support because `MSFT_texture_dds` extension is a vendor extension, not a Khronos KHR extension, and they encourage to use more standard compressed texture `KHR_texture_basisu` extension now.

If we want to keep supporting DDS texture, we can use an external DDS texture plugin I wrote 

https://github.com/takahirox/three-gltf-extensions/tree/main/loaders/MSFT_texture_dds

Any opinion on this? We should encourage KHR_texture_basisu, too, but also I think we don't need to drop DDS texture support and require the asset authors to update their works.

src/components/gltf-model-plus.js
takahirox: I didn't really want to override `parser.textureLoader` because `GLTFLoader` doesn't have a proper API to override the texture loader. Overriding `parser.textureLoader = new CustomLoader()` is hacky and the `GLTFLoader` doesn't expect the texture loader to be overridden. There might be a chance that it causes future errors especially when updating to the future glTF loader.

I did override the texture loader because there doesn't seem to be other options to let `GLTFLoader` use `ImageBitmapLoader` for Firefox. Yes, overriding the texture loader is hacky as I wrote but `GLTFLoader` expects that the texture loader can be `ImageBitmapLoader` so I think it would be less problematic than using our custom texture loader. And if Firefox and Three.js will be updated to support `ImageBitmapLoader` we will be able to simply drop this hacky code.

But if we realize that this traverse is slow I want to revisit.

src/components/gltf-model-plus.js
takahirox: If a plugin manage resources `afterRoot` hook would be a good place to release the resources.

src/components/gltf-model-plus.js
takahirox: Sounds good. Updated.

src/react-components/avatar-editor.js
takahirox: How about this?

```
await new Promise((resolve, reject) => {
  // GLTFBinarySplitterPlugin saves gltf and bin in gltf.files
  gltfLoader.load(gltfUrl, result => {
    this.inputFiles.gltf = result.files.gltf;
    this.inputFiles.bin = result.files.bin;
  }, onProgress, reject);
});
```

src/components/gltf-model-plus.js
takahirox: Added a comment so far and would like to think of the cleaning up later.

src/react-components/avatar-editor.js
takahirox: Done, thanks

takahirox:
Ah wait. I noticed that the current `gltf-model-plus.js` doesn't set `DDSLoader` to `GLTFLoader`. That means we don't support DDS texture even now. Then no DDS texture support won't be problem.

takahirox:
I reverted the removal of `HubsTextureLoader` from `gltf-model-plus.js` because it removes the resolved texture image url from Three.js Cache and I couldn't come up with a workaround with the more proper API like plugin API.

As we internally discussed on Discord, we may disable Three.js entire cache so we may want to revisit later.

I added a comment about why we use `HubsTextureLoader` and we should replace it later so far.

takahirox:
Somehow the glTF loader replacement resolves the Specular/Glossiness material problem #1613 The Specular/Glossiness material is rendered correctly on the testing server.

takahirox:
Regarding [`parser.cache.removeAll()`](https://github.com/mozilla/hubs/blob/ebfe192502f32cfda00fce22b8b675b65891cb4e/src/components/gltf-model-plus.js#L448), I thought I added it in `GLTFHubsPlugin` but it seems I somehow accidentally delete it.

I think it's ok because `glTFLoader` and `parser` aren't saved anywhere so the resources used by them should be garbage collected automatically.

https://github.com/mozilla/hubs/blob/693dc621377734a5391e78e18c4d7a2c947b169d/src/components/gltf-model-plus.js#L515-L539
https://github.com/mozilla/hubs/blob/693dc621377734a5391e78e18c4d7a2c947b169d/src/components/gltf-model-plus.js#L638-L718

PR #4199, opened 2021-04-28
Feature/people menu button count

MegaMotion:
Adds a presenceCount state variable to UIRoot class, to display number of people present in the room without clicking on the People Menu button.


takahirox:
Nice work! This feature is what I really wanted.

PR #4206, opened 2021-04-29
Handle XR controller disconnect and session end

### LINKED ISSUE #4134: VR + WebGL2 not working on Oculus Quest 1 and 2
**Description**
Clicking the "Enter VR" green button on the bottom will result in a WebXR session with black screen, making the VR experience unavailable. It works when first entering the room (hubs wants to start xr session, click allow), but using the button will not work.

The Invite button inside the initial WebXR session is also completely broken and does nothing.
Clicking the Create button will show a prompt "Continue" but will break the website by going out and then back into WebXR session but this time with a black screen.

**To Reproduce**
1. Go to hubs.mozilla.com on Oculus Browser
2. Enter the room, exit webxr session, click "Enter VR" button
3. See black screen

**Expected behavior**
Clicking "Enter VR" should start a WebXR session without pitch black darkness.

**Hardware** 
 - Device: Oculus Quest 2
 - Browser: Oculus Browser.

**Additional context**
Seems to have started after the UI update.
Avatar Head rotation and movement still works fine if you look at the avatar through another browser.

Here is the log from Chrome debugging / OQBrowser:
image

Here is another log from Chrome/OQBrowser on my hubscloud
image

netpro2k:
Exiting VR mode in Oculus browser now ends the XR Session (without triggering a `inputsourcechange` or `gamepaddisconnected` event. This PR correctly handles the `end` event on session to clean up controllers form the input system.

Also fixes the handling for the `inputsourcechange` event to correctly handle devices being lost not just added (this can happen for example if a controller runs out of batteries).

fixes #4134
fixes #4196

### DIFF
--- a/src/systems/userinput/userinput.js
+++ b/src/systems/userinput/userinput.js
@@ -412,13 +412,20 @@ AFRAME.registerSystem("userinput", {
       gamepad && gamepadConnected({ gamepad });
     }
 
-    const retrieveXRGamepads = ({ session }) => {
-      for (const inputSource of session.inputSources) {
-        inputSource.gamepad.isWebXRGamepad = true;
-        inputSource.gamepad.targetRaySpace = inputSource.targetRaySpace;
-        inputSource.gamepad.primaryProfile = inputSource.profiles[0];
-        inputSource.gamepad.hand = inputSource.handedness;
-        gamepadConnected(inputSource);
+    const retrieveXRGamepads = ({ added, removed }) => {
+      if (removed) {
+        for (const inputSource of removed) {
+          gamepadDisconnected(inputSource);
+        }
+      }
+      if (added) {
+        for (const inputSource of added) {
+          inputSource.gamepad.isWebXRGamepad = true;
+          inputSource.gamepad.targetRaySpace = inputSource.targetRaySpace;
+          inputSource.gamepad.primaryProfile = inputSource.profiles[0];
+          inputSource.gamepad.hand = inputSource.handedness;
+          gamepadConnected(inputSource);
+        }
       }
     };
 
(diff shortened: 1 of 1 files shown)


src/systems/userinput/userinput.js
```diff
@@ -429,6 +436,14 @@ AFRAME.registerSystem("userinput", {
...
           this.xrReferenceSpace = referenceSpace;
         });
+        xrSession.addEventListener("end", e => {
+          const xrDevices = this.activeDevices.items.filter(d => d.gamepad && d.gamepad.isWebXRGamepad);
+          for (const device of xrDevices) {
+            if (device.gamepad && device.gamepad.isWebXRGamepad) {
+              gamepadDisconnected(device);
+            }
+          }
+        });
```
takahirox, line 446: Why filtering two times? One is `this.activeDevices.items.filter(d => d.gamepad && d.gamepad.isWebXRGamepad)` and another one is `if (device.gamepad && device.gamepad.isWebXRGamepad)`. I think we can drop the latter one.

src/systems/userinput/userinput.js
```diff
@@ -412,13 +412,20 @@ AFRAME.registerSystem("userinput", {
...
     }
 
-    const retrieveXRGamepads = ({ session }) => {
-      for (const inputSource of session.inputSources) {
-        inputSource.gamepad.isWebXRGamepad = true;
-        inputSource.gamepad.targetRaySpace = inputSource.targetRaySpace;
-        inputSource.gamepad.primaryProfile = inputSource.profiles[0];
-        inputSource.gamepad.hand = inputSource.handedness;
-        gamepadConnected(inputSource);
+    const retrieveXRGamepads = ({ added, removed }) => {
```
takahirox, line 415: This line in `enter-vr` event listener calls `retrieveXRGamepads` with `session`.

https://github.com/mozilla/hubs/blob/fd81dd93f8b72c01d3d66e9e9458ee38b99386b5/src/systems/userinput/userinput.js#L447

Is it intentional? `added` and `removed` will be `undefined` so the function does nothing. Can't we simply drop the line?

src/systems/userinput/userinput.js
netpro2k: Ah yep good catch. I was originally just using the loop, but can't do that since the structure is being mutated so I added the pre-filter, forgot to remove the if.

src/systems/userinput/userinput.js
netpro2k: This ends up being a noop anyway since `inputsourceschange` will fire with the controllers anyway, but it does explain why I was seeing it called with just a session sometimes (which is why I added the `if(added)` and `if(removed)` checks. I think I can get rid of this and the checks.

Issue #3470, opened 2020-12-03
Performance problem on Firefox if video is placed in a room


takahirox:
**Description**

FPS number in a room is significantly worse on Firefox if video is placed in a room.

**To Reproduce**

Access this "room with video"

https://hubs.mozilla.com/rc7gbkQ/crazy-honored-nation

And also access this "room without video"

https://hubs.mozilla.com/PsvaJ6i/clear-authentic-space

And compare the fps number displayed at the bottom right.

In my Desktop Windows with NVIDIA GeForce GTX 1080

| Browser | With video | Without video |
| --- | --- | --- |
| Firefox Nightly| 30fps | 60fps |
| Chrome | 60fps | 60fps |

**Expected behavior**
The fps number in a "room with video" on Firefox should be as good as the one on Chrome.

**Hardware** 
 - Device: Desktop
 - OS: Windows
 - Browser: Firefox

**Additional context**

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-503)

takahirox:
Jim's case: Windows, Nvidia RTX 2080
| Browser | With video | Without video |
| ---- | ---- | ---- |
| Firefox | 90fps | 135fps |
| Chrome | 144fps | 144fps |

Dom's case: Linux, NVidia RTX 2080
| Browser | With video | Without video |
| ---- | ---- | ---- |
| Firefox | 39-45fps | 42-47fps |
| Chrome | 120fps | 120fps |

(In Dom's case, I guess the performance problem on Firefox seems from generic (JS engine?) issue rather than video stuffs.)

takahirox:
I took the profile on Firefox.

![image](https://user-images.githubusercontent.com/7637832/101076510-54bd0500-3558-11eb-9c6c-ad59d1598f2b.png)

The remarkable thing is `.setTexImage2D()` cpu time is huger with video. I guess Firefox has a problem uploading video to GPU?

takahirox:
Reported to bugzilla https://bugzilla.mozilla.org/show_bug.cgi?id=1680851

netpro2k:
The upstream `VideoTexture` in Three also makes this gap even worse since they now use `requestVideoFrameCallback` to reduce the work to only upload when there are actually new video frames, but only Chrome supports this currently https://github.com/mrdoob/three.js/blob/dev/src/textures/VideoTexture.js#L22. 

[...]

PR #4210, opened 2021-04-30
Add workaround for Firefox video play performance issue

takahirox:
From https://github.com/mozilla/hubs/issues/3470#issuecomment-829421706 and the following comments.

Firefox seems to have video play (or decode) performance issue. The FPS number becomes very bad if videos are played in viewport on Firefox.

[In Three.js repo, it has been reported that somehow setting RGBA format to video texture improves the performance very well.](https://github.com/mrdoob/three.js/pull/21746) (I guess it can avoid slow decoding in browser?).

[The ticket has been opened for the performance issue in bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1708491) but I don't think it can be resolved soon. So I would like to suggest to set RGBA format for Firefox as workaround so far.

11.2MB mp4 video + desktop Windows + NVIDIA GeForce GTX 1080.

| | RGB | RGBA |
|----|----|----|
| Firefox | 15fps | 60fps |
| Chrome | 60fps | 60fps |

Ideally we should test on more platforms (Mac, Mobile, other graphics boards and so on). I would be happy if you test on your end, too.

### DIFF
--- a/src/components/media-views.js
+++ b/src/components/media-views.js
@@ -731,6 +731,16 @@ AFRAME.registerComponent("media-video", {
         texture.minFilter = THREE.LinearFilter;
         texture.encoding = THREE.sRGBEncoding;
 
+        // Firefox seems to have video play (or decode) performance issue.
+        // Somehow setting RGBA format improves the performance very well.
+        // Some tickets have been opened for the performance issue but
+        // I don't think it will be fixed soon. So we set RGBA format for Firefox
+        // as workaround so far.
+        // See https://github.com/mozilla/hubs/issues/3470
+        if (/firefox/i.test(navigator.userAgent)) {
+          texture.format = THREE.RGBAFormat;
+        }
+
         isReady = () => {
           if (texture.hls && texture.hls.streamController.audioOnly) {
             audioEl = videoEl;


netpro2k (approved):
We should double check that this doesn't cause any issues in Firefox reality or Firefox for android but otherwise looks good to me. This is an awesomely simple "fix" to a big problem!

brianpeiris (approved):
My tests on various devices wiith a 4K video:

||RGB|RGBA|
-|-|-
|FF on Windows|6fps|60fps|
|FF on Mac|20fps|60fps|
|FF on Linux|14fps|14fps|
|FxR on Quest|35fps|35fps|

Firefox on Android: Could not test due to existing bug where videos don't render at all :( #4213

dezigns333:
The best video performance I have had was with the new Microsoft Edge
browser.
60 FPS even while screen sharing a video from a tab with either Youtube or
Netflix.
I was able to use VB-Audio virtual cable as a loopback to send the audio
through the mic so it was received perfectly in Hubs.

takahirox:
Thanks for the test! I would merge this PR.

dezigns333:
It works even better when Edge is capturing video from Firefox. Avoid
Chrome.

PR #4227, opened 2021-05-05
Disable chrome AEC hack by default

netpro2k:
This disables the workaround that was added to Chrome in https://github.com/mozilla/hubs/pull/2361 by default as it significantly reduces audio quality and largely breaks audio spatialization. It remains as a preference but we should plan to remove it once a fix lands upstream (which is slated for this year) https://bugs.chromium.org/p/chromium/issues/detail?id=687574

It is unclear if this was always an issue or something that changed in Chrome that caused this workaround to cause other issues... I am hoping for my own sanity that its the later, though it has been an issue since at least January since it was mentioned by a user on the original PR https://github.com/mozilla/hubs/pull/2361#issuecomment-766157504

Incidentally adds ability to the preference screen to have dependent properties:
image

### DIFF
--- a/src/assets/stylesheets/preferences-screen.scss
+++ b/src/assets/stylesheets/preferences-screen.scss
@@ -105,6 +105,13 @@ $semi-bold: 600;
   &:last-child {
     margin: 0px 16px 0px 16px;
   }
+  &.disabled {
+    opacity: 20%;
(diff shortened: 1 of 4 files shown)


johnshaughnessy (approved):
lgtm

PR #4235, opened 2021-05-06
"Change Scene" button is missing once the scene is set from a GLB

rawnsley:
Once a custom GLB URL is used to set the scene for a room, the scene cannot be set again to either an existing scene or another GLB URL.

This is particularly restrictive when rapidly iterating new scene files as you have to create a new room for each scene version.

**Initial Room Info Dialog**
Screenshot 2021-05-06 at 15 30 31

**Room Dialog After Using Custom Scene File**
Screenshot 2021-05-06 at 15 30 50

**Room Edit Dialog After Using Custom SceneFile**
Screenshot 2021-05-06 at 15 30 57

### DIFF
--- a/src/react-components/room/RoomSidebar.js
+++ b/src/react-components/room/RoomSidebar.js
@@ -80,7 +80,12 @@ function allowDisplayOfSceneLink(accountId, scene) {
 }
 
 export function SceneInfo({ accountId, scene, showAttributions, canChangeScene, onChangeScene }) {
-  if (!scene) return null;
+  const changeSceneButton = canChangeScene && (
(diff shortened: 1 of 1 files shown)


netpro2k:
Thanks! I definitely noticed that one when testing locally once but forgot to file it.

Softvision-GeluHaiduc:
I have verified this issue using Firefox 89.0 and Chrome 90.0.4430.212 on Windows 10.
Now, the "Change Scene" button is present for the rooms created with the GLB scenes. An interesting thing I notice is that the original issue is not present on Hubs Cloud.

PR #4247, opened 2021-05-07
Fix scene and media failing to load for local dev against hubs cloud

wmurphyrd:
See also #3972 

Two issues disrupting local dev against hubs cloud since February:

1. Assets are loaded via `https://${local ip address}:8080/...` instead of localhost. Browsers see this as a different domain and won't transfer your certificate exception from localhost, so the requests get rejected and you get "Scene failed to load"
2. Internal domains (AWS recipe 2) are not included in non-cors-proxy list, so they attempt to proxy and fail. Images and videos from the selfie cam just show as broken links

This fixes both for me, but I doubt it is the correct solution as it is just removing that `localIp` logic that must serve some purpose. Looking for guidance on what that purpose is to find a solution that works for both

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-802)

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -177,15 +177,14 @@ async function fetchAppConfigAndEnvironmentVars() {
     throw new Error(`Error fetching Hubs Cloud config "${hubsConfigsResponse.statusText}"`);
   }
 
-  const { shortlink_domain, thumbnail_server } = hubsConfigs.general;
+  const { shortlink_domain, thumbnail_server, non_cors_proxy_domains } = hubsConfigs.general;
 
-  const localIp = process.env.HOST_IP || (await internalIp.v4()) || "localhost";
-
+  const localIp = "localhost";
   process.env.RETICULUM_SERVER = host;
   process.env.SHORTLINK_DOMAIN = shortlink_domain;
   process.env.CORS_PROXY_SERVER = `${localIp}:8080/cors-proxy`;
   process.env.THUMBNAIL_SERVER = thumbnail_server;
-  process.env.NON_CORS_PROXY_DOMAINS = `${localIp},hubs.local,localhost`;
+  process.env.NON_CORS_PROXY_DOMAINS = `${non_cors_proxy_domains},${localIp},hubs.local,localhost`;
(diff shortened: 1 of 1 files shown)


yakyouk:
I think in here you can just comment the `internalIp` logic, keeping the ability to use HOST_IP:
(const localIp = process.env.HOST_IP || (await internalIp.v4()) || "localhost";)[https://github.com/mozilla/hubs/pull/4247/files#diff-1fb26bc12ac780c7ad7325730ed09fc4c2c3d757c276c3dacc44bfe20faf166fL182] 
Nice find on the proxying, been annoying me for a while.

yakyouk:
It doesn't seem to solve the issue of assets not loading for local development (`npm run start`) with Hubs Cloud:
The scene loads fine through the proxy ( https://localhost:8080/cors-proxy/https://internal-example-com-proxy.example.workers.dev/files/333ccccc-3333-4444-9999-26f26f26f26f.bin ). Other assets dropped into the room are using the proxy but never load; trying to fetch them unproxied from localhost:8080 will throw a CORS error: `Access to fetch at 'https://internal-example-com-proxy.example.workers.dev/files/01234567-ffff-4444-9999-cccfffddd777.png?token=1234567890' from origin 'https://localhost:8080' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.`.
Which is somewhat coherent with [this comment](https://github.com/mozilla/hubs/blob/41373e0078ce55eb09b6034fe2c621871c88ce96/src/components/media-loader.js#L373-L374) but it doesn't work.

procoder333:
I am trying to access with my IP address instead of localhost, but it occurs some errors in fetching.
So I can't join to the room with IP address.

netpro2k:
The "localIp" logic is used for accessing the local dev server via ip (such as on a mobile device). The intended way to add additional NON_CORS_PROXY_DOMAINs is to add them to a .env file. You will still need to manually make CORS/CSP changes in your hubs instance admin panel.

PR #4248, opened 2021-05-07
Allow spectate option when room is full

### LINKED ISSUE #4116: Spectate button not visible when room is full
With the current hubs (screenshot from hubs.mozilla.com today), opening a full room just shows an entry panel with no options and no explanation. Is this intended?

image

_Originally posted by @wmurphyrd in https://github.com/mozilla/hubs/discussions/4110_

I believe the issue is that `entryDisallowed` takes room occupancy / capacity into account. Spectate should be available even when the room is full.

https://github.com/mozilla/hubs/blob/41373e0078ce55eb09b6034fe2c621871c88ce96/src/react-components/ui-root.js#L806

wmurphyrd:
Fix #4116 - do not apply the full room logic to the spectate button


netpro2k:
Good catch. I think that was a regression from the new UI update.

Issue #4232, opened 2021-05-06
Camera videos recorded in Oculus Browser are blank


brianpeiris:
**Description**

Video clips do not get recorded correctly when using the Camera tool in Oculus Browser on the Quest. The recorded video is just  a blank black video. A warning in the console appears:

```
GL ERROR :GL_INVALID_OPERATION : glBlitFramebufferCHROMIUM: src framebuffer is multisampled, but src/dst regions are different
```

Which implies an issue with the blitFrameBuffer function: https://github.com/mozilla/hubs/blob/master/src/components/camera-tool.js#L68-L84

**To Reproduce**
1. Visit a room using the Oculus Browser on Quest
2. Spawn a camera
3. Record a video clip using the camera
4. See that the recorded video clip is blank

**Expected behavior**
Recorded video clips should not be blank.

**Hardware** 
 - Device: Oculus Quest
 - Browser: Oculus Browser

PR #4255, opened 2021-05-11
OB Camera multisampled FB blitting fix

keianhzo:
Closes https://github.com/mozilla/hubs/issues/4232

OB Chromium FBs are multisampled by default and they don't allow blitting for multisampled FBs when regions differ. https://chromium.googlesource.com/chromium/src/gpu/+/master/command_buffer/service/gles2_cmd_decoder.cc#8899
We are using `THREE.WebGLRenderTarget` so the render buffer shouldn't be multisampled but for some reason they are still forcing it. 

This PR avoiding blitting for OB and just flips the original camera render target pixels.

### DIFF
--- a/src/components/camera-tool.js
+++ b/src/components/camera-tool.js
@@ -39,8 +39,9 @@ const videoMimeType = videoCodec ? `video/webm; codecs=${videoCodec}` : null;
 const hasWebGL2 = !!document.createElement("canvas").getContext("webgl2");
 const allowVideo = !!videoMimeType && hasWebGL2;
 
-const CAPTURE_WIDTH = isMobileVR ? 640 : 1280;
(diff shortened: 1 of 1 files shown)


takahirox:
I think it would be better to ask Oculus Browser team about likely forced multisampling stuffs before going with this change.

keianhzo:
I agree, this was just a patch in case there wasn't a better short term alternative.

brianpeiris (approved):
Just to put it in writing here. This looks good to me as long as we're reasonably sure that it doesn't introduce an additional performance penalty. If we can find a more generalized fix in the future, that would also be good, but it seems fine to ship this fix for now. We can always improve on it.

keianhzo:
Ok, I'm going to land this to get it fixed until we can update ThreeJS.

PR #4261, opened 2021-05-14
Moveable media defined in the scene file

rawnsley:
This PR introduces the idea of `moveable` elements that can be defined in the scene file using an extension, which increases the interactivity of scenes and opens up many interesting use cases such a puzzles, scavenger hunts, disassembly/assembly tasks, board games, etc...

There is a live demo on my Hubs Cloud here: https://vrroom.us/hiW-3Os/moveable-example

https://user-images.githubusercontent.com/303516/118294047-3a596100-b4d2-11eb-88d0-1fad431ce09e.mp4

https://user-images.githubusercontent.com/303516/118294068-404f4200-b4d2-11eb-860f-b0ae472b936a.mp4

https://user-images.githubusercontent.com/303516/118294087-45ac8c80-b4d2-11eb-8c46-046e7f593bd7.mp4

The source Blender file for these video examples [is here](https://data.avncloud.com/activities/730777/files/moveable.blend?date=1621005004&size=45522564&md5=a115a8249f836691700a49834ca1677d) and the exported GLB file [is here](https://data.avncloud.com/activities/730777/files/%21moveable.glb?date=1621004990&size=22111904&md5=2d7f574da4cc72e4c0d40dea3ec79f62). Export relies on the addition of `model` and `moveable` components in [this Blender exporter PR](https://github.com/MozillaReality/hubs-blender-exporter/pull/33).

**Implementation Notes**

Adding the `moveable` component at any point in the scene graph makes that node moveable. If it's already a media component then the _interactive-media_ template is used, otherwise a new, simpler, _moveable-media_ template is used.

The puzzle use case exposed some issues with the way media-frame behaves when the frames are contiguous. I modified the code to prevent multiple frames accepting media in a single tick. I also made some changes to the way the scale fitting works to be better behaved with 3D models.

One limitation is that the code only generates physics hulls and the hover animation for nodes with single meshes. I could not understand the interplay between A-Frame, three.js, and ammo.js well enough to work out how to make it work. I know it's possible because that's it happens to models when they are dragged in at runtime.

I've kept the actual code changes relatively small, but my limited understanding of the Hubs codebase means there will be cases I have missed and improvements that can be made. I'm hoping that the use cases are compelling enough for one of the devs to help take it forward.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-765)


johnshaughnessy:
Hi @rawnsley

This PR has been here for a long time without a response from the Hubs team. It has been difficult for us to decide what to do with it. I'd like to acknowledge the effort you put in here, apologize for the delay, and explain how I've been thinking about it.

The feature itself is clearly useful. I suspect that if we support it, many people will be happy that it exists and will immediately start using it.

If people are going to use the feature, I want to make sure we implement it in a way that makes it unlikely for us to break their scenes in the future. Once it's out there, hopefully we won't have to change it too much.

Here are some challenges / questions we face in trying to get it merged.

[ Design Questions]
- How should scene-defined moveable objects behave?
  - Should these objects have menus?
  - Should they appear as entries in the objects menu? 
    - If not, what is the user's course of action if they misplace the object (in a wall, in the floor, etc).
  - Should users be able to delete them?
  - Is there a way to reset them?
  - How do we indicate the difference between a scene-defined moveable object and a "typical" room object to the end user?
  - Should users expect changes to the objects to persist between page-reloads? Can users "pin" these objects?
    - Right now, the changes only persist until the last person leaves the room. If the room is empty, all the previous state is lost. How do we communicate this expectation to users?
    - We could persist changes to these objects into the room's objects.gltf file, but we'd need to add logic to clear out or ignore that data if the room's scene changes.
    - We could add a new concept to reticulum that is something like, "server-backed object state" where the "object" is unique across scene+rooms. Then, if the scene changes, clients will be told to stop caring about that object. If the scene changes back, should clients care about that object again, or should those objects be deleted? 
  - If state is persisted on the server, should the user have to log in to move the objects?
    - Our current contract is basically that users must log in create persistent state on the server. That's why you have to log in to move pinned objects. The answer here might be, "no, they should not have to", but it's not clear to me.

[ Engineering Concerns ]
- As mentioned above, we have not laid the groundwork for "persistent object state that is tied to scenes + rooms". We also need to decide how we model "objects within objects" so that if you load a gltf file where some of the nodes inside it are "moveable", then delete the "parent object", all the persistent state for the children goes away too.
- This PR adds more code to the already-quite-confusing-and-brittle "networked, physics-enabled, scene-owned, semi-persistent, object loading path". This includes the "scale when loaded" code, the "create physics hull on load" code, the "move the root of the object to the center of the mesh" code, the "use this html template as a kind of prefab" code, and I'm sure I'm forgetting more. We would really like to simplify this, and the more code that depends on its quirky way of doing things, the harder it is to transition away from it.
- This PR adds another descriptive term ("moveable") to the already-long list of "things that affect interaction behavior" components and properties. This list includes everything in the client with the word "interactable" or "hoverable" or "targettable", the "tags" component properties, the "physics+constraints" stuff, "floaty-object", etc. This may be the same concern as the previous point. There's just a lot of code here that we don't feel very confident in. We'd like to change how some of it works in the near-ish future.
- Random parts of the client code assume that when you load an object, you're loading one "conceptual thing". An example would be how you had to change the nav mesh loading code to ignore "moveable" objects. I think we need to think more carefully about what the client should do with gltf models that are multiple "conceptual things".
- A related assumption is that the local space of networked objects are aligned with the global space. I don't think there is a particular reason for this other than that it was simple to write code with this assumption and it worked. If deeply-nested nodes in the gltf are allowed to become independently-networked nodes, some random parts of the client might do the wrong thing.

You successfully found a path through these concerns that met your needs. I ran this code and checked out the blender plugin and the example scene and it works great. After playing around with it a bit, I wrote some code that would reset objects to their initial positions if they strayed too far (1000 meters in any direction). I fixed some merge conflicts. Then I tried to re-implement the behavior "from scratch" to see what problems came up when I tried doing this with a more "naive" approach. (Spoiler alert -- it did not go well.)

I think what we should do is simplify the code that syncronizes object state across the network, simplify the loading code, simplify the interaction logic, simplify the creation and management of physics hulls, and think about how to add server-backed state for particular objects. Somewhere along that path, this feature should be easier and more straight-forward to implement, and hopefully make the design questions a little easier to answer.

I don't like this suggestion very much, because it feels disappointing not to merge this PR or find a way to get this feature into the core now, but that's my current recommendation.

rawnsley:
@johnshaughnessy Thanks for the detailed feedback. This is an unusual external PR and I didn't really expect it to be accepted "as-is" because of the complex issues you raise.

We developed the feature in response to our content team wanting "more interactivity" within scenes and although we have yet to actually take advantage of it I believe there is clear value. However, it's arguably a niche requirement that will only end up benefiting a minority of Hubs content creators, so I understand why you don't want to just merge-and-be-damned.

I submitted it as a PR for partly selfish reasons: firstly any PR you accept makes it easier for me to merge with your master branch, but secondly it is complex functionality and a maintenance burden that I would prefer to share. I will be happy to discard this PR in favour of a comprehensive and sustainable approach even if it means waiting a while for some more fundamental changes.

Regarding some of the specific points you raise: I feel a context menu and inclusion in the object list is unnecessary as "movable" elements are part of the scene. This of course leads to some the problems you describe. RecRoom deals with this sort of thing by having a configurable timer to replace objects in their starting position that is reset whenever an object is touched. This is at least logically consistent with the room reseting when everyone leaves. Differentiating between "moveable" and "interactable" objects for the user is harder; a different hover color would be a start (white perhaps rather than HTML anchor blue?), but the meaning would not be obvious.

Some of these design decisions may make themselves if we look forward a little more. A couple of interesting ideas suggested themselves once I had got this far. The first is that "moveable" is the logical component to store physics properties such as mass and friction. The second is that media frames are a blunt tool for reassembling models with multiple "moveable" parts such as engines or skeletons, which lead me to the idea of "snap-points". These would be empty GLTF nodes with a snap-point component that can have in and out string tag filters to allow for precise assembly (_the hip bone's connected to the leg bone_). In fact the snap-point component may also be the right place to store physics joint properties, but I haven't used ammo.js enough to be sure. If this approach worked it would allow us to take a car model from SketchFab, make the wheel joints "spin" and roll it down a hill or to take a skeleton model and dangle it from a rope.

There are a couple of parts of the PR that might be worth considering independently:

1. Loading models from URLs as we can with other media, which allows for a useful level of indirection in the scene file.
2. Fixes to adjacent media frames such as grids. Currently objects scoot about unpredictably.

If we close this one off for now I might resubmit those as independent PRs.

johnshaughnessy:
@rawnsley Regarding the 3D object scaling, it looks like @keianhzo has a fix for this in review. See https://github.com/mozilla/hubs/pull/4655

netpro2k:
As discussed we won't be landing this. The media frame system has since been rewritten which should account for the fixes here, and the new entity system should allow for the flexibility needed to make individual sub-parts of objects movable and networkable (thanks certainly in part to this PR as an existence proof)

Issue #3149, opened 2020-10-09
Waypoint link reloads room on first click


johnshaughnessy:
**Description**
See https://github.com/mozilla/hubs/discussions/2983

**To Reproduce**

**Expected behavior**

**Screenshots**

**Hardware** 
 - Device: [e.g. Desktop, phone, VR headset]
 - OS: [e.g. Windows, iOS, Linux]
 - Browser [e.g. Firefox]

**Additional context**

emclaren:
Just wanted to note that this is an issue that was also mentioned by a community member in an email last week

oli414:
I've been able to find a pattern with this issue. 

Our scene appears to have working teleports, but as soon as a user spawns an object in the scene, the teleports instead reload the page. I've been able to reproduce the issue on both Chrome and Firefox so I think it's fair to say that this might somehow be the cause of the issue.

Charlesc22:
We've also seen this bug, but have not drilled down to find the pattern. will check for the forward slash at the end though
Thanks for the tip @oli414

Exairnous:
From testing I've done it appears that links to a waypoint will reload the room if the link address does not match the current room URL regarding the trailing slash.  This is why adding an object (or changing avatars, etc.) will cause links to reload the room if a trailing slash is not included, because opening the object place menu will add a trailing slash to the current room URL (but the link URL doesn't contain a trailing slash).  The room will also be reloaded if the waypoint link URL contains a trailing slash, but the room URL does not (they must match or the room will reload).

The only way I've found to reliably work with waypoint links is to setup a link (or image with href) to contain the full room address, including a trailing slash, then the #WAYPOINTNAME, and then give everyone the room link with the trailing slash (potentially including query string parameters after the slash); the downside with this is that the favorite room links don't include the trailing slash, so if you visit a room from your favorites menu with waypoint links you'll run into the reloading problem :(

Example links:
WORKS ALL OF THE TIME:
waypoint link: https://hubs.mozilla.com/<room-code>/<room-name>/#WAYPOINTNAME
room link: https://hubs.mozilla.com/<room-code>/<room-name>/

WORKS SOME OF THE TIME (will work until some dialog changes the room URL to contain a slash):
waypoint link: https://hubs.mozilla.com/<room-code>/<room-name>#WAYPOINTNAME
room link: https://hubs.mozilla.com/<room-code>/<room-name>

NEVER WORKS:
[...]

PR #4272, opened 2021-05-20
#3149 fix waypoint link reloads room

yakyouk:
Fixes links pointing to waypoint hashes reloading the room.
Clicking on a link with a hash does not compare full urls anymore, instead it checks if the link's domain and room ID match those of the current room. As a side effect, waypoint links now also work on a local client.

Fixes back button sending to last waypoint: when using a waypoint link, the url with hash and the url with cleared hash was added to history. Now uses replaceState() instead.

Fixes RegEx for room/avatar/scene links to NOT match when IDs have more than 7 letters/numbers.

### DIFF
--- a/src/components/open-media-button.js
+++ b/src/components/open-media-button.js
@@ -57,7 +57,8 @@ AFRAME.registerComponent("open-media-button", {
       } else if ((hubId = await isHubsRoomUrl(this.src))) {
         const url = new URL(this.src);
         if (url.hash && APP.hub.hub_id === hubId) {
-          location.hash = url.hash;
+          // move to waypoint w/o writing to history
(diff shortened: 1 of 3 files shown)


netpro2k:
I don't love that `isHubsRoomUrl` now returns a non-boolean value (just because of the name) but this is a really impactful change. Thanks for the PR!

yakyouk:
> 
> 
> I don't love that `isHubsRoomUrl` now returns a non-boolean value (just because of the name) but this is a really impactful change. Thanks for the PR!

True, the rest of the code relies on it being truthy/falsy, so to prevent things from breaking if return value changes in the future, it would be cleaner to use a new function `idForRoomUrl`.

PR #4278, opened 2021-05-21
Audio zone fixes

### LINKED ISSUE #4276: Audio Source Components Do Not Move With Objects
**Description**
When adding a GLB with an audio source component the component stays at it's initial coordinates regardless of where the object is moved to in a scene/room.

**To Reproduce**
1. Export a GLB with mic + audio source component and speaker + audio target component. ([Mic_Speaker_test.zip](https://github.com/mozilla/hubs/files/6518138/Mic_Speaker_test.zip))
  Note: make sure to remove Mute self in the audio source and enable debug in both the audio source and audio target (done in test file).
2. Add it to a Spoke scene or drop it into a Hubs room.
3. Move it somewhere away from the world origin.
4. Go to the mic in the room and try to talk (you won't be amplified)
5. Go to the initial location of the audio source in the GLB file (0.9811 m on the blender Y axis from world origin) and try to talk (you will be amplified)

**Expected behavior**
To be amplified from the new position in the scene/room that the audio source has been moved to.

**Hardware**
 - Device: Laptop
 - OS: Linux
 - Browser: Chromium

**Additional context**
The audio target component appears to work correctly and move with the object.

netpro2k:
- Fixes #4276 by taking into account the world space position of zone audio sources when doing distances checks
- Uses the new `amplify_audio` permission added in https://github.com/mozilla/reticulum/pull/499 to enforce `modOnly` on zone audio sources. Previously this was using "owner" but this does not work well for bound rooms.
- Fixed an unrelated bug I noticed in hub-channel.js that was incorrectly concatenating arrays using the + operator.. It turned out this mostly just "worked" anyway since the arrays were contacted into a big string, and strings also have an `includes` method which happened to also be correct for this case. Oh javascript.

### DIFF
--- a/src/components/player-info.js
+++ b/src/components/player-info.js
@@ -128,7 +128,7 @@ AFRAME.registerComponent("player-info", {
     this.applyDisplayName();
   },
   can(perm) {
-    return !!this.permissiosn[perm];
+    return !!this.permissions && this.permissiosn[perm];
   },
   applyDisplayName() {
     const store = window.APP.store;
--- a/src/components/player-info.js
+++ b/src/components/player-info.js
@@ -120,11 +120,15 @@ AFRAME.registerComponent("player-info", {
     this.updateDisplayNameFromPresenceMeta(e.detail);
   },
   updateDisplayNameFromPresenceMeta(presenceMeta) {
+    this.permissions = presenceMeta.permissions;
     this.displayName = presenceMeta.profile.displayName;
     this.identityName = presenceMeta.profile.identityName;
     this.isRecording = !!(presenceMeta.streaming || presenceMeta.recording);
     this.isOwner = !!(presenceMeta.roles && presenceMeta.roles.owner);
     this.applyDisplayName();
+  },
+  can(perm) {
+    return !!this.permissiosn[perm];
   },
   applyDisplayName() {
     const store = window.APP.store;
--- a/src/components/player-info.js
+++ b/src/components/player-info.js
@@ -62,7 +62,7 @@ AFRAME.registerComponent("player-info", {
         this.playerSessionId = NAF.utils.getCreator(networkedEntity);
         const playerPresence = window.APP.hubChannel.presence.state[this.playerSessionId];
         if (playerPresence) {
-          this.updateDisplayNameFromPresenceMeta(playerPresence.metas[0]);
+          this.updateFromPresenceMeta(playerPresence.metas[0]);
         }
       });
     }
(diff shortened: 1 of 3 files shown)


src/components/player-info.js
```diff
@@ -120,12 +120,16 @@ AFRAME.registerComponent("player-info", {
...
   },
   updateDisplayNameFromPresenceMeta(presenceMeta) {
+    this.permissions = presenceMeta.permissions;
     this.displayName = presenceMeta.profile.displayName;
     this.identityName = presenceMeta.profile.identityName;
     this.isRecording = !!(presenceMeta.streaming || presenceMeta.recording);
     this.isOwner = !!(presenceMeta.roles && presenceMeta.roles.owner);
     this.applyDisplayName();
   },
+  can(perm) {
```
brianpeiris, line 130: The `can` function in `hub-channel.js` has a nice `VALID_PERMISSIONS` check that we might want to use here as well, if we continue to use this function more widely.

src/utils/hub-channel.js
```diff
@@ -22,11 +22,18 @@ const HUB_CREATOR_PERMISSIONS = [
...
-  ["tweet", "spawn_camera", "spawn_drawing", "spawn_and_move_media", "pin_objects", "spawn_emoji", "fly"];
+const VALID_PERMISSIONS = HUB_CREATOR_PERMISSIONS.concat([
+  "tweet",
+  "spawn_camera",
+  "spawn_drawing",
+  "spawn_and_move_media",
+  "pin_objects",
+  "spawn_emoji",
+  "fly"
+]);
```
brianpeiris, line 36: Oh wow, nice catch 😅

brianpeiris (changes requested):
Spotted a typo that needs fixing. Other minor comments, but looks good otherwise.

src/utils/hub-channel.js
netpro2k: Yeah at first I was like "Wait, js has an array concatenation operator? And then I realized what was going on... Oh javascript. Funny that it just happened to work mostly correctly by chance.

yakyouk:
> 
> 
>     * Fixes #4276 by taking into account the world space position of zone audio sources when doing distances checks
> 
>     * Uses the new `amplify_audio` permission added in [mozilla/reticulum#499](https://github.com/mozilla/reticulum/pull/499) to enforce `modOnly` on zone audio sources. Previously this was using "owner" but this does not work well for bound rooms.
> 
>     * Fixed an unrelated bug I noticed in hub-channel.js that was incorrectly concatenating arrays using the + operator.. It turned out this mostly just "worked" anyway since the arrays were contacted into a big string, and strings also have an `includes` method which happened to also be correct for this case. Oh javascript.

Ah good to see it's fixed, noticed that a year ago and I was just about to push a PR for that XD

src/utils/hub-channel.js
yakyouk: Ah good to see it's fixed, noticed that a year ago and I was just about to push a PR for that!
I wondered too at first if I was missing some obscure syntax  XD

Issue #4102, opened 2021-03-31
Media frames sometimes greyed out wont accept media


BonzoMartini:
**Description**
Media frames sometimes get greyed out wont accept media until, the Hub is reloaded

**To Reproduce**

1. in hubs import images and stick em into media frames
2. Sometimes all media frames go grey and won't accept media
3. The hub needs to be reloaded to fix

**Expected behavior**
User should be able to import media and stick it into media frames regardless of whether media frames have content if that content is not pinned.

**Screenshots**
![media frame bug grey mode](https://user-images.githubusercontent.com/66266175/113188628-2fb07880-9252-11eb-904d-2b4c123176f2.png)
This image shows the bug..
the second image shows after the hub has been reloaded - frames are all blue

![medaiframes after refresh](https://user-images.githubusercontent.com/66266175/113188788-65556180-9252-11eb-9dc2-3611e8ed1ab3.png)

**Hardware** 
Desktop, Pc Firefox

**Additional context**
This gives a bad experience for users who don't know whats happening

camelgod:
You think this could be related? https://github.com/mozilla/hubs/issues/4089

keianhzo:
Not related to the world pos issue, but I was a was able to reproduce, it seems to happen when an object with a larger scale than the media.frame is placed.

https://user-images.githubusercontent.com/837184/119980035-07c46380-bfbc-11eb-9e4c-0b278cb7acc4.mp4

Self ssigning for review.

rawnsley:
I've addressed this issue as part of this [moveable media PR](https://github.com/mozilla/hubs/pull/4261). This is the most relevant commit: https://github.com/mozilla/hubs/pull/4261/commits/a7a9a65d107e6c4e51cd836d795830477fbe5e05

The scale issue could be that the models are currently only constrained in two dimensions, because that was all that was required for 2D media. I've also corrected for that in the PR.

keianhzo:
Great, thanks for that PR, I'll take a look at it and hopefully we can also get the related issues solved.

Issue #4089, opened 2021-03-26
Media zones will sometimes make the object inserted enormous


camelgod:
**Description**
When adding an object to a media zone, the action results in the following:
- All nearby media zones are occupied as well (greyed out, no actual object inside)
- Removing the object from the original media zone will scale it so it covers the entire world and is impossible to delete

**To Reproduce**
Put a bunch of media zones together
Place object

Happens very inconsistently so its hard to reproduce, but i have gotten the bug to consistently happen when using 3d objects that have less than 1,1,1 scale, most likely the algorithm calculating new object size will go nuts. It also happens using images, often with our custom "post it note" feature but also other images.

**Expected behavior**
The object should not occupy other nearby media zones, and should pop out to its normal scale upon removal.

**Hardware** 
 - Device: Desktop
 - OS: MacOS, Windows, Linux
 - Browser: Chrome, FF

keianhzo:
Seems not to be related to the world space coordinates issues and happens pretty consistently, self assigning.

keianhzo:
From my tests and before a deeper inspection, this is probably related to https://github.com/mozilla/hubs/issues/4102

PR #4279, opened 2021-05-21
Place media frame content in world space coordinates

keianhzo:
We are currently placing elements in `media-frame`s based on local coordinates so nested `media-frame`s don't place them correctly.

Test scene: https://hubs.mozilla.com/scenes/4zXCrZE

### DIFF
--- a/src/systems/media-frames.js
+++ b/src/systems/media-frames.js
@@ -135,6 +135,7 @@ AFRAME.registerComponent("media-frame", {
   },
 
   init() {
+    this.tmpWorldPosition = new THREE.Vector3();
     //TODO these visuals need work
     this.el.setObject3D(
       "guide",
@@ -249,12 +250,9 @@ AFRAME.registerComponent("media-frame", {
   },
 
   snapObject(capturedEl) {
-    // TODO this assumes media frames are all in world space
-    const worldPosition = new THREE.Vector3();
-    this.el.object3D.getWorldPosition(worldPosition);
-    capturedEl.object3D.position.copy(worldPosition);
+    this.el.object3D.getWorldPosition(this.tmpWorldPosition);
+    capturedEl.object3D.position.copy(this.tmpWorldPosition);
     const worldQuat = new THREE.Quaternion();
-    this.el.object3D.updateWorldMatrix(true);
     this.el.object3D.getWorldQuaternion(worldQuat);
     capturedEl.object3D.setRotationFromQuaternion(worldQuat);
     capturedEl.object3D.matrixNeedsUpdate = true;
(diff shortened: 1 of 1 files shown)


src/systems/media-frames.js
```diff
@@ -250,8 +250,13 @@ AFRAME.registerComponent("media-frame", {
 
   snapObject(capturedEl) {
     // TODO this assumes media frames are all in world space
-    capturedEl.object3D.position.copy(this.el.object3D.position);
-    capturedEl.object3D.rotation.copy(this.el.object3D.rotation);
+    const worldPosition = new THREE.Vector3();
```
netpro2k, line 253: Not so bad since this isn't run every frame but we should avoid allocating these vectors every time, either by just pulling it out into the root of the file, or in a closure for this function alone. I will often just create a `const tmpWorldPos = new THREE.Vector3()` in a file to use for this sort of thing.

src/systems/media-frames.js
```diff
@@ -250,8 +250,13 @@ AFRAME.registerComponent("media-frame", {
...
+    this.el.object3D.getWorldPosition(worldPosition);
+    capturedEl.object3D.position.copy(worldPosition);
+    const worldQuat = new THREE.Quaternion();
+    this.el.object3D.updateWorldMatrix(true);
```
netpro2k, line 257: You should not need to manually call this, and in fact we don't want to because it will cause extra work. See https://github.com/mozilla/hubs/blob/master/src/utils/threejs-world-update.js#L22. Also worth reading through these rules and this file in general if you havent yet. Not following these rules is source of most of our hubs-specific threejs issues... I do hope we can get rid of it some day, but when we implemented it there was a big performance win, especially due to all the extra Object3Ds aframe introduces into the scene graph.

src/systems/media-frames.js
```diff
@@ -250,8 +250,13 @@ AFRAME.registerComponent("media-frame", {
 
   snapObject(capturedEl) {
     // TODO this assumes media frames are all in world space
```
netpro2k, line 252: This PR fixes this todo so we should remove it.

netpro2k (commented):
lgtm, some minor comments. I probably should have just done this originally instead of making a TODO heh. If I remember correctly I was fighting some annoying physics system bugs at the time so was trying to keep things as simple as possible.

src/systems/media-frames.js
keianhzo: Yeah, I thought that being non a tick function would be fine but it makes sense to avoid instancing if possible. Updated.

src/systems/media-frames.js
keianhzo: Humm, I didn't know about that file. I've remove the `updateWorldMatrix` call.

src/systems/media-frames.js
keianhzo: Updated.

keianhzo:
@netpro2k Updated.

camelgod:
Will this affect these issues at all?

https://github.com/mozilla/hubs/issues/4102
https://github.com/mozilla/hubs/issues/4089

PR #4285, opened 2021-05-23
Fix showing create room "more" menu option on servers with it disabled

wmurphyrd:
Checking truthiness of method rather than its return value means `canCreateRoom` would always be truthy


netpro2k:
Woops... Another loljavascript one. Thanks for the fix!

PR #4304, opened 2021-06-02
Feature/tokens page styling

robinkwilson:
View components for the /tokens page that manages API tokens for the hubs API. Functionality will be integrated in a later PR.

TODO: 
- The centered modal does not scroll when the height is too small


src/react-components/input/CheckboxInput.js
```diff
@@ -0,0 +1,30 @@
...
+import classNames from "classnames";
+import styles from "./CheckboxInput.scss";
+
+export const CheckboxInput = memo(
+  forwardRef(({ className, labelClassName, label, description, disabled, ...rest }, ref) => {
+    return (
+      <label className={classNames(styles.checkboxInput, { [styles.disabled]: disabled }, className)}>
+        <input type="checkbox" disabled={disabled} ref={ref} {...rest} />
+        <div className={classNames(styles.checkmark)} />
+        {label && (
```
johnshaughnessy, line 12: I wonder if this label ought to be its own component, such that this becomes a `LabeledCheckboxInput` that combines a `Label` and a `CheckboxInput`. 

Maybe not worth breaking this down unless we wanted to re-use the "label" portion.

src/react-components/layout/Row.js
```diff
@@ -0,0 +1,61 @@
...
+  spaceBetween,
+  breakpointColumn,
+  topMargin,
+  flexClassName,
+  flexBasis,
+  noWrap = false,
+  childrenMarginR,
+  className
+}) => {
+  const paddingClass = padding === true ? styleUtils.lgPadding : styleUtils[`${padding}Padding`];
```
johnshaughnessy, line 20: This pattern where a `prop` can either be `true` or one of (`"lg"`, `"sm"`, `"md"`, etc) is strange to me. If the purpose is to provide a default value for `padding`, should we use [`defaultProps`](https://reactjs.org/docs/typechecking-with-proptypes.html#default-prop-values)?

src/react-components/tokens/RevokeTokenModal.js
```diff
@@ -0,0 +1,66 @@
...
+import styles from "./Tokens.scss";
+import styleUtils from "../styles/style-utils.scss";
+import { Button } from "../input/Button";
+import { Row } from "../layout/Row";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { faExclamationTriangle } from "@fortawesome/free-solid-svg-icons/faExclamationTriangle";
+import { faTimes } from "@fortawesome/free-solid-svg-icons/faTimes";
+import { Column } from "../layout/Column";
+
+export function RevokeTokenModal({ onClose, revoke }) {
```
johnshaughnessy, line 14: To follow the naming convention, should this `revoke` be called something like `onRevoke`?

johnshaughnessy (approved):
lgtm!

src/react-components/input/CheckboxInput.js
robinkwilson: Good point for reuse on labels, I'll keep that in mind for future inputs (then redo RadioInput, CheckboxInput etc) -- but keeping this as is for now since it follows the RadioInput component standard.

src/react-components/layout/Row.js
robinkwilson: I also don't like this pattern at all -- I was following other components that do this. I'll change it to something that is better :P

src/react-components/tokens/RevokeTokenModal.js
robinkwilson: Yes

src/react-components/layout/Row.js
robinkwilson: Since this pattern is already used in Row and other components, I'm going to merge this one as is, then make another PR where gap / padding for Row and Column are strings only.

takahirox:
@robinkwilson I would like to advice you that you should run `npm run test`  before making a PR. I'm facing some lint errors which seem to be caused by this PR.

PR #4306, opened 2021-06-02
Feature/remove google poly

robinkwilson:
Removed all references to google poly because of deprecation.
1/3 PRs


johnshaughnessy (approved):
Thank you for doing this! Removed a few more references. Looks good to me!

Issue #4108, opened 2021-04-01
Replace our Three.js fork to the official one


takahirox:
We Hubs team have [our Three.js fork](https://github.com/MozillaReality/three.js) but maintaining it is costly and sometimes it is hard to bump into the upstream Three.js.

So we have been wanting to replace our fork to the official one by removing our specific code from Three.js as much as possible.

As far as I know the left tasks for the replacement are

Hubs (and Spoke?)
- [ ] How to keep WebVR support. Three.js has dropped WebVR support and focuses on WebXR. But some major web browsers don't support WebXR (by default) yet, eg. Firefox. So we want to keep WebVR support. Probably importing [webxr-polyfill](https://github.com/immersive-web/webxr-polyfill) may be a solution?
- [x] Rewrite our special glTF extension handlers in Hubs glTF loader to the ones with the Three.js glTF loader plugin system #4117
- [x] Multiview Support. Three.js has dropped multiview support because anti-aliasing with multiview is still unusable and it's useless. Is it ok for us to drop the support, too?
- [x] Remove regular `Geometry`. Three.js has dropped regular `Geometry` and it supports only `BufferGeometry`. If we use `Geometry` we need to replace it with `BufferGeometry`. At least it seems we need to update [three-pathfinding](https://github.com/donmccurdy/three-pathfinding/commit/b458c4087eb83bff11ecfff6539b2e2cbb3ae472)
- [x] Figure out what patches we have applied to our Threejs fork
- [ ] Figure out which patches we need/want to keep
- [x] Figure out the easy way to apply our patches (the ones we want to keep or future ones) to Three.js https://github.com/MozillaReality/three.js/wiki/Modifications-on-top-of-three.js Thanks @brianpeiris 

Spoke
- [ ] UBO support. [three-batch-manager](https://github.com/mozillareality/three-batch-manager), used in Spoke (and future Hubs?), uses UBO but Three.js UBO PR has not been merged yet. And it seems we added some changed from the PR.
- [ ] Rewrite the special loader/exporter for Spoke to the ones with the Three.js glTF loader/exporter plugin system (if possible)

Related: https://github.com/MozillaReality/three.js/pull/47

Feel free to update the task list if you have anything else.

brianpeiris:
Here's a link to my documentation about the patches we've applied in our fork: https://github.com/MozillaReality/three.js/wiki/Modifications-on-top-of-three.js

takahirox:
[...]

Issue #21, opened 2017-11-21
zombie avatars remaining on leave


gfodor:
when users leave the room, sometimes their  avatars remain. likely call stack: https://mozilla.slack.com/files/U723XJLDP/F835ECFCY/image.png

brianpeiris:
I haven't seen this lately. We can track it again in Jira if it becomes a problem.

gfodor:
I def saw this a few times today sadly :(

mauriciomartinez-itesm:
This still happens sometimes

Issue #3357, opened 2020-11-17
[New UI] The “!” sign is shown at first when creating a react emoji


Softvision-GeluHaiduc:
**[Affected Versions]:**
- Firefox Release 82.0.3
- Chrome 86.0.4240.198

**[Affected Platforms]:**
- Windows 10x64

**[Prerequisites]:**
- Have a new room created and entered.

**[Steps to reproduce]:**
1. Click the “React” button from the bottom menu.
2. Click one of the react emojis.
3. Observe the top of the emoji created.

**[Expected results]:**
- The react emoji is displayed and smaller emojis are being generated from the top.

**[Actual results]:**
- The react emoji is displayed and “!” signs are being generated from the top.

**[Notes]:**
- After a few seconds the correct emojis are generated from the top.
- Attached a screen recording of the issue below.
![react_emoji](https://user-images.githubusercontent.com/51400741/99388628-8d18de00-28de-11eb-9b30-a4b1f6dee3dc.gif)

rawnsley:
This is happening for me in Chrome on MacOS. It looks like a CORS error. Here's the developer console output:

Access to fetch at 'https://assets-prod.reticulum.io/hubs/assets/images/emojis/emoji_3-c6f4eb89cc99bcb1977111cb56905e9c.png' from origin 'https://hubs.mozilla.com' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
raven.js:1475 GET https://assets-prod.reticulum.io/hubs/assets/images/emojis/emoji_3-c6f4eb89cc99bcb1977111cb56905e9c.png net::ERR_FAILED
(anonymous) @ raven.js:1475
load @ three.module.js:41888
(anonymous) @ HubsTextureLoader.js:2
(anonymous) @ HubsTextureLoader.js:2
i @ HubsTextureLoader.js:39
a @ hub-05ccffcbb2e27e47ddae.js:1
s @ hub-05ccffcbb2e27e47ddae.js:1
(anonymous) @ hub-05ccffcbb2e27e47ddae.js:1
(anonymous) @ hub-05ccffcbb2e27e47ddae.js:1
loadTextureAsync @ hub-05ccffcbb2e27e47ddae.js:1
n @ particle-emitter.js:62
he @ hub-05ccffcbb2e27e47ddae.js:1
[...]

PR #4321, opened 2021-06-09
Bump Three.js into r128

takahirox:
Fixes: #4108

Requires: https://github.com/MozillaReality/aframe/pull/30
Requires: https://github.com/InfiniteLee/three-to-ammo/pull/25
Requires: https://github.com/InfiniteLee/buffered-interpolation/pull/7
Requires: https://github.com/MozillaReality/three-bmfont-text/pull/1
Requires: https://github.com/InfiniteLee/three-ammo/pull/22
Requires: https://github.com/MozillaReality/lib-hubs/pull/30

**Description**

This PR upgrades Three.js into r128. Currently the latest revision is r129 but the revision was r128 when I started to work on this. We can upgrade to r129 later. It should be much easier than upgrading from r111 to r128.

Please refer to #4108 for the purpose, benefits, and changes.

**Notes**

* r128 Three.js dropped deprecated WebVR API support. I import [`webxr-polyfill`](https://github.com/immersive-web/webxr-polyfill) from `hubs.js` for platforms supporting only WebVR e.g. Desktop Firefox + Wired VR device. I don't have such devices. If you have, would you please test to check if it works?

* I didn't apply this change so far because I don't see any errors so far. I guess it's from batching manager system or Three.js revision older than r128? Let me know if you encounter any shader material problems.

> MozillaReality/three.js@d83c538 Fix defines in RawShaderMaterial

* I dropped the change for Three.js `TextureLoader` so far. Instead I refactored `HubsTextureLoader` to follow the Three.js common Loader API and let Hubs source code explicitly use `HubsTextureLoader` rather than hacking Three.js `TextureLoader` for `ImageBitmap`. Only the place which uses Three.js `TextureLoader` I noticed is A-Frame material system. Do you  folks know what the system is for? And should we use `ImageBitmap` for A-Frame material system by hacking `TextureLoader` although we need to sacrifice the maintenance cost?

* I disable to use `ImageBitmap` for Firefox. Background is https://github.com/mrdoob/three.js/pull/21336. createImageBitmap() with option fails on Firefox due to the bug. r128 Three.js `ImageBitmapLoader` passes `{colorSpaceConversion: 'none'}` option. Without the option, the rendering result can be wrong if image file has ICC profiles. We may have an option to hack `ImageBitmapLoader` for Firefox not to pass the option. It's the tradeoff between rendering result vs maintenance cost. If we want to enable `ImageBitmap` for Firefox, I want to carefully change it for example researching how many ICC profiles use cases (I guess it may be rare tho) and want to do in another PR.

* It seems webpack tree shaking removes the modules under the r128 `three/examples/js` if I use `import`. Somehow it's resolve by using `require` instead. There might be better solutions but I use `require` in the PR so far.

* We disabled Batching manager system but it imports dependencies which break the build with Three.js r128. Then I replaced Batching manager system with a stub so far. We can revert it when we will enable it again.

* It seems this change is for Oculus browser specific performance problem. There is a chance that the problem has been already resolved. So I dropped this change so far. Let me know if you still see performance problem. We may be able to upload at the start with `renderer.initTexture()` without hacking Three.js.

> MozillaReality/three.js#32 Add option to upload videos at the start render

* Regarding the `flipY` + `PlaceBufferGeometry`, I made a utility function and let Hubs use it instead of hacking Three.js.

> MozillaReality/three.js#21 Prefer ImageBitmapLoader and dispose of textures after GPU upload #21

* Fortunately I didn't need to hack Three.js so far and I import pure Three.js r128. But if we want to apply some changes for example the possible changes mentioned above, I made [a Three.js r128 branch in MozillaReality repo](https://github.com/MozillaReality/three.js/tree/hubs/r128). We can start from it and we can discuss our applying match mechanism. But personally I want to discuss it in another PR because I don't want to make this PR bigger.

* I don't touch anything for Spoke. For Spoke I opened https://github.com/mozilla/Spoke/issues/1127 and hope someone works on it. (Of course I can help)

**Test**

Three.js is one of center cores in Hubs client and it's relevant to almost all the features in Hubs. Would you please help the test to check if all the functionalities are not broken? Please locally merge this PR for the test.

*For Hubs team internal:* I set up a testing instance. You can use it for the test. See the discord for the URL.

If you find any new bugs, please report in this thread.

Once I confirm that basic functionalities doesn't seem to be broken, I will work with the QA team for more systematic and better coverage testing.

I will merge this PR after we finish the tests with the QA team.

**TODOs**

- [x] Resolve the conflicts
- [x] Merge PRs in dependency repos
- [x] Update `package.json` and `package-lock.json`


takahirox:
I found a bug. The hover icons on video are partially hidden by video object. Which file should I look into? `media-views.js`? (And wondering why upgrading Three.js has an effect to this... They changed the rendering order?)

image

netpro2k:
You can try changing the z offset of the hover-menu https://github.com/takahirox/hubs/blob/LatestThree/src/components/position-at-border.js#L196 

Definitely interesting this z fighting is happening from the upgrade though. Did default near/far clip change, or some change to a default depthbuffer setting maybe?

netpro2k:
Actually it looks like 3D object menus are not being positioned correctly either, so its likely the same root issue 
image

takahirox:
> You can try changing the z offset of the hover-menu https://github.com/takahirox/hubs/blob/LatestThree/src/components/position-at-border.js#L196

Hm, it didn't resolve the problem.

takahirox:
OK, figured out.

`BufferAttribute._itemSize` is .`itemSize` in r128.

https://github.com/mozilla/hubs/blob/370afceea31bea641d7eb9814a94a8c3f5db3359/src/utils/auto-box-collider.js#L35

I could hardly notice this type of no-console-error issues beforehand..

netpro2k:
Oh man. Its times like this I really wish we were using something like Typescript heh.

netpro2k:
Found another issue. It seems like the buttons in object menus can be clicked even when the menu is hidden. 

Also, our loading cube asset probably needs to be reworked, it has issues with the transparency change...
image

Might be a nice time to give it a bit of a design refresh anyway.

takahirox:
> Also, our loading cube asset probably needs to be reworked, it has issues with the transparency change...

Is it a glTF?

I'm facing a bunch of bugs in Android Chrome. I'll report them here later. I would like you folks to test not only desktop but also other devices (mobile, XR devices, and so on).

netpro2k:
> Is it a glTF?

Yes it is. Jim should have the original asset. I posted about it in the Discord #design channel. I don't think we need to hold back this PR for it at all since its already like this in production since the GLTFLoader change (hard to notice honestly, only noticed because I was looking closely for issues). I think this will be a good chance to do a refreshed design anyway.

> I would like you folks to test not only desktop but also other devices (mobile, XR devices, and so on), too.

Yep I will test on Quest and my Pixel 4 XL

takahirox:
A bug on Android Chrome. Shader compile error with WebCam avatar.

image

No error on desktop.

image

takahirox:
A bug on Android Chrome.

`physics-utils.js` console error when entering a lobby. I don't see this error on desktop.

image

takahirox:
A bug on Android Chrome.

If I touch Emoji
* big emoji doesn't appear
* small emojis are upside down (flipY + PlaneBufferGeometry issue?)
* "WebGL: INVALID_OPERATION: useProgram: program not valid" console warning

image

Might be related: https://github.com/mozilla/hubs/issues/3357

takahirox:
A bug on Android Chrome. 

Pen doesn't work and "WebGL: INVALID_OPERATION: useProgram: program not valid" console warning.

image

netpro2k:
Re the comments in PR description: 

> I didn't apply this change so far because I don't see any errors so far. I guess it's from batching manager system or Three.js revision older than r128? Let me know if you encounter any shader material problems.

Yeah I think the only place we are currently using RawShaderMaterial with defines is in three-batch-manager. Seems fine to tackle that later when we revisit batching (or otherwise need RawShaderMaterial with defines).

> Only the place which uses Three.js TextureLoader I noticed is A-Frame material system. Do you folks know what the system is for?

This is just used when creating geometry with aframe and applying a material. As far as I know the only geometry we are still creating directly in a-frame is the a-sphere we use for the cursor, and it is not textured.

> I disable to use ImageBitmap for Firefox... 

Yep seems fine for us to revisit this later.

> It seems webpack tree shaking removes the modules under the r128 three/examples/js if I use import. Somehow it's resolve by using require instead. There might be better solutions but I use require in the PR so far.

This seems to be an acceptable solution for now. What I would like to eventually do instead is just use the jsm versions with `import`. That should avoid the tree-shaking issues, but I would want to amke sure we dont accidentally end up with multiple instances of THREE.

> I replaced Batching manager system with a stub so far. We can revert it when we will enable it again.

That seems fine. We will need to tackle this to get Spoke updated since batching is pretty important to the architecture kit being usable.

> It seems this change is for Oculus browser specific performance problem...

I will test on Quest and see if its still an issue. If I remember correctly it was mostly an issue with MultiView which we are no longer using anyway.

netpro2k:
> small emojis are upside down (ImageBitmap + flipY + PlaneBufferGeometry issue?)

Looks like the emoji particles are also upside down on desktop

image

package.json
```diff
@@ -134,15 +134,16 @@
...
-    "three-ammo": "^1.0.12",
-    "three-bmfont-text": "github:mozillareality/three-bmfont-text#hubs/master",
-    "three-mesh-bvh": "^0.1.2",
-    "three-pathfinding": "github:MozillaReality/three-pathfinding#hubs/master2",
-    "three-to-ammo": "github:infinitelee/three-to-ammo",
+    "three": "^0.128.0",
+    "three-ammo": "github:takahirox/three-ammo#LatestThree",
+    "three-bmfont-text": "github:takahirox/three-bmfont-text#hubs/LatestThree",
+    "three-mesh-bvh": "^0.3.7",
+    "three-pathfinding": "github:takahirox/three-pathfinding#hubs/master3",
```
netpro2k, line 141: This is a different branch than the PR opened on that repo is using.

src/components/simple-water.js
```diff
@@ -287,7 +280,7 @@ AFRAME.registerComponent("simple-water", {
 
   init() {
     if (!waterNormalMap) {
-      waterNormalMap = new TextureLoader().load(waterNormalsUrl);
+      waterNormalMap = new HubsTextureLoader().load(waterNormalsUrl);
```
netpro2k, line 283: Out of scope for this PR, but I wonder if we should just use a single instance of HubsTextureLoader for the whole app.

src/components/teleporter.js
```diff
@@ -24,7 +24,6 @@ const RayCurve = function(numPoints, width) {
   });
 
   this.mesh = new THREE.Mesh(this.geometry, this.material);
-  this.mesh.drawMode = THREE.TriangleStripDrawMode;
```
netpro2k, line 27: This causes the teleporter ray to render incorrectly. Since triangle strips are gone we will need to just adjust how we are generating the actual vertices. I can tackle that. 
image

src/components/water.js
```diff
@@ -140,9 +141,7 @@ MobileWater.prototype = Object.create(THREE.Mesh.prototype);
 MobileWater.prototype.constructor = THREE.Water;
 
 async function loadWaterNormals(url) {
-  const texture = await new Promise((resolve, reject) =>
-    new THREE.TextureLoader().load(url, resolve, undefined, reject)
-  );
+  const texture = await new Promise((resolve, reject) => new HubsTextureLoader().load(url, resolve, undefined, reject));
```
netpro2k, line 144: Out of scope for this PR, but its a bit annoying how we have this promise code everywhere. Would be nice to just add a function to HubsTextureLoader like `loadAsync` that just wraps `load` in a promise for you. (might not be a bad idea for the loader API updstream in general honestly)

src/loaders/HubsTextureLoader.js
```diff
@@ -52,13 +25,39 @@ export default class HubsTextureLoader {
     return texture;
   }
 
-  setCrossOrigin(value) {
-    this.crossOrigin = value;
-    return this;
+  async loadAsync(url, onProgress) {
```
netpro2k, line 28: Oh hah, I mentioned this earlier in the review. Apparently we already have it :)

src/react-components/ui-root.js
```diff
@@ -232,8 +232,22 @@ class UIRoot extends Component {
...
+              this.props.scene.renderer.compile(this.props.scene.object3D, this.props.scene.camera);
+              this.props.scene.object3D.traverse(obj => {
+                if (!obj.material) {
+                  return;
+                }
+                const materials = Array.isArray(obj.material) ? obj.material : [obj.material];
+                for (const material of materials) {
+                  for (const prop in material) {
+                    if (material[prop] && material[prop].isTexture) {
+                      this.props.scene.renderer.initTexture(material[prop]);
```
netpro2k, line 244: Didn't know about this and `renderer.compile` before you mentioned it. Good tools to know about.

src/systems/nav.js
```diff
@@ -1,116 +1,6 @@
 const { Pathfinding } = require("three-pathfinding");
 import qsTruthy from "../utils/qs_truthy";
 
-const Vector3 = THREE.Vector3;
-const Vector2 = THREE.Vector2;
-const Face3 = THREE.Face3;
-const Color = THREE.Color;
-
-// TODO this is backported from a later version of THREE.
```
netpro2k, line 9: Yay, 

- [X] TODO

src/utils/three-utils.js
```diff
@@ -357,3 +363,17 @@ export function traverseAnimationTargets(rootObject, animations, callback) {
     }
   }
 }
+
+export function createPlaneBufferGeometry(width, height, widthSegments, heightSegments, flipY = true) {
+  const geometry = new THREE.PlaneBufferGeometry(width, height, widthSegments, heightSegments);
+  // Three.js seems to assume texture flipY is true for all its built in geometry
```
netpro2k, line 369: Might not be worth dealing with right now, but this comment reminds me of the motivation for turning off flipY in HubsTextureLoader in the first place... Since we are now actually not using ImageBitmapLaoder in Firefox in the first place, maybe we should just get rid of this?

netpro2k (commented):
This is looking great, awesome work! Luckily the code changes in Hubs itself actually seem pretty minimal. And the code changes to dependencies also don't seem too scary. The fact that we are pointing directly at the upstream repo now without any patches is a great place to start fresh from again. We will need to make sure any patches we do need to apply in the future we do in a way that is easy to maintain. I am sure we will find a bunch of little issues but overall this is feeling quite good. And happy to help dig into fixing issues as we find them.

src/components/teleporter.js
netpro2k: Fixed this https://github.com/mozilla/hubs/pull/4321/commits/b3e197a427b94b56d9823e92f9f8a5759f9ac848

takahirox:
> A bug on Android Chrome.
> 
> physics-utils.js console error when entering a lobby. I don't see this error on desktop.

I noticed that I can reproduce this error even on desktop Chrome. The error seems to be dependent on scene. The error appears on Lunar Surface scene but doesn't on WebXR Main Hall scene.

src/components/teleporter.js
takahirox: Thanks!

takahirox:
>A bug on Android Chrome.
>
> physics-utils.js console error when entering a lobby. I don't see this error on desktop.

Fixed this issue. I needed to update `physics-utils.js` to follow the latest `three-mesh-bvh` API.

> It seems webpack tree shaking removes the modules under the r128 three/examples/js if I use import. Somehow it's resolve by using require instead. There might be better solutions but I use require in the PR so far.

This happens only with webpack production mode. We may add `npm run dev-production` command to run dev with production mode. If this command existed I could notice the problem before deploying to the testing instance.

> Looks like its using PlaneBufferGeometry here https://github.com/MozillaReality/lib-hubs/blob/master/packages/three-particle-emitter/src/index.ts#L125 ... I can make a PR to fix that.

Good catch. I didn't know well about `lib-hubs`. Should we move `createPlaneBufferGeometry` util function there?

src/utils/three-utils.js
takahirox: From the discussion on Discord...

Yeah, passing `imageOrientation: 'flipY'` to `ImageBitmapLoader(createImageBitmap)` may be one of the options. It's a good clean up. But I'm on the fence about it because I want to make the change minimal for the tests so far. I don't really add a change that there might be a chance to break some stuffs. (I guess it's rare tho.) We may think of it in follow up PRs?

takahirox:
@netpro2k I confirmed that your recent two commits resolved all the `useProgram` issues I posted above in low material quality mode (= Android Chrome). Nice work!

netpro2k:
I have done all my testing in Linux so far, so that's covered. 

> So, only the problems we noticed and are not resolved yet are upside down small emojis and loading cube transparent issues, aren't they.

Jim provided a new GLB for the loading cube, I will replace it. And yep the emoji issue is the flipy problem, so an easy fix.

takahirox:
The QA team seems to be busy for two-three weeks.
So I would like to finish the basic functionality check by they will have time.

I'm testing Oculus Quest + Oculus browser. I succeeded in entering immersive mode. I found a bug that the hands are not displayed but others looks fine. I'm investing the missing hands problem.

takahirox:
> I found a bug that the hands are not displayed but others looks fine. I'm investing the missing hands problem.

Sorry I just used an avatar which doesn't hands. I switched it to another one having hands, the hands are displayed.

takahirox:
I'm testing Oculus Quest + Firefox Reality and encountered a problem that I can't enter an immersive mode.

The error log is

image

takahirox:
I have fixed the `setDevice()` error.

And now I have a warning and an error on Firefox Reality.

image

I would like to suggest to ignore the both so far because there seems no function problem.

The details are following.

**1. "XRWebGLLayer antialiasing is not supported. Antialiasing will be diabling" warning**

[XRWebGLLayer.antialias is unsupported by Firefox](https://developer.mozilla.org/en-US/docs/Web/API/XRWebGLLayer/antialias).

Firefox reality seems to support WebVR and WebXR APIs by default. [The current Hubs uses WebVR API if the both WebXR and WebXR APIs are available.](https://github.com/mozilla/hubs/blob/e01c49781b7ad2457c87d526a1887d4fd4187bb5/src/webxr-bypass-hacks.js#L13) Firefox Reality WebVR seems to allow antialiasing. I speculate this is the reason why antialiasing works on Firefox Reality in the current hubs.mozilla.com.

I would like to suggest to just ignore this warning and accept non-antialiasing on Firefox Reality. Because I don't want us to the keep WebVR API specific codes only for Firefox Reality. (WebVR API is already deprecated.)

**2. "Uncaught (in promise) TypeError: Failed to find attached VR displays" error.**

This error seems to be caused by the simultaneous WebXR and WebVR request [according to the Gecko code](https://github.com/mozilla/gecko-dev/blob/1aaeb179e246e80c06cede578c6c2ec97f16d2b2/dom/base/Navigator.cpp#L1602).

With `$ find ./src -print0 | xargs -0 grep "navigator\.getVRDisplays"` command actually I have found some codes like

```
if (navigaror.getVRDisplays) {
  navigator.getVRDisplays()...
}
```

According to the Gecko code, it seems we can ignore that error without any problems. I confirmed that I can enter the immersive mode.

We are going to drop WebVR related code (unless there is any problems) so ideally we should remove such codes at some point and probably the error will be disabled. (I don't want to include such clean up in this PR because I don't want to make the PR bigger.)

takahirox:
I tested the very basic function (enter lobby/room/immersive mode, move, VR controls, and etc.) on Oculus Quest + Oculus Browser and Firefox Reality, Android Chrome/Firefox, and iOS safari. @netpro2k tested on Linux. So remaining platform we should test is

* VR Headset + WebVR (e.g. Desktop Firefox + wired VR headset)

And remaining bug we should fix is

* Upside-down small emojis

I would like to finish them and test more basic functions (e.g. placing an object, switching a scene, etc) in the two-three weeks. And then I would like to ask the QA team for the test as I mentioned above.

takahirox:
Fixed the upside-down small emojis problem.

takahirox:
Three.js for Hubs has been upgraded to r128, thanks everyone who helped me!

Now, the latest Three.js revision is r130 (and I guess r131 will be released soon).

https://github.com/mrdoob/three.js/releases

I'm wondering how often should we keep track of Three.js releases? Some options in my mind

1. Always use the latest revision.
2. Upgrade Three.js for every some Three.js releases (for example every three releases).
3. Track the change log of Three.js and upgrade Three.js only when the benefit of upgrading looks big to us.
4. The mix of 2 and 3

The pros and cons of more frequent upgrade are

* Pros:
  * Newer Three.js should have better optimization, stability, and new features. We can get the benefit of them more quickly.
  * One upgrading cost would be relatively smaller. Even if Three.js has breaking changes, it would be easier to follow and test than less frequently upgrading it.
* Cons:
  * Three.js is released every month. Three.js is one of core libraries for us and we should test well, ideally we should entirely test Hubs Client, for each upgrading. Testing well for example every month might be a bit costly.

and the pros and cons of less frequent upgrade are opposite to them.

Any thoughts? I haven't deeply thought yet but for me upgrading every one or two months may be good?

PR #4331, opened 2021-06-15
fix object list not updating

yakyouk:
the listed media list (Objects) does not update: object type remains undefined (cube icon) and object name is `...` .
The list gets updated when adding new objects, except for the just-added object.

Steps to reproduce:
- add an image into the room - see that it is not resolved in listed media
- add another image - see that the first image is now resolved but not the second one
- etc.

Cause:
Object type detection is done by checking the presence of either of the components `media-video`, `media-image`, `media-pdf`, `gltf-model-plus` on the media element but these components are not yet present for new objects when `objects` is being set in `ObjectListProvider`.

Solution:
Emit `listed_media_changed` when these components are playing to trigger `objects` update in `ObjectListProvider`
I'm not sure if this approach makes sense in terms of dependency. I'm thinking that `listed_media_changed` event is not necessarily specifically bound to the media list component, but something like `media_ready` could work better.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-793)

### DIFF
--- a/src/components/gltf-model-plus.js
+++ b/src/components/gltf-model-plus.js
@@ -597,6 +597,10 @@ AFRAME.registerComponent("gltf-model-plus", {
     this.loadTemplates();
   },
 
+  play() {
+    this.el.components["listed-media"] && this.el.sceneEl.emit("listed_media_changed");
+  },
+
   update() {
     this.applySrc(resolveAsset(this.data.src), this.data.contentType);
   },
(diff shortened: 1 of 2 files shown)


johnshaughnessy:
I fixed the merge conflict (that I created when I split `media-views.js` into multiple files) and then merged this fix here https://github.com/mozilla/hubs/pull/4609 

Thank you @yakyouk !

PR #4334, opened 2021-06-16
Fix forward and backward motion when using on-screen touch joysticks.

rawnsley:
The on-screen joysticks "stutter" when you try and go forward and back. This is caused by the Z component of velocity being inverted in order to map _joystick-up_ to _world-forward_, but `this.displacement` is passed by reference so the Z component gets inverted back-and-forth every frame, the net effect of which is zero forward motion. Wiggling the joystick means that `this.displacement` is set correctly every frame and overrides the inversion, so it's not immediately obvious there's a problem, but movement feels "sticky".

Inverting the value when it was calculated seemed the most appropriate fix as the code is already making joystick-to-world-coordinate mapping assumptions there anyway.


netpro2k:
Nice find, and good fix. Thanks!

PR #4347, opened 2021-06-22
Fast Room Switching

netpro2k:
You can now switch rooms without loading a new page or going back through the entry flow. This makes room switching much faster and less painful! You currently need to opt into this feature by enabling it in preferences. 

image

This PR adds the ability to switch rooms without a page refresh. This is done by switching phoenix channel topics, disconnecting from dialog, and connecting to a new dialog room. We also have to do work to clean up objects in the room, handle pinned objects, etc. There is still room to make this faster, but it is already much better than the current flow.

Associated backend change https://github.com/mozilla/reticulum/pull/507


blairmacintyre:
Neat.  Any sense when this will come to hubs cloud?

BabeleDunnit:
would love to see this working also in Oculus Quest...

PR #4349, opened 2021-06-22
isSignedIn should be a boolean, not credentials token(string)

StevenGastroSophy:
add !! to store.state.credentials.token


netpro2k:
Thanks for the PR!

PR #4354, opened 2021-06-22
Added max length 750 to chat messages

robinkwilson:
Message length is now set at 750. 

If we wanted to make the user experience better, we'd have a pop up over the send button or toast message that doesn't send your message until your characters are under 750 chars.

### DIFF
--- a/src/react-components/room/ChatSidebar.js
+++ b/src/react-components/room/ChatSidebar.js
@@ -1,4 +1,4 @@
-import React, { forwardRef } from "react";
+import React, { forwardRef, useState } from "react";
 import PropTypes from "prop-types";
 import classNames from "classnames";
 import { Sidebar } from "../sidebar/Sidebar";
@@ -14,7 +14,7 @@ import { ToolbarButton } from "../input/ToolbarButton";
 import { Popover } from "../popover/Popover";
 import { EmojiPicker } from "./EmojiPicker";
 import styles from "./ChatSidebar.scss";
-import { formatMessageBody, MAX_MESSAGE_LENGTH } from "../../utils/chat-message";
+import { formatMessageBody } from "../../utils/chat-message";
 import { FormattedMessage, useIntl, defineMessages, FormattedRelativeTime } from "react-intl";
 
 export function SpawnMessageButton(props) {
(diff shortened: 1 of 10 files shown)


src/react-components/room/ChatSidebar.js
```diff
@@ -80,6 +80,7 @@ export function ChatInput(props) {
   return (
     <div className={styles.chatInputContainer}>
       <TextAreaInput
+        maxLength={MAX_MESSAGE_LENGTH}
```
netpro2k, line 83: On the client side I think this check should be sufficient. It will provide the UX to let the user know they have entered too many characters (and we should make sure you cant click send when the form is invalid). I don't think we really need to check in send/receive as well. We ideally should add enforcement on the server side though.

netpro2k (commented):
Suggested maybe only checking on the input length rather than checking on send/receive as well. Doing so is fine but it feels a bit redundant. We should at some point add a server side check as well though.

keianhzo:
Same as Dom, this is good for a temp solution but we should enforce this on the server too at some point as it's still possible to repro using the console or we might expose this through the API at some point.

keianhzo (approved):
Same as Dom, this is good for a temp solution but we should enforce this on the server too at some point as it's still possible to repro using the console or we might expose this through the API at some point.

src/react-components/tokens/TokenPageLayout.js
```diff
@@ -8,13 +8,11 @@ import styleUtils from "../styles/style-utils.scss";
...
 
 export const TokenPageLayout = ({ children }) => (
-  <StorybookAuthContextProvider>
-    <PageContainer className={styles.tokenContainer}>
-      <Column gap="xl" className={classNames(styles.centerPaddingSides, styleUtils.xlPadding)}>
-        {children}
-      </Column>
-    </PageContainer>
-  </StorybookAuthContextProvider>
+  <PageContainer className={styles.tokenContainer}>
```
netpro2k, line 11: This looks unrelated. Don't know the context for this, if this is just an unrelated quick change you wanted to put in with this as well that's all good, just confirming its intentional. I usually try to note this in the PR description so whoever is reviewing knows its on purpose.

src/react-components/room/ChatSidebar.js
```diff
@@ -351,6 +371,7 @@ export function ChatSidebar({ onClose, children, ...rest }) {
       title={<FormattedMessage id="chat-sidebar.title" defaultMessage="Chat" />}
       beforeTitle={<CloseButton onClick={onClose} />}
       contentClassName={styles.content}
+      disableOverflowScroll
```
netpro2k, line 374: Is this actually still needed now? The message list and chat box should both now scroll themselves, so not sure when the container would ever end up overflowing. Not really a problem to explicitly disable it as well, but if it is needed it probably means there is a bug...

netpro2k (commented):
Mostly lgtm. Made a few suggestions around using classNames over ternary and integrating ChatLengthWarning more directly into ChatInput but they are pretty minor so totally fine to ignore if you just want to ship this and move on :)

src/react-components/room/ChatSidebar.js
robinkwilson: I kept this inside Sidebar, because in other views like People/Object menus that use the sidebar (not chat), I assume we'll want to enable overflow scroll. Chat is a special case, where the contents manage the scroll themselves.

src/react-components/room/ChatSidebarContainer.js
```diff
@@ -267,6 +272,18 @@ export function ChatSidebarContainer({ scene, canSpawnMessages, presences, occup
         onChange={e => setMessage(e.target.value)}
         placeholder={placeholder}
         value={message}
+        isOverMaxLength={isOverMaxLength}
+        warning={
```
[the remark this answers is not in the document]
robinkwilson: I find this more readable than adding a `maxLength` prop instead since I only need the calculation in the component. However, I removed the `isOverMaxLength` prop in the `<ChatLengthWarning>` because it already has `messageLength` and `maxLength` everything it needs to calculate it. The extra prop was redundant.

PR #4363, opened 2021-06-24
Fix bug that prevents dash videos from getting removed

johnshaughnessy:


netpro2k:
This was noticed testing fast room switching in the smoke test space.

PR #4364, opened 2021-06-24
Separate NAF and Dialog with a phoenix adapter

johnshaughnessy:
This PR changes a few things.

- There is a new `NAF` adapter called `PhoenixAdapter`. The `PhoenixAdapter` handles `naf` and `nafr` events on the `hub:${hubId}` phoenix channel, and tells `NAF` when participants join and leave the hub room. (It uses `Presence` to know about the other participants.)
- The `DialogAdapter` does not have to care about `NAF` anymore, so it doesn't. We give dialog only the information it needs (`serverParams`, `hubId`, `turn` info etc) to join the `protoo` room and establish `WebRTC` connections with the help of `mediasoup`.
- Mic state is simpler:
  - The source of truth is [`APP.dialog.isMicEnabled`](https://github.com/mozilla/hubs/blob/feature/phoenix-adapter/src/naf-dialog-adapter.js#L937-L939). 
  - The mic state is not stored in `localStorage` or in a `muted` css class on the `scene`. 
  - We do not toggle the microphone when selecting a new input device from the preference menu.
- Hub channel presence events are handled by various event handlers, rather than having a small number of event handlers do a bunch of work.
- The control flow in `hub.js` is mostly the same. The biggest difference is that we now need to explicitly tell dialog to connect because it's no longer tied to networked aframe. We should further simplify `hub.js` in a future PR. I tried to keep this diff manageable by postponing some changes that seem like improvements.

### DIFF
--- a/src/hub.js
+++ b/src/hub.js
@@ -37,14 +37,19 @@ import {
   getReticulumFetchUrl,
   getReticulumMeta,
   invalidateReticulumMeta,
-  migrateChannelToSocket
+  migrateChannelToSocket,
+  connectToReticulum,
+  denoisePresence,
+  presenceEventsForHub
 } from "./utils/phoenix-utils";
+import { Presence } from "phoenix";
+import { emitter } from "./emitter";
+import "./phoenix-adapter";
 
 import nextTick from "./utils/next-tick";
 import { addAnimationComponents } from "./utils/animation";
-import { authorizeOrSanitizeMessage } from "./utils/permissions-utils";
 import Cookies from "js-cookie";
-import "./naf-dialog-adapter";
+import DialogAdapter from "./naf-dialog-adapter";
 import "./change-hub";
 
 import "./components/scene-components";
@@ -138,7 +143,6 @@ import { ExitedRoomScreenContainer } from "./react-components/room/ExitedRoomScr
 import AuthChannel from "./utils/auth-channel";
 import HubChannel from "./utils/hub-channel";
 import LinkChannel from "./utils/link-channel";
-import { connectToReticulum } from "./utils/phoenix-utils";
 import { disableiOSZoom } from "./utils/disable-ios-zoom";
 import { proxiedUrlFor } from "./utils/media-url-utils";
 import { traverseMeshesAndAddShapes } from "./utils/physics-utils";
@@ -180,6 +184,7 @@ import { sleep } from "./utils/async-utils";
 import { platformUnsupported } from "./support";
 
 window.APP = new App();
+window.APP.dialog = new DialogAdapter();
 window.APP.RENDER_ORDER = {
   HUD_BACKGROUND: 1,
   HUD_ICONS: 2,
@@ -282,7 +287,6 @@ function setupLobbyCamera() {
 }
 
 let uiProps = {};
-let connectionErrorTimeout = null;
 
 // Hub ID and slug are the basename
 let routerBaseName = document.location.pathname
@@ -342,12 +346,6 @@ function mountUI(props = {}) {
 export function remountUI(props) {
   uiProps = { ...uiProps, ...props };
   mountUI(uiProps);
-}
-
-function setupPeerConnectionConfig(adapter) {
-  const forceTurn = qs.get("force_turn");
-  const forceTcp = qs.get("force_tcp");
-  adapter.setTurnConfig(forceTcp, forceTurn);
 }
 
 export async function getSceneUrlForHub(hub) {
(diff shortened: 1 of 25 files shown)


src/hub.js
```diff
@@ -1237,203 +1203,129 @@ document.addEventListener("DOMContentLoaded", async () => {
   APP.messageDispatch = messageDispatch;
   document.getElementById("avatar-rig").messageDispatch = messageDispatch;
 
-  let isInitialJoin = true;
+  hubChannel.channel = hubPhxChannel;
+  hubChannel.presence = new Presence(hubPhxChannel);
+  const { rawOnJoin, rawOnLeave } = denoisePresence(presenceEventsForHub(events, hubId));
```
johnshaughnessy, line 1205: Phoenix calls the `onJoin` and `onLeave` handlers whenever we receive any changes for the `meta`(s) associated with any `presence` key. (We use `session_id`'s as our keys.)

Here we introduce `denoisePresence` to make it easier to tell which of these things is happening:
- we receive a meta for a NEW presence (aka `session_id`), 
- ALL metas for a given presence (`session_id`) have left, or
- the meta for a given presence (`session_id`) changes.

This allows us to (for example) write `events.on("hub:join", ({key, meta})=>{ //... })` to handle a new participant joining the hub channel.

src/phoenix-adapter.js
```diff
@@ -0,0 +1,258 @@
+import { transportForChannel } from "./transport-for-channel";
+import { authorizeOrSanitizeMessage } from "./utils/permissions-utils";
+
+// TODO: Use the websocket connection, not HEAD requests
+const getTimeOffsetToServer = async () => {
```
johnshaughnessy, line 5: This is the same algorithm for computing the time offset to the server as was in `naf-dialog-adapter`.

src/phoenix-adapter.js
```diff
@@ -0,0 +1,258 @@
...
+    this.events.off("nafr", this.refs.get("nafr"));
+    this.events.off("hub:join", this.refs.get("hub:join"));
+    this.events.off("hub:leave", this.refs.get("hub:leave"));
+    this.refs.delete("naf");
+    this.refs.delete("nafr");
+    this.refs.delete("hub:join");
+    this.refs.delete("hub:leave");
+  }
+
+  toggleFreeze() {
```
johnshaughnessy, line 118: All of the code about freezing, unfreezing, flushing/storing/reducing updates is all taken from `naf-dialog-adapter`.

src/react-components/input/SelectInputField.js
```diff
@@ -110,7 +110,7 @@ SelectInputField.propTypes = {
       PropTypes.string,
       PropTypes.number,
       PropTypes.shape({
-        id: PropTypes.string.isRequired,
+        // id: PropTypes.string.isRequired,
```
johnshaughnessy, line 113: This change fixes a bug in the entry flow. When the microphone options menu is presented, there are no `id`s.

src/react-components/room/ContentMenu.js
```diff
@@ -36,13 +36,13 @@ export function PeopleMenuButton(props) {
...
       <span>
-        <FormattedMessage id="content-menu.people-menu-button" defaultMessage="People" />({props.presenceCount})
+        <FormattedMessage id="content-menu.people-menu-button" defaultMessage="People" />({`${props.presencecount}`})
       </span>
     </ContentMenuButton>
   );
 }
 PeopleMenuButton.propTypes = {
-  presenceCount: PropTypes.number
+  presencecount: PropTypes.number
```
johnshaughnessy, line 45: This change fixes a bug where React complains about `presenceCount` and tells us to use (lowercase) `presencecount` if we don't like red text in the console.

src/transport-for-channel.js
```diff
@@ -0,0 +1,40 @@
+export function transportForChannel(channel, reliable = true) {
```
johnshaughnessy, line 1: This is not new code. It was taken from `hub.js`

src/scene-entry-manager.js
```diff
@@ -142,8 +136,8 @@ export default class SceneEntryManager {
 
   exitScene = () => {
     this.scene.exitVR();
-    if (NAF.connection.adapter && NAF.connection.adapter.localMediaStream) {
-      NAF.connection.adapter.localMediaStream.getTracks().forEach(t => t.stop());
+    if (APP.dialog && NAF.connection.adapter.localMediaStream) {
```
yakyouk, line 139: should it be     `if (APP.dialog && APP.dialog.localMediaStream) {` ?

johnshaughnessy: Yes, thank you. Good catch!

johnshaughnessy:
Closing this in favor of https://github.com/mozilla/hubs/pull/4374

PR #4374, opened 2021-06-25
Add NAF Phoenix Adapter

johnshaughnessy:
This PR changes a few things.

- There is a new `NAF` adapter called `PhoenixAdapter`. The `PhoenixAdapter` handles `naf` and `nafr` events on the `hub:${hubId}` phoenix channel, and tells `NAF` when participants join and leave the hub room. (It uses `Presence` to know about the other participants.)
- The `DialogAdapter` does not have to care about `NAF` anymore, so it doesn't. We give dialog only the information it needs (`serverParams`, `hubId`, `turn` info etc) to join the `protoo` room and establish `WebRTC` connections with the help of `mediasoup`.
- Mic state is simpler:
  - The source of truth is [`APP.dialog.isMicEnabled`](https://github.com/mozilla/hubs/blob/feature/phoenix-adapter/src/naf-dialog-adapter.js#L937-L939). 
  - The mic state is not stored in `localStorage` or in a `muted` css class on the `scene`. 
  - We do not toggle the microphone when selecting a new input device from the preference menu.
- Hub channel presence events are handled by various event handlers, rather than having a small number of event handlers do a bunch of work.
    
    
Built on top of https://github.com/mozilla/hubs/pull/4364

### DIFF
--- a/src/hub.js
+++ b/src/hub.js
@@ -1,3 +1,4 @@
+import { getCurrentHubId } from "./utils/hub-utils";
 import "./utils/debug-log";
 import "./webxr-bypass-hacks";
 import configs from "./utils/configs";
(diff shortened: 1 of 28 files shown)


src/hub.js
```diff
@@ -739,13 +752,14 @@ document.addEventListener("DOMContentLoaded", async () => {
 
   const authChannel = new AuthChannel(store);
   const hubChannel = new HubChannel(store, hubId);
+  window.APP.hubChannel = hubChannel;
+
+  store.addEventListener("profilechanged", hubChannel.sendProfileUpdate.bind(hubChannel));
```
netpro2k, line 757: Since HubChannel is passed a reference to store, this can just move into its constructor.

PR #4392, opened 2021-07-05
MediaQueryList.addEventListener doesn't work in Safari and IE

StevenGastroSophy:
however you can use MediaQueryList.addListener

refer:
https://stackoverflow.com/questions/62028516/window-matchmedia-does-not-work-in-safari

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-785)

### DIFF
--- a/src/react-components/styles/theme.js
+++ b/src/react-components/styles/theme.js
@@ -33,9 +33,15 @@ function useDarkMode() {
 
     setDarkMode(darkmodeQuery.matches);
 
-    darkmodeQuery.addEventListener("change", event => {
-      setDarkMode(event.matches);
-    });
+    if (darkmodeQuery.addEventListener) {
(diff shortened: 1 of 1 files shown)


takahirox (approved):
According to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/addListener)

> Prior to Safari 14, MediaQueryList is based on EventTarget, so you must use addListener() and removeListener() to observe media query lists.

The latest Safari supports `MediaQueryList.addEventListener`. Actually my Safari 14.1.2 supports it.

But yes, some users may use older version. And the change doesn't really hurt on us. I think this change is ok to us.

I would like to add comment describing what this workaround is for and that we will remove this workaround at some point when no one will use old Safari 13 or before. I will make a follow up PR.

Thanks for the contribution!

Issue #4394, opened 2021-07-07
Open media button links can't be clicked from touchscreen devices.


MegaMotion:
**Description**
It is very difficult if not impossible to click on "visit room" links (and possibly other open-media-button links) from a touchscreen device, notably iPhone or Android devices.

**To Reproduce**
1. Go to this room, on a touchscreen device:  https://hubs.mozilla.com/dwghAPe/accomplished-powerless-festivity/
2. Attempt to click on a link to a breakout room.
3. Fail

**Hardware** 
 - Any iPhone or Android touchscreen device
 - 
**Additional context**
Was present before the 04/20  hubs client update, but got worse after the update.

robinkwilson:
I reproed on my Pixel 3. It captures clicks (sometimes) on the words but not the button itself.

Softvision-GeluHaiduc:
I have verified this issue using the latest mobile releases of Firefox and Chrome on an Android 11 device.
While verifying using the room from the above steps the issue reproduces, creating a new room with the same scene worked and the issue did not reproduce anymore. 
After further investigation, I noticed that if the “Create and move objects” setting from room settings is disabled the issue is reproducible again for other users than the room creator, on mobile.

Softvision-RemusDranca:
I have the same behavior on Iphone SE2 (iOS 15) with Safari as described by @Softvision-GeluHaiduc .

takahirox:
Thanks for the report. I confirmed I can reproduce the problem on my mobiles and changed the issue subject to clarify.

takahirox:
Sounds like the same issue as #3242

takahirox:
Should be fixed by #4740

Softvision-RemusDranca:
I have verified the issue and it's no longer reproducible. The links and waypoints are now working while the “Create and move objects” setting is disabled.
Tested on Windows 10 using Chrome(94) and Firefox(93), macOS(Big Sur 11.6) using Safari(15.0), iOS (15.0.1) with Safari Android (11) using Chrome(94) and Firefox(92.1.1).

Issue #4297, opened 2021-05-28
links-button not completely clickable on mobile devices


postelontwerpen:
**If I make a link to other rooms or to way-points (in Spoke and Hubs), the link on mobile devices is only clickable in the bottom part of the pink/red button.** 

- Issue on mobile devices in Chrome Safari and Firefox. 
- It concerns links in Spoke and links added in Hubs.
- The Button-text (visit link) also disappears on mobile.
- On desktop-browsers it works fine.

To illustrate the problem i made a small video where you can see me clicking in the top of the button (not working) and after that, at the bottom of the button (works)

https://user-images.githubusercontent.com/11406875/119941130-77bcf480-bf90-11eb-84f3-c211d65abb0e.mp4

Can you fix this?

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-117)

yakyouk:
I suspect this is a  problem of the button being too close to its mesh and not registering touch, as someone mentioned that scaling to 400% or 500% (which also scales the button's z-offset) seems to get the buttons to work reliably. Should be an easy fix.

postelontwerpen:
thank you, @yakyouk 
I only can scale the link-asset in Spoke to a huge format. Than it is easier to click, that's true. But that didn't change the clickable region in the bottom of the button. I also put it in a complete empty scene. No change and mesh influence. Am I looking in the wrong place or direction?

yakyouk:
hmm then maybe it's not that simple  XD

yakyouk:
This seems to be due to a miscalculation of the tap coordinates within the aframe element, there seems to be a constant (in screen coordinates, i.e. *before* raycasting) negative vertical offset between the tap and where it hits in 3D.
When getting close to a button, it's not obvious since the button is big and the offset is small, relatively, but when far away, the offset may be larger than the button itself so the button isn't clickable.

Issue #4075, opened 2021-03-23
[Mobile] Button on object is not work when avatar is far from the object


jasoncks:
**Description**
In mobile, button on link/image object is not work when avatar is far from the object, but when the avatar get close to object/button, the buttons function properly.
This problem occur in mobile only.

**To Reproduce**
1. Create a new room
2. Turn on the freeze mode
3. Move the avatar in front of the object and click the button, the button function normally
4. Move the avatar apart from the object, the button not work

**Expected behavior**
- The button should work properly whenever user can click on the button

**Screenshots**
[Screenshot Far](https://drive.google.com/file/d/1zWlEpJP-J4T4VOqVPF3vWyEXve1Mz1t9/view?usp=sharing)
[Screenshot Near](https://drive.google.com/file/d/1kVPXJLA8CwJW-Sf3HhvVq_KYP3CRK9i6/view?usp=sharing)

**Hardware** 
 - Device: iPhone X, iPhone X Max
 - OS: iOS
 - Browser: Safari

 - Device: Google Pixel 4 XL
 - OS: Android
 - Browser: Chrome

**Additional context**
Not sure if is the same problem(#4048, #4052), since it does not mention about the distance

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-593)

yakyouk:
[this pull request](https://github.com/mozilla/hubs/pull/4397) should fix this

takahirox:
Hi @jasoncks, thanks for the bug report.

Do you mind if testing again to check if the problem is still reproducible?

Softvision-GeluHaiduc:
I have verified this issue using the latest mobile releases of Firefox and Chrome on an Android 11 device.
Now, the buttons on objects are working even when the avatar is far from the object.

PR #4397, opened 2021-07-08
fix mobile touch coordinates

yakyouk:
Fixes buttons partially clickable or not clickable on mobile
https://github.com/mozilla/hubs/issues/4075
https://github.com/mozilla/hubs/issues/4297
https://github.com/mozilla/hubs/issues/4394

Touch coordinates calculation in `shouldMoveCursor` was using `window.innerWidth` and `window.innerHeight`, but the a-frame canvas is not always filling the window, e.g. with the new UI's toolbar.

Changed `window.innerWidth` and `window.innerHeight` to `a-canvas` element's width and height.
In case the canvas is further embedded in a UI in the future, the code also subtracts the canvas' left and top offsets from `touch.clientX` and `touch.clientY`.

**Notes:**
- This does not fix the issue where taps on `singleActionButton` buttons don't work for users who do not have the right to *move objects*.
- On an iPhone you have to go to  Settings > Safari and turn off Block Pop-ups and Fraudulent Website Warning, or else the link silently fails to open.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-784)

### DIFF
--- a/src/systems/userinput/devices/app-aware-touchscreen.js
+++ b/src/systems/userinput/devices/app-aware-touchscreen.js
@@ -41,10 +41,11 @@ function shouldMoveCursor(touch, raycaster) {
     return true;
   }
   const rawIntersections = [];
+  const canvas = document.querySelector(".a-canvas").getBoundingClientRect();
   raycaster.setFromCamera(
     {
-      x: (touch.clientX / window.innerWidth) * 2 - 1,
-      y: -(touch.clientY / window.innerHeight) * 2 + 1
+      x: ((touch.clientX - canvas.left) / canvas.width) * 2 - 1,
+      y: -((touch.clientY - canvas.top) / canvas.height) * 2 + 1
     },
     getPlayerCamera()
   );
--- a/src/systems/userinput/devices/app-aware-touchscreen.js
+++ b/src/systems/userinput/devices/app-aware-touchscreen.js
@@ -3,6 +3,7 @@ import { Pose } from "../pose";
 import { touchIsAssigned, jobIsAssigned, assign, unassign, findByJob, findByTouch } from "./touchscreen/assignments";
 import { findRemoteHoverTarget } from "../../interactions";
 import { canMove } from "../../../utils/permissions-utils";
+import ResizeObserver from "resize-observer-polyfill";
 
 const MOVE_CURSOR_JOB = "MOVE CURSOR";
 const MOVE_CAMERA_JOB = "MOVE CAMERA";
@@ -35,17 +36,16 @@ const getPlayerCamera = (() => {
   };
 })();
 
-function shouldMoveCursor(touch, raycaster) {
+function shouldMoveCursor(touch, rect, raycaster) {
   const isCursorGrabbing = !!AFRAME.scenes[0].systems.interaction.state.rightRemote.held;
   if (isCursorGrabbing) {
     return true;
   }
   const rawIntersections = [];
-  const canvas = document.querySelector(".a-canvas").getBoundingClientRect();
   raycaster.setFromCamera(
     {
-      x: ((touch.clientX - canvas.left) / canvas.width) * 2 - 1,
-      y: -((touch.clientY - canvas.top) / canvas.height) * 2 + 1
+      x: ((touch.clientX - rect.left) / rect.width) * 2 - 1,
+      y: -((touch.clientY - rect.top) / rect.height) * 2 + 1
     },
     getPlayerCamera()
   );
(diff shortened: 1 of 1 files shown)


src/systems/userinput/devices/app-aware-touchscreen.js
```diff
@@ -41,10 +41,11 @@ function shouldMoveCursor(touch, raycaster) {
     return true;
   }
   const rawIntersections = [];
+  const canvas = document.querySelector(".a-canvas").getBoundingClientRect();
```
netpro2k, line 44: This only gets called in touchStart anyway so its not a huge deal, but querySelectors can be really slow so we try to avoid them when we can. Also the use of `left` and `top` on the next lines threw me because of the variable name, didn't notice the `getBoundingClientRect()` at first
```suggestion
  const canvasRect = AFRAME.scense[0].getBoundingClientRect();
```

src/systems/userinput/devices/app-aware-touchscreen.js
```diff
@@ -41,10 +41,11 @@ function shouldMoveCursor(touch, raycaster) {
...
   raycaster.setFromCamera(
     {
-      x: (touch.clientX / window.innerWidth) * 2 - 1,
-      y: -(touch.clientY / window.innerHeight) * 2 + 1
+      x: ((touch.clientX - canvas.left) / canvas.width) * 2 - 1,
```
netpro2k, line 47:
```suggestion
      x: ((touch.clientX - canvasRect.left) / canvasRect.width) * 2 - 1,
```

src/systems/userinput/devices/app-aware-touchscreen.js
```diff
@@ -41,10 +41,11 @@ function shouldMoveCursor(touch, raycaster) {
...
+      y: -((touch.clientY - canvas.top) / canvas.height) * 2 + 1
```
netpro2k, line 48:
```suggestion
      y: -((touch.clientY - canvasRect.top) / canvasRect.height) * 2 + 1
```

netpro2k (commented):
Yikes good catch. I had assumed the reports we were getting about this were just about our touch controls in general being pretty floaty and unreliable (they need to be reworked) but this fixes a clear regression from adding the toolbar/sidebar. Made some minor suggestions you can apply if you like, but I can also just handle them when merging.

src/systems/userinput/devices/app-aware-touchscreen.js
yakyouk: I wanted to use `AFRAME.scene[0]` but for some reason it includes the toolbar as well, whereas the canvas doesn't.
image
But we can store a ref to canvas to avoid querying it every time.

yakyouk: I pushed some changes:
We actually have a ref to canvas in `this.canvas`, and I store canvas rect in `this.canvasRect` and update it on resize using `ResizeObserver`.
Since `shouldMoveCursor` is outside of the object and does not see `this.canvasRect`, I pass rect to the function as well.

src/systems/userinput/devices/app-aware-touchscreen.js
```diff
@@ -174,8 +176,8 @@ export class AppAwareTouchscreenDevice {
       case MOVE_CURSOR_JOB:
         assignment.cursorPose.fromCameraProjection(
           getPlayerCamera(),
-          (touch.clientX / this.canvas.clientWidth) * 2 - 1,
-          -(touch.clientY / this.canvas.clientHeight) * 2 + 1
+          ((touch.clientX - this.canvasRect.left) / this.canvasRect.width) * 2 - 1,
+          -((touch.clientY - this.canvasRect.top) / this.canvasRect.height) * 2 + 1
```
yakyouk, line 180: Updated calculation here too

wmurphyrd:
@netpro2k @yakyouk we also have a fix for this, since the main issue was that `shouldMoveCursor` was casting a slightly different ray than what `start` would later set as `assignment.cursorPose`, I just re-arranged the code to ensure that both places were able to share a single pose calculation. 

https://github.com/immers-space/hubs/pull/43/commits/298f8d0ddd526c10e269a2798cd075165eecb08e

We have this in production on https://vreign.space if you'd like to test. 

Dom, let me know if you'd like to see this in its own PR or combined yakyouk's work here

netpro2k:
> Dom, let me know if you'd like to see this in its own PR or combined yakyouk's work here
Yep I think a separate PR would be nice, since this is already ready to merge.

netpro2k (approved):

Softvision-GeluHaiduc:
After verifying some mobile issues I noticed that waypoints and links to other rooms or websites don’t work on mobile devices if the “Create and move objects” setting from room settings is disabled, the issues are not reproducible for the creator of the room but all the other users are affected.
I logged issue #4691 to track this.

PR #4418, opened 2021-07-16
fix permission token update, use dialog adapter token for kick

yakyouk:
Remove event `permissions-refreshed` that was redundant with `permissions_updated`.
Simplify permission refresh timer, clear existing timer if permissions are refreshed early.
In `hubChannel.kick`, do not refetch permissions so as to pass perms token to NAF dialog adapter's `kick`; instead, use NAF dialog adapter's `_joinToken` internally.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-783)

### DIFF
--- a/src/utils/hub-channel.js
+++ b/src/utils/hub-channel.js
@@ -150,10 +150,7 @@ export default class HubChannel extends EventTarget {
     if (this.fetchPermissionsTimeout) {
       clearTimeout(this.fetchPermissionsTimeout);
     }
-    this.fetchPermissionsTimeout = setTimeout(async () => {
-      const result = await this.fetchPermissions();
-      this.dispatchEvent(new CustomEvent("permissions-refreshed", { detail: result }));
-    }, nextRefresh);
+    this.fetchPermissionsTimeout = setTimeout(this.fetchPermissions, nextRefresh);
   };
 
   sendEnteringEvent = async () => {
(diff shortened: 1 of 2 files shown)


src/utils/hub-channel.js
```diff
@@ -145,14 +145,12 @@ export default class HubChannel extends EventTarget {
...
     this._permissions = jwtDecode(token);
     configs.setIsAdmin(this._permissions.postgrest_role === "ret_admin");
-    this.dispatchEvent(new CustomEvent("permissions_updated"));
+    this.dispatchEvent(new CustomEvent("permissions_updated", { detail: { permsToken: token } }));
 
     // Refresh the token 1 minute before it expires.
-    const nextRefresh = new Date(this._permissions.exp * 1000 - 60 * 1000) - new Date();
-    setTimeout(async () => {
-      const result = await this.fetchPermissions();
-      this.dispatchEvent(new CustomEvent("permissions-refreshed", { detail: result }));
```
yakyouk, line 154: `fetchPermissions` calls `setPermissionsFromToken` which dispatches `permissions_updated` so we don't need to dispatch here.

netpro2k:
The adapter code has changed a bit since you submitted this. This change definitely makes sense though. If you are up for updating it that would be great, otherwise just noting here for whichever one of us on the tip picks up this PR

yakyouk:
I updated the change

PR #4440, opened 2021-07-23
Load builtin audio-params for audio-targets

keianhzo:
We were no loading audio params for audio-targets with builtin audio params (instead of having a separate audio-params component)

### DIFF
--- a/src/gltf-component-mappings.js
+++ b/src/gltf-component-mappings.js
@@ -470,7 +470,26 @@ AFRAME.GLTFModelPlus.registerComponent(
       }
     }
 
-    el.setAttribute(componentName, { ...componentData, srcEl });
+    // Migrate audio-target component that still has builting audio params
+    if (componentData.positional !== undefined) {
+      el.setAttribute("audio-params", {
+        sourceType: SourceType.AUDIO_TARGET,
+        distanceModel: componentData.distanceModel,
+        rolloffFactor: componentData.rolloffFactor,
+        refDistance: componentData.refDistance,
+        maxDistance: componentData.maxDistance,
+        coneInnerAngle: componentData.coneInnerAngle,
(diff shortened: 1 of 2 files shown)


src/gltf-component-mappings.js
```diff
@@ -470,7 +470,26 @@ AFRAME.GLTFModelPlus.registerComponent(
       }
     }
 
-    el.setAttribute(componentName, { ...componentData, srcEl });
+    // Migrate audio-target component that still has builting audio params
+    if (componentData.positional !== undefined) {
```
netpro2k, line 474: Hmm, if the audio type is not positional then we won't actually add an `audio-params` object, but I think other code in `audio-source` like `createAudio()` assumes one always exists...

netpro2k (commented):
Importing the existing parameters into separate `audio-params` makes total sense. Not quite sure what happens now though for non positional sources...

src/gltf-component-mappings.js
keianhzo: The blender exported added that parameter for `audio-target` (before the `audio-params` refactor) so we consider that if that exists this is an old `audio-target` component and we create the `audio-params` component.

https://github.com/MozillaReality/hubs-blender-exporter/pull/29/commits/1be80cccfc0101017f9fd033a30c1a564cd17ae1#diff-a07a63e062368f64b62670e35510bc7194e1c23c69aaa0732d25586ff734eee6R795-R798

Not sure why we used that instead of the audioType enum that we used in other components.

Anyways I'm' closing this PR in  favor of https://github.com/mozilla/hubs/pull/4444  as we restoring the stereo support there and was missing here.

keianhzo:
Closing in favor of https://github.com/mozilla/hubs/pull/4444

Issue #4411, opened 2021-07-14
[Safari] The audio is broken on certain MacOS/Safari versions


Softvision-RemusDranca:
**[Notes]:**
- The issue is reproducible when communicating with other users and also when playing spawned videos.
- The performance also seems to be affected.

**[Affected Versions]:**
- Safari - Version 14.1.1 (15611.2.7.1.6, 15611)

**[Affected Platforms]:**
- MacOS Catalina 10.15.7 (19H1217)

**[Prerequisites]:**
- Have a room opened on "https://hubs.mozilla.com/".

**[Steps to reproduce]:**
1. Spawn a video inside the room.
2. Play the video for a small period of time.
3. Observe the behavior.

**[Expected results]:**
- The video is correctly played.

**[Actual results]:**
- The audio is broken.

**[Additional Notes]:**
- I reproduced this issue using different scenes.
- I reproduced the issue when playing videos inside the room.
- I reproduced the issue on 3 different devices using the same version.
- According to statcounter macOS Catalina is the most used version for MacOS. 
- Could be related to #4288 
- Attached is a screen recording of the issue where you can see the performance being affected: [link](https://drive.google.com/file/d/1zaJW_Ubk3rqvwovLI-7k-32ZtHtEEyjG/view?usp=sharing).
- Here are the logs received while reproducing the issue: [link](https://drive.google.com/file/d/18KJ1s0aOSVV0hGCs0codMHhdI0EulMSH/view?usp=sharing).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBSOLD-72)

Softvision-RemusDranca:
I've encountered an additional behavior while testing. If for example a YouTube video is playing in a different tab, when focusing the Hubs tab the video's audio from the Youtube tab is broken.

Softvision-RemusDranca:
- I've managed to get some help from another team and they were also able to reproduce the issue using:
	- MacBook Pro macOS Big Sur Version 11.3.1 (20E241), Safari Version 14.1 (16611.1.21.161.6).
	- MacBook Pro macOS Big Sur Version 11.4 (20F71), Safari Version 14.1.1 (16611.2.7.1.4).
[...]

PR #4441, opened 2021-07-23
Force disable panner audio in safari

brianpeiris:


brianpeiris:
This was a hotfix for hubs-cloud that disabled panner audio in safari in order to avoid browser regressions that broke audio playback in rooms. (See #4411). This was merged into `qa-test` in #4442

These changes were specific to the hubs-cloud branch at the time, and were not merged into `master`, since the audio code in `master` had changed considerably at that point. The changes in this PR should be overridden by the current state of audio code in `master`.

Issue #4438, opened 2021-07-22
GL_INVALID_OPERATION console warning in Outdoor Festival scene


takahirox:
**Description**

In Outdoor Festival scene repeatedly `GL_INVALID_OPERATION: Feedback loop formed between Framebuffer and active Texture.` warning shows up in the console.

**To Reproduce**

1. Go to https://hubs.mozilla.com
2. Enter a scene and change the scene to Outdoor Festival, or enter a scene whose scene is Outdoor Festival
3. See the console warnings

**Expected behavior**

No console warnings.

**Console logs**

```
core:schema:warn Unknown property `srcNode` for component/system `audio-target`. 
console.js:35 core:schema:warn Unknown property `gain` for component/system `audio-target`. 
console.js:35 core:schema:warn Unknown property `positional` for component/system `audio-target`. 
console.js:35 core:schema:warn Unknown property `distanceModel` for component/system `audio-target`. 
console.js:35 core:schema:warn Unknown property `refDistance` for component/system `audio-target`. 
console.js:35 core:schema:warn Unknown property `rolloffFactor` for component/system `audio-target`. 
console.js:35 core:schema:warn Unknown property `maxDistance` for component/system `audio-target`. 
console.js:35 core:schema:warn Unknown property `coneInnerAngle` for component/system `audio-target`. 
console.js:35 core:schema:warn Unknown property `coneOuterAngle` for component/system `audio-target`. 
console.js:35 core:schema:warn Unknown property `coneOuterGain` for component/system `audio-target`. 
console.js:35 core:schema:warn Unknown property `srcNode` for component/system `audio-target`. 
console.js:35 core:schema:warn Unknown property `gain` for component/system `audio-target`. 
console.js:35 core:schema:warn Unknown property `positional` for component/system `audio-target`. 
console.js:35 core:schema:warn Unknown property `distanceModel` for component/system `audio-target`. 
console.js:35 core:schema:warn Unknown property `refDistance` for component/system `audio-target`. 
console.js:35 core:schema:warn Unknown property `rolloffFactor` for component/system `audio-target`. 
console.js:35 core:schema:warn Unknown property `maxDistance` for component/system `audio-target`. 
console.js:35 core:schema:warn Unknown property `coneInnerAngle` for component/system `audio-target`. 
console.js:35 core:schema:warn Unknown property `coneOuterAngle` for component/system `audio-target`. 
[...]

PR #4444, opened 2021-07-23
Audio fixes

### LINKED ISSUE #4434: Whole-room sounds out of balance with localised sounds
**Description**
I've gone back into an existing Hubs piece today which I last touched in mid-May (at https://hubs.mozilla.com/niSUXry/haydn-space-opera ) and I'm finding that in every room that contains a heard-everywhere sound in combination with localised sounds, the balance between these two types of sound has changed although I have not touched the piece since. The balance difference is that the heard-everywhere sound is extremely quiet in relation to the localised sounds. 

For example, in the first room, the warmup singing used to be of an equivalent volume to the sounds of the videos and the main sonic focus of the space, but now can barely be heard while the video sound is prominent; in the Phyllida Barlow Playground (accessible from that first room), the main track used to be able to be clearly heard, significantly louder than the mumbled text audio (which is intentionally very quiet), but is now so quiet that the fragment of conversation at the start of it can't be heard at all, even with system volume at maximum (levels were managed to be what I felt optimal at about half-volume on my machine originally, to give scope for user adjustments in both directions).

I have changed nothing in either Spoke or Hubs since it was presented at a digital festival in May, when it was fine, and it was still fine a month ago when I captured a walkthrough for a conference presentation (link below).

**To Reproduce**
Visit [https://hubs.mozilla.com/boG8E2Q/phyllida-barlow-playground-haydn-space-opera](https://hubs.mozilla.com/boG8E2Q/phyllida-barlow-playground-haydn-space-opera):
1. Move around, note relative levels of the localised mumbling sounds (the sound by each of the individual eye videos [not part of the videos - separate sound files], group mumbling by the spawn point) to the main track which is hear-everywhere
2. compare levels with provided video of walkthrough captured a little over a month ago (doesn't start from the beginning of the room but from just before the hear-everywhere track begins with the "OK? OK... and go!" exchange - I think there's about a 30-45second lead in before it starts when you're actually in the room.) In particular, note how that OK exchange is either completely inaudible or background-volume whereas in the previous walkthrough it cuts through the other sounds to draw sonic focus.

**Expected behavior**
 View video at [https://youtu.be/qMzvIb_GsUg](https://youtu.be/qMzvIb_GsUg) to see how the balance used to be - I have changed nothing since then.

**Screenshots**
Not applicable

**Hardware** 
 - Device: Laptop
 - OS: Windows 10
 - Browser: Chrome & Firefox

**Additional context**
Also tested with headphones but it's much the same - you can hear that first exchange if you're looking for it, but it should be really clear and obvious, as per the video but it's close to inaudible.

### LINKED ISSUE #4454: Failure loading node:   TypeError: Cannot read property 'components' of null
The following exception has been reported when other users log in a session. 

```
core:a-node:error Failure loading node:   TypeError: Cannot read property 'components' of null
    at n.init (avatar-volume-controls.js:19)
    at n.initComponent (component.js:320)
    at n.updateProperties (component.js:302)
    at n.e.exports.Component (component.js:78)
    at new n (component.js:662)
    at HTMLElement.value (a-entity.js:332)
    at HTMLElement.value (a-entity.js:495)
    at HTMLElement.value (a-entity.js:463)
    at a-entity.js:249
    at a-node.js:127
```

keianhzo:
This PR fixes several audio related issues:

- We were restoring the component data with the panner values every tick so the component always got reset to default value. The first commit fixes that. Fixes #4434
- Restores the support for stereo audio elements that was accidentally removed during the audio parameters refactor.
- We weren't creating the `audio-params`  component for old audio-targets with builtin audio parameters. Related #4438
- We were still setting volume  on `avatar-audio-source` but we are now setting that on the `audio-params` component and that was logging an error. Fixes #4454

### DIFF
--- a/src/components/audio-params.js
+++ b/src/components/audio-params.js
@@ -197,13 +197,6 @@ AFRAME.registerComponent("audio-params", {
         audio.panner.orientationY.value,
         audio.panner.orientationZ.value
       );
-      this.data.distanceModel = audio.panner.distanceModel;
-      this.data.rolloffFactor = audio.panner.rolloffFactor;
-      this.data.refDistance = audio.panner.refDistance;
-      this.data.maxDistance = audio.panner.maxDistance;
-      this.data.coneInnerAngle = audio.panner.coneInnerAngle;
-      this.data.coneOuterAngle = audio.panner.coneOuterAngle;
-      this.data.coneOuterGain = audio.panner.coneOuterGain;
       this.updateDistances();
       this.updateAttenuation();
     }
--- a/src/components/audio-params.js
+++ b/src/components/audio-params.js
@@ -200,25 +200,21 @@ AFRAME.registerComponent("audio-params", {
     const audio = this.getAudio();
     if (audio) {
       if (audio.panner) {
-        this.el.setAttribute("audio-params", {
-          position: new THREE.Vector3(
-            audio.panner.positionX.value,
-            audio.panner.positionY.value,
-            audio.panner.positionZ.value
-          ),
-          orientation: new THREE.Vector3(
-            audio.panner.orientationX.value,
-            audio.panner.orientationY.value,
-            audio.panner.orientationZ.value
-          )
-        });
+        this.data.position = new THREE.Vector3(
+          audio.panner.positionX.value,
+          audio.panner.positionY.value,
+          audio.panner.positionZ.value
+        );
+        this.data.orientation = new THREE.Vector3(
+          audio.panner.orientationX.value,
+          audio.panner.orientationY.value,
+          audio.panner.orientationZ.value
+        );
         this.updateAttenuation();
       } else {
         this.el.object3D.getWorldDirection(this.data.orientation);
         this.el.object3D.getWorldPosition(this.data.position);
-        this.el.setAttribute("audio-params", {
-          rolloffFactor: 0
-        });
+        this.data.rolloffFactor = 0;
       }
       this.updateDistances();
     }
@@ -313,21 +309,17 @@ AFRAME.registerComponent("audio-params", {
 
   updateDistances() {
     this.el.sceneEl.audioListener.getWorldPosition(this.listenerPos);
-    this.el.setAttribute("audio-params", {
-      distance: this.data.position.distanceTo(this.listenerPos),
-      squaredDistance: this.data.position.distanceToSquared(this.listenerPos)
-    });
+    this.data.distance = this.data.position.distanceTo(this.listenerPos);
+    this.data.squaredDistance = this.data.position.distanceToSquared(this.listenerPos);
   },
 
   updateAttenuation() {
-    this.el.setAttribute("audio-params", {
-      attenuation: distanceModels[this.data.distanceModel](
-        this.data.distance,
-        this.data.rolloffFactor,
-        this.data.refDistance,
-        this.data.maxDistance
-      )
-    });
+    this.data.attenuation = distanceModels[this.data.distanceModel](
+      this.data.distance,
+      this.data.rolloffFactor,
+      this.data.refDistance,
+      this.data.maxDistance
+    );
   },
 
   clipGain(gain) {
--- a/src/components/audio-params.js
+++ b/src/components/audio-params.js
@@ -194,21 +194,25 @@ AFRAME.registerComponent("audio-params", {
     const audio = this.getAudio();
     if (audio) {
       if (audio.panner) {
-        this.data.position = new THREE.Vector3(
-          audio.panner.positionX.value,
-          audio.panner.positionY.value,
-          audio.panner.positionZ.value
-        );
-        this.data.orientation = new THREE.Vector3(
-          audio.panner.orientationX.value,
-          audio.panner.orientationY.value,
-          audio.panner.orientationZ.value
-        );
+        this.el.setAttribute("audio-params", {
+          position: new THREE.Vector3(
+            audio.panner.positionX.value,
+            audio.panner.positionY.value,
+            audio.panner.positionZ.value
+          ),
+          orientation: new THREE.Vector3(
+            audio.panner.orientationX.value,
+            audio.panner.orientationY.value,
+            audio.panner.orientationZ.value
+          )
+        });
         this.updateAttenuation();
       } else {
         this.el.object3D.getWorldDirection(this.data.orientation);
         this.el.object3D.getWorldPosition(this.data.position);
-        this.data.rolloffFactor = 0;
+        this.el.setAttribute("audio-params", {
+          rolloffFactor: 0
+        });
       }
       this.updateDistances();
     }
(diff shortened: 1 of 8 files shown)


src/components/audio-params.js
```diff
@@ -173,39 +177,40 @@ AFRAME.registerComponent("audio-params", {
...
-      );
-      this.data.distanceModel = audio.panner.distanceModel;
-      this.data.rolloffFactor = audio.panner.rolloffFactor;
-      this.data.refDistance = audio.panner.refDistance;
-      this.data.maxDistance = audio.panner.maxDistance;
-      this.data.coneInnerAngle = audio.panner.coneInnerAngle;
-      this.data.coneOuterAngle = audio.panner.coneOuterAngle;
-      this.data.coneOuterGain = audio.panner.coneOuterGain;
+      if (audio.panner) {
+        this.data.position = new THREE.Vector3(
```
johnshaughnessy, line 197: The contract for AFRAME components is to use `setAttribute` to set its `data`. If you set values on its `data` directly, subsequent calls to `setAttribute` may overwrite your changes (even if the subsequent call did not specify a value for the same key).

johnshaughnessy (changes requested):

src/components/audio-params.js
keianhzo: Fixed

src/components/audio-params.js
```diff
@@ -298,36 +307,42 @@ AFRAME.registerComponent("audio-params", {
...
     if (this.data.isClipped) {
       const audio = this.getAudio();
-      this.data.isClipped = false;
-      this.data.gain = this.data.preClipGain;
-      this.data.gain = Math.max(0.001, this.data.gain);
+      this.el.setAttribute("audio-params", {
+        isClipped: false,
+        gain: Math.max(0.001, this.data.preClipGain)
+      });
       audio?.gain?.gain.exponentialRampToValueAtTime(this.data.gain, audio.context.currentTime + MUTE_DELAY_SECS);
```
johnshaughnessy, line 346: We can remove this line because `gain` is applied in `update`

src/components/audio-params.js
```diff
@@ -298,36 +307,42 @@ AFRAME.registerComponent("audio-params", {
...
-      this.data.isClipped = true;
-      this.data.preClipGain = this.data.gain;
-      this.data.gain = gain;
-      this.data.gain = Math.max(0.001, this.data.gain);
+      this.el.setAttribute("audio-params", {
+        isClipped: true,
+        preClipGain: this.data.gain,
+        gain: Math.max(0.001, gain)
+      });
       audio.gain.gain.exponentialRampToValueAtTime(this.data.gain, audio.context.currentTime + MUTE_DELAY_SECS);
```
johnshaughnessy, line 335: We can remove this line because `gain` is applied in `update`

src/components/audio-params.js
```diff
@@ -173,39 +177,44 @@ AFRAME.registerComponent("audio-params", {
...
+      if (this.data.audioType === AudioType.PannerNode) {
+        this.audioRef.setDistanceModel(this.data.distanceModel);
+        this.audioRef.setRolloffFactor(this.data.rolloffFactor);
+        this.audioRef.setRefDistance(this.data.refDistance);
+        this.audioRef.setMaxDistance(this.data.maxDistance);
+        this.audioRef.panner.coneInnerAngle = this.data.coneInnerAngle;
+        this.audioRef.panner.coneOuterAngle = this.data.coneOuterAngle;
+        this.audioRef.panner.coneOuterGain = this.data.coneOuterGain;
+      }
       this.data.gain !== undefined && this.updateGain(this.data.gain);
```
johnshaughnessy, line 189: The `updateGain` function should not write to `this.data.gain`. (It should also not call `setAttribute` because then we would introduce an infinite loop here.) Instead, `updateGain` should apply the derived gain value to the underlying node, even if that differs from `this.data.gain`. If it is important for `this.data.gain` to be exactly equal to the gain of the underlying audio node, then whenever `update` receives an invalid gain node, it should call `setAttribute` to force it back to a valid value.

src/components/audio-params.js
```diff
@@ -173,39 +177,44 @@ AFRAME.registerComponent("audio-params", {
```
johnshaughnessy, line 177: This component is still attempting to write directly to `this.data` in its `init` method. The contract for setting component `data` is calling `setAttribute`. Basically, `this.data` is meant to be read-only.

johnshaughnessy (changes requested):

src/components/audio-params.js
keianhzo: We were actually updating the gain twice, once on the audio node gain and another one in the gain filter.

src/components/audio-params.js
keianhzo: Same as before, we were updating the gain twice, once on the audio node gain and another one in the gain filter.

src/components/audio-params.js
keianhzo: Updated

src/components/audio-params.js
keianhzo: Updated

keianhzo:
@johnshaughnessy I've updated the PR. Still there are a few things that doesn't feel right like calling `setAttribute` during tick to update the data from the audio element or having to correct the gain during `updateGain` in case we are in `audio` output mode. 

A lot of this is going to be fixed when we move away from the current "data book-keeping" design that so I'm going to leave it like this until the next audio iteration.

johnshaughnessy (commented):
I would like us to use a map from entities (`el`s) to objects with a shape something like this:
```
{
  audioNode: THREE.Audio || THREE.PositionAudio,
  defaultParams: { ... },
  sourceType
} 
```
With this map we can:
- remove the `audio-params` component
- remove the aframe component lifecycle concerns (i.e. the `setAttribute` / `update` / `this.data` conundrum)
- no longer keep a copy of the underlying state (that we are currently updating in `tick`)
- keep default values of audio sources separated from its "current values" (as influenced by things like audio zones). 

I'm concerned that the many small adjustments made to this code are not being made carefully or uniformly: 
- There is still code that overwrites a component property with assignment ( `this.data.gain = ...`  and with `getWorldDirection` / `getWorldPosition`).
- The function that adjusts gain has changed from `exponentialRampToValueAtTime` to `setTargetAtTime`. We determined that we could not use the former when the `gain` was `0`... but here you changed it for all values of `gain`. Is there an experiential difference between the two for certain `gain` values? Was `exponential` the preferred method of interpolation?
- In `audio-system`, this PR replaces calls to `exponentialRampToValueAtTime` to `setTargetAtTime` _even though the `newGain` values are prevented from being 0_! What was the benefit of the change in this case?
- Now that we call `setAttribute` to set component data, it has become especially awkward to update the component data in `tick` because `setAttribute` is slow and we generally discourage using it in hot loops (like `tick`). There does not seem to be a fundamental need to keep a second copy of this state in the component's `data`.

If this PR fixes the issues that are listed in its summary then I don't want to block its release. But I feel that we can remove quite a bit of complexity and book-keeping in this area if we take a step back and reconsider the tasks we are trying to accomplish with this code. I'd be happy to pair with you on this effort if you're interested.

keianhzo:
We have already discussed the redesign of the `audio-params` component and I agree with the proposal above and it's benefits but I don't want to delay the fixes in this PR anymore as they fix bugs that are affecting users in production so as long as this PR doesn't introduce new bugs I'm happy with landing it and addressing the audio redesign in a follow-up task.

johnshaughnessy (approved):

Issue #3658, opened 2021-01-06
Uploaded music files in a room continue to play at the same volume even when muted


Softvision-GeluHaiduc:
**[Affected Versions]:**
- Firefox Release 84.0.1
- Chrome 87.0.4280.88

**[Affected Platforms]:**
- Windows 10x64

**[Prerequisites]:**
- Have a new room created and entered.

**[Steps to reproduce]:**
1. Drag and drop a music file in the room or upload the file.
2. Click the “-” volume down button from the object until it is muted.
3. Observe the volume of the music.

**[Expected results]:**
- The volume is set to a lower and lower level until it is muted and no longer can be heard.

**[Actual results]:**
- The volume remains the same, the text displayed in the volume bar is “MUTE”

**[Notes]:**
- The volume remains the same when the “+” volume up button is clicked as well.
- Attached a screen recording of the issue [here](https://drive.google.com/file/d/1ZSvbXESTDFtG6YVKRyIWnxYcFCOfWwVC/view?usp=sharing).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-145)

johnshaughnessy:
This should block release until we understand what is happening here. (Should I add the [QA]:Blocker issue label?)

takahirox:
Probably this problem will be fixed by #4445 for Hubs Cloud, and already solved in hubs.mozilla.com.

Softvision-GeluHaiduc:
I have verified this issue using the latest releases of Firefox and Chrome on Windows 10.
Now, uploaded music files do not play at the same volume when muted.

PR #4445, opened 2021-07-24
Hubs Cloud: Fix broken media positional audio

takahirox:
In Hubs Cloud, somehow Media positional audio has been broken. This PR fixes it.

The root issue seems that the same two audio sounds come out for a media, one is a video element and another one is Three.js PositionalAudio (WebAudio panner). The former volume should be louder in general because no attenuation. So we feel that positional audio doesn't work although we just don't hear the positional audio because of the louder video element non-positional audio. This PR fixes this problem by muting the video element.

This change is a part of https://github.com/mozilla/hubs/commit/a53a59b1c19992a0758c48f3e4686018d8bd2b30 which is a commit already merged to master. So media positional audio works in hubs.mozilla.com.

@keianhzo Did you notice that the media positional audio was broken and fix it in the commit?

### DIFF
--- a/src/components/media-views.js
+++ b/src/components/media-views.js
@@ -862,6 +862,8 @@ AFRAME.registerComponent("media-video", {
         videoEl.onerror = failLoad;
 
         if (this.data.audioSrc) {
+          videoEl.muted = true;
+
           // If there's an audio src, create an audio element to play it that we keep in sync
           // with the video while this component is active.
           audioEl = createVideoOrAudioEl("audio");


keianhzo:
That was part of a big PR so hard to say but I guess we should have seen that bug at some point if it was happening with any audio element...

takahirox:
The QA team reported the issue in Jan https://github.com/mozilla/hubs/issues/3658 but unfortunately we didn't seem to notice... We should triage the issues more.

takahirox:
Regarding the solution in this PR (and that commit), to be honest I'm unsure if it is the best solution. I'm still wondering why we need to mute video element (sorry, I haven't fully understood the media view yet). I have never seen such WebAudio examples. So for me it is a temporal workaround for Hubs Cloud and we may need to refactor the media view later.

keianhzo:
This fixes the issue but I've some questions:

- Why are we creating a video element for audio only media and then muting the video element and creating a positional audio element?
- How did this ever work in the past if this video element was never being muted? I'll keep on digging for this one...

brianpeiris:
I think it may have been a requirement (in the past at least) that we create a video or audio element in order to use it with the WebAudio API. This may have been due to auto-play restrictions in the past, but I'm not sure if they still apply.
If we intend to attempt to simplify this code, I think it's safe to say that we should test these kinds of changes thoroughly in the various browsers that we support, because this code may have been put in place to workaround specific browser behaviors.

PR #4446, opened 2021-07-26
Disable audio debugger for WebGL < 2.0

keianhzo:
The audio debugging shader uses GLSL 3.00.0 ES features. This PR disables the shader for any device using WebGL < 2.0

### DIFF
--- a/src/systems/audio-debug-system.js
+++ b/src/systems/audio-debug-system.js
@@ -2,6 +2,7 @@ import { THREE } from "aframe";
 import audioDebugVert from "./audio-debug.vert";
 import audioDebugFrag from "./audio-debug.frag";
 import { DistanceModelType } from "../components/audio-params";
+import { getWebGLVersion } from "../utils/webgl";
 
 const MAX_DEBUG_SOURCES = 64;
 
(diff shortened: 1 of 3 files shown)


src/utils/webgl.js
```diff
@@ -34,3 +34,8 @@ export function patchWebGLRenderingContext() {
     };
   }
 }
+
+export function getWebGLVersion(renderer) {
+  const gl = renderer.getContext("webgl") || renderer.getContext("experimental-webgl");
```
takahirox, line 39: I don't think Three.js `WebGLRenderer.getContext()` takes an argument.

https://threejs.org/docs/#api/en/renderers/WebGLRenderer.getContext

```suggestion
  const gl = renderer.getContext();
```

keianhzo: Ops yep, fixed, thanks!

src/systems/audio-debug-system.js
```diff
@@ -143,7 +150,7 @@ AFRAME.registerSystem("audio-debug", {
   },
 
   enableDebugMode(enabled, force = false) {
-    if ((enabled === undefined || enabled === this.data.enabled) && !force) return;
+    if (((enabled === undefined || enabled === this.data.enabled) && !force) ||   this.unsupported) return;
```
takahirox, line 153: Nit: This line seems to have extra spaces.

```suggestion
    if (((enabled === undefined || enabled === this.data.enabled) && !force) || this.unsupported) return;
```

takahirox (approved):
WebGL2 detection looks good to me.

src/utils/webgl.js
```diff
@@ -34,3 +34,8 @@ export function patchWebGLRenderingContext() {
     };
   }
 }
+
+export function getWebGLVersion(renderer) {
```
takahirox, line 38: Nit: Just my preference. I prefer taking `gl`, rather than `renderer`, as an argument to reduce the dependency with Three.js because the file name is `webgl.js`. No blocking issue, just my preference.

PR #4451, opened 2021-07-27
Avoid rendering the render target texture.

keianhzo:
Chromium throws an error when trying to render the render target texture.
Fixes https://github.com/mozilla/hubs/issues/4438


netpro2k (commented):
Nice catch. I think this solution should work and should be pretty fast, but I wonder if we should just put render target on their own layer instead, Then we just make sure the video-texture-source cameras do not render that layer.

keianhzo:
I wonder if we wouldn't be over complicating the solution. All the scene cameras would need to render the default layer and the specific target object layer to render the whole scene, then the the render target would need to render only the default layer. So we would need to create all scene objects with both layers enabled just for the default cameras to render the scene and then make the render target just to render the default one.

Alternatively we could just disable all target object layers in the object hierarchy before rendering and then enabling them again which is pretty much the same as toggling their visibility.

keianhzo:
Updated to using layers.

src/components/video-texture-target.js
```diff
@@ -106,6 +107,12 @@ AFRAME.registerComponent("video-texture-target", {
   },
 
   init() {
+    // Make video-texture-target objects inivisible before rendering to the frame buffer
+    // Chromium checks for loops when drawing to a framebuffer so if we don't exclude the objects
+    // that are using that rendertarget's texture we get an error. Firefox does not check.
+    // https://chromium.googlesource.com/chromium/src/+/460cac969e2e9ac38a2611be1a32db0361d88bfb/gpu/command_buffer/service/gles2_cmd_decoder.cc#9516
+    this.el.object3D.traverse(o => o.layers.set(Layers.CAMERA_LAYER_VIDEO_TEXTURE_TARGET));
```
netpro2k, line 114: Probably doesn't really effect anything but this should probably be removed when the component is removed.

src/systems/camera-system.js
```diff
@@ -207,6 +204,9 @@ export class CameraSystem {
     this.mode = CAMERA_MODE_SCENE_PREVIEW;
     this.snapshot = { audioTransform: new THREE.Matrix4(), matrixWorld: new THREE.Matrix4() };
     this.audioSourceTargetTransform = new THREE.Matrix4();
+    scene.addEventListener("cameraready", ({ detail: { cameraEl } }) => {
+      cameraEl.getObject3D("camera").layers.mask |= LayerMasks.CAMERA_LAYER_VIDEO_TEXTURE_TARGET_MASK;
```
netpro2k, line 208: Can't you use `.enable(Layers.CAMERA_LAYER_VIDEO_TEXTURE_TARGET)` here?

src/components/layers.js
```diff
@@ -1,6 +1,17 @@
 export const Layers = {
   // Layers 0 - 2 reserverd by ThreeJS and AFrame.
-  reflection: 3
+  CAMERA_LAYER_REFLECTION: 3,
+  CAMERA_LAYER_INSPECT: 4,
+  CAMERA_LAYER_BATCH_INSPECT: 5,
+  CAMERA_LAYER_VIDEO_TEXTURE_TARGET: 6
+};
+
+export const LayerMasks = {
```
netpro2k, line 9: Hmm. What do we need this for? The THREE.Layers interface has methods for flipping individual bits.

netpro2k (commented):
I think we might be able to ditch LayerMasks, but other than that lgtm

src/systems/camera-system.js
keianhzo: That's true. I'm going to switch to enable and remove the Layers mask enum.

src/components/video-texture-target.js
keianhzo: Updated

PR #4461, opened 2021-07-30
Revert of "Fix OB blitting issues"

keianhzo:
This is already fixed in ThreeJS r128 so there is no need for this hack anymore.

Related: https://github.com/mozilla/hubs/pull/4255


takahirox (approved):
Reverting sounds good to me if it's fixed in Three.js r128

PR #4470, opened 2021-08-04
Fix wrong transparent object menu buttons and texts

### LINKED ISSUE #4416: The contents of the buttons in object menu can be tranaparent
**Description**

The contents of the buttons in object menu can be transparent.

**To Reproduce**

1. Go to hubs.mozilla.com
2. Place an thin object like gif
3. Make an angle of it and make some distance from it
4. Point the object and press space to open the object menu
5. Observe that the content of the buttons in the object menu can be transparent.

**Expected behavior**

They are opaque.

**Screenshots**

image

**Hardware** 
 - Device: Desktop
 - OS: Windows
 - Browser: Chrome

**Additional context**

They are opaque if I look at them at the front. The problem happens only if I make an angle and make some distance.

### LINKED ISSUE #4417: The around of the circled buttons in object menu can be weird
**Description**

The around of the circled buttons in object menu can be weird.

**To Reproduce**

1. Go to hubs.mozilla.com
2. Place a thin object
3. Make an angle and move it to a certain position
4. Point the object and press space to open an object menu
5. Observe that the around of the circled buttons in object menu can be weird

I'm really not sure the exact factor which causes this problem. If you can't reproduce the problem, please further try to move or rotate the object, or try another object.

**Expected behavior**

They are rendered correctly

**Screenshots**

image

image

**Hardware** 
 - Device: Desktop
 - OS: Windows
 - Browser: Chrome

**Additional context**

Could be render order issue?

takahirox:
This PR fixes #4416 and #4417 by applying `alphaTest` to `slice9` component and making `slice9` opaque for object menu buttons.

The screenshot of this PR. Please compare with the ones in #4416 and #4417.

image

Currently `slice9` component doesn't have `alphaTest` attribute so I made my fork.

If this change looks good and we merge, I will make a PR to `slice9` component. When it will be merged, I will update package.json and package-lock.json.

**The mechanism of the problems**

* The text is placed in front of the slice9 component and the slice9 component is placed in front of an object
* Currently all of them are somehow transparent and farther objects should be render first
* Three.js determines the render order by checking the distance of objects from the camera to generally the center of the objects
* If object group in which the text, the slice9 component, and the object are involved, is rotated and/or it is far from the camera, Three.js can wrongly determine the render order and render the texts or slice9 earlier than others (This is the limitation of the current efficient but non-perfect render order decision algorithm of Three.js)
* If it the wrong render order happens overlapped areas of slice9 and object are not rendered because of depth test. This is the reason why currently texts and the around of slice9 can be transparent to background

**How to solve**

* The `slice9` component seems to be expected to be transparent because the component renders to the plane geometry mesh and uses alpha blend to cut off the four corners
* But for that purpose, using alphaTest should be smarter and more efficient than alpha blend (=transparent). So I introduced `alphaTest` attribute to the `slice9` component and it resolves #4417 because the four corner will be discorded in the shader and (my understanding is) the depth will not be written neither.
* With alphaTest, we can make the `slice9` components for the object menu opaque. In Three.js opaque objects are rendered earlier than transparent object. So the render order between texts (transparent) and `slice9` (opaque) will be always correct and #4416 will be fixed.

**Note**

* The font we currently use seems to require `transparent`. So opaque texts can't be a solution unless we change the font.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15821,9 +15821,8 @@
       "integrity": "sha512-OQlwyK/U5ISEoEaZCOxCzkfqq6F1coUKXQuwkWMAmRmvKyAfrEyX63G8c87aPvZO35z+PItObcpYNtBmjk4SBQ=="
     },
     "aframe-slice9-component": {
-      "version": "1.0.0",
-      "resolved": "https://registry.yarnpkg.com/aframe-slice9-component/-/aframe-slice9-component-1.0.0.tgz",
-      "integrity": "sha1-+w+EQdrdHosRzCRRK6eqaS1iK+E="
+      "version": "github:takahirox/aframe-slice9-component#2ec0810cbf23f487a10c767d24dee6d5b7ede77f",
+      "from": "github:takahirox/aframe-slice9-component#AlphaTest"
     },
     "aggregate-error": {
       "version": "3.1.0",
--- a/package.json
+++ b/package.json
@@ -78,7 +78,7 @@
     "@popperjs/core": "^2.4.4",
     "aframe": "github:mozillareality/aframe#hubs/master",
     "aframe-rounded": "^1.0.3",
-    "aframe-slice9-component": "^1.0.0",
+    "aframe-slice9-component": "github:takahirox/aframe-slice9-component#AlphaTest",
     "ammo-debug-drawer": "github:infinitelee/ammo-debug-drawer",
     "ammo.js": "github:mozillareality/ammo.js#hubs/master",
     "animejs": "github:mozillareality/anime#hubs/master",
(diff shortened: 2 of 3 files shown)


takahirox:
Updated the PR comment.

netpro2k (commented):
The edges of the buttons definitely look a bit messier with this since they can't be anti-aliased but I think its a reasonable tradeoff, and things overall look better than with the bugs. Ultimately I think we want to rethink how our menus work pretty fundamentally (if we even want them at all) so this seems like a good near term fix.

takahirox:
> The edges of the buttons definitely look a bit messier with this since they can't be anti-aliased

Ah, good catch. But yeah I think the tradeoff is acceptable and this solution is reasonable.

PR #4471, opened 2021-08-04
Fix the wrong transparent object menu buttons

### LINKED ISSUE #4416: The contents of the buttons in object menu can be tranaparent

takahirox:
This PR fixes #4416 by introducing `renderOrder` attribute to `hubs-text` and rendering the texts of object menu buttons at last.

The screenshot with this PR. Please compare with the one in #4416.

image

My understanding of the mechanism how the problem was caused and how this PR fixes is

* The text is placed in font of the slice9 and object
* All of them are transparent and farther objects should be render first
* Three.js determines the render order by checking the distance of objects from the camera to generally the center of the objects
* If object group is rotated and/or it is far from the camera, Three.js can wrongly render the texts earlier than object and slice9 (This is the limitation of the current render order decision algorithm)
* If it happens overlapped area of slice9 and object are not rendered because of depth test. This is the reason why currently texts can be transparent to background
* An easy solution would be to force the render order of texts to render them at last by using [`Three.js Object3D.renderOrder`](https://threejs.org/docs/#api/en/core/Object3D.renderOrder)

### DIFF
--- a/src/components/hubs-text.js
+++ b/src/components/hubs-text.js
@@ -109,6 +109,7 @@ AFRAME.registerComponent("text", {
     // `negate` must be true for fonts generated with older versions of msdfgen (white background).
     negate: { type: "boolean", default: true },
     opacity: { type: "number", default: 1.0 },
+    renderOrder: { type: "number", default: 0.0 },
     side: { default: "front", oneOf: ["front", "back", "double"] },
     tabSize: { default: 4 },
     transparent: { default: true },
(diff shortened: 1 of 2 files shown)


takahirox:
Ah, wait. Making the slice9 opaque would be simpler solution (but it requires #4470). Let me try tomorrow.

takahirox:
Closing in favor of #4470

PR #4478, opened 2021-08-04
Teleport fix

keianhzo:
We were making the navMesh invisible when disabling the Audio Debug view and the Raycaster doesn't intersect with invisible objects. Changing the parent's visibility allows us to show/hide while keeping the actual navMesh intersectable

### DIFF
--- a/src/systems/audio-debug-system.js
+++ b/src/systems/audio-debug-system.js
@@ -162,7 +162,6 @@ AFRAME.registerSystem("audio-debug", {
       navMesh.visible = enabled;
       navMesh.traverse(obj => {
         if (obj.isMesh) {
-          obj.visible = enabled;
           if (obj.material) {
             if (enabled) {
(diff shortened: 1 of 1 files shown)


takahirox (approved):
I confirmed that this change fixes the problem.

FYI: About the visibility for raycasting, the latest Three.js raycasting determines whether it should ignore or not based on layers not visibility. But we have a lot of code relying on the old raycasting behavior which checks visibility not layers. So we applied the patch to Three.js to keep checking visibility. At some point we may update our code and remove the patch.

Refer to 
https://github.com/mozilla/hubs/blob/1a2668f20ad0b740a60f62758aa6c3ce2c18553a/src/utils/threejs-raycast-patches.js for the detail.

Note: Raycasting doesn't go into depth (children) if an object is invisible. So keep in mind that we shouldn't pass invisible navMesh to the raycaster instead of underlying visible meshes.

PR #4487, opened 2021-08-06
Make sure audio-params is available when media-video is created

### LINKED ISSUE #4456: Cannot read property 'data' of undefinedCannot read property 'data' of undefined
STRs:
- Enter in a Hubs room in two different sessions
- Create a Youtube video media element
- Check the console on the other end

The "audio-params" component in the "media-video" component seems not to initialize in the same order in local and remote.

```
media-views.js:904 Uncaught (in promise) TypeError: Cannot read property 'data' of undefined
    at n.updateVolumeLabel (media-views.js:904)
    at media-views.js:320
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-62)

### LINKED ISSUE #4472: Videos and GIFs objects appear broken when using the Magnify feature
**[Affected Versions]:**
- Safari 14.6
- Chrome 92
- Firefox 90.0.2

**[Affected Platforms]:**
- Windows 10
- Iphone SE 2

**[Prerequisites]:**
- Have a room opened on "https://hubs.mozilla.com/".
- Have a video or a GIF file spawned into the room.

**[Steps to reproduce]:**
1. Hover over the video or GIF object.
2. Press the Space key to display the object's menu.
3. Click the Magnify button from the menu.
4. Observe the low view of the object displayed.

**[Expected results]:**
- The view of the object is correctly displayed.

**[Actual results]:**
- A black view of the object is displayed.

**[Notes]:**
- The issue is not reproducible with images and PDF objects.
- The issue is not reproducible on Hubs Cloud production or beta.
- Attached is a screen recording of the issue: [link](https://drive.google.com/file/d/1KfRt4liiqHGh1moFXZE8DTKexxVYVj0T/view?usp=sharing).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-58)

### LINKED ISSUE #4477: The media objects volume bar is wrongly displayed at 100% when created
**[Affected Versions]:**
- Safari 14.6
- Chrome 92
- Firefox 90.0.22

**[Affected Platforms]:**
- Windows 10
- IphoneSE 2

**[Prerequisites]:**
- Have a room opened on "https://hubs.mozilla.com/".

**[Steps to reproduce]:**
1. Spawn a video or GIF file inside the room.
2. Hover over the object.
3. Observe the volume bar.

**[Expected results]:**
- The volume bar is displayed at 50%.

**[Actual results]:**
- The volume bar is displayed at 100%.

**[Notes]:**
- If the "+" or "-" buttons are clicked the volume bar jumps back to 50%.
- This issue is not reproducible on Hubs Cloud.
- Attached is a screen recording of the issue:
VolumeBar

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-43)

keianhzo:
After creating the `media-video` component from `media-loader`A-Frame calls `update` then we are calling `updateSrc` asynchronously which calls `setupAudio` and that's where the exceptions is thrown. 

My understanding is that there is some race condition happening that prevents the `audio-params` component to be created before the `setupAudio`call happens but it only happens in two scenarios that I know of, one is this one during a `linked-media` entity creation (#4472) and the other one is during the creation of a remote `media-video` (#4456). 

It's is unclear to me why is this happening only in those scenarios and not when creating a local `media-video` entity.

This PR moves the `audio-params` component creation from `media-loader` to `media-video` init/remove (as that's the only media view where we really need it) and that seems to guarantee that the component is available when `setupAudio` is called.
- Fixes #4472
- Fixes #4456
- Fixes #4477 We were not initializing the `audio-params` component with the default parameters for drag&drop media which doesn't have an `audio-params`.  Also I was tempted to add a to see  check if we already have and `audio-params` component to make sure we don't override the audio parameters for components were the `audio-params` component but as far as I know that could only happen in case the GLTF file has a `media-video` component declared before the `audio-params` component by manual manipulation so it should be a case.

### DIFF
--- a/src/components/media-loader.js
+++ b/src/components/media-loader.js
@@ -171,7 +171,6 @@ AFRAME.registerComponent("media-loader", {
     this.el.removeAttribute("gltf-model-plus");
     this.el.removeAttribute("media-pager");
     this.el.removeAttribute("media-video");
-    this.el.removeAttribute("audio-params");
     this.el.removeAttribute("audio-zone-source");
     this.el.removeAttribute("media-pdf");
     this.el.setAttribute("media-image", { src: "error" });
@@ -350,7 +349,6 @@ AFRAME.registerComponent("media-loader", {
       this.el.removeAttribute("gltf-model-plus");
       this.el.removeAttribute("media-pager");
       this.el.removeAttribute("media-video");
-      this.el.removeAttribute("audio-params");
       this.el.removeAttribute("audio-zone-source");
       this.el.removeAttribute("media-pdf");
       this.el.removeAttribute("media-image");
@@ -464,7 +462,6 @@ AFRAME.registerComponent("media-loader", {
             linkedMediaElementAudioSource
           })
         );
-        this.el.setAttribute("audio-params", {});
         this.el.setAttribute("audio-zone-source", {});
         if (this.el.components["position-at-border__freeze"]) {
           this.el.setAttribute("position-at-border__freeze", { isFlat: true });
@@ -475,7 +472,6 @@ AFRAME.registerComponent("media-loader", {
       } else if (contentType.startsWith("image/")) {
         this.el.removeAttribute("gltf-model-plus");
         this.el.removeAttribute("media-video");
-        this.el.removeAttribute("audio-params");
         this.el.removeAttribute("audio-zone-source");
         this.el.removeAttribute("media-pdf");
         this.el.removeAttribute("media-pager");
@@ -519,7 +515,6 @@ AFRAME.registerComponent("media-loader", {
       } else if (contentType.startsWith("application/pdf")) {
         this.el.removeAttribute("gltf-model-plus");
         this.el.removeAttribute("media-video");
-        this.el.removeAttribute("audio-params");
         this.el.removeAttribute("audio-zone-source");
         this.el.removeAttribute("media-image");
         this.el.setAttribute(
@@ -554,7 +549,6 @@ AFRAME.registerComponent("media-loader", {
       ) {
         this.el.removeAttribute("media-image");
         this.el.removeAttribute("media-video");
-        this.el.removeAttribute("audio-params");
         this.el.removeAttribute("audio-zone-source");
         this.el.removeAttribute("media-pdf");
         this.el.removeAttribute("media-pager");
@@ -589,7 +583,6 @@ AFRAME.registerComponent("media-loader", {
       } else if (contentType.startsWith("text/html")) {
         this.el.removeAttribute("gltf-model-plus");
         this.el.removeAttribute("media-video");
-        this.el.removeAttribute("audio-params");
         this.el.removeAttribute("audio-zone-source");
         this.el.removeAttribute("media-pdf");
         this.el.removeAttribute("media-pager");
--- a/src/components/media-views.js
+++ b/src/components/media-views.js
@@ -55,7 +55,7 @@ for (let i = 0; i <= 20; i++) {
 
 import { KTX2Loader } from "three/examples/jsm/loaders/KTX2Loader";
 import { rewriteBasisTranscoderUrls } from "../utils/media-url-utils";
-import { AudioType } from "./audio-params";
+import { AudioType, MediaAudioDefaults } from "./audio-params";
 const loadingManager = new THREE.LoadingManager();
 loadingManager.setURLModifier(rewriteBasisTranscoderUrls);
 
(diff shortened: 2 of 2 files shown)


takahirox (commented):
> My understanding is that there is some race condition happening that prevents the audio-params component to be created before the setupAudiocall happens

`media-loader` first set `media-video` and then set `audio-params`.

https://github.com/mozilla/hubs/blob/542591827c5bb382b85f69c9ef1e6157157a745b/src/components/media-loader.js#L455-L467

And `media-video` can wait depending on the condition before calling `setupAudio()` here

https://github.com/mozilla/hubs/blob/55f12c3b52626471cc7edb7e0de2c7027707d31b/src/components/media-views.js#L575-L583

This await seems to cause the race condition. If waiting, `audio-params` will be defined before calling `setupAudio()` then no error. Otherwise, causing the error.

Maybe the easiest solution would be changing the order in

https://github.com/mozilla/hubs/blob/542591827c5bb382b85f69c9ef1e6157157a745b/src/components/media-loader.js#L455-L467

to first set `audio-params` and then set `media-video`. But I don't fully understand the media related components so I can't say which is better.

And, with this PR or the order change, the following warnings and error will show up in the console if I test #4472. Better to fix them before merging.

image

keianhzo:
> media-loader first sets media-video and then sets audio-params.

Moving `this.el.setAttribute("audio-params", {});` before `this.el.setAttribute("media-video"...);` fixes the error for the magnifier issue but not for the remote instantiation issue (https://github.com/mozilla/hubs/issues/4456) as the `media-view` component is updated before that `media-loader` code that creates the `audio-params` component is called, that's why I moved the `audio-params` creation inside `media-view` initialization.

What happens in the networked instantiation scenario is:
1. `media-loader` update method is called
2. As we can't guess the content type for remote media there is an async call here:
https://github.com/mozilla/hubs/blob/542591827c5bb382b85f69c9ef1e6157157a745b/src/components/media-loader.js#L407
3. In the meantime `media-video` is initialized (I guess that's coming from the networked scheme?) and `updateVolumeLabel` is called which leads to the exception.
4. Then later on when the `fetchContentType` async call is has completed the `audio-params` component is created when `this.el.setAttribute("audio-params", {});` is called from `media-loader`
https://github.com/mozilla/hubs/blob/542591827c5bb382b85f69c9ef1e6157157a745b/src/components/media-loader.js#L467

That's why your solution works for the magnifier case but not for the networked instatiation case. All this async initialization code is really confusing and error prone.

keianhzo:
Regarding:

> I confirmed the both approaches resolve the problems but the both cause the following warnings and error in the console if I test #4472. Better to fix them before merging (if they are caused by this change).

This is because we are passing come properties when magnifying that are not in the `media-video` schema:

https://github.com/mozilla/hubs/blob/542591827c5bb382b85f69c9ef1e6157157a745b/src/components/media-loader.js#L455-L466

This was added a while ago in this commit https://github.com/mozilla/hubs/pull/2081/commits/970bb81cda98e55b4d4851e680dfe42b77fc87dc  but the schema properties were never added (maybe because A-Frame doesn't support those types?). Adding them and defaulting to null, seems to at least remove the error, not sure if that's something that we want to do though. Maybe someone else can help to solve that question, I'll ask around.

takahirox:
Thanks for the explanation in the details. I understand that the order change doesn't solve all the cases. 

> All this async initialization code is really confusing and error prone.

Yeah, it is very problematic.. 

So, I don't perfectly understand all the audio related components yet, but this change sounds ok to me. But please do double-check because of no confidence /cc @netpro2k @brianpeiris And we may need to refactor this initialization to clarify and simplify later. (Or does your @keianhzo WIP audio refactoring include the audio related components initialization refactoring?)

Regarding the console warnings, it sounds like that the problem has existed even before this change so I don't think it's a blocking issue for this PR. Let's fix in another PR. Adding the properties doesn't seem to solve the `[networked] component` error. Do you (or someone else) think this is also the existing issue?

keianhzo:
Yes, I'm working with @johnshaughnessy in a broader refactor that helps us move away from A-Frame and make this initialization code a bit more predictable. This is just a quick fix to solve production issues in the meantime.

takahirox (approved):
> Yes, I'm working with @johnshaughnessy in a broader refactor that helps us move away from A-Frame and make this initialization code a bit more predictable. This is just a quick fix to solve production issues in the meantime.

OK, then this change looks good to me as a quick fix. Thanks for resolving the problem!

takahirox:
> In the meantime media-video is initialized (I guess that's coming from the networked scheme?) 

It seems yes. It seems to directly create `media-video` component. So moving `audio-param` component creation into the inside of `media-video` makes sense for this scenario.

takahirox:
Opened #4499 for the console error

Issue #4405, opened 2021-07-12
[Safari] The window freezes when uploading video, GIF and audio files


Softvision-RemusDranca:
**[Affected Versions]:**
- Safari 14.0.2

**[Affected Platforms]:**
- macOS Catalina 10.15.7
- IphoneSE

**[Prerequisites]:**
- Have a room opened on "https://hubs.mozilla.com/".

**[Steps to reproduce]:**
1. Open the Chat box.
2. Insert a link for a video in the Chat input field.
3. Click the Create button.
4. Observe the behavior.

**[Expected results]:**
- The video is spawned into the room and played.

**[Actual results]:**
- The window freezes.

**[Notes]:**
- The issue is reproducible when uploading the files from the Place menu.
- The issue is also reproducible when the files are attached.
- The audio can still be heard in the background after the window freezes.
- Attached is a screen recording of the issue: [link](https://drive.google.com/file/d/1xUMnqKA4fZ0r1HoWhsbFK4swuKAd4X0l/view?usp=sharing).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-73)

takahirox:
Thanks for the report.

Any console errors?

> IphoneSE

I couldn't reproduce the error on iPhone X + iOS 14.6 + Safari 14.1.1.

Softvision-RemusDranca:
@takahirox [Here](https://drive.google.com/file/d/1exRNZpW91auZUx8nrPxPkDRXthCOGlSO/view?usp=sharing) are the console logs received while reproducing the issue on MacOS using Safari 14.0.2.

takahirox:
Thanks for the console logs.

This seems the error causing the problem.

> [Error] TypeError: undefined is not an object (evaluating 'e.panner.positionX.value') (frontend-1b3139bed401fce92cce.js:123:6827)

brianpeiris:
Was not able to reproduce this in Safari 14.1.1, on MacOS 11.4, on a Macbook Air (M1, 2020). I uploaded this video from the chat sidebar: https://github.com/brianpeiris/test-media/blob/main/big-buck-bunny-360p-short.mp4 and I also tried the vimeo video you used in your screen recording and I used the same Outdoor Workshop scene.
The video uploaded correctly without freezing the app, and it played correctly with audio.
[...]

PR #4503, opened 2021-08-10
Disable positional audio for Safari as a workaround for its positional audio bug

takahirox:
Currently audio in Hubs is choppy and distorted on Safari. It seems being caused by Safari Positional Audio bug. And the bug seems to have a bad impact to performance. (Refer to #4411).

We have reported the bug to Safari. And we have applied the temporal fix in #4442 by disabling positional audio for Hubs Cloud. Positional audio is one of the important features for us but non-positional audio should be much better than broken audio.

I would like to suggest to apply the same change to `master` (hubs.mozilla.com) because Safari users who tries hubs.mozilla.com may think Hubs is useless due to this problem and this problem blocks a lot of tests of the upstream on Safari.

@keianhzo Can you review to check if this temporal fix can affect new audio features you recently added?

### DIFF
--- a/src/hub.js
+++ b/src/hub.js
@@ -42,7 +42,7 @@ import "./utils/threejs-positional-audio-updatematrixworld";
 import "./utils/threejs-world-update";
 import "./utils/threejs-raycast-patches";
 import patchThreeAllocations from "./utils/threejs-allocation-patches";
-import { detectOS, detect } from "detect-browser";
+import { isSafari } from "./utils/detect-safari";
 import {
   getReticulumFetchUrl,
   getReticulumMeta,
@@ -670,11 +670,9 @@ document.addEventListener("DOMContentLoaded", async () => {
     return;
   }
 
-  const detectedOS = detectOS(navigator.userAgent);
-  const browser = detect();
   // HACK - it seems if we don't initialize the mic track up-front, voices can drop out on iOS
   // safari when initializing it later.
-  if (["iOS", "Mac OS"].includes(detectedOS) && ["safari", "ios"].includes(browser.name)) {
+  if (isSafari()) {
     try {
       await navigator.mediaDevices.getUserMedia({ audio: true });
     } catch (e) {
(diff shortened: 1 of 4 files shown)


brianpeiris (approved):
The code changes look good to me, but please actually test this in Safari. The audio code in master is very different and I'm not sure if the audioOutputMode flag is working correctly there.

keianhzo (changes requested):
One thing that we need to do here is to check if the output mode is stereo before creating the THREE audio instances because I think historically we haven't really maintained the `audioOutputMode` path so we are not really considering that right now.
So wherever we create a THREE audio instance we need to check first what the global `audioOutputMode` is to override the element's `audio-params `and create the right audio instance. 

I'm pushing a commit with that change.

keianhzo:
I'm requesting @johnshaughnessy review for my latest changes. We will need to factor in this patch for the audio refactor and if we remove the `audioOutputMode` pref we will need to keep track of this special case to instantiate the right THREE audio type.

@takahirox @brianpeiris I guess this updated patch should also be applied uplifted to HC?

Also it seems there are a number of media playback issues in Safari:
- Remote media: https://github.com/mozilla/hubs/issues/4516
- Uploaded media: https://github.com/mozilla/hubs/issues/4529
- Froze window: https://github.com/mozilla/hubs/issues/4405

takahirox:
Sorry, a bit confused. Have we dropped `autioOutputMode` although `/audiomode` command still seems to exist?

> @takahirox @brianpeiris I guess this updated patch should also be applied uplifted to HC?

I'm not sure if I understand the context correctly but I think Hubs Cloud still have `audioOutputMode` support because it seems to have been dropped just a few commits ago?

johnshaughnessy:
Hello @takahirox I would like to work on this problem with you. 

It is news to me that the `/audiomode` command regressed a while back. I would like to help fix it. I have been working with Manuel on some of the other audio related code. It should be easy for us to fix the `audioOutputMode` pref and also to use `Audio`s instead of `PositionalAudio`s on safari if that's what we want to do.

The work that Manuel and I have been doing to simplify the way we handle audio nodes and audio settings is here: https://github.com/mozilla/hubs/pull/4562

It would be convenient for me if we work together and avoid merge conflicts.

johnshaughnessy:
Takahiro and I paired to come up with an alternative fix (https://github.com/mozilla/hubs/pull/4594), so I am closing this PR.

PR #4508, opened 2021-08-11
Fix Audio Debug view not working for Blender nav meshes

### LINKED ISSUE #4506: Audio Debugging view doesn't work for Blender addon Nav meshes
**Description**
The debug view doesn't work for Blender exported scenes.

**To Reproduce**
1. Go to Blender
2. Export a scene with a nav mesh
3. Load the scene in Hubs
4. Enable the debug view
5. See that the Audio debug view doesn't draw anything

**Expected behavior**
The Audio debug view should be shown

**Additional context**
The Spoke generated nav meshes are tagged with `navMesh` or `triMesh` class but the ones coming from the Blender addon are tagged with `FloorPlan` class. We have to add support for that in the Debug View. We should probably look for the `nav-mesh` component instead.

keianhzo:
Fixes #4506
The Spoke generated nav meshes are have `navMesh` or `triMesh` class but the ones coming from the Blender add-on have `FloorPlan` class. 

Look for a `nav-mesh` component instead to be more consistent and cover all the cases.


src/systems/audio-debug-system.js
```diff
@@ -3,6 +3,7 @@ import audioDebugVert from "./audio-debug.vert";
 import audioDebugFrag from "./audio-debug.frag";
 import { DistanceModelType } from "../components/audio-params";
 import { getWebGLVersion } from "../utils/webgl";
+import { getMeshes } from "../components/teleporter";
```
brianpeiris, line 6: I'd suggest moving getMeshes out of teleporter.js and into a shared utility module, since it isn't actually specific to teleportation.

src/systems/audio-debug-system.js
```diff
@@ -154,14 +156,16 @@ AFRAME.registerSystem("audio-debug", {
     this.zones.forEach(zone => {
       zone.el.setAttribute("audio-zone", "debuggable", enabled);
     });
-    const envRoot = document.getElementById("environment-root");
-    const meshEl = envRoot.querySelector(".trimesh") || envRoot.querySelector(".navMesh");
-    if (meshEl) {
+
+    const collisionEntities = [].slice.call(this.el.sceneEl.querySelectorAll("[nav-mesh]"));
```
brianpeiris, line 160: Would be more idiomatic to use `Array.from()` here

src/systems/audio-debug-system.js
```diff
@@ -154,14 +156,16 @@ AFRAME.registerSystem("audio-debug", {
...
+    const meshes = getMeshes(collisionEntities);
+
+    if (meshes) {
```
brianpeiris, line 163: Isn't this always true? Maybe you meant `if (meshes.length)` instead?

brianpeiris (commented):
Minor comments. LGTM otherwise.

src/systems/audio-debug-system.js
keianhzo: Refactored to aframe-utils.js

src/systems/audio-debug-system.js
keianhzo: Yep, that's wrong. Fixed

src/systems/audio-debug-system.js
keianhzo: Updated

Issue #4510, opened 2021-08-11
Teleport is broken when the nav-mesh node is transformed


rawnsley:
[This room](https://hubs.mozilla.com/mFzsjfS/intelligent-ecstatic-gala/?debugNavmesh) is an example of the problem using [this simple scene file](https://data.avncloud.com/activities/735908/files/nav-teleport-bug.glb?date=1628690785&size=246488&md5=5f833092782777c04fc3759ee27d0f70) in which the nav-mesh node is visible and has been translated by 10m down. Note that the waypoint navigation mesh is shown in wireframe using debugNavmesh, but it is offset from the textured version.

<img width="1022" alt="Screenshot 2021-08-11 at 15 07 54" src="https://user-images.githubusercontent.com/303516/129044230-a91de9a0-27bf-4c4b-b83e-0ac44a67a0c2.png">

You can see the nav mesh below the start position and walking is in fact constrained by it, but you remain floating above. This is a problem for teleporting because it is keyed off the visible version of the mesh.

This is what the model looks like in glTF Viewer, where you can see the cylinder intersects the nav mesh:

<img width="951" alt="Screenshot 2021-08-11 at 15 08 12" src="https://user-images.githubusercontent.com/303516/129044333-c9602aa6-ed64-4c53-a720-88bf0a358ab2.png">

[Blender file](https://data.avncloud.com/activities/735908/files/nav-teleport-bug.blend?date=1628690819&size=905620&md5=877e9dcdbbc0d2c0d8235b6f1364b3f1) for reference.

PR #4513, opened 2021-08-11
Copy mesh geometry to prevent the transform matrix being applied twice

rawnsley:
Fixes https://github.com/mozilla/hubs/issues/4510

[This recent update to three.js ](https://github.com/mozilla/hubs/commit/4ea709ede09bce3cb1ff580916ed55261d502aaf#diff-e71a77a47ab916709aaf35571578358c7cb73db0d1970c008236a9ced02c6955L127) means that the transform matrix is applied directly to the underlying geometry rather than to a copy. This leads to the transform being applied twice to the three.js view of the nav-mesh and causing teleport problems and display problems in the rare cases where the nav-mesh is visible.

### DIFF
--- a/src/systems/nav.js
+++ b/src/systems/nav.js
@@ -14,7 +14,7 @@ AFRAME.registerSystem("nav", {
       console.error("tried to load multiple nav meshes");
       this.removeNavMeshData();
     }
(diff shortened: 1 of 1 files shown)


keianhzo (approved):
LGTM

PR #4535, opened 2021-08-17
Record logs for debugging

rawnsley:
_This an illustrative suggestion for improving debugging as discussed here: https://github.com/mozilla/hubs/discussions/4533_

When the `record_log` URL parameter is present, all log entries are stored in memory and can be downloaded using the "Save Logs" button that appears on the loading screen and in the support menu.

Screenshot 2021-08-17 at 14 04 49   Screenshot 2021-08-17 at 14 05 08

The `console` functions are intercepted and recorded ready for download as a JSON array like this

```
[
{"time":"2021-08-17T13:04:46.969Z","level":"log","args":["A-Frame Version: https://github.com/MozillaReality/aframe"]},
{"time":"2021-08-17T13:04:46.969Z","level":"log","args":["three Version: https://github.com/MozillaReality/three.js"]},
{"time":"2021-08-17T13:04:46.969Z","level":"log","args":["WebVR Polyfill Version:","^0.10.10"]},
{"time":"2021-08-17T13:04:46.988Z","level":"log","args":["%ccore:schema:warn %cDefault value `null` does not match type `string` in component `media-loader`%c ","color: orange","color: inherit","color: orange"]},
{"time":"2021-08-17T13:04:46.988Z","level":"log","args":["%ccore:schema:warn %cDefault value `null` does not match type `string` in component `media-loader`%c ","color: orange","color: inherit","color: orange"]},
{"time":"2021-08-17T13:04:46.988Z","level":"log","args":["%ccore:schema:warn %cDefault value `null` does not match type `string` in component `media-loader`%c ","color: orange","color: inherit","color: orange"]},
{"time":"2021-08-17T13:04:47.014Z","level":"info","args":["%cDownload the React DevTools for a better development experience: https://fb.me/react-devtools","font-weight:bold"]},
...
]

```
This can be read directly or displayed in a viewer such as the example provided in **record-log-viewer.html** as shown below

Screenshot 2021-08-17 at 14 07 58

Where possible the style and formatting has been matched closely to the typical browser console window.

Regarding the choice of storage format: there is no agreed standard and all the browsers do things differently. In most cases data and formatting is lost during the saving process. In addition to unifying the output from different browsers, this format also allows for easy parsing and comparison of multiple logs. It should be possible to accurately collate logs from multiple participants in a room, provided local clock drift is not too significant.

Limitations of this approach:
1. Call stack information is less available than in the console window, but it is captured for warnings and errors.
2. Browsers do not allow the interception of some messages, such as CORS violations, for security reasons.
3. Putting this functionality behind a URL parameter means the user must be directed to reload the page in some way. This is an unacceptable level of friction, but arguably necessary while this approach is evaluated.

When coupled with things like the `debug` flag, which adds considerable information to the log stream, this could be a powerful way to diagnose users problems.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-777)


yakyouk:
Very useful! how about integrating it with the `debug_log` flag? and for convenience for end users, make the debug log panel minimized by default in `debug-log.js` with:
```  let minimized = true;
  debugLog.className = minimized ? "minimized" : "";```
You'd probably also want the button on the `ExitedRoomScreenContainer`.

rawnsley:
> Very useful! how about integrating it with the `debug_log` flag? and for convenience for end users, make the debug log panel minimized by default in `debug-log.js` with:
> 
> ```
>   let minimized = true;
>   debugLog.className = minimized ? "minimized" : "";
> ```
> 
> You'd probably also want the button on the `ExitedRoomScreenContainer`.

That could work, although I'm trying not to modify too much of the existing code if I can help it. Your comment made me realize I didn't mean to actually delete debug-log.js in this branch, so I've restored it now. Honestly I'm not sure what debug-log does that you can't get from the developer console, but perhaps it was part of some larger plan like this one.

This button has already helped me a couple of times, particularly on platforms like Oculus Quest where it's a pain to get logs otherwise.

yakyouk:
I guess debug log is nice for a quick glance at the logs where it's not easy to get a console like on mobile, but I find the fact that it's expanded by default not very user-friendly.

>  particularly on platforms like Oculus Quest where it's a pain to get logs otherwise.

Agreed, and most importantly, like you mentioned, it's a great way to get logs from non techie users

rawnsley:
Here's [an example help request](https://discord.com/channels/498741086295031808/819204581140856832/891496894889619557):

> Tried to get my class of 31 to login to hubs, but many could not. I was running two rooms on two separate computers...kids had chrome books. Many just stuck in the boot up mode with the blue circle spinning. Any ideas to make this work..I was using the digital narrative room...was that the problem or the chromebooks. 

We can't help this user without logs and we can't ask a classroom full of kids to open the developer console.

rawnsley:
I'm withdrawing this PR as the main branch has drifted and it there's no point keeping it open if it's not wanted.

PR #4538, opened 2021-08-18
HDR Lightmaps & Environment Settings System

netpro2k:
This PR adds support for RGBE (.hdr) lightmaps and add a system to deal with environment settings. It also introduces the `environment-settings` component, which for now allows setting tone mapping, exposure, and custom equirectangular environment maps and backgrounds (also optionally HDR).

Currently only a single environment map is supporter per scene, but the plan would be to eventually allow multiple environment maps either by object or area (TBD)

This PR also reworks how references to GLTF types work on `MOZ_hubs_components`. Now instead of just listing the index, a special marker object is used so that the application loading the GLTF does not need to be aware of the component schema to be able to resolve the reference. This is important as we begin to add more tooling to our pipeline as there will be more things that need to "pass through" component data without breaking it. For now this applies to Spoke and the WIP "Hubs GLB tools". 

The marker object looks like:
```json
{ "__mhc_link_type": "texture", "index": 1}
```
This also begins to transition MOZ_hubs_components implementation to an actual GLTFLoader plugin, though right now does not change any of the existing "inflation" code.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-776)

### DIFF
--- a/src/components/gltf-model-plus.js
+++ b/src/components/gltf-model-plus.js
@@ -7,6 +7,7 @@ import { MeshBVH, acceleratedRaycast } from "three-mesh-bvh";
 import { disposeNode, cloneObject3D } from "../utils/three-utils";
 import HubsTextureLoader from "../loaders/HubsTextureLoader";
 import { KTX2Loader } from "three/examples/jsm/loaders/KTX2Loader";
+import { RGBELoader } from "three/examples/jsm/loaders/RGBELoader";
 
 THREE.Mesh.prototype.raycast = acceleratedRaycast;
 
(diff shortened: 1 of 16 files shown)


src/components/gltf-model-plus.js
```diff
@@ -398,6 +399,21 @@ class GLTFHubsPlugin {
...
+          return Promise.all(
+            parser._invokeAll(function(ext) {
+              return ext[hookName] && ext[hookName](idx);
+            })
+          );
+        })
+      );
+    }
+
+    // TODO decide if thse should get put into the GLTF loader itself
```
takahirox, line 415: Nit: typo
```suggestion
    // TODO decide if these should get put into the GLTF loader itself
```

PR #4565, opened 2021-08-23
Fix admin 403 images

robinkwilson:
HMC admin panel images were showing a 403 because the root url was hosting the images. This PR changes it so the admin panel images are being requested from the UPLOADS_HOST url.

Adds UPLOADS_HOST defined in Jenkins for HMC.
Will need to define UPLOADS_HOST in HC template for HC before sending deploying it to HC.

### DIFF
--- a/admin/src/react-components/fields.js
+++ b/admin/src/react-components/fields.js
@@ -68,6 +68,7 @@ OwnedFileImage.propTypes = {
 };
 
 function OwnedFileDownloadFieldInternal({ fileName, record, source }) {
+  console.log("OwnedFileDownloadFieldInternal()");
   return (
     <a
       download={fileName || true}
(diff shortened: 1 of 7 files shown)


admin/src/react-components/fields.js
```diff
@@ -2,7 +2,7 @@ import React from "react";
 import PropTypes from "prop-types";
 import { withStyles } from "@material-ui/core/styles";
 import LaunchIcon from "@material-ui/icons/Launch";
-import { getReticulumFetchUrl } from "hubs/src/utils/phoenix-utils";
+import { getReticulumFetchUrl, getAssetsFetchUrl } from "hubs/src/utils/phoenix-utils";
```
brianpeiris, line 5: Maybe `getUploadsUrl` instead of `getAssetsFetchUrl` ?

brianpeiris (approved):
Minor comment, LGTM otherwise.

PR #4569, opened 2021-08-25
Added more console logging to track progress during loading

rawnsley:
Collecting more information in the console logs makes debugging much easier, particularly outside of a development environment. These example additions were used to narrow down a hang during launch, but they will be generally useful in future so I thought I would submit them back upstream.

When coupled with better log gathering practices (such as [this suggestion](https://github.com/mozilla/hubs/pull/4535)), simple instrumentation can be a surprisingly powerful debugging tool.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-774)


netpro2k (approved):
We will need to come up with some more coherent logging, but this looks good. Thanks for the PR!

PR #4570, opened 2021-08-25
keyboard.getLayoutMap() throws an exception in iFrame on Chrome

rawnsley:
If you host Hubs in an iframe you get a console error from keyboard.getLayoutMap():

Screenshot 2021-08-25 at 10 50 19

I can't find any definitive reference to this behaviour, but [here is an example fix](https://github.com/eclipse-theia/theia/pull/5054/files) in another code base.

Downgrading this to a warning with an explicit message at least stops us having to think about it every time it happens. The fact that keyboard remapping will not work inside an iframe on Chrome is going to be harder to actually fix.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-773)


netpro2k (approved):
lgtm, thanks for flagging this.

PR #4585, opened 2021-08-30
landing page updates

matthewbcool:
adds a special check for HMC with a centered version of the logo for loading screen and entry modal. updates blues in themes to match design and for improved color contrast

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-767)


src/react-components/room/RoomEntryModal.js
```diff
@@ -29,13 +31,14 @@ export function RoomEntryModal({
   ...rest
 }) {
   const breakpoint = useCssBreakpoints();
+  const isHmc = configs.feature("show_cloud");
```
johnshaughnessy, line 34: Might be good to move this check into a function somewhere like
```
export function isHMC(){
  return configs.feature("show_cloud");
}
```
Then use it like
```
import { isHMC } from "./somewhere"

if (isHMC()) doThing();
```

Then everywhere that we are checking if we are on HMC can use this function, and if ever we decide that `"show_cloud"` is not the best `feature` to check, we only have to change it in one place (or, it is easy to check all the places that we are calling the function).

johnshaughnessy (approved):
lgtm

PR #4594, opened 2021-09-01
Bug/audio mode

johnshaughnessy:


takahirox:
I confirmed that Positional audio is disabled and `/audiomode` command has no effect on my Safari with this change. And also confirmed no distorted audio on Safari described in #4411.

PR #4597, opened 2021-09-01
Fix for URL waypoint fragments when using fast switching

rawnsley:
Fast room switching isn't currently working with URL fragments that deep-link to named waypoints.

This [example room](https://hubs.mozilla.com/CPMwEKv/stunning-dimwitted-assembly) contains two links to another Hubs room, but each link has a different fragment that corresponds to a particular named waypoint within the scene. Clicking on either link takes you to the same default position in the room rather than the intended waypoint.

This appears to be because of an assumption in the rather complicated waypoint system logic, in which fragment navigation is ignored until the first waypoint navigation has occurred, presumably to avoid this initial navigation overriding the one from the fragment.

The proposed change resets the waypoint system state when `moveToSpawnPoint` is called, which is done after the new environment model has finished loading.

Note that this is a different issue to the one preventing deep-linking working on initial load, which is covered by this PR https://github.com/mozilla/hubs/pull/4328.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-876)

### DIFF
--- a/src/systems/waypoint-system.js
+++ b/src/systems/waypoint-system.js
@@ -298,6 +298,9 @@ export class WaypointSystem {
         this.nextMoveToSpawnResolve = resolve;
       });
     }
+    // Reset system state to allow for waypoint selection based on URL fragments 
+    this.previousWaypointHash = null;
+    this.initialSpawnHappened = false;
     return this.nextMoveToSpawn;
   }
   moveToWaypoint(waypointComponent, instant) {


netpro2k:
This change seemed sane enough and testing room links locally is a bit of a pain so I merged it to test while pushing some other PRs. It does not seem to have the desired effect. @rawnsley wondering if this is working for you?

rawnsley:
@netpro2k It was working with the master branch, but not anymore. It may be related that something has changed with the way the links are displayed. This is how they look in Spoke:

Screenshot 2021-09-28 at 08 40 46

and this is how they currently look in Hubs

Screenshot 2021-09-28 at 08 40 42

When I hover over them the top one says "Open Link", but the URL is wrong: _https://smoke-hubs.mozilla.com/LLmN62m/nautical-modern-room#red_

The bottom one says "Visit Room", but I suspect it has lost the fragment.

rawnsley:
@netpro2k I've no idea what's happening in that other Spoke room, but I made a new version https://hubs.mozilla.com/scenes/CQC6oC7 and I don't get the double links or the "smoke" prefix.

That said, it still doesn't work with the fragment navigation.

netpro2k:
The smoke links were something I added when testing on our QA servers

PR #4598, opened 2021-09-01
Make camera tool render video-texture-targets

keianhzo:
This is a regression from https://github.com/mozilla/hubs/pull/4451 Adding the video-texture target layer to the camera-tool camera fixes it.


netpro2k (commented):
lgtm

PR #4602, opened 2021-09-02
fix: mobile sign out

### LINKED ISSUE #4582: [Mobile] The option to Sign Out and the account email info are no longer available on homepage
**[Affected Versions]:**
- Safari 14.6
- Chrome 92

**[Affected Platforms]:**
- Google Pixel
- Iphone SE 2

**[Prerequisites]:**
- You are logged into "https://hubs.mozilla.com/".

**[Steps to reproduce]:**
1. Navigate to https://hubs.mozilla.com/.
2. Observe if the option to Sign Out and if the logged in account email info are displayed.

**[Expected results]:**
- The option and the email account are displayed.

**[Actual results]:**
- Only the option to Sign In/Sign Up is displayed.

**[Notes]:**
- The issue is not reproducible within the More menu inside a Hubs room.
- This issue can confuse already logged in users and also new users by giving the impression that the login was not successful.
- Attached is a screen recording of the issue: [link](https://drive.google.com/file/d/14V7cv3YFbEdffPwvE_SFIBiftPFyMV2V/view?usp=sharing).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-769)

matthewbcool:
adds mobile sign-out to resolve #4582

### DIFF
--- a/src/react-components/home/HomePage.js
+++ b/src/react-components/home/HomePage.js
@@ -18,6 +18,7 @@ import { Button } from "../input/Button";
 import { Container } from "../layout/Container";
 import { SocialBar } from "../home/SocialBar";
 import { SignInButton } from "./SignInButton";
+import maskEmail from "../../utils/mask-email";
 
 export function HomePage() {
   const auth = useContext(AuthContext);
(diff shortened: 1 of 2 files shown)


src/react-components/home/HomePage.js
```diff
@@ -50,12 +51,27 @@ export function HomePage() {
...
 
   const canCreateRooms = !configs.feature("disable_room_creation") || auth.isAdmin;
-
+  const email = auth.email;
   return (
     <PageContainer className={styles.homePage}>
       <Container>
         <div className={styles.hero}>
-          <SignInButton mobile />
+          {auth.isSignedIn ? (
```
netpro2k, line 59: If we find ourselves copying this code some more we might want to turn it into a component that handles both logged in and out states.

src/react-components/home/HomePage.scss
```diff
@@ -35,8 +35,16 @@
 
 :local(.sign-in-container) {
```
netpro2k, line 36: I was confused about re-using this classname, but I don't actually see where it was being used before...

netpro2k (commented):
I would double check that `sign-in-container` isn't being used anywhere else, but otherwise LGTM

src/react-components/home/HomePage.scss
matthewbcool: I think it was some unused styling before.

PR #4603, opened 2021-09-02
Track total bytes and bitrate for model downloads.

rawnsley:
Simple tracking of download stats for model downloads using the `onProgress` callback event:

Screenshot 2021-09-02 at 13 59 22

It's useful for debugging and optimizing start times, particularly where scene assets are large or customer bandwidth is limited.

### DIFF
--- a/src/components/gltf-model-plus.js
+++ b/src/components/gltf-model-plus.js
@@ -631,7 +631,16 @@ export async function loadGLTF(src, contentType, onProgress, jsonPreprocessor) {
 }
 
 export async function loadModel(src, contentType = null, useCache = false, jsonPreprocessor = null) {
-  console.log(`Loading model ${src}`);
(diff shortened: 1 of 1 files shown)


netpro2k:
Hmm. This info (and quite a bit more) is pretty readily available in the network tab of web inspector. That also lets you distinguish between actual traffic and cached traffic. Is there a reason you prefer logging it?

rawnsley:
> Hmm. This info (and quite a bit more) is pretty readily available in the network tab of web inspector. That also lets you distinguish between actual traffic and cached traffic. Is there a reason you prefer logging it?

I agree that for local debugging the network tab is superior. I'm trying to get as much useful information into the console as I can so that is all a user needs to send me if I'm debugging a problem. Typically a user will say "it takes forever to enter a room" - this information will help narrow down the cause.

PR #4606, opened 2021-09-02
Make Hubs version number obvious to HC users

robinkwilson:
Let's the user know whether a custom client is deployed or not in both the admin panel and in a hubs room via console.log.

In Admin Panel:
If no custom client: 
image

If yes custom client:
image

QA TEST:
Ensure when a custom client is deployed, it shows this admin panel message. When it is undeployed it does not show this message in the admin panel landing page.

### DIFF
--- a/admin/src/react-components/system-editor.js
+++ b/admin/src/react-components/system-editor.js
@@ -344,10 +344,13 @@ class SystemEditorComponent extends Component {
             {configs.IS_LOCAL_OR_CUSTOM_CLIENT ? (
               <>
                 <Typography variant="body1" gutterBottom>
-                  {`App client: Custom client`}
+                  App client: Custom client
                 </Typography>
                 <Typography variant="body1" gutterBottom>
-                  {`(Undeploy custom client to run build v${process.env.BUILD_VERSION || "?"})`}
+                  {`(Undeploy custom client to run build ${process.env.BUILD_VERSION || "?"})`}
+                </Typography>
+                <Typography variant="body1" gutterBottom>
+                  {`(Remember to regularly pull in upstream changes from the "hubs-cloud" branch: https://github.com/mozilla/hubs)`}
                 </Typography>
               </>
             ) : (
--- a/admin/src/react-components/system-editor.js
+++ b/admin/src/react-components/system-editor.js
@@ -339,13 +339,22 @@ class SystemEditorComponent extends Component {
           <Title title="Hubs Cloud" />
           <CardContent className={this.props.classes.info}>
             <Typography variant="title" gutterBottom>
-              Your hub versions:
+              Your hub version:
             </Typography>
-            <Typography variant="body1" gutterBottom>
-              {`App Client: ${
-                configs.IS_LOCAL_OR_CUSTOM_CLIENT ? "Custom client or local client" : process.env.BUILD_VERSION || "?"
-              }`}
-            </Typography>
+            {configs.IS_LOCAL_OR_CUSTOM_CLIENT ? (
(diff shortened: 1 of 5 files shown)


admin/src/react-components/system-editor.js
```diff
@@ -334,6 +335,28 @@ class SystemEditorComponent extends Component {
...
+            <Typography variant="title" gutterBottom>
+              Your hub version:
+            </Typography>
+            {configs.IS_LOCAL_OR_CUSTOM_CLIENT ? (
+              <>
+                <Typography variant="body1" gutterBottom>
+                  {`App client: Custom client`}
+                </Typography>
+                <Typography variant="body1" gutterBottom>
+                  {`(Undeploy custom client to run build v${process.env.BUILD_VERSION || "?"})`}
```
brianpeiris, line 350: nit: remove the "v" for consistency
```suggestion
                  {`(Undeploy custom client to run build ${process.env.BUILD_VERSION || "?"})`}
```

admin/src/utils/configs.js
```diff
@@ -6,4 +6,6 @@ const configs = {};
   configs[x] = el ? el.getAttribute("content") : process.env[x];
 });
 
+configs.IS_LOCAL_OR_CUSTOM_CLIENT = document.querySelector("meta[name='env:reticulum_server']") ? false : true;
```
brianpeiris, line 9: Ideally we'd have some explicit way of determining whether we are using a local or custom client. For example, maybe the habitat run hook could always injects a specific meta tag like 
`<meta name="env:is_local_or_custom_client" content="false" />` 

Using the `reticulum_server` meta tag as an indirect proxy feels a bit fragile, and is especially confusing with out context.
I'd suggest at least explaining this line with a "why" comment, maybe pointing to a mana page with more explanation about how these env configs work.

brianpeiris: nit: Maybe this is more of a preference thing, but I find it redundant to have a ternary whose values are just `false` and `true`. It feels clearer just to negate the value you care about instead of using a conditional. Maybe naming the boolean is a better option:
```suggestion
const hasReticulumServerMetaEnvTag = !!document.querySelector("meta[name='env:reticulum_server']");
configs.IS_LOCAL_OR_CUSTOM_CLIENT =  !hasReticulumServerMetaEnvTag;
```

src/hub.js
```diff
@@ -9,7 +9,13 @@ import configs from "./utils/configs";
 import "./utils/theme";
 import "@babel/polyfill";
 
-console.log(`App version: ${process.env.BUILD_VERSION || "?"}`);
+console.log(
+  `App version: ${
+    configs.IS_LOCAL_OR_CUSTOM_CLIENT
+      ? `Custom client or local client (undeploy custom client to run build v${process.env.BUILD_VERSION})`
```
brianpeiris, line 15:
```suggestion
      ? `Custom client or local client (undeploy custom client to run build ${process.env.BUILD_VERSION})`
```

src/utils/configs.js
```diff
@@ -28,6 +27,8 @@ let isAdmin = false;
   }
 });
 
+configs.IS_LOCAL_OR_CUSTOM_CLIENT = document.querySelector("meta[name='env:thumbnail_server']") ? false : true;
```
brianpeiris, line 30: Same comment above. Generally not great to use an unrelated env config as a proxy. A "why" comment would be nice here too.

brianpeiris (commented):
Functionality looks good to me. Just left some comments and suggestions.

PR #4608, opened 2021-09-03
Disable artificial attenuation for stereo audio/video

### LINKED ISSUE #4604: Stereo audios shouldn't be attenuated
**Description**
Audio sources defined as stereo in Spoke/Blender are currently artificially attenuated whereas before the latest audio refactors we were only attenuating when the audio was set to stereo by using the **audioOutputMode** command. 

**To Reproduce**
1. Go to Spoke and create a new scene
2. Add an audio an set its type to stereo
3. Publish and open in Hubs
4. See that the audio is not spatialized but it's attenuated as you move away from it

**Expected behavior**
The audio shouldn't be attenuated artificially.

**Suggested solution**
The most straightforward solution right now should be to update Spoke so  when the audioType is set to `Stereo` we set the `distanceModel` to `linear` and the `rolloffFactor` to `0` so `AudioGainSystem::calculateAttenuation` returns a `1.0 `attenuation which would not apply any attenuation on the gain.

We should override this for legacy media-view components to make sure they have the right `distanceModel` and `rolloffFactor` in case they are `Stereo`.

After this update there is going to be an inconsistency between how this is handled in Spoke vs Blender. We are ok with it right now as this as would require major UI updates.

**Hardware** 
All

**Additional context**
https://discord.com/channels/498741086295031808/535606666708910101/882954477500510218

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-856)

keianhzo:
Disable artificial distance based attenuation for legacy audio/video elements.

Related https://github.com/mozilla/Spoke/pull/1169
Fixes #4604

### DIFF
--- a/src/gltf-component-mappings.js
+++ b/src/gltf-component-mappings.js
@@ -3,7 +3,7 @@ import "./components/gltf-model-plus";
 import { getSanitizedComponentMapping } from "./utils/component-mappings";
 import { TYPE, SHAPE, FIT } from "three-ammo/constants";
 const COLLISION_LAYERS = require("./constants").COLLISION_LAYERS;
-import { AudioType, SourceType } from "./components/audio-params";
+import { AudioType, DistanceModelType, SourceType } from "./components/audio-params";
 import { updateAudioSettings } from "./update-audio-settings";
 
 function registerRootSceneComponent(componentName) {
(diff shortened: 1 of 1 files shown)


netpro2k (commented):
Not fully up to date on the audio changes but this seems sane enough. I would like to explore if we actually need the artificial distance attenuation stuff at all. Without that, stereo would always be stereo with no attenuation. Possible we would still want to use PannerNode and override the attenuation for simplicity, but there is a performance overhead so it might be worth not doing, especially if we don't support changing audioType of a source at runtime (since that's where most of the complexity comes in)

johnshaughnessy (approved):

PR #4616, opened 2021-09-07
Tweak media query and avatar settings heights to avoid cutoff and scrolling

### LINKED ISSUE #4615: The bottom of the Avatar Setup screen is cut off in Oculus Browser
**Description**
Oculus Browser's default resolution is now 1000x505 and our Avatar Setup screen breaks into mobile view does not allow scrolling down. 

**To Reproduce**
1. Visit hubs.mozilla.com in the Oculus Browser
2. Clear your site storage, to ensure that the avatar setup screen appears
3. Create and enter a room
4. See that the avatar setup screen is cut off.

**Expected behavior**
The avatar setup screen should either scroll or fit into the available resolution.

**Screenshots**
image

**Hardware** 
 - Device: Quest
 - OS: Quest
 - Browser: Oculus Browser

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-881)

brianpeiris:
Fixes #4615

- Increase "breakpoint-vr" cutoff, so that we use mobile screens on Oculus Browser's new 1000x505px default resolution
- Also reduce heights of elements in the Avatar Setup panel, in order to avoid having to scroll the dialog

Before:
image

After:
image

### DIFF
--- a/src/react-components/room/AvatarSettingsContent.scss
+++ b/src/react-components/room/AvatarSettingsContent.scss
@@ -16,9 +16,12 @@
 
   /* TODO: This styling into AvatarPreview */
   & > :first-child {
+    /*
+    We need to set dimensions explicitly here, since AvatarPreview
+    resizes itself to its container's dimensions.
+    */
     width: 168px;
     height: 250px;
-    min-height: 250px;
     border-radius: 8px;
     background-color: theme.$tile-bg-color;
   }
--- a/src/react-components/room/AvatarSettingsContent.scss
+++ b/src/react-components/room/AvatarSettingsContent.scss
@@ -2,7 +2,7 @@
 
 :local(.content) {
   align-items: center;
-  padding: 24px;
+  padding: 10px;
   text-align: center;
 }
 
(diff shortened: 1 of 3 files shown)


src/react-components/room/AvatarSettingsContent.scss
```diff
@@ -16,8 +16,8 @@
   /* TODO: This styling into AvatarPreview */
   & > :first-child {
     width: 168px;
-    height: 300px;
-    min-height: 300px;
+    height: 250px;
```
matthewbcool, line 20: I don't think height is required here since min-height should be overwriting it AFAIK.

matthewbcool (approved):
lgtm. Tested on oculus browser, chrome and ff.

src/react-components/room/AvatarSettingsContent.scss
brianpeiris: Ah, good call. I'll double-check if I can remove that safely.

brianpeiris: Turns out that `height` was the required property here. It's a bit of quirk, since the `AvatarPreview` component that is contained by this component resizes itself to its container's dimensions. Removed min-height and added a comment to explain that.

PR #4620, opened 2021-09-09
Fix audio pause state

keianhzo:
There are several places we we are not handling the audio pause state updates correctly.

### DIFF
--- a/src/components/player-info.js
+++ b/src/components/player-info.js
@@ -68,7 +68,8 @@ AFRAME.registerComponent("player-info", {
     registerComponentInstance(this, "player-info");
   },
   remove() {
-    APP.isAudioPaused.delete(this.el);
+    const avatarEl = this.el.querySelector("[avatar-audio-source]");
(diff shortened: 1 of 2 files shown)


src/components/player-info.js
```diff
@@ -186,10 +186,11 @@ AFRAME.registerComponent("player-info", {
       }
     }
 
+    const avatarEl = this.el.querySelector("[avatar-audio-source]");
     if (this.data.muted) {
-      APP.isAudioPaused.add(this.el);
+      APP.isAudioPaused.add(avatarEl);
     } else {
-      APP.isAudioPaused.delete(this.el);
```
johnshaughnessy, line 192: Does this need to be `avatarEl` in `remove` too?

keianhzo: Isn't that the next line?

keianhzo:
Sorry I misunderstood the comment. Fixed.

Issue #4502, opened 2021-08-10
Hide volume controls for GIF


takahirox:
**Is your feature request related to a problem? Please describe.**

Our GIF plate has volume controls but GIF doesn't support audio. It would be confusing to users. We should hide volume controls for GIF.

![unknown](https://user-images.githubusercontent.com/7637832/128927806-23da2844-e8ef-4f9f-94e5-485a5706d3f2.png)

**Describe the solution you'd like**

Hide volume controls for GIF

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-743)

keianhzo:
@takahirox I've tried adding through the place button, Spoke and just copy pasting the link I can't reproduce this for Gifs. How where you adding the gif? Can you share a gif link?

takahirox:
Don't you see the GIF button in the place menu?

![image](https://user-images.githubusercontent.com/7637832/132035645-8b75c112-546c-44b6-86d9-c7d804c267b0.png)

keianhzo:
That was the only method that I didn't try 😅

I've been debugging further and we are showing the controls in that case because the Reticulum media query API is returning mp4 instead of gifs for the gif backend. Looking into Reticulum I see that even though the gif backend is returning a few formats we specifically reduce to :
https://github.com/mozilla/reticulum/blob/b5680121d3ade221d6d3f862ff1fc9a70bdfcd99/lib/ret/media_search.ex#L780-L798

I don't see any issue to switch to gif on that one but my understanding is that we did that for a reason but I'm not sure: https://github.com/mozilla/reticulum/pull/119

Maybe @brianpeiris or @johnshaughnessy have some more background on that one.

keianhzo:
Per our chat on the engineering meeting we want to keep using mp4 as the format from the gif backend and just (in general for videos) just remove the audio controls if there is no audio tracks.

Softvision-GeluHaiduc:
I have verified this issue using the latest releases of Firefox and Chrome on Windows 10.
Now, the volume controls for GIFs are no longer displayed.

PR #4640, opened 2021-09-13
Show audio controls for media only if there are audio tracks

keianhzo:
Fixes https://github.com/mozilla/hubs/issues/4502

This PR adds support for detecting if an audio element has audio tracks and depending on that:
- Create or not a audio THREE audio element
- Show/Hide audio controls

### DIFF
--- a/src/utils/media-utils.js
+++ b/src/utils/media-utils.js
@@ -544,3 +544,17 @@ export function closeExistingMediaMirror() {
     });
   }
 }
+
+export function hasAudioTracks(el) {
+  if (!el) return false;
+  // At some point browsers will implement the audioTracks property
+  if (el.audioTracks !== undefined) {
+    return el.audioTracks.length > 0;
+  } else if (el.webkitAudioDecodedByteCount !== undefined) {
+    return el.webkitAudioDecodedByteCount > 0;
+  } else if (el.mozHasAudio !== undefined) {
+    return el.mozHasAudio;
+  } else {
+    return false;
+  }
+}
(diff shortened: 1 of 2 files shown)


src/utils/media-utils.js
```diff
@@ -544,3 +544,17 @@ export function closeExistingMediaMirror() {
...
   }
 }
+
+export function hasAudioTracks(el) {
+  if (!el) return false;
+  // At some point browsers will implement the audioTracks property
+  if (el.audioTracks !== undefined) {
+    return el.audioTracks.length > 0;
+  } else if (el.webkitAudioDecodedByteCount !== undefined) {
+    return el.webkitAudioDecodedByteCount > 0;
```
netpro2k, line 554: Might this be 0 when the video is first loading? We wait in `createVideoTextureAudioSourceEl` for the video to have a width/height so I guess at that point we could probably also assume some audio should have been decoded.

src/utils/media-utils.js
```diff
@@ -544,3 +544,17 @@ export function closeExistingMediaMirror() {
     });
   }
 }
+
+export function hasAudioTracks(el) {
+  if (!el) return false;
+  // At some point browsers will implement the audioTracks property
+  if (el.audioTracks !== undefined) {
```
netpro2k, line 551: Minor nit, but I would have probably written this if/else chain as just a boolean expression

netpro2k (approved):
lgtm!

src/utils/media-utils.js
keianhzo: I'd say so, for what I see about media element loaddata event https://html.spec.whatwg.org/multipage/media.html#event-media-loadedmetadata that's when the dimensions are known for the whole media resource (complete set of media data). We are checking for the actual dimensions instead of the loaddata event (I don't why is that though) so I think it's safe to assume that all tracks are available by that time and we can rely on that instead of adding overhead to the tick.

Issue #4059, opened 2021-03-19
A pinned PDF file is enlarged if the next/previous page button is clicked repeatedly


Softvision-GeluHaiduc:
**[Affected Versions]:**
- Firefox Release 86.0.1
- Chrome 89.0.4389.82

**[Affected Platforms]:**
- Windows 10x64

**[Prerequisites]:**
- Make sure to have a room created and entered.

**[Steps to reproduce]:**
1. Click the "Place" button.
2. Click the "Upload" button from the menu and attach a PDF file.
3. Click the "Create Object" button.
4. Pin the PDF file in the room.
5. Click the next/previous page button and observe the behavior.
    
**[Expected results]:**
- The next/previous page of the PDF file is displayed.

**[Actual results]:**
-  The next/previous page of the PDF file is displayed but the PDF file is also enlarged.

**[Notes]:**
- The issue is not reproducible if the PDF file is not pinned.
- Attached a screen recording of the issue below
![pdf file](https://user-images.githubusercontent.com/51400741/111789757-2cb89e00-88ca-11eb-8272-dc5465f61dc0.gif)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-585)

mauriciomartinez-itesm:
This issue has already been reported in #2284

matthewbcool:
Seems to be fixed, can we confirm?

mauriciomartinez-itesm:
> Seems to be fixed, can we confirm? Never mind! was able to reproduce after moving the pdf, resizing it and re-pinning.

This issue has not been fixed in the production build yet, I just tested it... remember to pin the PDF before clicking right and left rapidly

Softvision-GeluHaiduc:
I have verified this issue using the latest releases of Firefox and Chrome on Windows 10.
Now, a pinned PDF file is no longer enlarged if the next/previous page button is clicked repeatedly.
While verifying this issue I encountered another issue that is logged here #4696.

PR #4642, opened 2021-09-14
Skip pinning animations if already running

keianhzo:
Fixes https://github.com/mozilla/hubs/issues/4059

When clicking on the pager next/prev while the previous animations os still running we use the updated object scale as the size reference and we end up with a different scaled object when the animations end.

This PR fixes just dropping animations if the previous animations are still running.

### DIFF
--- a/src/components/pinnable.js
+++ b/src/components/pinnable.js
@@ -86,7 +86,7 @@ AFRAME.registerComponent("pinnable", {
     let didFireThisFrame = false;
     if (!held && this.wasHeld && isMine) {
       didFireThisFrame = true;
-      this._fireEventsAndAnimate({ data: this.data, force: true });
+      this._fireEventsAndAnimate({ oldData: this.data, force: true });
     }
 
     this.wasHeld = held;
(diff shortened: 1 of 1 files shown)


keianhzo:
@netpro2k That makes sense, I didn't really know what was the expected behavior. I've updated to not animate but I kept the code to check if the animation are already running to avoid this scaling issue in case we can get multiple events.

src/components/pinnable.js
```diff
@@ -80,15 +86,15 @@ AFRAME.registerComponent("pinnable", {
     let didFireThisFrame = false;
     if (!held && this.wasHeld && isMine) {
       didFireThisFrame = true;
-      this._fireEventsAndAnimate(this.data, true);
+      this._fireEventsAndAnimate({ data: this.data, force: true });
```
johnshaughnessy, line 89: The function signature destructures `oldData` but this line creates an object with key `data`. Is this a bug?

keianhzo: Oh yep, good catch. Updated.

PR #4655, opened 2021-09-20
Fix 3D models Media Frame scaling issues and add support for 3D preview

### LINKED ISSUE #4089: Media zones will sometimes make the object inserted enormous

keianhzo:
Fixes #4089

This PR fixes a couple of issues with media frames and 3D object:
- Fixes the scale issues when the object is captured
- Adds full support for 3D object preview

### DIFF
--- a/src/systems/media-frames.js
+++ b/src/systems/media-frames.js
@@ -1,12 +1,14 @@
 import { MediaType } from "../utils/media-utils";
-import { TEXTURES_FLIP_Y } from "../loaders/HubsTextureLoader";
 import { applyPersistentSync } from "../utils/permissions-utils";
-import { createPlaneBufferGeometry } from "../utils/three-utils";
+import { cloneObject3D, disposeNode } from "../utils/three-utils";
+import qsTruthy from "../utils/qs_truthy";
 
 // TODO better handling for 3d objects
 function scaleForAspectFit(containerSize, itemSize) {
-  return Math.min(containerSize.x / itemSize.x, containerSize.y / itemSize.y);
+  return Math.min(containerSize.x / itemSize.x, containerSize.y / itemSize.y, containerSize.z / itemSize.z);
 }
+
+const DEBUG = qsTruthy("debug");
 
 const isCapturableByType = {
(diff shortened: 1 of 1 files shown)


src/systems/media-frames.js
```diff
@@ -262,22 +332,38 @@ AFRAME.registerComponent("media-frame", {
...
-      capturableEntity.components["floaty-object"].setLocked(true);
+      const update = () => {
+        this.el.setAttribute("media-frame", {
+          targetId: capturableEntity.id,
+          originalTargetScale: new THREE.Vector3().copy(capturableEntity.object3D.scale)
+        });
+
+        capturableEntity.object3D.scale.set(1, 1, 1);
+        capturableEntity.object3D.quaternion.set(0, 0, 0, 1);
+        capturableEntity.object3D.updateMatrixWorld();
```
netpro2k, line 343: Don't think this is needed but you should set `matrixNeedsUpdate` since scale/quaternion was just changed.

netpro2k (commented):
lgtm. Thanks for getting this fixed up! If I had implemented this I would have been too lazy to support animated previews, but that's a nice touch :) - Made a few comments, mostly around the matrix world update contract. I think most of my comments are noop changes so no strong opinions on if you bother changing them if you have already done a bunch of testing.

PR #4660, opened 2021-09-22
Fixes for environment-settings system

### LINKED ISSUE #4634: Scene preview doesn't include HDR environment map
The scene preview doesn't include the environment skybox as a background or as a contributor to light in the render. For example, [this scene](https://hubs.mozilla.com/scenes/327U8BL/hubhenge-day) ([source](https://github.com/LearnHub/HubHenge)):

Screenshot 2021-09-13 at 08 36 03 Screenshot 2021-09-13 at 08 36 19

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-893)

### LINKED ISSUE #4635: Environment settings doesn't support non-HDR texture for background
If you set the **Environment Settings** _Background Image_ property to be a non-HDR texture it is not displayed in Hubs. In [this example scene](https://hubs.mozilla.com/scenes/KaUPL2M/ldr-env-texture-test) the image was set to another scene texture:

Screenshot 2021-09-13 at 08 50 13

When the scene is loaded only the background color is shown:

Screenshot 2021-09-13 at 08 51 21

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-894)

netpro2k:
Fixes a few minor issues brought up with the environment settings system

fixes #4634
fixes #4635


src/components/gltf-model-plus.js
```diff
@@ -58,7 +58,8 @@ const extractZipFile = promisifyWorker(new SketchfabZipWorker());
 
 function defaultInflator(el, componentName, componentData) {
   if (!AFRAME.components[componentName]) {
-    throw new Error(`Inflator failed. "${componentName}" component does not exist.`);
+    console.warn(`Inflator failed. "${componentName}" component does not exist.`);
```
netpro2k, line 61: We intentionally don't load all components on the scene page, but this causes some models to fail to load. This doesn't seem like it should be fatal, so converted it to a warning.

PR #4665, opened 2021-09-23
Fix browser detection code in node.js

brianpeiris:
Note this PR will be merged into the hubs-cloud branch. There will be a separate PR for the master branch.

In the last Hubs Cloud release, we added some browser [detection code to store.js](https://github.com/mozilla/hubs/pull/4442/files#diff-e0d49eaa67ecc73f10739b451f265c21d0bcf90a58101d5e77e5035dd8c233cc) to workaround issues with Safari. This code is also in the code path for `npm run login`, but the detection code did not behave correctly in a node.js environment. This PR fixes that issue.

### DIFF
--- a/src/utils/detect-safari.js
+++ b/src/utils/detect-safari.js
@@ -3,5 +3,5 @@ import { detect } from "detect-browser";
 export function isSafari() {
   const browser = detect();
   if (!browser) return false;
-  return ["iOS", "Mac OS"].includes(browser.os) && ["safari", "ios"].includes(browser.name)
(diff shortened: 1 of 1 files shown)


brianpeiris:
Build is failing because of prettier errors in other parts of the code. I've run all the other test/build/lint steps locally to make sure they are okay.

brianpeiris:
Correction: We do not need to fix this issue in the master branch, since the safari workaround was implemented in a different way in #4594

PR #4666, opened 2021-09-23
change avatar preview background to background theme color

matthewbcool:
| Avatar-Preview Themed Before     | Avatar Preview Themed After  |
| ----------- | ----------- |
|avatar pre before|avatar pre after|

Adds a function to avatar-preview.js that checks for a theme background color (must be hex value) and updates the preview background to that color. Default preview background will remain the same for now.

Putting the Avatar in a generic 'blueprint room' scene may be a better option than using a color there as we can still have contrast issues since the color of the avatar is variable.

For now this PR is meant to bring us one step closer to 'acceptable dark mode'.

Random theme:  
anytheme

Default:
default

### DIFF
--- a/src/react-components/avatar-preview.js
+++ b/src/react-components/avatar-preview.js
@@ -72,7 +72,7 @@ function getThemeBackground() {
   }
   for (let i = 0; i < themes.length; i++) {
     if (themes[i].id === currentTheme) {
-      let bgHex = themes[i].variables["background1-color"];
+      let bgHex = themes[i].variables["background3-color"];
       bgHex = `0x${bgHex.substring(1)}`;
       if (bgHex.length !== 8) {
         return defaultColor;
--- a/src/react-components/avatar-preview.js
+++ b/src/react-components/avatar-preview.js
@@ -64,23 +64,11 @@ function fitBoxInFrustum(camera, box, center, margin = DEFAULT_MARGIN) {
 }
 
 function getThemeBackground() {
-  const currentTheme = APP.store.state.preferences.theme;
-  const themes = window.APP_CONFIG?.theme.themes;
-  const defaultColor = 0xeaeaea;
-  if (currentTheme === "hubs-default") {
-    return defaultColor;
-  }
-  for (let i = 0; i < themes.length; i++) {
-    if (themes[i].id === currentTheme) {
-      let bgHex = themes[i].variables["background3-color"];
-      bgHex = `0x${bgHex.substring(1)}`;
-      if (bgHex.length !== 8) {
-        return defaultColor;
-      }
-      bgHex = parseInt(bgHex, 16);
-      return bgHex;
-    }
-  }
+  const currentTheme = APP?.store?.state?.preferences?.theme;
+  const themes = window.APP_CONFIG?.theme?.themes;
+  const currentThemeObject = themes?.find(t => t.id === currentTheme);
+  const previewBackgroundColor = new THREE.Color(currentThemeObject?.variables["background3-color"] || 0xeaeaea);
+  return previewBackgroundColor;
 }
 
 class AvatarPreview extends Component {
--- a/src/react-components/avatar-preview.js
+++ b/src/react-components/avatar-preview.js
@@ -64,9 +64,9 @@ function fitBoxInFrustum(camera, box, center, margin = DEFAULT_MARGIN) {
 }
 
 function getThemeBackground() {
-  let currentTheme = APP.store.state.preferences.theme;
-  let themes = APP_CONFIG.theme.themes;
-  let defaultColor = 0xeaeaea;
+  const currentTheme = APP.store.state.preferences.theme;
+  const themes = window.APP_CONFIG?.theme.themes;
+  const defaultColor = 0xeaeaea;
   if (currentTheme === "hubs-default") {
     return defaultColor;
   }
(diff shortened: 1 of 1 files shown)


brianpeiris:
I don't mean to block a small change like this, but can we get some design input on it? IMO, I think the background served a specific purpose -- it indicated that the avatar preview area was separate from the panel, and indicated an interaction affordance, so that users had a hint that the avatar preview was live and that you could rotate the avatar.

For dark mode, I would rather us use a slightly darker shade to differentiate it from the background.

matthewbcool:
> I don't mean to block a small change like this, but can we get some design input on it? IMO, I think the background served a specific purpose -- it indicated that the avatar preview area was separate from the panel, and indicated an interaction affordance, so that users had a hint that the avatar preview was live and that you could rotate the avatar.
> 
> For dark mode, I would rather us use a slightly darker shade or slightly lighter shade, to differentiate it from the background.

It's a good point and valid concern, how about using background3 instead? Matt and I also came to this conclusion. 
image

brianpeiris:
That sounds better, though I wish we gave our theme colors semantic names instead of 1, 2, 3, but that's a separate issue.
I have some code review comments that I'll add shortly.

src/react-components/avatar-preview.js
```diff
@@ -63,6 +63,26 @@ function fitBoxInFrustum(camera, box, center, margin = DEFAULT_MARGIN) {
   camera.lookAt(center);
 }
 
+function getThemeBackground() {
+  const currentTheme = APP.store.state.preferences.theme;
+  const themes = window.APP_CONFIG?.theme.themes;
```
brianpeiris, line 68: The `?.` here implies that APP_CONFIG could be undefined, which means the `themes` variable could be undefined here too.

src/react-components/avatar-preview.js
```diff
@@ -63,6 +63,26 @@ function fitBoxInFrustum(camera, box, center, margin = DEFAULT_MARGIN) {
   camera.lookAt(center);
 }
 
+function getThemeBackground() {
+  const currentTheme = APP.store.state.preferences.theme;
```
brianpeiris, line 67: Are `preferences` and `preferences.theme` always going to be defined?

src/react-components/avatar-preview.js
```diff
@@ -63,6 +63,26 @@ function fitBoxInFrustum(camera, box, center, margin = DEFAULT_MARGIN) {
...
+  const themes = window.APP_CONFIG?.theme.themes;
+  const defaultColor = 0xeaeaea;
+  if (currentTheme === "hubs-default") {
```
brianpeiris, line 70: If `currentTheme` is "hubs-default", shouldn't it just take the value from the associated theme structure?

src/react-components/avatar-preview.js
```diff
@@ -63,6 +63,26 @@ function fitBoxInFrustum(camera, box, center, margin = DEFAULT_MARGIN) {
...
+    return defaultColor;
+  }
```
brianpeiris, line 72: I'd suggest guarding against these values possibly being undefined or empty instead: 

```suggestion
  if (!currentTheme || !themes || themes.length === 0) {
    return defaultColor;
  }
```

src/react-components/avatar-preview.js
```diff
@@ -63,6 +63,26 @@ function fitBoxInFrustum(camera, box, center, margin = DEFAULT_MARGIN) {
...
+    if (themes[i].id === currentTheme) {
+      let bgHex = themes[i].variables["background3-color"];
+      bgHex = `0x${bgHex.substring(1)}`;
+      if (bgHex.length !== 8) {
+        return defaultColor;
+      }
+      bgHex = parseInt(bgHex, 16);
+      return bgHex;
+    }
+  }
```
brianpeiris, line 83: In case the `currentTheme` was not found in the themes array, return the defaultColor here.

This is currently a bug if you look at the behavior of the code in light mode when using `npm run dev`, since the default theme preference is undefined. The theme preference could also be "Browser Default", which is arguably a bug in the preferences panel.

```suggestion
  }
  
  return defaultColor;
```

src/react-components/avatar-preview.js
```diff
@@ -63,6 +63,26 @@ function fitBoxInFrustum(camera, box, center, margin = DEFAULT_MARGIN) {
...
+  const themes = window.APP_CONFIG?.theme.themes;
+  const defaultColor = 0xeaeaea;
+  if (currentTheme === "hubs-default") {
+    return defaultColor;
+  }
+  for (let i = 0; i < themes.length; i++) {
+    if (themes[i].id === currentTheme) {
+      let bgHex = themes[i].variables["background3-color"];
+      bgHex = `0x${bgHex.substring(1)}`;
+      if (bgHex.length !== 8) {
```
brianpeiris, line 77: Not sure if we intend people to use any type of CSS color value when defining their themes, but if we wanted to support other variations of CSS colors, we could use THREE.Color to do the conversion for us, instead of assuming hex values here.

brianpeiris (changes requested):
Left some comments about handling edge cases where values are undefined, but looks good to me otherwise.

src/react-components/avatar-preview.js
matthewbcool: made a new commit taking all the above feedback into account. Thank you for all the great suggestions and questions.

brianpeiris (approved):
LGTM. This code is much cleaner. Nice work!

PR #4676, opened 2021-09-27
Fix deprecated Three.js Quaternion API in three-utils.js

takahirox:
This PR fixes deprecated Three.js Quaternion API in three-utils.js, which was I have missed when I upgraded Three.js.


netpro2k (approved):
This is currently only used when traveling to a waypoint, so isn't the primary source of log spam, but definitely should be fixed. This LGTM.

PR #4677, opened 2021-09-27
Fix default dev proxy

brianpeiris:
We've change the configuration on dev.reticulum.io to use hubs-proxy.com instead of cors-proxy-dev.reticulum.io. This PR changes the default environment variable to reflect that, so that proxying works correctly when using `npm run dev`.


takahirox (approved):
Looking good. Actually I have been locally editing it.

brianpeiris:
Thanks for the quick review

takahirox:
> Thanks for the quick review

NP! The reason why my review was so quick was I was about to make the same PR.

Issue #4573, opened 2021-08-25
Allow independent audio parameter overrides


keianhzo:
Right now we always override all the audio parameters at the  supported override levels (`defaults, scene, audio, debug panel and zones`) mainly because for any component coming from Spoke/Blender we don't have any way of overriding only specific properties, the UI only allows to override all the audio parameters as a block. We only override specific audio parameters for the debugging panel properties.

https://github.com/mozilla/hubs/blob/a35c28ba1965d1d0ed1324e9f2cbf26953042347/src/update-audio-settings.js#L24-L39

This has some drawbacks like for example when an audio zone needs to be applied, we need to override all the audio parameters instead of selectively overriding those we are interested. This is very inconvenient as for things like just overriding the gain, you are going to end up overriding unrelated parameters in a lot of different audio sources that might have different values.

On the Spoke/Blender UI side we should allow the user to be able to see the defaults for the different audio parameters and independently override them both at the component level and at the scene level. In the Hubs Client we need to handle this accordingly and update the zones behavior so we correctly override individual parameters.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-757)

PR #4681, opened 2021-09-28
Allow independent audio parameter overrides

keianhzo:
Closes https://github.com/mozilla/hubs/issues/4573
Related https://github.com/mozilla/Spoke/pull/1175

This PR along with it's Spoke counterpart add support for overriding individual audio parameters.

### DIFF
--- a/src/systems/audio-zones-system.js
+++ b/src/systems/audio-zones-system.js
@@ -1,13 +1,28 @@
 // We apply the most restrictive audio parameters
 function paramsReducer(acc, curr) {
+  if (!curr) return;
   if (acc === null) acc = curr;
-  acc.gain = Math.min(acc.gain, curr.gain);
-  acc.maxDistance = Math.min(acc.maxDistance, curr.maxDistance);
-  acc.refDistance = Math.min(acc.refDistance, curr.refDistance);
-  acc.rolloffFactor = Math.max(acc.rolloffFactor, curr.rolloffFactor);
-  acc.coneInnerAngle = Math.min(acc.coneInnerAngle, curr.coneInnerAngle);
-  acc.coneOuterAngle = Math.min(acc.coneOuterAngle, curr.coneOuterAngle);
-  acc.coneOuterGain = Math.min(acc.coneOuterGain, curr.coneOuterGain);
+  if (curr.gain !== undefined) {
+    acc.gain = Math.min(acc.gain, curr.gain);
+  }
+  if (curr.maxDistance !== undefined) {
+    acc.maxDistance = Math.min(acc.maxDistance, curr.maxDistance);
+  }
+  if (curr.refDistance !== undefined) {
+    acc.refDistance = Math.min(acc.refDistance, curr.refDistance);
+  }
+  if (curr.rolloffFactor !== undefined) {
+    acc.rolloffFactor = Math.max(acc.rolloffFactor, curr.rolloffFactor);
+  }
+  if (curr.coneInnerAngle !== undefined) {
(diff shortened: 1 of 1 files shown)


src/systems/audio-zones-system.js
```diff
@@ -1,13 +1,28 @@
...
   if (acc === null) acc = curr;
-  acc.gain = Math.min(acc.gain, curr.gain);
-  acc.maxDistance = Math.min(acc.maxDistance, curr.maxDistance);
-  acc.refDistance = Math.min(acc.refDistance, curr.refDistance);
-  acc.rolloffFactor = Math.max(acc.rolloffFactor, curr.rolloffFactor);
-  acc.coneInnerAngle = Math.min(acc.coneInnerAngle, curr.coneInnerAngle);
-  acc.coneOuterAngle = Math.min(acc.coneOuterAngle, curr.coneOuterAngle);
-  acc.coneOuterGain = Math.min(acc.coneOuterGain, curr.coneOuterGain);
+  if (curr.gain !== undefined) {
+    acc.gain = Math.min(acc.gain, curr.gain);
```
johnshaughnessy, line 6: Calling `paramsReducer({}, {gain: 1})` yields `{ gain: NaN }`

src/systems/audio-zones-system.js
```diff
@@ -1,13 +1,28 @@
...
+  if (curr.coneInnerAngle !== undefined) {
+    acc.coneInnerAngle = Math.min(acc.coneInnerAngle, curr.coneInnerAngle);
+  }
+  if (curr.coneOuterAngle !== undefined) {
+    acc.coneOuterAngle = Math.min(acc.coneOuterAngle, curr.coneOuterAngle);
+  }
+  if (curr.coneOuterGain !== undefined) {
+    acc.coneOuterGain = Math.min(acc.coneOuterGain, curr.coneOuterGain);
+  }
   return acc;
```
johnshaughnessy, line 26: Maybe this function would be simpler as a `reduce` over the possible parameter names:
```js
return ["gain", "maxDistance",...].reduce((result, [key]) => {
        if ( predicate ) {
            result[key] = value;
        }
        return result;
    }, {});
```

src/systems/audio-zones-system.js
```diff
@@ -81,8 +96,20 @@ const updateSource = (function() {
       source.apply(inOutParams);
     } else {
       const params = outInParams;
-      params.gain = Math.min(outInParams.gain, inOutParams.gain);
-      params.coneOuterAngle = Math.min(outInParams.coneOuterAngle, inOutParams.coneOuterAngle);
+      if (outInParams.gain !== undefined && inOutParams.gain !== undefined) {
```
johnshaughnessy, line 99: Why are `gain` and `coneOuterAngle` the only properties that can change here? (I know this is not new behavior, but I don't know why it's this way.)

johnshaughnessy (changes requested):

src/systems/audio-zones-system.js
keianhzo: This branch is covering the case where an audio is affected by an `inOut` Zone and also by an `outIn` Zone, so this case:

audio-zones

In this case the only clear thing to me at the time was that we want to use the `outIn` zone parameters as is the closest to the listener but still use the min gain between both. Not sure why also the conOuterAngle tbh, it doesn't make much sense now. I think we can probably remove that one.

src/systems/audio-zones-system.js
keianhzo: yep, that's much better, thanks. Updated.

PR #4689, opened 2021-09-29
Add pref and logic to invert y cursor when looking

### LINKED ISSUE #3877: Option to invert mouse vertical look
When using mouse to look around my vertical control is inverted 

Having the option to set the mouse control for vertical movements inverted or not. Like in every games

**Describe alternatives you've considered**
---

**Additional context**
---

matthewbcool:
Resolves #3877

We have had several requests for an invert-y option across various forums.

It might be a good idea to apply this to the camera joysticks on various controllers as well but this PR just covers the mouse.

### DIFF
--- a/src/react-components/preferences-screen.js
+++ b/src/react-components/preferences-screen.js
@@ -442,6 +442,10 @@ const preferenceLabels = defineMessages({
     id: "preferences-screen.preference.invert-touchscreen-camera-move",
     defaultMessage: "Invert direction of camera movement for touchscreens"
   },
+  invertCursorY: {
+    id: "preferences-screen.preference.invert-cursor-y",
+    defaultMessage: "Invert cursor Y-axis"
+  },
   locale: {
     id: "preferences-screen.preference.locale",
     defaultMessage: "Language"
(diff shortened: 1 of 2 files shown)


brianpeiris (changes requested):
This implementation feels broken to me. 
- The cursor jumps up and down when you click without dragging. 
- In some cases, clicking causes the camera to move. 
- The cursor jumps to a different position when you've finished dragging.
- This inverts object movement as well. It's unclear to me if users who prefer inverted camera movement would also want inverted object movement.

I think fundamentally this is a bit harder to do in Hubs since our cursor is not locked to the center of the screen, unlike reticles in games. So users who think they want inverted camera movement might not understand what that means with an independent cursor.

I can't say off the top of my head what the correct code would look like, I haven't touched this part of the code in a while. My guess is we'd want the cursor to be consistent and not jump around. Single clicks should never move the camera. And I'm guessing users do not want inverted object movement.

matthewbcool:
Let's close this one for now since I don't see a clear path to fixing the above issues with this implementation.   
Maybe this option overall makes more sense in hubs with controllers only given all the tricky mouse problems.

PR #4706, opened 2021-10-04
OpenEXR support

rawnsley:
I've added a `MOZ_texture_exr` component with a simple implementation to support OpenEXR.

The caveat is that the imported textures end up being interpreted as upside down. For this reason I've branched from some of the corrective code for SDR images as it appears to be related and it's not easy to see where the various flipping should be applied.

Here's [an example scene file](https://github.com/mozilla/hubs/files/7277953/hubhenge.glb.zip) file for testing that was exported using [this pending PR](https://github.com/MozillaReality/hubs-blender-exporter/pull/44) for the Blender exporter.


src/components/gltf-model-plus.js
```diff
@@ -591,6 +592,36 @@ class GLTFMozTextureRGBE {
...
+    const textureDef = json.textures[textureIndex];
+
+    if (!textureDef.extensions || !textureDef.extensions[this.name]) {
+      return null;
+    }
+
+    const extensionDef = textureDef.extensions[this.name];
+    const source = json.images[extensionDef.source];
+    return parser.loadTextureImage(textureIndex, source, this.loader).then(t => {
+      // TODO pretty severe artifacting when using mipmaps, disable for now (assumption copied from GLTFMozTextureRGBE)
```
netpro2k, line 614: I suspect this was to do with decoding the RGBM which shouldn't be a factor here, the mips should blend more smoothly here since EXR actually uses float values from the start. Didn't test it obviously.

netpro2k:
Is there any particular reason you want to use OpenEXR over Radiance HDR? My plan was to only support 1 non-compressed HDR format for now and wait for some compressed standard to solidify.

rawnsley:
> Is there any particular reason you want to use OpenEXR over Radiance HDR? My plan was to only support 1 non-compressed HDR format for now and wait for some compressed standard to solidify.

Good question with a somewhat complicated answer: I'm exporting from Blender in EXR to preserve the maximum precision ahead of some post-processing steps that might otherwise suffer from the noise introduced by RGBE. Actually the output from that process doesn't include EXR files (because it's so verbose) so support in Hubs client isn't vital, but it would be useful for making a visual comparison of the scene before and after post-processing.

I wasn't going to bother with the PR, but I realised it might have some small implications for the flipping logic.

src/components/gltf-model-plus.js
rawnsley: Makes sense. I'll remove and check.

netpro2k:
Did not intend to close this with the merge. But I also don't think I will land it unless there ends up being a particularly strong demand for it or something that cant be represented with a radiance HDR map, since it will just add further support burden to each part of the pipeline.

PR #4709, opened 2021-10-05
Notify avatar changes

### LINKED ISSUE #4623: Notify avatar change when clicking "use avatar" button in the avatar object hover menu

If I click "use avatar" button in an avatar object hover menu, my avatar is silently switched.

I think we should notify the avatar change for example by displaying "avatar is changed" message or sound something.

unknown

Notify the avatar change for example by displaying "avatar is changed" message or sound something.

keianhzo:
Fixes #4623

This PR logs a message when the user's avatar has changed. Note that this is not broadcasted to the rest of the people in the room, it's just a local message.

### DIFF
--- a/src/react-components/room/ChatSidebar.js
+++ b/src/react-components/room/ChatSidebar.js
@@ -144,7 +144,8 @@ export const LogMessageType = {
   invalidAudioNormalizationRange: "invalidAudioNormalizationRange",
   audioSuspended: "audioSuspended",
   audioResumed: "audioResumed",
-  joinFailed: "joinFailed"
+  joinFailed: "joinFailed",
+  avatarChanged: "avatarChanged"
 };
 
 const logMessages = defineMessages({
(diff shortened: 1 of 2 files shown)


keianhzo:
@takahirox Apparently that's handled here: 
https://github.com/mozilla/hubs/blob/b07c5ce3e3aac027dfb2f36e68c6b517f46652a8/src/components/open-media-button.js#L23-L38
and here:
https://github.com/mozilla/hubs/blob/b07c5ce3e3aac027dfb2f36e68c6b517f46652a8/src/utils/media-url-utils.js#L175
For some reason for local server the "hub-name" header is not available. Not sure why is that.

I have also notices that we support changing the avatar from an avatar link as you posted but that is not working, I think that might be a bug.

BTW If have updated the PR with another method of detecting the change that feels that makes more sense.

takahirox:
@takahirox Thanks for the investigation.

> For some reason for local server the "hub-name" header is not available. Not sure why is that.

Any historical reasons? @netpro2k @brianpeiris 

> I have also notices that we support changing the avatar from an avatar link as you posted but that is not working, I think that might be a bug.

Oh. Yeah it sounds like a bug and we should open another issue/PR for it.

takahirox:
@keianhzo So, how did you test?

brianpeiris:
> Any historical reasons?

I think I'm seeing the hub-name header on my local server -- it should appear when making a request to the root URL (https://hubs.local:4000). Its value may be blank if you haven't configured an "app name", but the header key should still exist, so the `headers.has` check should still pass.

PR #4738, opened 2021-10-14
rework and fix media-url-utils

yakyouk:
rework and fixes for `media-url-utils.js`
- `isHubsServer` fetches HEAD only once per tested origin even if the same origin is tested multiple times in quick successions
- fix `hubsSceneRegex`, `hubsAvatarRegex` and `hubsRoomRegex` to allow urls ending with a fragment
- more efficient synchronous method to determine `isLocalHubsUrl`, also works in local dev so avatar and scene links are detected properly in local dev
- all async methods that were calling `isLocalHubsUrl` changed to synchronous
- change `isHubsRoomUrl` to return a boolean to conform to naming pattern, add function `getHubsRoomIdFromUrl` to test a url and return its hubs id

This also seems to solve the issue where clicking links to rooms on other HC instances fail with error: No such Hub.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-1111)

### DIFF
--- a/src/utils/media-url-utils.js
+++ b/src/utils/media-url-utils.js
@@ -166,28 +166,17 @@ async function isHubsServer(url) {
   }
   const { origin } = new URL(url);
 
-  let ob = originIsHubsServer.get(origin);
-  if (ob) {
-    if (ob.hasOwnProperty("fetchPromise")) {
-      await ob.fetchPromise;
-    }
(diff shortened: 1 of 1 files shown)


yakyouk:
up :) is there anything else to review for this PR to be merged?

PR #4741, opened 2021-10-15
OMI Audio Emitter

robertlong:
Hey Hubs team :wave: long time no see! I'd like to get support for something we've been working on in OMI working in Hubs.

This PR adds support for the [OMI_audio_emitter](https://github.com/omigroup/gltf-extensions/pull/1) extension, a draft extension being worked on by the Open Metaverse Interoperability Group for embedding audio emitters in a glTF scene. Unlike with the `MOZ_hubs_components` `audio` component where audio media is referenced via a url, this extension is intended for audio assets to be embedded in the glTF file itself. This makes for more portable assets across various virtual world platforms. We've been building up tooling for producing these assets in [Blender](https://github.com/omigroup/omi-blender-gltf) and [from existing Spoke scenes](https://github.com/omigroup/omi-gltf-transform). We're also currently working on support for this in Three.js,  Unity, and Godot with the potential for editor-time import in Unity that will open up these assets for use in VR Chat, Neos, AltspaceVR, and any other Unity-based projects.

The extension is implemented in the [three-omi](https://github.com/omigroup/three-omi) package as a GLTFLoader plugin. It creates Three.js `Audio` and `PositionalAudio` objects in the loaded scene.

I tried to keep this PR as small as possible to make it easy to review. I think it's reasonably small, however there are a few fairly major changes that I should point out to get feedback on.

In my first pass at this PR I tried to patch Three.js to make `Audio` and `PositionalAudio` nodes clonable. This would enable using the existing glTF cache. The clone/copy methods did work for the most part, however they don't actually clone the underlying WebAudio nodes. As a result you can't safely have two of these glTFs in a scene because they are backed by a single Audio/PannerNode.

In my second pass I focused on making sure that we avoid caching these glTFs if they contain audio emitters. This is done with the new `gltf.cache` flag that is set in `loadGltf` which should be set to false when you don;t want the result to be cached. It get's a little funky when you have multiple inflight glTFs for the same model, but it calls `loadGltf` again and it should load fairly quickly with everything in your browser's cache.

The biggest change is in how uncached glTF objects are removed from the scene. Previously `parent.remove()` was used to clean them up, now I've switched it to use `this.el.object3DMap.mesh.traverse(disposeNode);`. I think this is a positive change, it means uncached models will be completely cleaned up when removed. This could have big implications for cleaning up after a scene change. However, it also could have unintended side effects.

Let me know what you think.

Here's a test asset to get going:
[OutdoorFestival_out.zip](https://github.com/mozilla/hubs/files/7352104/OutdoorFestival_out.zip)


mrmetaverse:
Bump.

antpb:
Just want to throw my +1 in here for this. My org is very interested in OMI as a standard to make our assets common across many platforms. Our aim at the moment is having two platforms to move content to and from (Hubs being one of them).

Supporting files with these components would be a huge win toward my company's goals of interoperability! In particular, we're using global audio emitters for ambient music in scenes. My hope would be that I could use that same file for our Hubs users as well as our other non-Hubs experiences that we have decided to build using `three-omi` standards.

robertlong:
I can update this PR at any point if there's any intention in merging it. Just let me know. If it's too big of a change, I can see what I can do to minimize impact further.

mrmetaverse:
Bump. ♥️ Please let us know how we can work better with Hubs.

PR #4780, opened 2021-10-26
Prevent media-frame race condition

CodingBitsDev:
I modified the release function to only update globaly when you are the owner of the mdia frame. If not then locally setting the frame to free should be enough.

### DIFF
--- a/src/systems/media-frames.js
+++ b/src/systems/media-frames.js
@@ -367,20 +367,23 @@ AFRAME.registerComponent("media-frame", {
   },
 
   release() {
-    if (NAF.utils.isMine(this.el) || NAF.utils.takeOwnership(this.el)) {
-      const capturedEl = document.getElementById(this.data.targetId);
-
+    const capturedEl = document.getElementById(this.data.targetId);
+    if (NAF.utils.isMine(this.el)) {
+      //Free media-frame globaly
       this.el.setAttribute("media-frame", {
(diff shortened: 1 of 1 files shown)


netpro2k:
Thanks for this PR, sorry for the very long delay on actually taking a look. I think you correctly identified the source of the problem. From what I can tell the root is this line https://github.com/mozilla/hubs/blob/c4386690043d6037242578576dbbebe92452ef6d/src/systems/media-frames.js#L99 since this will get called by every user every frame when an object has been removed. It should be only done by the owner of the frame (which is essentially what this PR does as well). I don/t think its correct for non owners to modify their local state, this should be handled automatically by the owner updating it, and a network update. Its likely it doesn't matter, but I would be a bit concerned about desyncing things by breaking the contract. I think my preferred fix would be to wrap this callsite with an owner check like all the other calls to release and capture, so only one user actually ends up running this code. 

Looking at this code I also noticed that in release we aren't checking for ownership of the capturedEl, just the frame. This may be different people, and should be treated separately. This would cause the scale change of the object to not be networked. I think simplest fix for that would be to have the person handling releasing the frame taking ownership of the released object. (this would not happen in the case causing this race condition anyway since the captured object no longer exists)

This code is definitely quite confusing and it would be better of the "owner-only" logic was split out from the "everyone" logic more.

PR #4788, opened 2021-10-27
Update to ThreeJS 133

netpro2k:
Using our new update process to update to ThreeJS 133 (from 128)

See https://github.com/mrdoob/three.js/wiki/Migration-Guide#128--129 and up.

I have gone through the migration guide and list of changes in the Hubs repo, still need to check in our dependency repos, though things appear to be working correctly so far in my tests.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -36307,7 +36307,7 @@
       "dev": true
     },
     "three": {
-      "version": "github:mozillareality/three.js#5415954e75dd12b3539339f7b7049ca655cf0330",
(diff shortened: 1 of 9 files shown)


netpro2k:
> do you know mozilla is working on a fix for this? [mrdoob/three.js#22631](https://github.com/mrdoob/three.js/issues/22631)

Good callout, hadn't noticed this issue in my testing but seems like we should maybe include https://github.com/mrdoob/three.js/pull/22759 for now...

PR #4790, opened 2021-10-27
Only try and release resources that have been cached

rawnsley:
The scene file and the object.gltf file are not cached, but when you run with "fast room switching" and they are updated to a new src URL, the `GLTFCache.release` function is called and this results in an erroneous console error.


netpro2k (approved):
Have had it in the back of my mind to investigate all those "releasing uncached GLTF" errors. Good to see they were probably mostly benign. I guess now I should stop mentally ignoring them. Thanks for the PR!

netpro2k:
Finally getting this merged.. thanks again!

PR #4797, opened 2021-11-01
Sign out on invalid auth tokens

johnshaughnessy:
This fixes an issue where the homepage doesn't render properly if the credentials saved to local storage are invalid. If the `ret` phoenix channel rejects a token with `reason: "invalid_token"` then we sign out (by deleting the offending credentials). 

Fixes https://github.com/mozilla/hubs-cloud/issues/187

### DIFF
--- a/src/react-components/auth/AuthContext.js
+++ b/src/react-components/auth/AuthContext.js
@@ -33,6 +33,10 @@ async function checkIsAdmin(socket, store) {
       })
       .receive("error", error => {
         console.error("Error joining Phoenix Channel", error);
+        if (error.reason === "invalid_token") {
+          console.error("Token is invalid. Signing out.");
+          store.update({ credentials: { token: null, email: null } });
+        }
       })
       .receive("timeout", () => {
         console.error("Phoenix Channel join timed out");
--- a/src/storage/store.js
+++ b/src/storage/store.js
@@ -34,6 +34,18 @@ export const defaultMaterialQualitySetting = (function() {
   return "high";
 })();
 
+function isInvalidOrExpired(token) {
+  try {
+    const expiry = jwtDecode(token).exp * 1000;
+    if (expiry <= Date.now()) {
+      return true;
+    }
+  } catch (e) {
+    return true; // If decoding throws an error, could be a malformed token. (e.name === "InvalidTokenError")
+  }
+  return false;
+}
+
 // Durable (via local-storage) schema-enforced state that is meant to be consumed via forward data flow.
 // (Think flux but with way less incidental complexity, at least for now :))
 export const SCHEMA = {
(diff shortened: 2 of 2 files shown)


src/storage/store.js
```diff
@@ -34,6 +34,18 @@ export const defaultMaterialQualitySetting = (function() {
...
 })();
 
+function isInvalidOrExpired(token) {
+  try {
+    const expiry = jwtDecode(token).exp * 1000;
+    if (expiry <= Date.now()) {
+      return true;
+    }
+  } catch (e) {
+    return true; // If decoding throws an error, could be a malformed token. (e.name === "InvalidTokenError")
```
netpro2k, line 44: We should log this at least.

src/storage/store.js
```diff
@@ -253,18 +265,24 @@ export default class Store extends EventTarget {
...
 
-    this._signOutOnExpiredAuthToken();
+    const signOutOnInvalidAuthToken = (() => {
+      let previousToken;
+      return () => {
+        const currentToken = this.state.credentials.token;
+        const tokenDidChange = currentToken !== previousToken;
+        previousToken = currentToken;
+        if (currentToken && tokenDidChange && isInvalidOrExpired(currentToken)) {
+          if (isInvalidOrExpired(currentToken)) {
```
netpro2k, line 275: This check seems redundant

src/storage/store.js
```diff
@@ -253,18 +265,24 @@ export default class Store extends EventTarget {
       Cookies.remove(OAUTH_FLOW_CREDENTIALS_KEY);
     }
 
-    this._signOutOnExpiredAuthToken();
+    const signOutOnInvalidAuthToken = (() => {
+      let previousToken;
+      return () => {
+        const currentToken = this.state.credentials.token;
+        const tokenDidChange = currentToken !== previousToken;
```
netpro2k, line 272: I think the expression reads just as clearly as the variable and it is only used once, so I would inline it.

src/react-components/auth/AuthContext.js
```diff
@@ -33,6 +33,10 @@ async function checkIsAdmin(socket, store) {
       })
       .receive("error", error => {
         console.error("Error joining Phoenix Channel", error);
+        if (error.reason === "invalid_token") {
+          console.error("Token is invalid. Signing out.");
+          store.update({ credentials: { token: null, email: null } });
```
netpro2k, line 38: You are doing this in a few spots. Probably good to add `signout()` to store. Though the actual sign out butotn calls `signOut()` on AuthChannel, which we might want to do instead, since it sets a few other things.

netpro2k (commented):
This makes sense. Made a few suggestions, mostly just cleanup, but I would confirm the AuthChannel signout thing. Also this change may hide larger issues so we should make sure we understand the cases where people are getting invalid tokens and make sure they are expected.

johnshaughnessy:
Closing this PR because I abandoned this work a long time ago

Issue #4800, opened 2021-11-03
Basis textures are not working


keianhzo:
**Description**
Basis texture loading is not not working in HMC

**To Reproduce**
1. Open a Hubs room an enter
2. Share an basis image
3. The object fails loading

**Expected behavior**
The Basis texture loads

**Screenshots**
<img width="1429" alt="Screen Shot 2021-11-03 at 09 01 02" src="https://user-images.githubusercontent.com/837184/140025983-cf041b11-4a5c-4095-9469-85bdca3bf883.png">

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-1135)

PR #4802, opened 2021-11-03
Update Basis loading code after ThreeJS 133.1 update

keianhzo:
Fixes https://github.com/mozilla/hubs/issues/4800


rawnsley:
FWIW this fixes BASISU import for me. We use a wide range of .basis textures in our scenes.

PR #4804, opened 2021-11-03
Allow objects to opt out of frustrum culling

rawnsley:
Frustrum culling can cause some animated models to disappear. While this situation could possibly be improved, it appears that extreme changes of mesh vertex positions due to skinning could not easily be accounted for.

This change allows the designer to force the animated model to be displayed regardless of whether the renderer thinks it is inside the frustrum or not.

### DIFF
--- a/src/components/frustrum.js
+++ b/src/components/frustrum.js
@@ -1,0 +1,24 @@
+/**
+ * Frustrum-based object visibilty
+ * @component frustrum
+ */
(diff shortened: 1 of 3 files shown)


rawnsley:
Add companion PR for Hubs Blender Exporter: https://github.com/MozillaReality/hubs-blender-exporter/pull/58

keianhzo (approved):
LGTM Looking forward to land this! Thanks!

Issue #4801, opened 2021-11-03
Personal space bubble has stopped working


keianhzo:
**Description**
The avatar personal space bubble is not working

**To Reproduce**
1. Enter a Hubs room
2. Make sure that there are two avatars in the room
3. Get inside the personal space of the other avatar
4. The avatar doesn't become transparent

**Expected behavior**
When an avatar enter another avatar's personal space bubble it should become transparent.

**Screenshot**
<img width="1137" alt="Screen Shot 2021-11-03 at 09 07 07" src="https://user-images.githubusercontent.com/837184/140026636-bd2c57d8-47d0-4a9d-aa4d-281329bc211b.png">

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-1136)

SV-AndreiC:
I have verified this issue and it's no longer reproducible. Now, the avatars become transparent when entering the personal space bubble.
Tested on Windows 10 using Firefox 94.0.1 and Chrome 95 browsers.

PR #4808, opened 2021-11-04
Fix bubble transparency issues

keianhzo:
Fixes https://github.com/mozilla/hubs/issues/4801

I've tracked it down to this PR: https://github.com/mrdoob/three.js/pull/22428 that affects the GLTF material import process. It seems that now the material format for GLTF imported materials is `RGBFormat` if blend mode is opaque so just enabling the transparency flag doesn't work anymore for imported materials with blend opaque modes, we need to switch to `RGBAFormat` first. It doesn't make much sense and there is an open issue about it https://github.com/mrdoob/three.js/issues/22598 so we can expect an update that fixes that soon. 

In the meantime I've patched that in the Hubs client code explicitly switching to `RGBAFormat` and also switch to `NormalBlending` (to make sure it works for all models) while the avatar is transparent. I had to also enable `DoubleSided` rendering, this shouldn't be necessary but it was failing for some models.

As far as I know this is the only place in the app where we need this patch, I can't think of any other place where we import GLTF models and we want to change their transparency on demand but I might be missing some things.

### DIFF
--- a/src/systems/personal-space-bubble.js
+++ b/src/systems/personal-space-bubble.js
@@ -254,7 +254,10 @@ AFRAME.registerComponent("personal-space-invader", {
           if (!material.userData.originalProps) {
             originalProps = material.userData.originalProps = {
               opacity: material.opacity,
-              transparent: material.transparent
+              transparent: material.transparent,
+              format: material.format,
+              blending: material.blending,
+              side: material.side
             };
           }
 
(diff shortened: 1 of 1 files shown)


netpro2k (approved):
Good digging, makes sense... Will definitely continue to monitor that ThreeJS discussion to see where it lands. Don't really have super strong opinions on it, but the current state feels weird for sure.

PR #4814, opened 2021-11-05
Refactor how components are loaded from GLTFs (fix video-texture-target)

netpro2k:
The latest version of ThreeJS broke node references on material components (used video-texture-target). This was due to a change inside GLTFLoader causing materials to be `.clone()`ed when they were not before (though there were certain cases in the past that would probably also trigger this bug). The implementation for `.clone()` on Material and Objec3D kills references on `userData` since it runs it through JSON.stringify/parse. Because of the timing of when this was happening relative to when we were resolving links on material components, the node reference was getting lost...

We want to run all our logic for resolving components after everything else in the GLTF has loaded, so we avoid issues like this. I took this as an opportunity to start cleaning this code up a bit and actually using "associations". This lets us get rid of the hack that had ThreeJS putting the component data into `userData` for us. We still want to skip putting it into `userData` at all, but that is a bigger change, and this gets us 1 step closer.

### DIFF
--- a/src/components/gltf-model-plus.js
+++ b/src/components/gltf-model-plus.js
@@ -1,5 +1,5 @@
 import nextTick from "../utils/next-tick";
-import { mapMaterials, convertStandardMaterial } from "../utils/material-utils";
+import { updateMaterials, mapMaterials, convertStandardMaterial } from "../utils/material-utils";
 import SketchfabZipWorker from "../workers/sketchfab-zip.worker.js";
 import { getCustomGLTFParserURLResolver } from "../utils/media-url-utils";
 import { promisifyWorker } from "../utils/promisify-worker.js";
@@ -401,25 +401,6 @@ class GLTFHubsPlugin {
         node.extras.gltfIndex = i;
       }
     }
-
-    function hookDef(defType, hookName) {
-      return Promise.all(
-        (parser.json[defType] || []).map((_def, idx) => {
-          return Promise.all(
-            parser._invokeAll(function(ext) {
-              return ext[hookName] && ext[hookName](idx);
-            })
-          );
-        })
-      );
-    }
-
-    // TODO decide if thse should get put into the GLTF loader itself
-    return Promise.all([
-      hookDef("scenes", "extendScene"),
-      hookDef("nodes", "extendNode"),
-      hookDef("materials", "extendMaterial")
-    ]);
   }
 
   afterRoot(gltf) {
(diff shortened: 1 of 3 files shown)


keianhzo (changes requested):
The `video-texture-target` component is not working for me using the hubs blender exporter 0.0.13. 

I'm still going over the new loader code but so far I see that since the introduction of the `__mhc_link_type` both in the `video-texture-target` and the `audio-target`components we are looking for an element inside the linked node but it's not there:

https://github.com/mozilla/hubs/blob/3f2893a4c2ac45d72e841f11d1487b0d86ed7a9c/src/gltf-component-mappings.js#L506

I guess as we are now relaying on the `GLTFHubsComponentsExtension` to resolve references, we don't have a aframe element linked from the node as we are not setting it anywhere?

Looking into the `indexToEntityMap` table fixes the issue:
` srcEl = indexToEntityMap[srcNode.userData.gltfIndex];`
But my understanding is that we want to get rid of that mapping.

Also the background doesn't seem to be rendering, I'm not sure if that was the case before...

I'll provide more insights when I'm done parsing code but I don't want to hold this review longer in case you already know where the problem is.

keianhzo:
I'm also seeing that the texture is video texture flipped in Y and the avatar head and the background is not rendered.

keianhzo (commented):
LGTM

keianhzo (approved):

keianhzo:
Now I understand the process better. The flow that caused the issue was:

1. The new GLTF object is loaded
2. `GLTFHubsComponentsExtension` was called and we resolve the `__mhc_link_type` replacing it with the loader created object instance.
3. Right after that the model is cached and cloned so we end up with a disparity in the instances in the `srcNode` reference in the component and the object in the scene graph:
https://github.com/mozilla/hubs/blob/3f2893a4c2ac45d72e841f11d1487b0d86ed7a9c/src/components/gltf-model-plus.js#L665-L670
4. The `.el` backlink is added to the cloned element
https://github.com/mozilla/hubs/blob/3f2893a4c2ac45d72e841f11d1487b0d86ed7a9c/src/components/gltf-model-plus.js#L198
5. The component is inflated:
https://github.com/mozilla/hubs/blob/3f2893a4c2ac45d72e841f11d1487b0d86ed7a9c/src/components/gltf-model-plus.js#L266

The result is that the `srcNode` object reference when the component is inflated is pointing to the original object that doesn't have the `el` link:
https://github.com/mozilla/hubs/blob/3f2893a4c2ac45d72e841f11d1487b0d86ed7a9c/src/gltf-component-mappings.js#L506

netpro2k:
> The result is that the `srcNode` object reference when the component is inflated is pointing to the original object that doesn't have the `el` backlink:
> 
> https://github.com/mozilla/hubs/blob/3f2893a4c2ac45d72e841f11d1487b0d86ed7a9c/src/gltf-component-mappings.js#L506

Correct. That is why we need to either update this link as part of cloning (like we do for animations) or just resolve the link later (what I did in this PR). It's a bit ugly that the logic for __mhc_link_type is now split into multiple parts, but I think a general refactor of all this component loading stuff is in order at some point anyway.

SV-AndreiC:
"I have verified this issue and it's working as intended. The video-texture-target is working and showing the intended captures. 
I used https://hubs.mozilla.com/scenes/Rpt8DJS/outdoor-festival for testing this.
Tested on Windows 10 using Firefox 94.0.1 and Chrome 95 browsers. "

PR #4818, opened 2021-11-05
Added IS_HMC config to configs, keep tweet button invisible for HC users

robinkwilson:


robinkwilson:
Better way to do this is use the "show_cloud" config for distinguishing between HC and HMC.

PR #4820, opened 2021-11-08
adding turkey gitops

tanfarming:
- added Dockerfile that builds nginx hosted static files packaged as per reticulum needs
  - container accepts following parameters
    - DOMAIN
    - SUB_DOMAIN
    - turkeyCfg_base_assets_path
    - turkeyCfg_non_cors_proxy_domains
    - turkeyCfg_reticulum_server
    - turkeyCfg_cors_proxy_server

### DIFF
--- a/RetPageOriginDockerfile
+++ b/RetPageOriginDockerfile
@@ -1,0 +1,32 @@
+###
+# this dockerfile produces image/container that serves customly packaged hubs and admin static files
+# the result container should serve reticulum as "hubs_page_origin" and "admin_page_origin" on (path) "/hubs/pages"
+###
+from node:16.13 as builder
+run mkdir -p /hubs/admin/ && cd /hubs
+copy package.json ./
+copy package-lock.json ./
+run npm ci
+copy admin/package.json admin/
+copy admin/package-lock.json admin/
+run cd admin && npm ci && cd ..
+copy . .
+env BASE_ASSETS_PATH="{{rawhubs-base-assets-path}}"
+run npm run build 1> /dev/null
+run cd admin && npm run build 1> /dev/null && cp -R dist/* ../dist && cd ..
+run mkdir -p dist/pages && mv dist/*.html dist/pages && mv dist/hub.service.js dist/pages && mv dist/schema.toml dist/pages          
+run mkdir /hubs/rawhubs && mv dist/pages /hubs/rawhubs && mv dist/assets /hubs/rawhubs && mv dist/react-components /hubs/rawhubs/pages && mv dist/favicon.ico /hubs/rawhubs/pages
+
+from alpine/openssl as ssl
+run mkdir /ssl && openssl req -x509 -newkey rsa:2048 -sha256 -days 36500 -nodes -keyout /ssl/key -out /ssl/cert -subj '/CN=hubs'
+
+from nginx:alpine
+run apk add bash
+run mkdir /ssl && mkdir -p /www/hubs && mkdir -p /www/hubs/pages && mkdir -p /www/hubs/assets
+copy --from=ssl /ssl /ssl
+copy --from=builder /hubs/rawhubs/pages /www/hubs/pages
+copy --from=builder /hubs/rawhubs/assets /www/hubs/assets
+copy scripts/docker/nginx.config /etc/nginx/conf.d/default.conf
+copy scripts/docker/run.sh /run.sh
+run chmod +x /run.sh && cat /run.sh
+cmd bash /run.sh
--- a/scripts/docker/run.sh
+++ b/scripts/docker/run.sh
@@ -1,0 +1,19 @@
+
+# TODO: need a better one
+healthcheck(){
+    while true; do (echo -e 'HTTP/1.1 200 OK\r\n\r\n 1') | nc -lp 1111 > /dev/null; done
+}
+
+find /www/hubs/ -type f -name *.html -exec sed -i "s/{{rawhubs-base-assets-path}}\//https:\/\/${SUB_DOMAIN}-assets.${DOMAIN}\/hubs\//g" {} \;           
+find /www/hubs/ -type f -name *.html -exec sed -i "s/{{rawhubs-base-assets-path}}/https:\/\/${SUB_DOMAIN}-assets.${DOMAIN}\/hubs\//g" {} \; 
+find /www/hubs/ -type f -name *.css -exec sed -i "s/{{rawhubs-base-assets-path}}\//https:\/\/${SUB_DOMAIN}-assets.${DOMAIN}\/hubs\//g" {} \; 
+find /www/hubs/ -type f -name *.css -exec sed -i "s/{{rawhubs-base-assets-path}}/https:\/\/${SUB_DOMAIN}-assets.${DOMAIN}\/hubs\//g" {} \;             
+anchor="<!-- DO NOT REMOVE\/EDIT THIS COMMENT - META_TAGS -->" 
+for f in /www/hubs/pages/*.html; do 
+    for var in $(printenv); do 
+    var=$(echo $var | cut -d"=" -f1 ); prefix="turkeyCfg_"; 
+    [[ $var == $prefix* ]] && sed -i "s/$anchor/ <meta name=\"env:${var#$prefix}\" content=\"${!var//\//\\\/}\"\/> $anchor/" $f; 
+    done 
+done 
+healthcheck &
+nginx -g "daemon off;"
--- a/.github/workflows/hubs-RetPageOrigin.yml
+++ b/.github/workflows/hubs-RetPageOrigin.yml
@@ -1,0 +1,78 @@
+name: hubs
+on:
+  push:
+    paths-ignore: ["README.md"]
+  workflow_dispatch:
+
+env:
+  registryName: mozillareality
+  containerName: hubs
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout-hubs
+        uses: actions/checkout@v2
+        with:
+          path: "./hubs"
+      - name: docker setup buildx
+        uses: docker/setup-buildx-action@v1
+        with:
+          install: true
+      - name: docker login
+        uses: docker/login-action@v1
+        with:
+          username: ${{ env.registryName }}
+          password: ${{ secrets.DOCKER_HUB_PWD }}
+      - name: docker build(x) push
+        uses: docker/build-push-action@v2
+        with:
+          context: hubs/
+          file: hubs/RetPageOriginDockerfile
+          tags: ${{ env.registryName }}/${{ env.containerName }}:${{ github.run_number  }}
+          cache-from: type=registry,ref=${{ env.registryName }}/${{ env.containerName }}:buildcache
+          cache-to: type=registry,ref=${{ env.registryName }}/${{ env.containerName }}:buildcache,mode=max
+          push: true
+
+  # turkeyGitops:
+  #   needs: build
+  #   uses: mozilla/hubs-ops/.github/workflows/turkeyGitops.yml@feature/gha-template
+  #   with:
+  #     promoteFrom: mozillareality/hubs:${{ github.run_number }}
+  #     promoteTo: mozillareality/hubs:${GITHUB_REF/'refs/heads/'/}
+  #     # https://docs.github.com/en/actions/learn-github-actions/reusing-workflows#limitations
+  #     #   Any environment variables set in an env context defined at the workflow level in the caller workflow are not propagated to the called workflow.
+  #     # promoteFrom: ${{ env.registryName }}/${{ env.containerName }}:${{ github.run_number }}
+  #     # promoteTo: ${{ env.registryName }}/${{ env.containerName }}:${GITHUB_REF/'refs/heads/'/}
+
+  promote:
+    if: github.ref == 'refs/heads/dev' || github.ref == 'refs/heads/staging' || github.ref == 'refs/heads/prod'
+    runs-on: ubuntu-latest
+    needs: build
+    steps:
+      - name: promote artifact
+        run: |
+          fromTag=${{ env.registryName }}/${{ env.containerName }}:${{ github.run_number }}
+          toTag=${{ env.registryName }}/${{ env.containerName }}:${GITHUB_REF/'refs/heads/'/}
+          echo ${{ secrets.DOCKER_HUB_PWD }} | sudo docker login --username $registryName --password-stdin
+          docker pull $fromTag && docker tag $fromTag $toTag && sudo docker push $toTag
+          echo "promoted :$fromTag to :$toTag"
+
+  staging_pr_to_prod:
+    if: github.ref == 'refs/heads/staging'
+    runs-on: ubuntu-latest
+    needs: promote
+    steps:
+      - name: create pr for staging -> prod
+        run: |
+          gh pr create -H staging -B prod
+
+  prod_cut_release_branch:
+    if: github.ref == 'refs/heads/prod'
+    runs-on: ubuntu-latest
+    needs: promote
+    steps:
+      - name: cut release branch for prod
+        run: |
+          git checkout -b releases/$(date '+%y%m%d').$GITHUB_RUN_NUMBER
--- a/scripts/docker/nginx.config
+++ b/scripts/docker/nginx.config
@@ -1,0 +1,10 @@
+server {
+    listen 8080 ssl;
+    ssl_certificate /ssl/cert;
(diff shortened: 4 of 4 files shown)


Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+run mkdir rawhubs && mv dist/pages rawhubs && mv dist/assets rawhubs && mv dist/react-components rawhubs/pages && mv dist/favicon.ico rawhubs/pages
+
+from alpine/openssl as ssl
+run mkdir /ssl && openssl req -x509 -newkey rsa:2048 -sha256 -days 36500 -nodes -keyout /ssl/key -out /ssl/cert -subj '/CN=hubs'
+
+from nginx:alpine
+run apk add bash
+run mkdir /ssl && mkdir -p /www/hubs && mkdir -p /www/hubs/pages && mkdir -p /www/hubs/assets
+copy --from=ssl /ssl /ssl
+copy --from=builder /hubs/rawhubs/pages /www/hubs/pages
```
robinkwilson, line 20: I'd add rawhubs/ to the `.gitignore` file.

Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+for f in /www/hubs/pages/*.html; do \n\
+    for var in $(printenv); do \n\
+    var=$(echo $var | cut -d"=" -f1 ); prefix="turkeyCfg_"; \n\
+    [[ $var == $prefix* ]] && sed -i "s/$anchor/ <meta name=\"env:${var#$prefix}\" content=\"${!var//\//\\\/}\"\/> $anchor/" $f; \n\
+    done \n\
+done \n\
+/healthcheck.sh& \n\
+nginx -g "daemon off;" \n\n' > /run.sh
+run chmod +x /run.sh && cat /run.sh
+cmd bash /run.sh
```
robinkwilson, line 39: Where is the run.sh file?

Dockerfile
```diff
@@ -0,0 +1,39 @@
+
```
robinkwilson, line 1: nit, if it's relevant: Are there any dependencies or instructions to add to run this in the README?

robinkwilson (commented):
beginner docker question: is there a file depicting all of the dependencies like node or npm versions that hubs runs on? Is that included here or another place.

robinkwilson:
Ah I see this is only as it relates to turkey. So this is NOT relevant to our users wanting to run hubs using a docker file. 

Can you add a comment at the top of this docker file explaining it.

Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+env BASE_ASSETS_PATH="{{rawhubs-base-assets-path}}"
+run npm ci 
+run npm run build
+run cd admin && npm ci && npm run build && cp -R dist/* ../dist && cd ..
+run mkdir -p dist/pages && mv dist/*.html dist/pages && mv dist/hub.service.js dist/pages && mv dist/schema.toml dist/pages          
+#[info] rearrange files
+run mkdir rawhubs && mv dist/pages rawhubs && mv dist/assets rawhubs && mv dist/react-components rawhubs/pages && mv dist/favicon.ico rawhubs/pages
+
+from alpine/openssl as ssl
+run mkdir /ssl && openssl req -x509 -newkey rsa:2048 -sha256 -days 36500 -nodes -keyout /ssl/key -out /ssl/cert -subj '/CN=hubs'
```
netpro2k, line 14: What is this for? I guess this is for ssl between reticulum and the hubs source assets, though these are self signed certs so I guess it has to ignore their validity? Is this just to ensure encryption over the wire? These are public assets anyway I think, so this nginx service can probably just be running http. I might also just be missing something obvious.

Dockerfile
netpro2k: It is generated in the lines above. Though I suggested we should actually pull this out and not generate it at build time.

netpro2k (commented):
This is looking good. I would pull out the two .sh files rather than generating them in the dockerfile and I had some questions re-caching to make sure we make builds as fast as possible. Getting `npm ci` in particular to only run when actually necessary would drastically cut down build times. Also was a bit confused why the ssl stuff is needed.

Dockerfile
```diff
@@ -0,0 +1,39 @@
+
+from node:lts as builder
+workdir hubs
+copy . .
```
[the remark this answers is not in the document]
tanfarming: yeah i was thinking to run in only in github action's hosted agent so caching's not applicable there, we do have option to use github action's caching feature but i'm not yet sure if the few minutes saved would worth the cost of extra pipeline complexity + losing the portability of wrapping everything inside dockerfile
I think for local testing a custom Dockerfile should be used, and depends on the individual developer it could be different, maybe we can later add an example in the readme for it?

Dockerfile
tanfarming: long story short / if i remembered correctly -- it's for ret, it only wants assets served on https
we technically can terminate https on ingress (fraefik / haproxy) otherwise, i just didn't want to bother with more ret code changes lol

Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+run mkdir /ssl && openssl req -x509 -newkey rsa:2048 -sha256 -days 36500 -nodes -keyout /ssl/key -out /ssl/cert -subj '/CN=hubs'
+
+from nginx:alpine
+run apk add bash
+run mkdir /ssl && mkdir -p /www/hubs && mkdir -p /www/hubs/pages && mkdir -p /www/hubs/assets
+copy --from=ssl /ssl /ssl
+copy --from=builder /hubs/rawhubs/pages /www/hubs/pages
+copy --from=builder /hubs/rawhubs/assets /www/hubs/assets
+run echo "server {listen 8080 ssl;ssl_certificate /ssl/cert;ssl_certificate_key /ssl/key; location / {root /www;autoindex off;add_header 'Access-Control-Allow-Origin' '*';}}" > /etc/nginx/conf.d/default.conf
+run printf 'while true; do (echo -e "HTTP/1.1 200 OK\r\n") | nc -lp 1111 > /dev/null; done' > /healthcheck.sh && chmod +x /healthcheck.sh
```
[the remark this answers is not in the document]
tanfarming: yep will do

Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+run npm run build
+run cd admin && npm ci && npm run build && cp -R dist/* ../dist && cd ..
+run mkdir -p dist/pages && mv dist/*.html dist/pages && mv dist/hub.service.js dist/pages && mv dist/schema.toml dist/pages          
+#[info] rearrange files
+run mkdir rawhubs && mv dist/pages rawhubs && mv dist/assets rawhubs && mv dist/react-components rawhubs/pages && mv dist/favicon.ico rawhubs/pages
+
+from alpine/openssl as ssl
+run mkdir /ssl && openssl req -x509 -newkey rsa:2048 -sha256 -days 36500 -nodes -keyout /ssl/key -out /ssl/cert -subj '/CN=hubs'
+
+from nginx:alpine
```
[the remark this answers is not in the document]
tanfarming: jargon's "multi stage docker build"
mainly to slim down the final product, ie we need sdk to build, but just runtime to host
i haven't optimized the dockerfile yet, ie security, further slim-downs, etc, this push is just to get the devops started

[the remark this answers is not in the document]
netpro2k: Cool. Will read up on that.

Dockerfile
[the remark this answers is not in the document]
netpro2k: Ah yeah, I assumed github actions just had built in handling for this... 

Some quick searching shows it shouldn't be too hard to setup https://github.com/progapandist/anycable_rails_demo/blob/master/.github/workflows/fake_deploy_multistage.yml (from https://evilmartians.com/chronicles/build-images-on-github-actions-with-docker-layer-caching) by using github's cache action to cache the build cache directory. 

Re  "extra pipeline complexity + losing the portability of wrapping everything inside dockerfile" I don't think this moves anything out of the dockerfile, just configures the environment it is running in. Not the highest priority ever, but shaving off build time definitely adds up. Our current CI builds take quite a long time, and its highly dominated by `npm ci` time which actually changes quite rarely. Even if we don't go through setting up github actions to enable caching, making this part of the dockerfile more cache friendly seems good (unless there is some tradeoff)

netpro2k:
@robinkwilson 
> beginner docker question: is there a file depicting all of the dependencies like node or npm versions that hubs runs on? Is that included here or another place.

The `from node:16.13 as builder` line in the Dockerfile is responsible for what version of node/npm gets used during the build. At runtime its just static assets so only nginx is involved.

Dockerfile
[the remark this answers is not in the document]
tanfarming: yup agreed i'll add the github action caching -- just discovered yesterday that file ops are like 4-6 times slower in "docker file system" so the original ~6 min build time becomes ~25 min inside docker
the extra github action caching step totally worth it in this case

tanfarming:
> @robinkwilson
> 
> > beginner docker question: is there a file depicting all of the dependencies like node or npm versions that hubs runs on? Is that included here or another place.
> 
> The `from node:16.13 as builder` line in the Dockerfile is responsible for what version of node/npm gets used during the build. At runtime its just static assets so only nginx is involved.
yep, not only node version, OS and all other files/packages are pinned down too -- it's basically a "docker managed VM", cloud jargon here is "immutable infra"
with npm ci / package-lock.json there's no moving parts other than our code

tanfarming:
@netpro2k @robinkwilson

updated:
- changed dockerfile's name and added comments
- inline run.sh moved to ./scripts/docker/
- inline healthcheck.sh become a function inside run.sh
- docker cache layers on github action enabled with buildx + type=registry after wasted days on the buggy docker/buildx/buildKit
  - type=local + diy move files: hackish, extra steps, a bit slower
  - type=gha: docker/buildx wants us to use this, still "experimental"
  - type=registry: seems to be the cleanest
  - docker's buildx' still buggy, [for example](https://github.com/mozilla/hubs/runs/4193826833?check_suite_focus=true#step:5:178), hope it will get better
  - docker file system and "non-ancient" npm versions are not playing well together, it seems npm runs many times slower inside docker, but lets worry about this later
    - [npm on agent with fresh npm ci: ~8min](https://github.com/mozilla/turkey-ops/actions/runs/1199621038)
    - [npm inside docker file system with fresh npm ci: ~33min](https://github.com/mozilla/hubs/actions/runs/1453911447)
    - [npm inside docker file system with npm ci cached: ~10min](https://github.com/mozilla/hubs/actions/runs/1454494526)

RetPageOriginDockerfile
```diff
@@ -0,0 +1,30 @@
+###
+# this dockerfile produces image/container that serves customly packaged hubs and admin static files
+# the result container should serve reticulum as "hubs_page_origin" and "admin_page_origin" on (path) "/hubs/pages"
+###
+from node:16.13 as builder
+run mkdir -p /hubs/admin/ && cd /hubs
+copy package*.json ./
+copy admin/package*.json admin/
+run npm ci
```
johnshaughnessy, line 9: Should the `package-lock.json` checked into the repo be built with node 16.13 to prevent the following?
```
Step 5/24 : run npm ci
 ---> Running in 45dcecf84c37
npm WARN old lockfile
npm WARN old lockfile The package-lock.json file was created with an old version of npm,
npm WARN old lockfile so supplemental metadata must be fetched from the registry.
npm WARN old lockfile
npm WARN old lockfile This is a one-time fix-up, please be patient...
```

RetPageOriginDockerfile
```diff
@@ -0,0 +1,30 @@
+###
+# this dockerfile produces image/container that serves customly packaged hubs and admin static files
+# the result container should serve reticulum as "hubs_page_origin" and "admin_page_origin" on (path) "/hubs/pages"
+###
+from node:16.13 as builder
+run mkdir -p /hubs/admin/ && cd /hubs
+copy package*.json ./
+copy admin/package*.json admin/
```
johnshaughnessy, line 8: nit: Since we only seek to copy `package.json` and `package-lock.json`, maybe these should list them explicitly instead of catching them with a wildcard:
```
copy package.json ./
copy package-lock.json ./
copy admin/package.json admin/
copy admin/package-lock.json admin/
```

johnshaughnessy:
> docker's buildx' still buggy, [for example](https://github.com/mozilla/hubs/runs/4193826833?check_suite_focus=true#step:5:178), hope it will get better

What bug you are showing in that example? I see the same output spammed repeatedly
```
#22 [builder  6/11] RUN cd admin && npm ci && cd ..
#22 CACHED
```
but don't know enough about buildx to know how to interpret this.

RetPageOriginDockerfile
tanfarming: yep, but we'll do it in future when we are updating hubs' dependencies, this is just an initial devops update for turkey and i'd like to minimize any disturbance to the existing codebase

RetPageOriginDockerfile
tanfarming: yep will do

tanfarming:
> > docker's buildx' still buggy, [for example](https://github.com/mozilla/hubs/runs/4193826833?check_suite_focus=true#step:5:178), hope it will get better
> 
> What bug you are showing in that example? I see the same output spammed repeatedly
> 
> ```
> #22 [builder  6/11] RUN cd admin && npm ci && cd ..
> #22 CACHED
> ```
> 
> but don't know enough about buildx to know how to interpret this.
yeaht that's a bad example, i saw the step 6/11 repeated so many times and got confused and thought it's a bug, googled it and find some other guys online also got confused
until yesterday Brian and Dom pointed out to me that it's probably just downloading
I'm pretty green with their buildkit/buildx too, tried it out a few years ago, but went back to vanilla docker build, 
there are other issues, I ran into a few of them last weekend, for example this one https://github.com/moby/buildkit/issues/1939
we'll see how it goes, I'm using buildkit because there's fewer steps and also it seems to be where things are going (with future native github action support etc) -- we do have other options, like using vanilla docker build with diy file copying etc to back up/restore local caches with github action's "file caches"

Dockerfile
brianpeiris: I think changing ret to remove the ssl requirement is worth it later on, but it's okay to do this for now.

RetPageOriginDockerfile
brianpeiris: I think this might be one of the reasons the build is slow. That "one-time fix-up" will happen every time and it downloads quite a lot of metadata, which slows things down. As you noted earlier, it is much faster with an older version of node/npm.

But I agree, this is a problem to be solved on the hubs side, not a problem with this workflow, so it can be tackled at a later time.

scripts/docker/run.sh
```diff
@@ -0,0 +1,19 @@
...
+
+find /www/hubs/ -type f -name *.html -exec sed -i "s/{{rawhubs-base-assets-path}}\//https:\/\/${SUB_DOMAIN}-assets.${DOMAIN}\/hubs\//g" {} \;           
+find /www/hubs/ -type f -name *.html -exec sed -i "s/{{rawhubs-base-assets-path}}/https:\/\/${SUB_DOMAIN}-assets.${DOMAIN}\/hubs\//g" {} \; 
+find /www/hubs/ -type f -name *.css -exec sed -i "s/{{rawhubs-base-assets-path}}\//https:\/\/${SUB_DOMAIN}-assets.${DOMAIN}\/hubs\//g" {} \; 
+find /www/hubs/ -type f -name *.css -exec sed -i "s/{{rawhubs-base-assets-path}}/https:\/\/${SUB_DOMAIN}-assets.${DOMAIN}\/hubs\//g" {} \;             
+anchor="<!-- DO NOT REMOVE\/EDIT THIS COMMENT - META_TAGS -->" 
+for f in /www/hubs/pages/*.html; do 
+    for var in $(printenv); do 
+    var=$(echo $var | cut -d"=" -f1 ); prefix="turkeyCfg_"; 
+    [[ $var == $prefix* ]] && sed -i "s/$anchor/ <meta name=\"env:${var#$prefix}\" content=\"${!var//\//\\\/}\"\/> $anchor/" $f;
```
brianpeiris, line 15:
```sh
${!var//\//\\\/}
```
Is this substitution actually doing what you intended? I stared at it for a while and tried to run it myself in a terminal, but I can't quite understand what the intent was. In fact, I think it might be wrong at the moment. Might be helpful to explain it in a comment in this script as well, since it's hard to read without thinking about it for a while.

tanfarming: yeah substitution part is just swapping `/` for `\/`, for the slashes in `base_assets_path`
both sed and bash parameter substitution do not like `/` so it needs to be escaped into `\/`
sed does not like `\` either, so the `\/` part end up got escaped into `\\\/`

PR #4821, opened 2021-11-08
add audio debug panel to sb + remove max distance from inverse/exponential

matthewbcool:
Adds audio debugger to storybook and removes the max distance select property for inverse and exponential since we only use maxDistance to calculate linear distance models.

[Image from Gyazo](https://gyazo.com/20c78cf2c834190144a9d2d566172bbf)

### DIFF
--- a/src/react-components/debug-panel/AudioDebugPanel.scss
+++ b/src/react-components/debug-panel/AudioDebugPanel.scss
@@ -8,6 +8,7 @@
   align-self: flex-start;
   flex: 0 1 auto;
   justify-content: center;
+  
 
   @media(max-width: 500px) {
     width: 100%;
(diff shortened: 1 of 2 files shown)


netpro2k:
Nice! 

I have been thinking a bit about our debug tooling, we have sort of a mishmosh of different things and it would be nice to consolidate. For some things we are using datGUI, others we have custom react UIs. 

I usually reach for datGUI because its very quick to configure something usable for debugging, but there is no reason we couldn't have a set of react widgets that are equally easy to set up. I am finding datGUI to be annoying to extend (I have run into this recently since I want to add a preview for the environment map to `envSettingsDebug`).

Definitely not super high priority but getting a design pass on a set of widgets and than a bit of engineering work on a nice way to set up some quick debug UIs would be nice to do. The design part definitely the lower of the 2 priorities. Function very much over form when it comes to debug tools 😃

Don't think we should necessarily actually use any of these but in looking into this came across a few projects for inspiration at least:
https://github.com/claus/react-dat-gui
https://github.com/mat-sz/react-var-ui
https://github.com/pmndrs/leva

PR #4824, opened 2021-11-09
Popup window closes itself on oauth-success

robinkwilson:
Firefox blocks the twitter pop-up so Hubs loses the reference to the window and can't close it. This enables the popup oauth window to close itself on success.


SV-AndreiC:
I have verified this issue and it's working as intended. The popup oauth window closes itself on success. 
Tested on Windows 10 using Firefox 94.0.1 and Chrome 95 browsers.

PR #4833, opened 2021-11-11
GLTF loader Draco support

keianhzo:
Add support for Draco compressed geometry int he GLTF loader. You can test it exporting any glb from Blender enabling the geometry compression.


netpro2k (approved):
lgtm

PR #4839, opened 2021-11-16
Expose left/right panning as a pref

### LINKED ISSUE #4639: Improve "disabling l/r panning" accessibility feature discoverability
The disable left/right panning feature is hidden behind a chat command, it's hard to discover and the name doesn't make it clear what it does.

We should improve the feature discoverability exposing it as a preference under Audio/Accessibility so it's easier to find.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-896)

keianhzo:
Closes #4639

This PR exposes Left/Right panning as an accessibility setting. Thing to note:
- We no longer use the `audioOutputMode` preference as it has been outlined in the code comments as confusing, now we have a `disableLeftRightPanning` setting and the `audiomode` command updates that pref.
https://github.com/mozilla/hubs/blob/fcf78a30652cb745f90d8e20159ade3031157d2a/src/message-dispatch.js#L179-L181
- This PR removes the initial panning mode pref resetting that was happening before.
https://github.com/mozilla/hubs/blob/fcf78a30652cb745f90d8e20159ade3031157d2a/src/systems/audio-settings-system.js#L8-L17
- Also we no longer log in the console when the mode is changed (we don't do that for any other pref either so it kind of makes sense now that it has graduated to pref)

### DIFF
--- a/src/components/avatar-audio-source.js
+++ b/src/components/avatar-audio-source.js
@@ -100,11 +100,11 @@ AFRAME.registerComponent("avatar-audio-source", {
     APP.dialog.on("stream_updated", this._onStreamUpdated, this);
     this.createAudio();
 
-    let audioOutputModePref = APP.store.state.preferences.audioOutputMode;
+    let disableLeftRightPanningPref = APP.store.state.preferences.disableLeftRightPanning === true;
     this.onPreferenceChanged = () => {
-      const newPref = APP.store.state.preferences.audioOutputMode;
-      const shouldRecreateAudio = audioOutputModePref !== newPref && !this.isCreatingAudio;
-      audioOutputModePref = newPref;
+      const newPref = APP.store.state.preferences.disableLeftRightPanning;
+      const shouldRecreateAudio = disableLeftRightPanningPref !== newPref && !this.isCreatingAudio;
+      disableLeftRightPanningPref = newPref;
       if (shouldRecreateAudio) {
         this.createAudio();
       }
--- a/src/components/avatar-audio-source.js
+++ b/src/components/avatar-audio-source.js
@@ -100,7 +100,7 @@ AFRAME.registerComponent("avatar-audio-source", {
     APP.dialog.on("stream_updated", this._onStreamUpdated, this);
     this.createAudio();
 
-    let disableLeftRightPanningPref = APP.store.state.preferences.disableLeftRightPanning === true;
+    let disableLeftRightPanningPref = APP.store.state.preferences.disableLeftRightPanning;
     this.onPreferenceChanged = () => {
       const newPref = APP.store.state.preferences.disableLeftRightPanning;
       const shouldRecreateAudio = disableLeftRightPanningPref !== newPref && !this.isCreatingAudio;
--- a/src/message-dispatch.js
+++ b/src/message-dispatch.js
@@ -66,8 +66,7 @@ export default class MessageDispatch extends EventTarget {
     uiRoot = uiRoot || document.getElementById("ui-root");
     const isGhost = !entered && uiRoot && uiRoot.firstChild && uiRoot.firstChild.classList.contains("isGhost");
 
-    // TODO: Some of the commands below should be available without requiring
-    //       room entry. For example, audiomode should not require room entry.
+    // TODO: Some of the commands below should be available without requiring room entry.
     if (!entered && (!isGhost || command === "duck")) {
       this.log(LogMessageType.roomEntryRequired);
       return;
@@ -170,13 +169,6 @@ export default class MessageDispatch extends EventTarget {
             captureSystem.start();
             this.log(LogMessageType.captureStarted);
           }
-        }
-        break;
-      case "audiomode":
-        {
-          window.APP.store.update({
-            preferences: { disableLeftRightPanning: args[0] === "panner" ? false : true }
-          });
         }
         break;
       case "audioNormalization":
--- a/src/message-dispatch.js
+++ b/src/message-dispatch.js
@@ -174,20 +174,9 @@ export default class MessageDispatch extends EventTarget {
         break;
       case "audiomode":
         {
-          const shouldEnablePositionalAudio = window.APP.store.state.preferences.audioOutputMode === "audio";
           window.APP.store.update({
-            // TODO: This should probably just be a boolean to disable panner node settings
-            // and even if it's not, "audio" is a weird name for the "audioOutputMode" that means
-            // "stereo" / "not panner".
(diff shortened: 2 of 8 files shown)


src/systems/sound-effects-system.js
```diff
@@ -138,7 +138,7 @@ export class SoundEffectsSystem {
     const audioBuffer = this.sounds.get(sound);
     if (!audioBuffer) return null;
 
-    const disablePositionalAudio = isSafari() || window.APP.store.state.preferences.audioOutputMode === "audio";
+    const disablePositionalAudio = isSafari() || window.APP.store.state.preferences.disableLeftRightPanning === true;
```
johnshaughnessy, line 141: Can `window.APP.store.state.preferences.disableLeftRightPanning === true` be reduced to `window.APP.store.state.preferences.disableLeftRightPanning` here?

src/storage/store.js
```diff
@@ -96,7 +96,6 @@ export const SCHEMA = {
         preferredMic: { type: "string" },
         preferredCamera: { type: "string" },
         muteMicOnEntry: { type: "bool" },
-        audioOutputMode: { type: "string" },
```
johnshaughnessy, line 99: If you remove this key from the store, but it was set in a user's local storage because in their most recent setting they set their `audioOutputMode` pref (via the command), then when the user loads the page, the data in their local storage will cause them to fail the schema validation and the page will fail to load.

Before schema validation runs the first time, we need to delete this entry from users' local storage. Alternatively, we can leave it in the schema definition for a while (with code anywhere that deletes the stored value) with a comment that says, "Deprecated, to be removed once we're confident local storage has been cleaned up."

I think I like the first option better, where before doing schema validation we delete now-deprecated keys if they exist.

src/message-dispatch.js
```diff
@@ -174,20 +174,9 @@ export default class MessageDispatch extends EventTarget {
         break;
       case "audiomode":
         {
-          const shouldEnablePositionalAudio = window.APP.store.state.preferences.audioOutputMode === "audio";
           window.APP.store.update({
-            // TODO: This should probably just be a boolean to disable panner node settings
-            // and even if it's not, "audio" is a weird name for the "audioOutputMode" that means
-            // "stereo" / "not panner".
-            preferences: { audioOutputMode: shouldEnablePositionalAudio ? "panner" : "audio" }
+            preferences: { disableLeftRightPanning: args[0] === "panner" ? false : true }
```
johnshaughnessy, line 178: We can remove this chat command

src/components/avatar-audio-source.js
```diff
@@ -100,11 +100,11 @@ AFRAME.registerComponent("avatar-audio-source", {
     APP.dialog.on("stream_updated", this._onStreamUpdated, this);
     this.createAudio();
 
-    let audioOutputModePref = APP.store.state.preferences.audioOutputMode;
+    let disableLeftRightPanningPref = APP.store.state.preferences.disableLeftRightPanning === true;
```
johnshaughnessy, line 103: I think you can drop `=== true`.

johnshaughnessy (changes requested):
My only major concern has to do with removing keys from the `store`. If a user has a key-value pair for `audioOutputMode`, this change will cause the schema validation to fail, which is an error we currently do not catch or recover from.

src/systems/sound-effects-system.js
keianhzo: yep, I think so, we are only using it for the ternary. Updated.

src/message-dispatch.js
keianhzo: I guess there is no point on keeping that around for compat. Removed.

src/components/avatar-audio-source.js
keianhzo: In this case (same as in media-video) I think we want to only recreate in case it went from undefined/false <-> true and not if it changed from undefined <-> false

keianhzo: Thinking about it again now that we don't have the chat command, I don't think there is a case where this goes from `undefined` to `false` so I think is safe to remove that comparison.

src/storage/store.js
keianhzo: I think we switched the preferences store to accept additional properties here:
https://github.com/mozilla/hubs/pull/4275/commits/e865eb78812980297b6ce6fae5236e3b737b0911#r636548037
That's why it doesn't really crash (and also why it doesn't crash even though I forgot to add the actual new `disableLeftRightPanning` pref 😬) I wonder if we should write cleanup code though so old prefs are removed and the store doesn't hold stall prefs.

keianhzo:
@johnshaughnessy I've fixed all the issues and added a final commit to always remove unsupported properties from the store after validating. As we don't want to accept unsupported properties I think this should help keep the store clean.

src/storage/store.js
johnshaughnessy: Oh, great!

src/storage/store.js
```diff
@@ -360,12 +359,13 @@ export default class Store extends EventTarget {
...
+    const { valid, errors } = validator.validate(finalState, SCHEMA);
 
+    // Cleanup unsupported properties
     if (!valid) {
-      // Intentionally not including details about the state or validation result here, since we don't want to leak
-      // sensitive data in the error message.
-      throw new Error(`Write to store failed schema validation.`);
+      errors.forEach(error => {
+        delete error.instance[error.argument];
+      });
```
johnshaughnessy, line 368: It seems like we should at least log these errors. Also, is this change necessary/desirable given the `allowAdditionalProps` change you referred to in an earlier comment.

src/update-audio-settings.js
```diff
@@ -36,7 +36,7 @@ export function getCurrentAudioSettings(el) {
   const audioDebugPanelOverrides = APP.audioDebugPanelOverrides.get(sourceType);
   const zoneSettings = APP.zoneOverrides.get(el);
   const preferencesOverrides =
-    APP.store.state.preferences.audioOutputMode === "audio" ? { audioType: AudioType.Stereo } : {};
+    APP.store.state.preferences.disableLeftRightPanning === true ? { audioType: AudioType.Stereo } : {};
```
johnshaughnessy, line 39: We can lose `=== true`

johnshaughnessy (approved):

src/storage/store.js
keianhzo: First comment here https://github.com/mozilla/hubs/pull/4275/files#r636548037 talked about cleaning up the store to avoid stall prefs to stick around forever. I thought it was a good time for adding that.

PR #4849, opened 2021-11-22
Fix/scene UI styles

matthewbcool:
Edits to scene-ui primarily to fix illegible color contrast on hover styles and bring us closer to the new design here: https://www.figma.com/file/J6YKkb2vWDlk81zm4kxUzA/Hubs-discoverability-and-onboarding?node-id=903%3A1324

Note scene-ui.scss and scene.ui is not using conventions outlined in https://github.com/mozilla/hubs/blob/1d747f73d558335b51b1e0811c5187dd91198ab2/doc/ui-best-practices.md and I was advised not to update/refactor to follow these conventions for now.

image

image


keianhzo (approved):
LGTM!

PR #4855, opened 2021-11-23
Troika Text node

jameskane05:
This is the Hubs PR for displaying Troika Text nodes as defined in https://github.com/mozilla/Spoke/pull/1195, based on the troika-three-text package. This package supports a wide range of styling options and has the advantage of being able to display 3D text consistently in Spoke (pure three.js) and Hubs (A-Frame), making it a superior choice for content management purposes versus A-Frame’s <a-text> nodes or Mozilla’s hubs-text.js method.

To test:

1. Follow testing directions in the Spoke PR: https://github.com/mozilla/Spoke/pull/1195
2. Pull and checkout `troika-text` branch of Hubs
3. Run `npm install` or `npm ci` to install the new package
4. Open the scene and check text settings have transferred and display correctly

image


src/components/troika-text.js
```diff
@@ -0,0 +1,137 @@
...
+    // opacity: {type: 'number', default: 1.0},
+    // transparent: {default: true},
+    // side: {default: 'front', oneOf: ['front', 'back', 'double']},
+  },
+
+  /**
+   * Called once when component is attached for initial setup.
+   */
+  init: function() {
+    console.log("init1");
```
keianhzo, line 81: Can be removed.

package-lock.json
```diff
@@ -18976,6 +18976,14 @@
         "tryer": "^1.0.0"
       }
     },
+    "bidi-js": {
```
keianhzo, line 18979: I think we only want the troika related packages here. I feel that some things have leaked.

keianhzo (changes requested):
Awesome, thanks again for this. Just a few nits but it looks great.

src/components/scene-components.js
```diff
@@ -29,3 +29,4 @@ import "./video-pause-state";
 import "./particle-emitter";
 import "./audio-zone";
 import "./audio-zone-source";
+import "./troika-text";
```
keianhzo, line 32: The linter complaints about the missing extra blank line at the end of this file.

keianhzo (changes requested):

jameskane05:
@keianhzo I made the two small edits, but as for the package-lock.json update, I think this is correct. `bidi-js` is a dependency of troika-three-text (https://github.com/protectwise/troika/blob/master/packages/troika-three-text/package.json), so wouldn't it make sense it would appear in the package-lock.json after running `npm install`? Let me know what you think.

jameskane05:
`bidi-js` is also present in the Spoke yarn.lock for this PR: https://github.com/mozilla/Spoke/pull/1195/files#diff-51e4f558fae534656963876761c95b83b6ef5da5103c4adef6768219ed76c2de

netpro2k:
Gunna take a stab at seeing if I can merge this with our existing text component so we don't need to maintain 2 different ones...

Issue #2672, opened 2020-07-21
Control users volume from the 2D ui + 🤖 + T (add to user info popup for now)


joshmarinacci:
There is no way to control another user's volume through the 2D UI. Instead you have to hold shift and use a possibly distant UI control inside of VR.  One solution is to add it to the user info popup. This is the popup window that you see when you select another user's name in the user list in the upper right corner. We could add simple volume controls here.

camelgod:
For desktop UI I would really love a volume slider right by the "mute" button or something in the actual dropdown user list instead of having to go manually into each user popup menu. For larger groups where alot of people are speaking I think Second Life Firestorm client does it well with their "Nearby voice" menu, where you can easily adjust volume and mute people in the user list:

![image](https://user-images.githubusercontent.com/28685654/88654572-d514ba80-d0cd-11ea-962e-65ecc9edf52c.png)

emstanley:
Needs discussion on how to integrate into new design.

Issue #2251, opened 2020-03-20
Enable users with screen readers to adjust avatar volume controls


emclaren:
**Is your feature request related to a problem? Please describe.**
At the a11y meetup in Hubs, a user with visual impairments was having trouble hearing a few people speak. There was no way for them to turn up the volume on these avatars to hear them better.

**Describe the solution you'd like**
Perhaps there could be controls accessed through the Room Members Menu to adjust user's volume

TechThomas:
I just want to add as someone that needs to explain this feature verbally numerous times at my meetup it seems like it could be made easier for everyone.  Explaining to everyone to hold down the SPACEBAR and then mouseover content and click on + and - buttons I frequently find it would be difficult for people to follow my directions.  

Sorry I don't have any ideas for a better design :)

PR #4883, opened 2021-11-30
Avatar volume sidebar

keianhzo:
This PR adds a number of things:
- A new UI inside the user profile sidebar to change the volume level or mute locally.
- Persists the updated volume/mute states into a preference
- Reuses `linkedMutedState` audio state storage (now `mutedState`) to handle any audio elements muted states (video and avatars atm)
- Adds a preference audio item to remove the stored levels
- Adds a new slider component that is reused both in the sidebar an in preferences (along with it's storybook story)

Related: https://mozilla-hub.atlassian.net/browse/HUBS-1157


src/react-components/input/Slider.js
```diff
@@ -0,0 +1,143 @@
...
+  );
+}
+
+Slider.propTypes = {
+  min: PropTypes.number,
+  max: PropTypes.number,
+  step: PropTypes.number,
+  digits: PropTypes.number,
+  value: PropTypes.number,
+  defaultValue: PropTypes.number,
```
netpro2k, line 130: Having a `defaultValue` on a widget seems surprising. Its the responsibility of the thing using this component to provide a value to display, this component just tells you when the user tries to drag to change it, and then you pass back in the updated value.

src/react-components/input/Slider.js
```diff
@@ -0,0 +1,143 @@
...
+          ref={inputRef}
+          type="range"
+          step={step}
+          min={min}
+          max={max}
+          value={currentValue}
+          disabled={disabled}
+          onChange={e => {
+            if (disabled) return;
+            const num = round(step, parseFloat(e.target.value));
```
netpro2k, line 114: Doesn't the ranged input take care of this rounding for you because of `step`?

src/react-components/input/Slider.js
```diff
@@ -0,0 +1,143 @@
...
+    v => {
+      setDisplayValue(v);
+      if (onChange) {
+        onChange(v);
+      }
+    },
+    [onChange, setDisplayValue]
+  );
+
+  const drag = useCallback(
```
netpro2k, line 61: Confused why we have to manually implement all this drag logic.. Doesn't the range input handle this for us? Something to do with styling?

src/react-components/input/Slider.js
```diff
@@ -0,0 +1,143 @@
...
+  if (el) {
+    const ratio = ((value - min) / (max - min)).toFixed(2);
+    el.style.backgroundImage = getLinearGradientCSS(ratio, trackColor, "transparent");
+  }
+}
+
+export function Slider({ min, max, step, digits, value, defaultValue, onChange, disabled, className }) {
+  const myRoot = useRef();
+  const inputRef = useRef();
+  const [displayValue, setDisplayValue] = useState(-1);
```
netpro2k, line 34: This extra state existing is surprising. I expect we just use the value passed in directly, which will get updated because we call `onChange` every time the `range` input changes. The usecase I could see for it existing would be if we only wanted to call `onChange` once the user has stopped dragging, but we aren't currently doing that.

src/react-components/input/Slider.js
```diff
@@ -0,0 +1,143 @@
...
+  const [isDragging, setIsDragging] = useState(false);
+
+  let currentValue = defaultValue;
+  if (isDragging) {
+    currentValue = displayValue !== -1 ? displayValue : defaultValue;
+  } else {
+    currentValue = value !== undefined ? value : displayValue !== -1 ? displayValue : defaultValue;
```
netpro2k, line 41: This feels extra defensive. If the proptype is a number and you don't provide it a number that's a bug in the calling component. But also as stated above I think defaultValue is kind of an odd concept which is also part of what makes this line feel odd.

src/react-components/input/Slider.js
```diff
@@ -0,0 +1,143 @@
...
+          type="range"
+          step={step}
+          min={min}
+          max={max}
+          value={currentValue}
+          disabled={disabled}
+          onChange={e => {
+            if (disabled) return;
+            const num = round(step, parseFloat(e.target.value));
+            const finalNum = parseFloat(num.toFixed(digits));
```
netpro2k, line 115: This widget doesn't actually display a number so it feels weird for it to b doing this truncation. Even if it did display a number it seems like that should only be done in the display logic, not in the actual value it outputs. We also seem to do this again outside of the component when using it.

src/react-components/room/useAvatarVolume.js
```diff
@@ -0,0 +1,71 @@
...
+} from "../../utils/avatar-volume-utils";
+
+const minLevel = 0;
+const maxLevel = MAX_VOLUME_LABELS - 1;
+
+export default function useAvatarVolume(sessionId) {
+  const playerInfo = getPlayerInfo(sessionId);
+  const controlsEl = playerInfo?.el.querySelector("[avatar-volume-controls]");
+  const controls = controlsEl?.components["avatar-volume-controls"];
+  const [levelStep, setLevelStep] = useState(0);
```
netpro2k, line 18: The concept of "levels" here seems to have leaked from the fact that I happend to use | characters as a quick hack to display volume in media-video... It feels a bit awkward that using `useAvatarVolume` you end up reading a value in  "levels", but then writing in "gainMultiplier" units. I think everything here should just be based on gain multiplier. The levels thing should only apply at the very edge when you want to display it as a series of bars (which actually seems to be sidestepping this code anyway).

src/react-components/room/useAvatarVolume.js
```diff
@@ -0,0 +1,71 @@
...
+      setLevelStep(newLevel > level ? calcLevelStepDown(gainMultiplier) : calcLevelStepUp(gainMultiplier));
+      setLevel(newLevel);
+    },
+    [level]
+  );
+
+  const updateGainMultiplier = useCallback(
+    value => {
+      if (!controls) return;
+      const gainMultiplier = calcGainMultiplier(value);
```
netpro2k, line 34: Actually looks like what you pass in is indeed in "level" units, so the name is confusing. But then you basically end up having to convert back to gain multiplier, the actual value we want (and then converting back to level in onGainMultiplierUpdated)

src/utils/avatar-volume-utils.js
```diff
@@ -0,0 +1,56 @@
+export const MAX_VOLUME_LABELS = 21;
```
netpro2k, line 1: I think most of the complexity in this file goes away if useAvatarVolume just operates only in gainMultiplier. The one spot that actually cares about level (for displaying the |||| volume bar) can just keep that logic inlined since its purely display logic.

src/react-components/room/UserProfileSidebarContainer.js
```diff
@@ -146,5 +151,12 @@ UserProfileSidebarContainer.propTypes = {
   onBack: PropTypes.func,
   onClose: PropTypes.func,
   onCloseDialog: PropTypes.func.isRequired,
-  showNonHistoriedDialog: PropTypes.func
+  showNonHistoriedDialog: PropTypes.func,
+  minVolume: PropTypes.number,
```
netpro2k, line 155: Confused about these properties. I was at first surprised why they would exist, but then it looks like they arent being used so I am guessing they just got left in by mistake

src/utils/avatar-volume-utils.js
```diff
@@ -0,0 +1,56 @@
...
+export function calcGainMultiplier(level) {
+  return Math.min(
+    MAX_GAIN_MULTIPLIER,
+    level <= MAX_VOLUME_LABELS / 2 ? level * SMALL_GAIN_STEP : 1 + (level - MAX_VOLUME_LABELS / 2) * BIG_GAIN_STEP
+  );
+}
+
+export function updateAvatarVolumesPref(userId, gainMultiplier, muted) {
+  const avatarVoiceLevels = APP.store.state.preferences.avatarVoiceLevels || {};
+  avatarVoiceLevels[userId] = {
```
netpro2k, line 43: I am actually kind of surprised this userId is consistent across sessions. I thought it was random every time you joined a room. If it's not we need to think about what that means privacy wise... But its kind of required to implement a feature like this...

src/components/avatar-volume-controls.js
```diff
@@ -1,63 +1,103 @@
...
+    const numBars = calcLevel(gainMultiplier);
+    this.volumeLabel.setAttribute(
+      "text",
+      "value",
+      gainMultiplier === 0 || this.isLocalMuted() ? "Muted" : VOLUME_LABELS[numBars]
     );
-    this.volumeLabel.setAttribute("text", "value", gainMultiplier === 0 ? "Muted" : VOLUME_LABELS[numBars]);
+  },
+
+  getGainMultiplier() {
```
netpro2k, line 92: Totally just personal style thing but I would not have made either of these getters as I think the line they are wrapping reads just as if not more clearly.

netpro2k (changes requested):
Been waiting for this feature, I use it all the time in Discord! 

I am actually kind of surprised we can get consistent user ids across sessions though, I was under the impression we intentionally make those random for privacy reasons. If not we need to think about that.. If so, this feature will only store the preference until a user refreshes which is not very much different from not persisting them at all. Certainly we want to make features like this so we need to think about the tradeoffs.

Code wise I made some suggestions that I think could simplify things quite a bit. Namely I think the concept of "levels" ended up leaking from a display concern, and I think things could be simplified a bunch only handling that at the edge. Also made some suggestions about how to simplify the Slider component, though I realize most of that code is coppied over from the inlined implementation in preferences screen, so there might be some reason we had to do things manually instead of relying on the browser implementation. That said, a "defaultValue" in a component immediately stood out to me.

src/utils/component-utils.js
```diff
@@ -27,3 +27,14 @@ export function getCurrentStreamer() {
 
   return null;
 }
+
+export function getPlayerInfo(sessionId) {
+  if (!window.APP || !window.APP.componentRegistry || !window.APP.hubChannel || !window.APP.hubChannel.presence)
```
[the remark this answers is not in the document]
keianhzo: I think I got this from:

https://github.com/mozilla/hubs/blob/749c7f4c0b4bd86aa40f74d478531078ceb54223/src/utils/component-utils.js#L14

Not sure why we need that there  but I kept it. I guess in this case we should never call this method if there is not a hub channel so probably it can be removed.

src/react-components/input/Slider.js
keianhzo: I think I just copy pasted this from the prefs code and didn't really questioned it that much. Looking at it now I think we can get rid of a bunch of code. I'll test it in different platforms, it might have been there for a reason...

src/react-components/input/Slider.js
keianhzo: True story. Also copied from the pref implementation. Updated.

src/react-components/input/Slider.js
keianhzo: All this state handling was original to deal with the double updating case where the slider itself could update it's value and also be updated by an external component from outside. As part of the refactor I've got rid of this an a lot of related code. Now the component should be much more lean and refresh only when strictly necessary.

src/react-components/input/Slider.js
keianhzo: Same as above.

src/react-components/input/Slider.js
keianhzo: same

src/react-components/room/UserProfileSidebarContainer.js
keianhzo: Yep, I think I left that from an initial implementation. Removed.

src/utils/avatar-volume-utils.js
keianhzo: That name leaked from an initial implementation also, the right name for that parameter is be `displayName` as we don't really have access to any `userId` so we use the display name to store volumes. Updated.

src/components/avatar-volume-controls.js
keianhzo: Old habits. Updated

src/react-components/room/useAvatarVolume.js
keianhzo: Yep, it makes sense. I've refactored all the level calculations outside `useAvatarVolume`.

src/react-components/room/useAvatarVolume.js
keianhzo: updated

src/utils/avatar-volume-utils.js
keianhzo: updated

keianhzo:
@netpro2k I've updated quite a few things.

src/utils/avatar-volume-utils.js
netpro2k: Hmm, displaynames are not guaranteed to be unique so that could pose a bit of an issue... I guess its probably the best we can do without making a decision on if we want to expose unique identifiers... Should be fairly easy to migrate later I guess. We should discuss this at the next engineering meeting.

src/react-components/room/UserProfileSidebar.js
```diff
@@ -38,7 +48,25 @@ export function UserProfileSidebar({
...
+    onMultiplierChanged
+  );
+  const onLevelChanged = useCallback(
+    level => {
+      updateMultiplier(calcGainMultiplier(level));
+    },
+    [updateMultiplier]
+  );
+  const newLevel = calcLevel(multiplier);
+  const step = multiplier > prevMultiplier ? calcLevelStepDown(newLevel) : calcLevelStepUp(newLevel);
```
netpro2k, line 69: Not really following why this step needs to be dynamic? Tried using the slider too and I can't really tell when its using one or the other. Is this a specific requirement from the designs or something?

src/react-components/room/UserProfileSidebar.js
```diff
@@ -61,13 +89,11 @@ export function UserProfileSidebar({
               disabled={isNetworkMuted}
             />
             <Slider
```
netpro2k, line 91: This sidebar component still seems fishy. The fact that its not getting any sort of value passed into it is a red flag (its storing state, which is being updated imperatively). Since its primary goal is essentially just restyling a native slider I would expect it to work almost exactly the same. I would expect something like this: 
```js
<Slider
      min={0}
      max={2}
      step={0.1}
      value={multiplier}
      onChange={updateMultiplier}
      className={styles.sliderInputContainer}
      disabled={isNetworkMuted || isMuted}
/>
```

src/react-components/input/Slider.js
```diff
@@ -28,105 +24,56 @@ function updateGradient(el, min, max, value) {
   }
 }
 
-export function Slider({ min, max, step, digits, value, defaultValue, onChange, disabled, className }) {
-  const myRoot = useRef();
-  const inputRef = useRef();
-  const [displayValue, setDisplayValue] = useState(-1);
-  const [isDragging, setIsDragging] = useState(false);
+export const Slider = memo(
+  forwardRef(({ min, max, step, defaultValue, onChange, disabled, className, ...rest }, ref) => {
```
netpro2k, line 28: Never really used forwardRef, seems like its being used correctly here but I am still not sure this component actually needs any state. I would expect it just gets passed in a `value` (and not a `defaultValue`) and then the thing using the component is responsolbe for passing in an updated `value` in response to `onChange`

src/react-components/input/Slider.js
```diff
@@ -28,105 +24,56 @@ function updateGradient(el, min, max, value) {
...
+          <input
+            ref={inputRef}
+            type="range"
+            step={step}
+            min={min}
+            max={max}
+            value={displayValue}
+            disabled={disabled}
+            onChange={e => {
+              if (disabled) return;
```
netpro2k, line 58: Don't think its possible for `onChange` to be fired from a disabled `input`

src/react-components/input/Slider.js
netpro2k: This still seems true (though is really just a carry through from this component having state being itself surprising). The main case I could see for a component like this having a default value would be if it internally handled "resetting" to some default, but even then I would probably just have it take an `onReset` function and have the thing using the component handle actually changing the value.

src/react-components/preferences-screen.js
```diff
@@ -154,18 +153,20 @@ export class NumberRangeSelector extends Component {
...
             }}
           />
         </div>
         <Slider
+          ref={this.sliderRef}
           step={this.props.step}
           min={this.props.min}
           max={this.props.max}
-          value={currentValue}
+          defaultValue={currentValue}
```
netpro2k, line 165: This is again quite suspicious, and misleading, since the value of this prop is only used when the component is first mounted (to set the initial state), so changing it has no effect. It works anyway since the ref is actually what is used to update the internal state (imperatively)

src/react-components/room/useAvatarVolume.js
```diff
@@ -1,40 +1,31 @@
...
-const maxLevel = MAX_VOLUME_LABELS - 1;
-
-export default function useAvatarVolume(sessionId) {
+export default function useAvatarVolume(sessionId, onMultiplierChanged) {
   const playerInfo = getPlayerInfo(sessionId);
   const controlsEl = playerInfo?.el.querySelector("[avatar-volume-controls]");
   const controls = controlsEl?.components["avatar-volume-controls"];
-  const [levelStep, setLevelStep] = useState(0);
-  const [level, setLevel] = useState(0);
+  const [prevMultiplier, setPrevMultiplier] = useState(0);
```
netpro2k, line 8: This is currently only being used to set the dynamic step size. If that actually turns out not to be a requirement this can go as well, further simplifying things.

netpro2k (commented):
This is looking closer. I think the Slider component still feels a bit more complex than necessary, made some suggestions to hopefully further simplify. Don/t want to keep dragging this PR out if you just want to be done with it heh.

src/react-components/input/Slider.stories.js
```diff
@@ -13,10 +13,10 @@ const Template = args => (
   </Column>
 );
 
-export const Base = Template.bind({});
-Base.argTypes = {
+export const Controlled = Template.bind({});
```
netpro2k, line 16: I have not really used storybook much, whats this change about?

netpro2k (commented):
Nice, this looks much cleaner now.

src/utils/component-utils.js
```diff
@@ -29,12 +29,10 @@ export function getCurrentStreamer() {
 }
 
 export function getPlayerInfo(sessionId) {
-  if (!window.APP || !window.APP.componentRegistry || !window.APP.hubChannel || !window.APP.hubChannel.presence)
-    return null;
   const playerInfos = window.APP.componentRegistry["player-info"] || [];
-  const playerInfo = playerInfos.filter(info => {
+  const playerInfo = playerInfos.find(info => {
```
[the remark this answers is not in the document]
keianhzo: Updated

src/react-components/input/Slider.js
keianhzo: Removed

src/react-components/input/Slider.stories.js
keianhzo: After learning about the Controlled/Uncontrolled ReactJS components I updated the slider story to show the slider component working as a controller and an uncontrolled component.

camelgod:
Related issues:
https://github.com/mozilla/hubs/issues/2672
https://github.com/mozilla/hubs/issues/2251

netpro2k (commented):
Changes to the Slider look good. I am still a bit confused why the "level" stuff is needed and we don't just directly operate on gain multiplier but doesn't seem like a huge deal either way.

Issue #4916, opened 2021-12-10
focus-action icon unaffected by notice-background-color variable


markusTraber:
**Description**
The focus-action.png is a notice icon, but its color is #FF3464 instead of #2F80ED. The result is, that the icon won't be affected by the notice-background-color variable and changed via color-shift.worker.js.

**To Reproduce**
1. Change the 'notice-background-color' variable in your theme.
2. Go into a room.
3. Insert for a example a picture in the room.
4. Toggle the object menu and see if the color of the focus-action icon changed.

**Expected behavior**
The focus-action icon color should be affected by notice-background-color and therefore change its color if set to a different color value.

**Screenshots**
![grafik](https://user-images.githubusercontent.com/12909076/145557802-b0ca0845-f8b3-474f-9401-1b9b4112e3b1.png)

**Hardware** 
 - Device: Desktop
 - OS: MacOs
 - Browser: Firefox

**Additional context**
I fixed the problem by changing the color of the focus-action.png from #FF3464 to #2F80ED. A pull request will follow soon.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-1203)

PR #4923, opened 2021-12-10
Changed focus-action icon color from #FF3464 to #2F80ED for it to wor…

markusTraber:
I reported and described this issue here: #4916

Changed focus-action icon color from #FF3464 to #2F80ED (notice color) for it to work with color-shift.
But I am currently unsure, if this is the best solution. Maybe this icon should be in the action-spritesheet instead?


keianhzo:
Hey @markusTraber Thanks for this, it seems that this change would require to change the icon color also to update the spritesheets and ideally also moving this sprite to the actions spritesheet that seems the most logical place. I've pushed a PR with this changes: https://github.com/mozilla/hubs/pull/4965 Feel free to review and close this one if you feel that it fixes the issue.

Issue #4918, opened 2021-12-10
getThemeColor() not returning default theme colors, if localStorage has no theme id defined.


markusTraber:
**Description**
The function getThemeColor() in theme.js not returning colors of the defined theme, if localStorage has no theme id defined. 

**To Reproduce**
1. Make sure your localStorage does not contain a theme id.
2. Go to the admin interface and change the theme color for "action-color".
3. Go into a room.
4. Insert for a example a picture in the room.
5. Toggle the object menu and see if the colors of the buttons have changed.

**Expected behavior**
The UI-Element colors of the object menu should change to the defined theme color.

**Screenshots**
I defined in the admin interface "action-color": "#D5AD50" (curry-yellow), but the UI elements stay pink, like in the following screenshot.
![grafik](https://user-images.githubusercontent.com/12909076/145564302-60bf5ed1-bfb8-471a-b400-138875a014c3.png)

**Hardware** 
 - Device: Desktop
 - OS: MacOs
 - Browser: Firefox

**Additional context**
I fixed the issue for myself, a pull request will follow soon.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-1205)

markusTraber:
Pull request has been merged and this issue is therefor resolved.

PR #4924, opened 2021-12-10
getThemeColor() fix if localStorage has no theme id defined.

markusTraber:
Described the issue here: #4918

The changes achieve, that the getThemeColor() function looks for a default theme, if no preferences are defined in the localStorage.


matthewbcool (approved):
Looks good to me, thank you for your contribution.

Issue #4646, opened 2021-09-16
The people counter is not displayed if the language is set to french


Softvision-GeluHaiduc:
**[Affected Versions]:**
- Firefox Release 92.0
- Chrome 93.0.4577.82

**[Affected Platforms]:**
- Windows 10x64

**[Prerequisites]:**
- Have a new room created.
- Have the room entered.

**[Steps to reproduce]:**
1. Open Preferences from the More menu on the "Misc" page.
2. Click the drop-down menu from the Language section.
3. Select the French language from the dropdown menu.
4. Dismiss the “Preferences” menu
5. Observe the “Personnes” button in the top right corner of the main view

**[Expected results]:**
- The people counter is displayed to the right of the “Personnes” string

**[Actual results]:**
- The people counter is not displayed

**[Notes]:**
- Attached is a screenshot of the issue below:
![image](https://user-images.githubusercontent.com/51400741/133620426-34d4d30c-734d-48e4-8244-692d8503b8d3.png)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-909)

Issue #4386, opened 2021-06-30
[French] The More menu is slightly enlarged when hovering over the "Unfavorite Room" button


Softvision-RemusDranca:
**[Affected Versions]:**
- Safari 14
- Chrome 91
- Firefox 89.0.2

**[Affected Platforms]:**
- MacOS 10.15.7
- Windows 10

**[Prerequisites]:**
- Hubs Cloud services are set up.
- Have a favorite room opened.
- Set the room's language to French.

**[Steps to reproduce]:**
1. Click the "More" menu.
2. Hover over the "Unfavorite Room" button.
3. Observe the behavior.

**[Expected results]:**
- The More menu's size isn't affected.

**[Actual results]:**
- The More menu is slightly enlarged.

**[Notes]:**
- The issue is also reproducible on "https://hubs.mozilla.com/" website.
- Attached is a screen recording of the issue: 
![FrenchRemoveFavorite](https://user-images.githubusercontent.com/40061643/123974607-66b73700-d9c5-11eb-835d-c3c968eec157.gif)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-83)

SV-AndreiC:
"I have verified this issue and it's no longer reproducible. Now, the name of the button is changed and the More menu is no longer enlarged when Hovering over the Unfavorite Room button.
Tested on: 
- Windows 10 using Firefox 94.0.2 and Chrome 96.0.4664.45.
- macOS BigSur v11.6.1 using Safari v15.1."

matthewbcool:
Reopening, some disagreement over appropriate translation among French speakers. Issue will need to be addressed as part of a larger issue of developing a system for dealing with longer strings elegantly in the UI.

PR #4936, opened 2021-12-14
updates french translations based on feedback and fixes people counter

matthewbcool:
Fixes the people counter for French (resolves https://github.com/mozilla/hubs/issues/4646) and also reverts some translations after we had some disagreement among french speakers over the most appropriate translation.  All three agree that the original is good so we will take that consensus over the change. Re-opens https://github.com/mozilla/hubs/issues/4386


brianpeiris (approved):
LGTM

Issue #4897, opened 2021-12-06
Scene link preview screen bugs 


matthewbcool:
**Description**
Certain scene previews will fail to load in the scene-ui screen. (scene links)
Additionally there is a broken attribution in the lower right and attributions can display multiple 'by's and incorrect spacing. The text doesn't wrap or respond well on smaller mobile devices.

**To Reproduce**
Go to a scene link such as https://hubs.mozilla.com/scenes/Rpt8DJS/outdoor-festival

**Expected behavior**
We should have the text be legible on smaller devices, attributions should not have repeat prepositions or missing spaces in the text. The scene preview should work as expected and if not fallback to something more aesthetically pleasing.  

**Screenshots**

<img width="1055" alt="Screen Shot 2021-12-06 at 12 48 23 PM" src="https://user-images.githubusercontent.com/4493657/144921553-b16612fd-ce5d-4b57-87b0-05a80e21336c.png">

**Hardware** 
 - Device: desktop, mobile
 - OS: Mac, android
 - Browser: firefox 94, chrome 96, 

**Additional context**

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-1187)

matthewbcool:
The issue remains mostly for spoke and potentially the blender exporter to potentially validate attribution inputs (e.g: By: By) or maybe this is better left to the discretion of the designer/builder of the scene.

PR #4967, opened 2021-12-22
scene link screen fixes

matthewbcool:
This PR addresses several issues with the scene screen, a screen that serves as a common entry point for many hubs users. Detailed notes can be found below.  [Design Link (requires access)](https://www.figma.com/file/J6YKkb2vWDlk81zm4kxUzA/Hubs-discoverability-and-onboarding?node-id=903%3A1324)

This PR does not include every design recommendation but resolves the issues outlined in #4897. The attribution issue mentioned in #4897  should be caught on the publishing side and is not addressed here.

| Mobile Scene Screen Before    | Mobile Scene Screen After  |
| ----------- | ----------- |
|s5 before|s5 after|

Notes:

- removed the spoke call out
    the lower right hand spoke call out has been removed from the [design](https://www.figma.com/file/J6YKkb2vWDlk81zm4kxUzA/Hubs-discoverability-and-onboarding?node-id=903%3A1324)

- use theme values for text color and shadow color
     note on shadow color: AFAIK we don't have an explicit inverted text color and that is effectively what we want. Adding 
     a text-inverted-color theme to accommodate a variety of themes and I removed text6 as we don't seem to be using 
     that variable anywhere. An alternative to consider might be to use filter() on the existing dominate text color.   
          
- removed white overlay on scene image

-  [clamped](https://caniuse.com/?search=clamp()) the width for logo
    these values look reasonable to me but feedback welcome  

- changed grid values and scene info to inside the grid for more control over it's layout.
     this piece was previously positioned absolutely, added a media query to changed alignment for mobile

- set a media query on the grid to lower the top row height when we are below the vr height breakpoint

- adjusted padding for tagline 

- removed now irrelevant positioning CSS on values that were previously positioned absolutely

### DIFF
--- a/src/react-components/scene-ui.js
+++ b/src/react-components/scene-ui.js
@@ -212,7 +212,6 @@ class SceneUI extends Component {
         >
           {this.state.showScreenshot && <img src={this.props.sceneScreenshotURL} />}
         </div>
-        <div className={styles.whiteOverlay} />
         <div className={styles.grid}>
           <div className={styles.mainPanel}>
             <a href="/" className={styles.logo}>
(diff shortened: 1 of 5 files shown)


src/react-components/scene-ui.js
```diff
@@ -273,28 +272,11 @@ class SceneUI extends Component {
               </a>
             </div>
           </div>
-        </div>
-        <div className={styles.info}>
-          <div className={styles.name}>{this.props.sceneName}</div>
-          <div className={styles.attribution}>{attributions}</div>
-        </div>
-        <IfFeature name="enable_spoke">
```
netpro2k, line 281: Wasn't able to check the design notes so I assume its covered in there, but just in case, just making sure this was intentionally removed.

netpro2k (commented):
lgtm

src/react-components/scene-ui.js
matthewbcool: Correct, spoke call out is intentionally removed. Thanks for checking.

PR #4977, opened 2021-12-29
target svg on hmc and use theme color on object attributions

matthewbcool:
- Use theme colors on object attributions and adjust text alignment in the info section.
- clamps the height and width specifically for the svg we use on hmc.
- removed a 'wrap' value from `white-space` since it is not an accepted keyword value of this property.

### DIFF
--- a/src/assets/stylesheets/scene-ui.scss
+++ b/src/assets/stylesheets/scene-ui.scss
@@ -54,8 +54,8 @@
     width: 100%;
   }
   svg {
-    width: 100%;
-    min-height: 210px;
+    width: clamp(200px, 100%, 17vw);
+    height: clamp(200px, 100%, 17vh);
   }
 }
 
@@ -67,10 +67,12 @@
   flex-direction: column;
   align-items: center;
   align-self: center;
+  text-align: center;
   @media (min-width: theme.$breakpoint-lg) {
     align-items: flex-start;
     align-self: left;
     margin-left: 16px;
+    text-align: left;
   }
 }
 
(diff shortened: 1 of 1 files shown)


src/assets/stylesheets/scene-ui.scss
```diff
@@ -81,15 +83,15 @@
...
+  white-space: normal;
 
   .remix {
     font-size: 0.8em;
   }
 
   a {
     font-size: 0.8em;
-    color: black;
+    color: theme.$text1-color;
```
keianhzo, line 94: Not sure about the target style for this page but we have moved from a dark font to a light one. I find it harder to read the text.

keianhzo (approved):
LGTM I just have a question about the font color change.

src/assets/stylesheets/scene-ui.scss
matthewbcool: Especially true against light-colored scene previews for dark and darker previews for light, also it's not very clear to me that this is a link to the glb. Will make notes here: [Design Link](https://www.figma.com/file/J6YKkb2vWDlk81zm4kxUzA/Hubs-discoverability-and-onboarding?node-id=903%3A1324)

https://user-images.githubusercontent.com/4493657/147691074-dd975838-39f5-4e72-867e-6d1effc8ed23.mp4

keianhzo: Thanks for the Figma link! that's what I was looking for. Makes sense.

Issue #4922, opened 2021-12-10
There is no recording spawned when using the record function of the “Camera”


SV-AndreiC:
**[Affected Versions]:**
- Firefox Release 92.01
- Chrome 94.0.4606.61
- Safari 15.1

**[Affected Platforms]:**
- Windows 10x64
- iOS15.1

**[Prerequisites]:**
- The Hubs Cloud stack is set up. 
- Have a room created and opened in Hubs Cloud.

**[Steps to reproduce]:**
1. Tap on the "Camera" button from the Place menu.
2. Tap on the "Video" button.
3. Wait for the recording to finish and observe the outcome.

**[Expected results]:**
- A video object containing the recording spawns in the room.

**[Actual results]:**
- There is no object containing the recording, spawned in the room.

**[Notes]:**
- There is no notification or error message received when the recording finishes.
- If there’s another user in the room a “Broken media link” can be seen spawning in the room. 
- Link to a screen recording of the issue:
https://drive.google.com/file/d/16v_-bKjd4gyTTVWkXbfbRAELmlhyRB_w/view?usp=sharing

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-1206)

brianpeiris:
Closing this as a duplicate of #4895, since they have the same root cause.

brianpeiris:
Actually, I'll keep it open so the QA team can verify it themselves.

brianpeiris:
The SpeelyCaptor and the CloudFormation template updates were deployed, and this is currently working on Hubs Cloud, so we can call this closed.

Issue #3742, opened 2021-01-21
[Intermittent] The recorded videos are sometimes broken, incomplete or not recorded 


Softvision-RemusDranca:
**[Affected Versions]:**
- Chrome 87.0.4280.141
- Firefox Release 84.0.2

**[Affected Platforms]:**
- Windows 10x64

**[Prerequisites]:**
- Hubs Cloud is set up.
- Have a room opened.

**[Steps to reproduce]:**
1. Click the "Camera" button from the main menu.
2. Click the "Recording" button from the camera.
3. Wait until the recording process is complete.
4. Observe the behavior.

**[Expected results]:**
- The video is successfully recorded.

**[Actual results]:**
- The videos are sometimes broken, incomplete or not recorded.

**[Notes]:**
- Attached are recordings with the issue:
	- Chrome video: [link](https://drive.google.com/file/d/1v-WAyOvAXHu0Z2MCo2ED_sMeG4oUiagR/view?usp=sharing).
	- Firefox video - [link](https://drive.google.com/file/d/1rQgw5758x5_izy5QpPPx4s7uuOTNBjJ1/view?usp=sharing).
	- The only error I get in the console is when the video is not recorded, here is a screenshot:
	
![ErrorVideo](https://user-images.githubusercontent.com/40061643/105331291-f4535700-5bdb-11eb-991a-f63cc754ad35.PNG)

Softvision-RemusDranca:
It seems that this issue is also reproducible on Oculus Browser using Oculus Quest.

Softvision-GeluHaiduc:
I also encountered this issue on hubs.mozilla.com using Chrome 90.0.4430.212 on Windows 10 with the new UI.

PR #4987, opened 2021-12-31
fix camera-tool video recordings broken or shortened

yakyouk:
Camera-tool video recordings sometimes generate zero-length video or, for longer recording times, videos miss a big chunk at the end.

This issue seems to be due to MediaRecorder events `ondataavailable` and `onstop` firing out of order, causing the last data chunk to be discarded.

This PR solves the issue by moving the video processing logic into `ondataavailable`, processing data only after the last chunk is received.

Fixing #4922 and #3742

### DIFF
--- a/src/components/camera-tool.js
+++ b/src/components/camera-tool.js
@@ -416,35 +416,15 @@ AFRAME.registerComponent("camera-tool", {
     const chunks = [];
     const recordingStartTime = performance.now();
 
-    this.videoRecorder.ondataavailable = e => chunks.push(e.data);
+    this.videoRecorder.ondataavailable = e => {
+      chunks.push(e.data);
+      if (this.videoRecorder) return;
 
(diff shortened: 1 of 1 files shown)


keianhzo:
Thanks for fixing this @yakyouk!

keianhzo:
Hey @yakyouk I was testing this and I can still get partial videos. During my testing I found out a related issue that I've fixed in another PR (#5000) and I couldn't seem to reproduce this issue using that branch anymore so I wonder if this could be somehow related to that fix. Could you please give it a try and see if you can reproduce it using that PR?

keianhzo:
Hi @yakyouk [The PR I talked about ](https://github.com/mozilla/hubs/pull/5000) has landed. Could you check if this is still happening in production and close if not?

keianhzo:
@yakyouk I'm going to close this for the reasons stated above, feel free to reopen if you see this behavior again

PR #4997, opened 2022-01-06
Handle twitter misconfiguration error

brianpeiris:
Adds error handling to twitter API calls, so that we react accordingly instead of crashing the game loop.

Goes with https://github.com/mozilla/reticulum/pull/543

### DIFF
--- a/src/react-components/room/TwitterOAuthModalContainer.js
+++ b/src/react-components/room/TwitterOAuthModalContainer.js
@@ -3,7 +3,7 @@ import PropTypes from "prop-types";
 import { TwitterOAuthModal } from "./TwitterOAuthModal";
 import configs from "../../utils/configs";
 
-export function TwitterOAuthModalContainer({ hubChannel, onConnected, onClose }) {
+export function TwitterOAuthModalContainer({ hubChannel, onConnected, onClose, isAdmin }) {
   const popupRef = useRef();
 
   const onConnect = useCallback(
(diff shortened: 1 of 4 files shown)


src/react-components/room/TwitterOAuthModalContainer.js
```diff
@@ -30,7 +30,14 @@ export function TwitterOAuthModalContainer({ hubChannel, onConnected, onClose })
         popup.focus();
         popupRef.current = popup;
       } catch (error) {
-        console.error(error);
+        if (error === "twitter_probably_misconfigured") {
```
brianpeiris, line 33: This seems wrong now that I look at it. `error` should probably be an `Error` object, not a string.

src/react-components/room/TwitterOAuthModalContainer.js
```diff
@@ -30,7 +30,14 @@ export function TwitterOAuthModalContainer({ hubChannel, onConnected, onClose })
...
+          console.error(
+            "Twitter might be misconfigured. Check the docs here: https://hubs.mozilla.com/docs/hubs-cloud-enable-media-browser.html#twitter"
```
brianpeiris, line 35: We might want to fix this URL in the docs before we merge this?

src/utils/hub-channel.js
```diff
@@ -360,7 +360,7 @@ export default class HubChannel extends EventTarget {
         .receive("ok", res => {
           resolve(res.oauth_url);
         })
-        .receive("error", reject);
+        .receive("error", err => reject(err.reason));
```
robinkwilson, line 363: Should we pass in the error object?

src/react-components/room/TwitterOAuthModalContainer.js
robinkwilson: I can change the way we handle the error if this still looks wrong.

PR #5000, opened 2022-01-07
Fix chrome camera recording issues with muted audio tracks

keianhzo:
Browsers seem to have issues recording videos that have empty or muted audio tracks. We had a hack in place for both FF and Chrome but it seems to have stopped working for Chrome at some point

This PR adds a Chrome specific hack that removes the audio tracks when there is none available as with the previous hack we were missing data chunks and ended up generating partial or empty videos.

To test the Chrome issues before this PR:
- Open the Camera Tool
- Change the video recording time to 60
- Mute your mic
- Start a video recording

The resulting video is empty or partial.

Related: https://github.com/mozilla/hubs/pull/4987

### DIFF
--- a/src/components/camera-tool.js
+++ b/src/components/camera-tool.js
@@ -10,6 +10,11 @@ import { waitForDOMContentLoaded } from "../utils/async-utils";
 import cameraModelSrc from "../assets/camera_tool.glb";
 import anime from "animejs";
 import { Layers } from "./layers";
+const { detect } = require("detect-browser");
+
+const browser = detect();
+
+const isFirefox = browser.name === "firefox";
 
 const cameraModelPromise = waitForDOMContentLoaded().then(() => loadModel(cameraModelSrc));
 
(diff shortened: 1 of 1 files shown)


src/components/camera-tool.js
```diff
@@ -372,17 +377,23 @@ AFRAME.registerComponent("camera-tool", {
...
+        const gain = context.createGain();
+        const destination = context.createMediaStreamDestination();
+        gain.gain.setValueAtTime(0.0001, context.currentTime);
+        oscillator.connect(destination);
+        gain.connect(destination);
+        stream.addTrack(destination.stream.getAudioTracks()[0]);
+      } else {
+        // Chrome has issues when the audio tracks are muted.
+        // https://bugs.chromium.org/p/chromium/issues/detail?id=1223382
+        stream.getAudioTracks().forEach(t => stream.removeTrack(t));
```
netpro2k, line 395: Shouldn't the video have no audio tracks already with the above code in a conditional for Firefox?

src/components/camera-tool.js
```diff
@@ -405,18 +416,21 @@ AFRAME.registerComponent("camera-tool", {
...
     }
 
     stream.addTrack(track);
     this.videoRecorder = new MediaRecorder(stream, { mimeType: videoMimeType });
     const chunks = [];
     const recordingStartTime = performance.now();
 
-    this.videoRecorder.ondataavailable = e => chunks.push(e.data);
+    this.videoRecorder.ondataavailable = e => {
+      console.log("Camera Tool: ondataavailable => " + e.data?.size);
```
netpro2k, line 431: remove console.log
```suggestion
```

netpro2k (commented):
This LGTM. I am guessing the chrome branch of the if that removes audio tracks is now a noop, but I didn't dig into it.

src/components/camera-tool.js
keianhzo: Removed

PR #5039, opened 2022-01-19
Reflection Probes

netpro2k:
You can now have multiple environment maps in your scenes using the new component `reflection-probe` component. Objects inside of the reflection probe's box will use the environment map configured for that probe rather than the one configured in environment-settings. When an object overlaps multiple probes, the object will use a mix of the 2 most overlapped probes' maps. This feature should be considered alpha and will likely be undergoing changes based on feedback.

image

This is a first pass at this, and there will likely be more changes/features (possibly breaking) added, but the best way to find out what those changes/features are is to have people start trying to build actual scenes with it. I plan to limit the blast radius on breaking changes by not immediately merging the blender-exporter changes for this, to limit people using it ot people we have a good direct line of communication with.

before merging:
- [x] merge MozillaReality/three.js/pull/59
- [x] rebase and update build on `three-patches-133` branch in three repo
- [x] update package.json in this repo to point back at `three-patches-133`
- [x] update package-lock.json to point at the updated sha at the head of `three-patches-133`

### DIFF
--- a/src/systems/environment-system.js
+++ b/src/systems/environment-system.js
@@ -98,6 +98,13 @@ export class EnvironmentSystem {
       Object.assign(envSettings, envSettingsEl.components["environment-settings"].data);
     }
 
+    // TODO animated objects should not be static
+    envEl.object3D.traverse(o => {
+      if (o.isMesh) {
+        o.reflectionProbeMode = "static";
+      }
+    });
+
     this.applyEnvSettings(envSettings);
   }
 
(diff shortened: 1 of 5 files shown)


src/react-components/room/RoomSidebar.js
```diff
@@ -135,7 +136,59 @@ export function SceneInfo({ accountId, scene, showAttributions, canChangeScene,
...
+    setDropping(false);
+  };
+
+  const handleDragOver = e => {
+    e.preventDefault();
+    e.dataTransfer.dropEffect = "move";
+  };
+
+  const handleDrop = e => {
+    e.preventDefault();
```
keianhzo, line 158: Not critical but it's confusing that the object dropping stops working during scene debug. I think we can stop the event bubbling propagation here using `e.stopPropagation()` and we can keep the top object drop behavior untouched.

keianhzo (approved):
Looks good to me, just a nit.

src/systems/environment-system.js
```diff
@@ -194,3 +201,24 @@ AFRAME.registerComponent("environment-settings", {
...
+  init: function() {
+    this.el.object3D.updateMatrices();
+
+    const box = new THREE.Box3()
+      .setFromCenterAndSize(new THREE.Vector3(), new THREE.Vector3().setScalar(this.data.size * 2))
+      .applyMatrix4(this.el.object3D.matrixWorld);
+
+    this.el.setObject3D("probe", new THREE.ReflectionProbe(box, this.data.envMapTexture));
+
+    this.el.sceneEl.object3D.add(
```
keianhzo, line 220: Not sure if this is important (as this is mostly for testing purposes) but we are showing the helper even in non scene debug mode and we are not cleaning the helper geo when the component is remove.

src/systems/environment-system.js
```diff
@@ -194,3 +201,24 @@ AFRAME.registerComponent("environment-settings", {
...
+      new THREE.Box3Helper(box, new THREE.Color(Math.random(), Math.random(), Math.random()))
+    );
```
keianhzo, line 222: Not sure if this is important (as this is mostly for testing purposes) but we are showing the helper even in non scene debug mode and we are not cleaning the helper geo when the component is remove.

keianhzo (approved):

src/react-components/room/RoomSidebar.js
netpro2k: Yeah removing this from this PR, I ended up doing it in a different PR

src/systems/environment-system.js
netpro2k: fixed to only show when evnSettingsDebug is enabled

Issue #4173, opened 2021-04-21
Can't link to Hubs rooms using named target


rawnsley:
If you have a link that looks like this:

```
<a href="https://hubs.mozilla.com/uipqcCJ/buttery-energetic-shindig" target="sometarget">
    Click to open in a new browser window
</a>
```
The Hubs room load can fail with the following error message in console:

`Uncaught DOMException: Blocked a frame with origin "https://hubs.mozilla.com" from accessing a cross-origin frame.
`

The offender [is some code](https://github.com/mozilla/hubs/blob/master/src/hub.js#L235) that's part of the OAuth project and probably temporary:

```
// OAuth popup handler
// TODO: Replace with a new oauth callback route that has this postMessage script.
if (window.opener && window.opener.doingTwitterOAuth) {
  window.opener.postMessage("oauth-successful");
  isOAuthModal = true;
}
```

Basically you are not supposed to reference `window.opener` unless you have CORS permission from the parent server.

Kind of obscure, but I happen to link to Hubs in exactly this way. [Here is an example page](https://alpha.avncloud.com/dev/cross-window-hubs-bug.html) with the two types of links, the first will work and the second will not.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-627)

oneWaveAdrian:
Experiencing the same issue. Only happens when linking with `target="_blank"` otherwise works fine. I will use intermediate solution as suggested by poster

matthewbcool:
I wasn't able to repro using `<a href={hubsroomurl} target="_blank">`

brianpeiris:
Just wanted to note something I realized about this bug -- when linking to Hubs rooms, users may not always have control of the particular HTML used in the link tag, for example when using a third-party site or CMS, so the `noopener` workaround may not always be usable. That bumps the priority of this issue, in my mind.

PR #5040, opened 2022-01-19
Catch DOM security exception when accessing properites of window.opener

rawnsley:
Addresses https://github.com/mozilla/hubs/issues/4173

In the unusual case that a window is opened from an anchor as a named target, accessing properties on the `window.opener` object will throw a security exception if they are cross-origin, which is often going to be the case.

There may be more elegant ways to address this, but the code being wrapped is likely to be replaced once OAuth is fully implemented anyway.

### DIFF
--- a/src/hub.js
+++ b/src/hub.js
@@ -264,10 +264,14 @@ let isOAuthModal = false;
 
 // OAuth popup handler
 // TODO: Replace with a new oauth callback route that has this postMessage script.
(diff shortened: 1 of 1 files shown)


netpro2k (approved):
lgtm. We need to ditch this code, but until we do this fixes an issue people keep running into... Thanks for the fix!

PR #5045, opened 2022-01-20
Audio zones transform fix and new shapes

keianhzo:


keianhzo:
@Charlesc22 This PR and the related https://github.com/mozilla/Spoke/pull/1208 should fix the issues that we have talked about. Feel free to review it.

src/components/audio-zone.js
```diff
@@ -42,17 +49,35 @@ AFRAME.registerComponent("audio-zone", {
...
     return function() {
+      this.debugMesh.geometry.computeBoundingBox();
       return bbaa.copy(this.debugMesh.geometry.boundingBox).applyMatrix4(this.el.object3D.matrixWorld);
     };
   })(),
 
+  getBoundingSphere: (function() {
+    const bbaa = new THREE.Sphere();
+    return function() {
+      this.debugMesh.geometry.computeBoundingSphere();
```
takahirox, line 76: Better to add a guard for efficiency unless this method is guaranteed to be called only once or the bounding sphere really needs to be recomputed for each call?

```
if (this.debugMesh.geometry.boundingSphere !== null) {
  this.debugMesh.geometry.computeBoundingSphere();
}
```

src/components/audio-zone.js
```diff
@@ -42,17 +49,35 @@ AFRAME.registerComponent("audio-zone", {
...
+  })(),
+
   isEnabled() {
     return this.data.enabled;
   },
 
   getBoundingBox: (function() {
     const bbaa = new THREE.Box3();
     return function() {
+      this.debugMesh.geometry.computeBoundingBox();
```
takahirox, line 68: Similar to `computeBoundingSphere()`, better to consider to add a guard?

src/components/audio-zone.js
```diff
@@ -42,17 +49,35 @@ AFRAME.registerComponent("audio-zone", {
     this.enableDebug(this.data.debuggable && window.APP.store.state.preferences.showAudioDebugPanel);
   },
 
+  tick: (() => {
+    const quat = new THREE.Quaternion();
+    return function() {
+      this.el.object3D.quaternion.set(0, 0, 0, 1);
+      this.el.object3D.quaternion.multiply(this.el.object3D.parent.getWorldQuaternion(quat).inverse());
```
takahirox, line 56: I'm not so familiar with math, but I guess 

```
quaternion.set(0,0,0,1)
quaternion.multiply(other);
```

can be rewritten to

```
quaternion.copy(other);
```

?

src/components/audio-zone.js
```diff
@@ -42,17 +49,35 @@ AFRAME.registerComponent("audio-zone", {
...
+      this.el.object3D.updateMatrix();
```
takahirox, line 57: To prevent unnecessarily raising the dirty flag, I would like to suggest to call `updateMatrix()` only if `quaternion` has diff with the previous values, like

```
const EPSILON = 0.00000000001;

...

this.el.object3D.parent.getWorldQuaternion(quat).inverse();
if (!this.el.object3D.quaternion.near(quat, EPSILON)) {
  this.el.object3D.quaternion.copy(quat);
  this.el.object3D.updateMatrix();
}
```

takahirox (commented):
Honestly I'm not so familiar with Audio zone and debugger. So I have added only Three.js API related comments.

src/components/audio-zone.js
keianhzo: Oh yep, there's no need at all. Updated.

src/components/audio-zone.js
keianhzo: Updated.

src/components/audio-zone.js
keianhzo: Yes, that's better.

src/components/audio-zone.js
keianhzo: Yes, that's more efficient. Updating.

keianhzo:
Thanks for the review @takahirox!

PR #5054, opened 2022-01-26
Improved cursor with adjustable size

brianpeiris:
The cursor now has a border, which makes it more visible on bright backgrounds and objects. You can now also adjust the cursor size up to 5 times its normal size.

cursor-improved

The implementation now uses a flat billboarded circle geometry for the visual representation of the cursor, instead of a sphere. The circle is textured with a bordered image. The cursor's behavior hasn't changed, since it still uses a sphere as the physics shape.


matthewbcool (commented):
Looks great! I think the scale state needs to be initialized somewhere with an explicit value. If you enter the space, users won't see a cursor until they set the scale.

Two suggestions (I can move later to part of a larger design discussion): 
-  moving the preference itself to the "control" section under a new "cursor" subsection. This makes "misc." less cluttered and gives us room to possibly add more options (color/texture/matcap) later.

-  changing the scaling to be less dramatic. The increments feel a bit too large right now to me.

brianpeiris:
Fixed the initialization issue. Nice catch! Also set the size slider to allow 0.5 increments.

I agree we ought to re-organize the preferences now, it's a bit messy, but I want to keep this particular PR small and simple.

PR #5060, opened 2022-01-27
optimized images and added min-height for popover menu

robinkwilson:
**Before:** The emoji menu loads slowly for slow internet connections/slow computers and you can't click on anything until the pngs load.

**Now:** Emoji menu loads quicker AND you can click on the words before the pngs load to get the reaction in world.

Replaced the emoji pngs with smaller ones. Ex - clap emoji file 150 kb -> 16kb.


matthewbcool (approved):
Looks good, nice optimization. 
may want to consider reducing min-height to 48px for some browsers when resizing to really narrow widths.

PR #5068, opened 2022-01-31
Replace cursor png with SVG

carmebar:
image
The PNG had its own transparency and then a 90% opacity is applied via CSS. The opacity could be adjusted via CSS if needed.

Saves around 8KB and prevents loading an extra resource from the server. If inlining is not desired using an SVG file is still advised since it would only be a few bytes.


matthewbcool (commented):
This looks good to me, thanks for your contribution. @brianpeiris ok to merge?

brianpeiris:
I would prefer using an SVG file or just optimizing the existing PNG.

carmebar:
Updated to use SVG file instead of inlined and changed opacity to 0.8 to match old transparency.

brianpeiris:
I'm not sure why, but the SVG does not seem to work in Firefox. Can you verify? If it gets more complicated to solve, I'd rather go with an optimized PNG.

carmebar:
Fixed. Seems related to:
https://bugzilla.mozilla.org/show_bug.cgi?id=700533

src/assets/images/cursor.svg
```diff
@@ -0,0 +1 @@
+<svg height="100" width="100" viewBox="-4 -4 8 8" xmlns="http://www.w3.org/2000/svg"><circle r="4"/><circle r="3" fill="#fff"/></svg>
```
brianpeiris, line 1:
```suggestion
<svg height="256" width="256" viewBox="-4 -4 8 8" xmlns="http://www.w3.org/2000/svg"><circle r="4"/><circle r="3" fill="#fff"/></svg>
```

src/hub.html
```diff
@@ -1069,7 +1069,7 @@
             <a-circle
                 id="right-cursor-visual"
                 radius="0.02"
-                material="shader: flat; src: #cursor-image; depthTest: false; opacity: 0.9;"
+                material="shader: flat; src: #cursor-image; depthTest: false; opacity: 0.8;"
```
brianpeiris, line 1072:
```suggestion
                material="shader: flat; src: #cursor-image; depthTest: false; opacity: 0.9;"
```

src/hub.html
```diff
@@ -1094,7 +1094,7 @@
             <a-circle
                 id="left-cursor-visual"
                 radius="0.02"
-                material="shader: flat; src: #cursor-image; depthTest: false; opacity: 0.9;"
+                material="shader: flat; src: #cursor-image; depthTest: false; opacity: 0.8;"
```
brianpeiris, line 1097:
```suggestion
                material="shader: flat; src: #cursor-image; depthTest: false; opacity: 0.9;"
```

brianpeiris (commented):
Thanks. I'd suggest setting the size to 256x256 so that it scales better and does not alias as much when rendered. I'd also suggest leaving the opacity as it was.

brianpeiris (approved):
LGTM

brianpeiris:
Inlining is a good thought. We should look into doing that more systematically for all our iconography. Ideally it would be an automated post-processing step so that we don't have to sacrifice readability and maintainability, especially for marginal optimizations like this one. Automating it is a big step though, and arguably we've got bigger fish to fry at the moment.

carmebar:
I will look into that next time I have some free time. Any pointers to implement that kind of feature in this codebase?

brianpeiris:
I'd have to put more thought into it to give you any pointers. I personally don't spend a lot of time modifying UI code currently, and honestly I'm not sure where it would fit into our overall plans with the UI, so I'd hesitate to get someone from the community involved in a large-ish project like that. If you're particularly interested in it, I'd suggest opening an enhancement ticket to at least start a discussion before spending a lot of time on it.

PR #5074, opened 2022-02-01
Fix video-texture-target rendering in Chrome

### LINKED ISSUE #5061: Video Textures broken in Chrome
**Description**
video textures appear to be broken in chrome right now. No issue in firefox and fine in other chromium browsers. 

**To Reproduce**
1. Enter a room in chrome and load the test glb
https://github.com/matthewbcool/rc3-hubs-assembly-assets/blob/main/glb/funhousemirror.glb?raw=true

**Expected behavior**
test glb should load without issue

**Screenshots**
Screen_Shot_2022-01-21_at_4 37 52_PM

**Hardware**
 - Device: desktop
 - OS: MacOs
 - Browser: Chrome

**Additional context**
Thanks @yakyouk for the report.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-1257)

netpro2k:
Due to a race condition in how aframe sets up cameras (which appears to be browser dependent) the main camera was never getting the video texture target layer added to it, so it was not rendering them. 

This happens because the `cameraready` event only fires in the case aframe created a default camera (in which case we would end up adding the layer to the wrong camera anyway) or it "detected" our "user" camera in the scene at startup (this is what is "supposed" to happen). For whatever reason (likely custom element API weirdness) this does not always happen (it usually does in Firefox, but rarely if ever does in Chrome). We work around this in the `set-active-camera` component, but this never ends up firing the `cemeraready` event so the layer was never being added.

To fix I matched the hacky code we have elsewhere to deal with this same issue... I also noticed the old code likely was not working for VR mode, so I fixed that as well. 

We really need to just rip out the built in aframe camera system, it's not providing anything to us and we continue to actively fight it. This is part of "taking back the render loop" body of work, which I now have a renewed desire to do sooner.

fixes #5061

### DIFF
--- a/src/systems/camera-system.js
+++ b/src/systems/camera-system.js
@@ -204,9 +204,6 @@ export class CameraSystem {
     this.mode = CAMERA_MODE_SCENE_PREVIEW;
     this.snapshot = { audioTransform: new THREE.Matrix4(), matrixWorld: new THREE.Matrix4() };
     this.audioSourceTargetTransform = new THREE.Matrix4();
-    scene.addEventListener("cameraready", ({ detail: { cameraEl } }) => {
-      cameraEl.getObject3D("camera").layers.enable(Layers.CAMERA_LAYER_VIDEO_TEXTURE_TARGET);
-    });
     waitForDOMContentLoaded().then(() => {
       this.avatarPOV = document.getElementById("avatar-pov-node");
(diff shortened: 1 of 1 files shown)


keianhzo (approved):
LGTM

PR #5079, opened 2022-02-03
Replace three-bmfont-text with troika-three-text

netpro2k:
This is based on the work in #4855 but goes further, replacing the existing text component completely and dropping three-bmfont-text. Support for old `text` component in gltf files (if any actually exists) will be lightly maintained by mapping old properties, but minimum effort has gone into making sure they actually render identically. Text in the client itself has been updated to look roughly the same as before.

image

This also touches up the nametag with an outline (since that is now a feature we have for text) and also fixes the inconsisten application of tonemapping on the elements of the in world UI. This currently means **not** tonemapping all the elements. We may instead choose to later tonemap all of it, but in their case it should certainly be consistent.

This sets us up to render unicode in nametags as well, but we will need to find a suitable font, and other work needs to be done along the chain to change what is considered a valid displayName.


src/gltf-component-mappings.js
```diff
@@ -24,6 +24,8 @@ function registerRootSceneComponent(componentName) {
 
 registerRootSceneComponent("fog");
 
+AFRAME.GLTFModelPlus.registerComponent("troika-text", "troika-text");
```
keianhzo, line 27: We need to either rename the component in Spoke or map this to the updated "text" component so the Spoke Troika PR works. I've already done that in a branch with a few other updates on that PR so I can push it similarly to what you have done with this one.

keianhzo (approved):
What's the plan with the Blender component? I see that you have a new one on the works but if we land this the old component will be broken.

The Spoke PR also needs some updates. In Spoke is a bit simpler as we don't have a preexisting text path so probably just naming that one straight to "text" will be ok and we can get rid of the "troika-text" -> "text" mapping from `gltf-component-mapping` here.

keianhzo:
I've gone ahead and updated the Spoke PR to align it with this work: https://github.com/mozilla/Spoke/pull/1217

src/gltf-component-mappings.js
keianhzo: I have renamed the component in the new Spoke PR so we can probably remove this mapping.

netpro2k: Ah yeah that's actually a bug, no component called `troika-text` exists anymore with this PR, so this shouldn't be there. I decided it was worth just replacing wholesale since its not exposed in Spoke and I very highly doubt it was being used in many Blender scenes since it was much too hard to configure correctly.

PR #5088, opened 2022-02-07
Fix `for-of` in deploy script

juunini:
# Doing

- Line 51 `in` to `of`

# Why

I think this code is intended to delete `./dist` and `./admin/dist`
But, code is doing `rmdir 0` and `rmdir 1`
So, I fix code doing `rmdir ./dist` and `rmdir ./dist/admin`

image

## Before

image

## After

image


netpro2k (approved):
Nice catch, lgtm

PR #5096, opened 2022-02-09
Account for cloudflare when loading admin panel assets

brianpeiris:
When a Hubs Cloud stack is configured to use Cloudflare, we need to change the way we load uploaded assets in the admin panel.


robinkwilson:
Code looks good. Can you change this into qa-stage, we'll move it to qa-test quickly afterward. Thank you!

brianpeiris:
Replaced by https://github.com/mozilla/hubs/pull/5100

PR #5100, opened 2022-02-09
Account for cloudflare when loading admin panel assets

brianpeiris:
When a Hubs Cloud stack is configured to use Cloudflare, we need to change the way we load uploaded assets in the admin panel.


brianpeiris:
Looks like the CI run is failing here because of the CI memory issue that we've now fixed on `master`. I ran the checks locally and they passed.

robinkwilson (approved):

PR #5125, opened 2022-02-12
Fixes broken admin scene/avatar listings

wmurphyrd:
BASE_ASSETS_PATH was not available in the `process.env` generated by webpack, so it was undefined when read by `config.js`

Got this after meting in latest hubs-cloud: 
image

After this fix:
image

Others reporting this issue on Discord:
https://discord.com/channels/498741086295031808/819203046931693589/941487090942742548

This fix doesn't really explain what went wrong - as this isn't a new env var and wasn't in the webpack defines before, but it works :upside_down_face:

### DIFF
--- a/admin/webpack.config.js
+++ b/admin/webpack.config.js
@@ -225,7 +225,8 @@ module.exports = (env, argv) => {
           RETICULUM_SERVER: process.env.RETICULUM_SERVER,
           CORS_PROXY_SERVER: process.env.CORS_PROXY_SERVER,
           POSTGREST_SERVER: process.env.POSTGREST_SERVER,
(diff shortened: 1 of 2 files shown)


Dayk0:
Hello,

Thank you for the correction ! It would be necessary to put the modified file in the main branch because the current client is unusable with this bug.

robinkwilson:
@wmurphyrd can you merge this PR into qa-stage branch?

wmurphyrd:
@robinkwilson updated target & rebased :+1:

robinkwilson:
Thank you! I'll take care of all the merges. It will make it's way into all of the branches we use for hubs-cloud. Thanks for the help with this!!

PR #5141, opened 2022-02-16
Restore microphone icon volume meter

mattrossman:
After #5014, the microphone icon in the mute button no longer displays the microphone volume, making it hard to tell if the microphone is working at a glance.

I copied the [old behavior](https://github.com/mozilla/hubs/blob/8adfe267f8350ed62be2c1c65a00f70d51aa95cd/src/react-components/room/VoiceButtonContainer.js) and changed the cutoff to 0.1 due to b66bfed18c606256d41e460baf7669744395fdc6.

If desired, this could be an opportunity to revisit https://github.com/mozilla/hubs/pull/5014#discussion_r788427276 to reduce re-renders.

### DIFF
--- a/src/react-components/room/AudioPopoverButton.js
+++ b/src/react-components/room/AudioPopoverButton.js
@@ -1,4 +1,4 @@
-import React, { useRef } from "react";
+import React, { useEffect, useRef } from "react";
 import PropTypes from "prop-types";
 import styles from "./AudioPopover.scss";
 import { Popover } from "../popover/Popover";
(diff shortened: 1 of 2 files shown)


matthewbcool (commented):
thanks for this contribution. Looks good to me, pending @keianhzo approval.

src/react-components/room/AudioPopoverButton.js
```diff
@@ -18,13 +18,31 @@ export const AudioPopoverButton = ({
...
   isMicrophoneEnabled,
+  micLevel,
   onChangeMicrophoneMuted
 }) => {
   const ref = useRef();
   const intl = useIntl();
   const title = intl.formatMessage(invitePopoverTitle);
   const popoverApiRef = useRef();
 
+  useEffect(
```
keianhzo, line 29: LGTM my only comment would be to use callback refs instead of just passing down the ref as we could end up with stale refs if the component changes downstream which in this case shouldn't happen but it's a good practice: https://reactjs.org/docs/hooks-faq.html#how-can-i-measure-a-dom-node

keianhzo (changes requested):

src/react-components/room/AudioPopoverButton.js
mattrossman: Swapped the useRef / useEffect for a useCallback

keianhzo: Thanks! I want to refactor this in another PR to try to limit the re-renders to the most inner button element.

keianhzo (approved):

PR #5149, opened 2022-02-17
Allow testing CSP with webpack-dev-server

netpro2k:
CSP issues often only surface when deploying. This PR adds the `DEV_CSP_SOURCE` environment variable which will have webpack-dev-server serve the same CSP as the destination host. 

Typically you might want to test things like this `DEV_CSP_SOURCE="dev.reticulum.io" npm run dev`

Keeping this as an environment variable to start as I want people to be able to test this a bit more and also we need to figure out how this should work with Hubs Cloud.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -252,6 +252,18 @@ module.exports = async (env, argv) => {
     ]
   };
 
+  const devServerHeaders = {
+    "Access-Control-Allow-Origin": "*"
+  };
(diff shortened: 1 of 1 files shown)


brianpeiris (approved):
I'd suggest `DEV_CSP_SOURCE`, to make the variable name more accurate. LGTM otherwise.

PR #5160, opened 2022-02-18
fix broken image for avatar_unavailable 

### LINKED ISSUE #5151: Hard coded image file requires CSP addition in Hubs Cloud
The fallback image used when an avatar thumbnail can't be found is hard coded to https://asset-bundles-prod.reticulum.io/bots/avatar_unavailable.png and referenced in several place ([such as this](https://github.com/mozilla/hubs/blob/41373e0078ce55eb09b6034fe2c621871c88ce96/src/utils/avatar-utils.js#L53)) in the code. This won't work on Hubs Cloud unless you include the host domain in the img-src CSP list.

Perhaps this should be replaced with a relative asset in the client or something similar?

FYI The same URL is also referenced [in the Reticulum code](https://github.com/mozilla/reticulum/blob/1254f62e50ba5e7bdd0d6a984c7e26850b0527ed/lib/ret/media_search.ex#L704) and may need to be duplicated there as a relative asset.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-1278)

yakyouk:
Partial fix for #5151
This fixes a broken image in the Client when viewing an avatar without a thumbnail (direct glb)
image


netpro2k (approved):
LGTM

PR #5192, opened 2022-02-26
Fix localization typo

juunini:
# Doing

- `src/avatar.js`: `"avatar-editor.field.alllow-remixing"` to `"avatar-editor.field.allow-remixing"`
- `src/react-components/avatar-editor.js`: `"avatar-page.copyied-avatar"` to `"avatar-page.copied-avatar"`

# Why

I found this typos when doing korean localization.
I'll be soon upload pull request korean localization.


matthewbcool (approved):
LGTM. General note- We need to be vigilant in double checking future translation json are using the fixed keys. I know there are several translations being worked on now using the keys with typos. Thank you very much for this fix!

PR #5195, opened 2022-02-28
New nametags

keianhzo:
This PR adds support for a new name tag design.

Screen Shot 2022-02-28 at 17 54 01

- New name-tag design
- State visualization for Talking, Typing and Hand Raised states.
- Support for name tag visualization filter: always, never, close (with distance slider), frozen only, speaking only.

This PR depends on this Reticulum PR: https://github.com/mozilla/reticulum/pull/566


src/components/name-tag.js
```diff
@@ -0,0 +1,377 @@
...
+      if (this.identityName.length > DISPLAY_NAME_LENGTH) {
+        this.identityName = this.identityName.slice(0, DISPLAY_NAME_LENGTH).concat("...");
+      }
+      this.identityNameEl.setAttribute("text", { value: this.identityName });
+      this.identityNameEl.object3D.visible = this.el.sceneEl.is("frozen");
+    }
+  },
+
+  onModelLoading() {
+    this.tmpNametagVisible = false;
```
netpro2k, line 201: This state seems redundant with `this.model`

src/components/name-tag.js
```diff
@@ -0,0 +1,377 @@
...
+    this.store = window.APP.store;
+    this.displayName = null;
+    this.identityName = null;
+    this.isTalking = false;
+    this.isTyping = false;
+    this.isOwner = false;
+    this.isRecording = false;
+    this.isHandRaised = false;
+    this.volumeAvg = new MovingAverage(128);
+    this.isNametagVisible = true;
```
netpro2k, line 39: I think this represents the desired state so maybe something like `shouldBeVisible` would be better. The actual state should live in `this.el.object3D.visibile`, and we should apply that once per frame, taking into consideration other overrides like hiding during model loading.

src/components/name-tag.js
```diff
@@ -0,0 +1,377 @@
...
+  updateNameTag() {
+    if (!this.model) return;
+    this.updateDisplayName();
+    this.updateContainer();
+    this.updateVolume();
+    this.updateState();
+  },
+
+  updateContainer() {
+    this.nametagBackgroundEl.setAttribute("visible", this.isNametagVisible);
```
netpro2k, line 297: Can we just set the visibility of the root element instead of individually managing each of the children?

src/components/name-tag.js
```diff
@@ -0,0 +1,377 @@
...
+    this.nametagBackgroundEl.setAttribute("slice9", {
+      width: this.size.x + NAMETAG_BACKGROUND_PADDING * 2,
+      height: NAMETAG_HEIGHT
+    });
+    this.updateBorder();
+  },
+
+  updateVolume() {
+    this.nametagVolumeEl.setAttribute("visible", this.isTalking && this.isNametagVisible);
```
netpro2k, line 306: As elsewhere, skip the aframe stuff, but also this object's visibility should only depend on `this.isTalking` (or better yet directly on the analyzer `isTalking` state). The nametag visibility should be handled by hiding the root.

src/components/name-tag.js
```diff
@@ -0,0 +1,377 @@
...
+    this.onModelLoading = this.onModelLoading.bind(this);
+    this.onModelLoaded = this.onModelLoaded.bind(this);
+    this.onAnalyserVolumeUpdated = this.onAnalyserVolumeUpdated.bind(this);
+    this.onStateChanged = this.onStateChanged.bind(this);
+
+    this.avatarRig = document.getElementById("avatar-rig");
+    this.ikRootEl = findAncestorWithComponent(this.el, "ik-root");
+    this.playerInfo = this.el.components["player-info"];
+
+    this.nametagEl = this.el.querySelector(".nametag");
```
netpro2k, line 57: I didn't check every one, but `nametagEl` for example we never use directly, we always get the `object3D` off of it, so we should just store a ref to that directly. Even when we do use both the element and object3D, we should probably prefer the object3D reference. I think I will add that to our best-practices.md

src/components/name-tag.js
```diff
@@ -0,0 +1,377 @@
...
+    this.identityName = presenceMeta.profile.identityName;
+    this.isRecording = !!(presenceMeta.streaming || presenceMeta.recording);
+    this.isOwner = !!(presenceMeta.roles && presenceMeta.roles.owner);
+    this.isTyping = !!presenceMeta.typing;
+    this.isHandRaised = !!presenceMeta.hand_raised;
+    this.updateNameTag();
+  },
+
+  updateDisplayName() {
+    if (this.displayName) {
```
netpro2k, line 173: It seems like we probably end up doing this quite often, even if the displayname has not changed

src/components/name-tag.js
```diff
@@ -0,0 +1,377 @@
...
+      this.nametagTextEl.addEventListener(
```
netpro2k, line 174: Related to the above, since we end up calling this multiple times even when the text doesn't need to be updated, this event listener will end up just sitting around waiting for the event to be fired (`sync()` in Troika does not fire the callback if the text wasn't dirty https://github.com/protectwise/troika/blob/main/packages/troika-three-text/src/Text.js#L402). This would result in a bunch of event listeners hanging around and then all firing at once, when you actually did change your displayname. Its all synchronous so I don't think anything would break but it would do a bunch of unnecessary work.

src/components/name-tag.js
```diff
@@ -0,0 +1,377 @@
...
+        value: this.displayName
+      });
+    }
+
+    if (this.identityName) {
+      if (this.identityName.length > DISPLAY_NAME_LENGTH) {
+        this.identityName = this.identityName.slice(0, DISPLAY_NAME_LENGTH).concat("...");
+      }
+      this.identityNameEl.setAttribute("text", { value: this.identityName });
+      this.identityNameEl.object3D.visible = this.el.sceneEl.is("frozen");
```
netpro2k, line 196: This entity already has `visibility-while-frozen` which is doing this same thing. They are going to be battling for control over visibility of this object. Ditching the aframe component is fine but then we should separate out where we actually apply the desired visibility state to 1 spot in tick as described in my `updateNametag` comment.

src/components/name-tag.js
```diff
@@ -0,0 +1,377 @@
...
+        this.wasNametagVisible = this.isNametagVisible;
+        this.isNametagVisible = true;
+        this.isNametagVisible !== this.wasNametagVisible && this.updateNameTag();
+      }
+    }
+    this.isNametagVisible && this.isTalking !== this.wasTalking && this.updateBorder();
+    this.isNametagVisible && this.updateVolume();
+  },
+
+  onStateChanged() {
```
netpro2k, line 262: This wrapping function feels superfluous.

src/components/name-tag.js
```diff
@@ -0,0 +1,377 @@
...
+
+  onModelLoading() {
+    this.tmpNametagVisible = false;
+    this.updateNameTagVisibility();
+    this.model = null;
+  },
+
+  onModelLoaded({ detail: { model } }) {
+    clearInterval(this.firstIkStepHandler);
+    const checkIkStep = () => {
```
netpro2k, line 208: Hmm, this feels pretty messy. Do we actually care to wait for ik-controller's `hadFirstTick`? If we do it probably makes sense in this case for `ik-controller` to fire an event. 

Like the other comments in this PR, it call a function directly on nametag instead of firing an event, but in this case it feels odd to me for `ik-controller` to know about nametag. (But does not for example feel odd for the chat box to be involved in notifying about typing status)

netpro2k (changes requested):
Really excited for this feature, it has been a long time coming. I left a bunch of "best practices" sort of feedback, but I do think name-tag.js in particular could use some cleanup. The code flow is pretty hard to follow. 

It is definitely a very tricky thing to implement dynamic UI like this on top of an imperative interface (react-three-fiber would probably do wonders here, but we are a ways off from that being viable even if it ends up making sense). I made some suggestions about how I might restructure some things, but its hard to tell if it would work out well without trying it.

Let me know if you want to chat more in depth on any of it.

src/components/audio-feedback.js
```diff
@@ -84,6 +84,7 @@ AFRAME.registerComponent("networked-audio-analyser", {
   tick: function(t) {
     if (!this.avatarIsQuiet) {
       this._updateAnalysis(t);
+      this.el.emit("analyser-volume-updated", { volume: this.volume });
```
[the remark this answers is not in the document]
keianhzo: Makes sense. Updated.

src/components/name-tag.js
keianhzo: Make sense. Updated.

src/components/name-tag.js
```diff
@@ -0,0 +1,377 @@
...
+    this.identityName = null;
+    this.isTalking = false;
+    this.isTyping = false;
+    this.isOwner = false;
+    this.isRecording = false;
+    this.isHandRaised = false;
+    this.volumeAvg = new MovingAverage(128);
+    this.isNametagVisible = true;
+    this.size = new THREE.Vector3();
+    this.avatarBBAA = new THREE.Box3();
```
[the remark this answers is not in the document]
keianhzo: Updated

src/components/name-tag.js
```diff
@@ -0,0 +1,377 @@
...
+    let typingAnimTime = 0;
+    const worldPos = new THREE.Vector3();
+    const avatarRigWorldPos = new THREE.Vector3();
+    const matPos = new THREE.Vector3();
+    const matRot = new THREE.Quaternion();
+    const matScale = new THREE.Vector3();
+    const mat = new THREE.Matrix4();
+    return function(t) {
+      if (this.model) {
+        this.nametagEl.object3D.getWorldPosition(worldPos);
```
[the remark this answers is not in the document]
keianhzo: Fixed

src/components/name-tag.js
keianhzo: Updated.

src/components/name-tag.js
keianhzo: Updated

src/components/name-tag.js
keianhzo: Updated

src/components/name-tag.js
keianhzo: Updated

src/components/name-tag.js
keianhzo: Removed

src/components/name-tag.js
keianhzo: We need to wait for the first tick otherwise the BBAA size is not correct. Makes sense to use an event, updated.

src/components/name-tag.js
```diff
@@ -0,0 +1,377 @@
...
+      Math.abs(tmpVector.subVectors(this.ikRootEl.object3D.position, this.avatarBBAACenter).y) +
+      this.avatarBBAASize.y / 2 +
+      NAMETAG_OFFSET;
+    this.nametagElPosY = this.nametagHeight;
+    this.size = this.nametagTextEl.components["text"].getSize();
+    this.size.x = Math.max(this.size.x, NAMETAG_MIN_WIDTH);
+    this.updateNameTag();
+  },
+
+  onAnalyserVolumeUpdated({ detail: { volume } }) {
```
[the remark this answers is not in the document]
keianhzo: Yep, makes sense. As we are using `avatarIsQuiet` for disabling the analyzer updates with a 10secs grace period and that too high for our purpose if introduced another flag to check if the avatar is talking with a 1sec time.

src/components/name-tag.js
```diff
@@ -0,0 +1,377 @@
...
+    this.volumeAvg.push(Date.now(), volume);
+    if (!this.playerInfo.data.muted) {
+      const average = this.volumeAvg.movingAverage();
+      isTalking = average > 0.01;
+    }
+    this.wasTalking = this.isTalking;
+    this.isTalking = isTalking;
+    if (this.nametagVisibility === "showSpeaking") {
+      if (!this.isTalking && this.wasTalking) {
+        this.frozenTimer = setTimeout(() => {
```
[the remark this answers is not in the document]
keianhzo: Updated

src/components/name-tag.js
keianhzo: Updated

src/components/name-tag.js
keianhzo: Updated

src/components/name-tag.js
keianhzo: Updated

src/components/name-tag.js
```diff
@@ -0,0 +1,377 @@
...
+    this.nametagStatusBorderEl.setAttribute(
+      "text-button",
+      `backgroundColor: ${getThemeColor(
+        this.isHandRaised ? "nametag-border-color-raised-hand" : "nametag-border-color"
+      )}`
+    );
+  },
+
+  updateState() {
+    this.recordingBadgeEl.setAttribute("visible", this.isRecording && this.isNametagVisible);
```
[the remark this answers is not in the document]
keianhzo: Updated

src/components/player-info.js
```diff
@@ -42,40 +37,36 @@ AFRAME.registerComponent("player-info", {
...
+      if (playerPresence) {
+        this.updatePermissions(playerPresence.metas[0]);
+      }
+    });
+
     registerComponentInstance(this, "player-info");
   },
+
   remove() {
+    clearTimeout(this.frozenTimer);
```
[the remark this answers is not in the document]
keianhzo: Updated

src/components/troika-text.js
```diff
@@ -134,5 +136,18 @@ AFRAME.registerComponent("text", {
   remove: function() {
     // Free memory
     this.troikaTextMesh.dispose();
-  }
+  },
+
+  getSize: (function() {
```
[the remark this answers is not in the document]
keianhzo: Updated

src/hub.html
```diff
@@ -127,48 +130,57 @@
...
+                            position="0 0.25 0.001"
+                            visible="false"
+                        ></a-entity>
+                        <a-entity
+                            class="identityName"
+                            text="side: double; textAlign: center; color: #ddd; fontSize: 0.1; outlineWidth: 1%;"
+                            visibility-while-frozen="withinDistance: 100; requireHoverOnNonMobile: false;"
+                            position="0 0.4 0"
+                        ></a-entity>
+                        <a-image class="hand-raised-id" scale="0.2 0.2 0.2" position="0 -0.3 0.001" src="#hand-raised" visible="false" material="alphaTest:0.1;"></a-image>
```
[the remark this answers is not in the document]
keianhzo: Sprites are currently tinted baased on the theme so I'm using an image to avoid that

src/hub.html
```diff
@@ -127,48 +130,57 @@
...
                                 ik-controller
                                 hand-pose__left
                                 hand-pose__right
                                 networked-audio-analyser
-                            ></a-entity>
+                            >
+                            </a-entity>
                         </template>
 
                         <template data-name="Neck">
```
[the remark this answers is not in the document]
keianhzo: Removed

src/hub.html
```diff
@@ -1264,9 +1303,7 @@
                 </template>
 
                 <template data-name="Neck">
```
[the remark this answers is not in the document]
keianhzo: Removed

src/hub.js
```diff
@@ -952,6 +953,12 @@ document.addEventListener("DOMContentLoaded", async () => {
   scene.addEventListener("action_camera_recording_started", () => hubChannel.beginRecording());
   scene.addEventListener("action_camera_recording_ended", () => hubChannel.endRecording());
 
+  scene.addEventListener("action_raise_hand", () => window.APP.hubChannel.raiseHand());
```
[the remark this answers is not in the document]
keianhzo: Updated

src/react-components/room/ChatSidebarContainer.js
```diff
@@ -182,8 +183,11 @@ export function ChatSidebarContainer({ scene, canSpawnMessages, presences, occup
       } else if (e.key === "Escape") {
         onClose();
       }
+      clearTimeout(typingTimeoutRef.current);
+      typingTimeoutRef.current = setTimeout(() => scene.emit("action_typing_ended"), 500);
```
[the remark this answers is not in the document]
keianhzo: Updated

src/utils/component-utils.js
```diff
@@ -34,3 +34,15 @@ export function getPlayerInfo(sessionId) {
     return info.playerSessionId === sessionId;
   });
 }
+
+export function getPlayerPresence(sessionId) {
```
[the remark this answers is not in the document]
keianhzo: Removed

src/utils/theme.js
```diff
@@ -65,6 +72,17 @@ function activateTheme() {
         "text-button",
         `textHoverColor: #fff; textColor: #fff; backgroundColor: ${actionColor}; backgroundHoverColor: ${actionHoverColor}`
       );
+
+    document
+      .querySelector("#nametag-background-mixin")
```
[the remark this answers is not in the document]
keianhzo: I've moved this to a manual update inside the name tag component and also add the volume bar to the theme which was missed here.

src/components/name-tag.js
```diff
@@ -0,0 +1,377 @@
...
+        mat.compose(
+          matPos,
+          matRot,
+          matScale
+        );
+        setMatrixWorld(this.nametagEl.object3D, mat);
+      }
+
+      if (this.nametagVisibility === "showClose") {
+        this.avatarRig.object3D.getWorldPosition(avatarRigWorldPos);
```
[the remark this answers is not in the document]
keianhzo: I've refactored the name tag visibility to a system to handle all the visibility cases.

src/components/name-tag.js
```diff
@@ -0,0 +1,377 @@
...
+    } else if (this.nametagVisibility === "showSpeaking") {
+      prefNametagVisibility = this.isTalking;
+    } else {
+      prefNametagVisibility = true;
+    }
+    this.isNametagVisible = prefNametagVisibility && this.tmpNametagVisible;
+    this.updateNameTag();
+  },
+
+  updateNameTag() {
```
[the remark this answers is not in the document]
keianhzo: My original approach was to try to keep as little code in the tick as possible but I think it was making it more confusing. I've refactored and moved most of the logic to the tick.

src/hub.html
```diff
@@ -127,48 +130,57 @@
 
                     <a-entity class="right-controller" matrix-auto-update></a-entity>
 
+                    <a-entity class="nametag" tags="ignoreSpaceBubble: true;" position="0 0 0" billboard="onlyY: true;">
+                        <a-entity class="nametag-background" mixin="nametag-background-mixin" slice9="width: 1">
+                            <a-entity class="nametag-status-border" mixin="nametag-status-border-mixin" visible="false" slice9="width: 1"></a-entity>
+                            <a-entity class="nametag-text" text="side: double; textAlign: center; color: #ddd; fontSize: 0.1;" text-raycast-hack position="0 0.025 0.001"></a-entity>
+                            <a-plane class="nametag-volume" visible="false" position="0 -0.075 0.001" material="shader:flat;" color="#7ED320" height="0.025" width="1"></a-plane>
+                            <a-entity class="nametag-typing" position="0 -0.075 0.001">
```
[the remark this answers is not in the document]
keianhzo: Right now the sprite reuse the same material so we would need to do some shader work to allow updating the opacity I guess. I've reduced the circle segments for the moment.

PR #5208, opened 2022-03-04
German localization

markusTraber:
Added German localization. It is based on #4233, but updated to the master branch. We fixed some translation errors and furthermore changed the salutation from formal ("Sie") to informal ("Du").

Bildschirmfoto 2022-03-04 um 09 28 31


matthewbcool (approved):
lgtm, Thanks so much for this huge contribution! Fantastic!

typo fixes to check: 
avilable
copyied
alllow

Issue #5062, opened 2022-01-28
As a room Owner, Mute/MuteAll no longer mutes participants.


matthewbcool:
**Description**
Selecting a person from the people menu and muting them does not mute participant. (Chrome) 

**To Reproduce**
1. Open people menu with multiple participants in room
2. Click on person name
3. Select 'Mute'

**Expected behavior**
participant should be muted

**Hardware**
 - Device: [Desktop]
 - OS: [MacOs]
 - Browser: [Chrome]

**Additional context**
Regression, previously these features have worked for me.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-1258)

Dayk0:
I tried again today and the "mute microphone" button doesn't work. Unable to mute participants' microphones.

matthewbcool:
This issue has been resolved: https://github.com/mozilla/hubs/commit/a185073540da114dba73469e579e1b6d586d9d69

Confirmed room owners can once again mute participants.

PR #5219, opened 2022-03-06
revert to dialog enable mic

matthewbcool:
reverting to `APP.dialog.enableMicrophone(false)` seems to fix the issue #5062. Disregard this PR if there is a reason for keeping this on `mediaDevicesManager`

https://user-images.githubusercontent.com/4493657/156936402-591a5803-cf2d-4d31-893d-bfd4af76cc6d.mp4

### DIFF
--- a/src/hub.js
+++ b/src/hub.js
@@ -1398,7 +1398,7 @@ document.addEventListener("DOMContentLoaded", async () => {
 
   hubPhxChannel.on("mute", ({ session_id }) => {
     if (session_id === NAF.clientId) {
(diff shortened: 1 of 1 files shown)


keianhzo:
Thanks for looking into this @matthewbcool . I just saw that we were wrongly still using the previous method in `MediaDevicesManager` that doesn't exist anymore. I've updated it to `APP.mediaDevicesManager.micEnabled = false;` so we route all media devices call through `MediaDevicesManager` and keep it as the authority for media devices related APIs

PR #5221, opened 2022-03-08
Fix typo `becuase` to `because`

juunini:
# Doing

- `src/components/camera-tool.js` line 641: `becuase` to `because`

I was find to way what makes to mirror with head.
So difficult T.T


takahirox (approved):
Good catch, thanks!

PR #5225, opened 2022-03-09
Add mirror component

juunini:
# Doing

Mar-09-2022 09-55-34

- Add mirror AFRAME component
- playerHead visible and hide repeatedly
- In Reference, camera-tool.js line 605 and line 663

# Etc

- will be create spoke mirror component soon

# Example

```html
<a-plane mirror scale="5 5 5" position="0 3 0"></a-plane>
<a-circle mirror="color: #7f7f7f" scale="5 5 5" position="0 3 0"></a-circle>
```

### DIFF
--- a/src/components/mirror.js
+++ b/src/components/mirror.js
@@ -39,8 +39,7 @@ AFRAME.registerComponent("mirror", {
   init() {
     const geometry = this.el.object3DMap?.mesh?.geometry;
     if (!geometry) {
-      console.error('can not find "object3DMap.mesh.geometry"');
-      return;
+      throw new Error('can not find "object3DMap.mesh.geometry"');
     }
 
     const reflector = new Reflector(geometry, {
--- a/src/components/mirror.js
+++ b/src/components/mirror.js
@@ -1,3 +1,4 @@
+import { THREE } from "aframe";
 import { Reflector } from "three/examples/jsm/objects/Reflector";
 
 /** @return { THREE.Object3D? } */
@@ -22,6 +23,10 @@ function setPlayerHeadVisible(visible) {
   playerHead.updateMatrixWorld(true, true);
 }
 
+const DEFAULT_MIRROR_GEOMETRY = new THREE.PlaneBufferGeometry();
+const DEFAULT_TEXTURE_WIDTH = window.innerWidth * window.devicePixelRatio;
+const DEFAULT_TEXTURE_HEIGHT = window.innerHeight * window.devicePixelRatio;
+
 /**
  * Should need to entity that has geometry primitive
  *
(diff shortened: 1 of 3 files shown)


src/components/mirror.js
```diff
@@ -0,0 +1,67 @@
...
+  schema: {
+    color: { type: "color", default: "#7f7f7f" }
+  },
+
+  init() {
+    const geometry = this.el.object3DMap?.mesh?.geometry;
+    if (!geometry) {
+      console.error('can not find "object3DMap.mesh.geometry"');
+      return;
+    }
```
hannut91, line 49: How about throwing an exception rather than simply printing the console?

src/components/mirror.js
```diff
@@ -0,0 +1,67 @@
...
+ */
+AFRAME.registerComponent("mirror", {
+  schema: {
+    color: { type: "color", default: "#7f7f7f" }
+  },
+
+  init() {
+    const geometry = this.el.object3DMap?.mesh?.geometry;
+    if (!geometry) {
+      console.error('can not find "object3DMap.mesh.geometry"');
```
hannut91, line 47: Strings in the code are using double quotes. How about using single quotes consistently?

```suggestion
      console.error("can not find 'object3DMap.mesh.geometry'");
```

juunini: image

Project's eslint rule T.T

src/components/mirror.js
juunini: Good. I'll accept it

src/components/mirror.js
juunini: Ah, I saw it wrong.
accept it. Thanks

juunini:
# Add

## Doing

- [ADD] default geometry is `PlaneGeometry`
    - Currently, only `PlaneGeometry` is supported, but will support geometry primitives such as `circle` in the future.
- [ADD] `DEFAULT_TEXTURE_WIDTH` and `DEFAULT_TEXTURE_HEIGHT`

## Reason why set `DEFAULT_TEXTURE_WIDTH` and `DEFAULT_TEXTURE_HEIGHT`

|When set texture width and height|When not set(as default) texture width and height|
|-|-|
|스크린샷 2022-03-10 오후 6 28 45|스크린샷 2022-03-10 오후 6 29 08|
|mirror renders Anti-aliased|mirror renders aliased|

johnshaughnessy (approved):
lgtm! 

Do you have any thoughts for what we can do to prevent performance degrading on low-performance hardware like phones and tablets when one or more mirrors are in the scene?

johnshaughnessy:
mirror

johnshaughnessy:
I've edited the PR summary so that this feature is shown on the [what's new page](https://hubs.mozilla.com/whats-new)

juunini:
- reduce view finder fps
- reduce texture width and height (aliased)
- when avatar is back of mirror, mirror not render. avatar is front of mirror, render mirror

PR #5226, opened 2022-03-09
Cloudflare instruction edit

Pattersonbl2:
Removed white spaces and added some lanaguge.

### DIFF
--- a/admin/src/react-components/content-cdn.js
+++ b/admin/src/react-components/content-cdn.js
@@ -310,7 +310,7 @@ class ContentCDNComponent extends Component {
                   <b>WARNING - Do not &quot;add site&quot; to Cloudflare</b>, you only need to create workers)
                 </li>
                 <li>
-                  In <b>Home</b> tab, click on <b>Workers</b> side panel. You&apos;ll be asked to create a workers subdomain. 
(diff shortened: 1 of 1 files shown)


admin/src/react-components/content-cdn.js
```diff
@@ -326,14 +326,14 @@ class ContentCDNComponent extends Component {
                 {hasValidWorkerDomain && (
                   <>
                     <li>
-                      In the Workers Dashboard click <b>Create Worker</b>.
+                      In the Workers Dashboard click <b>Create a Service</b>.
                     </li>
                     <li>
-                      Enter the name for the worker (at the top, above the script) as:
+                      Enter the name for the service (at the top, above the script) as:
```
johnshaughnessy, line 332: I'm surprised to see this change from `worker` to `service`. Is the thing that you're creating not a `worker`? (https://developers.cloudflare.com/workers/)

johnshaughnessy (approved):

admin/src/react-components/content-cdn.js
Pattersonbl2: Cloud flare called them services under worker now.   I believe because they want to push the serverless actions behind them.  
image

PR #5230, opened 2022-03-10
Fix default devices id not working on FF

keianhzo:
FF doesn't tag the default device as default so when using the "default" tag it wasn't choosing the right device. Use the first device as the default for all platforms.

### DIFF
--- a/src/utils/media-devices-manager.js
+++ b/src/utils/media-devices-manager.js
@@ -290,9 +290,7 @@ export default class MediaDevicesManager extends EventEmitter {
       this.audioTrack = newStream.getAudioTracks()[0];
       this.audioTrack.addEventListener("ended", async () => {
         this._scene.emit(MediaDevicesEvents.MIC_SHARE_ENDED);
-        if (this._permissionsStatus[MediaDevices.MICROPHONE] === PermissionStatus.GRANTED) {
-          this.startMicShare({ unmute: this.isMicEnabled });
-        }
+        this.startMicShare({ unmute: this.isMicEnabled });
       });
 
       if (/Oculus/.test(navigator.userAgent)) {
--- a/src/utils/media-devices-manager.js
+++ b/src/utils/media-devices-manager.js
@@ -265,15 +265,15 @@ export default class MediaDevicesManager extends EventEmitter {
     }
 
     if (result) {
-      this._scene.emit(MediaDevicesEvents.MIC_SHARE_STARTED);
       this._permissionsStatus[MediaDevices.MICROPHONE] = PermissionStatus.GRANTED;
+      this._scene.emit(MediaDevicesEvents.MIC_SHARE_STARTED);
       this.emit(MediaDevicesEvents.PERMISSIONS_STATUS_CHANGED, {
         mediaDevice: MediaDevices.MICROPHONE,
         status: PermissionStatus.GRANTED
       });
     } else {
-      this._scene.emit(MediaDevicesEvents.MIC_SHARE_ENDED);
       this._permissionsStatus[MediaDevices.MICROPHONE] = PermissionStatus.DENIED;
+      this._scene.emit(MediaDevicesEvents.MIC_SHARE_ENDED);
       this.emit(MediaDevicesEvents.PERMISSIONS_STATUS_CHANGED, {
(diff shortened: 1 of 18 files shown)


src/utils/media-devices-manager.js
```diff
@@ -86,7 +92,7 @@ export default class MediaDevicesManager extends EventEmitter {
         return this._micDevices;
       }
     } else {
-      return [{ value: DEFAULT_DEVICE_ID, label: "Default" }];
+      return [{ value: this.defaultInputDeviceId, label: this.micLabelForDeviceId(this.defaultInputDeviceId) }];
```
netpro2k, line 95: Its a bit hard to trace through but I think this branch will never actually be hit where `defaultInputDeviceId` will return anything but 'none' because `this._micDevices` will never be set since its guarded around the same condition https://github.com/mozilla/hubs/pull/5230/files#diff-68c343bc022cf6fd909117f937d2d42768cee117b0d0649e9ffc77bd22b83c78R237 (you don't actually see the relationship here because its hidden behind the getter.

I think this whole block can be simplified into:
```js
return MediaDevicesManager.isAudioInputSelectEnabled && this._permissionsStatus[MediaDevices.MICROPHONE] !== PermissionStatus.DENIED ?
         this._micDevices :
         [{ value: NO_DEVICE_ID, label: "None" }];
```

src/utils/media-devices-manager.js
```diff
@@ -127,10 +133,10 @@ export default class MediaDevicesManager extends EventEmitter {
       if (this._permissionsStatus[MediaDevices.MICROPHONE] === PermissionStatus.DENIED) {
         return NO_DEVICE_ID;
       } else {
-        return this.deviceIdForMicDeviceLabel(this.selectedMicLabel) || DEFAULT_DEVICE_ID;
+        return this.deviceIdForMicDeviceLabel(this.selectedMicLabel) || this.defaultInputDeviceId;
```
netpro2k, line 136: Again its pretty hard to trace through all the derived values. but I think `deviceIdForMicDeviceLabel` ends up calling the `micDevices` getter which ends up checking the same conditions that the surrounding if statement here is checking. That means it would already be returning `[{ value: NO_DEVICE_ID, label: "None" }]` as `micDevices`, which would then not pass the filter in `deviceIdForMicDeviceLabel`, so that would return `undefined`. I think this means you can just drop the entire if statement and just keep this line, but it should be `|| NO_DEVICE_ID` instead, which is I think is more correct anyway.

src/utils/media-devices-manager.js
```diff
@@ -78,6 +72,18 @@ export default class MediaDevicesManager extends EventEmitter {
     this._audioTrack = audioTrack;
   }
 
+  get defaultInputDeviceId() {
+    return this._micDevices && this._micDevices.length > 0 ? this._micDevices[0].value : NO_DEVICE_ID;
```
netpro2k, line 76: I don't see a case where `this._micDevices` can ever be null since its initialized to `[]` and only assigned the result of a `Array.prototype.map` which will always return an empty array, so no need to check for this here. (Having the check made me go digging through the code trying to figure out when this state could occur, but I couldn't find it)

I think though, with the other changes this might end up only getting called once, so I would just ditch the getter altogether as the data flow will be much clearer.

netpro2k (commented):
This change looks like it probably works correctly but I think its adding some additional layers of missdirection and the underlying code also can probably serve to have some missdirection removed. I only traced through the mic part in detail but I think its a similar case for speakers/video. 

Since I assume this is causing issues I am fine getting this fix out, but it does seem worth revisiting a cleanup here at some point.

src/utils/media-devices-manager.js
keianhzo: That's right, removing the check.

src/utils/media-devices-manager.js
keianhzo: Yep, true again. Simplifying.

arpu:
is this a bug in Firefox or not in the specs?

keianhzo:
> is this a bug in Firefox or not in the specs?

It seems to be a FF only thing, Chromium and Webkit always use unique Ids. The spec doesn't mention anything about using `default` as the device id.

PR #5233, opened 2022-03-10
Fix shadows for directinal lights

### LINKED ISSUE #4164: "Enable Dynamic Shadows" option does not work
**Description**
The "Enable Dynamic Shadows" option in the preferences panel does not seem to work. As far as I know, it should enable shadows in scenes that support them.

**To Reproduce**
1. Create a room with https://hubs.mozilla.com/scenes/zfWqEaF
2. Go to preferences and enable the dynamic shadows option
3. See that shadows do not appear under the door in the scene, even though they are enabled in Spoke

**Expected behavior**
The option should enable shadows if the scene supports them.

**Hardware** 
 - Device: Desktop
 - OS: Windows
 - Browser: Firefox

**Additional context**
The scene linked above is remixable, so you can verify and edit it in spoke if necessary.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-623)

takahirox:
Fixes #4164

This PR fixes a broken directional light shadow. (Other light types shadows seem to be fine.)

If I'm right, we should use `light.shadow.camera` instead of `light` for converting world matrix to shadow camera space matrix.

I haven't tested this change well other than with this scene https://hubs.mozilla.com/scenes/zfWqEaF I'm happy if anyone who has any scenes in which shadows are broken tests this change.

### DIFF
--- a/src/systems/shadow-system.js
+++ b/src/systems/shadow-system.js
@@ -1,7 +1,7 @@
 import { traverseAnimationTargets } from "../utils/three-utils";
 
 const frustumBox = new THREE.Box3();
-const inverseLightMatrixWorld = new THREE.Matrix4();
+const inverseShadowCameraMatrixWorld = new THREE.Matrix4();
 const tempBox = new THREE.Box3();
 const FRUSTUM_PADDING = 1;
 const NEAR_CLIPPING_PLANE = -500;
@@ -64,10 +64,12 @@ function resizeShadowCameraFrustum(light, boundingBox) {
   verts[7].set(max.x, max.y, max.z);
 
   light.updateMatrices();
-  inverseLightMatrixWorld.copy(light.matrixWorld).invert();
+  light.target.updateMatrices();
+  light.shadow.updateMatrices(light);
+  inverseShadowCameraMatrixWorld.copy(light.shadow.camera.matrixWorld).invert();
 
   for (let i = 0; i < verts.length; i++) {
-    verts[i].applyMatrix4(inverseLightMatrixWorld);
+    verts[i].applyMatrix4(inverseShadowCameraMatrixWorld);
     frustumBox.expandByPoint(verts[i]);
   }
 


src/systems/shadow-system.js
```diff
@@ -64,10 +64,11 @@ function resizeShadowCameraFrustum(light, boundingBox) {
   verts[7].set(max.x, max.y, max.z);
 
   light.updateMatrices();
-  inverseLightMatrixWorld.copy(light.matrixWorld).invert();
+  light.shadow.updateMatrices(light);
```
netpro2k, line 67: Hmm. Would you have to do this in upstream THREE? If not I bet its a result of our matrix dirty flags patch, so we should probably include fixing lights in that patch.

netpro2k (commented):
This stuff used to work, so I assume it regressed during one of the THREE upgrades. A bit surprising but these changes seems sane enough.

src/systems/shadow-system.js
takahirox: > Hmm. Would you have to do this in upstream THREE?

I don't think so, maybe. Shadow camera matrices are updated in `WebGLShadowMap` when rendering shadow depth to shadow map. It would be good for most of the use cases because generally light shadow camera matrices are not used in user code.

I guess the reason why the shadow camera matrices update is separated from light matrices update is, in addition to optimization, shadow camera matrices can be affected from `light.target` which can be anywhere in the scene graph (or can be out side of the scene graph). It doesn't greatly fit to the current `.updateMatrixWorld()` so I speculate Three.js wants to calculate shadow camera matrices after updateing the entire scene graph matrices.

> I bet its a result of our matrix dirty flags patch

I don't think so. Light matrices and shadow camera matrices are different. The latter but the latter taking account of looking at `light.target` but the former doesn't.

netpro2k: Ah. Then I am a bit confused why this is needed in Hubs but not upstream then. In GLTFLoader the lights target is added as a child and set up such that the light will be facing down its -z axis https://github.com/mrdoob/three.js/blob/dev/examples/jsm/loaders/GLTFLoader.js#L522

takahirox: Again, `light.target` can be anywhere in the scene graph in Three.js.

https://threejs.org/docs/#api/en/lights/DirectionalLight.target

Then Three.js wants to calculate shadow camera after the entire scene matrices are updated so it won't update in object's `.updateMatrixWorld()` but updates it in renderer after entire scene matrices are up to date. I don't think object `.updateMatrixWorld()` should update the shadow camera in the upstream.

netpro2k: Ah yes. What I kept missing is that we are reading the shadow camera's matrix here, so it needs to be up to date. So this is not about the shadow camera's matrix not getting computed correctly (THREE will update it correctly during render) but about *when* it is updated. In most cases you don't care about it being updated until it is about to be used during render. But in our case we care about it being updated here since we are using it to transform points to figure out the bounding box.

Issue #4861, opened 2021-11-26
[NodeJS 17 or later] “npm run dev”, “npm start”, and “npm run deploy” throw exception


esperecyan:
**Description**
When I run `npm run deploy` on NodeJS17, it throws an exception.

**To Reproduce**
1. Install NodeJS 17
1. Clone https://github.com/mozilla/hubs
1. `git checkout hubs-cloud` (52f293ed7f65329386adc50cbd192807166c83a5)
1. `npm ci`
1. `npm run login`
1. `npm run deploy`
1. See error

> PS ○○○○○○\hubs> npm run deploy
> 
> \> hubs@0.0.1 deploy
> \> node -r @babel/register -r esm -r ./scripts/shim scripts/deploy.js
> 
> Browserslist: caniuse-lite is outdated. Please run:
> npx browserslist@latest --update-db
> Deploying to exr-hubs.com.
> &nbsp;&nbsp;/ Building Client.node:child_process:397
> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;ex = new Error('Command failed: ' + cmd + '\n' + stderr);
> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;^
> 
> Error: Command failed: npm run build
> Browserslist: caniuse-lite is outdated. Please run:
> npx browserslist@latest --update-db
> node:internal/crypto/hash:67
> &nbsp;&nbsp;this[kHandle] = new _Hash(algorithm, xofLen);
> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;^
> 
> Error: error:0308010C:digital envelope routines::unsupported
> &nbsp;&nbsp;&nbsp;&nbsp;at new Hash (node:internal/crypto/hash:67:19)
> &nbsp;&nbsp;&nbsp;&nbsp;at Object.createHash (node:crypto:130:10)
> &nbsp;&nbsp;&nbsp;&nbsp;at module.exports (○○○○○○\hubs\node_modules\webpack\lib\util\createHash.js:135:53)
[...]

PR #5240, opened 2022-03-13
upgrade from webpack 4 to webpack 5

yakyouk:
This PR upgrades webpack to the latest version 5, along with some of the webpack plugins.

Aside from the advantages of wp 5 over wp 4, this solves an issue when npm packages use conditional chaining and nullish coalescing operator (one example is `screenfull 6`): wp 4 uses acorn 6 which does not support these in versions <7, so compilation fails.

Everything compiles well and seems to work. Compilation currently shows deprecation warnings that should be adressed, `process.traceDeprecation` is left on in webpack config to make these warnings verbose.

Process:
- Update `webpack.config.js` according to the first commit in the PR and run `npm i copy-webpack-plugin@6 html-webpack-plugin@4 webpack@4 webpack-bundle-analyzer@4 webpack-cli@4 webpack-dev-server@4`. Update the commands in `package.json` to be able to launch the local webpack server and verify that everything works.
- Update `webpack.config.js` according to the second commit in the PR and run `npm i copy-webpack-plugin@9 mini-css-extract-plugin@2 path-browserify@1 stream-browserify@3 webpack@5`.


esperecyan:
This Pull Request resolves https://github.com/mozilla/hubs/issues/4861 .

netpro2k:
Thanks for this PR. I used this as a basis for #5619, where I also upgraded all the dev dependencies and node version.

PR #5249, opened 2022-03-15
Upadate best practices doc

netpro2k:
Added a few more best practices to the doc and reorganized existing ones into some high level "themes".

New points:
- Prefer calling functions to emitting events
- Avoid defensive coding
- Prefer timestamps to timeouts
- Avoid using aframe's THREE property components
- Prefer storing Object3D references to aframe entities

### DIFF
--- a/doc/best-practices.md
+++ b/doc/best-practices.md
@@ -66,7 +66,7 @@ class Foo {
 
 ### Only store things on `this` that are actually persistent state.
 
-If you don't care about the state between frames and are only re-using it for [GC reasons](#dont-allocate-objects-every-frame))  use a closure or variable defined at the root of the file instead.
+If you don't care about the state between frames and are only re-using it for [GC reasons](#dont-allocate-objects-every-frame)  use a closure or variable defined at the root of the file instead.
 
 Don't:
 ```js
--- a/doc/best-practices.md
+++ b/doc/best-practices.md
@@ -1,10 +1,42 @@
 This is a set of random little best practices we have come to endorse over the years. They are examples of things that have often come up in PR reviews or patterns we have started to solidify on. There are plenty of counter examples of these in the codebase, they are probably mostly unintentional. Also, most of these are more suggestions than hard rules.
 
-These pertain mostly to the "in room experince" or "game" code. For best practices in the React UI code, see [UI Best Practices](ui-best-practices.md)
+These pertain mostly to the "in room experience" or "engine" code. For best practices in the React UI code, see [UI Best Practices](ui-best-practices.md)
+
+## Avoid Misdirection
+### if you have a setter/getter that just directly sets/gets a value, consider not having it instead
+
+This avoids an extra layer of indirection and makes the code easier to interpret. If the goal is "in case we need to do other things when this changes later", see [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it). Setters and getters also have added overhead in javascript vs raw values.
+
+Don't:
+```js
+class Foo {
+  constructor() {
+    this._val = 1;
+  }
+    
+  get val() {
+      return this._val;
+  }
+
+  set val(newVal) {
+      this._val = newVal;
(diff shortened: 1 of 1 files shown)


doc/best-practices.md
```diff
@@ -1,82 +1,72 @@
...
+    console.log(plusOne(3));
   }
 }
 ```
 
 
 ### Only store things on `this` that are actually persistent state.
 
-If you don't care about the state between frames and are only re-using it for GC reasons (see above point) use temp reusable variables, use a closure or variable defined at the root of the file instead.
+If you don't care about the state between frames and are only re-using it for [GC reasons](#dont-allocate-objects-every-frame))  use a closure or variable defined at the root of the file instead.
```
brianpeiris, line 69: Had an extra parentheses in there.
```suggestion
If you don't care about the state between frames and are only re-using it for [GC reasons](#dont-allocate-objects-every-frame)  use a closure or variable defined at the root of the file instead.
```

doc/best-practices.md
```diff
@@ -118,43 +108,95 @@ class Foo {
...
+### Avoid defensive coding
+
+Related to no failing silently; Do not guard against conditions that should not occur or indicate problems that are the responsibility of some other code. Adding unnecessary guards will force future readers of the code to have to try and figure out what cases that guard is intending to handle. Also, as with failing silently, unnecessary guards can lead to masking of serious issues.
+
+
+Don't:
+```js
+if(window.APP && window.APP.shouldAlwaysExist && window.APP.shouldAlwaysExist.doSomething) {
+  window.APP.shouldAlwaysExist.doSomething();
+}
```
brianpeiris, line 149: This reminded me of our increased use of optional chaining (`foo?.bar()`). Perhaps if we should watch out for that as a code smell too.

doc/best-practices.md
```diff
@@ -177,6 +219,90 @@ AFRAME.registerComponent("bad-example", {
 ```
 
 
-### Don’t fail silently.
+### Avoid using aframe's THREE property components
 
-Always at least log a warning or error to the console when something unexpected happens
+Do not use aframe's `position`, `rotation`, `scaale`, or `visibile` components, and instead just directly modify the Object3D properties. These add an additional layer of indirection and performance overhead.
```
brianpeiris, line 224:
```suggestion
Do not use aframe's `position`, `rotation`, `scale`, or `visibile` components, and instead just directly modify the Object3D properties. These add an additional layer of indirection and performance overhead.
```

brianpeiris (commented):
Minor comments and typos. LGTM.

doc/best-practices.md
keianhzo: Also `visibile`

keianhzo:
As someone that has fallen in a few of this mistakes I find this very useful and I hope this helps contributors. 

One thing that might make following this hard is that in our own code base there is a lot of old code that doesn't actually follow this guidelines and I guess that in many cases (at least that was my case when I started working on Hubs) people base their code in similar code parts more than they read this best practices guide so I feel that we are sending a confusing message of what we expect from contributions.

Not sure if spending some time in consolidating the code base to follow this guidelines would make sense though.

doc/best-practices.md
netpro2k: I agree, I actually don't love it because it makes this sort of defensive coding easier to do and easier to sneak in. I definitely have used it though, and I will admit it makes code that legitimately needs to be doing that sort of checking much less verbose

netpro2k:
Yeah. I think the most important thing is to try and follow best practices when introducing new code, or when reworking existing code. Going through old code and specifically reworking it though is also time well spent. I do think while doing that though is also a good time to then clean up the side effects and untangle the complexity that the old code may have been hiding as well.

takahirox:
Looking good to me

takahirox:
Any blocking issues? I want to add a new item to the best practice document and am waiting for this PR is merged.

Issue #5210, opened 2022-03-04
Improve shadow options in the preferences dialog


takahirox:
**Is your feature request related to a problem? Please describe.**

In the preferences there are separated options "Material quality" and "Enable Dynamic Shadow".

![image](https://user-images.githubusercontent.com/7637832/156835203-fef03b2f-63f1-4d75-8d60-5d01ab77ea27.png)

But Low material quality mode seems to force "Disable Dynamic Shadow" inside the Hubs source code.

This may be confusing to users because "Enable Dynamic Shadow" has no effect in case of low material quality mode.

**Describe the solution you'd like**

Two options in my mind.

1. Uncheck and disable "Enable Dynamic Shadow" if "Low material quality mode" is selected
2. Stop to force "Disable Dynamic Shadow" if "Low material quality mode" is selected, because "Enable Dynamic Shadow" option is separated from "Material quality mode" option in the preferences dialog.

**Additional context**

Forcing to disable dynamic shadow with low quality mode makes sense to me because mainly low quality mode is selected when devices are not powerful enough to run Hubs. What I want to mention is that the preference dialog for them a bit confusing.

PR #5250, opened 2022-03-16
Simplify the behavior of real-time shadows

### LINKED ISSUE #5212: Improve shadows preferences

Currently the shadows preferences and their behaviors may be confusing.

There is "Enable Dynamic Shadows" option in the preferences. But unchecking it doesn't disable shadows but updating shadow maps only once when environement (scene) is loaded.

This may be confusing to users. They may expect shadows are simply disabled (for the performance) if the option is unchecked. But even if updating the shadow maps only once shadow map textures are fetched and shadow depth is compared for rendering objects which receive shadows every frame, except for low quality materials. It affects the performance.

Simply disable shadows (`renderer.shadowMap.enabled = false`) if "Enable dynamic shadows` preference is unchecked. (And rename it to "Enable shadows"?)

The current behavior of disabled dynamic shadows seemed to be implemented when light map was not supported. Now we may be able to encourage to bake light or use light map instead.

**Additional context**

Maybe related in terms of improving the preferences menu for shadow #5210

takahirox:
Fixes #5212

This PR simplifies the behavior of real-time shadows preference.

Changes
* Rename "Dynamic shadows" to "Real-time shadows" in the label in the preference as suggested in https://github.com/mozilla/hubs/issues/5212#issuecomment-1064695862
* Simply enable real-time shadow (`renderer.shadowMap.enabled = true`) if "Enable Real-time shadows" checkbox is checked and disable it (`renderer.shadowMap.enabled = false`) if unchecked
* No longer page refresh is needed when checking/unchecking the checkbox

Benefits
* Probably the new behavior will be fit to uesrs' expectation
* Performance optimization. In the current behavior shadow map texture is fetched for rendering objects which can receive shadows every frame even if "Dynamic shadows" is disabled. In the new behavior, shadow map texture won't be fetched/used if "Real-time shadows" is disabled

Additional contexts
* This PR changes the behavior of disabled dynamic (real-time) shadows from updating shadow map only once when environment is loaded to simply disable shadows. We may be able to encourage to use light map instead.
* Most of assets don't support shadows (don't have Hubs shadow components) so that enabling real-time shadow doesn't have an effect in most of cases. It may be confusing to users. We may be able to consider to change the preference from "Enable real-time shadows" to "Force to disable real-time shadows" (later in another PR), it may be clearer.

### DIFF
--- a/src/systems/shadow-system.js
+++ b/src/systems/shadow-system.js
@@ -1,4 +1,4 @@
-import { traverseAnimationTargets } from "../utils/three-utils";
+import { updateMaterials } from "../utils/material-utils.js";
 
 const frustumBox = new THREE.Box3();
 const inverseShadowCameraMatrixWorld = new THREE.Matrix4();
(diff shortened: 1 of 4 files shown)


src/systems/shadow-system.js
```diff
@@ -90,46 +88,44 @@ function resizeShadowCameraFrustum(light, boundingBox) {
...
+      window.APP.store.state.preferences.enableDynamicShadows === true &&
+      window.APP.store.materialQualitySetting !== "low";
+    if (this.previousShadowsEnabled !== this.realtimeShadowsEnabled) {
+      this.sceneEl.setAttribute("shadow", { enabled: this.realtimeShadowsEnabled });
+
+      // If scene has already rendered, materials which can receive shadows must be updated.
+      if (this.sceneEl.hasLoaded) {
+        this.sceneEl.object3D.traverse(node => {
+          if (node.receiveShadow && node.material) {
+            const materials = Array.isArray(node.material) ? node.material : [node.material];
```
netpro2k, line 109: Fairly minor but in three-utils.js we have `updateMaterials` you could use here.

src/systems/shadow-system.js
```diff
@@ -90,46 +88,44 @@ function resizeShadowCameraFrustum(light, boundingBox) {
...
+    window.APP.store.addEventListener("statechanged", this.updatePrefs.bind(this));
+    this.updatePrefs();
   }
 
-  onEnvironmentSceneLoaded({ detail: environmentObject3D }) {
-    this.environmentObject3D = environmentObject3D;
-    this.needsUpdate = true;
-    this.sceneEl.renderer.shadowMap.autoUpdate = this.dynamicShadowsEnabled;
+  updatePrefs() {
+    this.realtimeShadowsEnabled =
```
netpro2k, line 99: It might be a bit confusing if you are in low material quality but check this box. Since we will not have it off by default maybe we should just directly follow the preference even in low quality? Another option would be to hide/disable the preference when in low quality mode.

netpro2k (commented):
Some minor notes but mostly LGTM. This definitely simplifies things quite a bit and I think will line up better with user expectations. It would be nice at some point to further optimize shadow rendering in general to update static and dynamic objects differently, but as you said, static objects will largely be taken care of now that we have lightmaps.

src/systems/shadow-system.js
takahirox: I feel the same. And I already opened an issue for it #5210
I don't really want to change the behavior of low quality mode in this PR. Let's make another PR if needed.

src/systems/shadow-system.js
takahirox: Updated, thanks

PR #5251, opened 2022-03-16
Upgrade Puppeteer to the latest

takahirox:
This PR upgrades [Puppeteer package](https://www.npmjs.com/package/puppeteer).

The current one uses a bit old Chromium Version 90.0.4403.0 (Developer Build) (64-bit) and it crashes when running `run-bot.js` on my Windows 10. Upgrading Puppeteer package resolved the problem.

Although others in the team couldn't reproduce the crash but I think upgrading it is safer. And using the newer Chromium would be more realistic for the tests. So I would like to suggest the Puppeteer package upgrade.

### DIFF
--- a/scripts/bot/package-lock.json
+++ b/scripts/bot/package-lock.json
@@ -1,8 +1,723 @@
 {
   "name": "hubs-bot",
   "version": "0.0.1",
-  "lockfileVersion": 1,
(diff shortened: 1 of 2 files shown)


brianpeiris (approved):
LGTM

PR #5257, opened 2022-03-17
Do not update the avatar preview component when unmounted

keianhzo:
When the Preview component is unmounted there might still be async operations ongoing and that's going to alter the component state after the operation resolves throwing a console error. This PR makes sure that we don't update the state after the component has been unmounted.

Screen Shot 2022-03-17 at 17 41 38


brianpeiris (approved):
LGTM. Thanks for cleaning this up.

PR #5260, opened 2022-03-18
Better first person head hiding

### LINKED ISSUE #1852: Muted Avatars Look Like They are Speaking in Selfie Camera
**Description**
Muted avatars look like their mics are being picked up in the selfie camera viewfinder. 

**To Reproduce**
1. Mute your mic
2. Open Up a selfie camera
3. Speak and see your avatar's head will change sizes

**Expected behavior**
I expect that the head wouldn't change size if my mic was muted at the time. I thought other people might be able to hear me even though I was muted. 

**Screenshots**
Two images showing head changing sizes with mute button on
Screen Shot 2019-10-22 at 1 48 25 PM
Screen Shot 2019-10-22 at 1 48 27 PM

**Hardware** 
 - Device: Desktop /  Phone
 - OS: Mac 10.14.6 / Android
 - Browser Firefox Nightly & Chrome / Chrome

**Additional context**

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-272)

netpro2k:
This reworks how we hide the avatar's head in first person mode. Instead of scaling down the head bone like we were doing previously, a new headless mesh is created and added to its now "first person only" layer. The original avatar mesh is put on a "third person only" layer. The various in game cameras are then adjusted to use the appropriate layers. This should fix the constant breaking of head scaling, head hiding, and just generally clean up code.

This is directly in line with how VRM avatars should work (they actually allow arbitrary "first person only" and "third person only" meshes) so it should be simple to swap once we do that (the code for creating the headless avatar is actually comes from three-vrm).

This PR also simplifies the behavior of scale-audio-feedback back to its original non distance dependent form, since that distance calculation was actually only correct for first person view and not the various in world cameras. Rather than make the logic more complicated I reverted it to the original intent of the component which was to allow the artist to specify this scale. With the incoming nametag changes you should not need to rely on avatar audio feedback to know who is speaking. This also fixes #1852 .

Lastly this PR slightly cleans up the code in camera-system but another pass will further be able to clean things up once we get rid of the built in AFRAME camera system.

Note: the `bone-visibility` component is still used for hiding avatar hands. We might consider doing something similar for them, but its less obviously correct, and feels out of scope for this PR.

TODO before merge:
- [ ] test mirror component
- [ ] test in VR


src/components/player-info.js
```diff
@@ -73,8 +75,20 @@ AFRAME.registerComponent("player-info", {
...
     deregisterComponentInstance(this, "player-info");
   },
+
+  onAvatarModelLoaded(e) {
+    this.applyProperties(e);
+
+    const modelEl = this.el.querySelector(".model");
+    if (this.isLocalPlayerInfo && e.target === modelEl) {
+      modelEl.object3D.traverse(function(o) {
+        if (o.isSkinnedMesh) o.parent.add(createHeadlessModelForSkinnedMesh(o));
```
keianhzo, line 85: We allow users to upload non rigged avatars and in that case we won't be hiding the mesh in first person mode.

src/components/mirror.js
```diff
@@ -48,15 +26,6 @@ AFRAME.registerComponent("mirror", {
       textureWidth: DEFAULT_TEXTURE_WIDTH,
       textureHeight: DEFAULT_TEXTURE_HEIGHT
     });
-    const { onBeforeRender } = reflector;
-
-    reflector.onBeforeRender = function(renderer, scene, camera) {
-      setPlayerHeadVisible(true);
```
keianhzo, line 54: I think we still need to hook on the onBeforeRender to set the third person layer in render target camera before rendering.

keianhzo (changes requested):
I'm very happy to remove so many hacks and adhoc code. There are a couple of things that I've seen.

src/components/player-info.js
netpro2k: Ah yes good catch. I guess in that case we should still put the whole mesh in the THIRD_PERSON_ONLY layer, but then have no FIRST_PERSON_ONLY mesh.

src/components/mirror.js
netpro2k: Hmm yeah. Ideally we should only need to do this once just to add the THIRD_PERSON_ONLY layer to the Reflector camera https://github.com/MozillaReality/three.js/blob/17375caff569ad48e1bef3e673e33f0bdd5d9bab/examples/js/objects/Reflector.js#L27 but the reference is not exposed so we will either need to modify Three to expose it or do some sort of workaround like this.

src/components/player-info.js
keianhzo: Yep I was thinking on something like that, if there was no headless mesh created just add it to the third person layer only.

src/components/mirror.js
keianhzo: Not sure about how expensive it's to hook into the rendering hooks vs setting the layers upfront but I like the former better than maintaining a patch.

Issue #5057, opened 2022-01-27
[Android] more then 20 people in the room, causese audio issue


juunini:
**Description**
- When more then 20 people in the room
  - Android device white noise sound.
- When more then 25 people in the room
  - Android device no sound.
  - video, audio and people's talking everything. all sound.

**To Reproduce**
1. More then 20 people in the room
2. Listen android device sound
  - my device is Galaxy S8, S10, Fold3 
3. More then 25 people in the room (or 30)
4. No sound to listen in android device

**Expected behavior**
- Can listen white noise (more 20)
- Can't listen any sound (more 25 or 30)

**Screenshots**
no screen shot. sound issue.

**Hardware** 
 - Device: Galaxy S8, Galaxy S10, Galaxy Fold
 - OS: Android
 - Browser: Chrome

Dayk0:
It's normal. Basically, hubs allows the possibility of having a maximum of 25 people in a single room. Depending on the device used (example: android phone) but also on the 3d environment in which you are, this can cause latencies.
For there to be 25 people without sound problems, the environment must be the most optimized with simple avatars but also a good connection for each user.

juunini:
@takahirox

1. Only android issue as far as I know. I'm not hear the white noise in non-Android device.
2. if Scene is more heavy, white noise early to hear when people exist 20.
  but, Scene is very light, I was hear the white noise people 28, 25 is in room and 3 is lobby.
3. Yes, low performance or old phone is early and loudly hear the white noise
4. I think the condition is when the device performs heavy calculations.
  The more people there are, the louder the white noise is.

takahirox:
@juunini Hi, thanks for the answers. Do you mind if checking the "Disable audio left/right panning" option checkbox in the preferences and checking if the problem can be mitigated? You can find the option in "More" - "Preferences" - "Audio" tab.

takahirox:
I tested and confirmed that the audio can be very choppy or even I hear nothing if people are many (about ten or more) in a room on my Samsung Galaxy S8+ + Chrome.

I realized that the problem can be worse if I move or rotate. So I speculate the problem can be related to Positional Audio (left/right panner), especially if CPU (or memory?) pressure is high. Actually if I check "Disable audio left/right panning" the audio problem is very mitigated.

If my speculation is true, potential solutions or workarounds may be

1. Force to disable Positional Audio on Android device or encourage to disable it if users face the audio problem
[...]

Issue #5278, opened 2022-03-23
Consider to use equalpower panning model


takahirox:
**Is your feature request related to a problem? Please describe.**

Three.js uses `HRTF` as [panning model](https://developer.mozilla.org/en-US/docs/Web/API/PannerNode/panningModel) by default. It is higher quality but probably more costly than `equalpower`.

**Describe the solution you'd like**

Consider to use `equalpower` as optimization. Some options how to use/switch.

1. Always use `equalpower`. It mighe be good enough. (We should test.)
2. Use `equalpower` only on (low-end) mobile devices
3. Use `equalpower` if the quality mode is low
4. Add preference to select panning model
5. Dynamically switch if pwerformance is bad, for example by watching the FPS counter

**Additional context**

Making benchmark test would help us determine whether we should use `equalpower` or not.

takahirox:
There are some documents that describe that `HRTF` is more expensive than `equalpower`.

* https://www.w3.org/TR/webaudio/#dom-panningmodeltype-equalpower
* https://developer.mozilla.org/en-US/docs/Web/API/PannerNode/panningModel
* https://padenot.github.io/web-audio-perf/#pannernode-when-panningmodel--hrtf

But I couldn't find any resource which compare the actual performance difference between them. And unclear how likely much gain we can get with `equalpower` in Hubs.

So I would like to start with adding "use equalpower panning mode" in the preferences. We can test in Hubs by switching it. If we get noticable performance benefit, we can keep the preference and/or introduce smarter optimization like dynamically switching it depending on the FPS number. If we don't see actual performance we may revert the change.

Any thoughts?

netpro2k:
From the tests I ran in https://github.com/mozilla/hubs/issues/5057#issuecomment-1074459540 equalpower definitely made a noticeable difference in perf on my Android device, especially in Firefox. The "web audio perf" doc you linked also suggests there is a pretty massive difference, calling `hrtf` "**Very** expensive" and `equalpower` "rather cheap"

Since we have had issues with audio on Android and were considering disabling panning altogether (#5291) I think just always using equalpower on Android might be a good idea (maybe all mobile devices?). This would probably also apply on Android based VR headsets which I don't think we have heard as many complaints about, so we might consider filtering those out. 

As you said, ultimately we probably want to decide this based on actual system performance instead of just device detection, but I think doing it based on device type would be a good start. 

[...]

PR #5291, opened 2022-03-24
Disable panner node for Android

takahirox:
Currently Android seems to have a panner node audio problem. With 10 or many Panner nodes (the number depends on Android OS version or device. The problem seems to happen more easily on older or lower-end devices) the audio can be crackled or broken. See #5057

We already reported it. Until it will be resolved on Android end, I would like to suggest to force to disable it as short-term workaround. The change is small. It may not be a perfect solution but it is a short-term workaround. I hope we can accept.

### DIFF
--- a/src/update-audio-settings.js
+++ b/src/update-audio-settings.js
@@ -5,6 +5,8 @@ import {
   AvatarAudioDefaults,
   TargetAudioDefaults
 } from "./components/audio-params";
+import isMobile from "./utils/is-mobile";
+import { isAndroid } from "./utils/detect-android";
 import { isSafari } from "./utils/detect-safari";
 
 const defaultSettingsForSourceType = Object.freeze(
@@ -38,6 +40,12 @@ export function getCurrentAudioSettings(el) {
   const preferencesOverrides = APP.store.state.preferences.disableLeftRightPanning
     ? { audioType: AudioType.Stereo }
     : {};
+  // Android has PannerNode audio problem. We already reported it.
+  // Until it will be fixed, we force to disable panner audio
+  // as short-term workaround. Disable only for Android phone
+  // because we haven't heard the problem on VR device so far.
+  // See #5057
+  const androidOverrides = isAndroid() && isMobile() ? { audioType: AudioType.Stereo } : {};
   const safariOverrides = isSafari() ? { audioType: AudioType.Stereo } : {};
   const settings = Object.assign(
     {},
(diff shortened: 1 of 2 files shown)


netpro2k:
We might try switching to `equalpower` as a less agressive workaround. I think losing positional audio is a pretty huge loss to the utility of Hubs. If we do end up doing this we should make sure it does not apply to android based VR headsets since I don't think we have had people report issues there, and positional audio is an even bigger loss there.

netpro2k:
Oops, did not mean to close the PR

takahirox:
> We might try switching to equalpower as a less agressive workaround.

Hm, from the test in #5057, `equalpower` just slightly mitigates the problem. But yeah, it should be less problematic. We might be able to try it first and see if we need other workarounds.

> we should make sure it does not apply to android based VR headsets since I don't think we have had people report issues there, and positional audio is an even bigger loss there.

Good point. If we force to disable it, we may be able to start to force only on Android phones.

brianpeiris:
I agree it's unfortunate we have to completely turn off positional audio for all android devices. I don't expect this to be fixed on Android or Chrome any time soon. Perhaps they will take months to fix it. If the fix ends up being in Android, then it's possible that many existing users with old phones may not receive the update for even longer, if at all.

If possible, perhaps we could disable positional audio on Android if there are more than 15 users in the room. That way users who often use Hubs with a small number of people will not be affected.

takahirox:
> If possible, perhaps we could disable positional audio on Android if there are more than 15 users in the room

At first I thought of that option but I started to feel I want to avoid because

1. The threshold number very depends on device. For example the number should be about 5-10 for my device.
2. The audio quality depends on the number of audio panner nodes so we also need to monitor the number of audio/video objects (any thing else?). The code can be complexed
3. Perhaps the problem will be fixed in months, I don't feel like to insert a complex code for short-term workaround

So I'm thinking of disabling on Android Phone devices regardless of the number of people. I may change my mind if the workaround code which disable positional audio depending on the number of used panner nodes, can be simple.

takahirox:
I have updated the PR to disable it only for Android Phones.

Some thoughts.

* Probably the problem will be fixed on the Android end sooner or later. I want very simple short-term workaround.
* Not disabling it for VR devices makes sense to me because we haven't heard the problem on VR devices (yet) and no positional audio may be big experience loss on VR devices.
* I'm thinking of introducing equalpower panning mode #5278 as optimization for low-end devices. But not as workaround for the Android panner audio problem because it seems to only slightly mitigate the problem.

takahirox:
Closing in favor of #5540

PR #5300, opened 2022-03-27
增加中文翻译

triston:


matthewbcool (changes requested):
Thank you for your contribution.  Could you add these translations to `zh-cn.json`? https://github.com/mozilla/hubs/blob/master/src/assets/locales/zh-cn.json

PR #5313, opened 2022-03-30
Fix handling of layers for VR cameras

### LINKED ISSUE #4553: [VR] Video-texture target components do not appear in VR
**Description**
Video-texture components are absent in scenes making use of this feature.

**To Reproduce**

1. Open a scene with a video source/texture on desktop and note position of video texture component. (https://hubs.mozilla.com/scenes/rUpFogH/galaxy-panel)
1. Open that scene with a video source/texture in VR  and note that the component is absent.

**Expected behavior**
Expect to see the video texture component.

**Screenshots**
image
inVR

**Hardware** 

* Device: [Oculus 2]
* Browser: [Oculus Browser]

**Additional context**
Component is visible in 2D, not available when immersed.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-721)

netpro2k:
This fixes a regression caused in #5260 where VR users could not see their own avatar. It also incidentally fixes #4553 and actually simplifies the code a bit, without losing support for single eye layers (useful if we want to do any stereoscopic effects).

This change should probably land in THREE instead, but putting it here for now since it hotfixes an issue and I am already actively refactoring this code in another branch.

### DIFF
--- a/src/components/layers.js
+++ b/src/components/layers.js
@@ -1,9 +1,8 @@
 export const Layers = {
   // These 3 layers are hardcoded in THREE
   CAMERA_LAYER_DEFAULT: 0,
-  // We disable support for these in camera-system to simplify dealing with layers
-  // CAMERA_LAYER_XR_LEFT_EYE: 1,
-  // CAMERA_LAYER_XR_RIGHT_EYE: 2,
+  CAMERA_LAYER_XR_LEFT_EYE: 1,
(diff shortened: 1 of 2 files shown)


cstfan:
Does it also fix the non-display of hands in VR?

netpro2k:
> Does it also fix the non-display of hands in VR?
Yes, fixing that regression is the primary goal

PR #5318, opened 2022-03-31
Refactor aframe startup and RAF loop

netpro2k:
This PR attempts to rework the way some of the early initialization happens in aframe. The main goal is to begin taking more direct control over the main RAF loop and start to untangle some incidental complexity around overriding many of the built in aframe systems. This first PR attempts not to change the code flow too much, but some of the async behavior around camera initialization was easier to simplify than temporarily maintain the old complexity.

Most of the interesting changes are in https://github.com/MozillaReality/aframe/pull/33, and comparing that to the changes in app.js here.

todo:
- [x] merge MozillaReality/aframe/pull/33
- [x] update package.json and package-lock.json to point at merged branch/commit

### DIFF
--- a/src/components/inject-main-camera-here.js
+++ b/src/components/inject-main-camera-here.js
@@ -1,4 +1,4 @@
-AFRAME.registerComponent("set-active-camera", {
+AFRAME.registerComponent("inject-main-camera-here", {
   init() {
     // This is what actually puts the camera into the scene graph.
     // Ideally we would just be doing this ourselves, but we have to do it after the entity we want to parent it to is created
--- a/src/components/set-active-camera.js
+++ b/src/components/set-active-camera.js
@@ -1,18 +1,8 @@
 AFRAME.registerComponent("set-active-camera", {
   init() {
-    // This is a hack needed to set the active camera because sometimes A-Frame will create the default camera and set it to the active camera on startup because the query selector to find an existing camera in the scene fails.
-
-    const cameraSystem = this.el.sceneEl.systems.camera;
-    const currentActiveCamera = cameraSystem.activeCameraEl;
-
-    if (currentActiveCamera !== this.el) {
-      cameraSystem.setActiveCamera(this.el);
-
-      // Also A-Frame fails to delete the default camera :P
-      currentActiveCamera.parentNode.removeChild(currentActiveCamera);
-
-      // Look controls leaves behind this CSS class.
-      this.el.sceneEl.canvas.classList.remove("a-grab-cursor");
-    }
+    // This is what actually puts the camera into the scene graph.
+    // Ideally we would just be doing this ourselves, but we have to do it after the entity we want to parent it to is created
+    // TODO any references to this entity should instead get the camera reference from the camera system
+    this.el.setObject3D("camera", this.el.sceneEl.camera);
   }
 });
--- a/src/components/spot-light.js
+++ b/src/components/spot-light.js
@@ -21,7 +21,6 @@ AFRAME.registerComponent("spot-light", {
     this.light.matrixNeedsUpdate = true;
     this.light.target.matrixNeedsUpdate = true;
     this.el.setObject3D("spot-light", this.light);
-    this.el.sceneEl.systems.light.registerLight(el);
   },
 
   update(prevData) {
--- a/src/components/spot-light.js
+++ b/src/components/spot-light.js
@@ -13,7 +13,6 @@ AFRAME.registerComponent("spot-light", {
   },
 
   init() {
-    const el = this.el;
     this.light = new THREE.SpotLight();
     this.light.position.set(0, 0, 0);
     this.light.target.position.set(0, 0, 1);
--- a/src/hub.html
+++ b/src/hub.html
@@ -55,8 +55,6 @@
     <div id="support-root"></div>
 
     <a-scene
-        disable-resize
-        embedded
         vr-mode-ui="enabled: false"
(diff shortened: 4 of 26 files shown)


src/components/inject-main-camera-here.js
```diff
@@ -0,0 +1,8 @@
+AFRAME.registerComponent("inject-main-camera-here", {
```
netpro2k, line 1: This is the thing I like least about this refactor, but its better than the hoops we were jumping through before. Fixing this probably requires moving the camera rig out of hub.html

src/components/set-active-camera.js
```diff
@@ -1,18 +0,0 @@
-AFRAME.registerComponent("set-active-camera", {
```
johnshaughnessy, line 1: :eyes: :wastebasket:

src/components/spot-light.js
```diff
@@ -13,15 +13,13 @@ AFRAME.registerComponent("spot-light", {
   },
 
   init() {
-    const el = this.el;
```
johnshaughnessy, line 16: :eyes: :wastebasket:

src/systems/camera-system.js
```diff
@@ -410,9 +401,9 @@ export class CameraSystem {
     const translation = new THREE.Matrix4();
     let uiRoot;
     return function tick(scene, dt) {
-      this.viewingCamera.object3DMap.camera.matrixNeedsUpdate = true;
-      this.viewingCamera.object3DMap.camera.updateMatrix();
-      this.viewingCamera.object3DMap.camera.updateMatrixWorld();
+      this.viewingCamera.matrixNeedsUpdate = true;
+      this.viewingCamera.updateMatrix();
+      this.viewingCamera.updateMatrixWorld();
```
johnshaughnessy, line 406: Interesting code here... I'm guessing I wrote this out of pure frustration

johnshaughnessy (approved):

src/hub.html
```diff
@@ -55,31 +55,20 @@
...
         local-audio-analyser
         class="grab-cursor"
-        renderer="
-            antialias: true;
-            colorManagement: true;
-            sortObjects: true;
-            physicallyCorrectLights: true;
-            alpha: false;
-            webgl2: true;
-            multiview: false;"
```
takahirox, line 74: Is it ok to remove renderer parameters? Some custom client devs might need to control them, for example turning off anti-aliasing for better performance. Or we will think how we allow them to control the renderer parameters later?

src/systems/camera-system.js
```diff
@@ -205,18 +205,41 @@ function getAudio(o) {
...
+
+    // xr.updateCamera gets called every render to copy the active cameras properties to the XR cameras. We also want to copy layers.
+    // TODO this logic should either be moved into THREE or removed when we ditch aframe camera system
+    const xrManager = renderer.xr;
+    const updateXRCamera = xrManager.updateCamera;
+    xrManager.updateCamera = function(camera) {
+      updateXRCamera(camera);
+      const xrCamera = xrManager.getCamera();
+      xrCamera.layers.mask = camera.layers.mask;
+      if (xrCamera.cameras.length) {
```
takahirox, line 232: Is it guaranteed that length equals to or is greater than 2 onless it is zero? If so, I think we need comment here for future readers.

takahirox (commented):
Decoupling from A-Frame sounds good to me for simpler design and better performance. And this change looks like a good first step.

I have just read through the changes. The changes basically look good although I addes some minor comments.

But I don't perfectly understand the inside of A-Frame and the change is a kind of relatively big. So I can't say whether this change has any problems or not. And reviewing has a limitation to perfectly check if the sort of big change in the core design doesn't cause any problems. Instead I feel like testing testing and testing for verification.

src/hub.html
netpro2k: We now create the renderer directly in app.js so they can modify these properties themselves, no hoops to jump through https://github.com/mozilla/hubs/pull/5318/files#diff-3d74dddefb6e35fbffe3c76ec0712d5c416352d9449e2fcc8210a9dee57dff67R35

src/systems/camera-system.js
netpro2k: Yeah this is internal state of XRManager https://github.com/MozillaReality/three.js/blob/hubs/master/src/renderers/webvr/WebXRManager.js#L47

This code should actually be applied as a threejs patch but I didn't want an additional thing to juggle in this PR

arpu:
any hints why this should bring  better performance?  i think the ideal way to do is to patch or update aframe direct?

netpro2k:
> any hints why this should bring better performance? i think the ideal way to do is to patch or update aframe direct?

This is not likely to make much difference in performance, its mostly just moving things around. Our aframe fork is already highly customized and we don't use/support most of the built in components and systems. As we continue to pair down our usage of aframe though we find ourselves constantly having to fight with some of thee built in systems and the fact that we are not the ones controlling execution order in the main render loop. This is a first in a series of PRs to "rip off the band aid" on some of these issues, and start a migration off aframe more earnestly (though it will likely still be quite some time before it is entirely removed).

src/systems/camera-system.js
netpro2k: Yeah I didn't want to change that in this PR. This camera system code is definitely worth a revisit in general. I think we should probably have multiple THREE.PerspectiveCamera instances rather than trying to mutate a single one and not clobber previous state.

PR #5324, opened 2022-04-02
Improved the chat emoji picker 👀 

### LINKED ISSUE #4250: Emoji search gets cleared when chat refreshes
**Description**
When using the emoji search box in chat, the search term gets wiped when a new message pops up in chat.

**To Reproduce**
1. Open the chat sidebar
2. Click the emoji icon
3. Type a search term in the emoji search
4. Wait...  for a change in the chat window (join/leave/chat, etc.)
5. The emoji search gets reset

**Expected behavior**
The emoji search should not get reset by chat events.

**Hardware** 
 - Device: Desktop
 - OS: Windows 10
 - Browser: Firefox

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBSOLD-645)

brianpeiris:
The emoji picker is now a bit nicer to use! You can hold down the `Shift` key to keep inserting emojis. The chat input also gets focused after picking an emoji, so that you can keep typing, and we've fixed an annoying bug that caused the emoji picker to reset itself when you received new chat messages 😤

hubs-chat-emoji

---

The core issue in #4250 is that emoji-mart, the emoji picker we use, does not seem to support re-renders well. It looses it search and filter state. I believe it's also quite heavy to load and render. So I used [React.memo](https://reactjs.org/docs/react-api.html#reactmemo) on `EmojiPickerPopoverButton` to avoid a re-render unless its `onSelectEmoji` prop changed.

While I was at it, I decided to improve the picker a bit more. Now, when you pick a single emoji, the chat input will be selected, using the existing `inputRef`, so that you can continue typing. Additionally, if you hold `Shift` down while picking emojis, the picker will stay open so that you can keep picking emojis. This was done using a [`useRef`](https://reactjs.org/docs/hooks-reference.html#useref) and keydown/up even handlers to track the current state of the Shift key. I also removed the implicit prop spreading (`{...props}`) from `Popover` to `EmojiPicker`, since we weren't actually using any of those props.

Fixes #4250

### DIFF
--- a/src/react-components/room/ChatSidebar.js
+++ b/src/react-components/room/ChatSidebar.js
@@ -1,4 +1,4 @@
-import React, { forwardRef } from "react";
+import React, { useEffect, useRef, forwardRef } from "react";
 import PropTypes from "prop-types";
 import classNames from "classnames";
 import { Sidebar } from "../sidebar/Sidebar";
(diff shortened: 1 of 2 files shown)


keianhzo (approved):
LGTM

PR #5335, opened 2022-04-05
Fix error on entry modal

keianhzo:
Sometimes the `react-use-css-breakpoint` hook tries to perform an update when the entry modal component has already been unmounted.

This has been fixed in the original library: https://github.com/matthewhall/react-use-css-breakpoints/pull/13
This PR just update the upstream version to use the fixed one


brianpeiris (approved):
LGTM

Issue #5290, opened 2022-03-24
Get rid of matrix-auto-update component


takahirox:
Related: #5294 

**Is your feature request related to a problem? Please describe.**

We turn off Three.js [`.matrixAutoUpdate`](https://threejs.org/docs/#api/en/core/Object3D.matrixAutoUpdate) by default. Instead we add dirty flag, we manually set it true when object's position/rotation/scale is updated, and update matrix only needed.

But we have `matrix-auto-update` component. `.matrixAutoUpdate` is set to true if an object of an entity having the `matrix-auto-update` attribute(component). The local matrices can be updated every frame even if they don't need to. If local matrices are updated, all the children world matrices need to be updated, too. It has a bad performance impact (I'm not sure how big it is tho).

Probably the component is for workaround. We want to let all the objects follow the dirty flag rule but we haven't perfectly handled all the objects under the rule yet.

**Describe the solution you'd like**

Get rid of `matrix-auto-update` component and handle all the objects properly under the dirty flag rule.

PR #5339, opened 2022-04-07
Remove matrix-auto-update component

takahirox:
Fixes: #5290 

Refer to the issue for details about why we need this change.

We may need more testing to check if nothing is broken before merging.


src/hub.html
```diff
@@ -121,14 +121,14 @@
                 </a-entity>
             </template>
             <template id="remote-avatar">
-                <a-entity tags="inspectable: true; preventAudioBoost: true" inspect-pivot-child-selector="selector:.inspect-pivot" networked-avatar ik-root player-info matrix-auto-update visible="false">
-                    <a-entity class="camera" matrix-auto-update>
+                <a-entity tags="inspectable: true; preventAudioBoost: true" inspect-pivot-child-selector="selector:.inspect-pivot" networked-avatar ik-root player-info visible="false">
```
netpro2k, line 124: This and all the other occurences in network templates are now having their dirty flags set by NAF.

src/hub.html
```diff
@@ -1166,14 +1161,14 @@
                 <a-entity in-world-hud visibility-while-frozen="visible: false; requireHoverOnNonMobile: false;" rotation="30 0 0" >
                     <a-rounded height="0.08" width="0.2" color="#000000" position="-0.44 -0.045 0" radius="0.040" opacity="0.35" class="hud bg"></a-rounded>
                     <a-image scale="0.07 0.06 0.06" position="-0.38 -0.005 0.001" src="#presence-count" material="alphaTest:0.1;"></a-image>
-                    <a-entity id="hud-presence-count" text="value: ; textAlign:center;" position="-0.30 -0.005 0.001" matrix-auto-update></a-entity>
+                    <a-entity id="hud-presence-count" text="value: ; textAlign:center;" position="-0.30 -0.005 0.001"></a-entity>
```
netpro2k, line 1164: Hmm I wonder why we had this on here. From what I can tell we never move this object directly.

src/hub.html
```diff
@@ -1166,14 +1161,14 @@
...
                     <a-entity class="invite-btn" position="0 0.18 0" scale="0.7 0.7 0.7" is-remote-hover-target tags="singleActionButton: true;">
                         <a-entity mixin="rounded-text-action-button" slice9="width: 0.7" scale="0.8 0.8 1.0">
                             <a-entity text=" value:invite; textAlign:center;" text-raycast-hack position="0 0 0.001"></a-entity>
                         </a-entity>
                     </a-entity>
                     <a-rounded visible="false" vr-mode-toggle-visibility id="hud-tooltip" height="0.08" width="0.4" color="#000000" position="-0.2 -0.2 0" rotation="-20 0 0" radius="0.025" opacity="0.35" class="hud bg">
-                        <a-entity text="value: ; textAlign:center; fontSize: 0.06;" position="0.2 0.04 0.001" matrix-auto-update></a-entity>
+                        <a-entity text="value: ; textAlign:center; fontSize: 0.06;" position="0.2 0.04 0.001"></a-entity>
```
netpro2k, line 1171: Same for this, but both seem to be text, so either there was an issue with the text component (now replaced by troika-text, so it might not be an issue anymore anyway) or it was just a copy-paste mistake.

src/hub.html
```diff
@@ -1264,7 +1259,6 @@
                 visibility-by-path="path: /actions/leftHand/matrix"
                 hand-controls2="left"
                 teleporter="start: /actions/leftHand/startTeleport; confirm: /actions/leftHand/stopTeleport; collisionEntities: [nav-mesh]"
-                matrix-auto-update
```
netpro2k, line 1267: **CursorPoseTrackingSystem** should be setting the dirty flag for this because it has the `track-pose` component.

src/hub.html
```diff
@@ -1278,7 +1272,6 @@
                 visibility-by-path="path: /actions/rightHand/matrix"
                 hand-controls2="right"
                 teleporter="start: /actions/rightHand/startTeleport; confirm: /actions/rightHand/stopTeleport; collisionEntities: [nav-mesh]"
-                matrix-auto-update
```
netpro2k, line 1281: Same as above, dirty flag set by **CursorPoseTrackingSystem**

src/hub.html
```diff
@@ -1338,7 +1331,7 @@
                 "
         ></a-entity>
 
-        <a-entity matrix-auto-update class="ui interactable-ui vr-notice" follow-in-fov="target: #avatar-pov-node; offset: 0 0 -0.8; angle: 39;" visible="false">
+        <a-entity class="ui interactable-ui vr-notice" follow-in-fov="target: #avatar-pov-node; offset: 0 0 -0.8; angle: 39;" visible="false">
```
netpro2k, line 1334: `follow-in-fov` component looks like it is setting the dirty flag.

src/hub.html
```diff
@@ -1356,7 +1349,7 @@
                 <a-entity text="value:View your desktop to continue; textAlign:center;" position="0 0 0.001"></a-entity>
             </a-entity>
         </a-entity>
-        <a-entity matrix-auto-update class="ui interactable-ui media-mirror" follow-in-fov="target: #avatar-pov-node; offset: 0 0 -0.8; angle: 39;" visible="false">
+        <a-entity class="ui interactable-ui media-mirror" follow-in-fov="target: #avatar-pov-node; offset: 0 0 -0.8; angle: 39;" visible="false">
```
netpro2k, line 1352: Same as above, dirty flag set by `follow-in-fov`

src/hub.html
```diff
@@ -1367,16 +1360,15 @@
                 ></a-entity>
             </a-entity>
         </a-entity>
-        <a-entity id="viewing-rig" set-yxz-order matrix-auto-update>
+        <a-entity id="viewing-rig" set-yxz-order>
```
netpro2k, line 1363: It looks like the dirty flag on this and the viewing-camera should be getting set by the Camera System, though the logic is a bit hard to follow so hard to say if there are edge cases this was capturing before.

netpro2k (commented):
I did a sanity check of all the places matrix-auto-update was being used and it seems like the dirty flag setting is now all accounted for so this LGTM. The only one that worries me slightly is the camera system one since its not super straightforward, but any issues there should be pretty obvious in testing so I suspect its fine.

src/hub.html
takahirox: Yes, the dirty flag for the viewing-camera is set every frame  in `camera-system`.

https://github.com/mozilla/hubs/blob/856a2c36efb54685867063c698aa280ac40b99b4/src/systems/camera-system.js#L404-L406

We need to optimize it and don't set if the viewing-camera doesn't move tho.

PR #5340, opened 2022-04-07
Update README.md

agriggs2:
Updated the version of Node as v12 is EOL May 2022. Node v14 is EOL May 2023.


brianpeiris:
Resolved in #5619

Issue #5322, opened 2022-04-01
Raise the dirty flag only if needed


takahirox:
**Is your feature request related to a problem? Please describe.**

The world matrices of a lot of objects in the scene graph seems to be updated every frame even if no object moves. These unnecessary world matrices update can cause the performance problem, especially on low-end devices.

I noticed that some or many of our Hubs components/systems update position/quaternion/scale/matrix of near root objects in their tick(). The problem is that they set the dirty flags even if the new values don't have diff with the current ones. Turning the dirty flag on invokes all children world matrices. This seems one of the causes of a lot of world matrices update.

**Describe the solution you'd like**

Raise the dirty flag in Hubs components/systems only if needed. For example compare the new values and the current ones, and if they are different raise the dirty flag.

**Additional context**

This additional check can be performance penalty for moving objects. But not all objects move every frame. Probably the benefit wins to the penalty.

takahirox:
Especially we need to be extra careful for turning on `.childrenNeedMatrixWorldUpdate` because if it's true all its children need to update their world matrices.

I locally tried to set it true only if the new world matrix value has diff from the current ones (and to remove `matrix-auto-update` component #5290). It cut off 80-90% of the number of world matrices in a certain frame in a room no object moves. Sounds like worth to try this optimization.

takahirox:
I made a PR for our Three.js repo https://github.com/MozillaReality/three.js/pull/64 
that we raise `childrenNeedMatrixWorldUpdate` flag if the new world matrix values have diff from the current values in `updateMatrices()` like.

```
if ( _m1.near( this.matrixWorld, _epsilon ) ) {
    this.matrixWorld.copy( _m1 );
} else {
    this.childrenNeedMatrixWorldUpdate = true;
}
```

Ideally we should do this check when updating position/quaternion/scale/local matrix/world matrix and raising the dirty flag in the Hubs components/systems instead. If we do that we can remove that costly check in `updateMatrices()` which runs for every world matrix update.

But there are many components and systems updating them, it may take time to optimize them. So I would like to suggest to do the check for world matrix update in `updateMatrices()` as short-term solution. And later we can optimize the Hubs components and systems, and remove the check in `updateMatrices()`.

PR #5341, opened 2022-04-07
Dirty flag performance optimization in setMatrixWorld()

takahirox:
Related: #5322
Requires: https://github.com/MozillaReality/three.js/pull/64

Similar to https://github.com/MozillaReality/three.js/pull/64, I would like to suggest to raise the `childrenNeedMatrixWorldUpdate` flag in `setMatrixWorld()` only if the new values have diff from the current values to avoid unnecessary matrices update.

### DIFF
--- a/src/utils/three-utils.js
+++ b/src/utils/three-utils.js
@@ -52,10 +52,13 @@ export function disposeNode(node) {
 }
 
 const IDENTITY = new THREE.Matrix4().identity();
+const tempMatrix4 = new THREE.Matrix4();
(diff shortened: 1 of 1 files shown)


netpro2k (commented):
LGTM

netpro2k (approved):
It definitely makes sense in the case of exact equality not to raise the `childrenNeedMatrixWolrdUpdate` flag, I am a bit more iffy on the "near" check but I guess it makes sense the floating point errors may creep in and we could end up flipping it for imperceptible changes. So this LGTM.

PR #5348, opened 2022-04-10
Refactor common webpack page pattern

brianpeiris:
This PR refactors our webpack config to extract the common pattern we use for registering our HTML page entry points. This change should not result in any functional difference.

In theory we could simplify things even further by removing the variations where we sometimes use the `chunksSortMode` and `inject` options, but I'll leave that for a future improvement, since it might change behavior that would require extra testing.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -190,6 +190,22 @@ async function fetchAppConfigAndEnvironmentVars() {
   return appConfig;
 }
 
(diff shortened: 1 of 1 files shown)


keianhzo (approved):
Thanks for simplifying this

brianpeiris:
Made a minor change after review to use filenames instead of bare names, so that we can still grep for "cloud.html" and find it in this file.

PR #5349, opened 2022-04-10
Test home page with browser stack

brianpeiris:
This is an experiment to test the home page in IE 11 using BrowserStack. The goal is to help test seldom-used browsers. In this case we add a basic test to load hubs.mozilla.com and ensure the home page renders in IE 11. The GitHub workflow added here runs on a weekly schedule.


keianhzo (commented):
I don't see the action being fired. This should have fired already, right?

brianpeiris:
Here's the list of runs for this branch: https://github.com/mozilla/hubs/actions/workflows/browser-stack.yml?query=branch%3Abrowser-stack-test

The latest run failed, but this is actually expected at the moment, since the app doesn't actually render in IE11 currently. That will have to be a separate fix.

brianpeiris:
The home page will be fixed in IE11 with this change: https://github.com/mozilla/reticulum/pull/589
Though we'll have to update this test to detect the "unsupported" message instead of the "create room" button.

keianhzo (approved):
Ok, I was expecting it as part of the PR actions but it makes sense that it's out of it.

PR #5350, opened 2022-04-11
Pared down unsupported browser message

brianpeiris:
Currently if you visit hubs.mozilla.com in IE 11, the app doesn't display anything at all. We'd like to at least display an unsupported message in legacy browsers. This PR pares down the unsupported message UI so that we don't have to spend the effort to target and polyfill old browsers. Unfortunately it means we can't support localization in this UI, but it's better than display nothing at all.

image

### DIFF
--- a/src/support.js
+++ b/src/support.js
@@ -1,19 +1,18 @@
+/* eslint-disable @calm/react-intl/missing-formatted-message */
+
 /*
 NOTE: support.js needs to be as self-contained as possible, since it needs to work in legacy browers
 that we do not support. Avoid adding imports to libraries or other modules from our own codebase
 that might break the support UI in legacy browsers.
+We also do not support localization in this file.
 */
 
 import React from "react";
 import ReactDOM from "react-dom";
 import copy from "copy-to-clipboard";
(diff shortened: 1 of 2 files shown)


src/support.js
```diff
@@ -82,75 +81,56 @@ class Support extends React.Component {
...
-              )}
-            </div>
-          </div>
-        </ThemeProvider>
-      </WrappedIntlProvider>
+      <div className={styles.supportMain}>
+        <div className={styles.supportContent}>
+          <h1>Mozilla Hubs</h1>
+          <br />
+          <p className={styles.unsupportedLangs}>
```
matthewbcool, line 88: `<span>不支援</span>` if we want to add Chinese here too.

matthewbcool (approved):
lgtm, great to have more feedback than we previously did. Thanks for adding this!

src/support.js
brianpeiris: Thanks!

brianpeiris:
This didn't end up working in IE 11 after I deployed it :( 
It seems IE 11 is borked for some unknown reason. I'll have to investigate later.

Issue #5287, opened 2022-03-24
Optimize Sprites system


takahirox:
**Is your feature request related to a problem? Please describe.**

Our Sprites system (src/systems/sprites.js) is used for rendering object menus. But its tick seems to be a bit costly. It consumes about 1ms every frame on my Samsung Galaxy S8+ (Android 9) Chrome even if object menus are not rendered, according to the profiler.

This 1024 * 2 loops every frame seems a bit costly.

https://github.com/mozilla/hubs/blob/6aaa86225a57b943838f332e01192a7a715ec165/src/systems/sprites.js#L294-L334

**Describe the solution you'd like**

Optimize Sprites system. Maybe tick can be skipped if object menus are inactive?

**Describe alternatives you've considered**

We may be able to use InstanceMesh or InstancedBufferGeometry, not sure if tick() can be faster with them tho.

PR #5360, opened 2022-04-13
Rewrite Sprites system for optimization

takahirox:
Fixes: #5287

This PR optimizes (and almost rewrites) Sprites system which is one of the slow systems in Hubs.

Changes:
1. Use simpler algorithm to manage sprite insertion and removal
2. Adopt draw range
3. Adopt buffer range update
4. Drop Multiview support
5. Clean up JavaScript and GLSL code

### DIFF
--- a/src/systems/sprites.js
+++ b/src/systems/sprites.js
@@ -7,47 +7,26 @@ import { createImageTexture } from "../utils/media-utils";
 import spritesheetActionPng from "../assets/images/spritesheets/sprite-system-action-spritesheet.png";
 import spritesheetNoticePng from "../assets/images/spritesheets/sprite-system-notice-spritesheet.png";
 import { waitForDOMContentLoaded } from "../utils/async-utils";
-import vert from "./sprites/sprite.vert";
-import frag from "./sprites/sprite.frag";
+import vertexShader from "./sprites/sprite.vert";
+import fragmentShader from "./sprites/sprite.frag";
 import { getThemeColorShifter } from "../utils/theme-sprites";
 import { disposeTexture } from "../utils/material-utils";
 
-const pngs = [[spritesheetActionPng, "action"], [spritesheetNoticePng, "notice"]];
+const MAX_SPRITES = 1024;
+const SHEET_TYPES = ["action", "notice"];
+const PNGS = {
+  action: spritesheetActionPng,
+  notice: spritesheetNoticePng
+};
+const ZEROS = new Array(16 * 4).fill(0);
 
-const multiviewVertPrefix = [
-  // GLSL 3.0 conversion
-  "#version 300 es",
-  "#define attribute in",
-  "#define varying out",
-  "#define texture2D texture",
-  "#extension GL_OVR_multiview : require",
-  "layout(num_views = 2) in;",
-  "uniform mat4 modelViewMatrix;",
-  "uniform mat4 modelViewMatrix2;",
-  "#define modelViewMatrix (gl_ViewID_OVR==0u?modelViewMatrix:modelViewMatrix2)",
-  "uniform mat4 projectionMatrix;",
-  "uniform mat4 projectionMatrix2;",
-  "#define projectionMatrix (gl_ViewID_OVR==0u?projectionMatrix:projectionMatrix2)",
-  ""
-].join("\n");
-const nonmultiviewVertPrefix = ["uniform mat4 modelViewMatrix;", "uniform mat4 projectionMatrix;", ""].join("\n");
+const getSheetType = sprite => (spritesheetAction.frames[sprite.data.name] ? "action" : "notice");
 
-const multiviewFragPrefix = [
-  "#version 300 es",
-  "#define varying in",
-  "out highp vec4 pc_fragColor;",
-  "#define gl_FragColor pc_fragColor",
-  "#define texture2D texture",
-  ""
-].join("\n");
-
-const nonmultiviewFragPrefix = "";
-
-function isVisible(o) {
+const isVisible = o => {
   if (!o.visible) return false;
   if (!o.parent) return true;
   return isVisible(o.parent);
-}
+};
 
 AFRAME.registerComponent("sprite", {
   schema: { name: { type: "string" } },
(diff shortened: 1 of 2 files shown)


src/systems/sprites.js
```diff
@@ -137,15 +110,15 @@ const raycastOnSprite = (function() {
   };
 })();
 
-function getHoverableVisuals(el) {
+const getHoverableVisuals = el => {
```
netpro2k, line 113: Doesn't actually matter but is there a reason you changed these? I tend to only use arrow functions when I need to actually bind `this` or use implicit return. No strong preference but we should probably come up with a rule for code consistency.

src/systems/sprites.js
```diff
@@ -156,29 +129,46 @@ function enableSweepingEffect(comp) {
   const isFrozen = hoverableVisuals.el.sceneEl.is("frozen");
   const hideDueToPinning = !isSpawner && isPinned && !isFrozen;
   return hoverableVisuals.data.enableSweepingEffect && !hideDueToPinning;
-}
+};
 
-function createGeometry(maxSprites) {
+const createGeometry = maxSprites => {
+  // TODO: Consider to use InstancedBufferGeometry which would be more efficient and performant
```
netpro2k, line 135: Yeah I think the sprite system predates that being added to Three (or possibly we just didn't know about it at the time). It definitely seems like the right way to handle this. We can probably tackle that whenever we actually get to reworking object menus, since that is by far the biggest usage of sprites.

src/systems/sprites.js
```diff
@@ -187,224 +177,234 @@ function createGeometry(maxSprites) {
     indices[i * 3 * 2 + 4] = i * 4 + 2;
     indices[i * 3 * 2 + 5] = i * 4 + 3;
   }
-  geometry.setIndex(indices);
-  return geometry;
-}
+  geometry.setIndex(new THREE.BufferAttribute(indices, 1));
 
-const ZEROS = [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0];
+  geometry.drawRange.count = 0;
```
netpro2k, line 182: Oh wow, we weren't setting this before? Very wasteful.

src/systems/sprites.js
```diff
@@ -187,224 +177,234 @@ function createGeometry(maxSprites) {
...
+    this.meshes = { action: null, notice: null };
+
+    waitForDOMContentLoaded().then(() => {
+      for (const type in PNGS) {
+        const spritesheetPng = PNGS[type];
+        createImageTexture(spritesheetPng, getThemeColorShifter(type)).then(spritesheetTexture => {
+          const geometry = createGeometry(MAX_SPRITES);
+          const material = createMaterial(spritesheetTexture);
+          const mesh = (this.meshes[type] = new THREE.Mesh(geometry, material));
+          const el = document.createElement("a-entity");
```
netpro2k, line 226: Minor since its just 2 entities total, but I wonder if we actually need this? We should be able to just add the Mesh to the scene directly.

src/systems/sprites.js
```diff
@@ -187,224 +177,234 @@ function createGeometry(maxSprites) {
...
+        });
       }
     });
 
-    APP.store.addEventListener("themechanged", async () => {
-      for (const [spritesheetPng, type] of pngs) {
+    APP.store.addEventListener("themechanged", () => {
+      for (const type in PNGS) {
+        const spritesheetPng = PNGS[type];
+        // TODO: Fix me if possible
```
netpro2k, line 240: Yeah. I would eventually like to rework this theming stuff such that sprites are colorless and can individually have a color property that is applied in the shader. We might then also use an SDF for the sprites themselves too, depending on if we end up needing sprites to support multiple tones.

src/systems/sprites.js
```diff
@@ -187,224 +177,234 @@ function createGeometry(maxSprites) {
...
       }
     });
   }
 
   tick(t) {
-    if (!this.meshes.action || !this.meshes.notice) return;
-
-    for (let i = 0; i < SHEET_TYPES.length; i++) {
-      const sheetType = SHEET_TYPES[i];
+    for (const sheetType of SHEET_TYPES) {
```
netpro2k, line 258: No clue if this is still the case, but at one point we observed these sort of loops created garbage so we tried to avoid them in hot code. Maybe just use a regular for just in case?

netpro2k (commented):
This LGTM, made a few minor notes. The only thing I would really consider changing before pushing is the for loop in tick, slightly concerned about creating garbage there.

src/systems/sprites.js
takahirox: Just for the consistency in the file (although still some `function`s are remained as inner functions). Yes, it's good if we have a rule.

src/systems/sprites.js
takahirox: > these sort of loops

Does it mean `for (const foo of array)`?

src/systems/sprites.js
takahirox: The current algorithm is more complex than it needs to be. The sprites array and the attribute arrays are sparse. So I guess `geometry.drawRange.count` can't be used (efficiently). With the new algorithm suggested in this PR is easier and the arrays will be contiguous so that we can set `geometry.drawRange.count`. A beauty of simplicity.

src/systems/sprites.js
netpro2k: Yep. for-of and for-in loops use iterators and may create garbage. Like I said not sure if this is still true and in which engines, but it was true at some point.

takahirox: OK, using old for loop sounds safer. Updated, thanks.

And it's good if it's written in the best practice document or somewhere else.

src/systems/sprites.js
takahirox: Does it mean

```
const mesh = (this.meshes[type] = new THREE.Mesh(geometry, material));
scene.object3D.add(mesh);
...
```

?

I just copied the block from the current code and I'm not sure if it's ok because I'm not familiar with A-Frame design.

Issue #4208, opened 2021-04-29
Admin pages frequently come up blank white for custom hubs cloud clients.


MegaMotion:
Bug can be erratic and does not affect everyone, but those affected see nothing but a blank white screen instead of admin pages, and very few to no messages in the browser logs. 

To reproduce: deploy a custom client, and you may or may not have this problem.

When active, the bug affects all browsers equally.

We experienced the issue from the point we installed our current stack several weeks ago all the way up until a couple of days ago, not immediately but a couple of days after installing the new hubs cloud client. At one point it just miraculously fixed itself, and worked for a day or two, but today it has gone back to a blank white screen, and several other users are reporting the problem now as well.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBSOLD-638)

blairmacintyre:
Ditto.  We just installed a custom client, seeing the same thing.  We had been running a custom client based on the previous client code, and ported our small changes to this.  Admin pages works fine without the custom client, and everything else in the client seems to work (spoke, rooms, etc).  Just the admin pages.  Signing out and back in doesn't help.

mattrossman:
In our case, the admin page is sending an empty Javascript bundle even though the expected 3 MB build file appears in `dist/` locally. Perhaps something is going wrong during the deploy script's upload process. After an `npm run undeploy`, the admin page becomes visible, it's just `npm run deploy` that doesn't work.

blairmacintyre:
FWIW, /hab/svc/hubs/var/dist/pages/admin.html  (Is this where it's served from?) is small.
```
app root@optimistic-troll:/hab/svc/hubs# ls -la var/dist/pages/
total 128
drwxr-xr-x 2 hab hab  4096 Apr 29 16:07 .
drwxr-xr-x 6 hab hab  4096 Apr 29 16:07 ..
-rw-r--r-- 1 hab hab  2274 Apr 29 16:07 admin.html
-rw-r--r-- 1 hab hab  2678 Apr 29 16:07 avatar.html
-rw-r--r-- 1 hab hab  2987 Apr 29 16:07 cloud.html
-rw-r--r-- 1 hab hab  3030 Apr 29 16:07 discord.html
-rw-r--r-- 1 hab hab 61479 Apr 29 16:07 hub.html
-rw-r--r-- 1 hab hab  1578 Apr 29 16:07 hub.service.js
[...]

PR #5363, opened 2022-04-14
Hack for unexpected ncp behaviour

rawnsley:
The is a temporary workaround to prevent deployment failures as discussed here https://github.com/mozilla/hubs/issues/4208#issuecomment-1006448926

The ncp command issues multiple callbacks during operation, which is unexpected and compromises the promise-based flow control.

Obviously this is only supposed to be a temporary solution. The full solution will probably require the selection of a new library to replace ncp and this more suitably tackled by the core dev team.

### DIFF
--- a/scripts/deploy.js
+++ b/scripts/deploy.js
@@ -101,6 +101,9 @@ const getTs = (() => {
   });
   step.text = "Preparing Deploy.";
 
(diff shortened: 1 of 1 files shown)


brianpeiris:
I've opened #5365, which should be a permanent fix.

rawnsley:
Perfect thanks.

PR #5365, opened 2022-04-16
Fix Hubs Cloud admin panel deploy by replacing ncp with fs-extra copy

brianpeiris:
As reported in #4208, the [`ncp`](https://www.npmjs.com/package/ncp) dependency we used to copy admin build files did not actually wait for the copy operations to complete, which caused the admin panel to break after the deploy. `ncp` has also been unmaintained for a while. This PR replaces `ncp` with [`fs-extra`](https://www.npmjs.com/package/fs-extra) which is relatively lightweight, maintained, battle-tested, and has a `copy` function that is a drop-in replacement for `ncp`.


FarhadG:
Confirmed this resolved my issue and removed the hack workaround on my end. Thank you, @brianpeiris

brianpeiris:
Thanks for verifying it!

keianhzo (approved):
LGTM 

Not critical but we still have ncp along with a number of hubs client dependencies locked in the admin package-lock.json that are outdated. We could probably work on aligning versions on both modules. I'll open an issue for that.

Issue #5388, opened 2022-04-26
scene.autoUpdate should be false in tock()


takahirox:
**Is your feature request related to a problem? Please describe.**

`video-texture-source` and `camera-tool` components call `renderer.render(scene, camera)` in their `tock()`. `scene.autoUpdate` is `true` so that the renderer tries to traverse the entire scene graph and update their matrices in every `renderer.render()`.

But the entire scene graph should be already updated in `tick()`. Thanks to our dirty flag optimization, the renderer doesn't unnecessarily recompute the matrices but still traverses the entire scene graph. It's unnecessary and costly.

**Describe the solution you'd like**

We may be able to set `scene.autoUpdate = false` in `tock()` like

```
const currentAutoUpdate = sceneEl.object3D.autoUpdate;
sceneEl.object3D.autoUpdate = false;
renderer.render(sceneEl.object3D, this.camera);
sceneEl.object3D.autoUpdate = currentAutoUpdate;
```

**Additional context**

Only the matrices which need to be updated in `tock()` may be the ones whose local matrices or parents world matrices are updated in other components/systems' `tock()`. But traversing the entire scene graph in each `renderer.render()` is inefficient. Instead we can make a rule either 1. If an object's matrix is updated in `tock()`, all its children should be updated, too, in the same `tock()` method to keep the entire scene graph up to date, or 2. Matrices shouldn't be updated in `tock()`.

takahirox:
Closing in favor of #5392 and #5400. And it would be better to add a tip, "`scene.autoUpdate` should be false", to our design document or somewhere. Or it can be set to `false` before our `tock()`s (this may need our A-Frame refactoring).

PR #5392, opened 2022-04-26
Set scene.autoUpdate false in video-texture-source.tock()

takahirox:
From: #5388

`video-texture-source.tick()` calls `renderer.render(scene, camera)`. The renderer attempts to update entire scene graph matrices due to `scene.autoUpdate = true`.

But the matrices should already be updated in `tick()`. Attempting to update the matrices again in `tock()` is unnecessary and costly.

`scene.autoUpdate` should be `false` in `tock()`.

I want to make another PR for `camera-tool.tock()` because it seems to update a matrix. I want to carefully review it.

### DIFF
--- a/src/components/video-texture-target.js
+++ b/src/components/video-texture-target.js
@@ -72,9 +72,16 @@ AFRAME.registerComponent("video-texture-source", {
     delete sceneEl.object3D.onAfterRender;
     renderer.xr.enabled = false;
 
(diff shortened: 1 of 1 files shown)


netpro2k (commented):
LGTM. Like with the other PR we may want to add a comment indicating why we are turning off matrixAutoUpdate

PR #5394, opened 2022-04-26
Cache the result of isSafari()

takahirox:
`detect-browser` seems to be slow. If panner audio is disabled it is called every frame via `isSafari()`.

image

Caching the result of `isSafari()` improves the Hubs Client performance.


netpro2k (commented):
This change LGTM. This reminds me that we have lots of different ways of doing browser checking that would be nice to consolidate some day.

PR #5400, opened 2022-04-29
Turn off scene.autoUpdate in camera-tool.tock()

takahirox:
From: #5388

The entire scene graph matrices should already be updated in `tick()` and traversing/recomputing them again in `tock()` is not necessary.

This commit turns it off and updates only needed matrices in `camera-tool.tock()`.

### DIFF
--- a/src/components/camera-tool.js
+++ b/src/components/camera-tool.js
@@ -584,6 +584,8 @@ AFRAME.registerComponent("camera-tool", {
       if (this.trackTarget) {
         if (this.trackTarget.parentNode) {
           this.lookAt(this.trackTarget);
+          // scene.autoUpdate will be false so explicitly update the world matrices
+          this.object3D.updateMatrixWorld();
         } else {
           this.trackTarget = null; // Target removed
         }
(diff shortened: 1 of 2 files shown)


src/components/camera-tool.js
```diff
@@ -620,12 +621,15 @@ AFRAME.registerComponent("camera-tool", {
...
           boneVisibilitySystem.tick();
+          boneVisibilitySystem.updateMatrices();
         }
 
         const tmpVRFlag = renderer.xr.enabled;
         const tmpOnAfterRender = sceneEl.object3D.onAfterRender;
+        const tmpAutoUpdate = sceneEl.object3D.autoUpdate;
         delete sceneEl.object3D.onAfterRender;
         renderer.xr.enabled = false;
+        sceneEl.object3D.autoUpdate = false;
```
netpro2k, line 632: Might be good to add a comment here explaining why we are turning off autoUpdate.

netpro2k (commented):
Might want to add a comment about this since its not super obvious. Other than that LGTM

src/components/camera-tool.js
takahirox: Added, thanks

PR #5410, opened 2022-05-04
Fix invalid material.side parameter set in troika-text

takahirox:
[Three.js `Material.side` takes `THREE.FrontSide/BackSide/DoubleSide` constants](https://threejs.org/docs/#api/en/materials/Material.side) but `troika-text` wrongly sets invalid strings `'front/back/double'`. Due to this problem, troika-text may not be rendered expectedly.

This PR fixes the problem by setting valid parameters.

### DIFF
--- a/src/components/troika-text.js
+++ b/src/components/troika-text.js
@@ -7,6 +7,12 @@ import { Text } from "troika-three-text";
 // Mark this type of object so we can filter in from our shader patching
 Text.prototype.isTroikaText = true;
 
+const THREE_SIDES = {
+  front: THREE.FrontSide,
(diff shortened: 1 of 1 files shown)


brianpeiris (approved):
LGTM

takahirox:
I merged this PR, but it seems still some components/systems may set invalid side parameters. If you run the following script in the console at the entrance of the smoke test, you will see some `material.side` has `undefined`.

```
document.getElementsByTagName('a-scene')[0].object3D.traverse(o => {
  if (o.material) {
    console.log(o.material.side)
  }
})
```

Issue #4779, opened 2021-10-26
Mediaframe race condition when a higher user number is present.


CodingBitsDev:
**Description**
There is currently a bug in media-frames.js which causes Every user to try to take ownership to a media-frame and change its data at once. This causes a reace condition when a sginificant ammount of users are present and lets every client spam the release function until enough clients leave the room.

**To Reproduce**
1. Enter the room with 25+ clients
2. Add an media Element to a media-frame
3. Delete the Media-Element
4. Watch the console beeing spamed with messages
5. The media frame should now switching between states of beeing blocked and beeing freed.

**Expected behavior**
It should not happen that many users are sending a new media-frame attribute at the same time

**Hardware** 
This was tested an many devices that mostly used Firefox and Chrome

**Additional context**
The error lies in the release function which is for example called when an element is removed and always tries to take ownership of the media-frame.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-1128)

matthewbcool:
https://user-images.githubusercontent.com/4493657/143784963-4ec43cbe-d0e5-4579-bf10-69cec982060c.mp4



![image](https://user-images.githubusercontent.com/4493657/143784802-ff0bce1b-b41f-4e05-a351-5a025dab4d95.png)

netpro2k:
This issue should be resolved by #5536

PR #5416, opened 2022-05-09
Fix media-frame release race condition

keianhzo:
This PR is based on this stalled community PR: https://github.com/mozilla/hubs/pull/4780

and Fixes https://github.com/mozilla/hubs/issues/4779

An issue that happens when multiple clients try to take ownership of a media-frame after it's been released. This PR limits the release only allowing the media frame owner to release it and also makes the releaser take ownership of the target object.


netpro2k:
The media frame system was rewritten in #5536 so this should no longer be an issue.

Issue #4951, opened 2021-12-15
[VR] The PDF pages are not turned


Softvision-RemusDranca:
**[Affected Versions]:**
- Oculus Browser

**[Affected Platforms]:**
- Oculus Quest

**[Prerequisites]:**
- Your Hubs Cloud stack is set up.
- Have a room opened.

**[Steps to reproduce]:**
1. Import a PDF file into the room.
2. Change the PDF pages and observe the behavior.

**[Expected results]:**
- The pages are turned.

**[Actual results]:**
- The pages number is changed but the file remains on the same page.

**[Notes]:**
- This issue is reproducible when trying to change to the previous or next page.

vincentfretin:
FYI the exact same issue I believe was reported to A-Frame repo https://github.com/aframevr/aframe/issues/4984
And #4181 may be related to this too.

vincentfretin:
#3968 may be related to this issue.

Renji3:
It seems that this is caused by functions added to window.requestAnimationFrame not beeing called anymore because the window is not redrawn in vr.
The PR [#5427](https://github.com/mozilla/hubs/pull/5427) should fix this

emclaren:
If we can implement a fix to this fairly quickly we should do this as P1, we may downgrade this to P2 if it requires a deeper dive.

Dayk0:
Hello, this bug is really annoying because PDFs are widely used for VR. Is there any news concerning the progress of the correction of this bug?

YasienKashef:
I will add myself into this thread because this bug is still as February 2023 present. In Hubs Cloud as well as the official Mozilla Hubs instances have this bug. It happens also on all VR devices so it doesnt depend which VR device is used.

I hope this bug fix will be taken faster in consideration as multiple clients as well as users cant use the PDF tool, which is very important especially if its used for introduction to the virtual room itself as well as to display controls and usability.
So most users are still hanging on the very first steps in the experience.

CodingBitsDev:
The proposed fix above solves the problem. 
It is currently uses aframe systems but that can be easily adapted to bitECS.

PlumCantaloupe:
> It seems that this is caused by functions added to window.requestAnimationFrame not beeing called anymore because the window is not redrawn in vr.
> This is however used by PDFjs to render the pdfs and resolve the page.render() promise in src/components/media-pdf.js
> The PR [#5427](https://github.com/mozilla/hubs/pull/5427) proposes a possible solution.

[...]

Issue #3968, opened 2021-02-26
[VR] PDFs don't load while in webxr on oculus browser


Sliferblade:
**Description**
While in WebXR on the oculus browser newly added pdfs wont load. The user must leave WebXR for the PDF.JS rendertask promise to complete.

**To Reproduce**
1. Enter hub on the oculus browser and enter in VR
2. Enter hub on desktop and load a PDF
3. PDF will load on desktop however wont on oculus browser
4. leave VR on oculus browser and PDF will load

**Expected behavior**
PDFs should load on oculus browser while in VR

**Hardware** 
 - Device: [Oculus Quest]
 - OS: []
 - Browser: [Oculus Browser]

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-558)

Softvision-RemusDranca:
I was also able to reproduce this issue using the same device and browser, the PDF file remains in a loading state.

After reentering in VR, the PDF is loaded, but when clicking the "Create" button for the file, a blank photo is created.

Exairnous:
This should be fixed now with the merge of https://github.com/Hubs-Foundation/hubs/pull/6575 but let us know if it reappears.

PR #5427, opened 2022-05-12
added fix system for pdfjs rendertask not resolving on oculus vr

Renji3:
The issues [#4951](https://github.com/mozilla/hubs/issues/4951) and [#3968](https://github.com/mozilla/hubs/issues/3968) talk about an issue with pdf rendering on the Oculus Browser in VR.

The following commit proposes a fix.

### DIFF
--- a/src/systems/hubs-systems.js
+++ b/src/systems/hubs-systems.js
@@ -28,6 +28,7 @@ import { AudioSettingsSystem } from "./audio-settings-system";
 import { AudioSystem } from "./audio-system";
 import { ShadowSystem } from "./shadow-system";
 import { MediaFramesSystem } from "./media-frames";
+import { MediaPDFOculusFix } from "./media-pdf-oculus-fix";
 import { InspectYourselfSystem } from "./inspect-yourself-system";
 import { EmojiSystem } from "./emoji-system";
 import { AudioZonesSystem } from "./audio-zones-system";
@@ -71,6 +72,7 @@ AFRAME.registerSystem("hubs-systems", {
     this.shadowSystem = new ShadowSystem(this.el);
     this.mediaFramesSystem = new MediaFramesSystem(this.physicsSystem, this.el.systems.interaction);
     this.inspectYourselfSystem = new InspectYourselfSystem();
+    this.mediaPDFOculusFix = new MediaPDFOculusFix();
     this.emojiSystem = new EmojiSystem(this.el);
     this.audioZonesSystem = new AudioZonesSystem();
     this.gainSystem = new GainSystem();
@@ -114,6 +116,7 @@ AFRAME.registerSystem("hubs-systems", {
     this.inspectYourselfSystem.tick(this.el, systems.userinput, this.cameraSystem);
     this.cameraSystem.tick(this.el, dt);
     this.waypointSystem.tick(t, dt);
+    this.mediaPDFOculusFix.tick(this.el)
     this.menuAnimationSystem.tick(t);
     this.spriteSystem.tick(t, dt);
     this.uvScrollSystem.tick(dt);
--- a/src/systems/media-pdf-oculus-fix.js
+++ b/src/systems/media-pdf-oculus-fix.js
@@ -1,0 +1,53 @@
+/*
+  PDFjs which is used to render pdfs depends on the window.requestAnimationFrame function.
+  This is not called when in mobileVR.
+  This system replaces the function with the requestAnimationFrame of the xrSession.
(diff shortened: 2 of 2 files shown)


src/systems/media-pdf-oculus-fix.js
```diff
@@ -0,0 +1,53 @@
...
+        cb();
+      };
+      this.cbList.add(myCb);
+      return xrSession.requestAnimationFrame.call(xrSession, myCb);
+    };
+    //Make sure tick is called
+    if (!this.interval)
+      this.interval = setInterval(() => {
+        this.tick(sceneEl);
+      }, 500);
```
vincentfretin, line 40: Not sure why this `setInterval` is needed, `this.mediaPDFOculusFix.tick(this.el)` should already be called right?

src/systems/media-pdf-oculus-fix.js
```diff
@@ -0,0 +1,53 @@
...
+    let isVR = !exit && sceneEl.is("vr-mode") && isMobileVR;
+    if (isVR != this.wasVR) {
+      if (isVR) {
+        this.onEnterVR(sceneEl);
+      } else if (this.wasVR) {
+        this.onExitVR();
+      }
+    }
+    this.wasVR = isVR;
+  }
```
vincentfretin, line 24: A-Frame has an enter-vr and exit-vr events so I was going to say to use addEventListener in constructor, but I guess you did it like this because you don't have the sceneEl is that constructor to add the listeners, that class is not an real aframe system.
Not sure why the isMobileVR is checked here, don't we have the pdf issue on any XR session anyway?

src/systems/hubs-systems.js
```diff
@@ -71,6 +72,7 @@ AFRAME.registerSystem("hubs-systems", {
     this.shadowSystem = new ShadowSystem(this.el);
     this.mediaFramesSystem = new MediaFramesSystem(this.physicsSystem, this.el.systems.interaction);
     this.inspectYourselfSystem = new InspectYourselfSystem();
+    this.mediaPDFOculusFix = new MediaPDFOculusFix();
```
vincentfretin, line 75: You actually can pass `this.el` here `new MediaPDFOculusFix(this.el)` so you have the sceneEl to add listeners.

src/systems/media-pdf-oculus-fix.js
```diff
@@ -0,0 +1,53 @@
...
+  This system replaces the function with the requestAnimationFrame of the xrSession.
+  It also adds a listener to ensure it is reset and call not called callbacks after the user leaves xr.
+*/
+export class MediaPDFOculusFix {
+  constructor() {
+    this.requestAnimationFramePrev = this.requestAnimationFramePrev || window.requestAnimationFrame;
+    this.cbList = new Set();
+  }
+
+  tick(sceneEl, exit) {
```
vincentfretin, line 13: You have a second param `exit` here that is apparently always undefined if you call `this.mediaPDFOculusFix.tick(this.el)`

src/systems/media-pdf-oculus-fix.js
vincentfretin: In some of the hubs systems, we actually pass `this.el` that is `sceneEl`.
So
```js
constructor(sceneEl) {
  this._sceneEl = sceneEl
  this._sceneEl.addEventListener('enter-vr', this.onEnterVR)
  this._sceneEl.addEventListener('exit-vr', this.onExitVR)
}
```
and removing tick altogether should probably work?

src/systems/media-pdf-oculus-fix.js
vincentfretin: That `setInterval` was maybe a previous fix before you patch `window.requestAnimationFrame`?
And that 500ms was to match the default minDurationToUpdateCanvas?
https://github.com/mozilla/pdf.js/blob/264046736d6cf3b726a111eb1461b01e56e774b2/web/base_pdf_page_view.js#L62C11-L62C36

DougReeder:
I think a workable solution is sketched out by @vincentfretin .  Do you have time to work on this, @CodingBitsDev ?  We'd love to have this working in VR.

DougReeder:
Closed in favor of a version cleaned-up along the lines discussed above: https://github.com/Hubs-Foundation/hubs/pull/6575

Exairnous:
Thank you for your work on this @CodingBitsDev.  Please feel free to keep track of, and comment on, PR https://github.com/Hubs-Foundation/hubs/pull/6575.

PR #5434, opened 2022-05-15
Implement VR mode fly button and add assets

msub2:
For context: I run monthly WebXR meetups in Hubs and almost every single time I find myself wanting to enable flight while in VR, which requires me to exit VR mode, open up chat, type "/fly" with the virtual keyboard on Quest, and then re-enter VR mode. This is frankly very cumbersome, so I decided to address it.

This PR adds a "Fly" button to the in-world HUD which allows the user to toggle flight mode on or off. It also includes new sprite assets and an updated system-notice spritesheet. The airplane icon is sourced from [here](https://commons.wikimedia.org/wiki/File:Plane_icon_nose_up.svg) and is in the public domain.


rawnsley:
Nice addition. It's worth mentioning that clicking the left joystick on the Quest will also toggle fly mode, but it's obscure and unfriendly to new users.

PR #5435, opened 2022-05-16
Support portable media frame scaling

wsxiaoys:


netpro2k:
Thanks for this PR. The media frame system was rewritten in #5536 and now supports scaling media frames!

PR #5436, opened 2022-05-16
Fix dark-mode logo. Adds an AppLogo component to unify logo usage.

brianpeiris:
- Added a new AppLogo component that unifies our use of the app logo.
- Fixed the dark-mode logo on hubs.mozilla.com (HMC). The root cause of this was the default value of "Browser Default" in `store.js`, which caused `getAppLogo` to return the wrong logo in dark mode.
- Refactored our use of `show_cloud` into a single helper function.
- Removed unused propTypes in `MobileNav.js`.
- Updated the fallback app-logo images to match the actual aspect ratio on the page.


matthewbcool (approved):
lgtm!

Issue #5295, opened 2022-03-26
Consider to resize textures for low-end devices


takahirox:
**Is your feature request related to a problem? Please describe.**

We uses MeshBasicMaterial based material if the material quality mode is low. But we reuse the original textures. If the original textures are huge, they can be slow to be rendered/fetched or consume memory too much on devices on which users want to use low quality mode.

**Describe the solution you'd like**

Consider to resize huge texture image sizes on the fly, for example to 512x512.

Note: Textures can't be resized if compressed textures?

**Describe alternatives you've considered**

Introduce LOD support. In low quality mode, we use only low levels. If the assets are created properly, probably low levels should use smaller textures.

takahirox:
I'm thinking how to expose the configuration. Some ideas.

1. Force to resize the textures on certain platforms. For example force to resize textures up to 512x512 on mobile devices
2. Bind to the material quality mode preference. For example resize textures up to 2048x2048 with medium mode and to 512x512 with low mode
3. Add a new preference "Maximum texture size"

I prefer 3, and the default value can be set on mobile devices, for example 512x512.

brianpeiris:
Option 3 with a default of 512x512 for mobile sounds like a good idea to me.

Some other things to consider:
- Should we do texture resizing for the entire scene, or only for objects in the scene?
- Should we avoid resizing textures on some objects? Like Image media, or PDFs? Maybe texture resizing will make some scenes look really bad?
- Should we do this for all media sources, or just ones that we know produce heavy objects (Sketchfab)
- Sketchfab already has an off-thread worker for unzipping the model. Should we do texture resizing inside that worker?
  - https://github.com/mozilla/hubs/blob/master/src/components/gltf-model-plus.js#L59
  - https://github.com/mozilla/hubs/blob/master/src/workers/sketchfab-zip.worker.js
- Should we avoid doing this on tablet devices that may have more VRAM?
- In the future, should we consider doing texture resizing as a backend service, so that it doesn't need to be done on each client device?

None of those are blockers though. I think we can benefit from this even if we do the simplest possible implementation, and then iterate on it as needed afterwards.

keianhzo:
I agree with all the comments above, just a couple of questions:
- How would this affect the loading times for big scenes with a lot of textures? I guess loading times can grow exponentially as we have to wait for all textures to be compressed in sequence or small batches?
- Can we cache the resized textures so we can skip the resizing step in future loads?
[...]

Issue #4669, opened 2021-09-24
[iOS] The page is refreshed when uploading certain objects from Sketchfab


Softvision-RemusDranca:
**[Affected Versions]:**
- Safari 15
- Chrome 93

**[Affected Platforms]:**
- Iphone SE2 

**[Prerequisites]:**
- A room is created and opened on https://hubs.mozilla.com/.

**[Steps to reproduce]:**
1. Tap on the "Place" button.
2. Click the "3D Model" button.
3. Select for example the "St Olaf" [object](https://sketchfab.com/3d-models/st-olaf-the-patron-saint-of-norway-342b6866ba7c46449392acbe27217aa2).
4. Observe the behavior.

**[Expected results]:**
- The object is spawned into the room.

**[Actual results]:**
- The Hubs page is refreshed.

**[Notes]:**
- Attached is a screen recording of the issue: [link](https://drive.google.com/file/d/15y3YdAaDLwjxREF8Pb-KMqHUK8ehYAaA/view?usp=sharing).

takahirox:
Thanks for the report. The model has 8192x8192 texture image. I guess it's too huge for iPhone SE2?

I first want to check WebGL stats on iPhone SE2. Would you mind if accessing https://webglreport.com/?v=2 on your iPhone SE2 and report "Max texture size" to me?

Softvision-RemusDranca:
@takahirox  The Max Texture Size on my Iphone SE2 is "16384".
I've encountered the same issue with the test page after refreshing the window (I did not encountered it when first accessing the page). At the first reload I reproduced the same issue and after some other page refreshes I got the message "A problem repeatedly occurred..". Here is a screenshot
![TestPage](https://user-images.githubusercontent.com/40061643/135082939-ed51b1d4-99f8-4404-8a64-2bb1dbfc02e6.jpeg)
:

Softvision-RemusDranca:
@takahirox I've encountered the same behaviour while changing scenes. Here is a [video](https://drive.google.com/file/d/1AF5VA1aa9NAKJmCQsQi4Z553S68GRBrd/view?usp=sharing). I reproduced it while having multiple tabs opened and changing the scene. Not sure, but could this be a safe mechanism on iOS, where when a tab consumes to many resources it will reload it? If it's the case it could be related to the high CPU usage issue...

takahirox:
@Softvision-RemusDranca

[...]

Issue #4771, opened 2021-10-23
The room is black on iOS in the old smoke test room


takahirox:
**Description**

The room is black on iOS in the old smoke test room.

**To Reproduce**

1. Go to the old smoke test room on iOS (Internally ask the team for the URL)
2. See the room is rendered as black

**Expected behavior**

The room is rendered correctly.

**Screenshots**

![df718560-10f5-4c4c-ad97-4ae76eae0772](https://user-images.githubusercontent.com/7637832/138536488-6af08032-3e6d-4a7b-af38-fd65ac7e38bd.png)

**Hardware** 
 - Device: iPhoneX
 - OS: iOS 15.0.2
 - Browser: Safari, Chrome

**Additional context**

I couldn't reproduce on Android. So maybe this is an iOS specific problem.
I couldn't reproduce in other scenes (yet). So maybe this issue is a scene specific problem.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBSOLD-1123)

takahirox:
Thank for the test. Hm, so the problem might be old or low-end model specific or latest iOS specific. Would you mind if upgrading your iOS and test again?

brianpeiris:
I cannot reproduce this anymore on my iPhone 11 with iOS 15.4.1 and Safari 15.4

takahirox:
I can no longer reproduce it neither on my iPhoneX iOS 15.4.1 Safari and Chrome 101.0.4951.44. I guess the problem has been fixed on iOS end?

Let's close this issue. Please reopen if someone still faces the problem.

PR #5437, opened 2022-05-16
Resize textures for iOS in low material quality mode.

takahirox:
From: #5295

Note: This change description is outdated. See https://github.com/mozilla/hubs/pull/5437#issuecomment-1168122622 and the following comments for update.

Resize huge textures up to 512x512 (Update: I changed to 1024x1024) for low material quality mode on `HubsMeshBasicMaterial` creation, for

* Better reliability on iOS devices. Currently browsers on iOS can crash with huge textures #4669
* Better performance. Users choose the low material quality mode for better performance. But fetching huge textures may be costly.
* Shorter main thread blocking time. It seems the main thread blocking time caused by texture image resize + smaller texture upload than the one caused by huge texture upload.

As written in https://github.com/mozilla/hubs/issues/5295#issuecomment-1126535213 I want to start with resizeing textures only for low material quality mode on `HubsMeshBasicMaterial` creation for simplicity and quickly fixing #4669. I hope we can think of further texture resizing optimization later if needed.

### DIFF
--- a/src/loaders/HubsTextureLoader.js
+++ b/src/loaders/HubsTextureLoader.js
@@ -1,10 +1,67 @@
+import { isIOS as detectIOS } from "../utils/is-mobile";
+
+const isIOS = detectIOS();
+
 // Disable ImageBitmap for Firefox so far because
 // createImageBitmap() with option fails on Firefox due to the bug.
 // Three.js ImageBitmapLoader passes {colorSpaceConversion: 'none'} option.
 // Without the option, the rendering result can be wrong if image file has ICC profiles.
 // See https://github.com/mrdoob/three.js/pull/21336
-const HAS_IMAGE_BITMAP = window.createImageBitmap !== undefined && /Firefox/.test(navigator.userAgent) === false;
+const HAS_IMAGE_BITMAP =
+  window.createImageBitmap !== undefined && /Firefox/.test(navigator.userAgent) === false && !isIOS;
 export const TEXTURES_FLIP_Y = !HAS_IMAGE_BITMAP;
+
+// Browsers on iOS devices can crash if huge or many textures are used in a room.
+// Perhaps huge VRAM usage may cause it. We already reported it to the webkit devs.
+// Until the root issue is fixed on iOS end, we resize huge texture images in the
+// low material quality mode as workaround. It isn't a perfect solution but it
+// should mitigate the problem. Also disable image bitmap (See the above)
+// because huge textures resize + image bitmap can still even crash browsers.
+//
+// See
+//   - https://github.com/mozilla/hubs/issues/5295
+//   - https://bugs.webkit.org/show_bug.cgi?id=241478
+
+const IOS_TEXTURE_MAX_WIDTH = 2048;
+const IOS_TEXTURE_MAX_HEIGHT = 2048;
+
+function resizeIfNeeded(texture) {
+  if (!isIOS || window.APP.store.state.preferences.materialQualitySetting !== "low") {
+    return texture;
+  }
+
+  const image = texture.image;
+
+  // Just in case.
+  if (!image || image.width === undefined || image.height === undefined) {
+    return texture;
+  }
+
+  if (image.width <= IOS_TEXTURE_MAX_WIDTH && image.height <= IOS_TEXTURE_MAX_HEIGHT) {
+    return texture;
+  }
+
+  const conversionRatio = Math.min(IOS_TEXTURE_MAX_WIDTH / image.width, IOS_TEXTURE_MAX_HEIGHT / image.height);
+
+  const newWidth = Math.round(image.width * conversionRatio);
+  const newHeight = Math.round(image.height * conversionRatio);
+
+  const canvas = document.createElement("canvas");
+  canvas.width = newWidth;
+  canvas.height = newHeight;
+  const context = canvas.getContext("2d");
+  context.drawImage(image, 0, 0, newWidth, newHeight);
+
+  texture.image = canvas;
+
+  if (image.close !== undefined) {
+    image.close();
+  }
+
+  texture.format = THREE.RGBAFormat;
+  texture.type = THREE.UnsignedByteType;
+  return texture;
+}
 
 export default class HubsTextureLoader extends THREE.TextureLoader {
   constructor(manager) {
@@ -12,9 +69,16 @@ export default class HubsTextureLoader extends THREE.TextureLoader {
   }
 
   load(url, onLoad, onProgress, onError) {
+    const callback = texture => {
+      resizeIfNeeded(texture);
+      if (onLoad) {
+        onLoad(texture);
+      }
+    };
+
     const texture = HAS_IMAGE_BITMAP
-      ? this.loadImageBitmapTexture(url, onLoad, onProgress, onError)
-      : super.load(url, onLoad, onProgress, onError);
+      ? this._loadImageBitmapTexture(url, callback, onProgress, onError)
+      : super.load(url, callback, onProgress, onError);
 
     texture.onUpdate = function() {
       // Delete texture data once it has been uploaded to the GPU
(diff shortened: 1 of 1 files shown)


takahirox:
Visual quality with resized 512x512 textures looks worse than I expected. Perhaps it isn't acceptable even with low material quality mode.

image

image

Ideally it's the best if we come up with a technique that can resize the textures while keeping the quality. (We need smoothing and/or uv adjustment?)

But what I really want to fix is #4669 , I want to think of better texture resizing later if needed.

Instead I'm thinking of relaxing the maximum texture size for the low quality mode, like from 512x512 (all textures must be small enough) to 1024x1024 or 2048x2048 (too huge textures must be avoided).

takahirox:
Changed from 512x512 to 1024x1024

image

Looks ok so far in this scene (tho I'm not really sure what textures in the scene are resized).

takahirox:
We internally discussed in the team. Resized textures visual quality may not be acceptable especially if scene authors tune well and it's much worse than author's expectation.

But we also agree that assets from Sketchfab may not be optimized.

So we are thinking of resizing textures only the assets imported from Sketchfab so far.

takahirox:
In today's team meeting in Hubs, I encountered the unexpected page refresh problem on my iPhoneX.

Probably no assets is imported from Sketchfab. And I can enter the room if no one there. I speculate some avatars were not optimized well.

So... I'm thinking what we should do for the problem. Ideally we create a simple test reproducing the problem and push the iOS or Safari team to fix the problem. Until the problem will be fixed on iOS or Safari end, what should the workaround be?

src/utils/material-utils.js
```diff
@@ -240,5 +255,68 @@ export function disposeTexture(texture) {
...
+  if (!image || image.width === undefined || image.height === undefined) {
+    return texture;
+  }
+
+  if (image.width <= width && image.height <= height) {
+    return texture;
+  }
+
+  const newWidth = Math.min(width, image.width);
+  const newHeight = Math.min(height, image.height);
```
wsxiaoys, line 294: Consider keep the aspectRatio here based on width / height?

Use case: we reused this function to resize some large media-image to avoid crash the safari.

takahirox: Sounds like a good idea. Probably I will do it if we introduce this feature, thanks for suggestion.

takahirox:
Let me switch this PR to draft for now because I started to worry if this feature or configuration is good because

* The resized texture visual quality is worse than my expectation
* Even if textures are resized only for assets from Sketchfab the problem can still happen
* This PR forces the texture resize in low material quality mode. But we might want to think of adding a new preference max texture size.

takahirox:
Rethinking the issue because I encountered the problem on an iOS device again...

I'm thinking if we can resize huge textures to 2K textures in the low material quality mode only on iOS devices (regardless of whether assets from Sketchfab or not).

Some thoughts.
* Crashing problem on iOS is a serious problem. For example if someone places an asset with huge textures all the iOS devices in the room can crash. We may need a workaround.
* We already reported the root issue to webkit dev. We may be able to expect that the problem can be fixed on webkit end. The workaround will be removed sooner or later. So I don't want to add a complex workaround.
* Not distinguishing assets whether from Sketchfab or not can make the workaround simpler.
* Resizing to 2K textures may not be a perfect solution. But it should metigate the problem. It would be good as workaround.
* Resizing textures can affect visual quality. But (I hope) 2K may be big enough. And doing it only on iOS devices can limit the devices affected by the workaround.
* At least, a bit worse visual quality should be better than broken app.

wsxiaoys:
Adding a few cases this PR won't solve (but still causing problem on mobile iOS) for consideration:
1. Large resolution image implemented by a-image still causes crash.
2. Large resolution pdf will not render (black plane)

takahirox:
Then resizing textures in `HubsTextureLoader` may be better? I assume all textures are loaded via it, but not sure if it's true.

src/utils/material-utils.js
takahirox: Done

takahirox:
Updated the PR to force to resize huge textures to 2K textures in `HubsTextureLoader` in the low material quality mode on iOS devices.

src/loaders/HubsTextureLoader.js
```diff
@@ -1,20 +1,84 @@
...
+  if (image.width <= IOS_TEXTURE_MAX_WIDTH && image.height <= IOS_TEXTURE_MAX_HEIGHT) {
+    return texture;
+  }
+
+  const conversionRatio = Math.min(IOS_TEXTURE_MAX_WIDTH / image.width, IOS_TEXTURE_MAX_HEIGHT / image.height);
+
+  const newWidth = Math.round(image.width * conversionRatio);
+  const newHeight = Math.round(image.height * conversionRatio);
+
+  const canvas = document.createElement("canvas");
```
netpro2k, line 49: We may want to reuse this cavnas/context and just resize it instead of making anew one for each texture. (I noticed the code in ThreeJS for resizing textures does this).

netpro2k (commented):
Still not in love with doing this but like you said I guess its better than the app crashing... I think the issue mentioned in the thread about `a-image` no longer applies since that has since been removed from our aframe, so everything should be going through our texture loader. 

While reviewing I dug into ThreeJS since I remember that it had texture resizing code. It does resize if textures are `> gl.MAX_TEXTURE_SIZE`. Another options for doing this patch would be to override **WebGLCapabilities** in our fork to report a lower `MAX_TEXTURE_SIZE` on iOS. Not sure if I like that solution better but figured i would mention it as an option. 

I also wonder what the typical `MAX_TEXTURE_SIZE` is on an iOS device, since we would have already been resizing anything larger than that anyway.

takahirox:
> I also wonder what the typical MAX_TEXTURE_SIZE is on an iOS device,

16K as far as I know

takahirox:
> I guess its better than the app crashing...

Yes, it should be much better. Actually I rarely be able to join our meeting and meetup rooms on my iOS devices due to this problem. And 2K may be big enouch. It may not affect many assets.

> While reviewing I dug into ThreeJS since I remember that it had texture resizing code. It does resize if textures are `> gl.MAX_TEXTURE_SIZE`. Another options for doing this patch would be to override **WebGLCapabilities** in our fork to report a lower `MAX_TEXTURE_SIZE` on iOS.

Hm, honestly I don't really agree with that option. (But thanks for mentioning the option anyways.)  I don't think we should really rely on Three.js internal implementation. It's an internal design. The behavior might be able to change in future revisions without any notice (internal change may not appear in the migration guide.) Overriding capabilities might have similar risk. In the future revision it might cause unexpected side effects.

Resizing image code is just about 50 lines. It may not be a big deal. I think having the 50 lines in our code is less problem and better maintainability.

If we want to avoid duplicated code with Three.js core, the best option may suggest to Three.js upstream for exposing the resize dunction.

src/loaders/HubsTextureLoader.js
takahirox: `canvas` is set to `texture.image`. If we reuse `canvas`, doesn't the update affect all the textures having resized images?

wsxiaoys:
> Still not in love with doing this but like you said I guess its better than the app crashing... I think the issue mentioned in the thread about `a-image` no longer applies since that has since been removed from our aframe, so everything should be going through our texture loader.

Great to know a-image is removed!

while it seems [media-pdf](https://github.com/mozilla/hubs/blob/master/src/components/media-pdf.js#L90) is still loading texture by its own. It won't cause crash on iOS device, but the content become black due to large dimension

takahirox:
If I remember correctly, the black texture problem was caused on iOS image bitmap. But I heard it has been fixed in the latest iOS. Do you use the latest iOS? (I have bad memory. Sorry if I'm wrong.)

wsxiaoys:
I'm on iPhone 12 Pro Max 15.5.

I haven't tried disabling image bitmap, but this MR fixed the pdf rendering issue for me on an 1920x1080 resolution pdf: https://github.com/mozilla/hubs/pull/5554

takahirox:
Sounds like the same issue as

* https://github.com/mozilla/hubs/issues/4771
* https://bugs.webkit.org/show_bug.cgi?id=232357

But the root issue seems to have been fixed on iOS end. If you can still reproduce the problem, first we should report to the iOS team.

Can you make and share a room that you can reproduce the problem?

PR #5438, opened 2022-05-17
Move fog handling to environment system

### LINKED ISSUE #5428: Fog no longer working in Mozilla Hubs
**Description**
Yesterday I noticed that fog doesn't seem to be working anymore in Mozilla Hubs. I tested it out in my own custom room and in rooms that others have made that I found online, and in both cases fog wasn't showing up. It works in Spoke, but not in the actual room viewer. I have tried to view rooms on multiple different devices, but no matter how I access it, the fog is gone.

**To Reproduce**
To see what I am talking about go to any room that has fog. For example, this foggy lake: https://hubs.mozilla.com/scenes/RAYtAUO/foggy-lake
In the preview of the room the fog is gone and when you create a room with the scene the fog disappears.
Oddly, when you hit "Remix in Spoke" the fog appears again.

**Expected behavior**
I expect the fog to appear more or less the same in spoke as in the actual room, but the fog doesn't seem to be working at all.

**Screenshots**
Screen Shot 2022-05-12 at 4 40 38 PM
Screen Shot 2022-05-12 at 4 41 11 PM

**Hardware**
 - Device: Multiple different Macbook Air Laptops
 - OS: Apple macOS Big Sur Version 11.6
 - Browser: Chrome

netpro2k:
#5361 removed the fog component but we were still relying on it. This PR moves the handling of fog into the environment system. 

For reference the previous behavior was implemented in (now removed) https://github.com/MozillaReality/aframe/blob/0cd5cac593a1000d59895aedfac38068060be04b/src/components/scene/fog.js

fixes #5428

### DIFF
--- a/src/systems/environment-system.js
+++ b/src/systems/environment-system.js
@@ -34,7 +34,13 @@ const defaultEnvSettings = {
   physicallyCorrectLights: true,
   envMapTexture: null,
   backgroundTexture: null,
-  backgroundColor: new THREE.Color("#000000")
+  backgroundColor: new THREE.Color("#000000"),
+
+  fogType: null,
+  fogColor: new THREE.Color("#000000"),
+  fogDensity: 0.00025,
+  fogFar: 1000,
+  fogNear: 1
 };
 
 let blenderLUTPromise; // lazy loaded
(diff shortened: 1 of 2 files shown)


src/systems/environment-system.js
```diff
@@ -34,7 +34,13 @@ const defaultEnvSettings = {
   physicallyCorrectLights: true,
   envMapTexture: null,
   backgroundTexture: null,
-  backgroundColor: new THREE.Color("#000000")
+  backgroundColor: new THREE.Color("#000000"),
```
keianhzo, line 37: We are using different default values in Spoke, Blender and here. We should decide reasonable defaults and unify for consistency:

https://github.com/mozilla/Spoke/blob/d98a535b948a3d144290b87d459a39e978683d7f/src/editor/nodes/SceneNode.js#L151-L155

Maybe #FFFFFF and a default density of 0.00025.

keianhzo (approved):
LGTM Just a nit regarding the defaults

src/systems/environment-system.js
netpro2k: Ah these defaults were from the fog component in aframe. I will pull what we have in Blender.

netpro2k: Actually yeah I think the `0.1` default we have in blender is probably way too high. Will use `0.00025` and we can update that in Blender as well.

PR #5441, opened 2022-05-18
Fix share screen toggle.

wsxiaoys:
There's a typo in screen sharing toggle caused the issue.


matthewbcool:
nice catch! Thank you!

takahirox (approved):
Good catch, thanks.

PR #5447, opened 2022-05-19
Rename certs to RootCA.*; Doc the existence of the generated RootCAs.

knev:
Make it easier for the user of the repo to know RootCAs are generated and how they can be used to create a self-signed certificate for `hubs.local`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,12 @@ The backend dev server is configured with CORS to only accept connections from "
 127.0.0.1	hubs-proxy.local
 ```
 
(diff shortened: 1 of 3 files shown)


knev:
Note: if this PR is accepted and people pull it; those that have used the certs before to create a self-signed certificate will have to rename their certs to `RootCA.*`
Otherwise: since the code detects the existence of the `certs` directory and not the name of the files an error will be generated instead of new certs being generated.

Issue #5431, opened 2022-05-13
Visual quality can be very bad if you switch from portrait mode to landscape mode on mobile devices


takahirox:
**Description**

Visual quality can be very bad if you switch from portrait mode to landscape mode on mobile devices

**To Reproduce**
1. Go to https://hubs.mozilla.com with portrait mode on your mobile device
2. Enter a room
3. Switch to landscape mode by rotating your device
4. See the worse visual quality

**Expected behavior**

Visual quality keeps the same even after switching to landscape mode.

**Hardware** 
 - Device: Phones
 - OS: Android, iOS
 - Browser: Chrome, Safari

**Additional context**

I guess the root issue is max resolution in the preference? There might be a chance that the width and height are not switcheg when switching the portrait/landscape mode.

takahirox:
I set large enough numbers to the max resolution preference and the visual quality keeps good even if rotating the device. So my speculation seems correct.

Probably a solution would be to detect the orientation change and switch the max resolution width and height on the orientation change.

PR #5450, opened 2022-05-21
Handle max resolution properly based on the screen orientation

takahirox:
Resolves: #5431

Currently Hubs Client doesn't care screen orientation. So for example if mobile device users rotate their devices the visual quality can be bad because max resolution width and height should be swapped but they aren't.

With this commit Hubs Client will manage max resolution based on the screen orientation and the problem will be resolved.

### DIFF
--- a/src/utils/screen-orientation-utils.js
+++ b/src/utils/screen-orientation-utils.js
@@ -1,0 +1,79 @@
+// What this screen-orientation-utils does are
+// 1. Hide the screen orientation API difference across browsers
+//    ScreenOrientation API should be used but Safari doesn't support it yet.
+//    Instead deprecated window.orientation and orientationchange event
+//    need to be used on Safari for now.
+// 2. Manage maxResolution preferences values based on the screen orientation
+
+import { isIOS as detectIOS } from "./is-mobile";
+
+const isIOS = detectIOS();
+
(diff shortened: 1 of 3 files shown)


takahirox:
It doesn't work correctly on Safari yet so still draft.

takahirox:
Works fine on iOS, too. Ready for review.

src/utils/screen-orientation-utils.js
```diff
@@ -0,0 +1,79 @@
...
+
+const isNaturalOrientation = () => {
+  return getAngle() % 180 === 0;
+};
+
+// Return the screen width based on the current screen orientation
+const getScreenWidth = () => {
+  // Is seems screen.width value is based on the natural screen orientation on iOS
+  // while it is based on the current screen orientation on Android (and other devices?).
+  if (isIOS) {
```
takahirox, line 40: FYI: Due to this bahavior on iOS, currenlty renderer width and height are not set properly on iOS in landscape mode.

takahirox:
I hope we can review and merge this PR before revisiting our (max) resolution logic and configuration #5465 #5466

brianpeiris (approved):
LGTM. Nice job separating the orientation code out like that, very clean.
It's a bit unfortunate that we have to use `forceUpdate` here, but I think it's justified in this case.

PR #5454, opened 2022-05-23
Introduce Audio PanningQuality preference

### LINKED ISSUE #5278: Consider to use equalpower panning model

takahirox:
Fixes #5278 

`WebAudio` and `Three.js` uses `HRTF` panning model for Panner audios by default. But `HRTF` is said to be better quality but much more costly than another pannind model `equalpower`. Some users, especially low-end or mobile device users, may think they want to set `equalpower` panning model for better performance by sacrificing the audio panning quality.

This PR introduces Audio PanningQuality preference. The high quality corresponds to `HRTF` while the low quality corresponds to `equalpower`.

**Additional Context**

* I didn't change the default model `HRTF` so far. We may think we want to set `equalpower` (on mobile devices) by default. But I want to think of the appropriate default value after testing across the devices.
* (For me) the quality difference is obvious. So currently I don't think to force the value and I hope we can keep in the preferences even after we decide the default value because the preferred value may depend on users.
* Some of us may think we can integrate the "Accessibility Disable audio left/right panning" and "Audio Panning Quality" preferences, like into "Audio Panning Quality High/Low/None"

### DIFF
--- a/src/systems/sound-effects-system.js
+++ b/src/systems/sound-effects-system.js
@@ -18,6 +18,7 @@ import URL_SPAWN_EMOJI from "../assets/sfx/emoji.mp3";
 import URL_SPEAKER_TONE from "../assets/sfx/tone.mp3";
 import { setMatrixWorld } from "../utils/three-utils";
 import { SourceType } from "../components/audio-params";
+import { getOverriddenPanningModelType } from "../update-audio-settings";
 
 let soundEnum = 0;
 export const SOUND_HOVER_OR_GRAB = soundEnum++;
@@ -147,6 +148,12 @@ export class SoundEffectsSystem {
       : new THREE.PositionalAudio(this.scene.audioListener);
     positionalAudio.setBuffer(audioBuffer);
     positionalAudio.loop = loop;
+    if (!disablePositionalAudio) {
+      const overriddenPanningModelType = getOverriddenPanningModelType();
+      if (overriddenPanningModelType !== null) {
+        positionalAudio.panner.panningModel = overriddenPanningModelType;
+      }
+    }
     this.pendingPositionalAudios.push(positionalAudio);
     this.scene.systems["hubs-systems"].audioSystem.addAudio({
       sourceType: SourceType.SFX,
--- a/src/update-audio-settings.js
+++ b/src/update-audio-settings.js
@@ -1,6 +1,7 @@
 import {
   AudioType,
   SourceType,
+  PanningModelType,
   MediaAudioDefaults,
   AvatarAudioDefaults,
   TargetAudioDefaults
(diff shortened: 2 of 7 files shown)


src/update-audio-settings.js
```diff
@@ -21,23 +22,37 @@ function applySettings(audio, settings) {
...
     audio.setMaxDistance(settings.maxDistance);
+    audio.panner.panningModel = settings.panningModel;
     audio.panner.coneInnerAngle = settings.coneInnerAngle;
     audio.panner.coneOuterAngle = settings.coneOuterAngle;
     audio.panner.coneOuterGain = settings.coneOuterGain;
   }
   audio.gain.gain.setTargetAtTime(settings.gain, audio.context.currentTime, 0.1);
 }
 
+// Return panning model depending on the preferences
```
johnshaughnessy, line 33: Can remove this comment

src/update-audio-settings.js
```diff
@@ -21,23 +22,37 @@ function applySettings(audio, settings) {
...
   const sceneOverrides = APP.sceneAudioDefaults.get(sourceType);
   const audioOverrides = APP.audioOverrides.get(el);
   const audioDebugPanelOverrides = APP.audioDebugPanelOverrides.get(sourceType);
   const zoneSettings = APP.zoneOverrides.get(el);
-  const preferencesOverrides = APP.store.state.preferences.disableLeftRightPanning
-    ? { audioType: AudioType.Stereo }
-    : {};
+
+  const preferencesOverrides = {
+    panningModel: getPanningModelType()
```
johnshaughnessy, line 49: If the user does not specify a preference, we should not add `panningModel` to `preferencesOverrides`. 

If the user does not have a preference, their preference is `null`:

```js
function getPanningModelType() {
  return APP.store.state.preferences.audioPanningQuality === "Low"  ? PanningModelType.EqualPower
            APP.store.state.preferences.audioPanningQuality === "High" ? PanningModelType.HRTF
            : null;
}
```

If we always add `panningModel` to `preferencesOverrides`, these layers can never affect the `panningModel`: `defaults`, `sceneOverrides`, `audioOverrides`, `audioDebugPanelOverrides`, `zoneSettings`.

src/systems/sound-effects-system.js
```diff
@@ -148,6 +149,9 @@ export class SoundEffectsSystem {
       : new THREE.PositionalAudio(this.scene.audioListener);
     positionalAudio.setBuffer(audioBuffer);
     positionalAudio.loop = loop;
+    if (!disablePositionalAudio) {
+      positionalAudio.panner.panningModel = getPanningModelType();
```
johnshaughnessy, line 153: If you make the change I recommend in the other comment, be careful that `getPanningModelType` could be `null`.

johnshaughnessy (approved):

takahirox:
@johnshaughnessy Would you mind reviewing the update to check if they look ok?

takahirox:
@johnshaughnessy Would you mind reviewing the update to check if they look ok?

takahirox:
@johnshaughnessy Would you mind if reviewing again to check if the updates look ok?

PR #5460, opened 2022-05-25
Re-enable Positional audio on Safari

takahirox:
Safari WebAudio Panner node seemed to have a problem that the audio can be broken. We decided to force to disable panning audio on Safari. See #4411 for details.

The problem seems to have been resolved on Safari end. We can re-enable the panning audio even on Safari.

Just in case we should test again to check if the problem is no longer reproducible before merging this PR.


takahirox:
We tested in Hubs team and confirmed that audio on iPhone Safari and Mac Safari is fine now. I think we can re-enable positional audio on them again.

brianpeiris (approved):
LGTM

PR #5479, opened 2022-06-02
Remove three-batch-manager

takahirox:
Our batching system is disabled. And Three.js may introduce official batching API (Currently at proposal phase). We may directly use it or write our own based on it when we will enable it again.

No pointing keeping three-batch-manager for now.


netpro2k (approved):
This LGTM. Definitely good to clean this up since w aren't using it anyway.  If we do add some form of dynamic batching in the future we can refer to this PR for hints on what we might need to change again.

src/components/media-image.js
```diff
@@ -162,7 +150,6 @@ AFRAME.registerComponent("media-image", {
           break;
         default:
           this.mesh.material.transparent =
-            !this.data.batch ||
```
takahirox, line 165: Here, I noticed that `this.mesh.material.transparent` should be alwayt `true` because `data.batch` has been removed (`false`). I'm making a follow up PR now.

PR #5484, opened 2022-06-06
Fix a media-image transparent bug

takahirox:
In the three-batch-manager removal commit (#5479), in `media-image` `mesh.material.transparent` has been wrongly set for default
alphaMode. This commit fixes it. Refer to https://github.com/mozilla/hubs/pull/5479/files#r890466313


brianpeiris (approved):
LGTM

PR #5485, opened 2022-06-06
Update node.js to v16. Minor and patch updates to dependencies

brianpeiris:
This PR updates our package-lock versions so that we can update to node.js v16. It updates the node.js version we use in our habitat build plan. It also updates several dependencies to newer patch and minor versions. Finally, it resolves some warnings from babel and SASS that resulted from newer versions of those libraries.
I've done some initial testing locally, and haven't spotted any regressions yet, but we might only be able to find some issues in a staging environment.

TODO:
- [ ] There seems to be an issue with the habitat build, where it fails to resolve some npm dependencies. This requires further investigation.


admin/package.json
```diff
@@ -70,5 +70,8 @@
     "webpack-cli": "^3.2.3",
     "webpack-dev-server": "^3.1.14",
     "worker-loader": "^2.0.0"
+  },
+  "overrides": {
+    "three": "github:mozillareality/three.js#hubs/master"
```
netpro2k, line 75: I think this should match what is in the main hubs package.json

brianpeiris: Makes sense. I think I just happen to open this PR when we were in the middle of a three.js update.

netpro2k:
Closed in favor of #5619

Issue #5414, opened 2022-05-06
Upgrade Three.js for better performance


takahirox:
**Is your feature request related to a problem? Please describe.**

`getProgramCacheKey()` in Three.js seems to be one of the performance bottlenecks.

![unknown](https://user-images.githubusercontent.com/7637832/167228289-7422b33c-3a8c-4ea3-862c-df85e808cc12.png)

**Describe the solution you'd like**

The performance of it seems to be improved in r136. Maybe worth to upgrade our Three.js to the latest 

https://github.com/mrdoob/three.js/pull/22960

**Additional context**

Three.js we use now is r133 based

blairmacintyre:
Just talked with Rik about issues I'm having on Quest2, and he pointed to some fixes in 137 that will help perf (multisampling and layers) and solve my issues (custom shader related).  

I too am wondering if upgrading is feasible.

takahirox:
Status: I'm working on upgrading Three.js but encountered some weird problem on compilation with `npm run dev`.

takahirox:
@blairmacintyre Still a bit at testing phase but I made a PR #5488 . You can locally merge and test.

blairmacintyre:
@takahirox  I'm not sure I can test this;  it's against the main branch of the hubs client, but we're using the hubs-cloud branch.

PR #5488, opened 2022-06-07
Upgrade Three.hs to r141

takahirox:
Fixes: #5414

Require: https://github.com/MozillaReality/three.js/tree/hubs-patches-141
Require: https://github.com/MozillaReality/aframe/pull/35
Require: https://github.com/MozillaReality/networked-aframe/pull/51

Three.js migration guide

https://github.com/mrdoob/three.js/wiki/Migration-Guide

Still testing, so draft now.


takahirox:
Let me switch this PR to ready for review.

src/components/tools/networked-drawing.js
```diff
@@ -204,7 +204,7 @@ AFRAME.registerComponent("networked-drawing", {
     };
 
     const glb = await new Promise(resolve => {
-      exporter.parse(mesh, resolve, {
+      exporter.parse(mesh, resolve, undefined, {
```
netpro2k, line 207: We should probably reject the promise if the error callback gets called. I assume before it was throwing an exception which would cause `serializeDrawing` to return a rejected promise since it is an async function.

src/utils/three-utils.js
```diff
@@ -37,14 +37,8 @@ export function disposeNode(node) {
   }
 
   if (node.material) {
-    let materialArray;
-    if (node.material instanceof THREE.MeshFaceMaterial || node.material instanceof THREE.MultiMaterial) {
```
netpro2k, line 41: Didn't see it in the migration guide, but dug in and looks like MeshFaceMaterial was removed so this makes sense. I don't think we were using it for anything anyway.

webpack.config.js
```diff
@@ -447,6 +447,19 @@ module.exports = async (env, argv) => {
...
+            __dirname,
+            "node_modules",
+            "three",
+            "examples",
+            "jsm",
+            "exporters",
+            "GLTFExporter.js"
+	  )],
+          loader: "babel-loader",
+          options: legacyBabelConfig
```
netpro2k, line 461: Still slightly confused why this needed the "legacy" loader config... but I guess its fine. We definitely need to do a pass on cleaning up and updating our bundling.

netpro2k (approved):
This was definitely one of the less scary ThreeJS diffs. Everything looks good to me.

src/components/tools/networked-drawing.js
takahirox: Good catch. Updated to reject the promise on error.

Issue #4942, opened 2021-12-14
There is no radar visuals from media and user when enabling Audio Debug Panel


SV-AndreiC:
**[Affected Versions]:**
- Safari 15.1
- Chrome 96

**[Affected Platforms]:**
- iOS 15.1
- Android 11

**[Prerequisites]:**
- The Hubs Cloud stack is set up.
- Have a room with media objects, created and opened 
- Have another user in the room.

**[Steps to reproduce]:**
1. Activate the Audio Debugging tool from preferences.
2. Tap Play on the media objects in the room.
3. Have the other user start talking.

**[Expected results]:**
- There should be a radar visual around every sound source in the room.

**[Actual results]:**
- There is no radar visual from any sound source in the room.

**[Notes]:**
- This works on Windows 10 Chrome and Firefox.
- Attached 2 screenshots with actual and expected results:
Expected:
![expected](https://user-images.githubusercontent.com/93765706/146029035-092b0302-eaff-4617-b788-670da4ddf8d7.PNG)
Actual:
![actual](https://user-images.githubusercontent.com/93765706/146029053-80f29a7d-b0b4-46cb-a112-0f604c7ee4a2.JPEG)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBSOLD-1216)

Softvision-RemusDranca:
I was also able to reproduce this issue on macOS BigSur 11.6.1 using Safari browser.

takahirox:
Removed "[MOBILE]" from the issue title because @Softvision-RemusDranca reproduced the problem on desktop Mac.

Pattersonbl2:
This has been tagged in the previous release has it been fixed?

keianhzo:
@Pattersonbl2 This is not yet solved, I just opened a PR that fixes this issue. When that one is merged this issue will be closed.

takahirox:
> In iOS I think it's related to the fact that we are using derivatives (fwidth) in the audio debug fragment shader and it might not be supported in some devices/iOS versions.

How did you find that iOS may not support fwidth? This page mentions that iOS Safari supports WebGL OES_standard_derivatives extension.

https://developer.mozilla.org/en-US/docs/Web/API/OES_standard_derivatives

But also I confirmed that #5492 resolves the problem tho.

cornelionce:
The radar visuals of the Audio debug panel are now displayed on iOS devices.

PR #5492, opened 2022-06-09
Add Safari compat for the audio debugger

keianhzo:
Remove `fwidth` dependency to support implementation without standard derivatives support.

### DIFF
--- a/src/systems/audio-debug-system.js
+++ b/src/systems/audio-debug-system.js
@@ -4,8 +4,9 @@ import audioDebugFrag from "./audio-debug.frag";
 import { DistanceModelType } from "../components/audio-params";
 import { getWebGLVersion } from "../utils/webgl";
 import { getMeshes } from "../utils/aframe-utils";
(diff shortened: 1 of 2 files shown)


takahirox (approved):
I didn't look into the change in details but I confirmed that this change fixes the problem on my iPhoneX Safari.

Would you mind if writing a note in the PR comment and commit log that fwidth caused a problem on iOS? It may help anyone who wants to track the problem.

takahirox:
> as Safari doesn't support the standard derivatives extension.

Just to clarify, this MDN page mentions that iOS Safari supports WebGL OES_standard_derivatives extension.

https://developer.mozilla.org/en-US/docs/Web/API/OES_standard_derivatives

Is the MDN page wrong?

keianhzo:
@takahirox It was right, I answered also here: https://github.com/mozilla/hubs/issues/4942#issuecomment-1163038159 and created a new PR with a better fix.

Issue #5466, opened 2022-05-26
Easier (max) resolution configuration in the preferences dialog


takahirox:
**Is your feature request related to a problem? Please describe.**

Users can configure max resolution in the preference.

![image](https://user-images.githubusercontent.com/7637832/170577798-0f876723-0817-45d6-b7bc-c48ff45d6c58.png)

But it has some problems.

* Unclear whether the resolution is in screen device resolution or CSS resolution (screen device resolution = CSS resolution * device pixel ratio)
* For generic users it's hard to know what numbers to enter even if they want to balance the performance and the visual quality

Might be related #5465

**Describe the solution you'd like**

Provides an easier UI to control the (max) resolution for performance - quality balance.

For example, resolution multiplier slide bar?

cornelionce:
A resolution slider was introduced and it functions correctly. It is much easier for the end-user to change the resolution now.

PR #5496, opened 2022-06-11
Clean up calculateRendererSize()

### LINKED ISSUE #5465: Set Canvas width/height correctly
**Description**

Hubs Client seems to unexpectedly set CSS screen resolution (`screen.width, .height`), not Screen device resolution (CSS resolution * device pixel ratio), to Canvas width and height.

* `maxResolution` is `screen.width` and `.height` by default (unless they are set from the preferences)
* Imagine that browser is full screen (ie. mobile device)
* [`getRenderResolution()` returns `maxResolution` / pixel ratio](https://github.com/mozilla/hubs/blob/01d8e31dbdf1f300174f0fdc17dc9d5cb4f73580/src/react-components/room/useResizeViewport.js#L7-L24)
* The returned values are passed to `renderer.setSize()`
* [`renderer.setSize()` sets the passed size * pixel ratio to Canvas width and height](https://github.com/MozillaReality/three.js/blob/58551805d11cbc5510989950beb35d0141ca9a71/src/renderers/WebGLRenderer.js#L409-L410)

In short, `screen.width, .height` / pixel ratio * pixel ratio are set to Canvas width and height.

But for the best visual quality, Screen device resolution (`screen.width, .height` * pixel ratio) should be set to Canvas width and height.

This problem causes the blurriness on mobile devices.

**Expected behavior**

Screen device resolution are set to Canvas width and height (in full screen).

**Hardware** 
 - Device: Any
 - OS: Any
 - Browser: Any

**Additional context**

Currently the canvas size are smaller than it needs to be and the visual quality is worse. If we fix the problem the visual quality can improve but the performance can be affected.

We may need to consider smaller resolution for mobile devices.

takahirox:
Fixes #5465

Currently `getRenderResolution()` in `useResizeViewport.js` is confusing because physical pixels (`screen.width, .height * window.devicePixelRatio`) and CSS pixels (`screen.width, .height`) are mixed.

This PR cleans up `getRenderResolution()`.

As written in #5465, I thought of changing the default max resolution from `screen.width, .height` to `screen.width, .height * window.devicePixelRatio` but I noticed bad performance impact. So I remained the values for now.

For users who want better visual quality, I'm thinking of easier resolution controls #5466.


src/react-components/room/useResizeViewport.js
```diff
@@ -8,35 +8,28 @@ import {
...
+    return canvasRect;
   }
 
-  if (canvasRect.height * pixelRatio > maxResolution.height && maxResolution.height !== -1) {
-    return {
-      height: Math.round(maxResolution.height / pixelRatio),
-      width: Math.round((maxResolution.height * aspectRatio) / pixelRatio)
-    };
-  }
+  const convertionRatio = Math.min(maxWidth / canvasRect.width, maxHeight / canvasRect.height);
```
netpro2k, line 27: This should be `conversionRatio`

src/react-components/room/useResizeViewport.js
```diff
@@ -8,35 +8,28 @@ import {
...
+  const maxWidth = maxResolution.width / pixelRatio;
+  const maxHeight = maxResolution.height / pixelRatio;
 
-  if (canvasRect.width * pixelRatio > maxResolution.width && maxResolution.width !== -1) {
-    return {
-      width: Math.round(maxResolution.width / pixelRatio),
-      height: Math.round(maxResolution.width / aspectRatio / pixelRatio)
-    };
+  if (canvasRect.width <= maxWidth && canvasRect.height <= maxHeight) {
+    return canvasRect;
```
netpro2k, line 24: Minor since we only ever read `width` and `height` from the result of this function anyway, but this function end sup having multiple return types. Here we are returning the passed in `DOMRectReadOnly`, but elsewhere we are returning a new object `{width, height}`.

netpro2k (commented):
Pointed out some minor issues, but otherwise LGTM

src/react-components/room/useResizeViewport.js
takahirox: Honestly I didn't really like that multiple return types. But the current implementation does it, too. I just didn't change. We may be able to clean up later.

src/react-components/room/useResizeViewport.js
takahirox: Updated, thanks.

Issue #5185, opened 2022-02-24
Enabling joysticks requires the web page reload


takahirox:
**Description**

**To Reproduce**
1. Enter a room
2. Click on 'Menu" - "Preferences" - "Controls"
3. Check "Enable left on-screen joystick for moving around" and "Enable right on-screen joystick for looking around" checkboxes
4. Confirm that "the joystics don't appear even if you tap left/right bottom of the screen"
5. Reload the page and enter the room again
6. Confirm that "the joystics appear if you tap left/right bottom of the screen"

**Expected behavior**
The joysticks appear after enabling them on Preferences even without the page repload. Or if it actually needs reloading, the reload require message should appear.

**Hardware** 
 - Device: Phone (Samsung S8+)
 - OS: Android
 - Browser: Chrome 98

takahirox:
This line causes an error

https://github.com/mozilla/hubs/blob/d038beb4fed73a4fa8d4891f4454a7c09ec1a3fa/src/components/virtual-gamepad-controls.js#L5

`referenceEl.parentNode` is `null`.

cornelionce:
The page refresh is no longer needed when the joysticks are enabled/disabled.

PR #5499, opened 2022-06-14
Fix Enable VirtualJoystick preference.

takahirox:
Fixes: #5185

Currently an error happens if virtual joysticks are enabled from the preferences because `virtual-gamepad-controls` component attempts to access `sceneEl.canvas.parentNode` even though it is `null`.

This commit fixes it by ensuring non-null `sceneEl.canvas.parentNode` before accessing it.

### DIFF
--- a/src/components/virtual-gamepad-controls.js
+++ b/src/components/virtual-gamepad-controls.js
@@ -79,6 +79,9 @@ AFRAME.registerComponent("virtual-gamepad-controls", {
     if (!isChanged) {
       return;
     }
+    if ((newEnableLeft || newEnableRight) && !this.mockJoystickContainer.parentNode) {
+      insertAfter(this.mockJoystickContainer, this.el.sceneEl.canvas);
+    }
     if (!this.enableLeft && newEnableLeft) {
       this.createLeftStick();
     } else if (this.enableLeft && !newEnableLeft) {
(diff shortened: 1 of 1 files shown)


netpro2k (commented):
This change looks fine, but I am a bit confused how it fixes a case where `sceneEl.canvas.parentNode` is null. Seems like the insert just moved to a slightly different spot in the same function unless I am missing some odd control flow.

takahirox:
https://github.com/mozilla/hubs/blob/05b1543a53ad3fe7f7ac3d5eaa25f40bb687578e/src/components/virtual-gamepad-controls.js#L83

`insertAfter(this.mockJoystickContainer, this.el.sceneEl.canvas);` inserts `mockJoystickContainer` under `sceneEl.canvas.parent` before `sceneEl.canvas`.

```
- sceneEl.canvas.parent
  - mockJoystickContainer // insert here
  - sceneEl.canvas
```

https://github.com/mozilla/hubs/blob/05b1543a53ad3fe7f7ac3d5eaa25f40bb687578e/src/components/virtual-gamepad-controls.js#L119
     `this.mockJoystickContainer.parentNode.removeChild(this.mockJoystickContainer)` removes `mockJoystickContainer` from `sceneEl.canvas.parent (== mockJoystickContainer.parent)`.

```
- sceneEl.canvas.parent
  - sceneEl.canvas
```

But I agree with that it may not be so easy to understand by just reading.

PR #5504, opened 2022-06-15
Introduce slider for max resolution preference

takahirox:
Fixes: #5466

Currently max resolution preference directly takes numbers but most users may not know what numbers to enter.

This commit introduces a slider for max resolution preference for easier controls. The slider value is multipler against screen resolution in physical pixels.

image

Note: I made the two right number input fields readonly for simplicity. Is there any use case that users want custom max resolution numbers?

### DIFF
--- a/src/react-components/preferences-screen.js
+++ b/src/react-components/preferences-screen.js
@@ -17,6 +17,8 @@ import {
   removeOrientationChangeListener,
   getMaxResolutionWidth,
   getMaxResolutionHeight,
+  getScreenResolutionWidth,
+  getScreenResolutionHeight,
   setMaxResolution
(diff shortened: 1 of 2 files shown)


takahirox:
> I still find the whole max resolution preference pretty confusing since I am almost always going to be running Hubs in a window, not fullscreen.

Do you want resolution multiplier preference? Like canvas.size(width, height) * window.devicePixelRatio * multiplier (0.1-1.0).

takahirox:
> I still find the whole max resolution preference pretty confusing since I am almost always going to be running Hubs in a window, not fullscreen.

Not sure if I understand correctly, but I think whole max resolution actually means no resolution limit.

PR #5506, opened 2022-06-15
Fix AvatarPreview

### LINKED ISSUE #5503: Avatar previews appear to be broken
**Description**
Avatar previews don't work in the preview URL links.

**To Reproduce**
1. Go to any Hubs room.
2. Click on 'More', 'Change Name and Avatar'
3. Click on 'Change Avatar'
4. Right-click on any avatar (such as a Featured avatar) and copy the link.
5. Paste the link into a new tab.

** Result **
The avatar preview area is blank and says "Loading failed, Please choose another avatar"

**Expected behavior**
The avatar preview window shows the avatar.

**Screenshots**
n/a

**Additional context**
The avatars still work when you choose one, and the 2D preview still works when you paste it into the Hubs room chat and press the magic wand button.

takahirox:
Fixes #5503

Currently AvatarPreview page fails to load model because of missing THREE.GLTFLoader.

This PR fixes it by importing the loader.

### DIFF
--- a/src/react-components/avatar-preview.js
+++ b/src/react-components/avatar-preview.js
@@ -8,6 +8,7 @@ import classNames from "classnames";
 // under the three/examples/js to avoid tree shaking
 // in webpack production mode.
 require("three/examples/js/controls/OrbitControls");
+require("three/examples/js/loaders/GLTFLoader");
 
 import { createDefaultEnvironmentMap } from "../components/environment-map";
 import { loadGLTF } from "../components/gltf-model-plus";


netpro2k (commented):
This fix should work. Thanks for getting to it so quickly. It might be more "correct" to do the import in each file we actually use GLTFLoader (avatar-editor.js and gltf-model-plus.js) and remove it from elsewhere, that way we won't run into this problem in other spots in the future. Or better yet, use the jsm version of GLTFLoader so we don't need the weird `require` hack.

Merging this as is is totally fine though.

takahirox:
Ah, sorry I have overlooked https://github.com/mozilla/hubs/issues/5503#issuecomment-1155785374, let's go with this PR for now to fix the problem. And let's clean up later when we will have time.

Issue #4933, opened 2021-12-13
[Android] Can’t join room if the scene is changed in the room lobby


SV-AndreiC:
**[Affected Versions]:**
- Chrome 94.0.4664.92

**[Affected Platforms]:**
- Android 11

**[Prerequisites]:**
- The Hubs Cloud stack is set up.
- Have the homepage opened.

**[Steps to reproduce]:**
1. Tap the “Create room” button.
2. Once in the lobby, tap the “More” button.
3. Select “Room Info and Settings” and tap “Change Scene”.
4. Select a new scene and dismiss the room settings pannel.
5. Tap “Join Room” and then “Enter Room” from the “Microphone Setup” panel.

**[Expected results]:**
- The room is entered by the user.

**[Actual results]:**
- The “Enter Room” button is highlighted but nothing happens.

**[Notes]:**
- If the browser "Back" button is used, the room is entered.
- Attached is a recording of the issue:
![mobileCR](https://user-images.githubusercontent.com/93765706/145846837-bfa78e24-405b-4737-96a4-19409b55c682.gif)

takahirox:
Confirmed this happens on my Samsung S8+ Android Chrome. And this doesn't happen on my Windows 10 Chrome. It may be a mobile or android specific issue.

takahirox:
I confirmed that this problem still happens on my Pixel6 Android 12 Chrome.

takahirox:
The root issue seems that `await screenfull.request()` isn't resolved in `src/utils/fullscreen.js`.

https://github.com/mozilla/hubs/blob/79b115695e7aee904ebd8dfc77f17528b701ef4c/src/utils/fullscreen.js#L18

When entering a room, attempting to get in fullscreen. But it isn't resolved so can't enter.

When opening the change scene dialog, Hubs Client gets in fullscreen here

https://github.com/mozilla/hubs/blob/7085199dc9e7f317ef6880b7170d7aa114a3eb61/src/react-components/ui-root.js#L748

and after that doesn't exit and keep the fullscreen mode even when closing the dialog. In fullscreen mode `screenfull.request()` doesn't seem to be resolved.

Solution ideas would be either one.
1. Don't get in fullscreen mode when opening the change scene dialog
2. Exit fullscreen mode when closing the change scene dialog
3. Skip `screenfull.request()` when entering a room if already in fullscreen mode

PR #5512, opened 2022-06-16
Prevent fullscreen request in fullscreen mode

takahirox:
Fixes: #4933 

`screenfull.request()` doesn't seem to be resolved if it is called in full screen mode. So add `if !screenfull.isFullscreen` guard`to prevent the call in full screen mode.


netpro2k (approved):
kinda surprising screenfull doesn't handle that for us, but this LGTM

PR #5513, opened 2022-06-17
Always visible and fixed position virtual joysticks

takahirox:
With this commit the virtual joysticks will be always visible if they are enabled. They are transparent and they will be less transparent if they are touched.

And they will be placed at fixed position.

These may be more intuitive to users.

image

### DIFF
--- a/src/components/virtual-gamepad-controls.js
+++ b/src/components/virtual-gamepad-controls.js
@@ -127,6 +127,8 @@ AFRAME.registerComponent("virtual-gamepad-controls", {
     this.rightTouchZone.classList.add(styles.touchZone, styles.right);
     insertAfter(this.rightTouchZone, this.mockJoystickContainer);
     this.rightStick = nipplejs.create({
+      mode: "static",
+      position: {left: "50%", top: "50%"},
       zone: this.rightTouchZone,
       color: "white",
       fadeTime: 0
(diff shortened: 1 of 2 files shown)


takahirox:
I'm not a CSS professional so comments especially for CSS are very welcome.

takahirox:
Ah wait, the joysticks should be hidden if the scene is hidden... What's the easiest way to hide it? Using zIndex?

image

takahirox:
Fixed it.

src/components/virtual-gamepad-controls.js
```diff
@@ -127,10 +127,16 @@ AFRAME.registerComponent("virtual-gamepad-controls", {
...
+      mode: "static",
+      position: {left: "50%", top: "50%"},
       zone: this.rightTouchZone,
       color: "white",
       fadeTime: 0
     });
+    // nipplejs sets z-index 999 but it makes the joysticks
+    // visible even if the scene is hidden for example by
+    // preference dialog. So remove z-index.
+    this.rightStick[0].ui.el.style.removeProperty("z-index");
```
netpro2k, line 139: I somehow did not know about `removeProperty`

netpro2k (commented):
Code changes LGTM

takahirox:
I got only positive comments for this change from the community for now. I will go with this change. I may revisit if I will get objections.

PR #5517, opened 2022-06-17
Allow newlines in troika-text

wmurphyrd:
With the switch to troika text, we lost processing of newline characters in text values. 
Screenshot from 2022-06-17 14-07-43

Troika's docs said they support them, so I took a peek at their source and found [this little preprocessing on the text value](https://github.com/lojjic/aframe-troika-text/blob/master/src/aframe-troika-text-component.js#L109-L111) was not present in the hubs version. 

Screenshot from 2022-06-17 14-24-53

### DIFF
--- a/src/components/troika-text.js
+++ b/src/components/troika-text.js
@@ -95,7 +95,7 @@ AFRAME.registerComponent("text", {
     const mesh = this.troikaTextMesh;
 
     // Update the text mesh
-    mesh.text = data.value || "";
+    mesh.text = (data.value || "").replace(/\\n/g, "\n").replace(/\\t/g, "\t");
     mesh.textAlign = data.textAlign;
     mesh.anchorX = data.anchorX;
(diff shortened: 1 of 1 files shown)


netpro2k:
Ah interesting, I wonder why the string is ending up with a literal `['\' 'n']` instead of a newline character. I guess you just cant type actual newlines in to the HTML (would probably work via `setAttribute`. Could try `&#10;`. In anycase this change seems sane enough.

wmurphyrd:
@netpro2k oh yeah I should have specified it's not literal new lines, but `<a-entity text="value: line 1\nline 2"></a-entity>` that is affected

matthewbcool:
This looks ok to me, @netpro2k do you see any issues with pulling this in?

wmurphyrd:
@matthewbcool thanks for following up. I expect this PR is obsolete with the refactor away from A-Frame underway

PR #5524, opened 2022-06-22
Restore derivatives support and limit the uniform vectors number

keianhzo:
After looking deeper into this I've seen that the issue was not related with the standard derivatives support but with the amount of fragment uniform vectors that Safari supports. We were limiting the number of displayed audio debug elements to 64 by default but in some platforms like Safari that exceeds the amount of supported uniform vectors (256 Vs 1024 in Chrome/FF). Our shader is using 10 uniform vectors right now to store audio elements data and 64x10 exceeded the Safari 256 limit and that's why the shader didn't work in that platform.

We can revisit this in the future to find a more efficient approach like using a texture to pass the uniform audio element data but for now this seems to work fine.

Fixes [#4942](https://github.com/mozilla/hubs/issues/4942)

### DIFF
--- a/src/systems/audio-debug-system.js
+++ b/src/systems/audio-debug-system.js
@@ -4,9 +4,6 @@ import audioDebugFrag from "./audio-debug.frag";
 import { DistanceModelType } from "../components/audio-params";
 import { getWebGLVersion } from "../utils/webgl";
 import { getMeshes } from "../utils/aframe-utils";
-import { isSafari } from "../utils/detect-safari";
-
-const MAX_DEBUG_SOURCES = isSafari() ? 24 : 64;
 
 const fakePanner = {
   distanceModel: DistanceModelType.Inverse,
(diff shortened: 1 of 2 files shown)


src/systems/audio-debug-system.js
```diff
@@ -23,10 +20,16 @@ AFRAME.registerSystem("audio-debug", {
   },
 
   init() {
+    this.max_debug_sources = 64;
```
takahirox, line 23: For consistent naming rule, `this.maxDebugSources` would be better?

src/systems/audio-debug-system.js
```diff
@@ -23,10 +20,16 @@ AFRAME.registerSystem("audio-debug", {
...
     this.unsupported = false;
     const webGLVersion = getWebGLVersion(this.el.sceneEl.renderer);
     if (webGLVersion < "2.0") {
       this.unsupported = true;
+    } else {
+      const gl = this.el.sceneEl.renderer.getContext();
+      const max_f_vectors = gl.getParameter(gl.MAX_FRAGMENT_UNIFORM_VECTORS);
```
takahirox, line 30: `maxFVectors` or `maxFragmentVectors`?

takahirox (approved):
I haven't tested yet and honestly I'm not so familiar with Audio debug implementation but the change is clearer. We can merge this if it works expectedly.

src/systems/audio-debug-system.js
keianhzo: Yep, can do, I think I've spent way too long in Python land 😅

src/systems/audio-debug-system.js
keianhzo: Sure thing.

Issue #5188, opened 2022-02-25
Allow to hide the bottom 2D UI bar


takahirox:
**Is your feature request related to a problem? Please describe.**

The bottom 2D UI bar may be too tall on some platforms, especially on mobile devices with landscape mode.

If users use the virtual joysticks the tall UI bar can be more problematic. The UI bar makes the joysticks too high up on the screen.

![image](https://user-images.githubusercontent.com/7637832/155801683-cf076a7a-cc8a-40d4-8a0f-7f51ecfa30d0.png)

**Describe the solution you'd like**

Allow to hide the bottom 2D UI bar. Maybe click something that slides it down and off the screen, with a single small button that will bring it back.

takahirox:
Related or duplicated with #3468

takahirox:
An important thing would be Voice (mic) and Share should always visible for user privacy. They should always be able to be aware of mic on and screen/camera share on.

So, an easy solution would be transparent tool bar. It would be opacity if it's focused.

![image](https://user-images.githubusercontent.com/7637832/174418083-b9a0fe4d-3a84-4226-a6f6-e834da6fa1e3.png)

vincentfretin:
Personally I'm not a fan of semi-transparent block that become solid on hover, and you can't really hover on touch devices. The low opacity value you used in your image is not accessible for people having bad eyes.
Using a semi-transparent like this doesn't resolve the fact you have less space in the scene part to move the avatar on mobiles.
I would prefer an arrow on left or right to minimize it and keep a smaller voice and share button maybe.
Or a button to make all the icons twice smaller and no text so reducing the bottom bar height by 2?

vincentfretin:
Or avoid a button to minimize entirely, and use an alternate bottom bar for low resolutions, so mobiles and tablets.
Some folks did an alternate UI with fully transparent background, smaller icons with no text on their Hubs Cloud, there is some screenshots on discord dev-general.

Exairnous:
I agree with what @vincentfretin has said, but I'd also like to mention that it can be hidden already on desktop via the backtick key and so I don't think the UI needs any modification on desktop, except maybe adding an option to the preferences to make this key configurable as not everyone has access to it.

takahirox:
> it can be hidden already on desktop via the backtick key 

Oh, I didn't know that...

vincentfretin:
[...]

PR #5525, opened 2022-06-23
Remove tool bar UI background

takahirox:
Fixes: #5188

The tool bar UI is too tall especially in the landscape mode on mobile devices.

This PR removes the background of the tool bar UI. The viewport becomes vertically wider.

I'm not sure if you like it because I'm not a designer. I hope designers brush up the design later even if we go with this change.

image

unknown

Concepts and thoughts
* Current my main focus is resolving the problem that the viewport is vertically narrow in landscape mode on mobile devices. #5188
* I'm not a designer. I hope we can go with this change so far, and then the designers or artists brush up, improve, or clean up it.
* I think Mic and Share icons should be always visible for user privacy. Users should be always aware of whther their mic, camera, or screen is shared. So I don't think of completely hiding the bottom tool bar.
* I don't want to add complexity. So I don't want to add event handlers (eg: Add button to show/hide tool bar.) And I want consistent design/layout across platforms or device orientation for simplicity. If we change them depending on platforms or device orientation, testing cost will be higher or we will easily ignore some cases in testing.
* Preferred design and layout should be very dependent on admins, scenes, users, events, or use cases. (Actually I have gotten the two feedbacks. Some users prefer always visible icons. Others prefer hide/show tray.) I hope admins or owners edit the design and layout in their custom clients if the default design or layout doesn't satisfy them. So I think it's good that the default is good design and simple implementation for them to edit rather than the default is perfect design and complex implementation.


takahirox:
Ah, wait. I need to fix them...

image

image

takahirox:
Fixed them

takahirox:
Lnadscape mode before and after. I'm not sure if the new one looks more beautiful, but at least more usable with vertically wider viewport.

unknown

themes.json
```diff
@@ -79,7 +79,10 @@
       "toolbar-icon-selected-bg": "#ffffff",
       "toolbar-basic-icon-color": "#ffffff",
       "toolbar-basic-selected-icon-color": "#2B313B",
-      "toolbar-basic-color-hover": "#ffffff",
```
keianhzo, line 82: It seems that removing this variable makes the menu button in dark mode to have the incorrect background on hover.

keianhzo (changes requested):
Other than the style issue that I point out looks good to me and I think it's more usable than what we had.

I think I'd like it a bit better if the labels had more padding around, something like:
Screen Shot 2022-07-05 at 17 57 25
But that's more a design decision.

takahirox:
@keianhzo 

Thanks for the review.

> I think I'd like it a bit better if the labels had more padding around

I hope someone (especially designers or artists) makes a follow up PR to adjust or improve the design.

themes.json
takahirox: Hm, it works fine here as the following pic shows. (The mouse pointer is invisible but it is on the mic button.)

image

`--toolbar-basic-color-hover` refers to `--basic-color-hover` and `-basic-color-hover` is updated for the dark mode in `theme.json`.

src/react-components/room/useResizeViewport.js
```diff
@@ -73,7 +73,7 @@ export function useResizeViewport(viewportRef, store, scene) {
           return;
         }
 
-        const canvasRect = entries[0].contentRect;
+        const canvasRect = { width: window.innerWidth, height: window.innerHeight };
```
takahirox, line 76: @netpro2k @keianhzo Do you think this is a proper way to set window-full-size canvas?

themes.json
keianhzo: Humm I just checkout the branch again and did a npm ci and it seems to be alright. Not sure what happened but looks good.

src/react-components/room/useResizeViewport.js
keianhzo: This looks good to me, I think it should work well but just curious, why did you change `entries[0].contentRect`?

keianhzo: Oh ok, thanks for the explanation. 

The `useResizeViewport` hook takes a node and resizes the viewport according to the latest node size. With this change we are always setting the viewport to the window size which works in this case but breaks the hook contract as we are not resizing based on the `viewportRef` anymore. It works in this case but it might be confusing for people using the hook in the future so I'd rather keeping the hook contract as it was and  pass the right viewport reference to it.

Once the `useResizeViewport`code has been reverted, you can assign the `viewportRef` to the room layout div in `RoomLayout.js`. That div should cover the entire window if I'm right.
```javascript
return (
    <div
      className={classNames(styles.roomLayout, { [styles.objectFocused]: objectFocused }, className)}
      ref={viewportRef}
      {...rest}
    >
      {sidebar && <div className={classNames(styles.sidebar, sidebarClassName)}>{sidebar}</div>}
      <div className={classNames(styles.modalContainer, styles.viewport)}>{modal}</div>
      {(toolbarLeft || toolbarCenter || toolbarRight) && (
        <Toolbar
          className={classNames(styles.main, styles.toolbar, toolbarClassName)}
          left={toolbarLeft}
          center={toolbarCenter}
          right={toolbarRight}
        />
      )}
      <div className={classNames(styles.main, styles.viewport, { [styles.streaming]: streaming }, viewportClassName)}>
        {viewport}
      </div>
    </div>
  );
```

Sorry for missing this on my first review.

keianhzo:
I think we can get rid of this polyfill?

https://github.com/mozilla/hubs/blob/cfa9fbf0ca8ffdbed36d9095c43b0a1855463635/src/react-components/room/useResizeViewport.js#L2-L3

It seems to be well supported in all browsers since a while ago: https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver

takahirox:
Thanks for the suggestion.

> I think we can get rid of this polyfill?

I want to think of it in another PR if needed.

src/react-components/room/useResizeViewport.js
takahirox: Thanks, updated.

keianhzo: You also need to remove the `viewportRef` from the last div otherwise otherwise we can get unexpected behavior.

takahirox: Updated, thanks.

keianhzo (approved):

Issue #5196, opened 2022-02-28
Virtual joysticks by default on mobile devices?


takahirox:
**Is your feature request related to a problem? Please describe.**

Hubs supports virtual joysticks behind the preference.

"Menu" - "Preferences" - "Controls" tab - "Touchscreen" section - "
Enable left on-screen joystick for moving around" and "Enable right on-screen joystick for looking around"

But I often heard that many mobile users prefer joysticks over than the default controls (pinching/dragging).

It may be worth to consider to enable the virtual joysticks by default on mobile devices. Personally I think joysticks is more intuitive than the default controls.

But also I heard some users prefer the default controls and some of them think the pinchin/dragging controls is one of good uniquness of Hubs compared to similar services.

So I would like to suggest to take a survey and do research about how the default controls should be (pinching/dragging or virtual joysticks) and change the default controls if needed.

**Describe the solution you'd like**

Enable joysticks by default on mobile devices.

**Describe alternatives you've considered**

Even if we decide not to change the default controls, there are lot of users who prefer joysticks and may not know the Hubs virtual joysticks support. So at least discoverability with the joysticks needs improvement.

**Additional context**

1. I might be biased to joysticks because I'm a video game fan.
2. It is woth to research the similar services (eg. Roblox) for the controls.

takahirox:
I collected some feedback from users that joysticks controls may be more intuitive for more many mobile users but if pinch/drag controls may be faster and smoother users get used to them. So I'm thinking of the followings instead of changing the default controls.

* Add and show the controls dialog where user chooses the controls, pinch/drag or joysticks, and store it to the preferences when entering a room for mobile users
* Improve the controls tutorial instead of just blue bubble message
* Make a quick controls tutorial video

Praharsh226:
I have collected feedback from various users and following are the key insights : 
1. Joystick is much more intuitive option than pinch/drag controls as most people have played atleast one game on mobile which has similar kind of control
2. The joystick circle should be always visible so that the usage becomes much more intuitive for new users
3. The right joystick circle which controls the view direction should only rotate along the z axis so that user only looks at a certain level

takahirox:
@Praharsh226

Thanks for the feedbacks! 

> 1. Joystick is much more intuitive option than pinch/drag controls as most people have played atleast one game on mobile which has similar kind of control

I think at least we need better disability with the joysticks even if we give up enabling them by default.

> 2. The joystick circle should be always visible so that the usage becomes much more intuitive for new users

[...]

PR #5527, opened 2022-06-24
Enable the virtual joysticks by default on mobiles

takahirox:
#5196 

The virtual joysticks are more intuitive to generic mobile users. And the recent and upcoming joystics and 2D UI improvement, the virtual joysticks become more useful. I would like to suggest to enable the virtual joysticks by default on mobiles.

Of course there are users who prefer the pinching/dragging controls. But probably most people who prefer the pinching/dragging controls may already got used to Hubs and easily find joystickswith the preferences. I hope it's acceptable.


netpro2k (commented):
Code change LGTM, no strong opinions on the defaults from me, not sure if anyone else feels more strongly.

PR #5529, opened 2022-06-25
Adding feature to block access and/or chat functionality (mozfest features)

ActiveReplica:
In Room settings, you have now two new toggles, one to block chat (both button and side bar) and one to block access (allowing only to spectate) for new comers.

blockaccesschat

Both features are an updated version of those being used during MozFest.

### DIFF
--- a/src/hub.html
+++ b/src/hub.html
@@ -1003,7 +1003,7 @@
         </a-assets>
 
         <!-- Interactables -->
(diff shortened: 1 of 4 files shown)


matthewbcool:
Thank you for this contribution. Although there is no timeline on this being reviewed and merged into Hubs main, I appreciate the added chat/room blocking moderation control for admins. A great reference for custom clients that would like these features as well!

Issue #4832, opened 2021-11-11
Reliable crash using fast-switching on Android Chrome


rawnsley:
When fast-switching between rooms on mobile, the web page eventually crashes. Before it crashes completely, some of the textures in the scene fail to load. This implies it's some sort of memory exhaustion or fragmentation.

This [test room](https://hubs.mozilla.com/z7J9foR/easygoing-used-huddle) links to another room, which in turn links back to it. If you follow the links a number of times (about ten in testing) the problem should emerge.

It is possible that this is also an issue on desktop, but the available memory is much higher. I haven't seen it in testing though, and I transit between a large number of heavy scenes frequently.

Test environment was Chrome 95.0.4638.74 on Android 10 on a POCOPHONE F1.

yakyouk:
This is reproducible as well on Oculus Quest, and causes the whole headset to malfunction, with the Quest menus appearing black, and the headset needs to be forcefully turned off. This seems to point towards *vram*  exhaustion.

rawnsley:
Relevant article that might be useful for investigating this issue: https://nolanlawson.com/2022/01/05/memory-leaks-the-forgotten-side-of-web-performance/

rawnsley:
I can reproduce at least one significant leak. Here is [a test room](https://dev.reticulum.io/Fyy6RPc) with a single link through to another test room that contains a large model. This room then links back to the first room.

If you enable fast room switching and follow the first link, then the second so that you end up back in the original room, you can see some significant (50MB) leaked allocations in the heap snapshot:

<img width="1078" alt="Screenshot 2022-06-27 at 11 29 01" src="https://user-images.githubusercontent.com/303516/175921402-328e5481-cbdc-40de-abe7-9def55be42e0.png">

These leaks multiply every time you repeat the process.

The most visible leak is due to HTML elements being registered with various global maps in [App.js](https://github.com/mozilla/hubs/blob/master/src/App.js) and not being removed when changing rooms.

The next most visible of these are from stray _networkEl_ properties not being nulled out when the component is removed, although there might be more pertinent roots below that (something must be keeping the component in scope). Either way, it's probably best practice to do this if the component has set _networkEl_ at the very least to simplify leak investigations. An example assignment that should be undone on remove: https://github.com/mozilla/hubs/blob/master/src/components/media-loader.js#L73

[...]

PR #5531, opened 2022-06-27
Improved reference management to reduce memory leaks

rawnsley:
Fast room switching [exposes a number of issues](https://github.com/mozilla/hubs/issues/4832) with reference management that lead to large memory leaks and ultimately instability, particularly on mobile platforms.

One root cause was the use of APP level globals for variables that should really be bound to the scope of the room. Rather than make any radical changes, this PR continues to rely on `AudioSettingsSystem` to reset globals when the scene changes.

The _networkedEl_ NAF parameter is now reset to null in the few components where it was being assigned. There weren't any specific leaks associated with this, but it does confuse the process of tracking down actual leaks.

The most insidious leak was in the media-video component, where captured objects and captured functions appear to confound the garbage collector. Explicitly nulling out some of the references fixes the problem.

I've tried to keep the changes small and positive in and of themselves. I expect there are some edge cases where leaks can still happen, particularly where timing is a factor.

I will keep this PR in draft until I have tested the changes more widely on my own Hubs Cloud. All suggestions are welcome.

### DIFF
--- a/src/components/action-to-remove.js
+++ b/src/components/action-to-remove.js
@@ -18,5 +18,9 @@ AFRAME.registerComponent("action-to-remove", {
     if (this.data.requireOwnership && this.networkedEl && !NAF.utils.isMine(this.networkedEl)) return;
 
     this.el.parentNode.removeChild(this.el);
+  },
+
+  remove() {
+    this.networkedEl = null;
   }
 });
--- a/src/components/media-loader.js
+++ b/src/components/media-loader.js
@@ -701,6 +701,7 @@ AFRAME.registerComponent("media-pager", {
     if (this.networkedEl) {
       this.networkedEl.removeEventListener("pinned", this.update);
       this.networkedEl.removeEventListener("unpinned", this.update);
+      this.networkedEl = null;
     }
 
     window.APP.hubChannel.removeEventListener("permissions_updated", this.update);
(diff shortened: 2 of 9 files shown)


vincentfretin:
FYI I did #5510 to fix disposing of the geometry and material of the simple-water component. There may be similar issues in other shader components, I didn't check.

rawnsley:
The first test of this on my own Hubs Cloud code isn't very encouraging and I'm still seeing apparent memory leaks. I took the (now obvious) step of just filtering the snapshot results for "Detatched" objects and it looks like this:

Screenshot 2022-06-30 at 10 10 50

I'm already out of my depth on large web app development, but let me swim out a little further by saying that there should be no detached objects references after garbage collection.

flankechen:
@rawnsley thanks for your report, by "It appears that at least one major leak was taken care of by @netpro2k during his recent refactoring. It was due to an attachment to event "camera-set-active" that was never removed." , did you mean this merge [New Entity Framework](https://github.com/mozilla/hubs/pull/5536) ? it seems hard to find "camera-set-active" related code.

rawnsley:
@flankechen I think it was [this commit](https://github.com/mozilla/hubs/commit/5f8229a38c669a1b7053860eaf1aa5ee7e8e82c0#diff-7f28f4412d51a4907de1643b1af1acc1dc988c71a92e16ee7f69efb96cf7f189) that removes the event attachment.

rawnsley:
I'm withdrawing this PR as the main branch has drifted significantly following the migration to bitECS, which should be inherently less leaky anyway.

PR #5535, opened 2022-06-27
Release texture image more properly

takahirox:
`HubsTextureLoader` removes the reference to texture image for GC by using `delete texture.image` once it is uploaded.

But `delete texture.image` should be replaced with `texture.image = null` because the latter follows the Three.js API better.


netpro2k (approved):
This also potentially has perf benefits in terms of JS engine not having to create a new internal representation for the texture object.

PR #5537, opened 2022-06-29
Improve useVolume refresh to avoid react re-renders

keianhzo:
This issue was reintroduced in this PR https://github.com/mozilla/hubs/pull/5141 but I never improved the re-renders then so I'm doing it now after Dom run into this recently.

For context: https://github.com/mozilla/hubs/pull/5014#discussion_r788427276

### DIFF
--- a/src/react-components/input/ToolbarMicButton.js
+++ b/src/react-components/input/ToolbarMicButton.js
@@ -6,30 +6,25 @@ import { useVolumeMeter } from "../misc/useVolumeMeter";
 
 export function ToolbarMicButton({ scene, ...rest }) {
   const ref = useRef();
-  const setRef = useCallback(
-    node => {
-      ref.current = node;
-    },
-    [ref]
-  );
-  const update = useCallback(
-    level => {
-      const node = ref.current;
-      if (node) {
-        const rect = node.querySelector("rect");
-        if (level <= 0.1) {
-          rect.setAttribute("height", 0);
-        } else if (level < 0.3) {
-          rect.setAttribute("y", 8);
-          rect.setAttribute("height", 4);
-        } else {
-          rect.setAttribute("y", 4);
-          rect.setAttribute("height", 8);
-        }
+  const setRef = useCallback(node => {
+    if (node) {
+      ref.current = node.querySelector("rect");
+    }
+  }, []);
+  const update = useCallback(level => {
+    const rect = ref.current;
+    if (rect) {
+      if (level <= 0.1) {
+        rect.setAttribute("height", 0);
+      } else if (level < 0.3) {
+        rect.setAttribute("y", 8);
+        rect.setAttribute("height", 4);
+      } else {
+        rect.setAttribute("y", 4);
+        rect.setAttribute("height", 8);
       }
-    },
-    [ref]
-  );
+    }
+  }, []);
   useVolumeMeter({
     analyser: scene.systems["hubs-systems"].audioSystem.outboundAnalyser,
     update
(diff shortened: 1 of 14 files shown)


src/react-components/input/ToolbarMicButton.js
```diff
@@ -0,0 +1,50 @@
...
+    },
+    [ref]
+  );
+  useVolumeMeter({
+    analyser: scene.systems["hubs-systems"].audioSystem.outboundAnalyser,
+    update
+  });
+  return (
+    <ToolbarButton
+      ref={setRef}
```
brianpeiris, line 39: Pretty sure `setRef` is redundant. You can use `ref` directly here.
```suggestion
      ref={ref}
```

src/react-components/input/ToolbarMicButton.js
```diff
@@ -0,0 +1,50 @@
...
+        } else if (level < 0.3) {
+          rect.setAttribute("y", 8);
+          rect.setAttribute("height", 4);
+        } else {
+          rect.setAttribute("y", 4);
+          rect.setAttribute("height", 8);
+        }
+      }
+    },
+    [ref]
```
brianpeiris, line 31: `ref` should never change, so you can have this `useCallback` use an empty dependency array.
```suggestion
    []
```

src/react-components/misc/VolumeLevelBar.js
```diff
@@ -0,0 +1,46 @@
...
+    [ref]
+  );
+  useVolumeMeter({
+    analyser:
+      type == "mic"
+        ? scene.systems["hubs-systems"].audioSystem.outboundAnalyser
+        : scene.systems["hubs-systems"].audioSystem.mixerAnalyser,
+    update
+  });
+  return <LevelBar ref={setRef} className={classNames(styles.levelBarContainer, className)} />;
```
brianpeiris, line 39: Same deal here.
```suggestion
  return <LevelBar ref={ref} className={classNames(styles.levelBarContainer, className)} />;
```

src/react-components/input/ToolbarMicButton.js
```diff
@@ -0,0 +1,50 @@
...
+    node => {
+      ref.current = node;
+    },
+    [ref]
+  );
+  const update = useCallback(
+    level => {
+      const node = ref.current;
+      if (node) {
+        const rect = node.querySelector("rect");
```
brianpeiris, line 19: Not an issue with this PR, since this was existing code, but we may want to look into the cost of this `querySelector` call at some point. It's unfortunate that we need to query for the rect every update.

src/react-components/room/MicSetupModal.scss
```diff
@@ -1,4 +1,4 @@
-@use "../styles/theme.scss";
+  @use "../styles/theme.scss";
```
brianpeiris, line 1: Looks like an accidental change here.

brianpeiris (approved):
Spotted some simplifications, but LGTM otherwise.

src/react-components/input/ToolbarMicButton.js
keianhzo: Oh yep that would be only necessary in case of using an effect but this is just a timed callback so it should be fine. 

Thinking about your later comment about removing the rect query, I think we can probably keep this and use it for gathering the rect once the node is ready to avoid the repeated querySelector.

src/react-components/input/ToolbarMicButton.js
keianhzo: Removed

src/react-components/misc/VolumeLevelBar.js
keianhzo: Removed setRef

src/react-components/input/ToolbarMicButton.js
keianhzo: Addressed int he previous comment using setRef.

PR #5538, opened 2022-06-29
Hubstrky 446

nickgrato:
Jira ticket [446 ](https://mozilla-hub.atlassian.net/browse/HUBSTRKY-446?atlOrigin=eyJpIjoiN2QyNzcwMTY1MzVkNDNmODgxMTY5OTA0YjUwYzBiZjUiLCJwIjoiaiJ9)

Adding Banner for email sign up for MVP2 launch


src/react-components/banner/Banner.js
```diff
@@ -0,0 +1,321 @@
...
+import { ReactComponent as ArrowIcon } from "../icons/Arrow.svg";
+import warning_icon from "../../assets/images/warning_icon.png";
+import { TextInputField } from "../input/TextInputField";
+import { CheckboxInput } from "../input/CheckboxInput";
+import { FormattedMessage } from "react-intl";
+import cone from "../../assets/images/cone.png";
+
+const Banner = () => {
+  const [email, setEmail] = useState("");
+  const [confirm, setConfirm] = useState(true);
```
brianpeiris, line 17: Should this be `false` by default, so that users can opt-in to consent?

src/react-components/banner/Banner.js
```diff
@@ -0,0 +1,321 @@
...
+    // Check Target Status
+    const status = resp.target.status;
+    if (status !== 200 || status > 300 || status < 200) {
+      newsletterError();
+      return;
+    }
+
+    // Check Response Status
+    let response = resp.target.response;
+    typeof response !== "object" ? (response = JSON.parse(response)) : "";
```
brianpeiris, line 58: Might want to wrap this JSON.parse in a `try/catch`

src/react-components/banner/Banner.js
```diff
@@ -0,0 +1,321 @@
...
+    // Note: this is a pattern that was suggesed by the basket creator
+    const url = "https://basket.mozilla.org/news/subscribe/";
+    const xhr = new XMLHttpRequest();
+    const params =
+      "email=" +
+      encodeURIComponent(data.email) +
+      "&newsletters=" +
+      "hubs" +
+      "&lang=" +
+      encodeURIComponent(navigator.languages[0]) +
```
brianpeiris, line 78: `navigator.language` is simpler and available more browser versions.
```suggestion
      encodeURIComponent(navigator.language) +
```

src/react-components/banner/Banner.js
```diff
@@ -0,0 +1,321 @@
...
+      "&source_url=" +
+      encodeURIComponent(window.location);
```
brianpeiris, line 80: We may want to avoid sending the entire URL here, since the query parameters may contain arbitrary information.

src/react-components/banner/Banner.js
```diff
@@ -0,0 +1,321 @@
...
+    xhr.timeout = 5000;
+    xhr.ontimeout = newsletterError;
+    xhr.responseType = "json";
+    xhr.send(params);
+  };
+
+  /**
+   * Emain input change
+   */
+  const onChangeEmail = useCallback(
```
brianpeiris, line 95: Out of curiosity, why `useCallback` on this handler?

src/react-components/banner/Banner.js
```diff
@@ -0,0 +1,321 @@
...
+
+  /**
+   * Checkbox Label
+   * @returns JSX
+   */
+  const Label = () => {
+    return (
+      <>
+        I&apos;m okay with Mozilla handling my info as explained in this{" "}
+        <a href="https://github.com/mozilla/hubs/blob/master/PRIVACY.md" rel="noopener noreferrer" target="_blank">
```
brianpeiris, line 117: Maybe this should link to
https://www.mozilla.org/en-US/privacy/websites/
or 
https://www.mozilla.org/en-US/privacy/hubs/

src/react-components/banner/Banner.js
```diff
@@ -0,0 +1,321 @@
...
+  };
+
+  /**
+   * Success Message
+   * @returns JSX
+   */
+  const Success = () => {
+    return (
+      <div className={styles.message_wrapper}>
+        <img src={cone} className={styles.success_image} />
```
brianpeiris, line 140: Probably best to use `alt=""` here at least. Also on the warning image below.

src/react-components/banner/Banner.js
```diff
@@ -0,0 +1,321 @@
...
+        </Button>
+      </div>
+
+      {/* EXPAND CONTAINER  */}
+      {isExpanded ? (
+        <div className={styles.expand_wrapper}>
+          <div className={styles.expand_container}>
+            <div className={styles.expand_header}>
+              <h2 className={styles.expand_title}>
+                <FormattedMessage defaultMessage="Join the next evolution of Hubs!" id="banner.expand_title" />
```
brianpeiris, line 241: Use a hyphen to keep these ids consistent.
```suggestion
                <FormattedMessage defaultMessage="Join the next evolution of Hubs!" id="banner.expand-title" />
```

src/react-components/banner/Banner.js
```diff
@@ -0,0 +1,321 @@
...
+                  <div>
+                    <TextInputField
+                      ref={register}
+                      name="email"
+                      type="email"
+                      label={<FormattedMessage id="email-address" defaultMessage="Email Address" />}
+                      required
+                      value={email}
+                      onChange={onChangeEmail}
+                      placeholder="example@example.com"
```
brianpeiris, line 276: nit: This might be a bit nicer.
```suggestion
                      placeholder="name@email.com"
```

src/react-components/banner/Banner.js
```diff
@@ -0,0 +1,321 @@
...
+              <form className={styles.expand_form} onSubmit={handleSubmit(onSubmit)}>
+                {submitted ? (
+                  <Messaging status={responseStatus} />
+                ) : (
+                  <div>
+                    <TextInputField
+                      ref={register}
+                      name="email"
+                      type="email"
+                      label={<FormattedMessage id="email-address" defaultMessage="Email Address" />}
```
brianpeiris, line 272:
```suggestion
                      label={<FormattedMessage id="banner.email-address" defaultMessage="Email Address" />}
```

src/react-components/banner/Banner.js
```diff
@@ -0,0 +1,321 @@
...
+                      required
+                      value={email}
+                      onChange={onChangeEmail}
+                      placeholder="example@example.com"
+                      className={styles.expand_form_field}
+                    />
+
+                    <RadioInputField className={styles.expand_form_field} label="Format">
+                      <RadioInputOption name="email_format" value="html" label="HTML" ref={register} />
+                      <RadioInputOption name="email_format" value="text" label="Text" ref={register} />
```
brianpeiris, line 282: I think you can just use `<FormattedMessage />` for the values of the `label` attributes, like you did for "Email Address" above.

src/react-components/banner/Banner.js
```diff
@@ -0,0 +1,321 @@
...
+
+    // Do success stuff here..
+    newsletterSuccess();
+  };
+
+  const onSubmit = async data => {
+    // Note: this is a pattern that was suggesed by the basket creator
+    const url = "https://basket.mozilla.org/news/subscribe/";
+    const xhr = new XMLHttpRequest();
+    const params =
```
brianpeiris, line 72: The form has the `email_format` radio buttons, but you don't actually use the value here.

brianpeiris (approved):

brianpeiris (commented):
It seems I accidentally chose "Approve" on the last review, when I meant to chose "Comment". Let's see if I can correct that by submitting a new review.

Edit: Apparently not.

src/react-components/banner/Banner.js
```diff
@@ -0,0 +1,321 @@
...
+import { Button } from "../input/Button";
+import { CloseButton } from "../input/CloseButton";
+import { useForm } from "react-hook-form";
+import styles from "./Banner.scss";
+import { ReactComponent as ArrowIcon } from "../icons/Arrow.svg";
+import warning_icon from "../../assets/images/warning_icon.png";
+import { TextInputField } from "../input/TextInputField";
+import { CheckboxInput } from "../input/CheckboxInput";
+import { FormattedMessage } from "react-intl";
+import cone from "../../assets/images/cone.png";
```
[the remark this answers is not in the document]
nickgrato: I would agree but the size of it in the design is actually pretty large i think i have it at like 175 px height / width. I can talk to kate about knocking it down it time permits though

src/react-components/banner/Banner.js
nickgrato: ops yep let's add

src/react-components/banner/Banner.js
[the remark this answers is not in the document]
brianpeiris: Yeah, not a big deal. There a bunch of large images already on the home page anyway, so this wouldn't make much of a difference. Just trying to keep optimization in mind. We should do a pass all the other front page assets before we do any marketing pushes.

src/react-components/banner/Banner.js
nickgrato: As a rule of thumb whenever prop-ing a function into a child component we should always use useCallback. This helps with unnecessary re-rendering. Components that receive primitive data types on re-render will treat the data the same. So like ok, it was a string before and it's the same string now, no re-render. But for reference types  even though its the same method the reference point changed and causes the child component to think it got a new function and it didn't.

src/react-components/banner/Banner.js
```diff
@@ -0,0 +1,349 @@
...
+          className={styles.success_image}
+          alt={<FormattedMessage defaultMessage="celebrate" id="banner.celebrate-alt" />}
+        />
+        <div>
+          <h3 className={styles.message_title}>
+            <FormattedMessage defaultMessage="You're on the list" id="banner.success-title" />
+          </h3>
+          <p className={styles.message_body}>
+            <FormattedMessage
+              defaultMessage="Keep an eye out for product updates and an invite to join us as a tester in August."
```
misslivirose, line 152: Talked with @emclaren - can we drop "in August" and just go "Keep an eye out for product updates and an invite to join us as a tester"?

brianpeiris (approved):
LGTM after resolving Liv's comment.

src/react-components/banner/Banner.js
nickgrato: yea those  should be too! updating.

PR #5540, opened 2022-06-29
Use low quality audio panning mode for Android

takahirox:
From #5057

`WebAudio` on Android devices (only non-VR devices?) seems to have a bug and audio can be broken if there are many people in a room.

We have reported the problem to the Android devs. We found that using equal power panning mode can mitigate the problem so I would like to suggest that we use low audio panning quality (= equal power mode) by default on Android as workaround until the root issue is fixed on Android end.


brianpeiris (approved):
LGTM

PR #5547, opened 2022-06-30
Remove all event handler when component is removed

keianhzo:
The `media-pager` component is not correctly removing the listeners when it's removed and this causes issues when media element is refreshed.

STRs:
- Add a PDF with multiple pages to the scene
- Turn a couple of pages
- Refresh it a couple of times
- Turn a page again

Result: Multiple pages are turned at the same time
Expected: Only one page should be turned


brianpeiris (approved):
Nice catch. LGTM!

Issue #2599, opened 2020-06-30
Video audio plays at full volume during asset loading stage


JacobErvin:
**Description**
Video assets play at full volume during asset loading stage, causing a loading to be very disruptive. May happen with standard audio assets too. 

**To Reproduce**
1. Have a video in Spoke. 
2. Publish
3. Load. 

**Expected behavior**
Audio assets don't play when scene is loading. 

**Hardware** 
 - Device: Desktop
 - OS: Mac
 - Browser: Firefox
- Other: Hubs Cloud

InfiniteLee:
@JacobErvin When you say "during asset loading stage" - do you mean when loading in a Hubs room (e.g. url: "https://hubs.mozilla.com/<room_sid>/<random_room_name>" or in a scene preview (e.g. url: "https://hubs.mozilla.com/scenes/<scene_sid>")?

JacobErvin:
This happens when loading into a room. I can try to grab a video recording of it a little later today. FYI - since this is a Hubs Cloud instance it's actually <customserver.com><room_sid>/<random_room_name>

Issue #4739, opened 2021-10-14
Apply Audio Zones also during spectator mode


keianhzo:
**Description**
Audio zones are now enabled only when the user enters the room and is embodied as we rely on the avatar-audio-source to track the listener position.

**Describe the solution you'd like**
We should apply Audio Zones also when the user is in spectator mode to better deliver the room audio design experience in all cases.

Issue #2814, opened 2020-08-07
No room sound until you enter, or choose to watch from lobby


emclaren:
I'm going through surveys from the XR access symposium, and feedback from a researcher who did a walk through of Hubs and there are a number of comments about sound being too much on first load. If I recall correctly, we've had people at the meetup flag similar issues. 

I believe sound specialization may not kick in right away, and it can mean that the voices/audio are very loud at first when you open a Hubs room, and then quiet down. If the default lobby position is right next to a video or an ongoing conversation it is continues to b every loud as you go through the steps to enter. 

I wonder if it might be a good idea to not play room audio at first, if we are doing a redesign of the entry flow it might be worthwhile having audio only start once you have decided to enter the room, or spectate from the lobby.

emclaren:
#2599 appears to be related to the loud audio during load

NearlyNorman:
I am having the same problem where all my audio files play at once when I open my husb and hang around in the lobby

PR #5550, opened 2022-06-30
Media audio can be very loud on room entry and scene changes

rawnsley:
When you enter a room with audio or video media there can be a short period of very load audio output. The principle reason for this is that when the HTML Audio and Video elements are created there can be a delay before the various audio systems adopt them and start controlling their volumes. This has been addressed by setting the volume to zero by default and then reseting it to one once control has been passed over.

A secondary issue is that the default audio gain for the THREE sources is one and volume is interpolated to it's proper value over time using `setTargetAtTime`. If the gain is set to zero by default then the audio will fade up rather than down, which is less jarring.

The last issue is that before the room is entered, the player is positioned at the scene preview camera position and the audio will be set accordingly. This PR does not address this issue because it is not clear what the desired behaviour is.

Here is a test room that demonstrates all of these elements: [https://dev.reticulum.io/2D9w2Y4](https://dev.reticulum.io/2D9w2Y4). Without the PR applied you will hear a loud blast on first load, followed by the volume decreasing because the preview position in near the audio source, then on scene entry the volume will drop to being low. With the PR you will not have the initial blast of sound and instead jump straight to the volume of the preview sound.

Note that you might need to click the page a bit during load to tell the browser you are happy to hear audio at all. While this would appear to make the PR less relevant, remember that this approval is already given in the case of scene changes or fast room switching, where the sound blast can be significant.

I'm happy to take advice on whether I'm using the right properties in the right places, but I hope the underlying principles are sound.

### DIFF
--- a/src/components/avatar-audio-source.js
+++ b/src/components/avatar-audio-source.js
@@ -61,6 +61,9 @@ AFRAME.registerComponent("avatar-audio-source", {
     } else {
       audio = new THREE.Audio(audioListener);
     }
+    // Default to being quiet so it fades in when volume is set by audio systems
+    audio.gain.gain.value = 0;
+
     this.audioSystem.addAudio({ sourceType: SourceType.AVATAR_AUDIO_SOURCE, node: audio });
 
     if (SHOULD_CREATE_SILENT_AUDIO_ELS) {
--- a/src/components/media-video.js
+++ b/src/components/media-video.js
@@ -358,6 +358,8 @@ AFRAME.registerComponent("media-video", {
     } else {
       this.audio = new THREE.Audio(audioListener);
     }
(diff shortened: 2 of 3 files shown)


rawnsley:
Fixes https://github.com/mozilla/hubs/issues/2599 and https://github.com/mozilla/hubs/issues/2814,

keianhzo (approved):
Thanks for this, it looks good to me. This is a small simple patch and I can see the benefit of it.

Related to this and the last issue that you pointed out is the preview + audio zones case where we are currently not applying the audio zones if the avatar is not yet spawned. We should probably apply it to the preview point so the behavior is more consistent. I opened an issue a while ago for this: https://github.com/mozilla/hubs/issues/4739

PR #5553, opened 2022-07-01
Don't use avatar-rig for audio-zone tracking if not entering scene yet

wsxiaoys:
Otherwise, getOwnerId will fails with following message if audio source captured the player in lobby.

`Could not find networked el. Entity does not have and is not a child of an entity with the [networked] component`


netpro2k:
Thanks for the PR! Sorry this took  ages to merge.

PR #5574, opened 2022-07-07
Extract only required references at Sketchfab Zip worker

### LINKED ISSUE #5469: glTF content from Sketchfab can memory leak
**Description**

glTF content from Sketchfab can memory leak

* [`scene.gltf` file content is loaded as blob to `fileMap["scene.gltf
]`](https://github.com/mozilla/hubs/blob/01d8e31dbdf1f300174f0fdc17dc9d5cb4f73580/src/workers/sketchfab-zip.worker.js#L9)
* [`fileMap["scene.gltf"]` has a new value](https://github.com/mozilla/hubs/blob/01d8e31dbdf1f300174f0fdc17dc9d5cb4f73580/src/workers/sketchfab-zip.worker.js#L15)
* [The blob resource is not released here](https://github.com/mozilla/hubs/blob/01d8e31dbdf1f300174f0fdc17dc9d5cb4f73580/src/components/gltf-model-plus.js#L675)

**Expected behavior**

No memory leak

**Hardware** 
 - Device: Any
 - OS: Any
 - Browser: Any

**Additional context**

Probably the easiest solution would be not load `scene.gltf` as the blob in the loop because it doesn't seem to be used.

```javascript
for (const fileName in zip.files) {
  if (fileName !== "scene.gltf") {
    fileMap[fileName] = zip.extractAsBlobUrl(fileName);
  }
}
```

sarhugo:
Fixes #5469

### DIFF
--- a/src/workers/sketchfab-zip.worker.js
+++ b/src/workers/sketchfab-zip.worker.js
@@ -5,13 +5,13 @@ async function fetchZipAndGetBlobs(src) {
   return zip.load().then(() => {
     // Rewrite any url references in the GLTF to blob urls
     const fileMap = {};
-    for (const fileName in zip.files) {
-      fileMap[fileName] = zip.extractAsBlobUrl(fileName);
-    }
-
+    const rewriteUri = uri => {
+      fileMap[uri] = fileMap[uri] || zip.extractAsBlobUrl(uri);
+      return fileMap[uri];
+    };
     const gltfJson = JSON.parse(zip.extractAsText("scene.gltf"));
-    gltfJson.buffers && gltfJson.buffers.forEach(b => (b.uri = fileMap[b.uri]));
-    gltfJson.images && gltfJson.images.forEach(i => (i.uri = fileMap[i.uri]));
+    gltfJson.buffers && gltfJson.buffers.forEach(b => (b.uri = rewriteUri(b.uri)));
+    gltfJson.images && gltfJson.images.forEach(i => (i.uri = rewriteUri(i.uri)));
     fileMap["scene.gtlf"] = URL.createObjectURL(new Blob([JSON.stringify(gltfJson, null, 2)], { type: "text/plain" }));
 
     return fileMap;


takahirox (approved):
Looking good to me, thanks.

src/workers/sketchfab-zip.worker.js
```diff
@@ -5,13 +5,13 @@ async function fetchZipAndGetBlobs(src) {
   return zip.load().then(() => {
     // Rewrite any url references in the GLTF to blob urls
     const fileMap = {};
-    for (const fileName in zip.files) {
-      fileMap[fileName] = zip.extractAsBlobUrl(fileName);
-    }
-
+    const rewriteUri = uri => {
+      fileMap[uri] = fileMap[uri] || zip.extractAsBlobUrl(uri);
```
takahirox, line 9: Note to other reviewers: I haven't checked the `ZipLoader` API yet but I assume the return value from `zip.extractAsBlobUrl(uri)` is not `null` or emptry strings.

src/workers/sketchfab-zip.worker.js
```diff
@@ -5,13 +5,13 @@ async function fetchZipAndGetBlobs(src) {
...
+      return fileMap[uri];
+    };
     const gltfJson = JSON.parse(zip.extractAsText("scene.gltf"));
-    gltfJson.buffers && gltfJson.buffers.forEach(b => (b.uri = fileMap[b.uri]));
-    gltfJson.images && gltfJson.images.forEach(i => (i.uri = fileMap[i.uri]));
+    gltfJson.buffers && gltfJson.buffers.forEach(b => (b.uri = rewriteUri(b.uri)));
+    gltfJson.images && gltfJson.images.forEach(i => (i.uri = rewriteUri(i.uri)));
     fileMap["scene.gtlf"] = URL.createObjectURL(new Blob([JSON.stringify(gltfJson, null, 2)], { type: "text/plain" }));
```
takahirox, line 15: Note to other reviewers: I assume the file name `scene.gtlf` is not in `gltfJson.buffers` or `gltfJson.images`.

takahirox: Note to other reviewers: Probably `scene.gtlf` should be renamed to `scene.gltf` but I want to do it in another PR because a single PR should include a single change logically.

src/workers/sketchfab-zip.worker.js
sarhugo: Checking [ziploader](https://github.com/yomotsu/ZipLoader/blob/master/src/ZipLoader.js#L89) It always returns a blob url. Also it already keeps track of extracted uris so we may simplify here just calling the ziploader method

takahirox: Thanks for checking.

> Also it already keeps track of extracted uris so we may simplify here just calling the ziploader method

The current `fileMap[uri] = fileMap[uri] || zip.extractAsBlobUrl(uri);` is ok to me. `it already keeps track of extracted uris` is ZipLoader internal design. There might be a chance they ZipLoader will change it in the future version.

takahirox:
Thanks for the contribution!

PR #5577, opened 2022-07-08
Fix darkened camera view

### LINKED ISSUE #5571: Camera view is darkened
**Description**

Camera view is darkened. Probably it was broken by the recent our Three.js upgrade #5488

**To Reproduce**
1. Go to https://hubs.mozilla.com
2. Create and enter a room
3. Switch the scene to Gathering Hall
4. Place a camera
5. Confirm that camera view is darkened

**Expected behavior**

Camera view is not darkened.

**Screenshots**

image

**Hardware** 
 - Device: Desktop but probably any
 - OS: Windows 10 but probably any
 - Browser: Chrome but probably any

takahirox:
Fixes #5571

The recent `Three.js` relies on `WebGL` for sRGB - Linear color space conversion. But it still does it in shader for video textures for a performance reason.

`camera-tool` sets `isVideoTexture = true` for the screen render target texture to update the render target only when the screens are in view as hack. But this hack causes a wrong color space conversion and the camera views are darkened.

This PR resolves the problem by removing the hack and rewriting it more properly.

### DIFF
--- a/src/components/camera-tool.js
+++ b/src/components/camera-tool.js
@@ -127,9 +127,8 @@ AFRAME.registerComponent("camera-tool", {
     });
     material.toneMapped = false;
 
-    // Bit of a hack here to only update the renderTarget when the screens are in view
-    material.map.isVideoTexture = true;
-    material.map.update = () => {
+    // Only update the renderTarget when the screens are in view
+    const onBeforeRender = () => {
       if (this.showCameraViewfinder) {
         this.viewfinderInViewThisFrame = true;
       }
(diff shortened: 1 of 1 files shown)


src/components/camera-tool.js
```diff
@@ -128,8 +128,7 @@ AFRAME.registerComponent("camera-tool", {
     material.toneMapped = false;
 
     // Bit of a hack here to only update the renderTarget when the screens are in view
```
netpro2k, line 130: We can remove this comment now

src/components/camera-tool.js
```diff
@@ -195,12 +194,14 @@ AFRAME.registerComponent("camera-tool", {
...
       this.screen.rotation.set(0, Math.PI, 0);
       this.screen.position.set(0, 0, -0.042);
+      this.screen.onBeforeRender = onBeforeRender;
       this.screen.matrixNeedsUpdate = true;
       this.el.setObject3D("screen", this.screen);
 
       this.selfieScreen = new THREE.Mesh(geometry, material);
       this.selfieScreen.position.set(0, 0.4, 0);
       this.selfieScreen.scale.set(-2, 2, 2);
+      this.selfieScreen.onBeforeRender = onBeforeRender;
```
netpro2k, line 204: Hmm don't remember why we didn't do this originally... I bet we were doing the actual rendering work in the `update` callback before and we didn't want to do it twice so we applied it to the material instead of the object. This definitely makes more sense now though given its only setting a flag anyway.

netpro2k (approved):
This LGTM. I rewrote the camera system in the new entity system PR so I will need to apply this same change there. I think it still makes sense to ship this.

src/components/camera-tool.js
takahirox: Not sure if it is a hack but I think `only update the renderTarget when the screens are in view` comment is still helpful. What do you think?

Issue #4810, opened 2021-11-04
The prompt to Sign in is wrongly displayed when trying to access certain options after a sign in performed through options that require a login 


Softvision-RemusDranca:
**[Affected Versions]:**
- Safari 15
- Firefox Release 94.0.1
- Chrome 95

**[Affected Platforms]:**
- MacOS Big Sur 11.6
- Windows 10x64

**[Prerequisites]:**
- Have a room created on the "https://hubs.mozilla.com/" website.
- The user is not logged in.
- Have a valid email account.

**[Steps to reproduce]:**
1. Import an object inside the room.
2. Try to pin the object.
3. Login trough the prompt displayed.
4. Navigate for example to Favorite Rooms and select a room.
5. Observe the behavior.

**[Expected results]:**
- The confirmation Leave Room prompt is displayed.

**[Actual results]:**
- The prompt to Log In is displayed.

**[Notes]:**
- This issue is also reproducible after a Login performed from the other locations that need login to be accessed.
- The issue is not reproducible when the Login is done through the normal flow.
- The Create Room and Close Room functionalities are also affected by this issue.
- The Pin/Unpin, Add room to Favorites, Room Settings functionalities are not affected by this issue.
- As a side effect of this issue, if a log out is performed after reproducing it, the Sign In prompt is displayed when trying to play media, move unpined objects, the Pin/Unpin states get mixed between Object View and Object Menu.
- Attached is a screen recording of the issue:
![SignIn](https://user-images.githubusercontent.com/40061643/140389310-7d67741d-4602-4d83-af85-9248f0237d4a.gif)

takahirox:
`performConditionalSignIn()` or its usage is suspicious. It calls `remountUI` with `onSignInDialogVisibilityChanged` callback which can show the sign in dialog.

https://github.com/mozilla/hubs/blob/ac2c2af5df75f0362fa2c65e6db0ec1cb2c65b4a/src/hub.js#L795-L801

`showNonHistoriedDialog` in `ui-root` which is called on leave room button click calls the callback.

https://github.com/mozilla/hubs/blob/7085199dc9e7f317ef6880b7170d7aa114a3eb61/src/react-components/ui-root.js#L663

Issue #4695, opened 2021-10-01
The “You are now signed in” prompt can’t be dismissed


Softvision-GeluHaiduc:
**[Affected Versions]:**
- Firefox Release 92.01
- Chrome 94.0.4606.61

**[Affected Platforms]:**
- Windows 10x64

**[Prerequisites]:**
- Have a new room created and opened on ”https://hubs.mozilla.com/”.
- The user is not signed in.
- Have a valid email account.

**[Steps to reproduce]:**
1. Place an object in the room.
2. Hover over the object, hold the space key and click the “Pin” button.
3. Enter the email from prerequisites in the prompt displayed and click the “Next" button.
4. Close or click the “Cancel” button on the prompt displayed, informing that an email was sent.
5. Go to your Inbox and click the link received, to log into Hubs.
6. Go back to Hubs and try to dismiss the “You are now signed in” prompt.

**[Expected results]:**
- The prompt is dismissed. 

**[Actual results]:**
- The prompt can’t be dismissed.

**[Notes]:**
- If the prompt that informs the user that an email was sent is not closed or canceled, the prompt can be dismissed as expected.
- This is reproducible for every action that requires a user to log in (Pin objects, Acces Room Settings, Favorite Room, Close Room from More menu or Options from Lobby.
- This issue is not reproducible when trying to log in from the “More” menu when clicking the “Sign In” button.
- Attached is a screen recording of the issue:
![YouAreNowSignedIn](https://user-images.githubusercontent.com/51400741/135587286-226002e1-9620-474a-8c87-fac64a61e0fd.gif)

Issue #4367, opened 2021-06-24
The Sign in prompt can't be dismissed after it's triggered a second time from a different location


Softvision-RemusDranca:
**[Affected Versions]:**
- Firefox Release 89.0.1
- Chrome 91.0.4472.106
- Safari 14

**[Affected Platforms]:**
- Windows 10x64
- IphoneSE

**[Prerequisites]:**
- Hubs Cloud services are set up.
- Have a room opened.
- You are not logged into Hubs.

**[Steps to reproduce]:**
1. Click the "Favorites Rooms" button from the More menu.
2. Dismiss the Sign In prompt.
3. Click the "Sign In" button from the "More" menu.
4. Close the Sign In prompt.

**[Expected results]:**
- The prompt is closed.

**[Actual results]:**
- The prompt can't be closed.

**[Notes]:**
- The issue persist even if the login process is completed, the prompts associated can't be dismissed.
- The issue is also reproducible on "https://hubs.mozilla.com/" website.
- Attached is a screen recording of the issue:
![SignInPromptDismiss](https://user-images.githubusercontent.com/40061643/123274999-a638db80-d50c-11eb-9723-d84659705685.gif)

Issue #3886, opened 2021-02-08
Sign in dialogue sometimes doesn't close on Chrome


emclaren:
Sign in dialogue sometimes doesn't close on Chrome. 
![EssbDvaUwAAFynH](https://user-images.githubusercontent.com/6835324/107279567-0d496e00-6a0c-11eb-8b16-538c361cb486.jpg)

Reported by community member here https://twitter.com/amoration/status/1354205993928822785 

I could not reproduce on my Mac & I followed up to get more info from the community member about the type of device they were using

emclaren:
Follow up - this was on a PC

PR #5581, opened 2022-07-12
Signing modal issues

keianhzo:
Fixes https://github.com/mozilla/hubs/issues/4810

We set the modals `onClose` callback to either `onContinueAfterSignIn` or `closeDialog` depending if the dialog needs to execute an action after signing or not but we never reset `onContinueAfterSignIn`. This means that if after showing a dialog that requires an action we show one that doesn't, the second one won't close as it will keep on calling the previous `onContinueAfterSignIn`callback.

This PR reverts a previous patch, resets the `onContinueAfterSignIn`callback after it's been called and fixes some component property types that are throwing warnings.

@SV-AndreiC I think it's worth revisiting these already closed issues. I've tested them but just in case: 
https://github.com/mozilla/hubs/issues/4695
https://github.com/mozilla/hubs/issues/4367
https://github.com/mozilla/hubs/issues/3886
https://github.com/mozilla/hubs/issues/4810

### DIFF
--- a/src/hub.js
+++ b/src/hub.js
@@ -777,7 +777,7 @@ document.addEventListener("DOMContentLoaded", async () => {
       showSignInDialog: true,
       signInMessage,
       onContinueAfterSignIn: async () => {
-        remountUI({ showSignInDialog: false });
+        remountUI({ showSignInDialog: false, onContinueAfterSignIn: null });
         let actionError = null;
         if (predicate()) {
           try {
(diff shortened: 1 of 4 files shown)


brianpeiris (approved):
LGTM. Nice catch.

SV-AndreiC:
@keianhzo I have tested this on hubs.mozilla.com and was not able to reproduce any of the related issues.
I will update once this is pushed and tested on Hubs-Cloud.
Thank you for the fix.

SV-AndreiC:
I can no longer reproduce this issue on the latest Hubs Cloud build.

PR #5585, opened 2022-07-14
Force AppLogo refresh when system theme changes

keianhzo:
The `AppLogo` component is not re-rendered when a system theme update happens. It only re-rendered when closing the preferences in the hub as that conditionally renders UI. This means that in any situations where the logo is visible and the system theme is updated we don't update the logo. As far as I know this only happens in the home screen.

This PR forces a `AppLogo` re-render when the system theme is updated by using the already existing `useDarkMode` hook. It also removes the `get-app-logo` util as it's only been used for the logo update but we get the same information from the hook.


keianhzo:
Adding you @brianpeiris as you were the last one updating that code.

src/react-components/misc/AppLogo.js
```diff
@@ -2,15 +2,22 @@ import React from "react";
...
+  const darkMode = useDarkMode();
   if (isHmc() && !forceConfigurableLogo) {
     return <HmcLogo className="hmc-logo" />;
   } else {
-    return <img className={className} alt={configs.translation("app-name")} src={getAppLogo()} />;
+    return (
+      <img
+        className={className}
+        alt={configs.translation("app-name")}
+        src={(darkMode && configs.image("logo_dark")) || configs.image("logo")}
```
brianpeiris, line 18: This isn't fully equivalent to the logic that was in `getAppLogo`. In particular, if your system theme is set to "light", but your Hubs theme is set to "Hubs Dark Mode" in our preferences panel, this will still use the "light" logo according to the system theme.
I think we still care about overriding the system theme with our theme preference, so we should account for that scenario.

src/react-components/styles/theme.js
```diff
@@ -39,14 +46,18 @@ function useDarkMode() {
...
     }
+
+    return () => {
+      if (darkmodeQuery.removeEventListener) {
+        darkmodeQuery.removeEventListener("change", changeListener);
+      } else {
+        darkmodeQuery.removeListener(changeListener);
+      }
+    };
   }, []);
```
brianpeiris, line 61: Technically I think you need to add `changeListener` as an effect dependency now (though in practice I suspect the `setDarkMode` reference never actually changes.)
```suggestion
  }, [changeListener]);
```

brianpeiris (changes requested):

src/react-components/misc/AppLogo.js
keianhzo: Thanks for catching that. I've updated.

src/react-components/styles/theme.js
keianhzo: Added

PR #5590, opened 2022-07-15
Set a default theme to fix the rest button

keianhzo:
I've seen that this is broken while working on another PR.

We set the default theme to undefined but that breaks the reset button.  This PR sets the theme default to the default theme or the first one in case none has been tagged as default.


brianpeiris (approved):
LGTM

Issue #5599, opened 2022-07-20
video-texture-target is darkened since THREE 141 update


netpro2k:
**Description**
Similar to #5571 video-texture-targets are also darkened.

**Screenshots**
![image](https://user-images.githubusercontent.com/130735/180087250-71259914-2925-4fe0-bcb9-001d2b104339.png)

**Additional context**
The fix we applied for camera tool isn't quite as easy to apply here since `onBeforeRender` has to be applied to each object that might have the `video-texture-target` material on it. Conceptually `video-texture-target` is a "material component" so it should apply to all objects using that material. In practice we can't actually add aframe components to materials, only objects, so it might be ok to just do the same trick and set `onBeforeRender` up on the mesh we grab the material from in `video-texture-target`.

takahirox:
Three.js has a `material.onBeforeRender` hook.

PR #5604, opened 2022-07-21
Fix darkened video-texture-target

takahirox:
Fixes: #5599 

This PR fixes darkened video-texture-target problem caused when upgrading our Three.js to r141.

Refer to "Fix darkened camera view" PR #5577 for the details because the root issue is same.

### DIFF
--- a/src/components/video-texture-target.js
+++ b/src/components/video-texture-target.js
@@ -153,9 +153,8 @@ AFRAME.registerComponent("video-texture-target", {
         const texture = videoTextureSource.renderTarget.texture;
         this.applyTexture(texture);
 
-        // Bit of a hack here to only update the renderTarget when the screens are in view
-        material.map.isVideoTexture = true;
-        material.map.update = () => {
+        // Only update the renderTarget when the screens are in view
+        material.onBeforeRender = () => {
           videoTextureSource.textureNeedsUpdate = true;
         };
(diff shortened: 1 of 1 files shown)


src/components/video-texture-target.js
```diff
@@ -153,9 +153,8 @@ AFRAME.registerComponent("video-texture-target", {
         const texture = videoTextureSource.renderTarget.texture;
         this.applyTexture(texture);
 
-        // Bit of a hack here to only update the renderTarget when the screens are in view
-        material.map.isVideoTexture = true;
-        material.map.update = () => {
+        // Only update the renderTarget when the screens are in view
+        material.onBeforeRender = () => {
```
takahirox, line 157: Note to reviews: I assume `material.onBeforeRender` isn't set anywhere else.

netpro2k: Oh wow I didn't realize this existed for materials. Well that makes things simple. We should probably use that for camera tool too then.

netpro2k (approved):
Did not realize materials had `onBeforeRender` as well, that makes this way simpler than I was thinking!

src/components/video-texture-target.js
takahirox: Sounds good to me. Opened #5607

PR #5605, opened 2022-07-22
Fix issue switching between scenes with many entities

netpro2k:
When switching between scenes with > 100 entities (which is not even all that many) BitECS's entity id recycling causes some strange behavior in our object3d cleanup code. 

image

This PR (along with a [patch](https://github.com/NateTheGreatt/bitECS/pull/93) to BitECS) allows entity id recycling to be configured, and configures it to 20% of our 10,000 entity max (2000). This points at our own fork of BitECS for now until the change lands upstream. This is still only a partial solution, as the issue would still happen if entities are added on a frame where > 2000 entities were also removed, but this is very unlikely to happen. 

There was some discussion of further solutions on [github](https://github.com/NateTheGreatt/bitECS/issues/68) and [discord](https://discord.com/channels/654487419407826945/817507396317741106/999788262900117714)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19022,9 +19022,8 @@
       "dev": true
     },
     "bitecs": {
-      "version": "0.3.38",
-      "resolved": "https://registry.npmjs.org/bitecs/-/bitecs-0.3.38.tgz",
-      "integrity": "sha512-vfz6wjPElg/0O7c06Ttb/BeWMInBspSuekRvNoLqPyEZV87tLWipxqpjtGPKZBhin8jv4OuQkrkZZfZpaqOndQ=="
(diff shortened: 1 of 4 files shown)


takahirox:
Curious to know, any downside of choosing 20% that is much greater than the default 1%?  The recycling process time can be longer?

netpro2k:
There is no cost to the "recycling" itself, its just a difference in `addEntity` between an id coming from the removed ids array or the next incremental entity id (https://github.com/NateTheGreatt/bitECS/pull/93/files#diff-4833659b31bb551bd2b6b1f136dee5686fcbcc1c29fb9de498950066d591eb62R87). The "cost" of not recycling is sooner is that you get a bit more memory fragmentation in your component data arrays.

takahirox:
Thanks, probably better memory fragmentation may not be our first priority then 20% is acceptable for us.

PR #5607, opened 2022-07-22
Clean up camera-tool

takahirox:
From: #5604

Use `material.onBeforeRender` instead of `object.onBeforeRender`.


brianpeiris (approved):
LGTM

PR #5615, opened 2022-07-26
Fix darkened video-texture-target

takahirox:
Fixes: #5599

This PR fixes darkened video-texture-target problem caused when upgrading our Three.js to r141.

Refer to "Fix darkened camera view" PR #5577 for the details because the root issue is same.

The previous same purpose PR #5604 was reverted because it didn't work correctly because of losing `material.onBeforeRender` when cloning materials. This PR resolves it by copying the function.


brianpeiris (approved):
LGTM

PR #5619, opened 2022-07-27
Upgrade to Node LTS (16) and Webpack 5

netpro2k:
This PR builds on work done in #5485 and #5240 to make long overdue updates to our build tooling. This should help alleviate all the build and install issues people have been having attempting to use Hubs with non ancient versions of node, reduce the number of build-server-only surprises, and reduce build time and bundle size. This also cleans up some strange bundling hacks we had to use and sets the stage nicely to add support for using typescript, and eventually breaking up hubs UI and "engine" code. Lastly it should allow us to finally close a number of the depndbot PRs, since most of them are for libs used at build time, not runtime.

This updates almost all of our dev dependencies so it definitely needs some testing, but no runtime dependencies should have changed (at least not in a major way, I did have to wipe out the package-lock file so we may have lost some exact pinnings)

Note this does not yet remove/update @babel/polyfill which is deprecated and has had a rather scary warning about core-js 

> Because of the V8 engine whims, feature detection in old core-js versions could cause a slowdown up to 100x even if nothing is polyfilled.

That we have been ignoring for quite some time... We should do that in it's own PR as it feels a bit more dangerous and hard to detect issues with (since they will likely only surface on random older browsers). Alongside that we might want to just hand-craft support.js to be usable in old engines (like IE11) without polyfills, so that we can ditch the confusing multi-target babel setup.

Also, I have updated eslint and prettier but several new rules mean some work needs to be done to make them happy. The work should be trivial but will touch a lot of files, so I also want to leave that for another PR. I have made the new eslint rules warnings and temporarily disabled the prettier eslint rule.


webpack.config.js
```diff
@@ -385,59 +416,24 @@ module.exports = async (env, argv) => {
...
+        // Some JS assets are loaded at runtime and should be copied unmodified and loaded using file-loader
         {
-          test: [
-            path.resolve(__dirname, "src", "utils", "configs.js"),
-            path.resolve(__dirname, "src", "utils", "i18n.js"),
-            path.resolve(__dirname, "src", "support.js")
-          ],
-          loader: "babel-loader",
-          options: legacyBabelConfig
-        },
```
brianpeiris, line 409: I think it's reasonable to re-evaluate whether we need this legacy support. Ideally we should only have to care about evergreen browsers at this point. As a lower priority, we'll probably still want to do spot checks on how things behave in old-but-still-popular browsers. Mainly we want to avoid the situation where nothing is rendered at all and you just see a blank page in unsupported browsers. 
Lastly, in addition to old browsers, there are still edge cases like in-app-browsers that are used in mobile apps like Twitter and Facebook, where we'll want to check that things don't break entirely.

brianpeiris (commented):
Just a few things I've spotted while skimming. Nothing major stands out though.

webpack.config.js
```diff
@@ -9,10 +9,10 @@ const MiniCssExtractPlugin = require("mini-css-extract-plugin");
 const CopyWebpackPlugin = require("copy-webpack-plugin");
 const BundleAnalyzerPlugin = require("webpack-bundle-analyzer").BundleAnalyzerPlugin;
 const TOML = require("@iarna/toml");
-const fetch = require("node-fetch");
+const fetch = import("node-fetch");
```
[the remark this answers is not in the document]
netpro2k: yes good call, I need to test custom client stuff still

webpack.config.js
netpro2k: From what I could tell, the way wee were polyfilling stuff we were only ever loading polyfills in hub.js anyway. Though I guess this could emit code with features that might not be supported. I think if the support page at least runs in IE 11 we can keep this out... How to test that is the question...

brianpeiris: You can use browserstack to do a quick check, but it would be easiest if it were deployed to a public server like dev.ret or smoke. Browserstack has the ability to test local servers, but hubs might be too complicated for that to work.

webpack.config.js
```diff
@@ -207,6 +209,12 @@ function htmlPagePlugin({ filename, extraChunks = [], chunksSortMode, inject })
   return new HTMLWebpackPlugin(options);
 }
 
+const threeExampelesDir = path.resolve(__dirname, "node_modules", "three", "examples");
```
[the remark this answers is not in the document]
netpro2k: Usually I am just bad at typing but currently my e key randomly multi-fires so I am going to blame this one on that. New kickstarter keyboard is supposed to be shipping soon so I have been resisting the urge to break out the soldering iron to replace the faulty switch.

src/react-components/styles/theme.js
```diff
@@ -37,90 +35,84 @@ function useDarkMode() {
...
+    if (!theme && darkMode) {
+      theme = themes.find(t => t.darkModeDefault);
+    }
 
-      if (!theme) {
-        theme = themes.find(t => t.default);
-      }
+    if (!theme) {
+      theme = themes.find(t => t.default);
+    }
```
nickgrato, line 85: could add a return in this (!then) and remove the next one.

keianhzo:
I'm getting some errors in the console when running against dev:
Screen Shot 2022-07-29 at 16 54 03
Screen Shot 2022-07-29 at 16 54 52

webpack.config.js
netpro2k: Tested on IE11 and page failed to load on a syntax error for `=>`.. I think I misunderstood the intent of this block before. Having a restricted legacy babel config makes sense, we had just started using it for other things. I have restored this but inlined it, overriding only the target and keeping all plugins (so we don't accidentally make some code indectly linked to the support bundle that breaks it. 

In the future I think it would be best to just hand craft support.js to be functional without polyfilling. We don't need anything fancy in there, not even React.

src/react-components/styles/theme.js
netpro2k: This suggestion makes sense, but this code only happened to change because of code formatting. The actual change in this file was making the `APP` lookup optional (to fix storybook). Gunna keep this as is in this PR.

netpro2k:
Re the ammo errors @keianhzo was seeing. Those were both actually harmless but incidentally have gone away re-introducing worker-loader. If we remove that again we will likely encounter them again, so noting for future:
- The "world not initialized" error is the ammo worker failing to correctly handle postmessages it is not expecting. In this case, messages from webpack de server.
- The build error about ammoWorker. three-ammo doesn't really correctly export its worker, it expects the thing using it to bundle it, but then exports it. Webpack doesn't know how to handle this. We circumvented it for our code, but the internal include (which nobody uses) still gets confused.

PR #5621, opened 2022-07-28
fix m1 npm ci fail by replacing lib-hubs with three-particle-emitter

kenykhung:
Related to discussion #5603 
Replace lib-hubs with three-particle-emitter to bypass legacy electron dependency and pass `npm ci` for mac m1

Tested environment:
Mac M1 Pro 12.1
NPM 8.1.2
Node v16.15.0


netpro2k:
Hmm. This unfortunately ends up using an outdated version of three-particle-emitter (since the changes we made in lib-hubs haven't been published). We can probably publish a new version but I would like to try and actually get to the bottom of what is importing electron and just fix that instead...

camelgod:
I just copy-pasted the lib-hubs .js files into hubs repo and reference it locally for now. Works fine! Will do that until Mozilla can clean up lib-hubs dependency issues.

knev:
@camelgod I don't suppose you could elaborate on that workaround. I found lib-hubs. Where do i need to make the local changes exactly?

markusTraber:
As a workaround, I am using rosetta.
Here are some instructions for it: https://github.com/nvm-sh/nvm#macos-troubleshooting

You might need to manually delete previously installed node versions for arm64.

keianhzo:
This other PR should fix this: https://github.com/MozillaReality/lib-hubs/pull/46

Closing for now.

PR #5623, opened 2022-07-28
Fix embed codes for anonymous users or non-owners

### LINKED ISSUE #3570: Embed Code field should be hidden when user does not have permissions to embed a room
**Description**
Currently the redesign shows an "Embed Code" field in the "invite" panel even if a user doesn't have embed permissions. The `embed_token` in the iframe src is "null".

**To Reproduce**
1. Visit a room that you did not create
2. Click on Invite
3. See that the Embed Code iframe src contains a null embed_token

**Expected behavior**
We should hide the Embed Code field when the user does not have that permission.

**Screenshots**

image

**Hardware** 
 - Device: Desktop
 - OS: Windows
 - Browser: Firefox

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-531)

### LINKED ISSUE #5622: Embedded rooms no longer work
**Description**
'Embed' link no longer works to embed a room inside of an iFrame.

**To Reproduce**
1. Create a room
2. Copy the iFrame code
3. Paste it into a web page

**Expected behavior**
The iFrame loads

**Screenshots**
image

**Hardware** 
 - Device: Desktop
 - OS: Windows 10
 - Browser: Firefox

brianpeiris:
If a user is not signed in and they create a room, we store the embed token for the room in the local store. We had some code that retrieved it from the store, but that code wasn't actually in use. This resulted in the embed token being null in the embed code.

Additionally, if a user is not an owner of a room, they will not have an embed token for the room, by design. In those cases, we should hide the embed code UI altogether.

Fixes #5622 Fixes #3570

### DIFF
--- a/src/hub.js
+++ b/src/hub.js
@@ -537,15 +537,6 @@ function handleHubChannelJoined(entryManager, hubChannel, messageDispatch, data,
   }
 
   const hub = data.hubs[0];
-  let embedToken = hub.embed_token;
-
-  if (!embedToken) {
-    const embedTokenEntry = store.state.embedTokens && store.state.embedTokens.find(t => t.hubId === hub.hub_id);
-
-    if (embedTokenEntry) {
-      embedToken = embedTokenEntry.embedToken;
-    }
-  }
 
   console.log(`Dialog host: ${hub.host}:${hub.port}`);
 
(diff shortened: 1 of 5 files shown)


src/hub.js
```diff
@@ -556,8 +547,7 @@ function handleHubChannelJoined(entryManager, hubChannel, messageDispatch, data,
     onMediaSearchResultEntrySelected: (entry, selectAction) =>
       scene.emit("action_selected_media_result_entry", { entry, selectAction }),
     onMediaSearchCancelled: entry => scene.emit("action_media_search_cancelled", entry),
-    onAvatarSaved: entry => scene.emit("action_avatar_saved", entry),
-    embedToken: embedToken
```
brianpeiris, line 560: The `embedToken` prop was a remnant from older an UI design. It is no longer used in `UIRoot`.

bryanenders (approved):

Issue #5576, opened 2022-07-08
Recorded camera videos don't play in "Open Link"


takahirox:
**Description**

Recorded camera videos don't play in "Open Link"

Update: The same problem should happen with uploaded videos

**To Reproduce**
1. Go to https://hubs.mozilla.com
2. Create and enter a room
3. Place a camera and record video
4. Clink "Open Link" on the object menu for recorded video
5. Confirm a new tab opens but the video doesn't play

**Expected behavior**

The video plays on the new tab

**Hardware** 
 - Device: Desktop
 - OS: Windows 10
 - Browser: Chrome

**Additional context**

Console error in the new tab

![image](https://user-images.githubusercontent.com/7637832/178045679-146494dd-6052-4110-95f5-c22d8133c46f.png)

It seems a Content Security Policy error.

takahirox:
![image](https://user-images.githubusercontent.com/7637832/181128910-49222190-49f7-469f-8395-90df3e1cc6d4.png)

Adding `media-src 'self' https://uploads-prod.reticulum.io;` to content security policy can resolve the problem?

takahirox:
The same problem happens even with `npm run dev`.

1. Run `npm run dev`
2. Go to https://hubs.local:8080/
3. Create and enter a room
4. Place a camera and record video
5. Clink "Open Link" on the object menu for recorded video
6. Confirm a new tab opens but the video doesn't play

Console error

![image](https://user-images.githubusercontent.com/7637832/181162143-cc2baf1b-0579-44b1-8135-d5b979a432d9.png)

takahirox:
Made a PR for reticulum https://github.com/mozilla/reticulum/pull/611 Probably it will fix the problem on `hubs.mozilla.com`.

But it may not resolve the problem with `npm run dev` yet because recorded videos (and uploaded videos) are loaded via the proxy `hubs-proxy.com`. We don't really want to relax the security policy for the proxy for security, `https://hubs-upload-cdn.com/files/*` should have CORS that allows `hubs.local`, and there doesn't seem to be reasons why they should be loaded via the proxy. So perhaps, in addition to https://github.com/mozilla/reticulum/pull/611, directly loading the original url would be a solution.

PR #5624, opened 2022-07-29
Add hubs-upload-cdn.com to NON_CORS_PROXY_DOMAINS in .defaults.env

takahirox:
Fixes: #5576

User upload files in `hubs.local` (`npm run dev`) are uploaded to `hubs-upload-cdn.com` and they are loaded via the proxy (`hubs-proxy.com`).

`hubs-upload-cdn.com` seems to have CORS that allows `hubs.local` then the files don't need to be loaded via the proxy.

To avoid the proxy, add `hubs-upload-cdn.com` to `NON_CORS_PROXY_DOMAINS` in `.defaults.env`.


brianpeiris (approved):
LGTM

PR #5629, opened 2022-08-02
Rewrite idle-detector with the one with the new framework

takahirox:
I rewrote `idle-detector` with the one with the new framework to get used to the new framework and to reduce the dependency with Aframe.

Notes:
* It needs some initializations. I used module locales for them.
* Should the userinputs be represented as entities?

### DIFF
--- a/src/systems/idle-detector.js
+++ b/src/systems/idle-detector.js
@@ -17,48 +17,45 @@ const BASIC_ACTIVITY_PATHS = [
   paths.actions.angularVelocity
 ];
 
-AFRAME.registerSystem("idle-detector", {
-  init() {
-    this.resetTimeout = this.resetTimeout.bind(this);
-    this.idleTimeout = null;
-    this.lastInputCheck = 0;
+let idleTimeout = null;
+const lastInputCheck = 0;
 
-    const events = ["click", "pointerdown", "touchstart", "keyup"];
+const onIdleTimeout = () => {
+  window.dispatchEvent(new CustomEvent("idle_detected"));
+};
 
-    for (const event of events) {
-      window.addEventListener(event, this.resetTimeout);
-    }
+const resetTimeout = () => {
+  if (idleTimeout) clearTimeout(idleTimeout);
+  idleTimeout = setTimeout(onIdleTimeout, IDLE_TIMEOUT_MS);
+  window.dispatchEvent(new CustomEvent("activity_detected"));
+};
 
-    this.resetTimeout();
-  },
-  resetTimeout() {
-    if (this.idleTimeout) clearTimeout(this.idleTimeout);
-    this.idleTimeout = setTimeout(this.onIdleTimeout, IDLE_TIMEOUT_MS);
-    window.dispatchEvent(new CustomEvent("activity_detected"));
-  },
-  onIdleTimeout() {
-    window.dispatchEvent(new CustomEvent("idle_detected"));
-  },
-  tick(time) {
-    if (time - this.lastInputCheck < INPUT_CHECK_INTERVAL_MS) return;
+for (const event of ["click", "pointerdown", "touchstart", "keyup"]) {
+  window.addEventListener(event, resetTimeout);
+}
 
-    const userinput = this.el.systems.userinput;
+resetTimeout();
 
-    let basicActivity = false;
-    for (const activityPath of BASIC_ACTIVITY_PATHS) {
-      basicActivity = basicActivity || !!userinput.get(activityPath);
-    }
+export const idleDetectSystem = time => {
+  // TODO: When should lastInputCheck be updated?
+  if (time - lastInputCheck < INPUT_CHECK_INTERVAL_MS) return;
 
-    const characterAcceleration = userinput.get(CHARACTER_ACCELERATION_PATH);
+  // TODO: Remove the dependency with AFRAME
+  const userinput = AFRAME.scenes[0].systems.userinput;
 
-    const active =
-      basicActivity ||
-      !!(characterAcceleration && characterAcceleration[0]) ||
-      !!(characterAcceleration && characterAcceleration[1]);
+  let basicActivity = false;
+  for (const activityPath of BASIC_ACTIVITY_PATHS) {
+    basicActivity = basicActivity || !!userinput.get(activityPath);
+  }
 
-    if (active) {
-      this.resetTimeout();
-    }
-  },
-  remove() {}
-});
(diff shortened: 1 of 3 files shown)


src/systems/idle-detector.js
```diff
@@ -17,48 +17,45 @@ const BASIC_ACTIVITY_PATHS = [
...
-    for (const activityPath of BASIC_ACTIVITY_PATHS) {
-      basicActivity = basicActivity || !!userinput.get(activityPath);
-    }
+export const idleDetectSystem = time => {
+  // TODO: When should lastInputCheck be updated?
+  if (time - lastInputCheck < INPUT_CHECK_INTERVAL_MS) return;
 
-    const characterAcceleration = userinput.get(CHARACTER_ACCELERATION_PATH);
+  // TODO: Remove the dependency with AFRAME
+  const userinput = AFRAME.scenes[0].systems.userinput;
```
takahirox, line 44: Any better way to get `systems.userinput` especially without the AFRAME dependency?

src/systems/idle-detector.js
```diff
@@ -17,48 +17,45 @@ const BASIC_ACTIVITY_PATHS = [
...
-    const userinput = this.el.systems.userinput;
+resetTimeout();
 
-    let basicActivity = false;
-    for (const activityPath of BASIC_ACTIVITY_PATHS) {
-      basicActivity = basicActivity || !!userinput.get(activityPath);
-    }
+export const idleDetectSystem = time => {
+  // TODO: When should lastInputCheck be updated?
+  if (time - lastInputCheck < INPUT_CHECK_INTERVAL_MS) return;
```
takahirox, line 41: `lastInputCheck` doesn't seem to be updated anywhere. It seems a bug but I kept it as is so far because I'm not sure when it should be updated and what this check is really for.

src/systems/idle-detector.js
```diff
@@ -17,48 +17,45 @@ const BASIC_ACTIVITY_PATHS = [
...
+  const userinput = AFRAME.scenes[0].systems.userinput;
 
-    const active =
-      basicActivity ||
-      !!(characterAcceleration && characterAcceleration[0]) ||
-      !!(characterAcceleration && characterAcceleration[1]);
+  let basicActivity = false;
+  for (const activityPath of BASIC_ACTIVITY_PATHS) {
+    basicActivity = basicActivity || !!userinput.get(activityPath);
+  }
```
takahirox, line 49: Should the userinputs be represented as entities like the following?

```
let basicActivity = false;
userinputQuery(world).forEach(foo => {
  basicActivity = basicActivity || ... ;
});
```

src/systems/idle-detector.js
netpro2k: We are unlikely to replace the user input system any time soon, and if we do we likely would keep a similar interface, just migrate it to be a non-aframe system. A slightly nicer way to access it would be to pass it into `idleDetectSystem`.

src/systems/idle-detector.js
netpro2k: I don't see `lastInputCheck` being set anywhere in the current code, so it does seem like a bug. We should use porting systems as an opportunity to rethink the problem from the beginning. I would try to avoid the temptation of just blindly porting things forward. We should be aggressive about cutting things that don't make sense, or even cutting things that might be features but add way more complexity than value.

src/systems/idle-detector.js
netpro2k: I think accessing input will be done through the userinput system for the foreseeable future. If we did want to make it more "ECS" I think it would just mean moving the input state into a component and then having a single entity with that component on it to grab the state off of. 

It would probably have something like:

 ```js
const eid = anyEntityWith(world, InputFrame);
if(InputFrame.actionTurnLeft[eid]) { ... }
```
I don't think we would really gain much from doing that. We should just see ECS as one of the tools we have, we don't need to try and fit everything into that box.

File: src/systems/idle-detector.js [hunk 4]
```diff
@@ -17,48 +17,45 @@ const BASIC_ACTIVITY_PATHS = [
...
-  init() {
-    this.resetTimeout = this.resetTimeout.bind(this);
-    this.idleTimeout = null;
-    this.lastInputCheck = 0;
+let idleTimeout = null;
+const lastInputCheck = 0;
 
-    const events = ["click", "pointerdown", "touchstart", "keyup"];
+const onIdleTimeout = () => {
+  window.dispatchEvent(new CustomEvent("idle_detected"));
```
netpro2k, line 24: One of the things we are trying to get out of our code is event dispatching. It makes the code flow harder to follow. 

I am not super familiar with this code but tracing this even it looks like the only thing using this (and the other event we are firing below) is the code to show the timeout UI. 

Essentially this code https://github.com/mozilla/hubs/blob/cd8e2a12d50126148f94d14d790dcb3db23eaad9/src/react-components/ui-root.js#L510-L536 is really part of what "idle-detector" is doing at its core and it would be nice if that was more obvious... Unfortunately we don't really have a nice clean way for communicating between our game code and react UI code, so I am not sure how I would go about untangling it. The countdown and exiting logic could be moved but not sure how I would go about showing/updating the UI or capturing clicking the cancel button...

I don't think this is particularly important, the code works fine and its not integral to main gameplay, but again I think porting systems gives us the opportunity to rethink problems and clean up our tech debt.

netpro2k (commented):
Left some comments. I am intentionally being over specific since I know the goal of this PR is just to get a taste for writing stuff in the "new style". The way I like to tackle it is to look through all the current code to understand the problem the existing thing is solving, but then completely throw out the current implementation in my mind, and rethink it from the start.

For idle detector that's "We want to exit the scene if a user hasn't done anything for a certain amount of time. Before exiting we want to show the user a message that will count down giving them an opportunity to cancel". 

Thinking through this problem quickly in my mind a system would look something like:

- There are 3 states I can be in. CHECKING, COUNTDOWN, and EXITED
- When I am in the CHECKING state I need to keep track of how long it has been since the last user input. If it has been long enough I go into the COUNTDOWN state
- When in the COUNTDOWN state I show a message to the user and start tracking a countdown time
- If the user doesn't click cancel before the countdown triggers I enter the EXITED state, and exit the scene. This system then does nothing anymore.

As I mentioned in the comments, I think this "ideal" system is complicated by the fact that we don't run our raf loop in the background (so our system doesn't run) and we don't have a clean path to interacting with the UI (so showing the countdown and capturing cancel is not straightforward). But it's useful to start from this ideal instead of the code we had before. Then we can try to solve the problems that make that ideal hard to implement (or just decide they are not worth it right now, but that should be a conscious choice).

PR #5647, opened 2022-08-08
Bitecs media-pdf migration

keianhzo:
BitECS version of the `media-pdf` component.


brianpeiris:
Would be good to revisit #2573 after this is merged.

netpro2k:
This was implemented already in another set of PRs

PR #5649, opened 2022-08-09
Replace CircleCI configuration with GitHub workflow

brianpeiris:
- Adds a test-and-deploy-storybook workflow that runs `npm test` (which mostly does linting) on the client and the admin panel.
- The workflow also builds and deploys the component storybook to gh-pages, but only for pushes to the master branch (https://mozilla.github.io/hubs)
- Changed some eslint rules to warnings in the admin panel config, similar to the changes we made in #5619 
- Removed the `npm run build` step from `npm test`, since that is already covered by the existing build workflow
- Fixed the `ava` configuration to conform to the updated version we now use.


.github/workflows/test-and-deploy-storybook.yml
```diff
@@ -0,0 +1,28 @@
...
+
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 16
+
+      - run: npm ci
+      - run: npm test
+
+      - run: cd admin && npm ci
+      - run: cd admin && npm test
```
bryanenders, line 23: >Removed the npm run build step from npm test, since that is already covered by the existing build workflow

Good catch! 👏

bryanenders (approved):
Fancy

brianpeiris:
Had to add one extra commit to this, since the changes in #5651 broke Storybook. https://github.com/mozilla/hubs/commit/7955a444a158b1de0eb102c8f813fc5595ecd169

PR #4332, opened 2021-06-15
fix button color on default theme settings

yakyouk:
- fix action button color: when user preference theme is 'default', the action button and action button highlight theme colors are not correctly applied on action buttons. Note that theme logic is duplicated in `utils/themes.json` and `react-components/styles/themes.js`. I did not find a good way to prevent this but I added a mention of this in both files.
- fixes for `webpack.config.json`:
  - fix localDev themes: themes JSON was not parsed
  - ~~fix remoteDev themes: moved theme from `appConfig.themes` to `appConfig.theme.themes`~~ (already fixed in master)
  - add `/link`, `/scenes` and `/avatars` paths for development
  - add `basis_transcoder.worker.js` to `.prettierignore` to prevent prettifying with `npm run prettier`

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-792)

PR #5651, opened 2022-08-10
Fix and unify theme retrieval

brianpeiris:
This builds on top of #4332 and unifies theme retrieval and theme switching into helpers that live in `src/utils/theme.js`. Tested this in local dev and against a remote hubs cloud instance.

- Adds an `onThemeChanged` helper that listens to both hubs preference changes and browser/system dark mode changes
  - This helper returns a cleanup function that can be called to remove the event listener
- Adds a `getThemeColor` helper which can retrieve a single color.
- Adds a `registerDarkModeQuery` helper which takes a listener, and returns the dark mode query, and a function to remove the listener.
- Moves a bunch of config and theme related logic from `src/react-components/styles/theme.js` to `src/utils/theme.js`, for reuse
- Moves the `themes` array export from `src/react-components/styles/theme.js` to `src/utils/theme.js`, so that it can be used more broadly
- Updates `sprites.js`, `avatar-preview.js`, and `pin-networked-object-button.js` to use the new `onThemeChanged` helper, so that they update immediately when dark mode is changed at the browser or system level, in addition to theme preference changes.
- Add an eslint exception to ignore destructured variables that begin with "_"


src/utils/theme.js
```diff
@@ -72,29 +144,68 @@ function activateTheme() {
...
+function onThemeChanged(listener) {
+  const storeIsAvailable = !!(window.APP?.store);
+  if (storeIsAvailable) {
+    registerThemeChangedListener(listener);
+  } else {
+    stashedThemeChangedListeners.add(listener);
+  }
+
+  return () => {
+    if (storeIsAvailable) {
```
johnshaughnessy, line 179: I don't think `storeIsAvailable` will be re-computed here, so a listener that started in the stash and then registered via `registerStashedThemeChangedListeners` may not be cleaned up.

We should either recompute `!!(window.APP?.store)` or handle both possible states:
```js
stashedThemeChangedListeners.delete(listener);
if (themeChangedListeners.has(listener)) {
  themeChangedListeners.get(listener).removeListener();
  themeChangedListeners.delete(listener);
}
```

johnshaughnessy (approved):

src/utils/theme.js
brianpeiris: Ah, yeah, wasn't sure what to do here. Handling both states is a good idea. Thanks.

PR #5660, opened 2022-08-18
Add third person camera view

jeus0630:
image
스크린샷 2022-08-18 오후 3 34 21

## DOING
- Added new feature of third-person-camera-view
- Added toggle option in the preference to change the camera POV

### DIFF
--- a/src/systems/camera-system.js
+++ b/src/systems/camera-system.js
@@ -158,6 +158,7 @@ export const CAMERA_MODE_THIRD_PERSON_NEAR = 1;
 export const CAMERA_MODE_THIRD_PERSON_FAR = 2;
 export const CAMERA_MODE_INSPECT = 3;
 export const CAMERA_MODE_SCENE_PREVIEW = 4;
+export const CAMERA_MODE_THIRD_PERSON_VIEW = 5;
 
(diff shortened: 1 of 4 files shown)


jeus0630:
New PR with updated code based on @markusTraber's comment.

src/systems/camera-system.js
```diff
@@ -158,6 +158,7 @@ export const CAMERA_MODE_THIRD_PERSON_NEAR = 1;
 export const CAMERA_MODE_THIRD_PERSON_FAR = 2;
```
netpro2k, line 158: Thanks for making this PR! I don't think this can be merged as is, since it conflicts with existing code for third person view.

As you can see here there is already code in the client to handle a third person camera view (with "near' and "far" toggles). I am pretty sure it is not working quite right (it was mostly just an experiment so has not been maintained), but it should serve as a good foundation for integrating this. A preference screen option for this instead of querystring is good but I suspect users will also want to be able to toggle in and out of it at runtime with a hotkey as well. We definitely should not have both codepaths continue to exist though.

piyush2greedy:
@jeus0630 Is it working fine

PR #5689, opened 2022-08-30
Fixes locale string extraction with formatjs

brianpeiris:
Newer versions of `@formatjs/cli` do not export an `extract` function. This changed with our recent node/npm upgrade, which broke the `npm run intl-extract` script that we use for translations.
The function now lives in `@formatjs/cli-lib`, so this PR adds that as a dependency and uses it in `intl-extract.js`


keianhzo (approved):
LGTM

PR #5692, opened 2022-09-01
Add startOffset and timeScale to loop-animation

netpro2k:
Adds startOffset (skip this many seconds on the first loop of the animation) and timeScale (scale the animation playback speed by this factor) to the "loop-animation" component.


keianhzo (approved):
LGTM

PR #5694, opened 2022-09-01
Refresh dialog join token when the phoenix channel reconnects

keianhzo:
When the phoenix channel is reconnected the JWT token might have changed but we weren't updating the dialog adapter token. This can cause issues if we try to reconnect to dialog after the token has changed.


brianpeiris (approved):
LGTM. Nice catch!

PR #5696, opened 2022-09-02
Add required reference to NAF in scene.js

keianhzo:


brianpeiris (approved):
Ideally we would work towards removing reliance on these globals, and import them explicitly instead, but this looks good as an immediate fix.

Issue #5323, opened 2022-04-01
Run tick() after the entire scene graph matrices update if no side effect to matrices


takahirox:
Related: #5294

**Is your feature request related to a problem? Please describe.**

Some systems need up-to-date world matrices but don't have side effect to matrices. 

They ensure world matrices to be up to date by directly or indirectly calling `.updateMatrices()` which traverses up to the root node and update their world matrices.

But traversing up to the root node, even without matrices update, seems costly especially on low-end devices.

**Describe the solution you'd like**

Run tick() after the entire scene graph matrices update if no side effect to matrices, like

```
// systems which have side effect to matrices
for (const system of systems1) {
  system.tick();
}
scene.updateMatrixWorld();
// systems which don't have side effect to matrices
for (const system of systems2) {
  system.tick();
}
scene.autoUpdate = false;
renderer.render(scene, camera);
```

they no longer need to call `.updateMatrices()` in tick().

**Additional context**

The solution may require changes in A-Frame?

PR #5701, opened 2022-09-08
Cleanup main tick

netpro2k:
This PR attempts to consolidate everything that happens in a frame into 1 spot so its easy to see the order things are executing in. It also cleans up several pending TODOs around it, namely unifying the various timers we were using and getting rid of some more globals being stored on a-scene. This doesn't yet attempt to handle the TODOs in bit-camera-sytem around turning off `scene.autoUpdate` (#5323). That feels like it warrants its own PR just because "matrix update bugs" are sneaky.

This is probably easiest to review commit-by-commit since I tried to break them down into logical chunks and also the intermediate commits had NOTES that might be useful to read.

### DIFF
--- a/src/systems/hubs-systems.ts
+++ b/src/systems/hubs-systems.ts
@@ -51,6 +51,8 @@ import { destroyAtExtremeDistanceSystem } from "./bit-destroy-at-extreme-distanc
 import { videoMenuSystem } from "../bit-systems/video-menu-system";
 import { deleteEntitySystem } from "../bit-systems/delete-entity-system";
 import type { HubsSystems } from "aframe";
+import { Camera, Clock, Scene, WebGLRenderer } from "three";
+import { HubsWorld } from "../app";
 
 declare global {
   interface Window {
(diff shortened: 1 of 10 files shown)


src/systems/hubs-systems.ts
```diff
@@ -92,83 +111,125 @@ AFRAME.registerSystem("hubs-systems", {
...
   }
 });
+
+export function mainTick(xrFrame: XRFrame, renderer: WebGLRenderer, scene: Scene, camera: Camera) {
+  const world = APP.world;
+  const sceneEl = AFRAME.scenes[0];
+  const aframeSystems = sceneEl.systems;
+  const hubsSystems = aframeSystems["hubs-systems"];
+
+  // TODO does anything actually ever pause the scene?
```
johnshaughnessy, line 125: I see that this check existed before, and has just been moved to here, but it seems like it would be a bug for anything to be able to pause the scene. At least -- I'm not aware of anything that would be doing that on purpose.

Along similar lines, perhaps we shouldn't start the `mainTick`until `DOMContentDidLoad` so we can remove the check here.

johnshaughnessy (approved):
lgtm

PR #5706, opened 2022-09-10
Add fonts to pages with the mailing list banner

brianpeiris:
We missed adding the new mailing list fonts to some of the pages with the banner.


brianpeiris:
@nickgrato Just for awareness, I fixed this on Friday. We just missed adding the fonts to the other pages where the banner appeared. See HUBSTRKY-622

PR #5708, opened 2022-09-12
Prepare for local orchestration

bryanenders:
Why
---
An “Invalid Host header” error message appears in the browser when Hubs is orchestrated with Docker Compose

What
----
Allow internal hostname to be supplied by OS var


brianpeiris (approved):
LGTM

PR #5709, opened 2022-09-12
Fix some lint warnings

brianpeiris:
Fix some of the obvious linter warnings that came with the recent node/npm upgrade, to cut down on noise in github PRs. The `useCallback` in banner.js was ineffective, since it didn't list dependencies, so the behavior is unaffected there.


brianpeiris:
Eh, on second thoughts, this is probably not very useful on its own. It might be better to address all the warnings in one go.

PR #5711, opened 2022-09-12
Adding new workflow call to change branch pointer and tag release. 

Pattersonbl2:
Tested workflow and everything works correctly.


tanfarming:
i think the title can be more descriptive
also since squash and merge is disabed on this repo i think we should squash/clean up the commits before PR

PR #5716, opened 2022-09-15
Add Admin Panel UI for account deletion

brianpeiris:
Adds a delete button to the account edit screen in the admin panel. Includes a confirmation dialog and states. Also removes the bulk delete capability from the account list.

Goes with https://github.com/mozilla/reticulum/pull/628

### DIFF
--- a/admin/src/react-components/account-edit-toolbar.js
+++ b/admin/src/react-components/account-edit-toolbar.js
@@ -1,0 +1,107 @@
+import React, { useState } from "react";
+import { Toolbar, SaveButton } from "react-admin";
+import { withStyles } from "@material-ui/core/styles";
+import Button from "@material-ui/core/Button";
+import { Dialog, DialogContent, DialogContentText, DialogActions } from "@material-ui/core";
+
+const accountEditToolbarStyles = {
+  spaceBetween: { justifyContent: "space-between" },
+  dialogActions: { padding: "0 5px" }
+};
+
+const DeleteStates = Object.freeze({
+  Confirming: Symbol("confirming"),
+  Deleting: Symbol("deleting"),
+  Succeeded: Symbol("succeeded"),
+  Failed: Symbol("failed")
+});
+
+export const AccountEditToolbar = withStyles(accountEditToolbarStyles)(props => {
+  const { classes, ...other } = props;
+  const { Confirming, Deleting, Succeeded, Failed } = DeleteStates;
+  const [openConfirmationDialog, setOpenConfirmationDialog] = useState(false);
+  const [deleteState, setDeleteState] = useState(Confirming);
+
+  const onDeleteAccount = async () => {
+    setDeleteState(Deleting);
+
+    try {
+      const resp = await fetch(`/api/v1/accounts/${props.id}`, {
+        method: "DELETE",
+        headers: {
+          "content-type": "application/json",
+          authorization: `bearer ${window.APP.store.state.credentials.token}`
+        }
+      });
+
+      setDeleteState(resp.ok ? Succeeded : Failed);
+    } catch {
+      setDeleteState(Failed);
+    }
+  };
+
+  return (
+    <Toolbar {...other} className={`${classes.spaceBetween}`}>
+      <SaveButton />
+
+      {!props.record.is_admin && (
+        <Button label="Delete" onClick={() => setOpenConfirmationDialog(true)} variant="outlined">
+          Delete
+        </Button>
+      )}
+
+      <Dialog open={openConfirmationDialog}>
+        <DialogContent>
+          <DialogContentText>
+            {(() => { 
+              switch (deleteState) {
+                case Confirming:
+                  return (
+                    <>
+                      Are you sure you want to delete account {props.id}?<br />
+                      <br />
+                      <b>WARNING!</b> This account will be permanently deleted, including all its scenes, assets, avatars,
+                      rooms and files. <b>This cannot be undone.</b>
+                    </>
+                  )
+                case Deleting:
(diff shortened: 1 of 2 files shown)


bryanenders:
Rebased

admin/src/react-components/account-edit-toolbar.js
```diff
@@ -0,0 +1,101 @@
+import React, { useState } from "react";
+import { Toolbar, SaveButton } from "react-admin";
+import { withStyles } from "@material-ui/core/styles";
+import Button from "@material-ui/core/Button";
+import { Dialog, DialogContent, DialogContentText, DialogActions } from "@material-ui/core";
+
+const accountEditToolbarStyles = {
```
nickgrato, line 7: not a fan of this but looks like this is how all the components work. I think this project just need a lot of overall help.

admin/src/react-components/account-edit-toolbar.js
```diff
@@ -0,0 +1,101 @@
...
+  spaceBetween: { justifyContent: "space-between" },
+  dialogActions: { padding: "0 5px" }
+};
+
+const deleteStates = {
```
nickgrato, line 12: when working if consts like this I usually like to use capitals 

const deleteStates = {
  Confirming: "confirming",
  Deleting: "deleting",
  Succeeded: "succeeded",
  Failed: "failed"
};

and then in the component i'd destruct that { Confirming, Deleting, Succeeded, Failed } = deleteStates

I like this because you can see where the consts come from in the top of the file "deleteStates" but then you can just throw around Confirm without dot notation. And then when i see a capital like "Confirming" then I know it's a const and not a var.

admin/src/react-components/account-edit-toolbar.js
```diff
@@ -0,0 +1,101 @@
...
+            {deleteState === deleteStates.confirming && (
+              <>
+                Are you sure you want to delete account {props.id}?<br />
+                <br />
+                <b>WARNING!</b> This account will be permanently deleted, including all its scenes, assets, avatars,
+                rooms and files. <b>This cannot be undone.</b>
+              </>
+            )}
+
+            {deleteState === deleteStates.deleting && <>Deleting account {props.id}...</>}
```
nickgrato, line 64: id consider changing this to something more like this 

```
{() => {
              switch (deleteState) {
                case Deleting:
                  return <>Deleting account {props.id}...</>;
                case Succeeded:
                  return <>Successfully deleted account {props.id}.</>;
                case Failed:
                  return <>Failed to delete account {props.id}.</>;
              }
            }}
```

admin/src/react-components/account-edit-toolbar.js
```diff
@@ -0,0 +1,101 @@
...
+
+            {deleteState === deleteStates.succeeded && <>Successfully deleted account {props.id}.</>}
+
+            {deleteState === deleteStates.failed && <>Failed to delete account {props.id}.</>}
+          </DialogContentText>
+        </DialogContent>
+
+        <DialogActions className={`${classes.dialogActions} ${classes.spaceBetween}`}>
+          {[deleteStates.succeeded, deleteStates.failed].includes(deleteState) && (
```
nickgrato, line 73: clever, I like this.

admin/src/react-components/account-edit-toolbar.js
```diff
@@ -0,0 +1,101 @@
...
+                if (deleteState === deleteStates.succeeded) {
+                  props.history.push("/accounts");
+                }
+              }}
+            >
+              Okay
+            </Button>
+          )}
+
+          {deleteState === deleteStates.confirming && (
```
nickgrato, line 87: might be nice to add a new function like this 

` /**
   * Is state delete state
   * @return boolean
   */
  const isDeleteState = useCallback(
    state => {
      return state === deleteState;
    },
    [deleteState]
  );`

and then refactor ones like this to..

`{isDeleteState(Confirming) && (`

admin/src/react-components/account-edit-toolbar.js
```diff
@@ -0,0 +1,101 @@
...
+
+const deleteStates = {
+  confirming: "confirming",
+  deleting: "deleting",
+  succeeded: "succeeded",
+  failed: "failed"
+};
+
+export const AccountEditToolbar = withStyles(accountEditToolbarStyles)(props => {
+  const { classes, ...other } = props;
```
nickgrato, line 20: i'd through "id" in here as well so you dont have to do prop.id

admin/src/react-components/account-edit-toolbar.js
```diff
@@ -0,0 +1,101 @@
...
+  const [openConfirmationDialog, setOpenConfirmationDialog] = useState(false);
+  const [deleteState, setDeleteState] = useState(deleteStates.confirming);
+
+  const onDeleteAccount = async accountId => {
```
nickgrato, line 24: up to you but you don't really need to pass account id here. Since it's only used once and you have access to prop.id in that function. 

 and do 
 
` const resp = await fetch(`/api/v1/accounts/${props.id}`, {`

nickgrato (approved):
Nothing here is a blocker - feel free to take any of the changes suggested but  I don't see any errors,

admin/src/react-components/account-edit-toolbar.js
bryanenders: Wouldn’t that change the value of `...other`, which is passed to the Toolbar component?

nickgrato: yea let's ignore this for now

bryanenders:
Squashed

PR #5718, opened 2022-09-16
Feat/record mode

djballowe:
This PR looks to solve a pain for folks trying to do screen captures in Hubs. 

Currently, we have a hide ui function but it fails to hide the cursor, debug menus and is not accessible on non-QWRTY keyboards.
Additionally it's annoying to have to close the hide ui tip manually each time you want to record.
This solution addresses all of the previously mentioned problems and provides a smooth transition to a state suitable for recording.

Alternate solutions we considered:

we thought about hiding nametags as well but decided to leave that up to the user.
we considered different inputs on the off chance a user presses the key by mistake. We decided 'R' was too close to movement keys.
we thought about using the built in toast messages but this got tricky since we were also using the built in functions for hide UI 

https://user-images.githubusercontent.com/101905433/190551367-01d6daef-a090-49f5-bb41-1954a1f221c8.mp4


netpro2k (commented):
This LGTM. Thanks for the PR!

The "b" keybinding seems fine for now but we may end up moving it elsewhere at some point.

netpro2k:
Actually it looks like this now has some merge conflicts. I suspect its from the changes made with chat/mic/moderation.

djballowe:
@netpro2k it should be resolved now and ready to merge if everything looks good to you!

matthewbcool:
Tested locally without issues. lgtm

netpro2k (approved):
LGTM

PR #5731, opened 2022-09-26
Mute room feature

keianhzo:
Related https://github.com/mozilla/reticulum/pull/629


src/react-components/room/useMicrophone.js
```diff
@@ -69,8 +71,16 @@ export function useMicrophone(scene) {
     [mediaDevicesManager]
   );
 
+  useEffect(() => {
+    if (canVoiceChat) {
+      mediaDevicesManager.startMicShare({});
```
netpro2k, line 76: Doesn't seem right to tie the actual behavior of mic toggling to a react hook. That would mean if no UI happens to be showing the uses `useMicrophone` this would not be toggled. Should just be handled outside of react land. Probably in MediaDevicesManager in the same spot we are toggling it off on permission changes?

src/react-components/room/RoomSettingsSidebarContainer.js
```diff
@@ -13,6 +15,16 @@ export function RoomSettingsSidebarContainer({ showBackButton, room, hubChannel,
     settings => {
       hubChannel.updateHub(settings);
       onClose();
+
+      NotifiablePermissions.forEach(perm => {
+        if (APP.hub.member_permissions[perm] !== settings.member_permissions[perm]) {
+          hubChannel.sendMessage({
```
netpro2k, line 21: Is there a reason this isn't just part of `hubChannel.updateHub(settings)` like the other settings?

src/components/avatar-audio-source.js
```diff
@@ -121,6 +128,17 @@ AFRAME.registerComponent("avatar-audio-source", {
     };
     APP.store.addEventListener("statechanged", this.onPreferenceChanged);
     this.el.addEventListener("audio_type_changed", this.createAudio);
+    APP.hubChannel.addEventListener("permissions_updated", this.onPermissionsUpdated);
+  },
+
+  onPermissionsUpdated() {
```
netpro2k, line 134: If a user loses owner role they should not be in `moderatorAudioSource` anymore.

src/message-dispatch.js
```diff
@@ -25,6 +25,13 @@ export default class MessageDispatch extends EventTarget {
   addToPresenceLog(entry) {
     entry.key = Date.now().toString();
 
+    const lastEntry = this.presenceLogEntries.length > 0 && this.presenceLogEntries[this.presenceLogEntries.length - 1];
+    if (lastEntry && entry.type === "permission" && lastEntry.type === "permission") {
+      if (lastEntry.body.permission === entry.body.permission && entry.key - lastEntry.key < 10000) {
```
netpro2k, line 30: `key` is technically a string so its slightly weird to be subtracting them, but loljavascript it converts them an ddoes the "right" thing anyway

src/react-components/presence-log.js
```diff
@@ -96,16 +106,27 @@ class PresenceLog extends Component {
...
-    const presenceClasses = {
-      [styles.presenceLog]: true,
-      [styles.presenceLogInRoom]: this.props.inRoom
-    };
+  }, [history, hubId, preset, intl, onViewProfile, presences, include, exclude, isMod]);
 
-    return <div className={classNames(presenceClasses)}>{this.props.entries.map(this.domForEntry)}</div>;
-  }
+  useEffect(() => {
+    setLogEntries(entries.map(domForEntry));
```
netpro2k, line 112: whats the goal with this effect and logEntries state vs just using entries directly?

src/react-components/presence-log.js
```diff
@@ -1,40 +1,40 @@
...
+  "InRoom",
+  "Notifications"
+];
 
-  constructor(props) {
-    super(props);
-  }
+export function PresenceLog({ entries, preset, hubId, history, presences, onViewProfile, include, exclude, ...rest }) {
+  const intl = useIntl();
+  const isMod = useCan("kick_users");
```
netpro2k, line 23: We seem to be mixing moderator and owner in a few places. Here we are checking of `kick_user` permission as "mod" but in other spots related to this we are checking if they have the `owner` role. Is that intended?

src/react-components/room/ChatSidebarContainer.js
```diff
@@ -102,18 +113,23 @@ function updateMessageGroups(messageGroups, newMessage) {
...
   const [messageGroups, setMessageGroups] = useState([]);
   const [unreadMessages, setUnreadMessages] = useState(false);
+  const isMod = useCan("kick_users");
 
   useEffect(
     () => {
       function onReceiveMessage(event) {
         const newMessage = event.detail;
+
+        if (isMod && newMessage.type === "permission") return;
```
netpro2k, line 123: Wouldn't we still want to see this in the log? Especially if a different moderator toggled it.

src/react-components/room/ChatSidebarContainer.js
```diff
@@ -164,12 +180,15 @@ export function ChatSidebarContainer({ scene, canSpawnMessages, presences, occup
...
   const [message, setMessage] = useState("");
+  const { text_chat: canTextChat, kick_users: isMod } = usePermissions();
+  const { text_chat: textChatEnabled } = useRoomPermissions();
   const typingTimeoutRef = useRef();
   const intl = useIntl();
   const inputRef = useRef();
 
   const onKeyDown = useCallback(
     e => {
+      if (!canTextChat) return;
```
netpro2k, line 191: To allow slash commands I think we can basically just check if the value starts with a `/` here.

src/react-components/room/ChatSidebarContainer.js
```diff
@@ -282,14 +301,21 @@ export function ChatSidebarContainer({ scene, canSpawnMessages, presences, occup
...
-        {messageGroups.map(({ id, systemMessage, ...rest }) => {
+        {messageGroups.map(entry => {
+          const { id, systemMessage, type } = entry;
           if (systemMessage) {
-            return <SystemMessage key={id} {...rest} />;
+            return <SystemMessage key={id} {...entry} />;
           } else {
-            return <ChatMessageGroup key={id} {...rest} />;
+            if (type === "permission") {
+              return !isMod && <PermissionMessageGroup key={id} {...entry} />;
```
netpro2k, line 310: Again should mods really not see these messages?

netpro2k (commented):
Mostly lgtm. I focused mostly on just auditing hooks/effects and didn't dig very deep on the actual components or css. Made a few notes, biggest one I would change before shipping is moving the toggling of the actual mic state out of an effect since its pretty weird to have the UI dictating that.

src/react-components/room/useMicrophone.js
keianhzo: This is mostly because the useMicrophone hook is used in MicSetupModal and the audio popover as that's where we want to conditionally start the mic based on if the voice permission is enabled. I don't have a strong opinion about this but I've moved it back to where we had it before, in ui-root in case that makes more sense.

src/components/avatar-audio-source.js
keianhzo: Ops, I missed that. I've added the condition branch.

src/message-dispatch.js
keianhzo: Yeah, we are relying on the key for the timestamp as it should be unique, that's why this.

keianhzo: Oh yep, I've updated to parse.

src/react-components/presence-log.js
keianhzo: Not sure if I get this one. Entries is the raw message data, logEntries are the DOM elements processed via map.

src/react-components/presence-log.js
keianhzo: I think in our case we want the room owners to be the ones with voice/chat permissions so I've introduced a `useRole` hook to check for owner permissions and add some consistency.

src/react-components/room/ChatSidebarContainer.js
keianhzo: I don't think we really considered that case. I've updated so other moderators also can see the permission updates notifications.

src/react-components/room/ChatSidebarContainer.js
keianhzo: Updated.

src/react-components/room/RoomSettingsSidebarContainer.js
keianhzo: We are not sending messages for room permissions updates. I guess we could just listen to `permission_update` events and call `hubChannel.receive` locally instead?

netpro2k: We send messages for other permission updates like "allow object spawning" and "allow drawing" no? Shouldn't this be the same?

src/react-components/presence-log.js
netpro2k: Right. But why have an effect and state instead of mapping in render like we were doing before. In either case we still have to render whenever a new entry comes in, so this just seems like an added layer of indirection.

src/react-components/presence-log.js
netpro2k: Should this be using `useRole` then?

src/react-components/room/useMicrophone.js
netpro2k: Having MicSetupModal have an effect that initially enables the mic makes sense, since that's part of what that screen does. But having it here means this ends up being what is re-enabling the mic in response to permission changes which seems wrong. MediaDeviceManager is what handles turning disabling it when permissions change, so I would think in the same spot it should also handle re-enabling it.

src/react-components/presence-log.js
keianhzo: I did that in https://github.com/mozilla/hubs/pull/5731/commits/960262572ee32e244d7bb07b672e504cfa6a061b maybe you are referring to some other one. I think I've moved all of them.

src/react-components/room/RoomSettingsSidebarContainer.js
keianhzo: We don't do for room permissions although if we are doing it for chat/voice probably we should also do for the rest? Maybe that could be part of a follow-up PR as this is how it's in production right now.

src/react-components/room/useMicrophone.js
keianhzo: There are two different things. One is mic sharing and another one is mic enabling. Mic sharing just triggers the sharing process and ask for permission if needed. Enabling pauses/resumes the producer so we are effectively sending voice data.

In `ui-root` we are listening to voice permission updates to start sharing the mic during the different entry steps.
In the `MediaDevicesManager` we are listening to voice permission updates to pause the mic producer and we rely on the user to manually resume it as it might be tricky if the voice permission is re-enabled by a moderator and that forces the user's mic to open.

src/react-components/presence-log.js
netpro2k: ah must have been looking at old code.

src/react-components/room/RoomSettingsSidebarContainer.js
netpro2k: Yeah seems worth changing on the server side so this can be removed as a special case.

src/react-components/room/useMicrophone.js
netpro2k: Ah right, that makes sense. We should still probably move this effect out of `useMicrophone` though, since if anything happens to be using it when a permission change happens its going to trigger, since it has `canVoiceChat` as a dependency, and `useCan` listens for permission change events.

Issue #5415, opened 2022-05-07
Depth of field shader


jywarren:
**Is your feature request related to a problem? Please describe.**

No, it's just a nice-to-have!

**Describe the solution you'd like**

I was using SketchFab and they have a really nice depth of field effect that I think would be really lovely in Hubs. Here's a model I'm using in hubs, with DOF and some other filters added

<img width="809" alt="image" src="https://user-images.githubusercontent.com/24359/167272579-27f1827b-3456-45c3-8e78-9e2c586b3f2a.png">

**Describe alternatives you've considered**

Manually blurring faraway things... obviously not flexible for large areas, and doesn't affect things near where a user can go close.

**Additional context**

I found a number of references to A-Frame custom shader use, including:

> In addition, there are components developed by the A-Frame developer community that allow the use of existing shaders from repositories such as [ShaderToy](https://shadertoy.com/) and [ShaderFrog](https://github.com/chenzlabs/aframe-import-shaderfrog). (https://aframe.io/docs/1.3.0/components/material.html#using-a-custom-shader-and-component-together
)

However I wasn't able to find more about this. There are definitely plenty of nice examples of depth of field (DOF) shaders out there. 

Three.js has an example too:

https://threejs.org/examples/webgl_postprocessing_dof.html (discussed at https://stackoverflow.com/questions/45842013/three-js-depth-of-field)

Also this one: https://threejs.org/examples/webgl_postprocessing_dof2

Aha -- OK, this is the thread where folks are trying to integrate a post-processing shader API into A-Frame: 

- https://github.com/aframevr/aframe/pull/3645
- https://github.com/mrdoob/three.js/pull/18846

It seems like a pretty complex project, things haven't made a lot of progress recently. Much respect for the folks working on this!

I think based on this research this isn't yet actionable, but if anyone knows shaders and A-Frame code very well maybe they could use some help upstream. 

Thanks, all!

jywarren:
This is happening!!! At least the postprocessing plugin has been added: https://github.com/mozilla/hubs/pull/5742

PR #5742, opened 2022-10-11
Post effects

netpro2k:
This PR adds support for Post Processing effects, starting with Bloom. At first this will be double opt-in, with both users needing to enable a preference, and scene authors needing to export with an updated environment-settings component configured for post effects. The plan is to make the user preference default enabled eventually on systems that can handle it. It is currently not supported in VR and the preference will be hidden if we detect the user is on a standalone VR headset.

Currently when opting into the new "HDR pipeline", tonemapping settings will be ignored and ACES will always be used. It's likely we will want this to be configurable eventually but supporting our existing tonemapping options (namely the Blender LUT one) is a bit complicated and requires some changes to the underlying postprocessing library. Its possible we will want to standardize on ACES and then offer additional LUT after tonemaping for artistic effects. We also likely want to explore adaptive exposure, which will be difficult to support for all tonemapping modes.

This PR also slightly re-organizes the graphics options in the preferences screen. Ideally we will want to eventually have a separate tab for this, but doing so requires rethinking the UI a bit for mobile so is out of scope for this PR. That likely would pair well with a PR around better configuration of graphics defaults.

Blender addon: https://github.com/MozillaReality/hubs-blender-exporter/pull/142


src/systems/camera-system.js
```diff
@@ -409,14 +406,7 @@ export class CameraSystem {
...
         position.y = position.y - 1.6;
-        setMatrixWorld(
-          this.avatarRig.object3D,
-          new THREE.Matrix4().compose(
-            position,
-            quat,
-            scale
-          )
-        );
+        setMatrixWorld(this.avatarRig.object3D, new THREE.Matrix4().compose(position, quat, scale));
```
takahirox, line 409: Better to avoid to create a new matrix instance in tick (Sorry, this is an existing issue. Making another PR is ok.)

takahirox:
Thanks for the PR.

> This PR also slightly re-organizes the graphics options in the preferences screen

Would you mind putting the screenshot of the new preferences screen here?

src/app.ts
```diff
@@ -142,20 +150,20 @@ export class App {
...
+
     const renderer = new WebGLRenderer({
-      // TODO we should not be using alpha: false https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API/WebGL_best_practices#avoid_alphafalse_which_can_be_expensive
       alpha: true,
-      antialias: true,
-      depth: true,
-      stencil: true,
-      premultipliedAlpha: true,
-      preserveDrawingBuffer: false,
-      logarithmicDepthBuffer: false,
```
takahirox, line 153: Just to clarify, you have removed `premultipliedAlpha`, `preserveDrawingBuffer`, and `logarithmicDepthBuffer` because they use the default values?

https://threejs.org/docs/#api/en/renderers/WebGLRenderer.preserveDrawingBuffer

src/app.ts
```diff
@@ -166,7 +174,7 @@ export class App {
 
     sceneEl.appendChild(renderer.domElement);
 
-    const camera = new PerspectiveCamera(80, window.innerWidth / window.innerHeight, 0.05, 10000);
+    const camera = new PerspectiveCamera(80, canvas.width / canvas.height, 0.05, 10000);
```
takahirox, line 177: Curious, what is this change for?

src/app.ts
```diff
@@ -179,13 +187,25 @@ export class App {
...
+    // We manually call scene.updateMatrixWolrd in mainTick
+    scene.autoUpdate = false;
+
+    if (enablePostEffects) {
+      this.fx = createEffectsComposer(canvas, renderer, camera, scene, sceneEl, this.store);
+    } else {
+      (sceneEl as any).addEventListener("rendererresize", function ({ detail }: { detail: DOMRectReadOnly }) {
+        renderer.setSize(detail.width, detail.height, true);
+      });
+    }
```
takahirox, line 201: Does here mean render resize handling is done in the effect composer if post effects is enabled? (Maybe adding comments is good)

src/effects.ts
```diff
@@ -0,0 +1,203 @@
...
+import qsTruthy from "./utils/qs_truthy";
+import ImageMessage from "./react-components/image-message";
+
+export enum AAModes {
+  NONE = "NONE",
+  SMAA = "SMAA",
+  MSAA_2X = "MASS_2X",
+  MSAA_4X = "MASS_4X",
+  MSAA_8X = "MASS_8X"
+}
```
takahirox, line 27: Just wondering whether we really need five aa options because it may add complexity in the preferences. For example aren't just two options (ON/OFF) good? No big preference tho...

src/app.ts
```diff
@@ -179,13 +187,25 @@ export class App {
     };
 
     this.world.scene = sceneEl.object3D;
+    const scene = sceneEl.object3D;
+
+    // We manually call scene.updateMatrixWolrd in mainTick
+    scene.autoUpdate = false;
```
takahirox, line 193: Is this manual updating for post processing? Can auto update break the post processing?

src/react-components/room/useResizeViewport.js
```diff
@@ -38,84 +38,70 @@ export function useResizeViewport(viewportRef, store, scene) {
...
-  useEffect(
-    () => {
-      const observer = new ResizeObserver(entries => {
-        const isPresenting = scene.renderer.xr.isPresenting;
-        const isVRPresenting = scene.renderer.xr.enabled && isPresenting;
+      // Do not update renderer, if a camera or a canvas have not been injected.
+      // In VR mode, three handles canvas resize based on the dimensions returned by
+      // the getEyeParameters function of the WebVR API. These dimensions are independent of
+      // the window size, therefore should not be overwritten with the window's width and
+      // height, // except when in fullscreen mode.
```
takahirox, line 67: Extra `//` after `height,`

src/react-components/room/useResizeViewport.js
```diff
@@ -38,84 +38,70 @@ export function useResizeViewport(viewportRef, store, scene) {
...
+      const isVRPresenting = scene.renderer.xr.enabled && isPresenting;
 
-  useEffect(
-    () => {
-      const observer = new ResizeObserver(entries => {
-        const isPresenting = scene.renderer.xr.isPresenting;
-        const isVRPresenting = scene.renderer.xr.enabled && isPresenting;
+      // Do not update renderer, if a camera or a canvas have not been injected.
+      // In VR mode, three handles canvas resize based on the dimensions returned by
+      // the getEyeParameters function of the WebVR API. These dimensions are independent of
```
takahirox, line 65: Does this mean Three.js still handles WebVR API? I thought it has dropped WebVR API support...

src/react-components/preferences-screen.js
```diff
@@ -1247,6 +1224,102 @@ class PreferencesScreen extends Component {
...
+            key: "enableBloom",
+            prefType: PREFERENCE_LIST_ITEM_TYPE.CHECK_BOX,
+            disableIfFalse: "enablePostEffects",
+            promptForRefresh: true
+          },
+          {
+            key: "aaMode",
+            prefType: PREFERENCE_LIST_ITEM_TYPE.SELECT,
+            disableIfFalse: "enablePostEffects",
+            promptForRefresh: true,
```
takahirox, line 1283: Do `enableBloom` and `aaMode` toggles really require the refresh? Can't we skip the effect passes if they are turned off?

src/effects.ts
```diff
@@ -0,0 +1,203 @@
...
+  canvas: HTMLCanvasElement,
+  renderer: WebGLRenderer,
+  camera: Camera,
+  scene: Scene,
+  sceneEl: AScene,
+  store: Store
+) {
+  const composer = new EffectComposer(renderer, {
+    frameBufferType: HalfFloatType
+  });
```
takahirox, line 39: The core post-processing logic is in an external library `postprocessing`. I don't think I have time to check the lib internal so want to check if it works fine on the test after landing the post-processing feature.

keianhzo (changes requested):

keianhzo:
Note: I mistakenly set the review to "request changes" but my comments are mostly questions.

src/app.ts
netpro2k: Yep, exactly

src/app.ts
netpro2k: Ah that may not actually be needed anymore. Early on when working on this I was using a static size canvas to simplify things. This value doesn't particularly matter anyway since it gets overwritten again once useResizeViewport runs, but I will ditch the change.

We end up doing way more resizing of the canvas than I would like but its a bit tricky to untangle right now since our initialization is still driven by aframe, then react, so at this point we don't actually know what the dimensions of the canvas will end up being.

src/effects.ts
netpro2k: I think we want at least off, SMAA, and some form of MSAA. I mostly included all of these so that people can test different options for visual differences and perf on different devices.

src/app.ts
takahirox: I hope we can have a separated clean up PR from a logic addition/change PR because review can be easier.

src/systems/camera-system.js
netpro2k: Yeah this was just a formatting change. I see a few spots in here we are creating some garbage.

src/app.ts
netpro2k: Post processing can call render multiple times, so it makes more sense to just manually update ourselves once. This is also something we wanted to do anyway so that we can easily see where in the frame matrix updates are happening and schedule systems before/after that.

src/systems/environment-system.js
```diff
@@ -61,30 +71,41 @@ export class EnvironmentSystem {
...
+    bloomController.enable(debugSettings.enableHDRPipeline);
+
+    const bloomFolder = gui.addFolder("bloom");
+    bloomFolder.show(debugSettings.enableHDRPipeline && debugSettings.enableBloom);
+    bloomFolder.add(debugSettings.bloom, "intensity", 0, 10, 0.01).onChange(updateDebug).listen();
+    bloomFolder.add(debugSettings.bloom, "threshold", 0, 10, 0.001).onChange(updateDebug).listen();
+    bloomFolder.add(debugSettings.bloom, "radius", 0, 1, 0.001).onChange(updateDebug).listen();
+    bloomFolder.add(debugSettings.bloom, "smoothing", 0, 1, 0.001).onChange(updateDebug).listen();
+    // bloomFolder.add(bloom.blendMode, "blendFunction", BlendFunction);
+    // bloomFolder.add(bloom.blendMode.opacity, "value", 0, 1).name("Opacity");
```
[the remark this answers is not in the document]
netpro2k: Yeah I agree. I was debating if I should just remove this debug stuff altogether but it seems really useful for early testing on this. I agree we should figure out a nicer way to split out debug logic.

src/effects.ts
netpro2k: Yep. Its definitely worth us reading through and understanding the internals well. I have read a lot of it myself, particularly for the effects we are using.

src/react-components/preferences-screen.js
netpro2k: I think both of these could probably be made to work but it seemed simpler just to have users refresh for now. Toggling the main option definitely requires a refresh since it changes how we initialize our WebGL context.

src/app.ts
takahirox: > Post processing can call render multiple times, so it makes more sense to just manually update ourselves once.

OK, makese sense to me. I thought `postprocessing` lib set `scene.autoUpdate` `false` but if it doesn't, setting `scene.autoUpdate` `false` and manually calling update matrix world in our code makes sense.

src/react-components/room/useResizeViewport.js
netpro2k: I think this comment is just out of date. It is still true though that WebXRManager handles the size of its framebuffers internally.

src/effects.ts
netpro2k: I did notice the constants for MSAA all have a typo though.. will fix

src/effects.ts
```diff
@@ -0,0 +1,203 @@
...
+        f.add(bloom.blendMode, "blendFunction", BlendFunction);
+        f.add(bloom.blendMode.opacity, "value", 0, 1).name("Opacity");
+      }
+      gui.add({ aaMode }, "aaMode", AAModes).onChange(applyAAMode);
+      gui.add(debugPass, "enabled").name("debug");
+    }, 0);
+  }
+
+  (sceneEl as any).addEventListener("rendererresize", function ({ detail }: { detail: DOMRectReadOnly }) {
+    composer.setSize(detail.width, detail.height, true);
```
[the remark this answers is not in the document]
netpro2k: Ah yeah good catch. Updated here and in EffectComposer `setSize` call as well. As written its actually a bug since sometimes we want to render at a lower resolution than the canvas is sized to.

keianhzo (approved):

jywarren:
Many wows! ❤️ This is very exciting!

Looking at https://www.npmjs.com/package/postprocessing, i see it can support depth of field (#5415), god rays, ambient occlusion, glitch, pixellation, and others. 

Nice demos too: https://pmndrs.github.io/postprocessing/public/demo/#depth-of-field

Are there plans to enable many of these? 

Also, I noticed a difference in how some items are lit, or maybe opacity differences, I believe, when this is enabled; these are alpha-transparent cloud PNGs in a particle system to simulate smoke; i can just adjust settings to fix but in case it's an issue you want to see reported:

Before:

Screen Shot 2022-12-11 at 2 16 00 PM

After:

Screen Shot 2022-12-11 at 2 16 48 PM

Issue #5781, opened 2022-11-09
LOD + SkinnedMesh can cause error


takahirox:
**Description**

If LOD contains SkinnedMesh object is placed, Hubs Client causes an error. 

**To Reproduce**
1. Go to https://hubs.mozilla.com
2. Create and enter a room
3. Upload a glTF file having LOD containing SkinnedMesh

**Expected behavior**

No Error

**Screenshots**

![image](https://user-images.githubusercontent.com/7637832/200952075-cc3b6d11-7adb-4a72-8eb0-874a817ec20f.png)

**Hardware** 
 - Device: Desktop but probably any
 - OS: Windows but probably any
 - Browser: Chrome but probably any

**Additional context**

We seem to need to update Three.js GLTFLoader and [three-gltf-extensions](https://github.com/takahirox/three-gltf-extensions/tree/main/loaders/MSFT_lod) to resolve the problem

PR #5782, opened 2022-11-10
Upgrade three-gltf-extensions

takahirox:
to ignore LOD for SkinnedMesh as workaround

Refer to https://github.com/takahirox/three-gltf-extensions/commit/b4737a2b2d855697b4b7391d25786edf2d946223 for the update.

From #5781


takahirox:
Emergently merging this because this PR because the change is simple and the problem is serious (if someone in a room upload LOD+SkinnedMesh model, all the clients in the room can be frozen).

PR #5786, opened 2022-11-11
Use qsTruthy to get the query param value

keianhzo:
In the context of this https://github.com/mozilla/hubs-compose/pull/16 I've seen that we are not currently respecting the user's force_turn and force_tcp parameters values.


keianhzo:
Forget about this, this shouldn0t be an issue.

PR #5802, opened 2022-11-21
Convert networking system to typescript

johnshaughnessy:
Convert networking system to typescript in preparation for future changes

### DIFF
--- a/src/bit-systems/networking.ts
+++ b/src/bit-systems/networking.ts
@@ -1,4 +1,14 @@
-import { addComponent, defineQuery, enterQuery, exitQuery, hasComponent, removeComponent, removeEntity } from "bitecs";
+import {
+  addComponent,
+  Component,
+  defineQuery,
+  enterQuery,
+  exitQuery,
+  hasComponent,
+  removeComponent,
+  removeEntity
+} from "bitecs";
+import { HubsWorld } from "../app";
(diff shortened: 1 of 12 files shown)


types/networked-aframe.d.ts
```diff
@@ -0,0 +1,22 @@
+import { AElement } from "aframe";
+
+// TODO Move this to a declaration file.
```
netpro2k, line 3: This TODO is no longer needed, this is a declaration file

src/bit-systems/networking.ts
```diff
@@ -36,11 +42,25 @@ export function takeOwnership(world, eid) {
   }
 }
 
-const createMessageDatas = new Map();
+type EntityID = number;
```
netpro2k, line 45: Might be nice to start using this universally just for ease of reading. Tradeoff is tricking ourselves into thinking its something special and making EntityID specific things when we are actually just doing number stuff.

netpro2k (commented):
lgtm other than the extra TODO

src/bit-systems/networking.ts
johnshaughnessy: It feels "correct" to call them `EntityID`s because (even though this isn't enforced), `EntityID`s are a narrower type than `number`s: They are (supposed to be) `uints` between `0` and `MAX_ENTITY_COUNT`.

PR #5803, opened 2022-11-22
Fix bug that causes missed naf / nafr messages

johnshaughnessy:


src/phoenix-adapter.js
```diff
@@ -104,8 +129,8 @@ export default class PhoenixAdapter {
 
   disconnect() {
     this.hubChannel.presence.list(key => key).forEach(this.nafOccupantLeave);
-    this.hubChannel.channel.off("naf", this.refs.get("naf"));
-    this.hubChannel.channel.off("nafr", this.refs.get("nafr"));
+    this.hubChannel.channel.off("naf");
+    this.hubChannel.channel.off("nafr");
```
johnshaughnessy, line 133: It appears that the second argument is unused: https://github.com/mcampa/phoenix-channels/blob/1eb71f723d3e76beb5d7600284eaa6910683ab8d/src/channel.js#L72

PR #5814, opened 2022-11-24
Feature/networking system refactor

johnshaughnessy:
Refactor networking system code so that each piece is smaller.

No behavior changes. 

Built on https://github.com/mozilla/hubs/pull/5813

### DIFF
--- a/src/bit-systems/network-send-system.ts
+++ b/src/bit-systems/network-send-system.ts
@@ -8,9 +8,9 @@ import {
   localClientID,
   pendingJoins,
   isNetworkInstantiated,
-  createMessageDatas,
-  EntityID
+  createMessageDatas
 } from "./networking";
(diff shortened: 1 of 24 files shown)


src/bit-systems/network-send-system.ts
```diff
@@ -0,0 +1,86 @@
...
+
+function isNetworkInstantiatedByMe(eid: EntityID) {
+  return isNetworkInstantiated(eid) && Networked.creator[eid] === APP.getSid(NAF.clientId);
+}
+
+const ticksPerSecond = 12;
+const millisecondsBetweenTicks = 1000 / ticksPerSecond;
+let nextTick = 0;
+
+export const ownedNetworkedQuery = defineQuery([Owned, Networked]);
```
johnshaughnessy, line 17: This doesn't need to be exported. It's only used here.

PR #5815, opened 2022-11-24
Pinning, networking, and waypoints

johnshaughnessy:
This PR makes changes to the networking system to support waypoints. It re-imagines pinning for new-netcode objects, but the actual calls to the backend are disabled until we write some reticulum code to separate new writes from existing room objects. It also makes some necessary tweaks to media frames and reimplements waypoints.


src/bit-systems/waypoint.ts
```diff
@@ -0,0 +1,209 @@
...
+      canBeOccupied &&
+      !NetworkedWaypoint.occupied[eid] &&
+      findAncestorWithComponent(world, SceneRoot, eid)
+    );
+  });
+}
+
+function* tryOccupyAndSpawn(world: HubsWorld, characterController: CharacterControllerSystem, spawnPoint: EntityID) {
+  moveToWaypoint(world, spawnPoint, characterController, true);
+  takeOwnershipWithTime(world, spawnPoint, 0);
```
johnshaughnessy, line 76: It seem this would fail if the waypoint was ever owned by anyone before us. 

We didn't think we could do a regular `takeOwnership` here, because we didn't want to assume that we'll have already received netcode messages about this entity. (We wanted to take "soft" ownership, instead.)

It would be nice for us to avoid having to add code here (and for networked videos, for example) that said, "Wait long enough to be sure that no one is trying to tell us about their ownership over this entity so we don't accidentally takeover that state." But in this case, the timestamp `0` doesn't seem correct.

Maybe this should be `takeOwnershipWithTime(world, spawnPoint, Networked.lastOwnerTime[spawnPoint] + 1)` , which will be `1` if we've never received a message about this entity. If it's not `1`, then it's as if we're saying, "I want to take ownership and believe I can according to the information I've received as of this timestamp."

johnshaughnessy: (This `lastOwnerTime + 1` is also what we do when taking ownership of the entities that are left behind by a parting client.)

src/utils/message-for.ts
```diff
@@ -63,3 +93,114 @@ export function messageFor(
...
+    deletes: []
+  };
+
+  objects.forEach(obj => {
+    const nid = obj.name;
+    const initialData: MediaLoaderParams = {
+      src: obj.extensions.HUBS_components.media.src,
+      resize: false,
+      recenter: false,
+      animateLoad: false,
```
johnshaughnessy, line 176: Maybe `resize`, `recenter` and `animateLoad` should be _true_, since that's presumably what would have happened to this object in the legacy system.

src/utils/message-for.ts
```diff
@@ -63,3 +93,114 @@ export function messageFor(
...
+    message.creates.push(createMessage);
+
+    const updateMessage: StorableUpdateMessage = {
+      data: {
+        "networked-transform": {
+          version: 1,
+          data: {
+            position: obj.translation,
+            rotation: obj.rotation,
+            scale: [1, 1, 1]
```
johnshaughnessy, line 189: We should also read scale, and also note that any of these can be undefined. https://github.com/mozilla/hubs/blob/master/src/utils/pinned-entity-to-gltf.js#L20-L22

src/utils/network-schemas.ts
```diff
@@ -13,10 +28,48 @@ interface NetworkSchema {
...
 export const schemas: Map<Component, NetworkSchema> = new Map();
-schemas.set(NetworkedMediaFrame, defineNetworkSchema(NetworkedMediaFrame));
-schemas.set(NetworkedTransform, defineNetworkSchema(NetworkedTransform));
-schemas.set(NetworkedVideo, defineNetworkSchema(NetworkedVideo));
+schemas.set(NetworkedMediaFrame, NetworkedMediaFrameSchema);
+schemas.set(NetworkedTransform, NetworkedTransformSchema);
+schemas.set(NetworkedVideo, NetworkedVideoSchema);
+schemas.set(NetworkedWaypoint, NetworkedWaypointSchema);
+
+// TODO: Write rest of schema for waypoints
```
johnshaughnessy, line 41: Remove this leftover TODO

src/utils/networked-waypoint-schema.ts
```diff
@@ -0,0 +1,28 @@
...
+  if (version !== 1) return false;
+  return true;
+}
+
+export const NetworkedWaypointSchema: NetworkSchema = {
+  componentName: "networked-waypoint",
+  serialize: runtimeSerde.serialize,
+  deserialize: runtimeSerde.deserialize,
+  // TODO: Strange to have to provide serializeForStorage for components
+  // that will not go to storage.
```
johnshaughnessy, line 20: Perhaps if a schema doesn't provide `serializeForStorage` and `deserializeFromStorage`, we can interpret that as, "don't store data for this component"

src/utils/store-networked-state.ts
```diff
@@ -0,0 +1,38 @@
...
+  takeOwnership(world, eid);
+  Networked.creator[eid] = APP.getSid("reticulum");
+
+  const nid = APP.getString(Networked.id[eid])!;
+
+  const storableMessage = messageForStorage(world, [eid], [eid], []);
+  const fileId = null;
+  const fileAccessToken = null;
+  const promotionToken = null;
+  console.log("Pinning:", { nid, storableMessage });
```
johnshaughnessy, line 25: Delete log

src/utils/store-networked-state.ts
```diff
@@ -0,0 +1,38 @@
...
+}
+
+export async function tryUnpin(world: HubsWorld, eid: EntityID, hubChannel: HubChannel) {
+  if (!localClientID) throw new Error("Tried to unpin before connected to the channel...");
+  takeOwnership(world, eid);
+  Networked.creator[eid] = APP.getSid(localClientID!);
+  const message = messageFor(world, [eid], [eid], [eid], [], false)!;
+  unpinMessages.push(message);
+  const fileId = null;
+  console.log("this is where i would unpin...");
```
johnshaughnessy, line 36: Delete log

src/systems/bit-media-frames.js
```diff
@@ -215,9 +229,32 @@ function mediaTypesOf(world, entities) {
   return mask;
 }
 
+const takeOwnershipOnTimeout = new Map();
 const heldQuery = defineQuery([Held]);
 // const droppedQuery = exitQuery(heldQuery);
 export function mediaFramesSystem(world) {
+  enteredMediaFramesQuery(world).forEach(eid => {
+    if (Networked.owner[eid] === APP.getSid("reticulum")) {
+      takeOwnershipOnTimeout.set(
```
netpro2k, line 238: Also for this timer pattern in general, like we have discussed a few times in the past already.

src/prefabs/waypoint-preview.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+let model: Object3D | null;
+export async function loadWaypointPreviewModel() {
+  model = ((await loadModel(modelSrc, null, false)) as LoadModelResult).scene;
+}
+
+export function WaypointPreview() {
+  if (!model) throw new Error("Must load the model before creating waypoint preview entity.");
+
+  return <entity name="Waypoint Preview" waypointPreview model={{ model: model! }}></entity>;
```
netpro2k, line 17: Kind of surprised you need the `!` here, the control flow should collapse the type...

src/change-hub.js
```diff
@@ -62,15 +83,32 @@ export async function changeHub(hubId, addToHistory = true, waypoint = null) {
...
+    unloadRoomObjects();
+  }
   NAF.connection.connectedClients = {};
   NAF.connection.activeDataChannels = {};
+  if (pendingMessages.length || storedUpdates.size) {
+    console.warn(
+      `Deleting ${pendingMessages.length + storedUpdates.size} unapplied network messages from previous hub.`
+    );
+    pendingMessages.forEach(console.warn);
+    for (const update of storedUpdates.entries()) {
```
netpro2k, line 96: excessive logging, and they are labeled as warnings.

src/change-hub.js
```diff
@@ -45,7 +59,14 @@ export async function changeHub(hubId, addToHistory = true, waypoint = null) {
   const hub = data.hubs[0];
 
   if (addToHistory) {
-    window.history.pushState(null, null, hubUrl(hubId, {}, hub.slug, waypoint));
+    const search = new URLSearchParams(location.search);
+    const params = {};
+    for (const [key, value] of search.entries()) {
```
netpro2k, line 64: I assume this was just an unrelated fix?

src/change-hub.js
```diff
@@ -29,10 +33,20 @@ export async function changeHub(hubId, addToHistory = true, waypoint = null) {
     console.log("Change hub called with the current hub id. This is a noop.");
     return;
   }
+
   // Suppress on-screen join and leave messages until we receive a sync.
   APP.hideHubPresenceEvents = true;
   const scene = AFRAME.scenes[0];
 
+  // Generate leave events for everyone in the room.
```
netpro2k, line 41: Why do we need to synthetically do this? Since we area actually leaving the room shouldn't we already get these for real?

src/inflators/waypoint.tsx
```diff
@@ -0,0 +1,63 @@
...
+import spawnPointSrc from "../assets/images/sprites/notice/spawn-point.png";
+import { Networked, NetworkedWaypoint, Waypoint } from "../bit-components";
+import { WaypointFlags } from "../bit-systems/waypoint";
+import { AlphaMode } from "../utils/create-image-mesh";
+import { createElementEntity, createRef, renderAsEntity } from "../utils/jsx-entity";
+import { loadTexture, loadTextureFromCache } from "../utils/load-texture";
+import { preload } from "../utils/preload";
+import { ProjectionMode } from "../utils/projection-mode";
+
+// We intentionally do not remove this model from the GLTF Cache
```
netpro2k, line 13: This comment is wrong

src/bit-systems/object-menu.ts
```diff
@@ -30,9 +33,13 @@ function moveToTarget(world: HubsWorld, menu: EntityID) {
   setMatrixWorld(menuObj, targetObj.matrixWorld);
 }
 
-function handleClicks(world: HubsWorld, menu: EntityID) {
+function handleClicks(world: HubsWorld, menu: EntityID, hubChannel: HubChannel) {
   if (clicked(world, ObjectMenu.pinButtonRef[menu])) {
     console.log("Clicked pin");
+    tryPin(world, ObjectMenu.targetRef[menu], hubChannel);
+  } else if (clicked(world, ObjectMenu.unpinButtonRef[menu])) {
+    console.log("Clicked unpin");
```
netpro2k, line 41: log

src/bit-systems/object-menu.ts
```diff
@@ -98,7 +113,7 @@ export function objectMenuSystem(world: HubsWorld, sceneIsFrozen: boolean, useri
   ObjectMenu.targetRef[menu] = objectMenuTarget(world, menu, sceneIsFrozen);
   if (ObjectMenu.targetRef[menu]) {
     moveToTarget(world, menu);
-    handleClicks(world, menu);
+    handleClicks(world, menu, hubChannel);
   }
   render(world, menu, sceneIsFrozen);
```
netpro2k, line 118: Minor, and not part of this PR but calling this `render` is confusing in this context since its not actually rendering anything. I realize you were using it more in the React sense than the Three sense, but we are in engine code here so my brain defaults to the later. In the camera tool system I called the similar function `updateUI`. In this case maybe its just `updateButtonVisibility`

src/bit-systems/scene-loading.ts
```diff
@@ -34,9 +47,10 @@ function* loadScene(world: HubsWorld, eid: number, signal: AbortSignal, environm
       return;
     }
 
+    add(world, scene, loaderEid);
     // TODO: Use a unique id for each scene as the root nid
```
netpro2k, line 51: This TODO is now slightly more appropriate above where we use "scene" as the nid for `setInitialNetworkedData`

src/bit-systems/waypoint.ts
```diff
@@ -0,0 +1,209 @@
...
+      willDisableMotion: false,
+      willDisableTeleporting: false,
+      snapToNavMesh: true,
+      willMaintainInitialOrientation: false
+    });
+  }
+}
+
+let spawnJob: Coroutine | null = null;
+export function moveToSpawnPoint(world: HubsWorld, characterController: CharacterControllerSystem) {
```
netpro2k, line 119: From all the calling sites its quite unclear that after this function call the player would not be at the spawn point. The function signature doesn't tell you either since its not async or a generator. 
1. We should probably try and come up with a better name
2. Should we make it return a promise that resolves when the coroutine finishes? We haven't needed it yet so maybe its not useful, but it would make it more clear that its an async process.

src/bit-systems/waypoint.ts
```diff
@@ -0,0 +1,209 @@
...
+  spawnJob = coroutine(moveToSpawnPointJob(world, characterController));
```
netpro2k, line 120: I don't have good suggestions but when discussing this at one of the meetups I got some confusion around "spawn" here since its also quite common to talk about "spawning" threads or jobs.

src/bit-systems/waypoint.ts
netpro2k: I wonder if this is actually what we always want to do in spots where we were using 0. We should talk through it.

src/bit-systems/waypoint.ts
```diff
@@ -0,0 +1,209 @@
...
+  world: HubsWorld,
+  characterController: CharacterControllerSystem,
+  sceneIsFrozen: boolean
+) {
+  if (exitedOwnedQuery(world).includes(myOccupiedWaypoint)) {
+    myOccupiedWaypoint = 0;
+  }
+
+  waypointQuery(world).forEach(eid => {
+    if (hasComponent(world, NetworkedWaypoint, eid) && hasComponent(world, Owned, eid) && eid !== myOccupiedWaypoint) {
```
netpro2k, line 160: Not sure its avoidable but noting it anyway.. Its still a bit awkward that each system will have to handle this "inherited" case (though maybe most don't need to do anything special?).

src/bit-systems/waypoint.ts
```diff
@@ -0,0 +1,209 @@
...
+      // Inherited this waypoint, clear its occupied state
+      if (NetworkedWaypoint.occupied[eid]) {
+        NetworkedWaypoint.occupied[eid] = 0;
+      }
+    }
+
+    if (hasComponent(world, Interacted, eid)) {
+      if (hasComponent(world, NetworkedWaypoint, eid)) {
+        if (NetworkedWaypoint.occupied[eid]) {
+          console.warn("Can't travel to occupied waypoint...");
```
netpro2k, line 170: They should ideally not even be clickable in the first place.

src/bit-systems/waypoint.ts
```diff
@@ -0,0 +1,209 @@
...
+      } else {
+        moveToWaypoint(world, eid, characterController, false);
+      }
+    }
+
+    const obj = world.eid2obj.get(eid)!;
+    obj.visible = sceneIsFrozen;
+    const isOccupied = hasComponent(world, NetworkedWaypoint, eid) && NetworkedWaypoint.occupied[eid];
+    if (Waypoint.flags[eid] & WaypointFlags.canBeOccupied && obj.children.length) {
+      ((obj.children[0] as Mesh).material as MeshStandardMaterial).color.setHex(isOccupied ? 0xff00aa : 0xffffff);
```
netpro2k, line 185: This was mostly for debugging, not sure if we want to keep this in, if we do we might want a nicer color for occupied.

src/utils/load-room-objects.ts
```diff
@@ -0,0 +1,68 @@
...
+  const legacyRoomObjects: LegacyRoomObject[] = roomData.nodes.filter(node => !isStorableMessage(node));
+
+  if (hubId === APP.hub!.hub_id) {
+    const message = messageForLegacyRoomObjects(legacyRoomObjects);
+    if (message) {
+      message.fromClientId = "reticulum";
+      message.hubId = hubId;
+
+      pendingMessages.push(message);
+      // TODO All clients must use the new loading path for this to work correctly,
```
netpro2k, line 62: I wonder if we should start treating this as a room property instead of a query param. Maybe both.

netpro2k (commented):
A few things we should go over again but mostly looking good and was easier to review even after a long break than expected which is encouraging.

src/bit-systems/network-send-system.ts
```diff
@@ -61,26 +67,50 @@ export function networkSendSystem(world: HubsWorld) {
...
   {
-    // Note: Many people may send delete messages about the same entity
-    const deletedEntities = exitedNetworkedQuery(world).filter(eid => {
-      return !world.deletedNids.has(Networked.id[eid]) && isNetworkInstantiated(eid);
+    const removedEntities = exitedNetworkedQuery(world).filter(isNetworkInstantiated);
+    const deletedEntities = removedEntities.filter(eid => {
+      return (
+        // Don't send delete messages for entities that were not explicitly deleted.
+        !softRemovedEntities.has(eid) &&
+        // Rebroadcast delete messages of entities I created, in case
```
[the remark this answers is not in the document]
johnshaughnessy: I think the case we worried about was with three clients:

I'm client A. I create the entity.
I receive client B `join`, so I send client B a `create` message.
I receive a `delete` and a `leave` from client C.
Client C didn't see client B join before it left, so never sent a `delete` to client B.
If I don't rebroadcast the delete, client B will receive a `create` (from me) and will never be told to delete it.

It's not clear if this is possible, but we didn't know whether we could rely on certain messages being in-order and didn't make assumptions about how much time it takes messages to be distributed to clients.

src/bit-systems/network-receive-system.ts
```diff
@@ -1,23 +1,49 @@
...
+function breakTie(a: ClientID, b: ClientID) {
+  if (a === "reticulum") return b;
+  if (b === "reticulum") return a;
+  // arbitrary (but consistent) tiebreak
+  return a < b ? a : b;
+}
+
+function isOutdatedMessage(eid: EntityID, updateMessage: UpdateMessage) {
+  const entityHasOwner = !!Networked.owner[eid];
+  const messageIsNewer = Networked.lastOwnerTime[eid] < updateMessage.lastOwnerTime;
```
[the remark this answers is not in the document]
johnshaughnessy: > It's not the message that is newer per se, but the owner.

True
> Might read more clearly as `ownerIsNewer`

There are two `owner`s (and `lastOwnerTime`s), so a name like `ownerIsNewer` would have me ask "_Which_ owner is newer?"

I think you would not have named this bool at all so I feel responsible for coming up with a better name... `messageOwnerIsNewer`? 

I think my favorite implementation of this function is still this one, even though `if ( predicate ) return bool` is a code smell and this can be expressed with ANDs and ORs:

```typescript
function isOutdatedMessage(eid: EntityID, updateMessage: UpdateMessage) {
  if (!Networked.owner[eid]) return false;
  if (Networked.lastOwnerTime[eid] < updateMessage.lastOwnerTime) return false;
  if (Networked.lastOwnerTime[eid] > updateMessage.lastOwnerTime) return true;
  return updateMessage.owner !== breakTie(APP.getString(Networked.owner[eid])!, updateMessage.owner);
}
```

src/bit-systems/network-receive-system.ts
```diff
@@ -28,32 +54,42 @@ export function networkReceiveSystem(world: HubsWorld) {
...
-          console.log("Rewriting update message from client who left.", JSON.stringify(update));
-          update.owner = NAF.clientId;
-          update.lastOwnerTime = update.timestamp;
+    for (let j = 0; j < message.deletes.length; j += 1) {
+      const nid = APP.getSid(message.deletes[j]);
+      if (world.deletedNids.has(nid)) continue;
+      world.deletedNids.add(nid);
+
+      const eid = world.nid2eid.get(nid);
+      if (eid) {
```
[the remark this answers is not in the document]
johnshaughnessy: I removed the stored updates for this entity. Its children's stored updates are trickier to remove. We should be able to match the beginning of their nids or something like that. See  https://github.com/mozilla/hubs/commit/abcdd8ffcb21b83fb425e208347f0d2eeb823b71#diff-e482992a52e7c00a4900660b7e8616aecf992b5e7144b3fbfcfde87045315924R91-R95

src/bit-systems/networking.ts
```diff
@@ -10,6 +10,17 @@ export const networkedQuery = defineQuery([Networked]);
...
 export const pendingJoins: StringID[] = [];
 export const pendingParts: StringID[] = [];
+export const softRemovedEntities = new Set<EntityID>();
 export function isNetworkInstantiated(eid: EntityID) {
   return createMessageDatas.has(eid);
 }
+
+let reticulum: StringID | undefined;
+export function isPinned(eid: EntityID) {
+  reticulum = reticulum || APP.getSid("reticulum");
```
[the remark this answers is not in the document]
johnshaughnessy: I removed the memoization because I don't actually know the impact and it seems like the kind of code someone would read and assume it was added with more thought / purpose than I actually put into it. https://github.com/mozilla/hubs/commit/7f51dc9a884821fef5b4c110ad3cacde1b15f014

src/bit-systems/network-receive-system.ts
```diff
@@ -71,26 +107,48 @@ export function networkReceiveSystem(world: HubsWorld) {
...
 
-        // If we were hanging onto updates for this nid we can now apply them. And they should be processed before other updates.
-        if (storedUpdates.has(nid)) {
-          console.log("had pending updates for", nidString, storedUpdates.get(nid));
-          Array.prototype.unshift.apply(message.updates, storedUpdates.get(nid));
-          storedUpdates.delete(nid);
+      for (let i = 0; i < updates.length; i++) {
+        const update = updates[i];
+        if (partedClientIds.has(APP.getSid(update.owner))) {
+          console.log("Rewriting update message from client who left.", JSON.stringify(update));
```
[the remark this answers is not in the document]
johnshaughnessy: This behavior seems suspicious.

We wanted to use the `timestamp` of the message as a way to ensure that whoever received the latest information from a previous owner takes ownership, but that could cause valid ownership transfers to be "undone." 

I would be more comfortable if this client used the `lastOwnerTime` or `lastOwnerTime + 1` of the message instead, so that any in-flight messages that we're about to receive take precedence over this stored message.

If we want to ensure that messages with the highest timestamp take precedence over those with lower timestamps, maybe that should be a condition in `isMessageOutdated`: First, check the `lastOwnerTime`. If they're the same, check the `timestamp`. If they're the same, then tiebreak on clientIds.

src/bit-systems/network-receive-system.ts
```diff
@@ -119,17 +177,11 @@ export function networkReceiveSystem(world: HubsWorld) {
...
-          APP.getString(Networked.owner[eid])! < updateMessage.owner) // arbitrary (but consistent) tiebreak
-      ) {
-        console.log(
-          "Received update from an old owner, skipping",
-          updateMessage.nid,
-          Networked.lastOwnerTime[eid],
-          updateMessage.lastOwnerTime
-        );
+      if (isOutdatedMessage(eid, updateMessage)) {
+        console.log("Skipping update from an old owner, skipping", {
```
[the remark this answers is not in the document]
johnshaughnessy: Yes, I would like to have more flexible logging.

src/bit-systems/network-receive-system.ts
```diff
@@ -146,27 +198,24 @@ export function networkReceiveSystem(world: HubsWorld) {
       Networked.lastOwnerTime[eid] = updateMessage.lastOwnerTime;
       Networked.timestamp[eid] = updateMessage.timestamp;
 
-      // TODO HACK simulating a buffer with a cursor using an array
-      updateMessage.data.cursor = 0;
-      for (let s = 0; s < updateMessage.componentIds.length; s++) {
-        const componentId = updateMessage.componentIds[s];
-        const schema = schemas.get(networkableComponents[componentId])!;
-        schema.deserialize(world, eid, updateMessage.data);
+      if (isCursorBufferUpdateMessage(updateMessage)) {
```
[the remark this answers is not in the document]
johnshaughnessy: I wonder if these different intuitions are because of a difference of opinion on how we should handle the `StorableUpdateMessage`s. 

You had expressed a preference for first transforming them into `CursorBufferUpdateMessage`s beforehand, and then applying them the same way as any others. That has the advantage of keeping the `networkReceiveSystem` very simple: It only deals with one update message shape. 

The cost is that we have to be able to write the transformer from JSON to a `CursorBufferUpdateMessage`. This seems more difficult than applying the JSON to a `NetworkedFoo` component, so I opted to add the complexity to the network receive system. 

This is intertwined with however we decide to do serialization / storage so I expect we'll do some iteration here.

As for removing it -- I'm a little hesitant because I feel like it will leave the storage / serialization code especially incomplete / incoherent. (Removing pinning seems OK because we don't want to write to the DB yet, but removing these lines makes this initial attempt at serialization / storage harder to understand and reason about.)

src/bit-systems/network-receive-system.ts
```diff
@@ -71,26 +107,48 @@ export function networkReceiveSystem(world: HubsWorld) {
...
         }
       }
+
+      // Process the stored message before other updates
+      pendingMessages.unshift({ creates: [], updates, deletes: [] });
+      storedUpdates.delete(nid);
     }
+  });
+
+  for (let i = 0; i < pendingMessages.length; i++) {
```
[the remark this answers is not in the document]
johnshaughnessy: I suppose I could go either way, but prefer to split it up like we've done:

```
// Handle parting clients
// Delete entities
// Create entities
// Update entities (include any stored updates)
// Takeover abandoned entities
```

Rather than:

```
// Handle parting clients
// For each message:
//   delete
//   create
//   update (including any stored updates)
// Takeover abandoned entities
```

It seems desirable that it's clear to readers that we don't have to process all of a given client's messages in order before moving on, and that instead the receive system can be focused more broadly on, "How am I going to manipulate the game state next?"

It's not a perfect analogy, but it feels similar to how we've written the ECS systems to be more about, "What's the next transformation to component datas (across all entities) do we need to make?" rather than, "What's the next object/entity we need to update?"

"What's the batch of entities I need to delete?", not "What are all the things I need to do in response to this message".

src/bit-systems/waypoint.ts
johnshaughnessy: (From our conversation) One way to avoid this special case here would be to store _who_ occupies the waypoint rather than just _whether_ it's occupied, and then checking if `NetworkedWaypoint.occupied[eid] === Networked.owner[eid]`.

src/bit-systems/network-receive-system.ts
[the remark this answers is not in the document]
johnshaughnessy: Edit: After discussing, we've settled on using `timestamp` most of the time, not `lastOwnerTime`. We expect there to be _rare_ cases when this results in an unexpected ownership transfer, such as when someone takes ownership of an object just before the previous owner disconnects.

src/utils/networking-types.ts
```diff
@@ -10,19 +10,28 @@ export type ClientID = string;
...
   data: CursorBuffer;
-};
+}
+export interface StorableUpdateMessage extends UpdateMessageBase {
+  data: any; // todo
+}
 export type DeleteMessage = NetworkID;
 export type Message = {
   fromClientId?: ClientID;
+  hubId?: string;
```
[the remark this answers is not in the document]
johnshaughnessy: We don't. Removed!

(I added this thinking that we may need to filter messages before processing them, but instead I don't queue them if we've already changed hubs (and also empty any messages in the queue during fast room switching)).

src/utils/networked-media-frame-schema.ts
```diff
@@ -0,0 +1,37 @@
+import { NetworkedMediaFrame } from "../bit-components";
```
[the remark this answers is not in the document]
johnshaughnessy: (As we discussed). It seems correct to not have to pay the cost (at all) if you don't intend to store the component data.

Still an open question whether you should have to pay the cost when you want to store data in the DB or _only_ when you update component props (that might already have been stored). 

Perhaps we can break the cost in two:
- If you want to store, you need to be explicit about how to store / load the data... This doesn't require versioning but hints at it.
- If you want to make a backwards-compatible change to a stored component, you'll want to think about versioning. Here is an example and some helpers we wrote to help you achieve this.

src/systems/bit-media-frames.js
```diff
@@ -215,9 +229,32 @@ function mediaTypesOf(world, entities) {
   return mask;
 }
 
+const takeOwnershipOnTimeout = new Map();
 const heldQuery = defineQuery([Held]);
 // const droppedQuery = exitQuery(heldQuery);
 export function mediaFramesSystem(world) {
+  enteredMediaFramesQuery(world).forEach(eid => {
+    if (Networked.owner[eid] === APP.getSid("reticulum")) {
```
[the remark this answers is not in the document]
johnshaughnessy: (In this same file) there's the more generic `isOwnedByRet` function, but hopefully most of the time we won't have to worry about `AEntity`s and we can use a function that only checks `Networked.owner[eid]`. https://github.com/mozilla/hubs/blob/f73418e3b1816439e97a8c97c31143735be184ea/src/systems/bit-media-frames.js#L52-L61

I don't like the "scene" in `isSceneOwned`. I also don't love `isOwnedByRet` (/`isOwnedByReticulum`). `Unowned` seems misleading (as if `Networked.owner[eid]` were `0`). Maybe `isServerOwned` or `isRoomOwned`?

src/prefabs/waypoint-preview.tsx
johnshaughnessy: Removed!

src/change-hub.js
johnshaughnessy: Yes

src/bit-systems/object-menu.ts
johnshaughnessy: done

src/bit-systems/waypoint.ts
johnshaughnessy: True, something would have had to go wrong to get to this point.

src/change-hub.js
johnshaughnessy: No -- the people we leave behind will receive a leave event for the client changing hub, but we don't receive leave events for everyone else. (This might be an artifact of how `migrateToHub` / `migrateToChannel` is implemented.)

Related, I've added this `TODO` about using `migrateToHub` in `changeHub`:

```
    // TODO Migrating to a new hub in one step makes state cleanup suspicious.
    //      Would prefer to disconnect, cleanup state, then connect to the new hub.
```

PR #5822, opened 2022-11-29
Replace deprecated prettier eslint integration with the vscode plugin

keianhzo:
The prettier eslint integration has been deprecated so in VSCode we are falling back to the default formatter on save that doesn't follow the same rules as the ones defined in eslintrc.

This only affect people using VSCode.

Read more here: https://github.com/prettier/prettier-vscode/issues/870

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -5,8 +5,20 @@
   "html.format.enable": false,
   // Disable the default javascript formatter
   "javascript.format.enable": false,
-  // Use 'prettier-eslint' instead of 'prettier'. Other settings will only be fallbacks in case they could not be inferred from eslint rules.
-  "prettier.eslintIntegration": true,
+  // Using the prettier-vscode plugin as the old eslint integration has been deprecated. More here: https://github.com/prettier/prettier-vscode/issues/870
+  "editor.defaultFormatter": "esbenp.prettier-vscode",
+  "[javascriptreact]": {
+    "editor.defaultFormatter": "esbenp.prettier-vscode"
+  },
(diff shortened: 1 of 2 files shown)


.vscode/settings.json
```diff
@@ -5,8 +5,20 @@
   "html.format.enable": false,
   // Disable the default javascript formatter
   "javascript.format.enable": false,
-  // Use 'prettier-eslint' instead of 'prettier'. Other settings will only be fallbacks in case they could not be inferred from eslint rules.
-  "prettier.eslintIntegration": true,
+  // Using the prettier-vscode plugin as the old eslint integration has been deprecated. More here: https://github.com/prettier/prettier-vscode/issues/870
+  "editor.defaultFormatter": "esbenp.prettier-vscode",
```
johnshaughnessy, line 9: Is this configuration specifically for vs code users?

keianhzo: I believe everything under the `.vscode/` folder is only applied by VSCode (maybe also some compatible forks?) But it should not affect other editors unless they purposefully read that file.

nickgrato (approved):
This looks good to me, I think the only thing ( and this could be a big change ) that I would change is the actual prettier file that this references. 

I have these in Subscriptions and Marketing project, some of these might be default and unnecessary. 

```
{
  "singleQuote": true,
  "jsxSingleQuote": false,
  "tabWidth": 2,
  "semi": true
}
```
Been a heavy adopter of  single quotes in Javascript and double quotes in HMTL/JSX however.

PR #5828, opened 2022-11-30
Fix scene deletion

bryanenders:
Why
---

The _Pending scenes_, _Approved scenes_, _Featured scenes_, and _Scenes_
admin pages offer a UI to delete scenes.  Scene deletion does not exist
as a back-end feature, so an error occurs.

What
----

Remove bulk action buttons from the affected pages


hrithikwins:
Much needed

robinkwilson (approved):
LGTM Ship it!

PR #5831, opened 2022-12-02
Fix / sometimes being cleared from chatbox

netpro2k:
Sometimes when using `/` to open chat the initial `/` was later cleared, making it difficult to type slash commands. This was due to the value of a controlled input being set in an effect. Since the only thing we were using this effect for anyway was to autoFocus the chat box I simplified this to a property, and have a separate property for specifying initial text.


johnshaughnessy (approved):
Woohoo! Quality of Life instantly improved!

PR #5842, opened 2022-12-04
Add network debug system

johnshaughnessy:


src/bit-systems/network-debug.ts
```diff
@@ -0,0 +1,47 @@
...
+export function networkDebugSystem(world: HubsWorld, scene: Scene) {
+  enteredNetworkedQuery(world).forEach(eid => {
+    const obj = world.eid2obj.get(eid);
+    if (obj) {
+      const textObj = world.eid2obj.get(renderAsEntity(world, NetworkDebugPrefab()))! as TroikaText;
+      texts.set(eid, textObj);
+      scene.add(textObj);
+      textObj.renderOrder = 999;
+      forEachMaterial(textObj, function (mat: Material) {
+        mat.depthTest = false;
```
takahirox, line 26: Let me confirm my understanding. This is for not hiding texts by other closer objects, isn't it?

src/bit-systems/network-debug.ts
```diff
@@ -0,0 +1,47 @@
...
+      });
+    }
+  });
+
+  exitedNetworkedQuery(world).forEach(eid => {
+    removeEntity(world, texts.get(eid)!.eid!);
+    texts.delete(eid);
```
takahirox, line 33: Don't we need to remove `textObj` from `scene`?

src/bit-systems/network-debug.ts
```diff
@@ -0,0 +1,47 @@
...
+import { Text as TroikaText } from "troika-three-text";
+import { HubsWorld } from "../app";
+import { Networked } from "../bit-components";
+import { NetworkDebugPrefab } from "../prefabs/network-debug";
+import { formatComponentProps, formatObjectName } from "../react-components/debug-panel/ECSSidebar";
+import { renderAsEntity } from "../utils/jsx-entity";
+import { forEachMaterial } from "../utils/material-utils";
+import { EntityID } from "../utils/networking-types";
+
+const texts = new Map<EntityID, TroikaText>();
```
takahirox, line 12: System design question:

Is it a good system design that a system has map from entity id to data? I thought in ECS a system shouldn't manage data by itself and all data should be represented as components.

/cc @netpro2k

src/bit-systems/network-debug.ts
```diff
@@ -0,0 +1,47 @@
...
+  exitedNetworkedQuery(world).forEach(eid => {
+    removeEntity(world, texts.get(eid)!.eid!);
+    texts.delete(eid);
+  });
+
+  networkedQuery(world).forEach(eid => {
+    const networkedObj = world.eid2obj.get(eid)!;
+    const textObj = texts.get(eid)!;
+
+    networkedObj.updateMatrices();
```
takahirox, line 40: Let me confirm my understanding. A networked object's translation/rotation/scale can be updated by remote peers but its matrix isn't updated yet because the scene graph will be updated in or right befor renderer.render() but this system needs updated world matrix so updating it here, correct?

src/bit-systems/network-debug.ts
```diff
@@ -0,0 +1,47 @@
...
+    networkedObj.matrixWorld.decompose(textObj.position, textObj.quaternion, _scale);
+    textObj.matrixNeedsUpdate = true;
+
+    textObj.text = formatObjectName(networkedObj) + "\nNetworked " + formatComponentProps(eid, Networked);
```
takahirox, line 44: This system is called every frame, isn't it? Networked component can be updated every frame? If not, updating text only if needed can be less GC pressure. But this sytem is for debug. Inefficiency should be acceptable. Not a blocking issue. But just a comment.

src/bit-systems/network-debug.ts
johnshaughnessy: You're probably right about that, though I don't feel strongly about it one way or another.

Here we are mapping from `EntityID` of the `Networked` entity to the text3D object of the entity with the `NetworkedDebug` component. If we want to manage this state using bitecs, we would add another component (e.g. `NetworkedDebugText` with a single property, which would be the a reference to the `NetworkedDebug` entity:

```ts
export const NetworkDebugText = defineComponent({
  targetRef: Types.eid
})
```

Then, when we render the entity using the `NetworkDebugPrefab`, we'd add the `NetworkDebugText` to the `NetworkedEntity` and set its `targetRef` to the `NetworkDebugPrefab`'s `eid`:

```diff
-      const textObj = world.eid2obj.get(renderAsEntity(world, NetworkDebugPrefab()))! as TroikaText;
-      texts.set(eid, textObj);
+      const debugEid = renderAsEntity(world, NetworkDebugPrefab());
+      addComponent(world, NetworkedDebugText, eid);
+      NetworkedDebugText.targetRef[eid] = debugEid;
+      const textObj = world.eid2obj.get(debugEid)! as TroikaText;
```

Similarly, in the `exitedQuery` we get the eid from the component store:
```diff
   exitedNetworkedQuery(world).forEach(eid => {
-    removeEntity(world, texts.get(eid)!.eid!);
-    texts.delete(eid);
+    const debugEid = NetworkedDebugText.targetRef[eid];
+    removeEntity(world, debugEid);
   });
   ```
   
   And where we are doing the updating, we just use the ref to access the object3D:
   
   ```diff
     networkedQuery(world).forEach(eid => {
     const networkedObj = world.eid2obj.get(eid)!;
-    const textObj = texts.get(eid)!;
+    const textObj = world.eid2obj.get(NetworkedDebugText.targetRef[eid])! as TroikaText;
```

If we like this better then I'll commit these changes and remove the map. 

In practice, the map will be smaller than the component store, but if we're hesitant to add components to store data then we need to solve that problem (e.g. by archetypes). This is something that Dom and I have discussed but have not yet explored.

src/bit-systems/network-debug.ts
johnshaughnessy: Yes, e.g. showing it through walls.

src/bit-systems/network-debug.ts
johnshaughnessy: Removing the object3D associated with an entity from the scene is handled by another system: 

https://github.com/mozilla/hubs/blob/d8e5400f8b123614f5e38dd5dd76d4ca05f47060/src/systems/remove-object3D-system.js#L100

src/bit-systems/network-debug.ts
johnshaughnessy: We call `updateMatrices` _anytime_ we access the `matrixWorld` or `matrix` of an `Object3D`. If it hasn't been updated, then very little work is done. This is part of the contract we follow because of the matrix optimizations we apply on top of threejs. (See commit description here: https://github.com/mozillareality/three.js/commit/b889d4e5617d95caa02064d6c8da5ae2b1f98e1a )

src/bit-systems/network-debug.ts
johnshaughnessy: Agreed. Not worth the effort or added complexity to optimize.

src/bit-systems/network-debug.ts
takahirox: Ah, ok. That is the one that skips the matrices update if it isn't dirty.

src/bit-systems/network-debug.ts
johnshaughnessy: Updated in https://github.com/mozilla/hubs/pull/5842/commits/38f6d6b89d770eb7645f8520736eb1ec22b5b2be

PR #5843, opened 2022-12-05
Fixed the stripping of the src folder on a windows build

albert-boersen:
When building on Windows the 'src' folder was not stripped.


albert-boersen:
Yeah, I know right, it should work. But unfortunately we needed this fix to make it work on our build setup and it also worked for other Windows setups.

PR #5857, opened 2022-12-11
Update browserstack ie11 test to use /demo

netpro2k:
The browserstack ie11 test checks the root of a hub to ensure the "unsupported browser" message doesn't break. The root of HMC is now a marketing site, so change to test to point at /demo


johnshaughnessy (approved):
Good catch

PR #5858, opened 2022-12-11
Japanese file updates December 2022

lavender-snow:
Japanese file has been updated to December 2022.


johnshaughnessy (approved):
Thank you @lavender-snow !

Issue #5859, opened 2022-12-11
Camera control not functioning on mobile with virtual joysticks enabled


LunarS8:
**Description**
I am unable to move inside a virtual room from my mobile device, I have 2 issues.
- When using screens: I can only look around when tapping on the edges of the screen. I can only move by dragging both fingers from the edges and outwards, the tactile window is so short that it moves really slow.
- When using the joysticks: they get stuck! Tried looking around for a bit, it got stuck with no way of fixing it unless I left the room. Same happened with the other one, it got stuck and my avatar would Sprint non-stop.

**To Reproduce**
1. Enter a room
2. Use the joysticks to move around for about 30 seconds
3. See error of either joystick getting stuck
OR
2.Try moving using the screen
3. See that movement only happens when you tap and drag from the edges of the screen

**Expected behavior**
I expect to be able to move around the room without getting the joysticks stuck in one position, they go back to center and I'm able to move them again.
I expect to be able to move by touching my entire screen not just the edges.

**Screenshots**
![Screenshot_20221211-070954](https://user-images.githubusercontent.com/61944097/206903309-2a4213ab-fbeb-4f4f-89ea-2ba24906bc43.png)

**Hardware** 
 - Device: Pixel 6A
 - OS: Android 12
 - Browser: Firefox

**Additional context**
As an FYI, I have the gestures enabled on my phone. I tried hubs in a laptop before and it worked fine.

matthewbcool:
durocher43#8329 from discord:

> I’m having the same issue of not being able to pinch! And the issue is on iphone 13

naoki1219aic:
I have same problrem.
Hardware
Device: iphone7,iphone13pro
Browser: safari.chrome both

nicolascolombo:
Same problem in Iphone XR.
Basically the tap doesn't work using the phone in vertical. It works horizontally.

takahirox:
Confirmed and investigated the problem.

The root issue seems to be z-index fight.

The event listeners are added to the canvas but this Notifications Container component seems to  hide the canvas.

![image](https://user-images.githubusercontent.com/7637832/208541115-db568b53-cc53-4241-9405-58edab335ae0.png)

For example if I set z-index of the Notifications Container dev to -1, the problem can be resolved.

Notifications Container seems to be introduced in #5731 /cc @keianhzo

takahirox:
I found a similar problem even on desktop. If a tip is displayed, the highlighted area is unclickable because it hides the canvas.

[...]

PR #5868, opened 2022-12-21
Update tooltip location to not block the canvas in mobile

keianhzo:
Fixes https://github.com/mozilla/hubs/issues/5859

The Notification container was blocking the canvas in mobile mode. If updated to to be at the top in mobile and bottom in Desktop and only take the required space.


takahirox (commented):
Thanks for working on it. I confirmed that the unclickable problem has been resolved with this change both on mobile and desktop.

But the tooltip looks being placed a bit too top on mobile portrait mode. Is this instentional?

image

keianhzo:
Thanks for testing that @takahirox there were a few possible mobile/viewport size combinations that I missed to test. sI've updated to account for them.

PR #5894, opened 2023-01-13
More bit gltf components

netpro2k:
This PR implements `skybox`, `fog`, `background`, `shadow`, `spawn-point` and `spawner` components in BitECS. Most of them are considered deprecated, but are needed for loading many existing scenes.

The spawner component still needs a bit of work to be grabbable with VR hands, but this requires deeper work on the physics integration into BitECS so leaving that for a later PR.

### DIFF
--- a/src/bit-systems/object-spawner.ts
+++ b/src/bit-systems/object-spawner.ts
@@ -7,6 +7,7 @@ import { coroutine, nextFrame } from "../utils/coroutine";
 import { createNetworkedEntity } from "../utils/create-networked-entity";
 import { EntityID } from "../utils/networking-types";
 import { setMatrixWorld } from "../utils/three-utils";
+import { animateScale } from "./media-loading";
 
 export enum OBJECT_SPAWNER_FLAGS {
   /** Apply gravity to spawned objects */
@@ -44,6 +45,9 @@ function* spawnObjectJob(world: HubsWorld, spawner: EntityID) {
   spawnerObj.visible = false;
   yield sleep(1000);
   spawnerObj.visible = true;
+
+  // TODO we should come up with a nicer way to get at the media that was loaded by a MediaLoader
+  yield* animateScale(world, spawnerObj.children[0]!.eid!);
 }
 
 // TODO type for coroutine
--- a/src/bit-systems/object-spawner.ts
+++ b/src/bit-systems/object-spawner.ts
@@ -1,0 +1,65 @@
+import { addComponent, defineQuery, enterQuery, exitQuery, hasComponent } from "bitecs";
+import { HubsWorld } from "../app";
+import { FloatyObject, Held, HeldRemoteRight, Interacted, MediaLoader, ObjectSpawner } from "../bit-components";
+import { FLOATY_OBJECT_FLAGS } from "../systems/floaty-object-system";
+import { sleep } from "../utils/async-utils";
+import { coroutine, nextFrame } from "../utils/coroutine";
+import { createNetworkedEntity } from "../utils/create-networked-entity";
+import { EntityID } from "../utils/networking-types";
+import { setMatrixWorld } from "../utils/three-utils";
+
+export enum OBJECT_SPAWNER_FLAGS {
+  /** Apply gravity to spawned objects */
+  APPLY_GRAVITY = 1 << 0
+}
+
+function* waitForMediaLoaded(world: HubsWorld, eid: EntityID) {
+  while (hasComponent(world, MediaLoader, eid)) {
+    yield nextFrame();
+  }
+}
+
+function* spawnObjectJob(world: HubsWorld, spawner: EntityID) {
+  const spawned = createNetworkedEntity(world, "media", {
+    src: APP.getString(ObjectSpawner.src[spawner]),
+    recenter: false,
+    resize: false,
+    animateLoad: false,
+    isObjectMenuTarget: true
+  });
+
+  if (ObjectSpawner.flags[spawner] & OBJECT_SPAWNER_FLAGS.APPLY_GRAVITY) {
+    FloatyObject.flags[spawned] &= ~FLOATY_OBJECT_FLAGS.MODIFY_GRAVITY_ON_RELEASE;
+  }
+
+  addComponent(world, HeldRemoteRight, spawned);
+  addComponent(world, Held, spawned);
+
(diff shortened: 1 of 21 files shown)


src/utils/coroutine.js
```diff
@@ -103,6 +103,11 @@ export function cancelable(iter, signal) {
   })();
 }
 
+const nextFramePromise = Promise.resolve();
+export function nextFrame() {
+  return nextFramePromise;
+}
```
johnshaughnessy, line 109: The name `nextFrame` is a little weird to me, since it assumes information about the calling context (i.e. this coroutine that ticks every frame). It's accurate everywhere we've wanted to use it, but still weird to me. I don't have a better name in mind though.

src/utils/network-schemas.ts
```diff
@@ -28,15 +30,19 @@ export interface NetworkSchema {
...
+  deserializeFromStorage?: (eid: EntityID, storedComponent: StoredComponent) => void;
 }
 
 export const schemas: Map<Component, NetworkSchema> = new Map();
 schemas.set(NetworkedMediaFrame, NetworkedMediaFrameSchema);
 schemas.set(NetworkedTransform, NetworkedTransformSchema);
 schemas.set(NetworkedVideo, NetworkedVideoSchema);
 schemas.set(NetworkedWaypoint, NetworkedWaypointSchema);
+schemas.set(NetworkedFloatyObject, {
+  componentName: "floaty-object",
```
johnshaughnessy, line 43: Should this be `networked-floaty-object`? (I guess the `networked-` prefix might be implied... I wasn't as clever with any of the other ones so they all have `networked-` in their names.)

src/utils/network-schemas.ts
```diff
@@ -28,15 +30,19 @@ export interface NetworkSchema {
     writeToShadow: boolean
   ) => boolean;
   deserialize: (world: HubsWorld, eid: EntityID, data: CursorBuffer) => void;
-  serializeForStorage: (eid: EntityID) => StoredComponent;
-  deserializeFromStorage: (eid: EntityID, storedComponent: StoredComponent) => void;
+  serializeForStorage?: (eid: EntityID) => StoredComponent;
+  deserializeFromStorage?: (eid: EntityID, storedComponent: StoredComponent) => void;
```
johnshaughnessy, line 34: :+1:

johnshaughnessy (approved):

src/bit-systems/object-spawner.ts
```diff
@@ -0,0 +1,65 @@
...
+  const spawned = createNetworkedEntity(world, "media", {
+    src: APP.getString(ObjectSpawner.src[spawner]),
+    recenter: false,
+    resize: false,
+    animateLoad: false,
+    isObjectMenuTarget: true
+  });
+
+  if (ObjectSpawner.flags[spawner] & OBJECT_SPAWNER_FLAGS.APPLY_GRAVITY) {
+    FloatyObject.flags[spawned] &= ~FLOATY_OBJECT_FLAGS.MODIFY_GRAVITY_ON_RELEASE;
```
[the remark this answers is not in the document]
netpro2k: Yep. `~` is bitwise not https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_NOT

src/bit-systems/object-spawner.ts
```diff
@@ -0,0 +1,65 @@
...
+
+// TODO type for coroutine
+type Coroutine = () => IteratorResult<undefined, any>;
+const jobs = new Map<EntityID, Coroutine>();
+
+const interactedSpanersEnterQuery = enterQuery(defineQuery([ObjectSpawner, Interacted]));
+const spawnerExitQuery = exitQuery(defineQuery([ObjectSpawner]));
+export function objectSpawnerSystem(world: HubsWorld) {
+  interactedSpanersEnterQuery(world).forEach(spawner => {
+    if (!jobs.has(spawner)) jobs.set(spawner, coroutine(spawnObjectJob(world, spawner)));
```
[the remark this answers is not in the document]
netpro2k: The ritual of each system managing these also feels a bit silly.

src/inflators/model.tsx
```diff
@@ -63,7 +63,20 @@ export function inflateModel(world: HubsWorld, rootEid: number, { model }: Model
     const components = obj.userData.gltfExtensions?.MOZ_hubs_components || {};
     if (components.visible) {
       const { visible } = components.visible;
-      obj.visible = !!visible;
+      obj.visible = visible;
```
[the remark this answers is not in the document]
netpro2k: The `visible` component has 1 property `visible` which is a boolean. We technically don't validate component schemas anywhere, but if we want to it should probably be upstream from here.

src/inflators/spawner.ts
```diff
@@ -0,0 +1,44 @@
...
+  addComponent(world, HandCollisionTarget, eid);
+  addComponent(world, CursorRaycastable, eid);
+  addComponent(world, RemoteHoverTarget, eid);
+  addComponent(world, SingleActionButton, eid);
+
+  addComponent(world, ObjectSpawner, eid);
+  ObjectSpawner.src[eid] = APP.getSid(props.src);
+  ObjectSpawner.flags[eid] = props.mediaOptions?.applyGravity ? OBJECT_SPAWNER_FLAGS.APPLY_GRAVITY : 0;
+
+  // TODO rigidbody+shape for hand grabbing
```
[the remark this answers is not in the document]
netpro2k: Ah perhaps. It should hopefully be much more straightforward though. We also may be able to avoid that whole class of problem if we rethink physics shapes a bit.

src/utils/coroutine.js
netpro2k: Yeah. This is part of the "courtine API" like timers stuff we experimented with so I guess I will add `cr`. I don't like that much either but at least its consistent. Maybe we want to do something like `Coroutine.nextFrame()`

src/utils/network-schemas.ts
netpro2k: No strong opinions here. Would like to get rid of this somehow as its just another thing you may mess up on copy/paste when making a new component (I almost screwed up matching the two instances of `NetworkedFloatyObject` in this block.

PR #5896, opened 2023-01-14
Add pdf support

johnshaughnessy:
Add support for loading pdfs with bitecs.

Also rework `coroutines` by introducing `JobRunner`.


src/utils/load-pdf.tsx
```diff
@@ -0,0 +1,15 @@
+/** @jsx createElementEntity */
+import * as pdfjs from "pdfjs-dist";
+import { HubsWorld } from "../app";
+import { createElementEntity, renderAsEntity } from "../utils/jsx-entity";
+
+export function* loadPDF(world: HubsWorld, url: string) {
+  const pdf = (yield pdfjs.getDocument(url).promise) as pdfjs.PDFDocumentProxy;
+
+  // TODO Should we create the canvas texture here and load/render the first page?
```
johnshaughnessy, line 9: In my opinion, we should not. Instead, we should just let the pdf system load whatever page needs to be loaded. If for some reason a page takes a long time to load, oh well. If a page throws an error while loading, we'll have to handle that somehow. (Currently, we don't.) Either way, it seems nice to keep the system simple by having only one place that pages are loaded rather than doing another page load here. Thoughts?

src/bit-systems/pdf-system.ts
```diff
@@ -0,0 +1,76 @@
...
+  material.map = component.texture;
+  material.map!.needsUpdate = true;
+  material.needsUpdate = true;
+  scaleMeshToAspectRatio(mesh, ratio);
+
+  // HACK The loading animation in media loader
+  // might still be controlling this object's scale,
+  // so wait for that long and then set the scale again.
+  yield sleep(mediaScaleAnimationDurationMS);
+  scaleMeshToAspectRatio(mesh, ratio);
```
johnshaughnessy, line 48: This is a tricky problem that will come up anywhere that we have animations (or other long-running processes) setting object properties. I don't know how to solve it. I don't mind shipping with this hacky workaround for now.

src/utils/networked-pdf-schema.ts
```diff
@@ -0,0 +1,31 @@
...
+      // TODO It would be nice to have a "strict mode" for development / testing
+      // that would elevate the severity of this to throw an error.
+      console.error("Failed to deserialize stored networked-pdf component data.", component);
+      return;
+    }
+
+    const { page }: { page: number } = component.data;
+    write(NetworkedPDF.page, eid, page);
+  }
+};
```
johnshaughnessy, line 31: In https://github.com/mozilla/hubs/pull/5896/commits/07e63899bffdf0acc10a1700027702a7e0effae7 I simplified `deserializeFromStorage` so that it didn't handle migrations from one `version` to another. This is the "simpler" version, but I don't like it better. 

My argument for writing all (storable) component schemas with "builtin" support/awareness of migrations/versions is:
- If all components that will be stored in the db have version numbers and support migrations the same way, then there's one way these storable components are ever written. Anyone who wants to store data in the database needs to learn this concept and conform to the spec, so we write a doc with an example to explain it. It's a little cumbersome everytime we want to store new component data, but at least we can copy/paste from any of the other component schemas that have only 1 version supported, so the marginal inconvenience is really only an extra 5 minutes converting old property names to new property names, fixing typos etc.
- If components only _sometimes_ have version numbers and handle migrations, and most components _don't need them_, then what I expect to happen is that the Foo component has a version _string_ and the Bar component handles versions in some bespoke way and the Baz component was written by a developer who hadn't encountered the idea at all and they are writing a post asking why their scenes with old Baz components is broken. In other words, it won't be obvious that we offered a solution to the problem or wanting to change your component schema (for a stored component), so everyone invented their own way to do it.

I would like to revert that commit to restore the "complicated" implementation.

src/bit-systems/pdf-menu-system.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  if (target) {
+    const numPages = (MediaPDF.map as PDFComponentMap).get(target)!.pdf.numPages;
+    (world.eid2obj.get(PDFMenu.pageLabelRef[menu]) as Text).text = `${NetworkedPDF.page[target]} / ${numPages}`;
+  }
+}
+
+const hoveredQuery = defineQuery([HoveredRemoteRight]);
+export function pdfMenuSystem(world: HubsWorld, sceneIsFrozen: boolean) {
+  const menu = anyEntityWith(world, PDFMenu)!;
```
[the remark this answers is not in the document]
netpro2k: For video menu we just made a query and grabbed the first element, which is basically the same as anyEntityWith. 

I definitely like the idea of doing it in a preload. The setInterval seems surprising. Ideally we should be able to do it synchronously. Maybe the functions you pass to prelaod get called after APP has been set up and passed a world. Some obviously will also be async after that, but this one shouldn't need to be. 

We could also just have a way to get the world with a promise which would remove the need to modify preload.

src/bit-systems/pdf-menu-system.ts
```diff
@@ -0,0 +1,99 @@
...
+
+function findPDFMenuTarget(world: HubsWorld, menu: EntityID, sceneIsFrozen: boolean) {
+  if (sceneIsFrozen) {
+    PDFMenu.targetRef[menu] = 0;
+    PDFMenu.clearTargetTimer[menu] = 0;
+    return;
+  }
+
+  const hovered = hoveredQuery(world);
+  const target = hovered.map(eid => findAncestorWithComponent(world, MediaPDF, eid))[0] || 0;
```
netpro2k, line 23: will just keep pointing out that I am very worried about using map/filter/etc in code called every frame. Wish I had a solid way of proving to myself that this isn't a problem...

src/bit-systems/pdf-menu-system.ts
```diff
@@ -0,0 +1,102 @@
...
+  if (target) {
+    PDFMenu.targetRef[menu] = target;
+    PDFMenu.clearTargetTimer[menu] = world.time.elapsed + 1000;
+    return;
+  }
+
+  if (PDFMenu.targetRef[menu] && !entityExists(world, PDFMenu.targetRef[menu])) {
+    // Invalid entity reference. (The pdf entity was removed).
+    // TODO Should this be handled in removeObject3DSystem instead?
```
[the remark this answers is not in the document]
netpro2k: I think here makes sense. I think of the removeObject3DSystem as cleaning up threejs resources. We have stretched that slightly in 1 or 2 spots to cover the weird `map` component pattern we have. Those might have better been handled in exit queries in their respective systems, similar to how we clean up "system state", like camera tool render targets. We definitely have lots of spots where this sort of stuff can be done and we should start to think about "rules", just for clarity.

src/bit-systems/pdf-menu-system.ts
```diff
@@ -0,0 +1,99 @@
...
+    PDFMenu.targetRef[menu] = 0;
+    PDFMenu.clearTargetTimer[menu] = 0;
+    return;
+  }
+
+  if (hovered.map(eid => findAncestorWithComponent(world, PDFMenu, eid))[0]) {
```
netpro2k, line 38: perf aside, `Array.prototype.some()` is probably more appropriate.

src/inflators/pdf-loader.ts
```diff
@@ -0,0 +1,16 @@
...
+export function inflatePDFLoader(world: HubsWorld, eid: number, params: PDFLoaderParams) {
+  inflateMediaLoader(world, eid, {
+    src: params.src,
+    recenter: false,
+    resize: false,
+    animateLoad: false,
+    isObjectMenuTarget: false
+  });
+
+  // TODO Should the PDF be controlled by users? (Should it be a PDFMenuTarget?)
```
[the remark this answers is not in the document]
netpro2k: Don't quite follow this TODO anyway

netpro2k (commented):
I think the core is pretty much exactly what you would expect to see, which is great, no big surprises. Made a bunch of little comments but most of them are just stylistic or trying to land on some form of pattern/rules for things that we can do in many ways. 

The main non-stylistic issue I saw was that we are definitely leaking some resources (though it seems we were sadly already doing this in the aframe version too). We definitely need better tools for auditing this sort of thing.

src/bit-systems/pdf-system.ts
johnshaughnessy: True... We could write `yield* waitForMediaLoaded(world, mesh.parent!.eid!)`, since the pdf's parent is the root media entity. And I guess it'll be good to reuse that function because later we can more easily say, "here are all the places that we are waiting for media to finish loading."  Fixed in https://github.com/mozilla/hubs/commit/31360a08d417cd1af5950417fc8f600030a84b4a

johnshaughnessy (commented):
Still working thru the comments...

src/bit-systems/pdf-system.ts
```diff
@@ -0,0 +1,76 @@
...
+  const page: pdfjs.PDFPageProxy = yield component.pdf.getPage(pageNumber);
+  const viewport = page.getViewport({ scale: 3 });
+  const ratio = viewport.height / viewport.width;
+  component.canvas.width = viewport.width;
+  component.canvas.height = viewport.height;
+  const renderTask = page.render({ canvasContext: component.canvasContext, viewport });
+  yield renderTask.promise;
+  const mesh = world.eid2obj.get(eid)! as Mesh;
+  const material = mesh.material as MeshBasicMaterial;
+  material.map = component.texture;
```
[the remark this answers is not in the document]
netpro2k: Ah yes, if you try to use a CanvasTexture that you have never rendered to yet I think that will be an issue. Another bonus point for making rendering *something* part of the initial "load"

src/bit-systems/pdf-system.ts
```diff
@@ -0,0 +1,85 @@
...
+const jobs = new JobRunner();
+const pdfQuery = defineQuery([MediaPDF, NetworkedPDF]);
+const pdfExitQuery = exitQuery(pdfQuery);
+export function pdfSystem(world: HubsWorld) {
+  pdfQuery(world).forEach(function (eid) {
+    if (MediaPDF.pageNumber[eid] !== NetworkedPDF.pageNumber[eid]) {
+      MediaPDF.pageNumber[eid] = NetworkedPDF.pageNumber[eid];
+
+      jobs.stop(eid);
+      jobs.add(eid, () => loadPageAndSetScale(world, eid, NetworkedPDF.pageNumber[eid]));
```
keianhzo, line 72: We talked about all of this in a 1:1  but I'm leaving it here too for discussion. 

We are explicitly ticking every job in each system. My understanding is that if we ticked all the jobs at the same time that would still guarantee the execution in the same frame point as if we were ticking them individually. Based on that assumption the API I imagine would be quite similar to this but hiding the `jobId` map inside the jobs runner and ticking all jobs from there called from the main tick. Something like:
```javascript
const jobId = APP.jobRunner.add((clear: ClearFunction, abort: AbortSignal) => myJob(clear, abort));
...
APP.jobRunner.cancel(jobId);
```
or:
```javascript
const myJob = (clear: ClearFunction, abort: AbortSignal) => myJob(clear, abort);
APP.jobRunner.add(myJob);
...
APP.jobRunner.cancel(myJob);
```

johnshaughnessy: > that would still guarantee the execution in the same frame point as if we were ticking them individually.

Not quite -- If we tick all jobs (registered by any system) at some point in the frame, that point is when they'll run (not at the point when they were registered). 

That's probably fine, but introduces some additional indirection we'd have to keep in mind.

netpro2k: Also, so far as we have used these we have almost always wanted to stop the job when the entity a job is "about" goes away. If job ids are assigned, we could make a separate map from entity to jobid, but that seems unnecessary.

src/utils/coroutine.js
```diff
@@ -139,13 +81,21 @@ export function coroutine(iter) {
...
+
       doThrow = false;
       timers = null;
       if (done) {
         return value;
       }
 
+      if (isCancelablePromise(value)) {
+        rollbacks.push(value.rollback);
+        value = value.promise;
```
netpro2k, line 96: Slightly weird that technically this could be a non-promise and would fail the check below. Typescript protects us from that, and I guess we end up throwing anway so its not a big deal, but just slightly odd how that works out.

src/utils/coroutine-utils.ts
```diff
@@ -0,0 +1,87 @@
...
+  rollback: RollbackFunction;
+  $rollback = $rollbackSymbol;
+  constructor(promise: Promise<T>, rollback: RollbackFunction) {
+    this.promise = promise;
+    this.rollback = rollback;
+  }
+}
+
+export function isCancelablePromise<T>(c: any): c is CancelablePromise<T> {
+  return c.$rollback === $rollbackSymbol;
```
netpro2k, line 82: It was late when we wrote this. I think what we wanted was
```ts
c[$rollbackSymbol] === true
```

with
```ts
static [$rollbackSymbol] = true;
```

What we have works, but the whole point of using a symbol for this sort of thing is that you shouldn't be able to get a reference to it externally. (this really doesn't matter)

src/bit-systems/media-loading.ts
```diff
@@ -44,51 +53,45 @@ function resizeAndRecenter(world: HubsWorld, media: EntityID, eid: EntityID) {
...
-  onAnimate([startPosition, startScale]);
+  onAnimate([startScale]);
   yield Promise.resolve();
 
   yield* animate({
-    properties: [
-      [startPosition, endPosition],
-      [startScale, endScale]
-    ],
+    properties: [[startScale, endScale]],
```
netpro2k, line 94: Why did this change?

src/inflators/pdf.ts
```diff
@@ -0,0 +1,33 @@
+import { addComponent } from "bitecs";
+import { PDFDocumentProxy } from "pdfjs-dist";
+import { Mesh, MeshBasicMaterial } from "three";
+import { HubsWorld } from "../app";
+import { MediaPDF, Networked, NetworkedPDF } from "../bit-components";
+import { PDFResourcesMap } from "../bit-systems/pdf-system";
```
netpro2k, line 6: I think we generally don't want to be leaking "system state" outside of the system like this. We were previously using `MediaPDF.map` for this (and we actually are still defining that). I think I like that better for this, since it IS component data, it just happens to contain references, so we have to store it a special way. (A good hint of that is that an inflator is writing to it).

I think of "system state" as incidental stuff a system needs to do its jobs, but is localized to that system.

src/bit-systems/pdf-system.ts
```diff
@@ -0,0 +1,85 @@
...
+    }
+  });
+
+  pdfExitQuery(world).forEach(function (eid) {
```
netpro2k, line 76: We aren't cleaning up `MediaPDF.map`, though it seems we aren't writing to it anymore either... See my note in the pdf inflator.

netpro2k (commented):
Already reviewed most of this, and paired on a good chunk of what was new, so only skimmed. Left a few comments. Mostly around use of system state and the `.map` on components.

src/inflators/pdf.ts
johnshaughnessy: It's not "system state" -- It's just wrapping `MediaPDF.map` with type info because it's better to write
`PDFResourcesMap.get(eid).pdf.numPages` than `((MediaPDF as any).map as Map<EntityID, PDFResources>).pdf.numPages`:

```ts
// TODO This can go away if we make bit-components becomes typescript file
export const PDFResourcesMap = (MediaPDF as any).map as Map<EntityID, PDFResources>;
```

netpro2k: Oh I missed this. That makes sense. Just a side effect of bit-components not being typescript yet.

src/bit-systems/media-loading.ts
netpro2k: I thought we had to to keep the object cetnered

johnshaughnessy: Hmm... That rings a bell... The `mediaObj` might have been `recenter`ed and have a non-zero offset applied (as its `position`). The offset was calculated based on the object being its "real size", which means we have to scale the offset as we scale the object. What we had before assumed uniform scale... So I guess it should be:
```diff 
- const startPosition = new Vector3().copy(mediaObj.position).multiplyScalar(startScale.x);
+ const startPosition = new Vector3().set(mediaObj.position.x * startScale.x, mediaObj.position.y * startScale.y, mediaObj.position.z * startScale.z);
```
Is that correct?

johnshaughnessy: Fixed in https://github.com/mozilla/hubs/commit/1ec67491c7aecf79cf4aba0f62b32148d7d91e1f

Good catch! Thanks!

src/utils/coroutine-utils.ts
johnshaughnessy: We tried this and it didn't work, so we changed it. I went back and tried some variations. This is what I've found to work:
```ts
  return c.__proto__.constructor[$rollbackSymbol] === true;
```
or (without `$rollbackSymbol` at all) 
```ts
c.__proto__ === CancelablePromise.prototype
```

Is either of these better? For the first, I probably also need to guard against `__proto__` or `constructor` being `undefined` since `c` is `any`.

src/utils/coroutine.js
johnshaughnessy: Yeah. `isCancelablePromise` doesn't actually check the `promise` part, does it...

src/utils/coroutine-utils.ts
```diff
@@ -51,17 +46,16 @@ export class JobRunner<T> {
...
     this.jobs.forEach((job, eid) => {
+      if (!job.coroutine) {
+        const clearRollbacks = () => {
+          job.rollbacks.length = 0;
+        };
+        job.coroutine = coroutine(job.startCallback(clearRollbacks, job.abortController.signal), job.rollbacks);
+      }
+
       if (job.coroutine!().done) {
-        this.jobs.delete(eid); // TODO Is this safe?
```
johnshaughnessy: I used this test to confirm that it was safe to delete from the map during iteration:
```
{
  const map = new Map([
    ["foo", 3],
    ["bar", 4],
    ["xxx", 5],
    ["yyy", 4],
    ["zzz", 9]
  ]);
  map.forEach((v, k) => {
    console.log(v, k);
    if (v === 4) map.delete(k);
  });
  console.log(map);
}
```

src/bit-systems/pdf-system.ts
keianhzo: Got it, all that makes sense.

PR #5901, opened 2023-01-19
Migrate billboard to bit-ecs

keianhzo:
This PR migrates and also make some choices that I'd like comments on:
- Adds a `inflatorWrapper` function to wrap a component inflator. I assume we will need to do this quite a lot for some components so it might be useful to have a convenience function for it
- Creates a single Billboard inflator that is reused in a few different places. In this specific case I'm are using the default inflator in a few places so it might make sense to reuse the same one. As we add more it might make sense to expose them in an object
- Exposes the main camera in App. I think we are accessing to this through the scene in a few places so I thought it might make sense to expose it here


src/bit-systems/billboard.ts
```diff
@@ -0,0 +1,93 @@
...
+    const object3D = world.eid2obj.get(billboard);
+    if (!object3D || !object3D.visible) {
+      inView.set(billboard, false);
+      return;
+    }
+
+    if (!APP.camera) return;
+
+    const inVR = APP.scene?.is("vr-mode");
+    inView.set(billboard, inVR ? true : isInViewOfCamera(object3D, APP.camera));
```
johnshaughnessy, line 53: Perhaps you should write this data to the Billboard component, rather than in a map.
```ts
if (isInView) {
  Billboard.flags[eid] |= BillboardFlags.IN_VIEW;
} else {
  Billboard.flags[eid] &= !BillboardFlags.IN_VIEW;
}
```

(Assuming that `BillboardFlags` is:
```ts
enum BillboardFlags {
  onlyY = 1 << 0,
  isInView = 1 << 1,
}
```

src/gltf-component-mappings.js
```diff
@@ -57,7 +60,7 @@ AFRAME.GLTFModelPlus.registerComponent("directional-light", "directional-light")
 AFRAME.GLTFModelPlus.registerComponent("hemisphere-light", "hemisphere-light");
 AFRAME.GLTFModelPlus.registerComponent("point-light", "point-light");
 AFRAME.GLTFModelPlus.registerComponent("spot-light", "spot-light");
-AFRAME.GLTFModelPlus.registerComponent("billboard", "billboard");
+AFRAME.GLTFModelPlus.registerComponent("billboard", "billboard", inflatorWrapper(billboardInflator));
```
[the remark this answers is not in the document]
netpro2k: No super strong opinion here solong as it doesn't add complexity to the system itself. (Ex: I would be weary if I saw some `hasComponent(world, AEntity, edi)` checks in the system). I agree that I might not have bothered, but deleting code is good.

src/bit-systems/billboard.ts
```diff
@@ -0,0 +1,93 @@
...
+import { HubsWorld } from "../app";
+import { Billboard } from "../bit-components";
+
+const billboardQuery = defineQuery([Billboard]);
+const billboardEnterQuery = enterQuery(billboardQuery);
+const billboardExitQuery = exitQuery(billboardQuery);
+
+const isMobileVR = AFRAME.utils.device.isMobileVR();
+
+const targetPos = new THREE.Vector3();
```
netpro2k, line 12: We are trying not to use the `THREE` global in new code, import the individual things you use instead.

netpro2k (commented):
Looking good. Made a few small comments. 

Biggest one would be getting rid of use of frame scheduler, I suspect we won't need that anymore going forward but maybe we will want some utility functions for doing something similar if we find ourselves writing the same code in many systems.

Also lets make sure we preserver that TODO, billboard as it is kind of stopped making sense since we now have many potential cameras in the scene. (Maybe it just needs to be renamed `face-player`. I would definitely like to also explore something more akin to how other engines handle billboarding and have it be more of a rendering concept, sitting at the same level as something like Mesh. That allows us to do the billboarding as part of the vertex shader.

src/app.ts
```diff
@@ -195,6 +196,7 @@ export class App {
     const audioListener = new AudioListener();
     this.audioListener = audioListener;
     camera.add(audioListener);
+    this.camera = camera;
```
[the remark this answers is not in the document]
keianhzo: It makes sense to pass the camera explicitly instead of having more stuff hanging from App.

src/bit-systems/billboard.ts
```diff
@@ -0,0 +1,93 @@
...
+    }
+  });
+  billboardQuery(world).forEach(billboard => {
+    const isInView = inView.get(billboard);
+    if (isInView || !isMobileVR) {
+      const object3D = world.eid2obj.get(billboard);
+      const onlyY = Billboard.onlyY[billboard];
+      if (!object3D || !object3D.visible) return;
+
+      const camera = APP.camera;
```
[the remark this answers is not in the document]
keianhzo: Yeah not sure what can be a better alternative but rendering with each camera can be pretty expensive specially if people are using many render targets for mirrors, video texture targets, photo cameras, etc. I think the rename can help to make it clear what it really is at least.

src/bit-systems/billboard.ts
```diff
@@ -0,0 +1,78 @@
...
+  frustum.setFromProjectionMatrix(frustumMatrix);
+  box.makeEmpty();
+  object3D.traverse(expandBox);
+
+  // NOTE: not using box.setFromObject here because text nodes do not have Z values in their geometry buffer,
+  // and that routine ultimately assumes they do.
+  return frustum.intersectsBox(box);
+};
+
+const updateIsInView = (world: HubsWorld, billboard: number, camera: Camera): boolean => {
```
johnshaughnessy, line 38: nit: Might call this something like `shouldUpdateBillboard` since it's not storing `inView` state in a map

src/utils/jsx-entity.ts
```diff
@@ -338,7 +340,8 @@ export const commonInflators: Required<{ [K in keyof ComponentData]: InflatorFn
   grabbable: inflateGrabbable,
 
   // inflators that create Object3Ds
-  directionalLight: inflateDirectionalLight
+  directionalLight: inflateDirectionalLight,
+  billboard: createDefaultInflator(Billboard)
```
johnshaughnessy, line 344: nit: There's a comment that separates inflators that do / don't create `Object3D`s. Not sure this comment is useful, but we should either remove the comment or make sure it's accurate (by moving `billboard` above it)

johnshaughnessy (approved):

src/bit-systems/billboard.ts
keianhzo: Oh yep, makes sense.

src/utils/jsx-entity.ts
keianhzo: I'll leave the comment for now and move the creator above.

PR #5903, opened 2023-01-20
Add spherical projections to videos / images

johnshaughnessy:
- Give `InitialData` a more specific type than `any`. (More can be done to tie a specific shape for each `PrefabName`, but I haven't done that here.) 
- Pipe projection information through media loader to support 360 videos / images loaded from gltf files. (Currently, copy/pasting an image will always result in it being flat. We can fix this separately if we figure out how to determine that an image should be projected spherically.)


src/systems/floaty-object-system.js
```diff
@@ -124,9 +124,9 @@ export const floatyObjectSystem = world => {
 
   networkedFloatyObjectsQuery(world).forEach(eid => {
     if (hasComponent(world, Owned, eid)) {
-      NetworkedFloatyObject.flags = FloatyObject.flags;
+      NetworkedFloatyObject.flags[eid] = FloatyObject.flags[eid];
```
netpro2k, line 127: This is a js file, but like we talked about, would be nice if we can modify the BitECS typescript defs to catch this sort of mistake. I think we have made it more than once.

src/utils/create-networked-entity.ts
```diff
@@ -25,12 +25,12 @@ export function renderAsNetworkedEntity(
   nid: NetworkID,
   creator: ClientID
 ) {
-  const eid = renderAsEntity(world, prefabs.get(prefabName)!.template(initialData));
+  const eid = renderAsEntity(world, prefabs.get(prefabName)!.template(initialData as any));
```
netpro2k, line 28: I think we looked into this already and the answer was no. But can't this be an `InitialData` instead of `any`?

src/utils/load-image.tsx
```diff
@@ -6,18 +6,22 @@ import { renderAsEntity } from "../utils/jsx-entity";
...
 import { Texture } from "three";
 import { AlphaMode } from "./create-image-mesh";
+import { MEDIA_LOADER_FLAGS } from "../bit-systems/media-loading";
 
-export function* loadImage(world: HubsWorld, url: string, contentType: string) {
+export function* loadImage(world: HubsWorld, flags: number, url: string, contentType: string) {
   const { texture, ratio, cacheKey }: { texture: Texture; ratio: number; cacheKey: string } =
     yield loadTextureCancellable(url, 1, contentType);
+  const projection =
+    flags & MEDIA_LOADER_FLAGS.SPHERICAL_PROJECTION ? ProjectionMode.SPHERE_EQUIRECTANGULAR : ProjectionMode.FLAT;
```
netpro2k, line 15: Feels weird that this gets a MEDIA_LOADER_FLAGS instead of a PROJECTION_MODE. Feels like the conversion between those should be at the thing calling `loadImage` this would also get rid of the confusing `0` being passed in (it would become `ProjectionMode.FLAT`.

netpro2k (commented):
Mostly lgtm. Made a few minor comments. Biggest thing that feels off to me is the MeidaLoader flags "leaking" outside of media loading. Its a pretty minor feeling though and I don't think it actually particularly matters much.

johnshaughnessy:
The base branch absorbed all but these two commits:
https://github.com/mozilla/hubs/pull/5903/commits/1c065cd27f4bd3fd16289693c72c1cf11824fe7c
https://github.com/mozilla/hubs/pull/5903/commits/a8e0376b05d8ddd6a3d81c4b0cc5d58e4c1655e7

PR #5906, opened 2023-01-21
adding SCSS and TS and components

nickgrato:
TODO


admin/.vscode/settings.json
```diff
@@ -0,0 +1,22 @@
+{
+  "workbench.colorCustomizations": {
+    "activityBar.activeBackground": "#ab307e",
+    "activityBar.background": "#ab307e",
+    "activityBar.foreground": "#e7e7e7",
+    "activityBar.inactiveForeground": "#e7e7e799",
+    "activityBarBadge.background": "#25320e",
```
nickgrato, line 7: i'll get ignore this.

admin/src/assets/locales/README.md
```diff
@@ -0,0 +1,253 @@
+# Localization
+
```
nickgrato, line 2: I just copied over the locales folder, maybe we delete this.

admin/src/react-components/lily.tsx
```diff
@@ -0,0 +1,56 @@
...
+  return false;
+}
+
+// export function getReticulumFetchUrl(path, absolute = false, host = null, port = null) {
+//   if (host || hasReticulumServer()) {
+//     return `https://${host || configs.RETICULUM_SERVER}${port ? `:${port}` : ""}${path}`;
+//   } else if (absolute) {
+//     resolverLink.href = path;
+//     return resolverLink.href;
+//   } else {
```
nickgrato, line 15: I won't check this in

admin/.prettierrc.json
```diff
@@ -0,0 +1,6 @@
+{
```
netpro2k, line 1: Admin project was being covered by Hubs prettierrc before. Fine to give it its own but we might not want to change the properties. since it will leed to A LOT of noise in the PR. If we really want to change it we should do just the formatting change in its own PR otherwise its pretty hard to look at the diff.

PR #5909, opened 2023-01-24
Bitecs simple-water

keianhzo:
This PR migrates the simple water object to bit-ecs. 

Initially I tried to get rid of the AFrame simple-water component to reuse the JSX inflator but I didn't like adding code to the common inflator to cover the AFrame case (as it's something to remove later) so in the end I have kept the previous path apart and I have just reused the refactored the common custom mesh code into a `SimpleWaterMesh` object.

### DIFF
--- a/src/bit-systems/simple-water.ts
+++ b/src/bit-systems/simple-water.ts
@@ -1,25 +1,11 @@
 import { defineQuery, exitQuery } from "bitecs";
 import { HubsWorld } from "../app";
 import { SimpleWater } from "../bit-components";
-import { SimpleWaterMesh } from "../inflators/simple-water";
-import { disposeMaterial } from "../utils/three-utils";
+import { SimpleWaterMesh } from "../objects/SimpleWaterMesh";
 
 const simpleWaterQuery = defineQuery([SimpleWater]);
-const exitSimpleWaterQuery = exitQuery(simpleWaterQuery);
 
 export function simpleWaterSystem(world: HubsWorld) {
-  exitSimpleWaterQuery(world).forEach(eid => {
-    const obj = world.eid2obj.get(eid) as SimpleWaterMesh;
-    if (obj) {
-      obj.geometry.dispose();
-      if (Array.isArray(obj.material)) {
-        obj.material.forEach(material => disposeMaterial(material));
-      } else {
-        disposeMaterial(obj.material);
-      }
-    }
-  });
-
   simpleWaterQuery(world).forEach(eid => {
     const simpleWater = world.eid2obj.get(eid) as SimpleWaterMesh;
     simpleWater.update(world.time.elapsed / 1000);
--- a/src/bit-systems/simple-water.ts
+++ b/src/bit-systems/simple-water.ts
@@ -1,0 +1,27 @@
+import { defineQuery, exitQuery } from "bitecs";
+import { HubsWorld } from "../app";
+import { SimpleWater } from "../bit-components";
+import { SimpleWaterMesh } from "../inflators/simple-water";
+import { disposeMaterial } from "../utils/three-utils";
+
+const simpleWaterQuery = defineQuery([SimpleWater]);
+const exitSimpleWaterQuery = exitQuery(simpleWaterQuery);
+
+export function simpleWaterSystem(world: HubsWorld) {
+  exitSimpleWaterQuery(world).forEach(eid => {
+    const obj = world.eid2obj.get(eid) as SimpleWaterMesh;
+    if (obj) {
+      obj.geometry.dispose();
+      if (Array.isArray(obj.material)) {
+        obj.material.forEach(material => disposeMaterial(material));
+      } else {
+        disposeMaterial(obj.material);
+      }
+    }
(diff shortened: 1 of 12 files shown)


src/systems/remove-object3D-system.js
```diff
@@ -62,6 +63,14 @@ const cleanupSkyboxes = cleanupObjOnExit(Skybox, obj => {
   disposeMaterial(obj.sky.material);
   obj.sky.geometry.dispose();
 });
+const cleanupSimpleWaters = cleanupObjOnExit(SimpleWater, obj => {
+  obj.geometry.dispose();
+  if (Array.isArray(obj.material)) {
+    obj.material.forEach(material => disposeMaterial(material));
+  } else {
+    disposeMaterial(obj.material);
+  }
```
johnshaughnessy, line 72: You may be able to replace this cleanup with `disposeNode`: https://github.com/mozilla/hubs/blob/eb5da30aa145133ce822d29928978850a75c0b1c/src/utils/three-utils.js#L36-L50

I'm not sure if a `SimpleWater` is a `THREE.Mesh`, but maybe `disposeNode` can handle that if necessary.

src/components/simple-water.js
```diff
@@ -279,12 +19,8 @@ AFRAME.registerComponent("simple-water", {
...
 
   init() {
-    if (!waterNormalMap) {
-      waterNormalMap = new HubsTextureLoader().load(waterNormalsUrl);
-    }
-
-    const usePhongShader = window.APP.store.state.preferences.materialQualitySetting !== "high";
-    this.water = new SimpleWater(waterNormalMap, undefined, usePhongShader);
+    const usePhongShader = APP.store.state.preferences.materialQualitySetting !== "high";
+    this.water = new SimpleWaterMesh({ lowQuality: usePhongShader });
```
johnshaughnessy, line 23: Does this entity need a `SimpleWater` (tag) component so that the (bit) system updates it?

johnshaughnessy (approved):
Nice work!

src/components/simple-water.js
keianhzo: For the AFrame component this is handled by the component tick as it was before. I didn't wanted to do any more refactoring in that component other than the common `SimpleWaterMesh` object that's the same for the new and the old component.

src/systems/remove-object3D-system.js
keianhzo: Oops thanks for pointing that, I just realized that I we are using a shared normal map for the water mesh so we should only dispose geo and material but not the textures and both `disposeMaterial` and `disposeNode` do that.

Also I updated the AFrame component to free resources upon removal, we were not doing that.

src/bit-systems/simple-water.ts
```diff
@@ -0,0 +1,13 @@
+import { defineQuery, exitQuery } from "bitecs";
```
takahirox, line 1: `exitQuery` seems to be unused.

src/objects/SimpleWaterMesh.ts
```diff
@@ -0,0 +1,301 @@
...
+  get opacity(): number {
+    return this.waterMaterial.opacity;
+  }
+
+  set opacity(value: number) {
+    this.waterMaterial.opacity = value;
+    if (value !== 1) {
+      this.waterMaterial.transparent = true;
+    }
+  }
```
takahirox, line 189: Is it intentional that `.transparent` won't be `false` even if `.opacity` is 1?

keianhzo: Yeah, that's true, it seems to me that we should always set...

takahirox: We may fix it or add a comment in another PR because the main purpose of this PR is migration, not fixing a bug.

johnshaughnessy: I don't know for sure but it does seem like this is supposed to say something like
```js
this.waterMaterial.transparent = value !== 1;
```
And please _do_ fix bugs like these as we migrate things because it's a great opportunity to make minor improvements.

takahirox: We do. And separating bug fix PR from migration PR is good to track the change later.

netpro2k: Yep, fixing bugs like this is absolutely something we should do as we migrate these. Also find opportunities to simplify by dropping features we wrote code for but never exposed (check spoke and blender addon), etc. Re separate PRs, use best judgment. I think its ok to combine fixing with migrating if the migration PR isn't very complicated. Just make sure to note the change in the PR notes though, so the reviewer knows it was intentionally changed.

takahirox: To clarify, I didn't say we shouldn't fix bugs when migrating. We should fix problems when we find. The first reason why I suggested to make another PR is we didn't get quick response from you two. There are many historical and unclear code in the old Hubs Client code that new devs couldn't determine whether intentional or not. I didn't think it's good to block this PR until we get the comment from you.

Issue #5899, opened 2023-01-18
BitECS Component Burndown


netpro2k:
## Simple Object3Ds
These will likely just be inflators that direclty read the properties, create an Object3D, and call `addObject3DComponent`. Depending on what they do they may or may not also add a tag component if they need to do any special cleanup outside of what Object3DTag is used for.
- [x] simple-water [keianhzo] #5909
- [x] mirror [Takahiro] #5919
- [x] text [Takahiro] #5933
 (JSX component done but might need some properties remapped, maybe think about TODOS around a system for calling `sync()` on all texts once per frame instead of in every system that uses them.)
- [x] particle-emitter [keianhzo] #5926
- [x] ambient-light [Takahiro] #5918
- [x] directional-light
- [x] hemisphere-light [Takahiro] #5934 
- [x] point-light [Takahiro] #5932
- [x] spot-light [Takahiro] #5931

## Media Loaders 
largely done, need some property mapping and link needs to be figured out and need to handle "controls" and other mediaOptions in existing loaders like image and video
- [x] model [Takahiro] #5953
JSX component is done, needs a "loader" version for GLTFs like we did for image/video.
- [x] audio [Takahiro] #5949
- [x] video [Takahiro] #5962 #5966 #5969
Mostly done. Needs support for HLS / DASH added in, and work done for owner selection for pinned videos.
- [x] link [Takahiro] #5976
- [x] html [Takahiro] #5977
- [x] pdf [John]

## Animation
- [x] loop-animation [Takahiro] #5938
(likely right time to do some form of animationMixerSystem, start to think about networking, though not implementing that part)
- [x] uv-scroll [Dom]
(see if we can deal with material components in a less hacky way now)

## Physics
Most of the current implementation of physics in BitECS now is hardcoded. Note: we can probably get pretty far with most environments by just re-implementing the auto shape generation we do for scenes without any physics components.
- [x] ammo-shape [keianhzo] https://github.com/mozilla/hubs/pull/5978
- [x] heightfield (likely maps to ammo-shape and becomes deprecated) [keianhzo] https://github.com/mozilla/hubs/pull/5978
- [x] trimesh (likely maps to ammo-shape and becomes deprecated) [keianhzo] https://github.com/mozilla/hubs/pull/5978
- [x] box-collier (likely maps to ammo-shape and becomes deprecated) [keianhzo] https://github.com/mozilla/hubs/pull/5978
[...]

PR #5911, opened 2023-01-25
Material Components (uv-scroll, video-texture-target, video-texture-source)

netpro2k:
This PR adds support for "material components". We allow adding hubs components to materials in GLTF files. So far we only used this for `video-texture-source` (and intended to go back and use it for `uv-scroll`). Previously we had no good way of implementing this, since components could only be on AFRAME entities, and entities could only be Object3Ds. This PR turns Materials into entities as well, so components can be added to them. 

I am still slightly uneasy about the concept of making materials entitles (and the number of entities this adds), but it feels nice to actually be able to implement this correctly. If we discover we really do not end up creating more material components then we can consider special casing these instead, similar to what we did before.

Part of implementing video-texture-source was factoring out the code for updating render targets used in the camera tool system. Ideally we should probably refactor that to actually use video-texture-target and video-texture-source components. The code in this function was also simplified a bit resolving some (but not all) of the TODOs.

This also incidentally fixes issues that were present resolving node references in components (these are the first components we have implemented that use them).

part of #5899


src/bit-components.js
```diff
@@ -197,3 +197,17 @@ export const ObjectSpawner = defineComponent({
...
+  speed: [Types.f32, 2],
+  increment: [Types.f32, 2],
+  offset: [Types.f32, 2]
+});
+export const VideoTextureSource = defineComponent({
+  fps: Types.ui8,
+  resolution: [Types.ui16, 2]
+});
+export const VideoTextureTarget = defineComponent({
+  srcNode: Types.eid,
```
johnshaughnessy, line 211: `srcNode` -> `source`? (May not be worth the trouble since gltf components say `srcNode` now.)

src/systems/remove-object3D-system.js
```diff
@@ -116,6 +124,7 @@ export function removeObject3DSystem(world) {
   cleanupSkyboxes(world);
 
   // Finally remove all the entities we just removed from the eid2obj map
-  entities.forEach(removeFromMap);
-  exitedObject3DQuery(world).forEach(removeFromMap);
+  entities.forEach(removeObjFromMap);
+  exitedObject3DQuery(world).forEach(removeObjFromMap);
+  exitedMaterialQuery(world).forEach(removeFromMatMap);
```
johnshaughnessy, line 129: (Checking my understanding...) 
It is OK not to call `o.material ? o.material.eid ? removeEntity(world, o.material.eid)` in the `traverseSome` above, because the various `cleanup` functions are going to call `disposeMaterial` as needed. We don't risk missing any material `eid`s. Is that correct?

src/bit-systems/uv-scroll.ts
```diff
@@ -0,0 +1,55 @@
...
+    removeComponent(world, UVScroll, eid);
+  });
+}
+
+const uvScrollQuery = defineQuery([UVScroll, MaterialTag]);
+export function uvScrollSystem(world: HubsWorld) {
+  migrateLegacyComponents(world);
+  uvScrollQuery(world).forEach(function (eid) {
+    const map = (world.eid2mat.get(eid)! as MeshBasicMaterial).map;
+    if (!map) return; // This would not exactly be expected to happen but is not a "bug" either. There is just no work to do in this case.
```
johnshaughnessy, line 43: I'm confused why this wouldn't be a bug. Is it valid to attach a UVScroll (material) component to something that isn't a material? Alternatively, is it valid to have a material `eid` that is not in `eid2mat`?

src/components/gltf-model-plus.js
```diff
@@ -757,18 +772,9 @@ export async function loadGLTF(src, contentType, onProgress, jsonPreprocessor) {
...
-            if (obj.material.specularMap) disposables.add(obj.material.specularMap);
-            if (obj.material.envMap) disposables.add(obj.material.envMap);
-            if (obj.material.aoMap) disposables.add(obj.material.aoMap);
-            if (obj.material.metalnessMap) disposables.add(obj.material.metalnessMap);
-            if (obj.material.roughnessMap) disposables.add(obj.material.roughnessMap);
-            if (obj.material.emissiveMap) disposables.add(obj.material.emissiveMap);
-          }
+          mapMaterials(obj, function (m) {
+            disposables.add(m);
+          });
```
johnshaughnessy, line 777: smart

src/components/gltf-model-plus.js
```diff
@@ -505,6 +505,21 @@ class GLTFHubsComponentsExtension {
...
+            typeof props[propName] === "number"
+          ) {
+            console.warn(
+              `Found an outdated ${componentName} 'srcNode' property, fixing. Make sure you are using the latest exporter.`
+            );
+            props[propName] = {
+              __mhc_link_type: "node",
+              index: props[propName]
+            };
+          }
```
johnshaughnessy, line 521: If traversing were free (which it's not), it'd be nice to run all the migrations we want to first and separately. If we have more of these over time, then at least we can consider extracting it to functions (like you did in the `uv-scroll` system).

src/inflators/model.tsx
```diff
@@ -1,30 +1,82 @@
...
+  Object.keys(components).forEach(name => {
+    const inflatorName = camelCase(name);
+    if (ignoredComponents.includes(inflatorName)) return;
+    if (!gltfInflatorExists(inflatorName)) {
+      console.warn(`Failed to inflate unknown component called ${inflatorName}`, components[name]);
+      return;
+    }
+
+    const props = components[name];
+    Object.keys(components[name]).forEach(propName => {
```
johnshaughnessy, line 33: `Object.entries(props).forEach(([propName, value]) => {` ?

src/inflators/model.tsx
```diff
@@ -1,30 +1,82 @@
...
+// These components are all handled in some special way, not through inflators
+const ignoredComponents = ["visible", "frustum", "frustrum", "shadow", "networked"];
+
+function inflateComponents(
+  world: HubsWorld,
+  eid: number,
+  components: { [componentName: string]: any },
+  idx2eid: Map<number, EntityID>
+) {
+  Object.keys(components).forEach(name => {
```
johnshaughnessy, line 24: `Object.entries(components).forEach(([name, props]) => {` ?

src/inflators/model.tsx
```diff
@@ -1,30 +1,82 @@
...
 function camelCase(s: string) {
   return s.replace(/-(\w)/g, (_, m) => m.toUpperCase());
 }
 
 export type ModelParams = { model: Object3D };
 
+// These components are all handled in some special way, not through inflators
+const ignoredComponents = ["visible", "frustum", "frustrum", "shadow", "networked"];
+
+function inflateComponents(
```
johnshaughnessy, line 18: Despite the trickiness/book-keeping for refs, I like how simple this function is: For each non-ignored component, resolve its refs and call its inflator.

src/inflators/model.tsx
```diff
@@ -1,30 +1,82 @@
...
+    } else {
+      eid = addEntity(world);
+    }
+
+    if (gltfIndex !== undefined) idx2eid.set(gltfIndex, eid);
+
+    const components = obj.userData.gltfExtensions?.MOZ_hubs_components;
+    if (components) inflateComponents(world, eid, components, idx2eid);
+
+    mapMaterials(obj, function (mat: Material) {
```
johnshaughnessy, line 73: Should this check if the object was replaced by an inflator?
`mapMaterials( world.eid2obj.get(eid) || obj, `

src/bit-systems/video-texture.ts
```diff
@@ -0,0 +1,173 @@
...
+      mat.emissiveMap = targetData.originalMap;
+    }
+    targetDataMap.delete(eid);
+  });
+
+  videoTextureSourceQuery(world).forEach(function (eid) {
+    const data = sourceDataMap.get(eid);
+    if (data && world.time.elapsed > data.lastUpdated + 1000 / VideoTextureSource.fps[eid]) {
+      updateRenderTarget(world, data.renderTarget, data.camera);
+      data.lastUpdated = world.time.elapsed;
```
johnshaughnessy, line 170: Why does video-texture.ts set `renderTarget.needsUpdate = false` after `updateRenderTarget` whereas that does not happen here?

src/bit-systems/video-texture.ts
```diff
@@ -0,0 +1,173 @@
...
+import { HubsWorld } from "../app";
+import { MaterialTag, VideoTextureSource, VideoTextureTarget } from "../bit-components";
+import { Layers } from "../camera-layers";
+import { VIDEO_TEXTURE_TARGET_FLAGS } from "../inflators/video-texture-target";
+import { EntityID } from "../utils/networking-types";
+import { findNode } from "../utils/three-utils";
+
+function noop() {}
+
+export function updateRenderTarget(world: HubsWorld, renderTarget: WebGLRenderTarget, camera: EntityID) {
```
johnshaughnessy, line 22: This function is tricky. (Not new)

johnshaughnessy (approved):
Nice work! Left some comments. I'll be curious if/when we ever bump into an invalidated assumption about entities always being associated with `object3D`s. Even if down the road we decide again material entities, it's encouraging to me to see that we can "just" have entities and component data and it's _fine_

src/app.ts
```diff
@@ -113,6 +115,7 @@ export class App {
     this.store = store;
     // TODO: Create accessor / update methods for these maps / set
     this.world.eid2obj = new Map();
+    this.world.eid2mat = new Map();
```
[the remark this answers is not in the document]
netpro2k: I think the way I would network sync a lot of the animation type stuff is just using serverTime instead. But your point stands it should be completely trivial to network stuff about material entities too, they are not special.

src/systems/remove-object3D-system.js
netpro2k: That's the idea yeah. I don't think it would be appropriate to remove a material entity just because an Mesh pointing at that material was removed. We specifically dispose of materials when we are done with them through various different means. All of them should always call `disposeMaterial`, which will call `removeEntity`, and then it will be removed from the map here.

src/bit-systems/uv-scroll.ts
netpro2k: No. It must be on a material, and the material will be in the map (hence the !). What is not guaranteed, and I don't think inherently a bug is that a material can choose to not have a map. In practice you would not do this in a GLTF file, but I could see cases where we would be UV scrolling something and changing its map. Maybe turning it off? Not sure. Maybe that is not a real usecase.

src/components/gltf-model-plus.js
netpro2k: Yeah, in making sure we were calling `disposeMaterial` here it became clear that we didn't need to be so verbose. Also it was actually a bug before because we weren't calling dispose on the material itself.

src/components/gltf-model-plus.js
netpro2k: Yeah I like the idea of moving all these sorts of migrations "up" into GLTF loading. Basically making the GLTF file we wish we had before we load it, that way ll of this logic happens up front and doesn't leak elsewhere. Main downside I see of that is that its then just a big clump of random things, rather than being "near" related stuff. I could also imagine some of the transformations will be slightly more tricky to do up front. But this doesn't seem worth trying to do until we get rid of gltf-mdoel-plus and gltf-component-mappings.

src/inflators/model.tsx
netpro2k: The materials in the GLTF file certainly would not be on an object created by an inflator. If `obj` is a `Mesh` (only thing that can have materials from the GLTFLoader) and swapped, that's a bug, and we throw an exception below.

src/bit-systems/video-texture.ts
netpro2k: That was just an additional property I sloppily stuck on `renderTarget` in the camera-tool system. This system does not use that, though the purpose of that in camera-tool system was to only update the render target when the material preview is visible, which we should actually do here too...

src/app.ts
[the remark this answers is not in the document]
johnshaughnessy: Network IDs would need to be assigned (deterministically) to materials in `setNetworkedData...`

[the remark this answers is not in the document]
netpro2k: Ah yeah that makes sense. That makes things a bit more tricky but still seems doable.

src/bit-systems/video-texture.ts
```diff
@@ -131,43 +177,33 @@ export function videoTextureSystem(world: HubsWorld) {
...
+        unbindMaterial(world, eid);
+        VideoTextureTarget.source[eid] = 0;
+      } else if (source !== targetDataMap.get(eid)!.boundTo) {
+        unbindMaterial(world, eid);
+        bindMaterial(world, eid);
+      }
+    } else if (source && entityExists(world, source)) {
+      bindMaterial(world, eid);
+    }
+  });
```
johnshaughnessy, line 199: What happens if the `source` entity goes away, and this `VideoTextureTarget.source[eid]` pointed to it? Do we need to check for that and set `VideoTextureTarget.source[eid] = 0`, is this handled already, or is this not possible / unnecessary?

netpro2k: This should be handled just above `if (!source || !entityExists(world, source)) {`

PR #5915, opened 2023-01-27
Add LightTag and call Three.js light.dispose when removed

takahirox:
Add `LightTag` component that must be set for all lights. Call Three.js `light.dispose()` when removed in `remove-object3D-system`.

This new tag and flow provides a proper light resource disposing.

Currently `light.dispose()` is called right after `DirectionalLight` creation in `inflator/directional-light` if `light.shadow.map` is non-null but it seems a bug because `light.shadow.map` should be null when light is created. This commit also fixes it by removing the lines. Lights will be properly disposed when removed with the change.

### DIFF
--- a/src/inflators/directional-light.ts
+++ b/src/inflators/directional-light.ts
@@ -1,6 +1,6 @@
 import { addComponent } from "bitecs";
 import { addObject3DComponent } from "../utils/jsx-entity";
-import { DirectionalLight } from "../bit-components";
+import { DirectionalLight, LightTag } from "../bit-components";
 import { DirectionalLight as DL } from "three";
 import { HubsWorld } from "../app";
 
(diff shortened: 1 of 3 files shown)


netpro2k (approved):
lgtm

src/inflators/directional-light.ts
```diff
@@ -24,12 +24,9 @@ export function inflateDirectionalLight(world: HubsWorld, eid: number, params: D
   light.shadow.bias = params.shadowBias;
   light.shadow.radius = params.shadowRadius;
   light.shadow.mapSize.set(params.shadowMapResolution[0], params.shadowMapResolution[1]);
-  if (light.shadow.map) {
-    light.shadow.map.dispose();
-    (light.shadow.map as any) = null; // TODO: Correct the Typescript definition in three. This /is/ nullable.
-  }
```
johnshaughnessy, line 30: Oops! Thanks for removing this

PR #5917, opened 2023-01-30
Set transparent properly in SimpleWaterMesh

takahirox:
Related: https://github.com/mozilla/hubs/pull/5909#pullrequestreview-1271996219

Currently opacity setter in `SimpleWaterMesh` updates `waterMaterial.transparent` only when opacity is not one but it should be always updated even opacity is one.


netpro2k (approved):
LGTM

PR #5918, opened 2023-01-30
Add AmbientLights to bitecs

takahirox:
This PR adds `AmbientLights` to bitecs in the same way as `DirectionalLight`.

### DIFF
--- a/src/bit-components.js
+++ b/src/bit-components.js
@@ -47,6 +47,7 @@ export const AEntity = defineComponent();
 export const Object3DTag = defineComponent();
 export const GLTFModel = defineComponent();
 export const LightTag = defineComponent();
+export const AmbientLightTag = defineComponent();
 export const DirectionalLight = defineComponent();
 export const CursorRaycastable = defineComponent();
 export const RemoteHoverTarget = defineComponent();
--- a/src/inflators/ambient-light.ts
+++ b/src/inflators/ambient-light.ts
@@ -1,0 +1,26 @@
+import { addComponent } from "bitecs";
+import { addObject3DComponent } from "../utils/jsx-entity";
+import { AmbientLightTag, LightTag } from "../bit-components";
+import { AmbientLight } from "three";
+import { HubsWorld } from "../app";
+
+export type AmbientLightParams = {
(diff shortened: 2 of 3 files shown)


src/bit-components.js
```diff
@@ -47,6 +47,7 @@ export const AEntity = defineComponent();
 export const Object3DTag = defineComponent();
 export const GLTFModel = defineComponent();
 export const LightTag = defineComponent();
+export const AmbientLight = defineComponent();
```
netpro2k, line 50: We might want to call this `AmbientLightTag` just to avoid ambiguity and having to rename when importing `AmbientLight` from Three.

src/inflators/ambient-light.ts
```diff
@@ -0,0 +1,21 @@
...
+import { AmbientLight, LightTag } from "../bit-components";
+import { AmbientLight as AL } from "three";
+import { HubsWorld } from "../app";
+
+export type AmbientLightParams = {
+  color: string;
+  intensity: number;
+};
+
+export function inflateAmbientLight(world: HubsWorld, eid: number, params: AmbientLightParams) {
```
netpro2k, line 12: In the aframe component we have default values. We probably want to preserve those. https://github.com/mozilla/hubs/blob/master/src/components/ambient-light.js#L3-L4

We can do this easily with `Object.assign()` like https://github.com/mozilla/hubs/blob/master/src/inflators/simple-water.ts#L34

netpro2k (commented):
Don't feel super strong since I want to deprecate light components, but we should probably carry forward the defaults. Otherwise, LGTM

takahirox:
> I want to deprecate light components

Deprecate light components and encourage glTF lights extension instead? If so, yes I absolutely agree.

src/bit-components.js
takahirox: Updated, thanks

src/inflators/ambient-light.ts
takahirox: Updated, thanks

hrithikwins (approved):

PR #5919, opened 2023-01-30
Add Mirror to bitecs

takahirox:
This PR adds `Mirror` to bitecs.

### DIFF
--- a/src/inflators/mirror.ts
+++ b/src/inflators/mirror.ts
@@ -1,0 +1,53 @@
+import { addComponent } from "bitecs";
+import { PlaneGeometry } from "three";
+import { Reflector } from "three/examples/jsm/objects/Reflector";
+import { addObject3DComponent } from "../utils/jsx-entity";
+import { Mirror } from "../bit-components";
+import { HubsWorld } from "../app";
+import { Layers } from "../camera-layers";
+
+const DEFAULT_MIRROR_GEOMETRY = new PlaneGeometry();
+// We may need to revisit the default texture size
+// because the full window resolution may be too huge.
+const DEFAULT_TEXTURE_WIDTH = window.innerWidth * window.devicePixelRatio;
+const DEFAULT_TEXTURE_HEIGHT = window.innerHeight * window.devicePixelRatio;
+
+export type MirrorParams = {
+  color?: string;
+};
+
+const DEFAULTS = {
+  color: '#7f7f7f'
+};
+
+export function inflateMirror(world: HubsWorld, eid: number, params: MirrorParams) {
+  params = Object.assign({}, DEFAULTS, params);
+  const geometry = DEFAULT_MIRROR_GEOMETRY;
+  const reflector = new Reflector(geometry, {
+    color: params.color,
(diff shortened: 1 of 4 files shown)


src/inflators/mirror.ts
```diff
@@ -0,0 +1,51 @@
...
+    };
+    originalOnBeforeRender(renderer, scene, camera, geometry, material, group);
+    renderer.render = originalRender;
+  };
+
+  addObject3DComponent(world, eid, reflector);
+  addComponent(world, Mirror, eid);
+  return eid;
+}
+
```
takahirox, line 51: Duplicated code with https://github.com/mozilla/hubs/blob/master/src/components/mirror.js I thought it would be ok because not so many codes and src/components/mirror.js may be removed when the new framework will land.

If we want to avoid duplicated code we can add a new file and let `src/inflators/mirror.ts` and `src/components/mirror.js` refer to it.

src/inflators/mirror.ts
```diff
@@ -0,0 +1,51 @@
...
+  color?: string;
+};
+
+const DEFAULTS = {
+  color: '#7f7f7f'
+};
+
+export function inflateMirror(world: HubsWorld, eid: number, params: MirrorParams) {
+  params = Object.assign({}, DEFAULTS, params);
+  const geometry = DEFAULT_MIRROR_GEOMETRY;
```
takahirox, line 23: In the old code, `this.el.object3DMap?.mesh?.geometry` is used if defined.
 https://github.com/mozilla/hubs/blob/40bbcc41c6b5c2eac2982b0a8bf2bbebd44db195/src/components/mirror.js#L24

In the new framework, is there a chance that `geometry` is already defined when inflating?

netpro2k: Since mirror is currently only exposed in Spoke, and Spoke doesn't do this, I don't think we need to support that right now. 

Just to think through what it would look like if we wanted to support this later): It would be slightly tricky since in infiltrators you can not access the Object3D for the entity you are inflating on (since inflators may be what creates the Object3D) so we would need to instead write a system that swaps the existing mesh with a Reflector using the same Geometry.

src/inflators/mirror.ts
```diff
@@ -0,0 +1,51 @@
...
+  const reflector = new Reflector(geometry, {
+    color: params.color,
+    textureWidth: DEFAULT_TEXTURE_WIDTH,
```
netpro2k, line 26: This is not new and we don't necessarily need to fix this right now, but I quite dislike that we make this so high resolution... Not sure what a more sane value would be.

src/inflators/mirror.ts
netpro2k: I think its ok, since the plan is to remove the old code once we remove the `newLaoder` flag and flip to using the new system by default.

netpro2k (approved):
Made some minor notes, but LGTM

src/inflators/mirror.ts
takahirox: Yes, probably lower resolution will be better. Added a comment for now.

PR #5920, opened 2023-01-31
Use material-utils where possible

takahirox:
There some codes in `src` like

```javascript
if (object.material) {
  if (Array.isArray(object.material)) {
    object.material.forEach(material => func(material));
  } else {
    func(object.material);
  }
}
```

`utils/material-utils` has `forEachMaterial` helper function that does the above. Other files can just import and use it.


netpro2k (approved):
LGTM. 

Related. I suspect as we convert to typescript we will find more spots where we are assuming material is a single object but can be an array.

takahirox:
> Related. I suspect as we convert to typescript we will find more spots where we are assuming material is a single object but can be an array.

Perhaps right. Let's digging into in another PR (at some point).

PR #5926, opened 2023-02-01
Bitecs particle emitter

keianhzo:
This PR adds support for the bit-ecs particle-system counterpart.

We also should move the particle system to a more modern GPU based particle system like https://github.com/fazeaction/three-gpu-particle-system/ but I have left that for a post bit-ecs migration step.

### DIFF
--- a/src/bit-systems/particle-emitter.ts
+++ b/src/bit-systems/particle-emitter.ts
@@ -1,0 +1,52 @@
+import { defineQuery, enterQuery } from "bitecs";
+import { ParticleEmitter } from "lib-hubs/packages/three-particle-emitter/lib/esm/index";
+
+import { HubsWorld } from "../app";
+import { ParticleEmitterTag } from "../bit-components";
+import { proxiedUrlFor } from "../utils/media-url-utils";
+import { resolveUrl, textureLoader } from "../utils/media-utils";
+
+const particleEmitterQuery = defineQuery([ParticleEmitterTag]);
+const particleEmitterEnterQuery = enterQuery(particleEmitterQuery);
+
+const setTexture = async (world: HubsWorld, eid: number) => {
+  let src = APP.sid2str.get(ParticleEmitterTag.src[eid]);
+
+  const result = await resolveUrl(src);
+  let canonicalUrl = result.origin;
+
+  // handle protocol relative urls
+  if (canonicalUrl.startsWith("//")) {
+    canonicalUrl = location.protocol + canonicalUrl;
+  }
+
+  // todo: we don't need to proxy for many things if the canonical URL has permissive CORS headers
+  src = proxiedUrlFor(canonicalUrl);
+
+  const texture = await textureLoader.loadAsync(src);
+
+  const particleEmitter: ParticleEmitter = world.eid2obj.get(eid)! as ParticleEmitter;
+  particleEmitter.material.uniforms.map.value = texture;
+  particleEmitter.visible = true;
+
+  ParticleEmitterTag.updateParticles[eid] = 1;
+};
+
+export function particleEmitterSystem(world: HubsWorld) {
+  particleEmitterEnterQuery(world).forEach(eid => {
+    setTexture(world, eid);
+  });
+  particleEmitterQuery(world).forEach(eid => {
+    const particleEmitter: ParticleEmitter = world.eid2obj.get(eid)! as ParticleEmitter;
+
+    if (ParticleEmitterTag.updateParticles[eid]) {
+      particleEmitter.updateParticles();
+      ParticleEmitterTag.updateParticles[eid] = 0;
+    }
+
+    if (particleEmitter.visible) {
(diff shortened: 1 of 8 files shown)


src/inflators/particle-emitter.ts
```diff
@@ -0,0 +1,89 @@
...
+  lifetimeRandomness: number;
+  particleCount: number;
+  startVelocity: [number, number];
+  endVelocity: [number, number, number];
+  velocityCurve: string;
+  angularVelocity: number;
+};
+
+const DEFAULTS = {
+  src: defaultSrcImage,
```
netpro2k, line 34: I wan to get it so that `inflateParticleEmitter` takes a texture instead of a `src` (similar to how `inflateImage` takes a texture instead of a src). Since the gltf-exposed component expects a src we made a seperate loader for `image` in gltf called `inflateImageLoader`, that ends up using a MediaLoader.

Not sure its worth trying to do that here, might make more sense to just try and design it that we when we make the GPUParticleEmitter component and deprecate this one.

types/lib-hubs.d.ts
```diff
@@ -0,0 +1,8 @@
+declare module "lib-hubs/packages/three-particle-emitter/lib/esm/index" {
```
netpro2k, line 1: Not worth spending very much time on. But lib-hubs is actually written in typescript and has types https://github.com/MozillaReality/lib-hubs/blob/master/packages/three-particle-emitter/lib/types/index.d.ts I wonder if we can get that to work somehow.

src/systems/remove-object3D-system.js
```diff
@@ -74,6 +75,9 @@ const cleanupSimpleWaters = cleanupObjOnExit(SimpleWater, obj => {
     obj.material.dispose();
   }
 });
+const cleanupParticleSystem = cleanupObjOnExit(ParticleEmitterTag, obj => {
```
netpro2k, line 78: technically don't need the wrapping function here.

src/bit-systems/particle-emitter.ts
```diff
@@ -0,0 +1,61 @@
...
+}
+
+const particleEmitterQuery = defineQuery([ParticleEmitterTag]);
+const particleEmitterEnterQuery = enterQuery(particleEmitterQuery);
+const particleEmitterExitQuery = exitQuery(particleEmitterQuery);
+
+const jobs = new JobRunner();
+export function particleEmitterSystem(world: HubsWorld) {
+  particleEmitterEnterQuery(world).forEach(eid => {
+    jobs.stop(eid);
```
netpro2k, line 41: We only ever add job for an entity in the enter query, and always stop it in the exit query so I don't think its possible for there to be a job for this entity already. At first reading this was confusing because I went looking for a way this might happen.

src/bit-systems/particle-emitter.ts
```diff
@@ -0,0 +1,61 @@
...
+    jobs.add(eid, () => setTexture(world, eid));
+  });
+  particleEmitterExitQuery(world).forEach(function (eid) {
+    // Object3D resources release happens in remove-object3D-system
+    jobs.stop(eid);
+  });
+  particleEmitterQuery(world).forEach(eid => {
+    const particleEmitter: ParticleEmitter = world.eid2obj.get(eid)! as ParticleEmitter;
+
+    if (ParticleEmitterTag.updateParticles[eid]) {
```
netpro2k, line 51: Seems like this flag only gets flipped on in `setTexture` so we could get rid of this property and just call `particleEmitter.updateParticles();` directly in the job.

src/bit-systems/particle-emitter.ts
```diff
@@ -0,0 +1,61 @@
...
+      particleEmitter.updateParticles();
+      ParticleEmitterTag.updateParticles[eid] = 0;
+    }
+
+    if (particleEmitter.visible) {
+      particleEmitter.update(world.time.delta / 1000);
+    }
+  });
+  jobs.tick();
```
netpro2k, line 60: Fairly minor, but because this is run after the block above, it means we won't update the particle systems on the frame the job finishes.

netpro2k (commented):
Left some notes but mostly LGTM. I definitely want to revisit the "src" / texture loading stuff when we get around to a second pass on this.

types/lib-hubs.d.ts
keianhzo: I'd have used those types as is but we are not exposing the material property there so I just preferred to add the d.ts file that updating that in the external module (considering also that we will eventually move away from this implementation)

src/inflators/particle-emitter.ts
keianhzo: Yeah, that makes sense. It will also make sense in the future to allow direct texture references in glTF components like `image` or `particle-emitter` similarly to what we do for other properties like `envMap`. Probably even allow both paths, URLs and texture refs.

I'm going to leave it as it is for now as I'll like to do a pass after the migration an add support for a more modern particle system.

PR #5927, opened 2023-02-01
Add button to remove bit-ecs entities

keianhzo:
I find the ecsDebug panel pretty useful while debugging components, I just missed a way to easily remove a bit-ecs entity to test exit cases. This PRs adds such a button to the object properties panel.

### DIFF
--- a/src/react-components/debug-panel/ECSSidebar.js
+++ b/src/react-components/debug-panel/ECSSidebar.js
@@ -7,7 +7,7 @@ import { IconButton } from "../input/IconButton";
 import { FormattedMessage } from "react-intl";
 
 import * as bitComponents from "../../bit-components";
-import { defineQuery, getEntityComponents } from "bitecs";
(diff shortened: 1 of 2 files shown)


netpro2k (approved):
This seems like a useful debug tool. Note that the user-facing way to delete a spawned entity (something you paste in or upload) is hovering over it and pushing the X key. Obviously we need to make that more discoverable (probably y implementing the handler for the button in the object menu)

hrithikwins (approved):

johnshaughnessy:
RE "Implement the handler for the button in the object menu"

I've done this in another commit (unfortunately part of an unrelated PR): https://github.com/mozilla/hubs/pull/5904/commits/02b9a4d879cf203c4cbc7d8563b4ce2e60d91643

keianhzo:
I wasn't aware of that shortcut in the new loader. It seems to work for spawned nodes from the media toolbar but not from the chat. For testing the new bit-ecs components I was using mostly the chat command that's why this remove button.

PR #5931, opened 2023-02-06
Add SpotLight to bitecs

takahirox:
This PR adds `SpotLight` to bitecs, similar to #5918.

### DIFF
--- a/src/inflators/spot-light.ts
+++ b/src/inflators/spot-light.ts
@@ -1,0 +1,53 @@
+import { addComponent } from "bitecs";
+import { addObject3DComponent } from "../utils/jsx-entity";
+import { SpotLightTag, LightTag } from "../bit-components";
+import { SpotLight } from "three";
+import { HubsWorld } from "../app";
+
+export type SpotLightParams = {
+  color: string;
+  intensity: number;
+  range: number;
+  decay: number;
+  innerConeAngle: number;
+  outerConeAngle: number;
+  castShadow: boolean;
+  shadowMapResolution: [number, number];
+  shadowBias: number;
+  shadowRadius: number;
+};
+
+const DEFAULTS = {
+  intensity: 1.0,
+  range: 0,
+  decay: 2.0,
+  innerConeAngle: 0,
+  outerConeAngle: Math.PI / 4.0,
+  castShadow: true,
+  shadowMapResolution: [512, 512],
+  shadowBias: 0,
+  shadowRadius: 1.0
+};
+
+export function inflateSpotLight(world: HubsWorld, eid: number, params: SpotLightParams) {
+  params = Object.assign({}, DEFAULTS, params);
+  const light = new SpotLight();
+  light.position.set(0, 0, 0);
+  light.target.position.set(0, 0, 1);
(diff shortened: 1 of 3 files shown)


johnshaughnessy (approved):

src/inflators/spot-light.ts
```diff
@@ -0,0 +1,53 @@
...
+  decay: number;
+  innerConeAngle: number;
+  outerConeAngle: number;
+  castShadow: boolean;
+  shadowMapResolution: [number, number];
+  shadowBias: number;
+  shadowRadius: number;
+};
+
+const DEFAULTS = {
```
netpro2k, line 20: Not sure if I like it better but I notice we have to duplicate a lot of information with these defaults... 

We could use typeof: [example](https://www.typescriptlang.org/play?#code/MYewdgzgLgBAIgUQGIEECqAZAKgZRgXhgG8BYAKBhgEswoBTSKqATwC4YBGAOgAYAacpQBOAQzABzOu36CYAEzrARbGACZeAitTBg6QgMLg6KCQBspMGVpABXegaMnx59gFkRUABZcACgEkYAHoYABYNWSVoHE8RORAAd3YoIRs6TUoIGLj49wAHACU6CBBTOypwdgBtAFYOVT4YWtUAXRgRCBhK+Ko5L1YwGwBbACM9Bs86KnFPKH6h0aFm9JhM2ISAISp26WXV7PzYqhsIdm4ecgBfAG5ycjoAD1yQIVgWXLoYHCeoDCmZnxEokGHUIbzoIAAZvBkOhsDgbmQgA)

src/inflators/spot-light.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export type SpotLightParams = {
+  color: string;
+  intensity: number;
+  range: number;
+  decay: number;
+  innerConeAngle: number;
+  outerConeAngle: number;
+  castShadow: boolean;
+  shadowMapResolution: [number, number];
```
netpro2k, line 15: Not particularly important, but you can name the elements of a "tuple". Gives you slightly nicer type hinting:

```ts
  shadowMapResolution: [width: number, height: number];
```

src/inflators/spot-light.ts
```diff
@@ -0,0 +1,53 @@
...
+  decay: 2.0,
+  innerConeAngle: 0,
+  outerConeAngle: Math.PI / 4.0,
+  castShadow: true,
+  shadowMapResolution: [512, 512],
+  shadowBias: 0,
+  shadowRadius: 1.0
+};
+
+export function inflateSpotLight(world: HubsWorld, eid: number, params: SpotLightParams) {
```
netpro2k, line 32: This was discussed in anotehr PR, so whatever we decide there should apply here too... Since we provide default values the properties passed to the inflator (and exposed to JSX) should be optional, but then the type after applying defaults should have all fields be required. We can do that with typescript Partial/Required utility types.

netpro2k (commented):
LGTM. Some minor thoughts around typescript stuff but I wouldn't even necessarily change them.

PR #5932, opened 2023-02-06
Add PointLight to bitecs

takahirox:
This PR adds `PointLight` to bitecs, similar to #5918.


johnshaughnessy (approved):

netpro2k (approved):
LGTM. Same thoughts on typing as the other similar inflator PRs. We should just figure out what we want to do and apply it to all of them. We can do it all at once in a separate PR too, its not particularly important.

PR #5933, opened 2023-02-07
Polish text bitecs

takahirox:
This commit pilishes text bitecs. The changes are

* Rename Text component to TextTag to avoid the confliction with the WebAPI Text
* Rewrite inflators/text from JavaScript to TypeScript
* Explicitly declare text inflator params
* Add Text system that calls sync() and remove sync() calls from other systems
* Update type/troika-three-text.d.ts to match the troika-three-text API
* Fix some bugs


src/inflators/text.ts
```diff
@@ -0,0 +1,120 @@
...
+  depthOffset?: number;
+  direction?: "auto" | "ltr" | "trl";
+  fillOpacity?: number;
+  fontUrl?: string | null;
+  fontSize?: number;
+  letterSpacing?: number;
+  lineHeight?: number | "normal";
+  maxWidth?: number;
+  opacity?: number;
+  outlineBlur?: number | `${number}%`;
```
johnshaughnessy, line 24: I've never seen this type definition before. What does `${number}%` mean?

johnshaughnessy (approved):
Nice work! Thanks for adding all of the type infos

src/inflators/text.ts
takahirox: It might be similar to regex. (If I understand correctly) `${number}%` means number string that ends with "%", like "25%" or "0.5%".

https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html#handbook-content
https://stackoverflow.com/questions/51445767/how-to-define-a-regex-matched-string-type-in-typescript

johnshaughnessy: TIL. Thanks!

johnshaughnessy (approved):

src/inflators/text.ts
```diff
@@ -0,0 +1,127 @@
...
+  text.outlineOffsetY = params.outlineOffsetY!;
+  text.outlineOpacity = params.outlineOpacity!;
+  text.outlineWidth = params.outlineWidth!;
+  text.overflowWrap = params.overflowWrap!;
+  text.strokeColor = params.strokeColor!;
+  text.strokeOpacity = params.strokeOpacity!;
+  text.strokeWidth = params.strokeWidth!;
+  text.textAlign = params.textAlign!;
+  text.textIndent = params.textIndent!;
+  text.whiteSpace = params.whiteSpace!;
```
johnshaughnessy, line 121: As we discussed, we might be able to get rid of the `!`s by defining two types:

`OptionalTextParams` where each of the properties is optional and `TextParams` where each property is required.

The function signature becomes:
```
export function inflateText(world: HubsWorld, eid: number, params: OptionalTextParams) {
```

The defaults become

```
const DEFAULTS: TextParams = {
```

And in the body of the inflator we write:
```
const finalParams : TextParams = Object.assign({}, DEFAULTS, params)
```

takahirox: Unfortunately

```
ERROR in ./src/inflators/text.ts:122:9
TS2322: Type 'TextParams' is not assignable to type 'TextRequiredParams'.
  Types of property 'anchorX' are incompatible.
    Type '"left" | "center" | "right" | undefined' is not assignable to type '"left" | "center" | "right"'.
      Type 'undefined' is not assignable to type '"left" | "center" | "right"'.
    120 | export function inflateText(world: HubsWorld, eid: number, params: TextParams) {
    121 |   params = Object.assign({}, DEFAULTS, params);
  > 122 |   const params2: TextRequiredParams = Object.assign({}, DEFAULTS, params);
        |         ^^^^^^^
    123 |   const text = new TroikaText();
    124 |   text.material!.toneMapped = false;
    125 |
```

takahirox: I experimented. It works (type check passes).

But some disadvantages.

1. Many duplicated type declaration
2. We need to carefully manage that type declarations are synched. Even if types are different between them (ex: string / number) type check didn't catch.

So, I would like to suggest to experiment more and make another PR for it.

src/inflators/text.ts
netpro2k: Didn't know that was supported either. Neat!

src/inflators/text.ts
netpro2k: Typescript has a generic utility type to make all the fields of a type optional (or required) https://www.typescriptlang.org/docs/handbook/utility-types.html#partialtype - Worth looking through the other types in that doc too just to know what tools are available.

So we could either do:
```ts
export function inflateText(world: HubsWorld, eid: number, params: Partial<TextParams>) {
```
or
```ts
const finalParams = Object.assign({}, DEFAULTS, params) as Required<TextParams>;
```

(I think the `Partial` one is more natural)

src/bit-systems/text.ts
```diff
@@ -0,0 +1,31 @@
...
+
+    // sync() invokes async text processing in workers.
+    // https://github.com/protectwise/troika/tree/main/packages/troika-three-text#handling-asynchronous-updates
+    //
+    // It is safe to call sync() every frame from the
+    // performance and efficiency perspective because
+    // sync() checks whether to invoke costly processing
+    // inside.
+    //
+    // sync() call can happen one frame after that text
```
netpro2k, line 20: We choose when to run this system, I think we should just enforce this guarantee.

src/bit-systems/text.ts
```diff
@@ -0,0 +1,31 @@
...
+    // properties are updated by some other systems unless
+    // this system is guaranteed that it runs after all
+    // other systems in the frame. But it may not be
+    // a big deal because what sync() invokes is async
+    // processing.
+    //
+    // Question: Is it safe even if text object is
```
netpro2k, line 27: This query won't match on removed entities, and the text wont be disposed of until the `removeObejct3dSystem` runs (or in this system once we refactor where we do cleanup). So it should always be safe to call this. I assume TriokaText takes care of disposing of any in flight rendering when you dispose of a text.

src/inflators/text.ts
```diff
@@ -0,0 +1,129 @@
...
+import { Text as TroikaText } from "troika-three-text";
+import { HubsWorld } from "../app";
+import { TextTag } from "../bit-components";
+import { addObject3DComponent } from "../utils/jsx-entity";
+
+export type TextParams = {
+  value: string;
+  anchorX?: "left" | "center" | "right";
+  anchorY?: "top" | "top-baseline" | "top-cap" | "top-ex" | "middle" | "bottom-baseline" | "bottom";
+  clipRect?: [number, number, number, number] | null;
```
netpro2k, line 12: does this need to support both null and undefined?

src/inflators/text.ts
```diff
@@ -0,0 +1,129 @@
...
+
+const THREE_SIDES = {
+  front: FrontSide,
+  back: BackSide,
+  double: DoubleSide
+};
+
+// Defaults values must be set for all the optional properties
+// to safely use foo! operator in the inflator.
+const DEFAULTS: TextParams = {
```
netpro2k, line 51: What I like about using Partial/Required as discussed below is this would be type checked for us so we couldn't accidentally miss a default value.

src/systems/hubs-systems.ts
```diff
@@ -245,7 +246,8 @@ export function mainTick(xrFrame: XRFrame, renderer: WebGLRenderer, scene: Scene
   hubsSystems.gainSystem.tick();
   hubsSystems.nameTagSystem.tick();
   simpleWaterSystem(world);
-
+  textSystem(world);
```
netpro2k, line 249: Its very crude but we may want to add a blank line above this system to make it clear its a bit "special". Or maybe better to just add a comment like `// all systems that update text properties should run before this`

netpro2k (commented):
LGTM. Thanks for cleaning that up!

src/inflators/text.ts
takahirox: Doesn't "?" imply that accepts `undeinfed`?

PR #5934, opened 2023-02-08
Add HemisphereLight to bitecs

takahirox:
This PR adds `HemisphereLight` to bitecs, similar to #5918.

### DIFF
--- a/src/inflators/hemisphere-light.ts
+++ b/src/inflators/hemisphere-light.ts
@@ -1,0 +1,31 @@
+import { addComponent } from "bitecs";
+import { addObject3DComponent } from "../utils/jsx-entity";
+import { HemisphereLightTag, LightTag } from "../bit-components";
+import { HemisphereLight } from "three";
+import { HubsWorld } from "../app";
+
+export type HemisphereLightParams = {
+  skyColor?: string;
+  groundColor?: string;
+  intensity?: number;
+};
+
+const DEFAULTS: Required<HemisphereLightParams> = {
+  skyColor: "#ffffff",
+  groundColor: "#ffffff",
+  intensity: 1.0
+};
+
+export function inflateHemisphereLight(world: HubsWorld, eid: number, params: HemisphereLightParams) {
+  const requiredParams = Object.assign({}, DEFAULTS, params) as Required<HemisphereLightParams>;
+  const light = new HemisphereLight();
+  light.position.set(0, 0, 0);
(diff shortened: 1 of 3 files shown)


src/inflators/hemisphere-light.ts
```diff
@@ -0,0 +1,29 @@
...
+const DEFAULTS = {
+  intensity: 1.0
+};
+
+export function inflateHemisphereLight(world: HubsWorld, eid: number, params: HemisphereLightParams) {
+  params = Object.assign({}, DEFAULTS, params);
+  const light = new HemisphereLight();
+  light.position.set(0, 0, 0);
+  light.color.set(params.skyColor).convertSRGBToLinear();
+  light.groundColor.set(params.groundColor).convertSRGBToLinear();
```
johnshaughnessy, line 22: How do we decide which parameters get defaults and which do not? 

If a user tries to load an invalid asset (e.g. one that does not include a `skyColor` in `params`), I am OK with either of these options:
- Inflating fails and throws an error (possibly crashing the animation loop, if the error is not caught).
- Default values are applied and an error or warning is simply logged to the console.

I think in the code above, we will throw an error because we will pass `undefined` into `light.color.set`. Again, that is OK. Just want to check / discuss how you think we should decide what to do in these cases.

johnshaughnessy (approved):
Left a comment to ask your opinions of component defaults.

src/inflators/hemisphere-light.ts
takahirox: I don't get one thing. Can `skyColor` and `groundColor` be `undefined` although they are not optional?

(By the way, the reason why I define default values only for intensity is I just followed the old hemisphere-light component because I though there is historical reasons.)

netpro2k: Type information is only present at compile time, there is no runtime enforcement. It would be a type error at compile time to manually call `inflateHemisphereLight` with undefined values for required properties, and similarly it would be a compile error to write JSX without the required properties, but GLTF files would not have any type checks applied to them since they end up calling the inflator dynamically.

That said, if some property is required, and the GLTF file somehow does not have it, that component data is invalid and the asset should probably fail to load and show an error. I think its correct for the inflator to assume its being passed valid data. If we want to add validation we should do it upstream from the inflator.

netpro2k (approved):
LGTM

PR #5938, opened 2023-02-10
Add Mixer Animation to bitecs

takahirox:
This PR adds animation handled with `AnimationMixer` to bitecs.

The changes are
* Add `MixerAnimatable` component which should be added to an entity whose `Object3D` has `AnimationClips` in `object.animations` and manages `AnimationMixer` for the `Object3D`.
* Add `MixerAnimatableInitialize` component which triggers `MixerAnimatable` component initialization that will be done in a system. `AnimationMixer` needs to be created with `Object3D` so it needs to be initialized after `Object3D` is initialized.
* Add `LoopAnimation` component which manages `AnimationActions` composed from `AnimationClips` assigned to `Object3D`.
* Add `LoopAnimationInitialize` component which triggers `LoopAnimation` component initialization that will be done in a system. `AnimationActions` needs to be created from `AnimcationClips` with `AnimationMixer` so it needs to be initialized after `MixerAnimatable` is initialized.
* Add `mixerAnimatableSystem` which initializes and cleans up `MixerAnimatable` component and calls `AnimationMixer.update()` for animatable entities every tick.
* Add `loopAnimationSystem` which initializes and cleans up `LoopAnimation` component.
* Add `MixerAnimatable` and `LoopAnimation` inflators that set up `MixerAnimatableInitialize` and `LoopAnimationInitialize`. (`MixerAnimatable inflator` is a default inflator.)
* Register the inflators to JSX inflators
* Add a special path in model inflator for inflating `MixerAnimatableInitialize` and `LoopAnimationInitialize` components because Hubs `animation-mixer` and `loop-animation` components don't greatly fit to the bitECS `MixerAnimatable` and `LoopAnimation` components and systems.

Note that this PR attempts to move the current functionality and capability of the old `animation-mixer` and `loop-animation` components. Advanced features (eg: Network animation) should be in another commit.

### DIFF
--- a/src/bit-components.js
+++ b/src/bit-components.js
@@ -178,7 +178,28 @@ MediaPDF.map = new Map();
 export const MediaVideo = defineComponent({
   autoPlay: Types.ui8
 });
-export const AnimationMixer = defineComponent();
+export const MixerAnimatableInitialize = defineComponent({});
+export const MixerAnimatable = defineComponent({});
+/**
+ * @type {Map<EntityId, AnimationMixer}>}
+ */
+export const MixerAnimatableData = new Map();
+export const LoopAnimationInitialize = defineComponent({});
+/**
+ * @type {Map<EntityId, {
+ *          activeClipIndices: number[],
+ *          clip: number,
+ *          paused: boolean,
+ *          startOffset: number,
+ *          timeScale: number
+ *        }[]>}
+ */
+export const LoopAnimationInitializeData = new Map();
+export const LoopAnimation = defineComponent();
+/**
+ * @type {Map<EntityId, AnimationAction[]>}
+ */
+export const LoopAnimationData = new Map();
 export const NetworkedVideo = defineComponent({
   time: Types.f32,
   flags: Types.ui8
--- a/src/bit-systems/loop-animation.ts
+++ b/src/bit-systems/loop-animation.ts
@@ -1,0 +1,105 @@
+import { addComponent, defineQuery, enterQuery, exitQuery, hasComponent, removeComponent } from "bitecs";
+import { AnimationAction, AnimationClip, AnimationMixer, LoopRepeat } from "three";
+import {
+  MixerAnimatable,
+  MixerAnimatableData,
+  LoopAnimation,
+  LoopAnimationData,
+  LoopAnimationInitialize,
+  LoopAnimationInitializeData,
+  Object3DTag
+} from "../bit-components";
+import { HubsWorld } from "../app";
+
+const loopAnimationInitializeQuery = defineQuery([LoopAnimationInitialize, MixerAnimatable, Object3DTag]);
+const loopAnimationInitializeEnterQuery = enterQuery(loopAnimationInitializeQuery);
+
+const loopAnimationQuery = defineQuery([LoopAnimation, MixerAnimatable, Object3DTag]);
+const loopAnimationExitQuery = exitQuery(loopAnimationQuery);
+
+// Question: Who should have AnimationMixer?
+//           MixerAnimatable component or Scene/App?
+
+const getActiveClips = (
+  animations: Array<AnimationClip>,
+  activeClipIndices: number[],
+  clip: string
+): AnimationClip[] => {
+  if (clip !== "") {
+    const activeClips = [];
+    const clipNames = clip.split(",");
+    for (let i = 0; i < clipNames.length; i++) {
+      const clipName = clipNames[i];
+      const foundClip = animations.find((clip: AnimationClip) => {
+        return clip.name === clipName;
+      });
+      if (foundClip) {
+        activeClips.push(foundClip);
+      } else {
+        console.warn(`Could not find animation names '${clipName}' in `, animations);
+      }
+    }
+    return activeClips;
+  } else {
+    return activeClipIndices.map((index: number) => animations[index]);
+  }
+};
+
+export function loopAnimationSystem(world: HubsWorld): void {
+  loopAnimationInitializeEnterQuery(world).forEach((eid: number): void => {
+    const object = world.eid2obj.get(eid)!;
+    const mixer = MixerAnimatableData.get(eid)!;
+
+    addComponent(world, LoopAnimation, eid);
+
+    const params = LoopAnimationInitializeData.get(eid)!;
+    const activeAnimations = [];
+
+    for (let i = 0; i < params.length; i++) {
+      const p = params[i];
+      // TODO: Consider where animations are stored.
+      //       Currently they are stored in Object3D but
+      //       it isn't aligned with the Three.js official API.
+      //       Move to Component?
+      const activeClips = getActiveClips(object.animations, p.activeClipIndices, APP.getString(p.clip)!);
+
+      for (let j = 0; j < activeClips.length; j++) {
+        const clip = activeClips[j];
+
+        // Ignore if activeClipIndex is out of range from the animations
+        if (!clip) {
+          continue;
+        }
+
+        const action = mixer.clipAction(activeClips[j], object);
+        action.enabled = true;
+        action.paused = p.paused;
+        action.time = p.startOffset;
+        action.timeScale = p.timeScale;
+        action.setLoop(LoopRepeat, Infinity);
+        action.play();
+
+        activeAnimations.push(action);
+      }
+    }
+
+    LoopAnimationData.set(eid, activeAnimations);
+
+    removeComponent(world, LoopAnimationInitialize, eid);
+    LoopAnimationInitializeData.delete(eid);
+  });
+
+  loopAnimationExitQuery(world).forEach((eid: number): void => {
+    const mixer = hasComponent(world, MixerAnimatable, eid) ? MixerAnimatableData.get(eid)! : null;
+    const activeAnimations = LoopAnimationData.get(eid)!;
+    for (let i = 0; i < activeAnimations.length; i++) {
+      const action = activeAnimations[i];
+      action.enabled = false;
+      action.stop();
+      if (mixer !== null) {
+        mixer.uncacheAction(action);
+      }
+    }
+    LoopAnimationData.delete(eid);
+  });
+}
--- a/src/bit-systems/mixer-animatable.ts
+++ b/src/bit-systems/mixer-animatable.ts
@@ -1,0 +1,33 @@
+import { addComponent, defineQuery, enterQuery, exitQuery, removeComponent } from "bitecs";
+import { AnimationMixer } from "three";
+import { MixerAnimatable, MixerAnimatableInitialize, MixerAnimatableData, Object3DTag } from "../bit-components";
+import { HubsWorld } from "../app";
+
+const initializeQuery = defineQuery([MixerAnimatableInitialize, Object3DTag]);
+const initializeEnterQuery = enterQuery(initializeQuery);
+const mixerQuery = defineQuery([MixerAnimatable, Object3DTag]);
+const mixerExitQuery = exitQuery(mixerQuery);
+
+export function mixerAnimatableSystem(world: HubsWorld): void {
+  initializeEnterQuery(world).forEach(eid => {
+    addComponent(world, MixerAnimatable, eid);
+
+    const object = world.eid2obj.get(eid)!;
+    const mixer = new AnimationMixer(object);
+    MixerAnimatableData.set(eid, mixer);
+
+    removeComponent(world, MixerAnimatableInitialize, eid);
+  });
+
+  mixerQuery(world).forEach(eid => {
+    const mixer = MixerAnimatableData.get(eid)!;
+    mixer.update(world.time.delta / 1000.0);
+  });
+
+  mixerExitQuery(world).forEach(eid => {
+    const mixer = MixerAnimatableData.get(eid)!;
+    mixer.stopAllAction();
+    mixer.uncacheRoot(mixer.getRoot());
+    MixerAnimatableData.delete(eid);
+  });
+}
--- a/src/inflators/loop-animation.ts
+++ b/src/inflators/loop-animation.ts
@@ -1,0 +1,60 @@
+import { addComponent } from "bitecs";
+import { LoopAnimationInitialize, LoopAnimationInitializeData } from "../bit-components";
+import { HubsWorld } from "../app";
+
+type ElementParams = {
+  activeClipIndex?: number;
+  // TODO: Do we need to keep supporting the following two params?
+  //       Perhaps we should detemine which one should be the
+  //       the canonical presentation and then handle converting
+  //       the others to that.
+  // DEPRECATED: Use activeClipIndex instead since animation
+  //             names are not unique
+  clip?: string;
+  // Support for Spoke->Hubs activeClipIndices struct
+  activeClipIndices?: number[];
+  paused?: boolean;
+  startOffset?: number;
+  timeScale?: number;
+};
+
+export type LoopAnimationParams = ElementParams[];
+
+const ELEMENT_DEFAULTS: Required<ElementParams> = {
+  activeClipIndex: 0,
+  clip: "",
+  activeClipIndices: [],
+  paused: false,
+  startOffset: 0,
+  timeScale: 1.0
+};
+
+const DEFAULTS: Required<LoopAnimationParams> = [ELEMENT_DEFAULTS];
+
+export function inflateLoopAnimationInitialize(
+  world: HubsWorld,
+  eid: number,
+  params: LoopAnimationParams = []
+): number {
+  if (params.length === 0) {
+    params = DEFAULTS;
+  }
+
+  const componentParams = [];
+  for (let i = 0; i < params.length; i++) {
+    const requiredParams = Object.assign({}, ELEMENT_DEFAULTS, params[i]) as Required<ElementParams>;
+    const activeClipIndices =
+      requiredParams.activeClipIndices.length > 0 ? requiredParams.activeClipIndices : [requiredParams.activeClipIndex];
+    componentParams.push({
+      activeClipIndices,
+      clip: APP.getSid(requiredParams.clip),
+      paused: requiredParams.paused,
+      startOffset: requiredParams.startOffset,
+      timeScale: requiredParams.timeScale
+    });
+  }
+
+  addComponent(world, LoopAnimationInitialize, eid);
+  LoopAnimationInitializeData.set(eid, componentParams);
+  return eid;
+}
--- a/src/inflators/model.tsx
+++ b/src/inflators/model.tsx
@@ -1,10 +1,11 @@
 import { addComponent, addEntity, hasComponent } from "bitecs";
 import { Material, Object3D } from "three";
 import { HubsWorld } from "../app";
-import { GLTFModel, MaterialTag } from "../bit-components";
+import { GLTFModel, MaterialTag, MixerAnimatableInitialize } from "../bit-components";
 import { addMaterialComponent, addObject3DComponent, gltfInflatorExists, gltfInflators } from "../utils/jsx-entity";
 import { mapMaterials } from "../utils/material-utils";
 import { EntityID } from "../utils/networking-types";
+import { inflateLoopAnimationInitialize, LoopAnimationParams } from "./loop-animation";
 
 function camelCase(s: string) {
   return s.replace(/-(\w)/g, (_, m) => m.toUpperCase());
@@ -13,7 +14,15 @@ function camelCase(s: string) {
 export type ModelParams = { model: Object3D };
 
 // These components are all handled in some special way, not through inflators
-const ignoredComponents = ["visible", "frustum", "frustrum", "shadow", "networked"];
+const ignoredComponents = [
+  "visible",
+  "frustum",
+  "frustrum",
+  "shadow",
+  "networked",
+  "animation-mixer",
+  "loop-animation"
+];
 
 function inflateComponents(
   world: HubsWorld,
@@ -111,6 +120,7 @@ export function inflateModel(world: HubsWorld, rootEid: number, { model }: Model
   // These components are special because we want to do a one-off action
   // that we can't do in a regular inflator (because they depend on the object3D).
   // If more things need to run at this point, we may need to expand the api here.
+  const loopAnimationParams: LoopAnimationParams = [];
   model.traverse(obj => {
     const components = obj.userData.gltfExtensions?.MOZ_hubs_components || {};
     if (components.visible) {
@@ -126,6 +136,10 @@ export function inflateModel(world: HubsWorld, rootEid: number, { model }: Model
       });
     }
 
+    if (components["loop-animation"]) {
+      loopAnimationParams.push(components["loop-animation"]);
+    }
+
     // We have had both spellings at different times.
     if (components.frustrum) {
       components.frustum = components.frustrum;
@@ -139,6 +153,16 @@ export function inflateModel(world: HubsWorld, rootEid: number, { model }: Model
     }
   });
 
+  // Hubs loop-animation component is defined at glTF node level.
+  // The component data are collected above in the scene graph traverse and
+  // bitECS LoopAnimation Component is set to root entity for simpler implementation.
+  // TODO: Probably the Hubs loop-animation component should be defined at scene
+  //       or root level. Revisit the specification.
+  // See https://github.com/mozilla/hubs/pull/5938#discussion_r1163410185
+  if (model.animations !== undefined && model.animations.length > 0) {
+    addComponent(world, MixerAnimatableInitialize, rootEid);
+    inflateLoopAnimationInitialize(world, rootEid, loopAnimationParams);
+  }
+
   addComponent(world, GLTFModel, rootEid);
-  // TODO Animation Mixer
 }
--- a/src/prefabs/loading-object.js
+++ b/src/prefabs/loading-object.js
@@ -1,16 +1,19 @@
 /** @jsx createElementEntity */
+import { BoxGeometry, Mesh, MeshBasicMaterial } from "three";
 import { createElementEntity } from "../utils/jsx-entity";
 import { loadModel } from "../components/gltf-model-plus";
 import loadingObjectSrc from "../assets/models/LoadingObject_Atom.glb";
-import { disposeNode } from "../utils/three-utils";
+import { cloneObject3D, disposeNode } from "../utils/three-utils";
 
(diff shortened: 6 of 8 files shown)


src/inflators/loop-animation.ts
```diff
@@ -0,0 +1,124 @@
...
+
+export type LoopAnimationParams = {
+  activeClipIndex: number;
+  paused: boolean;
+  startOffset: number;
+  timeScale: number;
+  // TODO: Do we need to keep supporting the following two params?
+  // DEPRECATED: Use activeClipIndex instead since animation
+  //             names are not unique
+  clip?: string;
```
netpro2k, line 35: It looks like Blender currently uses `clip` a comma separated list of names. Spoke uses `activeClipIndices`. I don't see anything using `activeClipIndex` but I bet there are old Spoke scenes that use it. We probably need to support all 3 for now :(

Lets discuss which one we want to be the canonical representation (probably `activeClipIndicies`) and then handle converting the others to that as early as possible in the flow, so the rest of the code does not need to think about it.

We have discussed in other places moving this sort of legacy migration all to an earlier point in GLTF loading after we completely remove gltf-model-plus, so we should add a TODO about that part of the code here.

src/inflators/loop-animation.ts
```diff
@@ -0,0 +1,124 @@
...
+      }
+    }
+    return activeClips;
+  } else {
+    return [animations[params.activeClipIndex]];
+  }
+};
+
+export function inflateLoopAnimation(world: HubsWorld, eid: number, params: LoopAnimationParams) : number {
+  // Is it possible to guarantee that LoopAnimation
```
netpro2k, line 73: No it's not. It's invalid to try and get the Object3D in an inflator. An inflator may be the thing that ends up creating the Object3D for an entity. Needing to get the object is a good sign that this code should be in an enter query instead. 

inflators should just set up the component data and (if they are a component managing *the* Object3D for an entity) create the Object3D and call `addObject3DComponent`. We definitely need to make sure to call this out clearly in the docs since I suspect it will be a repeated point of confusion.

src/systems/hubs-systems.ts
```diff
@@ -245,6 +246,7 @@ export function mainTick(xrFrame: XRFrame, renderer: WebGLRenderer, scene: Scene
   hubsSystems.gainSystem.tick();
   hubsSystems.nameTagSystem.tick();
   simpleWaterSystem(world);
+  mixerAnimationSystem(world, dt);
```
netpro2k, line 249: You can grab `dt` off of `world.time.delta`

src/bit-systems/mixer-animation.ts
```diff
@@ -0,0 +1,42 @@
...
+const exitedMixerAnimatableQuery = exitQuery(mixerAnimatableQuery);
+
+const loopAnimationQuery = defineQuery([LoopAnimation]);
+const exitedLoopAnimationQuery = exitQuery(loopAnimationQuery);
+
+// Question: Who should have AnimationMixer?
+//           MixerAnimatable component or Scene/App?
+
+export function mixerAnimationSystem(world: HubsWorld, delta: number) : void {
+  mixerAnimatableQuery(world).forEach(eid => {
```
netpro2k, line 28: Before this loop I think we can add an enteryQuery here for `MixerAnimatable` where we will actually set up the mixer. Then we can have an enter query for `[MixerAnimatable,LoopAnimation]` to set up the clip for the looping animation and add it to the mixer.
Then by the time we hit this loop everything is guaranteed to be set up.

src/inflators/model.tsx
```diff
@@ -139,6 +141,20 @@ export function inflateModel(world: HubsWorld, rootEid: number, { model }: Model
     }
   });
 
+  // TODO: Not sure if here is the best place and
+  //       this is the best way to inflate animations.
+  //       We may need to revisit.
+  if (model.animations !== undefined && model.animations.length > 0) {
+    inflateMixerAnimatable(world, rootEid, {
+      animations: model.animations
```
netpro2k, line 149: I don't think we need to copy the animations off the model. This will be the thing in eid2obj for rootEid, so we can just get the animations from that in the system when we need it. Here we can just call `addComponent(world, MixerAnimatable, rootEid)`, I don't think we even need an inflator.

src/inflators/model.tsx
```diff
@@ -139,6 +141,20 @@ export function inflateModel(world: HubsWorld, rootEid: number, { model }: Model
...
+    });
+    inflateLoopAnimation(world, rootEid, {
```
netpro2k, line 151: This still makes sense to do here to support the behavior we have now of: "If a gltf has animations and does not have any loop-animation component, loop the first animation"

src/inflators/loop-animation.ts
```diff
@@ -0,0 +1,124 @@
...
+  if (params.activeClipIndices !== undefined &&
+      params.activeClipIndices.length > 0) {
+    return params.activeClipIndices.map(index => animations[index]);
+  } else if (params.clip !== undefined &&
+             params.clip !== "") {
+    const activeClips = [];
+    const clipNames = params.clip!.split(",");
+    for (let i = 0; i < clipNames.length; i++) {
+      const clipName = clipNames[i];
+      const foundClip = animations.find((clip: AnimationClip) => {
```
netpro2k, line 57: Unfortunately since we need to support named animation lookup we will need to pass that all the way through to the system where we have the enterQuery since we can't really look up the animations here.

src/prefabs/loading-object.js
```diff
@@ -12,5 +12,15 @@ loadModel(loadingObjectSrc, null, true).then(gltf => {
 });
 
 export function LoadingObject() {
-  return <entity name="Loading Object" object3D={loadingObject.clone()} />;
+  const clonedObject = cloneObject3D(loadingObject);
+  return (
+    <entity
+	name="Loading Object"
```
netpro2k, line 18: linting errors in this file

src/utils/jsx-entity.ts
```diff
@@ -319,8 +320,9 @@ export interface JSXComponentData extends ComponentData {
     captureDurLblRef: Ref;
     sndToggleRef: Ref;
   };
-  animationMixer?: any;
   mediaLoader?: MediaLoaderParams;
+  mixerAnimatable?: MixerAnimatableParams;
```
netpro2k, line 324: not sure if we will end up needing an inflator for this or if we need to expose it to JSX. I think it's only valid to be inserted internally as part of a model component, but maybe we will find a use for manually making one in JSX as well... Not sure.

src/inflators/model.tsx
```diff
@@ -139,6 +141,20 @@ export function inflateModel(world: HubsWorld, rootEid: number, { model }: Model
     }
   });
 
+  // TODO: Not sure if here is the best place and
```
netpro2k, line 144: This seems like a reasonable spot to set up the MixerAnimatable component and set up a default LoopAnimation component if one does not exist already.

src/bit-systems/mixer-animation.ts
```diff
@@ -0,0 +1,42 @@
...
+    const { mixer } = MixerAnimatableData.get(eid)!;
+    mixer.update(delta / 1000.0);
+  });
+
+  // Not sure here is the best place to clean up
+  // these two components. We might need to revisit.
+  exitedLoopAnimationQuery(world).forEach(eid => {
```
netpro2k, line 35: Kind of minor, but I think we want to run this before we tick the animation mixers this frame otherwise we will run an additional frame of that animation.

src/bit-systems/mixer-animation.ts
```diff
@@ -0,0 +1,42 @@
...
+    cleanupLoopAnimation(eid);
+  });
+
+  exitedMixerAnimatableQuery(world).forEach(eid => {
+    cleanupMixerAnimatable(eid);
```
netpro2k, line 40: nit: No need for the wrapping function here and above, though I think most of the logic in there will be moving into this system anyway, so this code may just end up getting inlined instead of being with the inflator.

src/bit-components.js
```diff
@@ -150,7 +150,12 @@ MediaPDF.map = new Map();
 export const MediaVideo = defineComponent({
   autoPlay: Types.ui8
 });
-export const AnimationMixer = defineComponent();
+export const MixerAnimatable = defineComponent({});
+// Map<EntityId, {mixer: AnimationMixer: animations: Array<AnimationClip>}>
```
netpro2k, line 154: Not sure exactly the syntax but typescript can pickup on jsdoc comments for type info, so we might be able to use that here.  Similar to what we did here https://github.com/mozilla/hubs/blob/c78cec579eb422521e06cf1459469860277a47cd/src/components/gltf-model-plus.js#L818-L823

netpro2k (changes requested):
This is looking pretty close but I think we should move some stuff out of inflators into enter queries. I think what you have now will technically work for animated models without loop-animation components (they will play their first animation) but will not correctly handle models with actual `loop-animation` components (since the root object wont yet have an Object3DTag on it when calling the inflators). 

Its definitely very clear that inflators are quite a confusing concept and we need to better figure out how to describe what they should (and should not) be used for.

I do think we unfortunately do need to keep supporting all 3 methods of targeting animations since they are used in existing scenes...

Made a few other suggestions, but nothing major.

src/inflators/model.tsx
takahirox: I'm thinking of the opposite. Thinking if we can move animations away from Object3D to component. Object3D doesn't have animations property in the Three.js official API. It polluts the official Object3D Three.js and we need a type definition for it. And if we move animations to component, attaching/detaching animation might be easier and simpler (maybe).

src/inflators/loop-animation.ts
keianhzo: Related to Takahiro's comment, is there any reason why we are adding the Object3D here the last thing?:
https://github.com/mozilla/hubs/blob/a183e23e609f86db0612ccfb0b3b8e495625087a/src/inflators/model.tsx#L73-L92

If the Object3D entity was added first thing the component could have access to the Object3D in case it needs certain properties for any sort of initialization (scale, rotation, etc) instead of needing to implement a system/query just for that.

takahirox:
Updated to reflect the review comments.

I don't really like some of the designs in the PR, but want to focus on making it work for now because it is at very last minutes in this cycle. I want to revisit and refactor in the next or later cycle.

takahirox:
I started to think that the first design is not intuitive and is buggy. I rewrote the implementation. Please refer to the PR for the changes.

src/bit-systems/loop-animation.ts
```diff
@@ -0,0 +1,87 @@
+import { addComponent, defineQuery, enterQuery, exitQuery, hasComponent } from "bitecs";
+import { AnimationAction, AnimationClip, AnimationMixer, LoopRepeat } from "three";
+import { MixerAnimatable, MixerAnimatableData, LoopAnimation, LoopAnimationData, Object3DTag } from "../bit-components";
+import { HubsWorld } from "../app";
+
+const loopAnimationQuery = defineQuery([MixerAnimatable, LoopAnimation, Object3DTag]);
```
keianhzo, line 6: This works good for JSX entities like the LoadingObject where the `MixerAnimatable` and `LoopAnimation` components are both added to the root object where the animations are but when a glTF model is loaded the `MixerAnimatableInitialize` (and hence the `MixerAnimatable` component) is added to the glTF root node but the `LoopAnimation` component can be in a child node so `loopAnimationQuery` doesn't iterate over the right `LoopAnimation` component.

My understanding is that we need to add a `MixerAnimatable` and create a mixer for any object with animations and every child with a `LoopAnimation` component will look for a parent with a `MixerAnimatable` and get the mixer and animations and then create `AnimationActions` based on its `LoopAnimation`  parameters.

My other comment referred to this, when a model is loaded in `inflateModel` if there are any animations we can already create a mixer and link the animations, something like:

```typescript
if (model.animations !== undefined && model.animations.length > 0) {
    const mixer = new AnimationMixer(model);
    MixerAnimatableData.set(rootEid, {
      mixer,
      animations: model.animations
    });
    addComponent(world, MixerAnimatable, rootEid);
    if (!findChildWithComponent(world, LoopAnimation, rootEid)) {
      inflateLoopAnimation(world, rootEid, {});
    }
  }
```

and then during the `LoopAnimation` query enter loop retrieve the mixer and animations from closest parent:
```typescript
const mixerEid = findAncestorWithComponent(world, MixerAnimatable, eid);
const { mixer, animations } = MixerAnimatableData.get(mixerEid)!;
```

keianhzo (changes requested):
I've added some comments, let me know what you think.

src/bit-systems/mixer-animatable.ts
```diff
@@ -0,0 +1,33 @@
...
+export function mixerAnimatableSystem(world: HubsWorld): void {
+  initializeEnterQuery(world).forEach(eid => {
+    addComponent(world, MixerAnimatable, eid);
+
+    const object = world.eid2obj.get(eid)!;
+    const mixer = new AnimationMixer(object);
+    MixerAnimatableData.set(eid, mixer);
+
+    removeComponent(world, MixerAnimatableInitialize, eid);
+  });
```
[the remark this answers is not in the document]
takahirox: I have a mix feeling about it.

Pros: Yes, the implementation in model will be simpler. And the two step initializations (MixerAnimatableInitialize -> MixerAnimatable) for model will not be needed.
Cons: Object (AnimationMixer in this case) creation and clean up are spreaded out, creation is here while clean up is in mixerAnimatableSystem. I think it would be more readable and maintainable if both are placed in a single place. And MixerAnimatable initializations can be inconsistend between glTF and JSX. I think it would be more readable and maintainable if they are consistent.

Personally I prefer consistency and both creation and clean up are placed in a sinble place, so I want to go as is unless there is any critical problems.

src/bit-systems/loop-animation.ts
takahirox: I checked through the old code and you seem to be right. It seems that `AnimationMixer` needs to be found from ancestors and root entity needs to be aware of `LoopAnimation`s in its children. I don't really like this design, but we need to make the new code compatible with the old code for now. I will update the code.

src/bit-systems/loop-animation.ts
```diff
@@ -0,0 +1,87 @@
+import { addComponent, defineQuery, enterQuery, exitQuery, hasComponent } from "bitecs";
```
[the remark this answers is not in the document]
takahirox: Updated, thanks.

takahirox:
As @keianhzo mentioned, the previous implementation didn't handle Hubs loop-animation component in glTF correctly. I think it may be better to revisit the Hubs component specification but I don't think it's the best time to do that now. We should focus on migration and making the new entity system compatible with the old ones.

I updated the implementation to handle it. Refer to the PR for the update.

src/bit-systems/mixer-animatable.ts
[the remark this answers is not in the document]
keianhzo: Yes, makes sense, I think it's better to prioritize consistency.

src/bit-systems/loop-animation.ts
keianhzo: We have talked about this privately but for future reference about how the `loop-animation` component works.

> 1. Is loop-animation Hubs component a glTF node level extension, not a glTF scene level extension?

As of this writing the `loop-animation` component can the attached to any mesh node. In Blender we expose the animation actions/nla tracks of that mesh for selection so each mesh with a `loop-animation` will only reference it's own animations and it can't reference animations in other meshes. Same applies for Spoke, only meshes with animations will show the Loop Animation component options and list that mesh animations.

> 2. Can there be multiple loop-animations in glTF nodes in a single glTF, for example gltf.nodes[0], gltf.nodes[1], and gltf.nodes[2] have loop-animation component?

Yes, any mesh node can have one.

> 3. If 2. is yes, can loop-animation components have different component parameters from each other?

Yes, each `loop-animation` component can have their own independent parameters.

> 4. If 3. is yes, can loop-animation component parameters that are contradict from each other be valid, for example gltf.nodes[0].loop-animation = {activeClipIndex: 0, paused: false}, gltf.nodes[1].loop-animation = {activeClipIndex: 0, paused: true}?

No, each `loop-animation` component will only reference it's own meshes animations and each mesh node can on ly have one `loop-animation` component so conflict is not possible.

src/inflators/loop-animation.ts
```diff
@@ -0,0 +1,65 @@
...
+const DEFAULTS: Required<LoopAnimationParams> = [
+  {
+    activeClipIndex: 0,
+    clip: "",
+    activeClipIndices: [],
+    paused: false,
+    startOffset: 0,
+    timeScale: 1.0
+  } as Required<ElementParams>
+];
```
keianhzo, line 41: Couldn't this be reduced to?
```javascript
const DEFAULTS: LoopAnimationParams = [ELEMENT_DEFAULTS];
```

keianhzo (commented):
I'd like to understand better the pros/cons of this approach Vs keeping the `LoopAnimation` component in each entity and rely on an ancestor having a mixer. Any entity that has a `LoopAnimation` component should have an ancestor with a mixer as they are actually completely coupled so I think it would be fair to rely on that relationship. I feel we are adding a few more hops here to avoid that.

The concept of inflator seems to be diverging here from what we are doing for other components. I would expect it to add a `LoopAnimation` component but it's not doing that, the component is added in the system. Maybe we could rename it to something like `inflateLoopAnimationInitialize`.

Most of it is design related and the above it's just my preference so I don't want to block as many of this will be discussed eventually and this is working correctly.

src/inflators/loop-animation.ts
takahirox: Good idea. Updated, thanks.

src/bit-systems/loop-animation.ts
takahirox: Thanks for the explanation. My impression is Hubs animation system assumes a little bit too much implied rules in glTF, for example it assumes that an animation clip consists of a ceratain root glTF node and its child nodes but there is no such a limitation in the glTF animation specification.

I think it would be better to let `loop-animation` component be aligned with glTF spec more and remove such implied rules. Probably the component specification will be more straightforward and less problematic, for example when processing a Hubs glTF assets with external glTF tools.

If we want to keep the animation representation in Blender, for example binding a `loop-animation` copmponent to a certain node, we may add info needed in glTF extra field.

It isn't the best time to revisit the `loop-animation` component spec now. We may revisit the spec later.

takahirox:
> Any entity that has a LoopAnimation component should have an ancestor with a mixer as they are actually completely coupled so I think it would be fair to rely on that relationship.

Some thoughts.

1. No dependency from an entity with others is better for sure in terms of readability, maintainability, and so on.
2. I don't think the current the Hubs `loop-animation` component specification is good. It should be moved to root or scene level. Then `LoopAnimation` bitECS component will be able to be set to the root entity without any problem.
3. But we have to support the current Hubs `loop-animation` component. We may need hack anywhere anyways. Basically we need to choose which one, my implementation or having dependency with other entities. I prefer mine because hack is placed to be only in `model` inflator. And I don't really want a system to be complex for the component specification we may revisit later.

Regarding the naming, ok `inflateLoopAnimationInitialize` is less confusing. Updated.

takahirox:
This PR needs to be unlocked by the reviewers to get merged.

keianhzo (approved):

keianhzo:
@takahirox unlocked

src/bit-systems/mixer-animatable.ts
```diff
@@ -0,0 +1,33 @@
...
+import { MixerAnimatable, MixerAnimatableInitialize, MixerAnimatableData, Object3DTag } from "../bit-components";
+import { HubsWorld } from "../app";
+
+const initializeQuery = defineQuery([MixerAnimatableInitialize, Object3DTag]);
+const initializeEnterQuery = enterQuery(initializeQuery);
+const mixerQuery = defineQuery([MixerAnimatable, Object3DTag]);
+const mixerExitQuery = exitQuery(mixerQuery);
+
+export function mixerAnimatableSystem(world: HubsWorld): void {
+  initializeEnterQuery(world).forEach(eid => {
```
netpro2k, line 12: Confused why we need 2 components here (MixerAnimatableInitialize and MixerAnimatable). Can't we just create the Mixer on enter query for MixerAnimatable? That's what I did in the quick stub of this I am using in my behavior graph branch and it seems to work well https://github.com/mozilla/hubs/blob/behavior-graphs-spike/src/bit-systems/behavior-graph.ts#L82-L101. Since we have strong control over exectuion order we can guarantee that the mixer is set up before any other code using it runs (both in this system and the loop-animation system).

I would also prefer to keep the map on the MixerAnimatable component itself instead of having it floating seperatly. It makes it feel more like any other component property: https://github.com/mozilla/hubs/blob/behavior-graphs-spike/src/bit-components.js#L185-L190. With the type comment I have there it also typechecks just fine too.

src/bit-systems/mixer-animatable.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    removeComponent(world, MixerAnimatableInitialize, eid);
+  });
+
+  mixerQuery(world).forEach(eid => {
+    const mixer = MixerAnimatableData.get(eid)!;
+    mixer.update(world.time.delta / 1000.0);
+  });
+
+  mixerExitQuery(world).forEach(eid => {
```
netpro2k, line 27: I don't think it particularly matters in this case, but I tend to do enter query, exit query, and then the general work of the system.

src/bit-components.js
```diff
@@ -178,7 +178,28 @@ MediaPDF.map = new Map();
 export const MediaVideo = defineComponent({
   autoPlay: Types.ui8
 });
-export const AnimationMixer = defineComponent();
+export const MixerAnimatableInitialize = defineComponent({});
+export const MixerAnimatable = defineComponent({});
+/**
+ * @type {Map<EntityId, AnimationMixer}>}
+ */
+export const MixerAnimatableData = new Map();
```
netpro2k, line 186: Would prefer we start putting component properties on this on the component and just using type annotations, like https://github.com/mozilla/hubs/blob/behavior-graphs-spike/src/bit-components.js#L185-L190

src/bit-systems/loop-animation.ts
```diff
@@ -0,0 +1,105 @@
...
+import { HubsWorld } from "../app";
+
+const loopAnimationInitializeQuery = defineQuery([LoopAnimationInitialize, MixerAnimatable, Object3DTag]);
+const loopAnimationInitializeEnterQuery = enterQuery(loopAnimationInitializeQuery);
+
+const loopAnimationQuery = defineQuery([LoopAnimation, MixerAnimatable, Object3DTag]);
+const loopAnimationExitQuery = exitQuery(loopAnimationQuery);
+
+// Question: Who should have AnimationMixer?
+//           MixerAnimatable component or Scene/App?
```
netpro2k, line 21: I think it makes sense to have "on" a MixerAnimatable component . Then when loop-animation components are added we find the nearest ancestor that is MixerAnimatable and use its mixer.

src/inflators/loop-animation.ts
```diff
@@ -0,0 +1,60 @@
...
+  //             names are not unique
+  clip?: string;
+  // Support for Spoke->Hubs activeClipIndices struct
+  activeClipIndices?: number[];
+  paused?: boolean;
+  startOffset?: number;
+  timeScale?: number;
+};
+
+export type LoopAnimationParams = ElementParams[];
```
netpro2k, line 21: Was quite confused why this is an array, but I see in model inflator that you are attempting to consolidate all the loop-animation components in a model into a single one. I think that makes some sense. It's the component we would like to have. Though if we are going to do that anyway, I think the component data should be consolidated earlier too. Maybe it should just be an array of clips or actions (since that's what you end up doing in the system anyway). That way we can move the messy handling of the legacy properties all up front and not have the rest of the code need to think about it.

src/inflators/model.tsx
```diff
@@ -139,6 +153,16 @@ export function inflateModel(world: HubsWorld, rootEid: number, { model }: Model
     }
   });
 
+  // Hubs loop-animation component is defined at glTF node level.
```
netpro2k, line 156: Not sure I would have bothered trying to consolidate these at this stage. Writing an inflator for the GLTF data that added a LoopAnimation component, who's data is an array of clips, and then having a system that find's the ancestor mixer to play those clips on could have worked. I suppose this gets us closer to where we would eventually like to be with it but adding more special cases to `ignoredComponents` is definitely something we should really try and minimize. Animations in three are tricky though and quite tied to GLTF loading and the threejs hierarchy, so I can see why it might be justified.

src/prefabs/loading-object.js
```diff
@@ -1,16 +1,19 @@
...
 
 // TODO We should have an explicit "preload assets" step
-let loadingObject = new THREE.Mesh(new THREE.BoxGeometry(), new THREE.MeshBasicMaterial());
+let loadingObject = new Mesh(new BoxGeometry(), new MeshBasicMaterial());
 loadModel(loadingObjectSrc, null, true).then(gltf => {
   disposeNode(loadingObject);
   loadingObject = gltf.scene;
 });
 
+// TODO: Do we really need to clone the loadingObject every time?
```
netpro2k, line 15: We may have more than one so we definitely need multiple. We could render it as instanced geometry which might be a good idea, probably even better than pooling.

src/bit-systems/loop-animation.ts
```diff
@@ -0,0 +1,105 @@
...
+    return activeClipIndices.map((index: number) => animations[index]);
+  }
+};
+
+export function loopAnimationSystem(world: HubsWorld): void {
+  loopAnimationInitializeEnterQuery(world).forEach((eid: number): void => {
+    const object = world.eid2obj.get(eid)!;
+    const mixer = MixerAnimatableData.get(eid)!;
+
+    addComponent(world, LoopAnimation, eid);
```
netpro2k, line 53: I might be missing something since you ended up doing it twice. What does the additional `LoopAnimationInitialize` component get us vs doing this in an enter query for `LoopAnimation` itself?

netpro2k (commented):
This now seems to handle both caes but I left some feedback into how things are structured. The additional "Initialize" components feel quite strange to me, and I am not too sure what they are buying us. Also suggested possibly cleaner ways of exposing the `Map` data for the components. I am slightly uneasy about how the LoopAnimation components have been consolidated, but I think that's fine since that probably the direction we wanted to eventually go anyway.

netpro2k (approved):

PR #5944, opened 2023-02-15
bitecs audio-zone-system & audio-debug-system

keianhzo:
This PR migrates the audio zones and audio debug systems (mainly because having the debugger is handy to test the audio zones).

- Audio emitter is now a separate system that only handles `media-video` emitters for now (will handle all the audio emitter sources at a later stage)
- The new audio zones and audio debugger only work with `media-video` emitters yet. I prefer to migrate the rest of the audio related components first (`audio-target`, `zone-audio-source`) in other separater PRs before consolidating them in the audio emitters system.
- I've added a `MediaVideoPlaybackChanged` component to handle media-video playback state changes.
- There is a new `AudioListenerTag` component and that makes the `audio-zone-system` to handle the audio zone entities in two different groups: emitters and listeners. Another option would have been to have a specific `AudioZoneEntity` tag but in most cases that would require to double tag most of the entities so I opted for reusing existing tags.
- I'm using `getScene` as the init point for the service, not sure if this is what we want or we prefer a direct init method call in `hubs-systems` init as we currently do for the previous systems.
- I've updated the audio debug shader to work in world space instead of local space. That removed the need to transforming the emitters to local object space and also allows reusing that material for any object that wants to show audio debug info (ie. in case we support multiple nav meshes in the future)
- New audio zones dynamic parameter to avoid world matrix updates unless the audio zone is marked as movable. Requires: https://github.com/MozillaReality/hubs-blender-exporter/pull/198

Next steps on audio migration:
- Migrate `audio-target` and `audio-zone-source`
- Migrate `audio-gain-system`
- Migrate `audio-settings-system`
- Integrate avatar audio related systems (`avatar-audio-source` and `avatar-volume-controls`) in the new bitecs audio systems.
- Integrate `linked-media` in the new bitecs audio systems.
- Look into bypassing the ThreeJS Audio nodes and handle them by just using the WebAudio API.
- As I make progress in the above tasks I'll be separating the previous APP audio stores into a separated set of stores hanging from `AudioEmitter`: `AudioEmitter.sourceType`, `AudioEmitter.audioOverrides`, etc to keep both AFrame and BitECS systems separated.

### DIFF
--- a/src/bit-components.js
+++ b/src/bit-components.js
@@ -235,3 +235,19 @@ export const Mirror = defineComponent();
 export const ParticleEmitterTag = defineComponent({
   src: Types.ui32
 });
+export const AudioZone = defineComponent({
+  flags: Types.ui8
+});
+export const AudioListenerTag = defineComponent();
+export const AudioParams = defineComponent({
+  audioType: Types.ui32,
+  distanceModel: Types.ui32,
+  panningModel: Types.ui32,
+  rolloffFactor: Types.ui32,
+  refDistance: Types.ui32,
+  maxDistance: Types.ui32,
+  coneInnerAngle: Types.ui32,
+  coneOuterAngle: Types.ui32,
+  coneOuterGain: Types.ui32,
+  gain: Types.ui32
+});
--- a/src/bit-components.js
+++ b/src/bit-components.js
@@ -150,6 +150,7 @@ MediaPDF.map = new Map();
 export const MediaVideo = defineComponent({
   autoPlay: Types.ui8
 });
+export const MediaVideoPlaybackChanged = defineComponent();
 export const AnimationMixer = defineComponent();
 export const NetworkedVideo = defineComponent({
   time: Types.f32,
--- a/src/bit-components.js
+++ b/src/bit-components.js
@@ -242,7 +242,6 @@ export const ParticleEmitterTag = defineComponent({
 export const AudioZone = defineComponent({
   flags: Types.ui8
 });
-export const AudioListenerTag = defineComponent();
 export const AudioParams = defineComponent({
   audioType: Types.ui32,
   distanceModel: Types.ui32,
--- a/src/bit-components.js
+++ b/src/bit-components.js
@@ -128,6 +128,9 @@ export const MediaLoader = defineComponent({
   flags: Types.ui8
 });
 MediaLoader.src[$isStringType] = true;
+export const MediaLoaded = defineComponent({
+  rootRef: Types.eid
+});
 
 export const SceneRoot = defineComponent();
 export const NavMesh = defineComponent();
--- a/src/bit-components.js
+++ b/src/bit-components.js
@@ -242,15 +242,3 @@ export const ParticleEmitterTag = defineComponent({
 export const AudioZone = defineComponent({
   flags: Types.ui8
 });
-export const AudioParams = defineComponent({
-  audioType: Types.ui32,
-  distanceModel: Types.ui32,
-  panningModel: Types.ui32,
-  rolloffFactor: Types.ui32,
-  refDistance: Types.ui32,
-  maxDistance: Types.ui32,
-  coneInnerAngle: Types.ui32,
-  coneOuterAngle: Types.ui32,
-  coneOuterGain: Types.ui32,
-  gain: Types.ui32
-});
--- a/src/bit-components.js
+++ b/src/bit-components.js
@@ -252,4 +252,3 @@ export const AudioParams = defineComponent({
   coneOuterGain: Types.ui32,
   gain: Types.ui32
 });
-export const AudioDebugChanged = defineComponent();
--- a/src/bit-components.js
+++ b/src/bit-components.js
@@ -128,9 +128,7 @@ export const MediaLoader = defineComponent({
   flags: Types.ui8
 });
 MediaLoader.src[$isStringType] = true;
-export const MediaLoaded = defineComponent({
-  rootRef: Types.eid
-});
+export const MediaLoaded = defineComponent();
 
 export const SceneRoot = defineComponent();
 export const NavMesh = defineComponent();
@@ -153,7 +151,6 @@ MediaPDF.map = new Map();
 export const MediaVideo = defineComponent({
   autoPlay: Types.ui8
 });
-MediaVideo.map = new Map();
 export const AnimationMixer = defineComponent();
 export const NetworkedVideo = defineComponent({
   time: Types.f32,
@@ -171,6 +168,8 @@ export const VideoMenu = defineComponent({
 export const AudioEmitter = defineComponent({
   flags: Types.ui8
 });
+AudioEmitter.audios = new Map();
+AudioEmitter.params = new Map();
 export const AudioSettingsChanged = defineComponent();
 export const Deletable = defineComponent();
 export const EnvironmentSettings = defineComponent();
@@ -244,3 +243,4 @@ export const ParticleEmitterTag = defineComponent({
 export const AudioZone = defineComponent({
   flags: Types.ui8
 });
+export const AudioParams = defineComponent();
--- a/src/bit-components.js
+++ b/src/bit-components.js
@@ -153,7 +153,7 @@ MediaPDF.map = new Map();
 export const MediaVideo = defineComponent({
   autoPlay: Types.ui8
 });
-export const MediaVideoPlaybackChanged = defineComponent();
+MediaVideo.map = new Map();
 export const AnimationMixer = defineComponent();
 export const NetworkedVideo = defineComponent({
   time: Types.f32,
@@ -168,7 +168,9 @@ export const VideoMenu = defineComponent({
   playIndicatorRef: Types.eid,
   pauseIndicatorRef: Types.eid
 });
-export const AudioEmitter = defineComponent();
+export const AudioEmitter = defineComponent({
+  flags: Types.ui8
+});
 export const AudioSettingsChanged = defineComponent();
 export const Deletable = defineComponent();
 export const EnvironmentSettings = defineComponent();
--- a/src/bit-components.js
+++ b/src/bit-components.js
@@ -251,3 +251,4 @@ export const AudioParams = defineComponent({
   coneOuterGain: Types.ui32,
   gain: Types.ui32
 });
+export const AudioDebugChanged = defineComponent();
--- a/src/bit-systems/audio-emitter-system.ts
+++ b/src/bit-systems/audio-emitter-system.ts
@@ -14,6 +14,16 @@ function isPositionalAudio(node: AudioObject3D): node is PositionalAudio {
   return (node as any).panner !== undefined;
 }
 
+export function cleanupAudio(audio: AudioObject3D) {
+  const eid = audio.eid!;
+  audio.disconnect();
+  const audioSystem = APP.scene?.systems["hubs-systems"].audioSystem;
+  APP.audios.delete(eid);
+  APP.supplementaryAttenuation.delete(eid);
+  APP.isAudioPaused.delete(eid);
+  audioSystem.removeAudio({ node: audio });
+}
+
 function swapAudioType<T extends AudioObject3D>(
   world: HubsWorld,
   audioSystem: AudioSystem,
@@ -22,6 +32,10 @@ function swapAudioType<T extends AudioObject3D>(
 ) {
   const audio = world.eid2obj.get(eid)! as AudioObject3D;
   audio.disconnect();
+  APP.gainMultipliers.delete(eid);
+  APP.sourceType.set(eid, SourceType.MEDIA_VIDEO);
+  APP.supplementaryAttenuation.delete(eid);
+  APP.audios.delete(eid);
   audioSystem.removeAudio({ node: audio });
 
   const newAudio = new NewType(APP.audioListener);
--- a/src/bit-systems/audio-emitter-system.ts
+++ b/src/bit-systems/audio-emitter-system.ts
@@ -10,7 +10,7 @@ import { addObject3DComponent, swapObject3DComponent } from "../utils/jsx-entity
 type AudioObject3D = StereoAudio | PositionalAudio;
 type AudioConstructor<T> = new (listener: ThreeAudioListener) => T;
 
-function isPositionalAudio(node: AudioObject3D): node is PositionalAudio {
+export function isPositionalAudio(node: AudioObject3D): node is PositionalAudio {
   return (node as any).panner !== undefined;
 }
 
--- a/src/bit-systems/audio-emitter-system.ts
+++ b/src/bit-systems/audio-emitter-system.ts
@@ -7,8 +7,11 @@ import { AudioSystem } from "../systems/audio-system";
 import { applySettings, getCurrentAudioSettings, updateAudioSettings } from "../update-audio-settings";
 import { addObject3DComponent, swapObject3DComponent } from "../utils/jsx-entity";
 
-type AudioObject3D = StereoAudio | PositionalAudio;
+export type AudioObject3D = StereoAudio | PositionalAudio;
 type AudioConstructor<T> = new (listener: ThreeAudioListener) => T;
+
+export const Emitter2Audio = (AudioEmitter as any).audios as Map<number, number>;
+export const Emitter2Params = (AudioEmitter as any).params as Map<number, number>;
 
 export const EMITTER_FLAGS = {
   MUTED: 1 << 0,
@@ -26,6 +29,7 @@ export function cleanupAudio(audio: AudioObject3D) {
   const audioSystem = APP.scene?.systems["hubs-systems"].audioSystem;
   APP.audios.delete(eid);
   APP.supplementaryAttenuation.delete(eid);
+  APP.audioOverrides.delete(eid);
   audioSystem.removeAudio({ node: audio });
 }
 
@@ -40,6 +44,7 @@ function swapAudioType<T extends AudioObject3D>(
   APP.gainMultipliers.delete(eid);
   APP.sourceType.set(eid, SourceType.MEDIA_VIDEO);
   APP.supplementaryAttenuation.delete(eid);
+  APP.audioOverrides.delete(eid);
   APP.audios.delete(eid);
   audioSystem.removeAudio({ node: audio });
 
--- a/src/bit-systems/audio-emitter-system.ts
+++ b/src/bit-systems/audio-emitter-system.ts
@@ -41,10 +41,8 @@ function swapAudioType<T extends AudioObject3D>(
 ) {
   const audio = world.eid2obj.get(eid)! as AudioObject3D;
   audio.disconnect();
-  APP.gainMultipliers.delete(eid);
   APP.sourceType.set(eid, SourceType.MEDIA_VIDEO);
   APP.supplementaryAttenuation.delete(eid);
-  APP.audioOverrides.delete(eid);
   APP.audios.delete(eid);
   audioSystem.removeAudio({ node: audio });
 
--- a/src/bit-systems/audio-emitter-system.ts
+++ b/src/bit-systems/audio-emitter-system.ts
@@ -1,0 +1,89 @@
+import { addComponent, addEntity, defineQuery, removeComponent } from "bitecs";
+import { PositionalAudio, Audio as StereoAudio, AudioListener as ThreeAudioListener } from "three";
+import { HubsWorld } from "../app";
+import { AudioEmitter, AudioSettingsChanged } from "../bit-components";
+import { AudioType, SourceType } from "../components/audio-params";
+import { AudioSystem } from "../systems/audio-system";
+import { applySettings, getCurrentAudioSettings, updateAudioSettings } from "../update-audio-settings";
+import { addObject3DComponent, swapObject3DComponent } from "../utils/jsx-entity";
+
+type AudioObject3D = StereoAudio | PositionalAudio;
+type AudioConstructor<T> = new (listener: ThreeAudioListener) => T;
+
+function isPositionalAudio(node: AudioObject3D): node is PositionalAudio {
+  return (node as any).panner !== undefined;
+}
+
+function swapAudioType<T extends AudioObject3D>(
+  world: HubsWorld,
+  audioSystem: AudioSystem,
+  eid: number,
+  NewType: AudioConstructor<T>
+) {
+  const audio = world.eid2obj.get(eid)! as AudioObject3D;
+  audio.disconnect();
+  audioSystem.removeAudio({ node: audio });
+
+  const newAudio = new NewType(APP.audioListener);
+  newAudio.setNodeSource(audio.source!);
+  audioSystem.addAudio({ sourceType: SourceType.MEDIA_VIDEO, node: newAudio });
+  APP.audios.set(eid, newAudio);
+
+  audio.parent!.add(newAudio);
+  audio.removeFromParent();
+
+  swapObject3DComponent(world, eid, newAudio);
+}
+
+export function makeAudioSourceEntity(world: HubsWorld, video: HTMLVideoElement, audioSystem: AudioSystem) {
+  const eid = addEntity(world);
+  APP.sourceType.set(eid, SourceType.MEDIA_VIDEO);
+  if (video.paused) {
+    APP.isAudioPaused.add(eid);
+  } else {
+    APP.isAudioPaused.delete(eid);
+  }
+
+  let audio;
+  const { audioType } = getCurrentAudioSettings(eid);
+  const audioListener = APP.audioListener;
+  if (audioType === AudioType.PannerNode) {
+    audio = new PositionalAudio(audioListener);
+  } else {
+    audio = new StereoAudio(audioListener);
+  }
+  addComponent(world, AudioEmitter, eid);
+  addObject3DComponent(world, eid, audio);
+
+  audio.gain.gain.value = 0;
+  audioSystem.addAudio({ sourceType: SourceType.MEDIA_VIDEO, node: audio });
+
+  const audioSrcEl = video;
+  audio.setMediaElementSource(audioSrcEl);
+
+  APP.audios.set(eid, audio);
+  updateAudioSettings(eid, audio);
+
+  // Original audio source volume can now be restored as audio systems will take over
+  audioSrcEl.volume = 1;
+  return eid;
+}
+
+const staleAudioEmittersQuery = defineQuery([AudioEmitter, AudioSettingsChanged]);
+export function audioEmitterSystem(world: HubsWorld, audioSystem: AudioSystem) {
+  staleAudioEmittersQuery(world).forEach(function (eid) {
+    const audio = world.eid2obj.get(eid)! as PositionalAudio | StereoAudio;
+    const settings = getCurrentAudioSettings(eid);
+    const isPannerNode = isPositionalAudio(audio);
+
+    // TODO this needs more testing
+    if (!isPannerNode && settings.audioType === AudioType.PannerNode) {
+      swapAudioType(world, audioSystem, eid, PositionalAudio);
+    } else if (isPannerNode && settings.audioType === AudioType.Stereo) {
+      swapAudioType(world, audioSystem, eid, StereoAudio);
+    }
+
+    applySettings(audio, settings);
+    removeComponent(world, AudioSettingsChanged, eid);
+  });
+}
--- a/src/bit-systems/audio-emitter-system.ts
+++ b/src/bit-systems/audio-emitter-system.ts
@@ -10,6 +10,12 @@ import { addObject3DComponent, swapObject3DComponent } from "../utils/jsx-entity
 type AudioObject3D = StereoAudio | PositionalAudio;
 type AudioConstructor<T> = new (listener: ThreeAudioListener) => T;
 
+export const EMITTER_FLAGS = {
+  MUTED: 1 << 0,
+  PAUSED: 1 << 1,
+  CLIPPED: 1 << 2
+};
+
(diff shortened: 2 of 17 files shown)


src/bit-systems/audio-zone-system.ts
```diff
@@ -0,0 +1,318 @@
...
+const prevZones = new Map<number, Set<number>>();
+const aabbs = new Map<number, Box3>();
+
+const listenerPos = new Vector3();
+const audioZoneQuery = defineQuery([AudioZone]);
+const audioZoneEnterQuery = enterQuery(audioZoneQuery);
+const audioZoneExitQuery = exitQuery(audioZoneQuery);
+const audioZoneSourceQuery = defineQuery([AudioEmitter]);
+const audioZoneSourceEnterQuery = enterQuery(audioZoneSourceQuery);
+const audioZoneSourceExitQuery = exitQuery(audioZoneSourceQuery);
```
johnshaughnessy, line 256: The query looks for `AudioEmitter`, so these names might be clearer:

 `audioEmitterQuery`, `audioEmitterEnterQuery`, `audioEmitterExitQuery` OR
 `audioEmittersQuery`, `enteredAudioEmittersQuery`, `exitedAudioEmittersQuery`.
 
 (We should probably pick one or the other pattern and use it everywhere, but we haven't been consistent between these two.)

src/bit-systems/audio-zone-system.ts
```diff
@@ -0,0 +1,318 @@
...
+  audioZoneExitQuery(world).forEach(zone => {
+    aabbs.delete(zone);
+    debugObjects.delete(zone);
+
+    const isDebugEnabled = APP.store.state.preferences.showAudioDebugPanel;
+    isDebugEnabled && debugObjects.has(zone) && releaseZoneDebugObject(APP.world, zone);
+  });
+
+  const listener = APP.audioListener.eid!;
+  const listenersQuery = audioZoneListenerQuery(world);
```
johnshaughnessy, line 292: Do we ever expect more than one `AudioListenerTag`? If the expected result of `listenersQuery(world)` is `[listener]`, then perhaps we don't need the `AudioListenerTag` queries.

src/bit-systems/audio-zone-system.ts
```diff
@@ -0,0 +1,318 @@
...
+  const sourcesQuery = audioZoneSourceQuery(world);
+
+  zones.forEach(zone => {
+    APP.audioListener.getWorldPosition(listenerPos);
```
johnshaughnessy, line 296: This can be done before the `forEach`

src/inflators/audio-params.ts
```diff
@@ -0,0 +1,11 @@
+import { HubsWorld } from "../app";
+import { AudioSettings } from "../components/audio-params";
+import { updateAudioSettings } from "../update-audio-settings";
+
+export function inflateAudioParams(world: HubsWorld, eid: number, params: AudioSettings) {
+  APP.audioOverrides.set(eid, params);
+  const audio = APP.audios.get(eid);
+  if (audio) {
+    updateAudioSettings(eid, audio);
+  }
```
johnshaughnessy, line 10: I don't think you need to call `updateAudioSettings` here, because reaching into `APP.audios` is analogous to reaching into `world.eid2obj`. That is -- it's outside of the scope of responsibilities of an inflator. 

I am surprised to see that there is a bitecs `AudioParams` component defined in `bit-components.js`, which is not used here.

That said, I don't think you should have the same information duplicated in multiple places, so if `APP.audioOverrides` is going to continue to exist, I we should probably not define a bitecs `AudioParams` component with all the same values. (A tag component might be necessary for the purpose of querying, but I'm not sure of even that.)

src/bit-components.js
```diff
@@ -235,3 +236,19 @@ export const Mirror = defineComponent();
...
+  distanceModel: Types.ui32,
+  panningModel: Types.ui32,
+  rolloffFactor: Types.ui32,
+  refDistance: Types.ui32,
+  maxDistance: Types.ui32,
+  coneInnerAngle: Types.ui32,
+  coneOuterAngle: Types.ui32,
+  coneOuterGain: Types.ui32,
+  gain: Types.ui32
+});
```
johnshaughnessy, line 254: It seems like this is not being used, which makes sense because we already store this information in a map (`APP.audioOverrides`). 

At some point it might make sense to move these data into bitecs components, but it doesn't seem like a necessary step for what we're trying to do now:

```
  audios = new Map<AElement | number, PositionalAudio | Audio>();
  sourceType = new Map<AElement | number, SourceType>();
  audioOverrides = new Map<AElement | number, AudioSettings>();
  zoneOverrides = new Map<AElement | number, AudioSettings>();
  gainMultipliers = new Map<AElement | number, number>();
  supplementaryAttenuation = new Map<AElement | number, number>();
  clippingState = new Set<AElement | number>();
  mutedState = new Set<AElement | number>();
  isAudioPaused = new Set<AElement | number>();
```

src/bit-systems/audio-zone-system.ts
```diff
@@ -0,0 +1,318 @@
...
+const listenerPos = new Vector3();
+const audioZoneQuery = defineQuery([AudioZone]);
+const audioZoneEnterQuery = enterQuery(audioZoneQuery);
+const audioZoneExitQuery = exitQuery(audioZoneQuery);
+const audioZoneSourceQuery = defineQuery([AudioEmitter]);
+const audioZoneSourceEnterQuery = enterQuery(audioZoneSourceQuery);
+const audioZoneSourceExitQuery = exitQuery(audioZoneSourceQuery);
+const audioZoneListenerQuery = defineQuery([AudioListenerTag]);
+const audioZoneListenerEnterQuery = enterQuery(audioZoneListenerQuery);
+const audioZoneListenerExitQuery = exitQuery(audioZoneListenerQuery);
```
johnshaughnessy, line 259: Similarly, these can be `audioListenerQuery`, `audioListenerEnterQuery`, `audioListenerExitQuery`. I don't think having "Zone" in the name is helpful here.

src/bit-systems/audio-zone-system.ts
```diff
@@ -0,0 +1,318 @@
...
+  zones.forEach(zone => {
+    APP.audioListener.getWorldPosition(listenerPos);
+    addOrRemoveZone(currZones.get(listener)!, zone, listenerPos);
+    sourcesQuery.forEach((source: number) => {
+      addOrRemoveZone(currZones.get(source)!, zone, getSourcePosition(source));
+    });
+  });
+
+  sourcesQuery.forEach(source => {
+    const isListenerUpdated = isUpdated(currZones.get(listener)!, prevZones.get(listener)!);
```
johnshaughnessy, line 304: This can be done before the `forEach`

src/bit-systems/audio-zone-system.ts
```diff
@@ -0,0 +1,318 @@
...
+const audioZoneSourceQuery = defineQuery([AudioEmitter]);
+const audioZoneSourceEnterQuery = enterQuery(audioZoneSourceQuery);
+const audioZoneSourceExitQuery = exitQuery(audioZoneSourceQuery);
+const audioZoneListenerQuery = defineQuery([AudioListenerTag]);
+const audioZoneListenerEnterQuery = enterQuery(audioZoneListenerQuery);
+const audioZoneListenerExitQuery = exitQuery(audioZoneListenerQuery);
+
+export function audioZoneSystem(world: HubsWorld) {
+  const zones = audioZoneQuery(world);
+  if (!zones.length) return;
```
johnshaughnessy, line 263: We can't early out here. Exit queries still need to be allowed to run after the last `AudioZone` component disappears.

src/bit-systems/audio-zone-system.ts
```diff
@@ -0,0 +1,318 @@
...
+
+  [...audioZoneSourceEnterQuery(world), ...audioZoneListenerEnterQuery(world)].forEach(entity => {
+    currZones.set(entity, new Set());
+    prevZones.set(entity, new Set());
+  });
+  [...audioZoneSourceExitQuery(world), ...audioZoneListenerExitQuery(world)].forEach(entity => {
+    currZones.delete(entity);
+    prevZones.delete(entity);
+  });
```
johnshaughnessy, line 272: It feels strange that the audio listener needs to keep track of the zones that it is in. Isn't each zone keeping track of which entities are inside it? In this case, the zones that an audio listener is in each frame is a _derived property_. Even if we want to cache it to compare it each frame, it feels strange to store the information in the same maps.

In other words, `currZones` and `prevZones` are used to store two very different types of associations:
- On the one hand, they associate the `EntityID` of an `audioZone` to a `Set` of entities that are inside that zone. 
- On the other hand, they also associate the `EntityID` of an `audioListener` to a `Set` of `audioZone EntityIDs` that the listener is inside of.

src/bit-systems/audio-zone-system.ts
```diff
@@ -0,0 +1,318 @@
...
+    aabbs.delete(zone);
+    debugObjects.delete(zone);
+
+    const isDebugEnabled = APP.store.state.preferences.showAudioDebugPanel;
+    isDebugEnabled && debugObjects.has(zone) && releaseZoneDebugObject(APP.world, zone);
+  });
+
+  const listener = APP.audioListener.eid!;
+  const listenersQuery = audioZoneListenerQuery(world);
+  const sourcesQuery = audioZoneSourceQuery(world);
```
johnshaughnessy, line 293: Suggestion: Rename `sourcesQuery` -> `audioEmitters`. 

Reason: `sourcesQuery` is not a query -- it's a list of entity ID's (of entities with `AudioEmitter` components).

src/components/audio-params.ts
```diff
@@ -28,16 +28,16 @@ export enum PanningModelType {
...
+  audioType?: AudioType;
+  distanceModel?: DistanceModelType;
+  panningModel?: PanningModelType;
+  rolloffFactor?: number;
+  refDistance?: number;
+  maxDistance?: number;
+  coneInnerAngle?: number;
+  coneOuterAngle?: number;
+  coneOuterGain?: number;
+  gain?: number;
```
johnshaughnessy, line 40: I don't think these properties should be optional. 

If you made these optional so that each property of the `params` sent to `inflateAudioParams` can be optional, then those `params` should be a different type. 

Dom mentioned some typescript utilities that will let us handle this without duplicating a bunch of lines (and risking mistakes): https://github.com/mozilla/hubs/pull/5933#discussion_r1105232292

src/bit-systems/audio-zone-system.ts
```diff
@@ -0,0 +1,318 @@
...
+  if (isInZone && !wasInZone) {
+    zones.add(zone);
+  } else if (!isInZone && wasInZone) {
+    zones.delete(zone);
+  }
+};
+
+const currZones = new Map<number, Set<number>>();
+const prevZones = new Map<number, Set<number>>();
+const aabbs = new Map<number, Box3>();
```
johnshaughnessy, line 248: It seems like this is (non-numeric) component data that we can store using the standard "escape hatch".

johnshaughnessy (changes requested):
The audio zone system seems mostly correct. I left some suggestions, mostly around naming things."
 
I don't think we need to do this now, but I suspect that as we have more interactions triggered by collision volumes ("zones"), we will pretty quickly want to standardize how we check collisions (with masks, etc). In other words, `AudioEmitter`s should be `Collidable`s (on the "audio emitter" layer). Audio zones are just `CollisionVolumes` that collide with the "audio emitter" layer. We can probably also avoid using `Set`s of entity ids by reversing the relationship: `Collidables` would have a `zoneCollisionMask` which is a bitmask. Each `CollisionVolume` would be given an id and we'd set the `zoneCollisions` with `1 << zoneId`. This would limit the number of CollisionVolumes we had (e.g. 64 if `zoneCollisions` was a `ui64`), but this also seems like an easy limitation to remove if we needed more zones.

It seems like an `AudioParams` bitecs component was introduced, but doesn't seem to be used (and would duplicate data we have elsewhere, if it is). We should remove it if that's the case.

There's some fishy business happening with the maps where we're storing extra component data and with the (singleton) `AudioListener` that we should try to sort out. 

I haven't had a chance to read the audio debug system yet.

src/bit-systems/audio-zone-system.ts
```diff
@@ -0,0 +1,318 @@
...
+      }
+    });
+    if (!isDebugEnabled && debugMaterial) {
+      disposeMaterial(debugMaterial);
+      debugMaterial = null;
+    }
+  });
+  const isDebugEnabled = APP.store.state.preferences.showAudioDebugPanel;
+  if (isDebugEnabled) createMaterial();
+});
```
johnshaughnessy, line 45: I think this block can be moved to a regular system. It doesn't need the to use `getScene` or listen to `statechange`s from the store:

```ts
 let isShowing = false;
 function audioDebugPanelToggleSystem(world: HubsWorld) {
   const shouldShow = APP.store.state.preferences.showAudioDebugPanel;
   if (isShowing && !shouldShow) {
     // teardown
   } else if (!isShowing && shouldShow){
     // setup
   }
   isShowing = shouldShow;
 }
 ```

File: src/bit-components.js (hunk 5 above)
Comment (johnshaughnessy) [reply src/bit-components.js:254]: These are the same as bitecs components. We just didn't have bitecs available to us when we wrote them.

File: src/bit-systems/audio-zone-system.ts (hunk 1 above)
Comment (keianhzo) [reply src/bit-systems/audio-zone-system.ts:256]: Makes sense to consolidate to emitter everywhere so I'm going to change all strings to refect that. And I agree on the styling guide. We might benefit from having a more structured naming convention for things like:
- Component names (when to include "Tag" or other Pre/Suffixes)
- Queries names (enter[Component]Query, [Component]enterQuery, enter[Component], etc)
- Eid naming in queries, loops, variables, etc. No we use just eid everywhere, I tried using [component]Eid as much as possible to avoid confusion. Maybe other naming conventions.
- Inflators naming seem to be more consistent.

File: src/bit-systems/audio-zone-system.ts (hunk 8 above)
Comment (keianhzo) [reply src/bit-systems/audio-zone-system.ts:263]: Oops thanks for catching that. I've moved it to a later stage.

File: src/bit-systems/audio-zone-system.ts (hunk 10 above)
Comment (keianhzo) [reply src/bit-systems/audio-zone-system.ts:293]: Makes sense.

File: src/bit-systems/audio-zone-system.ts (hunk 2 above)
Comment (keianhzo) [reply src/bit-systems/audio-zone-system.ts:292]: The reason was more to keep track of the `AudioListener` in case it's removed or replaced in the scene. I don't thing there is any situation right now when we update the audio listener so maybe I was anticipating to something that might never happen.

File: src/inflators/audio-params.ts (hunk 4 above)
Comment (keianhzo) [reply src/inflators/audio-params.ts:10]: I agree, I think this is code I moved at an early stage and never updated. It's worth noting that I added commit after the initial review that adds support for updating the audio params in media video using a `MediaLoaded` state component to set them on the actual audio entity as this inflator does not yet know about it at this stage. It's not ideal as we need to duplicate the audio params in `audioOverrides` but so far it's what I came up with.

File: src/bit-components.js (hunk 5 above)
Comment (keianhzo) [reply src/bit-components.js:257]: Yep, removed.

File: src/bit-systems/audio-zone-system.ts (hunk 12 above)
Comment (keianhzo) [reply src/bit-systems/audio-zone-system.ts:248]: I would like to know more about what's the benefit of following the escape hatch pattern vs having module level maps. If I was an external client developer what would following this pattern buy me in this case? At first glance the benefit is having a entry point for all component bound data structures and I think that's a good thing. In this case this maps are storing transient data that don't have much interest for external systems. Is this assumption right or am I missing the point?

File: src/bit-systems/audio-zone-system.ts (hunk 13 above)
Comment (keianhzo) [reply src/bit-systems/audio-zone-system.ts:45]: IIUC this would make us repeat this same code in any system where we want to read this preferences parameter (ie. we also need to do this in the audio debug system) or centralize it in a system that calls functions on other systems explicitly.

An alternative to this would be to move preferences to bitecs where we have a `[Xyz]Preference` components, then any system interested could just query enter/exit on the prefference they are monitoring and then access it through the store so you only query when it has actually changed.

File: src/bit-systems/audio-zone-system.ts (hunk 6 above)
Comment (keianhzo) [reply src/bit-systems/audio-zone-system.ts:259]: Updated

File: src/bit-systems/audio-zone-system.ts (hunk 7 above)
Comment (keianhzo) [reply src/bit-systems/audio-zone-system.ts:304]: Updated

File: src/components/audio-params.ts (hunk 11 above)
Comment (keianhzo) [reply src/components/audio-params.ts:40]: I understand how `Partial` works but I'm not sure why here is the correct option in this case. What's the difference in this case of making all properties optional or using `Partial<AudioSettings>`? Those settings can be optional everywhere so this just adds the need of making it `Partial` everywhere where we use `AudioSettings`.

File: src/bit-systems/audio-zone-system.ts (hunk 9 above)
Comment (keianhzo) [reply src/bit-systems/audio-zone-system.ts:272]: In current approach each entity keeps track of what zones it's in, so `currZones` and `prevZones` both have `entityEid->Set<zoneEid>` associations. Both the audio emitter and the audio listener are audio zone entities but we keep the listener as a special case because we handle things like the world pos calculation in a different way because of the different nature of the entity.

Conceptually both the `AudioListener` and the `AudioEmitter` components should also be `AudioZoneEntity` but it didn't make sense to add a tag when you can handle all of them in an homogenous way.

Comment (johnshaughnessy) [reply src/bit-systems/audio-zone-system.ts:277]: Ah, right. My mistake. I misread `audioZoneSourceEnterQuery` as `audioZoneEnterQuery` and assumed it was `EntityID`s of `AudioZone`s

File: src/components/audio-params.ts (hunk 11 above)
Comment (johnshaughnessy) [reply src/components/audio-params.ts:40]: Defaults should not be `Partial`, and the reducer in `update-audio-settings` should end up with something that is also not a `Partial`

File: src/bit-systems/video-system.ts [hunk 14]
```diff
@@ -1,126 +1,69 @@
...
+    APP.audioOverrides.delete(rootEid);
+    video2audio.delete(videoEid);
+  });
+  mediaPlaybackUpdated(world).forEach(videoEid => {
+    const audioEid = video2audio.get(videoEid)!;
+    const videoEl = (world.eid2obj.get(videoEid) as any).material.map.image as HTMLVideoElement;
+    if (videoEl.paused) {
+      APP.isAudioPaused.add(audioEid);
+    } else {
+      APP.isAudioPaused.delete(audioEid);
```
johnshaughnessy, line 63: Whenever we change any of the audio settings maps on APP, we have to call updateAudioSettings, or else the changes won't be flushed to the audios.

src/bit-systems/audio-zone-system.ts
```diff
@@ -0,0 +1,326 @@
...
+  const audio = APP.audios.get(emitterEid);
+  if (audio) {
+    updateAudioSettings(emitterEid, audio);
+  }
+};
+
+const restoreEmitterParams = (emitterEid: number) => {
+  APP.zoneOverrides.delete(emitterEid);
+  const audio = APP.audios.get(emitterEid);
+  if (audio) {
```
johnshaughnessy, line 125: If there is no audio, is it a bug?

src/bit-systems/video-system.ts
```diff
@@ -1,126 +1,69 @@
...
+  mediaUnloadedQuery(world).forEach(videoEid => {
+    const audioEid = video2audio.get(videoEid)!;
+    APP.audioOverrides.delete(audioEid);
+  });
+  mediaVideoExitQuery(world).forEach(videoEid => {
+    const rootEid = MediaLoaded.rootRef[videoEid];
+    APP.audioOverrides.delete(rootEid);
+    video2audio.delete(videoEid);
+  });
+  mediaPlaybackUpdated(world).forEach(videoEid => {
```
johnshaughnessy, line 57: Rather than adding the component, querying for it, then changing APP.isAudioPaused, we could do the last part directly where we are toggling the video. APP.isAudioPaused is basically a "proto-tag-component" anyway

src/bit-systems/video-system.ts
keianhzo: `isAudioPaused` doesn't affect the audio node settings so `updateAudioSettings` it's not necessary but this is confusing because and it might not be easy to know when calling `updateAudioSettings` it's required. This raises a good point. We have a few state stores (`isAudioPaused`, `clippingState`, `mutedState`, etc)  that we might be able to replace for flags in the `AudioEmitter` component. That would move that state to the right place and remove a bit of complexity on the maps. So I'm going to do that for `isAudioPaused` and `mutedState` which are the ones that we are touching now and move the rest as I migrate other parts.

src/bit-systems/audio-zone-system.ts
keianhzo: Yeah this is called as part of a query so if the audio eid is there the audio element should be in the store too, otherwise there is a bug somewhere else.

src/bit-systems/video-system.ts
keianhzo: I think the second part of your comment is related to what I mentioned in my previous comment. Regarding the first part, the main reason to send that "event" is that there is a `video2audio` association map that lives in the video system. But I think this is a good application of the escape hatch pattern. This would live in a `MediaVideo` component map and the we can get the `AudioEmitter` and update the flags from anywhere. I'm going to push that.

src/bit-systems/audio-emitter-system.ts
```diff
@@ -0,0 +1,113 @@
...
+
+function swapAudioType<T extends AudioObject3D>(
+  world: HubsWorld,
+  audioSystem: AudioSystem,
+  eid: number,
+  NewType: AudioConstructor<T>
+) {
+  const audio = world.eid2obj.get(eid)! as AudioObject3D;
+  audio.disconnect();
+  APP.gainMultipliers.delete(eid);
```
johnshaughnessy, line 44: I don't think `gainMultipliers` should be deleted when swapping audio types.

src/bit-systems/audio-emitter-system.ts
```diff
@@ -0,0 +1,113 @@
...
+  APP.sourceType.set(eid, SourceType.MEDIA_VIDEO);
+  APP.supplementaryAttenuation.delete(eid);
+  APP.audioOverrides.delete(eid);
```
johnshaughnessy, line 47: I don't think `audioOverrides` should be deleted when swapping audio types, should they?

src/bit-systems/audio-emitter-system.ts
```diff
@@ -0,0 +1,113 @@
...
+
+  const newAudio = new NewType(APP.audioListener);
+  newAudio.setNodeSource(audio.source!);
+  audioSystem.addAudio({ sourceType: SourceType.MEDIA_VIDEO, node: newAudio });
+  APP.audios.set(eid, newAudio);
+
+  audio.parent!.add(newAudio);
+  audio.removeFromParent();
+
+  swapObject3DComponent(world, eid, newAudio);
```
johnshaughnessy, line 59: I know this is not new, but this the main complexity we incure by using three.js audio / positional audio nodes instead of using the webaudio api directly. I believe this is the only place in the app where we swap an entity id from one object3D to another. It would be good to revisit this (not in this PR).

src/bit-systems/audio-emitter-system.ts
```diff
@@ -0,0 +1,113 @@
...
+    const isPannerNode = isPositionalAudio(audio);
+
+    // TODO this needs more testing
+    if (!isPannerNode && settings.audioType === AudioType.PannerNode) {
+      swapAudioType(world, audioSystem, eid, PositionalAudio);
+    } else if (isPannerNode && settings.audioType === AudioType.Stereo) {
+      swapAudioType(world, audioSystem, eid, StereoAudio);
+    }
+
+    applySettings(audio, settings);
```
johnshaughnessy, line 110: If the settings can change during `swapAudioType`, then we should `getCurrentAudioSettings` after we (maybe) swap.

src/bit-systems/audio-zone-system.ts
```diff
@@ -0,0 +1,325 @@
...
+    debugMaterial = new MeshBasicMaterial({
+      color: DEBUG_BBAA_COLOR,
+      transparent: true,
+      opacity: 0.25,
+      side: DoubleSide
+    });
+  }
+};
+
+getScene().then(() => {
```
johnshaughnessy, line 27: I don't think it's correct to use `getScene` as a hook like this. Strong evidence (to me) is that you aren't _using_ the scene returned by `getScene`. (`getScene` is already a hack that we should find a way to remove -- We know when the scene exists, so if there's work that needs to happen after the scene exists, then we should schedule the work then.) 



If you need to do some async work before rendering begins, you can use `preload`. But this doesn't seem to be async. It seems like it can happen as soon as `world` and `store` exist. If that's the case, you can wrap this in a function:

```ts
const audioZoneQuery = defineQuery([AudioZone]);
export function maybeToggleDebugPanel(world: HubsWorld, store: Store) {
    const isDebugEnabled = store.state.preferences.showAudioDebugPanel;
    if (isDebugEnabled) createMaterial();
    audioZoneQuery(APP.world).forEach(zoneEid => {
      if (isDebugEnabled) {
        !debugObjects.has(zoneEid) && addZoneDebugObject(APP.world, zoneEid);
      } else {
        debugObjects.has(zoneEid) && releaseZoneDebugObject(APP.world, zoneEid);
      }
    });
    if (!isDebugEnabled && debugMaterial) {
      disposeMaterial(debugMaterial);
      debugMaterial = null;
    }
}
```
Then we can call this function whenever we need to (e.g. in hub.js or app.tsx, where you have non-global references to both store and world)
```ts
// As soon as we have store and world...
store.addEventListener("statechanged", () => {
  maybeToggleDebugPanel(world, store);
);
maybeToggleDebugPanel(world, store);
```

src/bit-systems/audio-zone-system.ts
```diff
@@ -0,0 +1,325 @@
...
+
+    const isDebugEnabled = APP.store.state.preferences.showAudioDebugPanel;
+    isDebugEnabled && !debugObjects.has(zoneEid) && addZoneDebugObject(APP.world, zoneEid);
+  });
+  audioZoneExitQuery(world).forEach(zoneEid => {
+    aabbs.delete(zoneEid);
+    debugObjects.delete(zoneEid);
+
+    const isDebugEnabled = APP.store.state.preferences.showAudioDebugPanel;
+    isDebugEnabled && debugObjects.has(zoneEid) && releaseZoneDebugObject(APP.world, zoneEid);
```
johnshaughnessy, line 292: It seems like these lines belong in the audio-debug-system

johnshaughnessy (approved):
The problem you encountered with having `AudioParams` on an entity that is _not_ the `LoadedEntity` is a tricky one. We should keep track of any times where we are running into this problem. I don't know yet if it's something we should just accept or if we will eventually want to revisit the "load this as a child entity" pattern we have for various media types. 

Left some comments about things I recommend changing, but don't want to block your progress so leaving an "approve" review.

johnshaughnessy:
Regarding next steps, which you listed here:

> Next steps on audio migration:
    Migrate audio-target and audio-zone-source
    Migrate audio-gain-system
    Migrate audio-settings-system
    Integrate avatar audio related systems (avatar-audio-source and avatar-volume-controls) in the new bitecs audio systems.
    Integrate linked-media in the new bitecs audio systems.
    Look into bypassing the ThreeJS Audio nodes and handle them by just using the WebAudio API.
    As I make progress in the above tasks I'll be separating the previous APP audio stores into a separated set of stores hanging from AudioEmitter: AudioEmitter.sourceType, AudioEmitter.audioOverrides, etc to keep both AFrame and BitECS systems separated.

These do not all need to be rewritten. For example, (if I understand correctly), the `audio-gain-system` can remain unchanged throughout the whole bitecs migration. Keep in mind that the primary goal right now is just to be able to load scenes and media without aframe. Very little of the audio handling code is coupled with aframe right now. If it isn't coupled, we don't need to change it (for now).

src/bit-systems/audio-emitter-system.ts
keianhzo: Yes, I think that true.

src/bit-systems/audio-emitter-system.ts
keianhzo: Same, I think this should stay intact.

src/bit-systems/audio-emitter-system.ts
keianhzo: Yep, I agree. The goal is to address when I'm do migrating all the services. In but-ecs all the Audio logic should be living in the `audio-emitter-system.ts` so it should be easier to remove that indirection layer when we have migrated all the systems/components.

src/bit-systems/audio-emitter-system.ts
keianhzo: Not sure... I'd rather leave this as is and keep an eye on it rather than doing defensive coding.

src/bit-systems/audio-zone-system.ts
keianhzo: I agree that this is not ideal and what you mention seems reasonable. The only thing that I don't like about that is that it forces the system developer to spread system code around the codebase. I'd like to discuss more the idea that I shared in my previous comment here: https://github.com/mozilla/hubs/pull/5944#discussion_r1109761735

i think I'm going to leave it as is for now and add a TODO to revisit this when we have discussed a better path for preferences updates.

src/bit-systems/audio-zone-system.ts
keianhzo: This are in charge of showing/hiding the audio zone debug objects based on the debug panel preference status when an audio zone is added/removed. In this case we can't rely on an preference status update event, we need to explicitly check.

PR #5949, opened 2023-02-16
Move AudioLoader to bitecs

takahirox:
This PR adds AudioLoader to bitecs.

This PR focuses on just works with minimal change. It may need be rewritten elegantly later.

Rendering audio icon is not supported yet. It is rendered as black for now. Supporting it may require some refactoring and should be in another PR.

### DIFF
--- a/src/utils/load-audio.tsx
+++ b/src/utils/load-audio.tsx
@@ -1,0 +1,34 @@
+/** @jsx createElementEntity */
+import { createElementEntity } from "../utils/jsx-entity";
+import { ProjectionMode } from "./projection-mode";
+import { VideoTexture } from "three";
+import { renderAsEntity } from "../utils/jsx-entity";
+import { loadAudioTexture } from "../utils/load-audio-texture";
+import { HubsWorld } from "../app";
+
+export function* loadAudio(world: HubsWorld, url: string) {
+  const { texture, ratio }: { texture: VideoTexture; ratio: number } = yield loadAudioTexture(url);
+
+  // TODO: VideoTexture.image must be content that be played
+  //       in video-system. And it is also used to render.
+  //       It is audio here so the object will be rendered as
+  //       black. Audio icon must be rendered. Fix this.
+
+  return renderAsEntity(
+    world,
+    <entity
+      name="Audio"
+      networked
+      networkedVideo
+      grabbable={{ cursor: true, hand: false }}
+      // Audio and Video are handled very similarly in 3D scene
+      // so create as video
+      video={{
+        texture,
+        ratio,
+        autoPlay: true,
+        projection: ProjectionMode.FLAT
+      }}
+    ></entity>
+  );
+}
--- a/src/bit-systems/media-loading.ts
+++ b/src/bit-systems/media-loading.ts
@@ -14,6 +14,7 @@ import { loadImage } from "../utils/load-image";
 import { loadModel } from "../utils/load-model";
 import { loadPDF } from "../utils/load-pdf";
 import { loadVideo } from "../utils/load-video";
+import { loadAudio } from "../utils/load-audio";
 import { MediaType, mediaTypeName, resolveMediaInfo } from "../utils/media-utils";
 import { EntityID } from "../utils/networking-types";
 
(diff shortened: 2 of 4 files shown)


src/utils/load-audio.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { loadAudioTexture } from "../utils/load-audio-texture";
+import { HubsWorld } from "../app";
+
+export function* loadAudio(world: HubsWorld, url: string) {
+  const { texture, ratio }: { texture: VideoTexture; ratio: number } = yield loadAudioTexture(url);
+
+  // TODO: VideoTexture.image must be content that be played
+  //       in video-system. And it is also used to render.
+  //       It is audio here so the object will be rendered as
+  //       black. Audio icon must be rendered. Fix this.
```
johnshaughnessy, line 15: If it helps, we have icons in the video menu system that can be used as an example for how to do this. You could add one on each side of the entity below (as children) to display an image.

johnshaughnessy (approved):
This seems to support audio files that are copy/pasted into the room. 

Was this also meant to include code to load audio from gltf/scene files (via a `gltfInflator`)?

src/utils/load-audio.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { ProjectionMode } from "./projection-mode";
+import { VideoTexture } from "three";
+import { renderAsEntity } from "../utils/jsx-entity";
+import { loadAudioTexture } from "../utils/load-audio-texture";
+import { HubsWorld } from "../app";
+
+export function* loadAudio(world: HubsWorld, url: string) {
+  const { texture, ratio }: { texture: VideoTexture; ratio: number } = yield loadAudioTexture(url);
+
+  // TODO: VideoTexture.image must be content that be played
```
netpro2k, line 12: video-system gets the video element off of obj.material.map.image but it doesn't have to, we can store the video element elsewhere so that we don't have to fake the texture like this.

src/utils/load-audio.tsx
netpro2k: a single double sided material would be better. In the old code we use this https://github.com/mozilla/hubs/blob/aaa4584a822545a8ea998c3c14edc62078fb542b/src/components/media-video.js#L30

Ideally we would just do that in an inflator for "audio" and change how we are getting the video element reference so we don't need to do any weird workaround.

src/utils/load-audio.tsx
```diff
@@ -0,0 +1,34 @@
...
+  return renderAsEntity(
+    world,
+    <entity
+      name="Audio"
+      networked
+      networkedVideo
+      grabbable={{ cursor: true, hand: false }}
+      // Audio and Video are handled very similarly in 3D scene
+      // so create as video
+      video={{
```
netpro2k, line 26: Since we want to change the visual representation it might be nicer to have an audio inflator. It could end up using the MediaVideo component if it needs to, but it could create a different mesh with a static image.

netpro2k (commented):
I don't love the workaround for texture, but its shipable. As John mentioned this does not seem to include a loader for the GLTF "audio" component. Similar to others this one should be mapped to something like an "audioLoaderInflator" that expects a src instead of a loaded video element.

src/utils/load-audio.tsx
takahirox: I think media files to play can be stored in component. I didn't have time to test in this cycle. I want to try in next cycle.

src/utils/load-audio.tsx
takahirox: > it could create a different mesh with a static image.

If I'm right, currently the video system is designed that `object.material.map` is used both for rendering and playing media so it's impossible to render a static image for audio. I didn't want to add workaround in the video system or write a separated audio system (because most of the processing would be like the same), and didn't have time to refactor the design. I want to think of properer design/implementation in the next cycle.

takahirox:
> Was this also meant to include code to load audio from gltf/scene files (via a gltfInflator)?

I was just missing it. I added to `gltfInflator`. I think perhaps video loader inflator can be reused for now. I haven't tested it tho because I don't have a test asset. I want to test later.

And in the next (or later) cycle I want to refactor the video system and revisit the audio inflators.

PR #5952, opened 2023-02-17
Fix type issues for simple-water and particle-emitter

keianhzo:
This fixes a bug that I've seen in earlier code of mine. I think the reason the bug was introduced was that after adding support for the glTF inflator I wanted the component to also work for JSX entities and I changed the component parameters data to array obviously breaking the glTF support. I've fixed that.

I've also limited the `simple-water` and `particle-emitter` inflators support only to glTF as I've seen that for this type of components we are not really adding support for JSX (see `UVScroll` component) and adding support for JSX would require adding another inflator which I'm not sure if it's worth i. Is this assumption correct or we want parity?

### DIFF
--- a/src/inflators/particle-emitter.ts
+++ b/src/inflators/particle-emitter.ts
@@ -24,8 +24,16 @@ export type ParticleEmitterParams = {
   lifetime: number;
   lifetimeRandomness: number;
   particleCount: number;
-  startVelocity: [number, number];
-  endVelocity: [number, number, number];
+  startVelocity: {
+    x: number;
+    y: number;
+    z: number;
+  };
+  endVelocity: {
+    x: number;
+    y: number;
+    z: number;
+  };
   velocityCurve: string;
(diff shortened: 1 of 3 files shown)


johnshaughnessy (approved):
> I've also limited the simple-water and particle-emitter inflators support only to glTF as I've seen that for this type of components we are not really adding support for JSX (see UVScroll component) and adding support for JSX would require adding another inflator which I'm not sure if it's worth i. Is this assumption correct or we want parity?

If we don't have an `EntityDef` written in JSX that needs to create a `simpleWater` or `particleEmitter`, then it's fine to only support these components in `gltfInflators`. 

>  adding support for JSX would require adding another inflator 

This is not necessarily true. It depends whether the data passed in via JSX would be the same shape as what is passed in via gltf components. If the shape of the data is the same for both, then you can re-use the sample inflator. If they are different, then you would need separate inflators.

PR #5953, opened 2023-02-17
Add ModelLoader to bitecs

takahirox:
This PR adds ModelLoader inflator for glTF model component.

### DIFF
--- a/src/inflators/model-loader.ts
+++ b/src/inflators/model-loader.ts
@@ -1,0 +1,16 @@
+import { HubsWorld } from "../app";
+import { inflateMediaLoader } from "./media-loader";
+
+export type ModelLoaderParams = {
+  src: string;
+}
+
+export function inflateModelLoader(world: HubsWorld, eid: number, params: ModelLoaderParams) {
+  inflateMediaLoader(world, eid, {
+    src: params.src,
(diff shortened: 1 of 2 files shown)


johnshaughnessy (approved):

src/inflators/model-loader.ts
```diff
@@ -0,0 +1,16 @@
+import { HubsWorld } from "../app";
+import { inflateMediaLoader } from "./media-loader";
+
+export type ModelLoaderParams = {
+  src: string;
+}
+
+export function inflateModelLoader(world: HubsWorld, eid: number, params: ModelLoaderParams) {
+  inflateMediaLoader(world, eid, {
```
netpro2k, line 9: Based on the old defaults https://github.com/mozilla/hubs/blob/aaa4584a822545a8ea998c3c14edc62078fb542b/src/gltf-component-mappings.js#L267-L275 I think resize and recenter should be true. Technically models like this had an object menu too, though it was a restricted one with only "open link", "refrresh" , "inspect', and "focus" - I don't think its particularly important to preserve that.

netpro2k (commented):
This LGTM other than the resize/recenter settings

PR #5955, opened 2023-02-18
Fix "system default" option for media devices

johnshaughnessy:
Prepend a `System Default` option for each `MediaDeviceKind` (`audioinput`, `audiooutput` and `videoinput`).

Also remove unused functions, unnecessary indirection, and duplicate code where possible.


keianhzo (changes requested):
This looks mostly good. The main thing I see is that we are reverting the `System Default` setting to the actual default device in some places including the Mic Setup Modal or the Audio Toolbar.

In the preferences screen everything works as expected because we are saving the pref and calling `startMicShare` with `updatePrefs` set to `false` so we don't overwrite the store.

When calling it from other places like `useMicrophone`, `useMicrophoneStatus` or `ui-root` we en up saving the currently used device in the preferences so we are overwriting it: https://github.com/mozilla/hubs/blob/c03f521c288fa405cabe0b0b390b3139a563227b/src/utils/media-devices-manager.js#L182

We have to save in preferences the currently used `deviceId` not the one coming from the audio track. Also we probably need to update `selectedMicDeviceId` to don't return the audio track based label but the currently stored `deviceId` similarly to how we do it in `selectedSpeakersDeviceId`.

keianhzo:
Thinking about this again, I think the best approach is not move the mic setting responsibility to the media devices manager and just update the preferences from anywhere else. I've made a proposal in this fork: https://github.com/mozilla/hubs/commit/eec77409c993be5deac22f2a72a6eac7af4d2065 Feel free to merge that commit or disregard if you don't feel it's the right thing.

PR #5960, opened 2023-02-22
bitecs audio-target and zone-audio-source

keianhzo:
This builds on top of https://github.com/mozilla/hubs/pull/5944 and adds support for audio-target and zone-audio-source.
- Now audio targets supports any audio emitter (only supported avatars before)
- Avatar audio sources are not yet an Audio emitter so is not yet supported in any audio system including this. This will come in a later PR.
- This PR removes dependency on `APP.audios` for the bit-ecs systems as the audio object is already stored in `APP.world.obj2eid`. More store migration will happen in next PRs as we move bit-ecs systems away from APP audio stores.

If you have reviewed https://github.com/mozilla/hubs/pull/5944 you should only care about the commits after https://github.com/mozilla/hubs/pull/5960/commits/5329bfec3605887414edd817c8b22dd2be748640


src/react-components/debug-panel/AudioDebugPanel.js
```diff
@@ -109,8 +113,14 @@ export function AudioDebugPanel({ isNarrow, collapsed, onCollapsed }) {
     } else {
       setAvatarSettings(getCurrentAudioSettingsForSourceType(SourceType.AVATAR_AUDIO_SOURCE));
     }
-    for (const [el, audio] of APP.audios.entries()) {
-      updateAudioSettings(el, audio);
+    if (qsTruthy("newLoader")) {
```
johnshaughnessy, line 116: Unless I'm misunderstanding things, `"newLoader"` does not mean all audios in the app are loaded with bitecs... For example, avatars will continue to be loaded through the old code paths, and avatar gltfs can have `audios` in them.

Beyond that though, I don't understand the purpose of this change. `updateAudioSettings` already handles whether the `audio` is a bitecs entity or aframe element. Why move that check outside to here?

src/bit-systems/audio-emitter-system.ts
```diff
@@ -23,11 +37,86 @@ export function isPositionalAudio(node: AudioObject3D): node is PositionalAudio
...
+    audio.setRolloffFactor(settings.rolloffFactor);
+    audio.setRefDistance(settings.refDistance);
+    audio.setMaxDistance(settings.maxDistance);
+    audio.panner.panningModel = settings.panningModel;
+    audio.panner.coneInnerAngle = settings.coneInnerAngle;
+    audio.panner.coneOuterAngle = settings.coneOuterAngle;
+    audio.panner.coneOuterGain = settings.coneOuterGain;
+  }
+  audio.gain.gain.setTargetAtTime(settings.gain, audio.context.currentTime, 0.1);
+};
```
johnshaughnessy, line 114: Isn't this code in `update-audio-settings.js`? Why is it duplicated here?

src/react-components/debug-panel/AudioDebugPanel.js
netpro2k: Yep thats right. We are already using the audio system in things like the video component, it should already support both. And yes they will need to co-exist since avatars and the pen tool will continue to be aframe entities even after we make newLoader the default

keianhzo: This PR is closely related to https://github.com/mozilla/hubs/pull/5974

For components like video, we are already adding audio objects to `APP.audios` based on eid but for components that we are not yet migrating like `avatar-audio-source` we still add them by element. New bit-ecs systems like `audio-zones-system` or `audio-debug-system` rely on eids. It looked simpler to me to make `avatar-audio-source` have two paths for the new and old loaders: 
https://github.com/mozilla/hubs/blob/d6118639a6d1f2d3cd4d04e7457bc675bf783f28/src/components/avatar-audio-source.js#L86-L91
than moving them to adding audios to `APP.audios` using el.eid or adding conditional code in the new systems to check for el/eid Vs creating similar methods that operate on eid and just use `APP.world.eid2obj` to get to get the audio objects, which also sounds like a logical next step as it avoids having duplicated references in `APP.audios` and `APP.world.eid2obj`. This also simplifies the migration of  components like `scale-audio-feedback` and `morph-audio-feedback`.

This duplicated code it's just a first step toward that decoupling of the audios stores. Both old and new loaders would work the same but new loader will use in `APP.world.eid2` and eids for all audios stores and the old loader will keep on using elements.

keianhzo: Please not that the links above make reference to the https://github.com/mozilla/hubs/pull/5974 PR which builds on top of this one. I recommend to take a look at that one as part of this review to have the full picture.

src/bit-systems/audio-target-system.ts
```diff
@@ -0,0 +1,206 @@
+import { addComponent, defineQuery, enterQuery, exitQuery } from "bitecs";
+import { Vector3, Object3D, LineSegments, WireframeGeometry, SphereBufferGeometry, PositionalAudio } from "three";
+import { HubsWorld } from "../app";
+import { AudioEmitter, AudioSettingsChanged, AudioSource, AudioTarget } from "../bit-components";
+import { SourceType } from "../components/audio-params";
+import { AUDIO_SOURCE_FLAGS } from "../inflators/audio-source";
+import { AudioSystem } from "../systems/audio-system";
+import { AudioObject3D, Emitter2Audio, EMITTER_FLAGS, makeAudioEntity } from "./audio-emitter-system";
+
+const createWhiteNoise = (audioContext: AudioContext, gain: number): AudioBufferSourceNode => {
```
johnshaughnessy, line 10: This function exists in `avatar-audio-source.js`. If you are moving it here in order to add type info, remove the function from the original file.

src/bit-systems/audio-debug-system.ts
```diff
@@ -168,7 +168,7 @@ export function audioDebugSystem(world: HubsWorld) {
       }
       if (idx >= maxDebugEmitters) return;
 
-      const audio = APP.audios.get(emitterEid)!;
+      const audio = APP.world.eid2obj.get(emitterEid)! as AudioObject3D;
```
johnshaughnessy, line 171: We should _not_ change how we're storing these audios until after the migration. 

`APP.audios` and the rest of the audio state maps have key types of `AEntity | EntityID` because they are used by both code paths. There's no advantage to changing these now, and it will be significantly easier to do (with confidence) after we remove `AEntity`s. Please reverse these changes.

src/react-components/debug-panel/AudioDebugPanel.js
johnshaughnessy: In your PR summary you said,

> This PR removes dependency on `APP.audios` for the bit-ecs systems as the audio object is already stored in `APP.world.obj2eid`. More store migration will happen in next PRs as we move bit-ecs systems away from APP audio stores.

We should not make any changes to the audio stores this until after the migration. 

`APP.audios` and the rest of the audio state maps have key types of `AEntity | EntityID` because they are used by both code paths. There's no advantage to changing these now, and increases the risk that we introduce regressions. This line is an example of such a regression. When `newLoader` is enabled, this code will (incorrectly) ignore any audios that are associated with `AEntity`s. 

It will be significantly easier to change how we store audio data after we remove `AEntity`s, because then the audio state maps will be `Map<EntityID, Foo>`. Until then, please reverse the changes to how audio data is stored.

src/components/audio-params.ts
```diff
@@ -79,4 +79,30 @@ export const TargetAudioDefaults: AudioSettings = {
...
+  rolloffFactor: 5,
+  refDistance: 5,
+  maxDistance: 10000,
+  coneInnerAngle: 180,
+  coneOuterAngle: 360,
+  coneOuterGain: 0.9,
+  gain: 1.0
+};
+
+export const ZoneAudioDefaults: AudioSettings = {
```
johnshaughnessy, line 95: Why were new defaults introduced?

johnshaughnessy (changes requested):

src/react-components/debug-panel/AudioDebugPanel.js
johnshaughnessy: > It looked simpler to me to make avatar-audio-source have two paths for the new and old loaders... than moving them to adding audios to APP.audios using el.eid or adding conditional code in the new systems to check for el/eid Vs creating similar methods that operate on eid and just use APP.world.eid2obj to get to get the audio objects

I don't understand this concern. To get the audio in the old (existing) code, we use `APP.audios.get`. To get the audio in the new code, we can still use `APP.audios.get`. Similarly for all of the other audio related state. That's what will allow us to migrate with very little effort.

>  it avoids having duplicated references in APP.audios and APP.world.eid2obj
This duplicate state goes away trivially once we're done with the migration.

keianhzo: Answering to my own question I think we can avoid bridging the two systems by iterating `APP.adios` instead of using the `AudioEmitter` query. It will require to update the existing bit systems but I think it can work and it will avoid the code duplication for bridging until we migrate the avatar source component

keianhzo:
@johnshaughnessy I've moved the systems that were using `AudioEmitter` queries  to use `APP.audios` and also move back the `audio-target-system` to use the same code as the AFrame system.

src/bit-systems/audio-target-system.ts
```diff
@@ -0,0 +1,209 @@
...
+};
+
+const addSourceToAudioTarget = (audioSourceEid: number, source: AudioNode) => {
+  const audioTargetEids = source2Target.get(audioSourceEid);
+  audioTargetEids?.forEach(audioTargetEid => {
+    const audioEid = Emitter2Audio.get(audioTargetEid)!;
+    const targetAudio = APP.audios.get(audioEid)!;
+    try {
+      targetAudio.disconnect();
+    } catch (e) {}
```
johnshaughnessy, line 35: May want to leave a comment explaining why this try/catch is necessary (and why we don't log the error)

src/bit-systems/audio-target-system.ts
```diff
@@ -0,0 +1,209 @@
...
+};
+
+const removeSourceFromAudioTarget = (audioSourceEid: number) => {
+  const audioTargetEids = source2Target.get(audioSourceEid);
+  audioTargetEids?.forEach(audioTargetEid => {
+    const audioEid = Emitter2Audio.get(audioTargetEid)!;
+    const targetAudio = APP.audios.get(audioEid)!;
+    try {
+      targetAudio.disconnect();
+    } catch (e) {}
```
johnshaughnessy, line 51: Same thing here

johnshaughnessy (approved):

PR #5961, opened 2023-02-23
Home Page Updates

nickgrato:
- Adding "--legacy-peer-deps" to npm ci because lilypad uses React 18
- Converting home page to functional react component and TS
- Adding new content to home page ( icons copy etc)
- Adding style library we use in marketing page and dashboard so all out utility and css vars are available.  ( this is about 25 files, a majority of the PR. This is not brand new SCSS, it is approved code im copying over to be used and edit in the context of the admin.)


admin/types.d.ts
```diff
@@ -0,0 +1,43 @@
+export type ReticulumStorageT = {
+  in_quota: boolean;
+};
+
+export type ReticulumSceneListingT = {
```
nickgrato, line 5: @keianhzo do you know where a good place for me to look to build these types out?

keianhzo: I believe that would be the reticulum source code:
- https://github.com/mozilla/reticulum/blob/master/lib/ret/meta.ex
- https://github.com/mozilla/reticulum/blob/master/lib/ret/avatar_listing.ex
- https://github.com/mozilla/reticulum/blob/master/lib/ret/scene_listing.ex
- https://github.com/mozilla/reticulum/blob/master/lib/ret/account.ex
- https://github.com/mozilla/reticulum/blob/master/lib/ret/storage.ex

nickgrato: amazing thank you

nickgrato: @keianhzo ok I think this is ready to review when you get a chance - it's a big departure from the current way we do things here so let me know if you want me to walk through my future plans if something seems odd.

keianhzo (approved):
Nice to see how we are consolidating things here and in the dashboard. Looking good. Just left a few nits, mostly things that we were already doing in a weird way before so I don't want to block on this so take them as suggestions in case you want to look into them now or prefer to let them for the future.

admin/src/react-components/pages/system-editor.tsx
```diff
@@ -0,0 +1,348 @@
...
+   * - Needs Avatars
+   * - Needs Scenes
+   * - Is in a Sandbox
+   * - is Using cloud flate
+   */
+  const needsAvatars = reticulumMeta.repo && !reticulumMeta.repo.avatar_listings.any;
+  const needsScenes = reticulumMeta.repo && !reticulumMeta.repo.scene_listings.any;
+  const exceededStorageQuota = reticulumMeta.repo && !reticulumMeta.repo.storage.in_quota;
+  const isInSESSandbox =
+    adminInfo && adminInfo.using_ses && adminInfo.ses_max_24_hour_send <= MAX_AWS_SES_QUOTA_FOR_SANDBOX;
```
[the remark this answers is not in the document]
nickgrato: check what types are optional for this.

admin/types.d.ts
```diff
@@ -0,0 +1,41 @@
+export type ReticulumMetaT = {
```
nickgrato, line 1: @bryanenders @netpro2k Was hoping to get some insight on if any of these types are conditional so I can throw the "?" on them or if we think they are all expected every time?

admin/src/react-components/pages/system-editor.tsx
```diff
@@ -0,0 +1,348 @@
...
+  const isUsingCloudflare =
+    adminInfo &&
+    retConfig &&
+    retConfig.phx &&
+    retConfig.phx.cors_proxy_url_host === `cors-proxy.${adminInfo.worker_domain}`;
+
+  return (
```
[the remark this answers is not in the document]
nickgrato: putting a todo note for this!

PR #5962, opened 2023-02-23
Add Dash video play support to bitecs

takahirox:
This PR adds Dash video play support to bitecs.

The main changes are
* Make `createDashPlayer()` util function
* Add `DashVideoTexture` which holds Dash video player and dispose it when the texture is disposed
  * Currently player is saved as `texture.dash` but I don't think polluting texture is good. Making a new class may be cleaner
* `loadVideoTexture()` creates `VideoTexture` or `DashVideoTexture` depending on contentType

### DIFF
--- a/src/textures/DashVideoTexture.ts
+++ b/src/textures/DashVideoTexture.ts
@@ -34,18 +34,14 @@ export class DashVideoTexture extends VideoTexture {
   ) { 
     super(video, mapping, wrapS, wrapT, magFilter, minFilter, format, type, anisotropy);
     this.isDashVideoTexture = true;
-    this.player = null; // Set via setPlayer()
-  }
-
-  setPlayer(player: MediaPlayer) : DashVideoTexture {
-    this.player = player;
-    return this;
+    this.player = null; // Set by user later
   }
 
   dispose() {
     super.dispose();
     if (this.player !== null) {
       this.player.reset();
+      this.player = null;
     }
   }
 }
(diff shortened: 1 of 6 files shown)


src/textures/DashVideoTexture.ts
```diff
@@ -0,0 +1,51 @@
...
+    format?: PixelFormat,
+    type?: TextureDataType,
+    anisotropy?: number
+  ) { 
+    super(video, mapping, wrapS, wrapT, magFilter, minFilter, format, type, anisotropy);
+    this.isDashVideoTexture = true;
+    this.player = null; // Set via setPlayer()
+  }
+
+  setPlayer(player: MediaPlayer) : DashVideoTexture {
```
netpro2k, line 40: I don't see an advantage to having this setter vs setting `player` directly.

src/utils/media-utils.js
```diff
@@ -10,6 +10,7 @@ import { isIOS as detectIOS } from "./is-mobile";
 import Linkify from "linkify-it";
 import tlds from "tlds";
 import { mediaTypeFor } from "./media-type";
+import { MediaPlayer } from "dashjs";
```
netpro2k, line 13: Since we are going to be re-testing this it might be a good time to update this dependency, its probably quite out of date.

netpro2k (commented):
made a super minor styling note, and we might want to use this as a chance to update dashjs since we will need to do testing anyway. Other than that LGTM

src/utils/media-utils.js
takahirox: I want to test it in another PR (in next cycle?) because of it is at very last minute in this cycle now.

PR #5966, opened 2023-02-23
Add HLS video play support to bitecs

takahirox:
This PR adds HLS video play support to bitecs.

The main changes are

* Make `createHLSPlayer()` util function
* Add `HLSVideoTexture` which holds HLS video player and cleans up it when the texture is disposed
* `loadVideoTexture()` creates `HLSVideoTexture` if video is HLS


netpro2k (commented):
Some comments as DASH PR, I don't think we need the setter, and this is probably a good time to bump hlsjs since we need to do testing anyway.

PR #5969, opened 2023-02-24
Take soft ownership for video with bitecs

takahirox:
Currently Hubs Client doesn't try to take soft ownwership for videos with bitecs so even if you upload a video you don't take ownership and the video doesn't play correctly.

With this commit the client tries to take soft ownership and the problem will be resolved.

### DIFF
--- a/src/bit-systems/video-system.ts
+++ b/src/bit-systems/video-system.ts
@@ -1,17 +1,19 @@
 import { addComponent, defineQuery, enterQuery, exitQuery, hasComponent } from "bitecs";
 import { Mesh, MeshStandardMaterial } from "three";
 import { HubsWorld } from "../app";
-import { AudioParams, AudioSettingsChanged, MediaLoaded, MediaVideo, NetworkedVideo, Owned } from "../bit-components";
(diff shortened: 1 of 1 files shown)


src/bit-systems/video-system.ts
```diff
@@ -100,7 +101,8 @@ function audioEmitterSystem(world: HubsWorld, audioSystem: AudioSystem) {
 }
 
 const OUT_OF_SYNC_SEC = 5;
-const networkedVideoQuery = defineQuery([NetworkedVideo]);
+const networkedVideoQuery = defineQuery([Networked, NetworkedVideo]);
+const networkedVideoEntryQuery = enterQuery(networkedVideoQuery);
```
johnshaughnessy, line 105: Suggestion: Rename `...EntryQuery` -> `...EnterQuery`

johnshaughnessy (approved):

netpro2k (commented):
I still have a slight feeling that we are missing something but if actually became this simple thats great, since this was one of the more complex and error prone things in the old networking system.

src/bit-systems/video-system.ts
takahirox: OK, updated

PR #5976, opened 2023-03-02
Add Link to bitecs

takahirox:
This PR adds Link to bitecs

image

### DIFF
--- a/src/bit-components.js
+++ b/src/bit-components.js
@@ -198,7 +198,7 @@ export const ObjectMenu = defineComponent({
 });
 // TODO: Store this data elsewhere, since only one or two will ever exist.
 export const LinkHoverMenu = defineComponent({
-  targetObject: Types.eid,
+  targetObjectRef: Types.eid,
   linkButtonRef: Types.eid
 });
 export const LinkHoverMenuItem = defineComponent();
--- a/src/bit-components.js
+++ b/src/bit-components.js
@@ -197,6 +197,16 @@ export const ObjectMenu = defineComponent({
   targetRef: Types.eid
 });
 // TODO: Store this data elsewhere, since only one or two will ever exist.
+export const LinkHoverMenu = defineComponent({
+  targetObject: Types.eid,
+  linkButtonRef: Types.eid
+});
+export const LinkHoverMenuItem = defineComponent();
+export const Link = defineComponent({
+  url: Types.ui32
+});
+Link.url[$isStringType] = true;
+// TODO: Store this data elsewhere, since only one or two will ever exist.
 export const PDFMenu = defineComponent({
   prevButtonRef: Types.eid,
   nextButtonRef: Types.eid,
--- a/src/bit-systems/link-hover-menu.ts
+++ b/src/bit-systems/link-hover-menu.ts
@@ -1,0 +1,51 @@
+import { defineQuery, enterQuery, entityExists } from "bitecs";
+import type { HubsWorld } from "../app";
+import { Interacted, Link, LinkHoverMenu, LinkHoverMenuItem, HoveredRemoteRight } from "../bit-components";
+import { anyEntityWith } from "../utils/bit-utils";
+
+const NULL_EID = 0;
+
+const hoveredQuery = defineQuery([HoveredRemoteRight, Link]);
+const hoveredEnterQuery = enterQuery(hoveredQuery);
+const hoveredMenuItemQuery = defineQuery([HoveredRemoteRight, LinkHoverMenuItem]);
+const clickedMenuItemQuery = defineQuery([Interacted, LinkHoverMenuItem]);
+
+export function linkHoverMenuSystem(world: HubsWorld) {
+  // Assumes always only single LinkHoverMenu entity exists.
+  const menuEid = anyEntityWith(world, LinkHoverMenu)!;
+  const menuObject = world.eid2obj.get(menuEid)!;
+
+  // Save Link object eid in LinkHoverMenu when hovered.
+  hoveredEnterQuery(world).forEach((eid: number) => {
+    LinkHoverMenu.targetObject[menuEid] = eid;
+    const targetObject = world.eid2obj.get(eid)!;
+    targetObject.add(menuObject);
+  });
+
+  // Check if the cursor it hovered on Link object or Link menu button object.
+  const hovered = hoveredQuery(world).length > 0 || hoveredMenuItemQuery(world).length > 0;
+
+  if (hovered) {
+    // Hovered then make the menu visible and handle clicks if needed.
+    menuObject.visible = true;
+    if (clickedMenuItemQuery(world).length > 0) {
+      const linkEid = LinkHoverMenu.targetObject[menuEid];
+      if (linkEid !== NULL_EID) {
+        // Just in case
+        if (entityExists(world, linkEid)) {
+          window.open(APP.getString(Link.url[linkEid])!);
+        } else {
+          // Link object seems to have been already removed so forget it.
+          LinkHoverMenu.targetObject[menuEid] = NULL_EID;
+        }
+      }
+    }
+  } else {
+    // Not hovered then make the menu invisible and forget the Link object.
+    menuObject.visible = false;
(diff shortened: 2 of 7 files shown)


src/bit-systems/link-hover-menu.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  // Save Link object eid in LinkHoverMenu when hovered.
+  hoveredEnterQuery(world).forEach((eid: number) => {
+    LinkHoverMenu.targetObject[menuEid] = eid;
+    const targetObject = world.eid2obj.get(eid)!;
+    targetObject.add(menuObject);
+  });
+
+  // Check if the cursor it hovered on Link object or Link menu button object.
+  const hovered = hoveredQuery(world).length > 0 || hoveredMenuItemQuery(world).length > 0;
```
netpro2k, line 26: We have now implemented something similar to this in 3 different spots in 3 slightly different ways (video menu, pdf menu, and this link menu). This is probably ok for burndown but we should really think about the general pattern here. In this and video menu we tag the sub items with a component, in PDFMenu we check if the hovered thing is itself or is a child of an interesting object. I think I like the second one better design wise, but walking the hierarchy every frame sounds slightly concerning perf wise.

src/bit-systems/link-hover-menu.ts
```diff
@@ -0,0 +1,51 @@
...
+
+const NULL_EID = 0;
+
+const hoveredQuery = defineQuery([HoveredRemoteRight, Link]);
+const hoveredEnterQuery = enterQuery(hoveredQuery);
+const hoveredMenuItemQuery = defineQuery([HoveredRemoteRight, LinkHoverMenuItem]);
+const clickedMenuItemQuery = defineQuery([Interacted, LinkHoverMenuItem]);
+
+export function linkHoverMenuSystem(world: HubsWorld) {
+  // Assumes always only single LinkHoverMenu entity exists.
```
netpro2k, line 14: For VR we are probably going to need to account for more than one, though maybe we can get away with only allowing a single cursor and just moving which hand it comes from.

src/bit-components.js
```diff
@@ -194,6 +194,16 @@ export const ObjectMenu = defineComponent({
   targetRef: Types.eid
 });
 // TODO: Store this data elsewhere, since only one or two will ever exist.
+export const LinkHoverMenu = defineComponent({
+  targetObject: Types.eid,
```
netpro2k, line 198: We have been using `ref` like `targetRef` for entity reference properties.

src/bit-systems/link-hover-menu.ts
```diff
@@ -0,0 +1,51 @@
...
+        if (entityExists(world, linkEid)) {
+          window.open(APP.getString(Link.url[linkEid])!);
+        } else {
+          // Link object seems to have been already removed so forget it.
+          LinkHoverMenu.targetObject[menuEid] = NULL_EID;
+        }
+      }
+    }
+  } else {
+    // Not hovered then make the menu invisible and forget the Link object.
```
netpro2k, line 44: I think if we move this condition up, and also check for `!entityExists(LinkHoverMenu.targetObject[menuEid])` we can get rid of the defensive checks around that above and the control flow will be a bit more clear.

src/bit-systems/link-hover-menu.ts
```diff
@@ -0,0 +1,51 @@
...
+  // Check if the cursor it hovered on Link object or Link menu button object.
+  const hovered = hoveredQuery(world).length > 0 || hoveredMenuItemQuery(world).length > 0;
+
+  if (hovered) {
+    // Hovered then make the menu visible and handle clicks if needed.
+    menuObject.visible = true;
+    if (clickedMenuItemQuery(world).length > 0) {
+      const linkEid = LinkHoverMenu.targetObject[menuEid];
+      if (linkEid !== NULL_EID) {
+        // Just in case
```
netpro2k, line 34: I don't like that we got as far as being able to click on something for an object that was deleted. Ideally we can clean up the control flow so that when the target of this menu is deleted we wouldn't even get here.

src/bit-systems/link-hover-menu.ts
```diff
@@ -0,0 +1,51 @@
...
+        if (entityExists(world, linkEid)) {
```
netpro2k, line 35: Same as above, we don't want to be checking this, the control flow should be such that we didn't even make it this far.

src/bit-systems/link-hover-menu.ts
```diff
@@ -0,0 +1,51 @@
...
+          window.open(APP.getString(Link.url[linkEid])!);
```
netpro2k, line 36: This accounts for the simple case of an external url but the current link component also supports linking to anchors in the same room, switching scenes, and switching avatars. See https://github.com/mozilla/hubs/blob/aaa4584a822545a8ea998c3c14edc62078fb542b/src/components/open-media-button.js#L43-L75 - Those should ideally also be reflected in the label.

src/prefabs/link-hover-menu.tsx
```diff
@@ -0,0 +1,42 @@
...
+const BUTTON_WIDTH = 0.6;
+const BUTTON_SCALE: ArrayVec3 = [1.0, 1.0, 1.0];
+const UI_Z = 0.001;
+const BUTTON_POSITION: ArrayVec3 = [0, 0, UI_Z];
+
+interface LinkButtonAttrs extends Attrs {
+  linkHoverMenuItem: boolean;
+}
+
+function LinkButton(props: LinkButtonAttrs) {
```
netpro2k, line 15: I don't feel super strongly about this, and we have done it in some other spots, but I personally wouldn't make a component (in the JSX sense) for this, reads just as clearly with the Button3D inlined. I know some people feel the opposite though.

netpro2k (commented):
This is looking pretty good. I think the control flow around the hovered entity being removed feels a bit confusing to follow and defensive. One of the main advantages we get in our new system is that execution order should be very predictable, so we don't have to second guess about "when might this happen". I think just slightly rearranging the order of checks will make it so that you can never get into the blocks where you have some of the defensive checks. Also, the old link component supported linking to rooms, scenes, avatars, and waypoints with different labels and handling for each, I think we need to preserve that behavior.

src/bit-systems/link-hover-menu.ts
takahirox: Do you think it's ok to support them in another PR in this or later cycle because I'm not really sure how difficult they are yet. I hope I can push the very basic function to the repository for now.

src/prefabs/link-hover-menu.tsx
takahirox: I want to go as is for now, and want to revisit in next or later cycle.

src/bit-systems/link-hover-menu.ts
takahirox: > We have now implemented something similar to this in 3 different spots in 3 slightly different ways

Yes, I noticed it. I focused on making it just work in this cycle. In next or later cycle, I hope we can work more on interactive stuffs and update the code to handle more consistently.

src/bit-systems/link-hover-menu.ts
takahirox: Yes, true. I hope we can work on more proper interactive flow in next cycle.

src/bit-systems/link-hover-menu.ts
takahirox: I will add TODO comment for now

src/bit-systems/link-hover-menu.ts
takahirox: I will add TODO comment for now

src/bit-systems/link-hover-menu.ts
takahirox: What I have been thinking is placing invisible plane or box and showing/hiding the menu depending on whether it is raycasted. It can avoid the hierarchy traverse. And there may be gaps between buttons or button and object. Even if curson points to the gaps the menu keeps appearing. Perhaps it is what users expect.

PR #5977, opened 2023-03-03
Add Media loading HTML support to bitecs

takahirox:
This PR adds media loading HTML support to bitecs.

"Open Link" support requires #5976 so draft now.

image

### DIFF
--- a/src/bit-systems/media-loading.ts
+++ b/src/bit-systems/media-loading.ts
@@ -24,6 +24,7 @@ import { loadModel } from "../utils/load-model";
 import { loadPDF } from "../utils/load-pdf";
 import { loadVideo } from "../utils/load-video";
 import { loadAudio } from "../utils/load-audio";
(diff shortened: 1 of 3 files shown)


takahirox:
Switched to "ready for review" because #5976 has been merged.

keianhzo (commented):
This LGTM but I expected the Link hover menu to work after https://github.com/mozilla/hubs/pull/5976 has landed. Is it missing?

takahirox:
I thought it would work but, right, it doesn't... Let me investigate.

takahirox:
I noticed that I had forgotten to add `link` inflator key to `EntityDef`. But if I add it, the object won't move by cursor. `RemoteHoverTarget` component seems to be the cause.

So I didn't add `link`, instead I added a TODO comment for it for now. I hope we can investigate and resolve the root issue in the follow up PRs.

Issue #5608, opened 2022-07-22
Animations stop working when using "Enable Fast Room Switching"


scrappymeals:
**Description**
The animations stop playing when using fast room switching between scenes.

**To Reproduce**
create a simple test scene with a Hubs link component that links to another test scene. These scenes should have some kind of looping animation. Enable fast room switching and see if the animations play after switching back and forth.

1. Go to 'Preferences...'
2. Click on 'Misc....'
3. Scroll down to 'Fast Room Loading....'
4. Enable and return to space.

**Expected behavior**
The animations should still be playing jumping between rooms. Currently they stop playing. The only way to get them to work is to refresh the scene and browser, which is not a viable solution.

**Screenshots**

**Hardware** 
 - Device:  MSI Gaming Laptop
 - OS: Windows 10
 - Browser: Chrome

**Additional context**

matthewbcool:
Thank you for this report! Good catch.

emclaren:
Fast room switching is still experimental, we would definitely like to take a look at this and a few other fast room switching bugs to improve the experience in the future.

PR #5981, opened 2023-03-04
Enable fast room switching by default

johnshaughnessy:


steDalessio:
Fast room switching by default leads to always having stuck animation https://github.com/mozilla/hubs/issues/5608

PR #5984, opened 2023-03-07
changing image sizes

nickgrato:
updating logo size in nav and modal.

### DIFF
--- a/src/react-components/layout/Header.scss
+++ b/src/react-components/layout/Header.scss
@@ -26,12 +26,6 @@ header {
     li {
       &:first-child {
         padding-top: 20px;
-        @media(min-width: theme.$breakpoint-lg) {
-          img {
-            width: 100%;
-            height: 100%;
-          }
-        }
       }
       padding-top: 20px;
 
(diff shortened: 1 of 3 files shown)


src/react-components/layout/Header.scss
```diff
@@ -77,11 +71,10 @@ header {
...
   width: 124px;
-  height: 83px;
-  @media(min-width: theme.$breakpoint-lg) {
-   img {
-    height: 32px;
-    }
+ 
+  img {
+    height: 83px;
+    width: 83px;
```
keianhzo, line 77: I think this will make non-square images to look stretched.

src/react-components/room/RoomEntryModal.scss
```diff
@@ -18,6 +18,11 @@
   margin-bottom: 32px;
 }
 
+:local(.logo) {
+  height: 250px;
+  width: 250px;
```
keianhzo, line 23: Same with this

keianhzo (changes requested):
I think we still support non-square images for logos, if that's the case setting fixed widths will break proportional scaling.

nickgrato:
> I think we still support non-square images for logos, if that's the case setting fixed widths will break proportional scaling.

@keianhzo  the direction I'm getting from the desing and UX is to only support square. There is a ticket for me to update the copy for the logo inputs to say this .

Screen Shot 2023-03-09 at 7 26 44 AM

PR #5985, opened 2023-03-08
Implement Clone and Open link button function with bitecs

takahirox:
This PR adds the implementation of object menu Clone and Open link button functions with bitecs.

### DIFF
--- a/src/bit-systems/object-menu.ts
+++ b/src/bit-systems/object-menu.ts
@@ -1,13 +1,20 @@
 import { defineQuery, entityExists, hasComponent } from "bitecs";
+import { Matrix4, Vector3 } from "three";
 import type { HubsWorld } from "../app";
 import { HoveredRemoteRight, Interacted, ObjectMenu, ObjectMenuTarget } from "../bit-components";
 import { anyEntityWith, findAncestorWithComponent } from "../utils/bit-utils";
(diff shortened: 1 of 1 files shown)


src/bit-systems/object-menu.ts
```diff
@@ -38,6 +45,39 @@ function moveToTarget(world: HubsWorld, menu: EntityID) {
   setMatrixWorld(menuObj, targetObj.matrixWorld);
 }
 
+function openLink(world: HubsWorld, eid: EntityID) {
+  const { initialData } = createMessageDatas.get(eid)!;
+  const src = initialData.src;
+  // TODO: Currently only accounts for the simple case of an external url
+  //       but should support other type actions(eg: avatar update for avatar
+  //       url, room switch for Hubs room url).
+  //       See src/components/open-media-button.js
```
johnshaughnessy, line 54: Now that we aren't duplicated the object menus for every object, we could have each of these actions called from their own buttons if we wanted to, and then show/hide the buttons as needed. Otherwise, we'll need to set the text of the button for "use avatar" / "use scene" / "visit room" / "open link". It doesn't seem that important which way we choose, but since the only thing the buttons will have in common will be their position (not their text or functionality when clicked), I thought it might make sense to make them different buttons.

johnshaughnessy (approved):

PR #5991, opened 2023-03-10
Revert "Enable fast room switching by default"

johnshaughnessy:
This reverts commit 4c3bc39ef9c97e0ccbe63d48c9c8aa023c1ba331.


johnshaughnessy:
Closing with merging because the linked commit did not cause any regressions. (It just revealed one, which we fixed in https://github.com/mozilla/hubs/pull/5992

PR #5993, opened 2023-03-11
Implement rotate and scale button functions with bitecs

takahirox:
This commit adds the implementations of Object menu rotate and scale button functions.

Note that this commit focuses on making them just functional with minimal change.

The implementation relys on the old systems to save development time. It should be rewritten more elegantly to fit more to bitECS.

The main logic of src/components/scale-button is exposed to be reused from the new and old systems.

### DIFF
--- a/src/bit-systems/object-menu.ts
+++ b/src/bit-systems/object-menu.ts
@@ -1,7 +1,15 @@
-import { defineQuery, entityExists, hasComponent } from "bitecs";
-import { Matrix4, Vector3 } from "three";
+import { defineQuery, enterQuery, entityExists, exitQuery, hasComponent } from "bitecs";
+import { Matrix4, Quaternion, Vector3 } from "three";
 import type { HubsWorld } from "../app";
-import { HoveredRemoteRight, Interacted, ObjectMenu, ObjectMenuTarget } from "../bit-components";
+import {
+  HeldRemoteRight,
+  HoveredRemoteRight,
+  Interacted,
+  ObjectMenu,
+  ObjectMenuTarget,
+  RemoteRight,
+  Rigidbody
+} from "../bit-components";
 import { anyEntityWith, findAncestorWithComponent } from "../utils/bit-utils";
 import { createNetworkedEntity } from "../utils/create-networked-entity";
 import { createEntityState, deleteEntityState } from "../utils/entity-state-utils";
@@ -10,11 +18,16 @@ import type { EntityID } from "../utils/networking-types";
 import { setMatrixWorld } from "../utils/three-utils";
 import { deleteTheDeletableAncestor } from "./delete-entity-system";
 import { createMessageDatas, isPinned } from "./networking";
+import { TRANSFORM_MODE } from "../components/transform-object-button";
+import { ScalingHandler } from "../components/scale-button";
 
 // Working variables.
-const _vec1 = new Vector3();
-const _vec2 = new Vector3();
+const _vec3_1 = new Vector3();
+const _vec3_2 = new Vector3();
+const _quat = new Quaternion();
 const _mat4 = new Matrix4();
+
+let scalingHandler: ScalingHandler | null = null;
 
 function clicked(world: HubsWorld, eid: EntityID) {
   return hasComponent(world, Interacted, eid);
@@ -39,10 +52,65 @@ function moveToTarget(world: HubsWorld, menu: EntityID) {
   const targetObj = world.eid2obj.get(ObjectMenu.targetRef[menu])!;
   targetObj.updateMatrices();
 
-  const menuObj = world.eid2obj.get(menu)!;
-
   // TODO: position the menu more carefully...
-  setMatrixWorld(menuObj, targetObj.matrixWorld);
+  //       For example, if a menu object is just placed at a target
+  //       object's position the menu object can be hidden by a large
+  //       target object or the menu object looks too small for a far
+  //       target object.
+  _mat4.copy(targetObj.matrixWorld);
+
+  // Keeps world scale (1, 1, 1) because
+  // a menu object is a child of a target object
+  // and the target object's scale can be changed.
+  // Another option may be making the menu object
+  // a sibling of the target object.
+  _mat4.decompose(_vec3_1, _quat, _vec3_2);
+  _vec3_2.set(1.0, 1.0, 1.0);
+  _mat4.compose(_vec3_1, _quat, _vec3_2);
+
+  const menuObj = world.eid2obj.get(menu)!;
+  setMatrixWorld(menuObj, _mat4);
+
+  // TODO: Remove the dependency with AFRAME
+  const camera = AFRAME.scenes[0].systems["hubs-systems"].cameraSystem.viewingCamera;
+  camera.updateMatrices();
+  menuObj.lookAt(_vec3_1.setFromMatrixPosition(camera.matrixWorld));
+}
+
+// TODO: startRotation/Scaling() and stopRotation/Scaling() are
+//       temporary implementation that rely on the old systems.
+//       They should be rewritten more elegantly with bitecs.
+function startRotation(world: HubsWorld, targetEid: EntityID) {
+  const transformSystem = APP.scene!.systems["transform-selected-object"];
+  const physicsSystem = AFRAME.scenes[0].systems["hubs-systems"].physicsSystem;
+  physicsSystem.updateBodyOptions(Rigidbody.bodyId[targetEid], { type: "kinematic" });
+  const rightCursorEid = anyEntityWith(world, RemoteRight)!;
+  transformSystem.startTransform(world.eid2obj.get(targetEid)!, world.eid2obj.get(rightCursorEid)!, {
+    mode: TRANSFORM_MODE.CURSOR
+  });
+}
+
+function stopRotation() {
+  const transformSystem = APP.scene!.systems["transform-selected-object"];
+  transformSystem.stopTransform();
+}
+
+function startScaling(world: HubsWorld, targetEid: EntityID) {
+  // TODO: Don't use any
+  // TODO: Remove the dependency with AFRAME
+  const transformSystem = (AFRAME as any).scenes[0].systems["transform-selected-object"];
+  const physicsSystem = AFRAME.scenes[0].systems["hubs-systems"].physicsSystem;
+  physicsSystem.updateBodyOptions(Rigidbody.bodyId[targetEid], { type: "kinematic" });
+  const rightCursorEid = anyEntityWith(world, RemoteRight)!;
+  scalingHandler = new ScalingHandler(world.eid2obj.get(targetEid), transformSystem);
+  scalingHandler!.objectToScale = world.eid2obj.get(targetEid);
+  scalingHandler!.startScaling(world.eid2obj.get(rightCursorEid));
+}
+
+function stopScaling(world: HubsWorld) {
+  const rightCursorEid = anyEntityWith(world, RemoteRight)!;
+  scalingHandler!.endScaling(world.eid2obj.get(rightCursorEid));
+  scalingHandler = null;
 }
 
 function openLink(world: HubsWorld, eid: EntityID) {
@@ -71,8 +139,8 @@ function cloneObject(world: HubsWorld, sourceEid: EntityID) {
   const sourceObj = world.eid2obj.get(sourceEid)!;
   sourceObj.updateMatrices();
 
-  const objPos = _vec1.setFromMatrixPosition(sourceObj.matrixWorld);
-  const cameraPos = _vec2.setFromMatrixPosition(camera.matrixWorld);
+  const objPos = _vec3_1.setFromMatrixPosition(sourceObj.matrixWorld);
+  const cameraPos = _vec3_2.setFromMatrixPosition(camera.matrixWorld);
   objPos.add(cameraPos.sub(objPos).normalize().multiplyScalar(0.2));
   const clonedMatrixWorld = _mat4.copy(sourceObj.matrixWorld).setPosition(objPos);
   setMatrixWorld(clonedObj, clonedMatrixWorld);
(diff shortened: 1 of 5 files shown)


src/components/scale-button.js
```diff
@@ -169,4 +162,30 @@ AFRAME.registerComponent("scale-button", {
     this.objectMatrix.scale(this.deltaScale);
     setMatrixWorld(this.objectToScale, this.objectMatrix);
   }
+}
+
+AFRAME.registerComponent("scale-button", {
+  init() {
+    this.handler = new ScalingHandler(this.el.object3D, this.el.sceneEl.systems["transform-selected-object"]);
```
takahirox, line 169: I'm not familiar with A-Frame component/system lifecycles. Is it safe to assume that `transform-selected-object` system is already registered here?

src/bit-systems/object-menu.ts
```diff
@@ -10,12 +18,19 @@ import type { EntityID } from "../utils/networking-types";
...
 
 // Working variables.
-const _vec1 = new Vector3();
-const _vec2 = new Vector3();
+const _vec3_1 = new Vector3();
+const _vec3_2 = new Vector3();
+const _quat = new Quaternion();
 const _mat4 = new Matrix4();
 
+// TODO: Avoid any
```
keianhzo, line 31: Why can't this be of type `ScalingHandler`?
```javascript
let scalingHandler: ScalingHandler | null = null;
```

src/bit-systems/object-menu.ts
```diff
@@ -39,10 +54,69 @@ function moveToTarget(world: HubsWorld, menu: EntityID) {
...
+  camera.updateMatrices();
+  menuObj.lookAt(_vec3_1.setFromMatrixPosition(camera.matrixWorld));
+}
+
+// TODO: startRotation/Scaling() and stopRotation/Scaling() are
+//       temporarl implementation that rely on the old systems.
+//       They should be rewritten more elegantly with bitecs.
+function startRotation(world: HubsWorld, targetEid: EntityID) {
+  // TODO: Don't use any
+  // TODO: Remove the dependency with AFRAME
```
keianhzo, line 87: If you want avoid using any and avoid using the `AFRAME` symbol you can also use:
```javascript
APP.scene!
```

and the add the type to the `aframe.d.ts`:
```javascript
 interface TransformSelectedObjectSystem extends ASystem {
    startTransform(Object3D, EntityID, any): void;
    stopTransform(): void;
  }

  interface AScene extends AElement {
    object3D: Scene;
    renderStarted: boolean;
    renderer: WebGLRenderer;
    tick(time: number, delta: number): void;
    isPlaying: boolean;
    behaviors: {
      tick: AComponent[];
      tock: AComponent[];
    };
    systemNames: Array<keyof AScene["systems"]>;
    systems: {
      "hubs-systems": HubsSystems;
      "personal-space-bubble": PersonalSpaceBubbleSystem;
      "transform-selected-object": TransformSelectedObjectSystem;
      userinput: UserInputSystem;
      /** @deprecated see bit-interaction-system */
      interaction: InteractionSystem;
      nav: NavSystem;
    };
    emit(string, any?): void;
    addState(string): void;
    is(string): boolean;
  }
```

keianhzo (changes requested):
LGTM, just a few nits. The menu placement improvements can probably be addressed in a follow-up.

src/components/scale-button.js
keianhzo: I think it should per:
https://github.com/mozilla/hubs/blob/3c897c0461fca5591646ce531af0a76c2cef3f32/src/hub.js#L123-L124

takahirox: Thanks.

And if I'm right, system seems to be registered to `sceneEl.systems` when `AFRAME.registerSystem` is called (when a file is imported) while `Component.init()` seems to be called later (when the component first initialized?). So probably it is safe to assume here.

src/bit-systems/object-menu.ts
```diff
@@ -1,7 +1,15 @@
-import { defineQuery, entityExists, hasComponent } from "bitecs";
-import { Matrix4, Vector3 } from "three";
+import { addComponent, defineQuery, enterQuery, entityExists, exitQuery, hasComponent } from "bitecs";
```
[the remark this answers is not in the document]
takahirox: Thanks, updated.

src/bit-systems/object-menu.ts
takahirox: Good catch, updated.

src/bit-systems/object-menu.ts
```diff
@@ -39,10 +54,69 @@ function moveToTarget(world: HubsWorld, menu: EntityID) {
   const targetObj = world.eid2obj.get(ObjectMenu.targetRef[menu])!;
   targetObj.updateMatrices();
 
+  // TODO: position the menu more carefully...
```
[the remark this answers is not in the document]
takahirox: > We can do it in follow-up PR.

Yes, that's the plan.

src/bit-systems/object-menu.ts
```diff
@@ -39,10 +54,69 @@ function moveToTarget(world: HubsWorld, menu: EntityID) {
...
-  // TODO: position the menu more carefully...
-  setMatrixWorld(menuObj, targetObj.matrixWorld);
+  // TODO: Remove the dependency with AFRAME
+  const camera = AFRAME.scenes[0].systems["hubs-systems"].cameraSystem.viewingCamera;
+  camera.updateMatrices();
+  menuObj.lookAt(_vec3_1.setFromMatrixPosition(camera.matrixWorld));
+}
+
+// TODO: startRotation/Scaling() and stopRotation/Scaling() are
+//       temporarl implementation that rely on the old systems.
```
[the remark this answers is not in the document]
takahirox: Thanks, updated.

src/bit-systems/object-menu.ts
```diff
@@ -39,10 +54,69 @@ function moveToTarget(world: HubsWorld, menu: EntityID) {
...
+  const transformSystem = (AFRAME as any).scenes[0].systems["transform-selected-object"];
+  const physicsSystem = AFRAME.scenes[0].systems["hubs-systems"].physicsSystem;
+  physicsSystem.updateBodyOptions(Rigidbody.bodyId[targetEid], { type: "kinematic" });
+  const rightCursorEid = anyEntityWith(world, RemoteRight)!;
+  scalingHandler = new ScalingHandler(world.eid2obj.get(targetEid), transformSystem);
+  scalingHandler.objectToScale = world.eid2obj.get(targetEid);
+  scalingHandler.startScaling(world.eid2obj.get(rightCursorEid));
+}
+
+function stopScaling(world: HubsWorld, targetEid: EntityID) {
```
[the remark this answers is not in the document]
takahirox: Good catch, updated.

src/bit-systems/object-menu.ts
takahirox: There are some other bitecs systems that have AFRAME dependency. I hope we will work on it when we remove the dependency.

takahirox:
This PR needs to be unlocked by the reviewers to get merged.

src/bit-systems/object-menu.ts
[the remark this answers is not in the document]
netpro2k: We should probably just aim for roughly what we have now in terms of menu positioning, we still would like to ultimately not have these in world object menus in their current form on most platforms. On desktop and mobile we likely end up with more traditional 2d context menus, and in VR we likely don't need most of what the menus do since direct manipulation is perdurable (we still need to figure out what to do with the rest of the options)

src/bit-systems/object-menu.ts
netpro2k: Agree with Manuel that just adding the type to the aframe.d.ts would be nice. Its fairly minor either way but prevents from making some accidental mistakes leaking the any around.

src/bit-systems/object-menu.ts
```diff
@@ -39,10 +52,69 @@ function moveToTarget(world: HubsWorld, menu: EntityID) {
...
+
+// TODO: startRotation/Scaling() and stopRotation/Scaling() are
+//       temporary implementation that rely on the old systems.
+//       They should be rewritten more elegantly with bitecs.
+function startRotation(world: HubsWorld, targetEid: EntityID) {
+  // TODO: Don't use any
+  // TODO: Remove the dependency with AFRAME
+  const transformSystem = (AFRAME as any).scenes[0].systems["transform-selected-object"];
+  const physicsSystem = AFRAME.scenes[0].systems["hubs-systems"].physicsSystem;
+  physicsSystem.updateBodyOptions(Rigidbody.bodyId[targetEid], { type: "kinematic" });
```
netpro2k, line 88: Not something to address here but just to get it bouncing around in everyone's head. We change these physics properties pretty arbitrarily all over and its very easy for things to end up in weird states and to overwrite one system's desired physics settings with another. We need to think of how to manage that. It mostly comes about when something wants to set the position/rotation of a physics object since it must be kinematic or static to do so, but then it needs to be able to return to its previous state.

src/bit-systems/object-menu.ts
```diff
@@ -142,13 +232,29 @@ function updateVisibility(world: HubsWorld, menu: EntityID, frozen: boolean) {
...
+const heldExitQuery = exitQuery(heldQuery);
 export function objectMenuSystem(world: HubsWorld, sceneIsFrozen: boolean, hubChannel: HubChannel) {
   const menu = anyEntityWith(world, ObjectMenu)!;
 
   ObjectMenu.targetRef[menu] = objectMenuTarget(world, menu, sceneIsFrozen);
   if (ObjectMenu.targetRef[menu]) {
-    moveToTarget(world, menu);
     handleClicks(world, menu, hubChannel);
+
+    heldExitQuery(world).forEach(eid => {
```
netpro2k, line 245: My first thought would have been to do `hasComponent(world, Held, ObjectMenu.rotateButtonRef[menuEid)` comparison to the current transform system state.

Using quires is fine, and maybe its nice that the previous frame state is handled by the enter/exit queries but just noting. Not sure which I like better.

src/prefabs/object-menu.tsx
```diff
@@ -198,6 +199,8 @@ function ScaleButton(props: Attrs) {
       height={buttonHeight}
       type={BUTTON_TYPES.ACTION}
       text={"scale"}
+      holdable
+      holdableButton
```
netpro2k, line 203: Bit weird to have to specify `holdableButton` on a `Button3D` should probably be handled in `Button3D` based on just `holdable`.

The disconnect between things being components and things just being properties to a JSX function here is definitely something I don't love, but its just how JSX works. Like here "height" is not a component nor is "text". 

I think mixing components "holdable" and "holdableButton" into it makes it even more confusing.

src/components/scale-button.js
```diff
@@ -63,85 +72,71 @@ AFRAME.registerComponent("scale-button", {
...
-    this.startScaling = e => {
-      if (this.isScaling || !this.objectToScale) {
-        return;
-      }
+  }
 
-      if (!(NAF.utils.isMine(this.networkedEl) || NAF.utils.takeOwnership(this.networkedEl))) {
-        return;
-      }
+  startScaling(object3D) {
```
netpro2k, line 77: This is temporary transition code so it doesn't really matter much, but we might try to name this parameter a bit better. Maybe cursorObj? interactor?

src/bit-systems/object-menu.ts
```diff
@@ -39,10 +52,69 @@ function moveToTarget(world: HubsWorld, menu: EntityID) {
...
+  const physicsSystem = AFRAME.scenes[0].systems["hubs-systems"].physicsSystem;
+  physicsSystem.updateBodyOptions(Rigidbody.bodyId[targetEid], { type: "kinematic" });
+  const rightCursorEid = anyEntityWith(world, RemoteRight)!;
+  scalingHandler = new ScalingHandler(world.eid2obj.get(targetEid), transformSystem);
+  scalingHandler!.objectToScale = world.eid2obj.get(targetEid);
+  scalingHandler!.startScaling(world.eid2obj.get(rightCursorEid));
+}
+
+function stopScaling(world: HubsWorld) {
+  const rightCursorEid = anyEntityWith(world, RemoteRight)!;
```
netpro2k, line 115: Not a suggestion for here but reminds me of an observation I want us to think about. 

I find it annoying that we don't have a good way to statically define references like this and have to "look them up" every frame with a query like this. Its not particularly expensive or anything, just silly because we actually know what number entity this is going to be every time. The only thing really stopping us from doing it is module loading order.

netpro2k (commented):
This looks good. Made a few comments but most are just observations or thoughts I had while reviewing and not really suggestions for changes.

src/bit-systems/object-menu.ts
takahirox: I don't love `anyEntityWith()`. Added this topic in the internal design document.

src/bit-systems/object-menu.ts
takahirox: I can understand.

Using `hasComponent(world, Held, ObjectMenu.rotateButtonRef[menuEid])` may be more efficient because `Held` component may be a common Component so `heldQuery` might include a way more entity ids than needed here.

But it required to manage previous frame state. I want to avoid to manage a previous frame state in a system for simplicity.

Using queries may be inefficient but we don't need to manually manage previous frame state so it is simpler.

I want to go as is for now. We may revisit later if we face performance problem.

src/bit-systems/object-menu.ts
takahirox: OK, two developers suggested the same thing. And the change is minor. Updated.

src/prefabs/object-menu.tsx
takahirox: Updated to remove `holdableButton` here.

keianhzo (approved):

PR #5995, opened 2023-03-13
Theme editor

nickgrato:
TODO


src/schema.toml
```diff
@@ -71,9 +71,9 @@ theme.notice-background-color = { category = "theme", type = "color", name = "Se
 theme.notice-text-color = { category = "theme", type = "color", name = "Secondary Text Color", description = "[deprecated] Notice and indicator text color.", deprecated="true" }
 theme.notice-widget-color = { category = "theme", type = "color", name = "Secondary Widget Color", description = "[deprecated] Notice and indicator UI widget color.", deprecated="true" }
 theme.favorited-color = { category = "theme", type = "color", name = "Favorited Color", description = "[deprecated] Favorited Color", deprecated = "true" }
-theme.nametag-color = { category = "theme", type = "color", name = "Nametag Color", description = "Avatar nametag color." }
-theme.nametag-text-color = { category = "theme", type = "color", name = "Nametag Color", description = "Avatar nametag text color." }
-theme.themes = { category = "theme", type = "json", name = "Themes", description = "A json array describing custom client themes." }
+theme.nametag-color = { category = "theme", type = "color", name = "Background Color", description="The background color of nametags which float above each room guest." }
```
nickgrato, line 74: @johnshaughnessy  I don't know enough about this to tell if changing the name corrupts the form?

johnshaughnessy: It's OK to change the name, description, etc because those are only used within the admin panel. 

You can't change `theme.nametag-color` without consequence because that's the key that's been written to the database. (You'd have to write code in the client `theme.foo || theme.nametag-color` if you wanted to change it to `theme.foo`, so that existing configs continued to work correctly.)

This ends up in `APP_CONFIG`:
image

admin/src/react-components/service-editor.js
```diff
@@ -325,12 +388,14 @@ class ConfigurationEditor extends Component {
     const displayPath = path.join(" > ");
     return (
       <div key={displayPath} className={this.props.classes.colorInput}>
+        {descriptor.name === "Background Color" && <h3 className="heading-sm mb-24">Nametags</h3>}
```
johnshaughnessy, line 391: It feels wrong to insert logic like this here and in service-editor (for handling themes). It belongs at a higher level. It would not be difficult to move it up (to where we decide to render the themes section). It just means we won't blindly render the controls for that chunk of config variables.

That being said, I understand this was meant to be done as quickly as possible, and you're still getting familiar with the admin panel.

If this code still exists in 6 months, if there are additional incremental changes planned for the admin panel, or if what I'm saying isn't clear, then we should move this logic to a higher level. If what I'm describing is unclear, then I'm happy to chat through it.

johnshaughnessy (approved):
The logic inserted here is at the wrong layer. If you're trying to ship this today because you have bigger fish to fry, don't let me stop you. If you're not sure how to move it to a higher layer, then we should pair to do that so that you don't repeat this for future changes you want to make to the admin panel.

src/schema.toml
nickgrato: @johnshaughnessy  perfect that helps a ton

admin/src/react-components/service-editor.js
```diff
@@ -240,42 +261,66 @@ class ConfigurationEditor extends Component {
...
+              this.onChange(path, pretty);
+            }
+          }}
+          helperText={description}
+          type="text"
+          fullWidth
+          margin="normal"
+        />
+
+        {isTheme && (
```
nickgrato, line 305: I'm thinking this still might be good here. I have ideas on how to make this more flexible in the future. but because the button ties in so much with the inputs hooks and current value it seems logical to keep in the component for now.

admin/src/react-components/service-editor.js
```diff
@@ -359,6 +405,60 @@ class ConfigurationEditor extends Component {
...
+  renderCategory(schema, category, config) {
+    switch (category) {
+      case "theme":
+        return this.renderThemeSection(schema.theme, config);
+      default:
+        return this.renderTree(schema, category, config);
+    }
+  }
+
+  renderThemeSection(theme, config) {
```
nickgrato, line 417: @johnshaughnessy  here is the new theme section.

admin/src/react-components/service-editor.js
```diff
@@ -359,11 +404,73 @@ class ConfigurationEditor extends Component {
...
+   */
+  renderThemeSection(theme, config) {
+    const configurables = this.getFilteredDescriptors(theme);
+    const getInput = ([path, descriptor]) => this.renderConfigurable(path, descriptor, getConfigValue(config, path));
+
+    return (
+      <form onSubmit={this.onSubmit.bind(this)}>
+        <h3 className="heading-sm mb-24">Nametags</h3>
+        {getInput(configurables[0])}
+        {getInput(configurables[1])}
```
johnshaughnessy, line 433: Since you know which configurables you need to render (`theme.nametag-color` and `theme.nametag-text-color`) -- it might be better to specify them by `path` rather than by index.

Something like

```ts
const configurable = (path) => {
  const descriptor = getDescriptors( path.split(".")[0] ).find( ([path, desc]) => desc.path === path )[1];
  return this.renderConfigurable(path, descriptor, getConfigValue(config, path));
}

// and in place of getInput, we would write:
configurable("theme.nametag-color")
configurable("theme.nametag-text-color")
...
configurable("theme.themes")
```

admin/src/react-components/service-editor.js
johnshaughnessy: I don't know how to lift the `currentValue` out from the `TextField` but in theory that's how I would _want_ to handle this. Conceptually this part is meant to be a presentational component. When I added the json validation way back when, I didn't know how to do that logic at the higher level so I just stuffed it in here. If you figure out how to handle the logic in a container component, please lift both the code you added (the button and themes specific stuff) AND the JSON validation stuff I added a while back.

But I don't mind if you don't do that now.

johnshaughnessy (approved):
I think there's state in `renderLongTextInput` that should be pulled out.

Specifically, I think the `renderThemeSection` function could call `renderLongTextInput` directly (instead of `renderConfigurable`) and handle things like the default value and the reset to default theme button.

But, it's not a big deal and I don't want to force the issue.

PR #6003, opened 2023-03-21
Bit media frames physics

keianhzo:
Built on top of https://github.com/mozilla/hubs/pull/5978. Only the last commit applies.

As commented in a previous PR, 

> It's a bit unfortunate that we have to go back and forth between bitECS and Ammo to keep everything in sync. As the
entry point for the physics life cycle is the physics system, I wonder it would be better to encapsulate all the > component life cycle there and let that system manage it all. Instead of going through the hops of adding a component and have that intermediate bit-physics-system to map that to the physics system, just call the physics system and it will add/update the component under the hood and keep everything in sync.

### DIFF
--- a/src/bit-systems/media-loading.ts
+++ b/src/bit-systems/media-loading.ts
@@ -1,12 +1,13 @@
 import { addComponent, defineQuery, enterQuery, exitQuery, hasComponent, removeComponent, removeEntity } from "bitecs";
 import { Vector3 } from "three";
 import { HubsWorld } from "../app";
-import { GLTFModel, MediaLoaded, MediaLoader, Networked, ObjectMenuTarget } from "../bit-components";
+import { GLTFModel, MediaLoaded, MediaLoader, Networked, ObjectMenuTarget, PhysicsShape } from "../bit-components";
 import { inflatePhysicsShape, Shape } from "../inflators/physics-shape";
 import { ErrorObject } from "../prefabs/error-object";
 import { LoadingObject } from "../prefabs/loading-object";
 import { animate } from "../utils/animate";
 import { setNetworkedDataWithoutRoot } from "../utils/assign-network-ids";
+import { computeObjectAABB } from "../utils/auto-box-collider";
 import { crClearTimeout, crNextFrame, crTimeout } from "../utils/coroutine";
 import { ClearFunction, JobRunner, withRollback } from "../utils/coroutine-utils";
 import { easeOutQuadratic } from "../utils/easing";
@@ -18,6 +19,39 @@ import { loadVideo } from "../utils/load-video";
 import { loadAudio } from "../utils/load-audio";
 import { MediaType, mediaTypeName, resolveMediaInfo } from "../utils/media-utils";
 import { EntityID } from "../utils/networking-types";
+
+export const MediaContentBounds = (MediaLoaded as any).contentBounds as Map<EntityID, Vector3>;
+
+const getBox = (() => {
+  const rotation = new THREE.Euler();
+  return (world: HubsWorld, eid: EntityID, rootEid: EntityID, worldSpace?: boolean) => {
+    const box = new THREE.Box3();
+    const obj = world.eid2obj.get(eid)!;
+    const rootObj = world.eid2obj.get(rootEid)!;
+
+    rotation.copy(obj.rotation);
+    obj.rotation.set(0, 0, 0);
+    obj.updateMatrices(true, true);
+    rootObj.updateMatrices(true, true);
+    rootObj.updateMatrixWorld(true);
+
+    computeObjectAABB(rootObj, box, false);
+
+    if (!box.isEmpty()) {
+      if (!worldSpace) {
+        obj.worldToLocal(box.min);
+        obj.worldToLocal(box.max);
+      }
+      obj.rotation.copy(rotation);
+      obj.matrixNeedsUpdate = true;
+    }
+
+    rootObj.matrixWorldNeedsUpdate = true;
+    rootObj.updateMatrices();
+
+    return box;
+  };
+})();
 
 export function* waitForMediaLoaded(world: HubsWorld, eid: EntityID) {
   while (hasComponent(world, MediaLoader, eid)) {
@@ -176,8 +210,11 @@ function* loadAndAnimateMedia(world: HubsWorld, eid: EntityID, clearRollbacks: C
   removeComponent(world, MediaLoader, eid);
 
   if (media) {
+    const box = getBox(world, eid, media);
+    MediaContentBounds.set(media, box.getSize(new Vector3()));
     // TODO update scale?
-    inflatePhysicsShape(world, media, {
+    removeComponent(world, PhysicsShape, eid);
+    inflatePhysicsShape(world, eid, {
       type: hasComponent(world, GLTFModel, media) ? Shape.HULL : Shape.BOX,
       minHalfExtent: 0.04
     });
@@ -188,6 +225,8 @@ const jobs = new JobRunner();
 const mediaLoaderQuery = defineQuery([MediaLoader]);
 const mediaLoaderEnterQuery = enterQuery(mediaLoaderQuery);
 const mediaLoaderExitQuery = exitQuery(mediaLoaderQuery);
+const mediaLoadedQuery = defineQuery([MediaLoaded]);
+const mediaLoadedExitQuery = exitQuery(mediaLoadedQuery);
 export function mediaLoadingSystem(world: HubsWorld) {
   mediaLoaderEnterQuery(world).forEach(function (eid) {
     jobs.add(eid, clearRollbacks => loadAndAnimateMedia(world, eid, clearRollbacks));
@@ -196,6 +235,8 @@ export function mediaLoadingSystem(world: HubsWorld) {
   mediaLoaderExitQuery(world).forEach(function (eid) {
     jobs.stop(eid);
   });
+
+  mediaLoadedExitQuery(world).forEach(eid => MediaContentBounds.delete(eid));
 
   jobs.tick();
 }
--- a/src/bit-systems/media-loading.ts
+++ b/src/bit-systems/media-loading.ts
@@ -1,5 +1,5 @@
 import { addComponent, defineQuery, enterQuery, exitQuery, hasComponent, removeComponent, removeEntity } from "bitecs";
-import { Vector3 } from "three";
+import { Box3, Euler, Vector3 } from "three";
 import { HubsWorld } from "../app";
 import {
   GLTFModel,
@@ -7,8 +7,7 @@ import {
   MediaLoaded,
   MediaLoader,
   Networked,
-  ObjectMenuTarget,
-  PhysicsShape
+  ObjectMenuTarget
 } from "../bit-components";
 import { inflatePhysicsShape, Shape } from "../inflators/physics-shape";
 import { ErrorObject } from "../prefabs/error-object";
@@ -29,9 +28,9 @@ import { MediaType, mediaTypeName, resolveMediaInfo } from "../utils/media-utils
 import { EntityID } from "../utils/networking-types";
 
 const getBox = (() => {
-  const rotation = new THREE.Euler();
+  const rotation = new Euler();
   return (world: HubsWorld, eid: EntityID, rootEid: EntityID, worldSpace?: boolean) => {
-    const box = new THREE.Box3();
+    const box = new Box3();
     const obj = world.eid2obj.get(eid)!;
     const rootObj = world.eid2obj.get(rootEid)!;
 
@@ -95,12 +94,12 @@ function resizeAndRecenter(world: HubsWorld, media: EntityID, eid: EntityID) {
   if (!resize && !recenter) return;
 
   const mediaObj = world.eid2obj.get(media)!;
-  const box = new THREE.Box3();
+  const box = new Box3();
   box.setFromObject(mediaObj);
 
   let scalar = 1;
   if (resize) {
-    const size = new THREE.Vector3();
+    const size = new Vector3();
     box.getSize(size);
     scalar = 1 / Math.max(size.x, size.y, size.z);
     if (hasComponent(world, GLTFModel, media)) scalar = scalar * 0.5;
@@ -109,7 +108,7 @@ function resizeAndRecenter(world: HubsWorld, media: EntityID, eid: EntityID) {
   }
 
   if (recenter) {
-    const center = new THREE.Vector3();
+    const center = new Vector3();
     box.getCenter(center);
     mediaObj.position.copy(center).multiplyScalar(-1 * scalar);
     mediaObj.matrixNeedsUpdate = true;
--- a/src/bit-systems/media-loading.ts
+++ b/src/bit-systems/media-loading.ts
@@ -217,10 +217,12 @@ function* loadAndAnimateMedia(world: HubsWorld, eid: EntityID, clearRollbacks: C
   removeComponent(world, MediaLoader, eid);
 
   if (media) {
-    const box = getBox(world, eid, media);
-    addComponent(world, MediaContentBounds, media);
-    box.getSize(tmpVector);
-    MediaContentBounds.bounds[media].set(tmpVector.toArray());
+    if (hasComponent(world, MediaLoaded, media)) {
+      const box = getBox(world, eid, media);
+      addComponent(world, MediaContentBounds, media);
+      box.getSize(tmpVector);
+      MediaContentBounds.bounds[media].set(tmpVector.toArray());
+    }
     // TODO update scale?
     removeComponent(world, PhysicsShape, eid);
     inflatePhysicsShape(world, eid, {
@@ -234,8 +236,6 @@ const jobs = new JobRunner();
 const mediaLoaderQuery = defineQuery([MediaLoader]);
 const mediaLoaderEnterQuery = enterQuery(mediaLoaderQuery);
 const mediaLoaderExitQuery = exitQuery(mediaLoaderQuery);
-const mediaLoadedQuery = defineQuery([MediaLoaded]);
-const mediaLoadedExitQuery = exitQuery(mediaLoadedQuery);
 export function mediaLoadingSystem(world: HubsWorld) {
   mediaLoaderEnterQuery(world).forEach(function (eid) {
     jobs.add(eid, clearRollbacks => loadAndAnimateMedia(world, eid, clearRollbacks));
@@ -244,8 +244,6 @@ export function mediaLoadingSystem(world: HubsWorld) {
   mediaLoaderExitQuery(world).forEach(function (eid) {
     jobs.stop(eid);
   });
-
-  mediaLoadedExitQuery(world).forEach(eid => removeComponent(world, MediaContentBounds, eid));
 
   jobs.tick();
 }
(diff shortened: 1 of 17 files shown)


src/systems/bit-media-frames.js
```diff
@@ -229,10 +316,16 @@ function mediaTypesOf(world, entities) {
...
+  obj.traverse(child => {
+    disposeNode(child);
+  });
+}
+
 const takeOwnershipOnTimeout = new Map();
 const heldQuery = defineQuery([Held]);
 // const droppedQuery = exitQuery(heldQuery);
-export function mediaFramesSystem(world) {
+export function mediaFramesSystem(world, physicsSystem) {
```
takahirox, line 328: Not a blocking issue, but a comment for future clean up.

I think ideally a system shouldn't have a dependency with another system.

src/bit-systems/media-loading.ts
```diff
@@ -19,6 +20,39 @@ import { loadAudio } from "../utils/load-audio";
 import { MediaType, mediaTypeName, resolveMediaInfo } from "../utils/media-utils";
 import { EntityID } from "../utils/networking-types";
 
+export const MediaContentBounds = (MediaLoaded as any).contentBounds as Map<EntityID, Vector3>;
+
+const getBox = (() => {
+  const rotation = new THREE.Euler();
+  return (world: HubsWorld, eid: EntityID, rootEid: EntityID, worldSpace?: boolean) => {
+    const box = new THREE.Box3();
```
takahirox, line 28: You may import `Euler` and `Box3`, and remove `THREE.` prefix here.

src/bit-systems/media-loading.ts
```diff
@@ -176,8 +210,11 @@ function* loadAndAnimateMedia(world: HubsWorld, eid: EntityID, clearRollbacks: C
   removeComponent(world, MediaLoader, eid);
 
   if (media) {
+    const box = getBox(world, eid, media);
+    MediaContentBounds.set(media, box.getSize(new Vector3()));
```
takahirox, line 214: Similar to `getBox`, ideally creating Vector3 for each call should be avoided?

takahirox, line 214: What happens if an exception is caught in `loadMedia()`? `media` is non-null but `MediaLoaded` component is not added to an entity. So, here `MediaContentBounds` is accessed even without `MediaLoaded` component.

src/bit-systems/media-loading.ts
```diff
@@ -19,6 +20,39 @@ import { loadAudio } from "../utils/load-audio";
 import { MediaType, mediaTypeName, resolveMediaInfo } from "../utils/media-utils";
 import { EntityID } from "../utils/networking-types";
 
+export const MediaContentBounds = (MediaLoaded as any).contentBounds as Map<EntityID, Vector3>;
```
takahirox, line 23: I'm not really sure if this alias pattern is readable because it is unclear `MediaContentBounds` data is associated with `MediaLoaded` component. At least, the name should clearly indicate the data is associated with `MediaLoaded` component.

src/bit-systems/media-loading.ts
```diff
@@ -19,6 +20,39 @@ import { loadAudio } from "../utils/load-audio";
...
+      obj.rotation.copy(rotation);
+      obj.matrixNeedsUpdate = true;
+    }
+
+    rootObj.matrixWorldNeedsUpdate = true;
+    rootObj.updateMatrices();
+
+    return box;
+  };
+})();
```
takahirox, line 54: Not a blocking issue, but a comment for future clean up.

I think `media-loading` system is one of the most complex systems and it may be hard to be maintained. And now we are adding a new code.

Perhaps the reasons of the complexity is `media-loading` system processes multiple things, like resolving URL, loading media, animating object after loading. If we can break `media-loading` system up to multiple systems, they can be simpler. And if we do that, this media content bound setting would be move to either one, or a new system.

src/systems/bit-media-frames.js
```diff
@@ -133,89 +157,152 @@ const snapToFrame = (() => {
...
     });
-  });
 
-  // TODO play animations on previews
+    if (hasComponent(world, AEntity, capturable)) {
+      const loopAnimation = el.components["loop-animation"];
+      if (loopAnimation && loopAnimation.isPlaying) {
+        const originalAnimation = loopAnimation.currentActions[loopAnimation.data.activeClipIndex];
+        const animation = previewMesh.animations[loopAnimation.data.activeClipIndex];
+        const mixer = new THREE.AnimationMixer(previewMesh);
```
takahirox, line 231: Probably we want to avoid `THREE` prefix. We may import Three.js objects from `three` and remove the prefix.

src/systems/bit-media-frames.js
```diff
@@ -102,26 +114,38 @@ function isColliding(world, eidA, eidB) {
   return false;
 }
 
+// TODO we only allow capturing media-loader so rely on its bounds calculations for now
+function getEntityBounds(world, target) {
```
takahirox, line 118: Rename `target` to `targetEid` as, IIRC, you suggested to add `Eid` suffix to name of variable that  holds eid?

src/systems/bit-media-frames.js
```diff
@@ -102,26 +114,38 @@ function isColliding(world, eidA, eidB) {
...
+  const targetObj = world.eid2obj.get(target);
+
+  let contentBounds;
+  if (hasComponent(world, AEntity, target)) {
+    contentBounds = targetObj.el.components["media-loader"].contentBounds;
+  } else {
+    const mediaEid = findChildWithComponent(world, MediaLoaded, target);
```
takahirox, line 125: Not a blocking issue, but a comment for future discussion.

I have been thinking that `findChild/AncestorWithComponent()` may be a bit less readable because for example it is often unclear who manages the relationships with child/ancestor entity from an entity and developers carefully need to be aware of the relationships not to break. This pattern maybe should be avoided where possible.

I added this topic to the internal design document.

src/systems/bit-media-frames.js
```diff
@@ -133,89 +157,152 @@ const snapToFrame = (() => {
...
+        (video.videoTexture.image.videoWidth || video.videoTexture.image.width);
+    }
+    srcMesh = el.getObject3D("mesh");
+  } else {
+    const mediaEid = findChildWithComponent(world, MediaLoaded, capturable);
+    isVideo = hasComponent(world, MediaVideo, mediaEid);
+    if (isVideo) {
+      ratio = MediaVideo.ratio[mediaEid];
+    }
+    srcMesh = APP.world.eid2obj.get(mediaEid);
```
takahirox, line 196: I guess `APP` prefix is not needed here?

src/systems/bit-media-frames.js
```diff
@@ -297,26 +399,30 @@ export function mediaFramesSystem(world) {
...
-      physicsSystem.updateBodyOptions(Rigidbody.bodyId[captured], { type: "dynamic" });
+      physicsSystem.updateBodyOptions(captured, { type: "dynamic" });
     }
 
     MediaFrame.capturedNid[frame] = NetworkedMediaFrame.capturedNid[frame];
     MediaFrame.scale[frame].set(NetworkedMediaFrame.scale[frame]);
 
-    display(world, frame, heldMediaTypes);
+    frame2mixer.forEach(mixer => {
+      mixer.update(APP.world.time.delta / 1000);
```
takahirox, line 423: I guess `APP` prefix is not needed here?

src/systems/bit-media-frames.js
```diff
@@ -133,89 +157,152 @@ const snapToFrame = (() => {
...
-      node.material = mat;
-      return mat;
+  return (obj, scaleArray) => {
+    obj.updateMatrices();
+    obj.matrixWorld.decompose(position, quaternion, scale);
+    setMatrixWorld(obj, m4.compose(position, quaternion, scale.fromArray(scaleArray)));
+  };
+})();
+
+const frame2mixer = new Map();
```
takahirox, line 173: To confirm, this `frame2mixer` is used only for `AEntity` (in old entity system) isn't it? I was not sure if it is good to manage `AnimationMixer` in this system, but it may be ok if it is only for old system that we will remove sooner or later. Adding a comment about it may help.

takahirox (commented):
I reviewed the PR mainly for bitecs design and readability. Most my comments are for future clean up and discussion, and don't block merging.

I haven't reviewed the core logic because I'm not familiar with media frames physics implementation. It may be ok if it keeps working because we focus on migration now. Or another reviewer may need to check.

Honestly `bit-systems/media-loading.ts`, `systems/bit-media-frames` (and others) are getting complex and harder to be maintained. But we may refactor them later because our development is at the migration phase now.

src/systems/bit-media-frames.js
```diff
@@ -264,16 +361,19 @@ export function mediaFramesSystem(world) {
...
-    if (captured && hasComponent(world, Owned, captured) && !hasComponent(world, Held, captured) && colliding) {
-      snapToFrame(world, frame, captured);
-      physicsSystem.updateBodyOptions(Rigidbody.bodyId[captured], { type: "kinematic" });
+    const isCapturedOwned = hasComponent(world, Owned, captured);
+    const isCapturedHeld = findChildWithComponent(world, Held, captured);
+    const colliding = captured && isColliding(world, physicsSystem, frame, captured);
+    const isFrameDeleting = findAncestorWithComponent(world, Deleting, frame);
+    const isFrameOwned = hasComponent(world, Owned, frame);
+
+    if (captured && isCapturedOwned && !isCapturedHeld && !isFrameDeleting && colliding) {
```
netpro2k, line 370: Was slightly uneasy when I saw the addition of the `Deleting` component at the start of reviewing this PR. I see here we are using it to prevent scaling the object down as the media frame it was in gets shrunk during the delete animation. It definitely seems really odd to have to think about that delete animation in systems and this definitely wont be the last time we have to... Don't have a nicer solution off the top of my head but we should definitely brainstorm on this more.

src/systems/bit-media-frames.js
```diff
@@ -229,10 +316,16 @@ function mediaTypesOf(world, entities) {
   return mask;
 }
 
+export function cleanupMediaFrame(obj) {
+  obj.traverse(child => {
```
netpro2k, line 320: No need for the wrapping function.

That said not sure this is correct. The only thing the media frame "owns" in terms of disposable resources is the geometry for the guide. The cloned preview object is all just references to geometries/textures that something else owns. Its actually incorrect to try and `.dispose()` a geometry/texture for an object we happened to be previewing just because this media frame is going away, since the object that actually "owns" those things is not going away. 

With the way threejs resource management works you won't actually notice this issue since it will lazily re-create everything. So I think what will happen in practice is when you delete a media frame capturing an object we will delete the GPU side resources for them and then re-upload them to the GPU next frame.

netpro2k: Oof, actually looking at the code some more, for videos we actually are creating a new Geometry so we technically do want to clean that up. We actually don't need a new geometry for each media frame and I am fine with statically keeping that around for the lifetime of the application so we shouldn't need to clean that up. Will add a note above about that as well.

src/systems/bit-media-frames.js
```diff
@@ -133,89 +157,152 @@ const snapToFrame = (() => {
...
+  }
+
+  // Audios can't be cloned so we take a different path for them
+  if (isVideo) {
+    const previewMaterial = new THREE.MeshBasicMaterial();
+    previewMaterial.side = THREE.DoubleSide;
+    previewMaterial.transparent = true;
+    previewMaterial.opacity = 0.5;
+
+    const geometry = createPlaneBufferGeometry(1, 1, 1, 1, TEXTURES_FLIP_Y);
```
netpro2k, line 206: I noted in the cleanup code, we would technically need to clean this up (since we are the ones creating a disposable resource), but we *don't* want to dipose of the `previewMesh`'s geometry for the other paths (since we are just referencing them. Since we don't actually mutate this geometry at all we can just statically define it once and share it among all media frames, and never `dispose()` of it.

netpro2k (commented):
This is looking good. I am happy to see the media frame system didn't need to change all that much and most of the code changes there were are just around generating the preview and getting the object bounds (which is what you would expect)

I made a bunch of little suggestions and observations, but none are sticking out to me as extreme blockers. (I think the disposal stuff probably sticks out the most, but its mostly an issue of "correctness", since in reality media frames are very rarely removed except for at scene transition time when the objects they are previewing are also going away).

The object bounding box stuff still strikes me as incredibly more complicated than I would expect (not a slight on your code, its happened ever time we try and write code like it). I have a very strong suspicion we are just tripping over ourselves because of some underlying complexity, and my best bet is that its in shape generation code in the physics system. When I think back I sort of remember the "just write the obvious thing" version of that code working except "but the physics shape doesn't line up now"... Been a long time since I looked at it though so I could be miss-remembering. I think once we are down to 1 codepath for all this stuff and we have had time to rethink physics a bit, its worth a revisit.

src/bit-systems/media-loading.ts
keianhzo: Right, I think I started using this for storing something different but ended being just a vector. I've refactored this to its own component.

src/bit-systems/media-loading.ts
keianhzo: Updated

src/bit-systems/media-loading.ts
keianhzo: Opps, yep, updated.

src/bit-systems/media-loading.ts
keianhzo: Right, it would be harmless but we don't need to add `MediaContentBounds` in that case.

src/systems/bit-media-frames.js
```diff
@@ -133,89 +157,152 @@ const snapToFrame = (() => {
...
 
   const cloneObj = new THREE.Group();
-  cloneObj.el = el; // We rely on media-loader component for bounds
-  cloneObj.add(meshClone);
-  AFRAME.scenes[0].object3D.add(cloneObj);
+  cloneObj.add(previewMesh);
+  APP.world.scene.add(cloneObj);
 
-  return addObject3DComponent(world, addEntity(world), cloneObj);
+  hasComponent(world, AEntity, capturable) && (cloneObj.el = el); // We rely on media-loader component for bounds
```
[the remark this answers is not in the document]
keianhzo: I generally use it for simple conditions like:
```
A & B;
```
to avoid braces clutter. I don't have a strong opinion about it, if people find that makes the code hard to read I'm happy to stop using it.

src/systems/bit-media-frames.js
keianhzo: Removed as it just adds unnecessary complexity.

src/bit-systems/media-loading.ts
```diff
@@ -176,8 +210,11 @@ function* loadAndAnimateMedia(world: HubsWorld, eid: EntityID, clearRollbacks: C
   removeComponent(world, MediaLoader, eid);
 
   if (media) {
+    const box = getBox(world, eid, media);
+    MediaContentBounds.set(media, box.getSize(new Vector3()));
     // TODO update scale?
-    inflatePhysicsShape(world, media, {
+    removeComponent(world, PhysicsShape, eid);
```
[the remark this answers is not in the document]
keianhzo: Yeah, that doesn't make sense, maybe it's a leftover by mistake. Removing.

src/systems/physics-system.js
```diff
@@ -241,6 +243,12 @@ export class PhysicsSystem {
     return bodyId;
   }
 
+  updateRigidBody(eid, options) {
```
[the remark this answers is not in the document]
keianhzo: Yep, I think I moved most of the methods but I didn't update the old `body-helper.js` one. I've removed the old `updateBody` and `updateBodyOptions` methods.

I'd like to revisit this as mentioned in https://github.com/mozilla/hubs/pull/5978#discussion_r1142076303. I think we can get rid of `bit-physics.js` as it's acting only as a forward layer. We should always call the physics system add/update methods for an entity and update the underlying components from there.

src/inflators/media-frame.js
```diff
@@ -51,9 +51,8 @@ export function inflateMediaFrame(world, eid, componentProps) {
     })
   );
   guide.layers.set(Layers.CAMERA_LAYER_UI);
-  // TODO: This is a hack around the physics system addBody call requiring its body to have parent
-  guide.parent = new Group();
```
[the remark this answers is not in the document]
keianhzo: Now we add the guide to the media frame object on `enteredMediaFramesQuery`. As the media frame is already a group we don't need to create an extra node.

src/systems/bit-media-frames.js
```diff
@@ -297,26 +399,30 @@ export function mediaFramesSystem(world) {
...
       setMatrixScale(world.eid2obj.get(captured), MediaFrame.scale[frame]);
-      physicsSystem.updateBodyOptions(Rigidbody.bodyId[captured], { type: "dynamic" });
+      physicsSystem.updateBodyOptions(captured, { type: "dynamic" });
     }
 
     MediaFrame.capturedNid[frame] = NetworkedMediaFrame.capturedNid[frame];
     MediaFrame.scale[frame].set(NetworkedMediaFrame.scale[frame]);
 
-    display(world, frame, heldMediaTypes);
+    frame2mixer.forEach(mixer => {
```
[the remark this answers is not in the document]
keianhzo: I've got rid of this for now.

src/systems/bit-media-frames.js
```diff
@@ -102,26 +114,38 @@ function isColliding(world, eidA, eidB) {
...
   const frameQuat = new THREE.Quaternion();
   const frameScale = new THREE.Vector3();
   const m4 = new THREE.Matrix4();
 
-  function scaleForAspectFit(containerSize, itemSize) {
-    return Math.min(containerSize[0] / itemSize.x, containerSize[1] / itemSize.y, containerSize[2] / itemSize.z);
-  }
-
-  return function snapToFrame(world, frame, target) {
+  return (world, frame, target, contentBounds) => {
```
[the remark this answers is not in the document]
keianhzo: Updated

src/systems/bit-media-frames.js
```diff
@@ -264,16 +361,19 @@ export function mediaFramesSystem(world) {
     const frame = mediaFrames[i];
 
     const captured = world.nid2eid.get(MediaFrame.capturedNid[frame]) || 0;
-    const colliding = captured && isColliding(world, frame, captured);
-
-    if (captured && hasComponent(world, Owned, captured) && !hasComponent(world, Held, captured) && colliding) {
-      snapToFrame(world, frame, captured);
-      physicsSystem.updateBodyOptions(Rigidbody.bodyId[captured], { type: "kinematic" });
+    const isCapturedOwned = hasComponent(world, Owned, captured);
```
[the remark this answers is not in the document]
keianhzo: Updated to try to improve readability.

src/systems/bit-media-frames.js
keianhzo: This should only be disposing the guide as the preview is transient as is removed when hidden so it shouldn't be affected by the cleanup. But you are right, the video geometry is not being cleaned up correctly. I didn't want to preallocate that to avoid allocating resources for something that can potentially never be used but maybe that's better that disposing it every time a preview is hidden so I'm going o do that.

PR #6018, opened 2023-04-04
Do not resize model loaded as part of the scene using the model component

keianhzo:
We are resizing objects that have a media component attached (image/video(model) and that causes issues as scene designers expect those objects to stay untouched.

Resizing makes sense for dynamically spawned objects but not so much for scene objects.

We were already not resizing image/video component in bitECS but model was still being resized, I've changed that. I've also updated the related AFrame components to behave the same.

Closes https://github.com/MozillaReality/hubs-blender-exporter/issues/104

### DIFF
--- a/src/gltf-component-mappings.js
+++ b/src/gltf-component-mappings.js
@@ -183,7 +183,7 @@ AFRAME.GLTFModelPlus.registerComponent("media", "media", (el, componentName, com
   }
 });
 
-async function mediaInflator(el, componentName, componentData, components) {
+async function mediaInflator(el, componentName, componentData, components, fitToBox = true) {
   let isControlled = true;
 
   if (componentName === "link" && (components.video || components.image)) {
@@ -266,7 +266,7 @@ async function mediaInflator(el, componentName, componentData, components) {
 
   el.setAttribute("media-loader", {
     src: sanitizeUrl(src),
-    fitToBox: true,
+    fitToBox,
     resolve: true,
     fileIsOwned: true,
     animate: false,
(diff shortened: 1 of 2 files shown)


takahirox (approved):
As I mentioned in the internal chat, I think this change is ok to me unless there is any special and historical reasons for auto resizing that I don't know. It makes sense that artists don't like auto resizing.

As I said there, it may be good to apply clamping to all the medias that clamps the size of too small (eg: 0.00001) or too huge (eg: 10000) sized model into a certain range as a guard protecting from invisible object (too small model) or broken scene (too hoge model). This clamping does't have an effect to regular sized models. It can be implemented in another PR if sounds good.

keianhzo:
Not sure if you refer to that but spawned media is already being clamped when fit-to-box is applied so it should never be too large/small. This PR only affects scene elements as designed in Spoke/Blender but maybe @johnshaughnessy @netpro2k have more background info about this.

takahirox:
My undrestanding is, this change is for model component, isn't it? I though thete might be a malicious person that creates a too huge/small asset, creates another asset including model component that points to the too huge/small asset, and uploads by drag&drop.

keianhzo:
Yeah, that could happen but you can also just upload the regular asset and scale it through the UI scale button. Not sure what implications limiting the spawnable assets scaling have but it sounds good to me that we clamp it both when spawning and also when scaling manually through UI.

takahirox:
Right, it is ideal if we can always check the size and scale and apply clamping when needed.

Anyways, my clamping idea doesn't block this PR. We can make another PR if the idea sounds good.

takahirox (approved):
The change in the inflator looks good to me.

I'm not really sure about AFrame code so another reviewer may be needed.

PR #6022, opened 2023-04-06
Bitecs-scene-preview-camera

keianhzo:
This PR adds support for scene preview camera in bitECS for feature parity.

Initially I thought about adding full glTF support so we can treat preview cameras same way as we do any other component as that's something that was missing in the AFrame component. Then I thoguht that it might introduce a potential preview camera hijacking risk if someone spawns a component with a scene-preview-camera component so I removed the glTF support and I have kept the functionality as it is in AFrame.


src/bit-systems/scene-loading.ts
```diff
@@ -81,7 +90,23 @@ function* loadScene(
     }
     environmentSystem.updateEnvironmentSettings(envSettings);
 
-    const sceneEl = AFRAME.scenes[0];
+    const cameraNode = (document.getElementById("scene-preview-node") as AElement).object3D!;
+    removeComponent(world, ScenePreviewCamera, cameraNode.eid!);
```
takahirox, line 94: This comment doesn't block this PR but is for future clean up.

Maybe using query to get camera node and its eid may be more consistent with other systems?

```
// in bit-components
export const ScenePreviewNode = defineComponent();

// in system
const query = defineQuery([ScenePreviewNode]);
// Assume there is a single entity having ScenePreviewNode component
const cameraNodeEid = query(world)[0];
const cameraNode = world.eid2obj.get(cameraNodeEid)!;
```

src/bit-systems/scene-loading.ts
```diff
@@ -81,7 +90,23 @@ function* loadScene(
...
+    const sceneObj = world.eid2obj.get(scene)!;
+    const previewCamera = sceneObj.getObjectByName("scene-preview-camera");
```
takahirox, line 96: For future clean up, similar to the above. We may use query and component to access preview camera.

src/bit-systems/scene-loading.ts
```diff
@@ -81,7 +90,23 @@ function* loadScene(
...
+      cameraNode.rotation.copy(previewCamera.rotation);
+      cameraNode.rotation.reorder("YXZ");
+    } else {
+      const cameraPos = cameraNode.position;
+      cameraNode.position.set(cameraPos.x, 2.5, cameraPos.z);
+    }
+    cameraNode.matrixNeedsUpdate = true;
+    addComponent(world, ScenePreviewCamera, cameraNode.eid!);
+    ScenePreviewCamera.duration[cameraNode.eid!] = 60;
+    ScenePreviewCamera.positionOnly[cameraNode.eid!] = 1;
```
takahirox, line 108: Probably I just don't understand the code yet... What is this "remove and then add ScenePreviewCamera component" for?

src/bit-systems/scene-preview-camera-system.ts
```diff
@@ -0,0 +1,119 @@
...
+import { lerp } from "three/src/math/MathUtils";
+
+const startPositions = new Map<number, Vector3>();
+const startRotations = new Map<number, Quaternion>();
+const targetPositions = new Map<number, Vector3>();
+const targetRotations = new Map<number, Quaternion>();
+const startTimes = new Map<number, number>();
+const backwards = new Map<number, boolean>();
+const ranOnePass = new Map<number, boolean>();
+const newRot = new Quaternion();
```
takahirox, line 17: This comment doesn't block the PR but is for future discussion.

Perhaps it would be good if we can define a standard pattern about where working variables (or states) required by a system are stored for the consistency. Ex: In a system file, in component, or another working data component. I added this topic to the internal design document.

src/bit-systems/scene-preview-camera-system.ts
```diff
@@ -0,0 +1,119 @@
...
+    startPositions.delete(eid);
+    startRotations.delete(eid);
+    targetPositions.delete(eid);
+    targetRotations.delete(eid);
+    startTimes.delete(eid);
+    backwards.delete(eid);
+    ranOnePass.delete(eid);
+  });
+  scenePreviewCameraEnterQuery(world).forEach(eid => {
+    const previewCamera = world.eid2obj.get(eid)!;
```
takahirox, line 82: This comment doesn't block the PR but is for future clean up.

If a system accesses Object3D, adding Object3DTag component to query would be safer (and less confusing for new readers)?

```typescript
const scenePreviewCameraQuery = defineQuery([ScenePreviewCamera, Object3DTag]);
```

src/bit-systems/scene-preview-camera-system.ts
```diff
@@ -0,0 +1,119 @@
...
+      backwards.set(eid, !back);
+      startTimes.set(eid, performance.now());
+    }
+  }
+};
+
+const scenePreviewCameraQuery = defineQuery([ScenePreviewCamera]);
+const scenePreviewCameraEnterQuery = enterQuery(scenePreviewCameraQuery);
+const scenePreviewCameraExitQuery = exitQuery(scenePreviewCameraQuery);
+export function scenePreviewCameraSystem(world: HubsWorld, cameraSystem: CameraSystem) {
```
takahirox, line 71: This comment doesn't block the PR but is for future clean up.

Ideally a system shouldn't have a dependency with another system (CameraSystem)? We may add TODO comment for now.

src/bit-systems/scene-preview-camera-system.ts
```diff
@@ -0,0 +1,119 @@
...
+  const streamerCamera = getStreamerCamera();
+  if (streamerCamera) {
+    setMatrixWorld(previewCamera, streamerCamera.matrixWorld);
+    // Move camera forward just a bit so that we don't see the avatar's eye cylinders.
+    previewCamera.translateZ(-0.1);
+    previewCamera.matrixNeedsUpdate = true;
+  } else {
+    const startTime = startTimes.get(eid)!;
+    const duration = ScenePreviewCamera.duration[eid];
+    let t = (performance.now() - startTime) / (1000.0 * duration);
```
takahirox, line 30: `performance.now()` may be replaced with `world.time.elapsed` where `world` is `HubsWorld`.

takahirox (approved):
I mostly commented from the view of ECS design. Most of the comments are for future clean up or future discussion.

`duration` variable type and using `world.time.elapsed` instead of `performance.now()` are only the changes required by me (if the suggestions look good).

Honestly I don't deeply look though the main preview camera logic (and I don't have time to do it right now). I assume it just basically follows the A-Frame one. Another reviewer may be needed for the core logic.

src/bit-systems/scene-preview-camera-system.ts
```diff
@@ -0,0 +1,119 @@
...
+    t = Math.min(1.0, Math.max(0.0, t));
+
+    if (!ranOnePass.get(eid)) {
+      t = t * (2 - t);
+    } else {
+      t = t < 0.5 ? 2 * t * t : -1 + (4 - 2 * t) * t;
+    }
```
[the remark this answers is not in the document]
keianhzo: If only I knew... This comes from https://github.com/mozilla/hubs/blob/5fcdf74a169c83077390deab88b409609bb4eee7/src/components/scene-preview-camera.js#L75 but I don't really have much idea of why those numbers.

src/bit-systems/scene-preview-camera-system.ts
keianhzo: yep, that makes sense. Thanks.

src/bit-systems/scene-preview-camera-system.ts
keianhzo: Added TODO to decouple that when it's migrated.

src/bit-systems/scene-loading.ts
keianhzo: This component/system it's a bit special. We don't really add the `scene-preview-camera` component in the glTF, we add a node with `scene-preview-camera` name (not component). When the scene is loading we get that node using `getObjectByName` and we add the `ScenePreviewCamera` component to a pre-exiting scene node (`scene-preview-node`). 

Every time a scene is loaded we add the `ScenePreviewCamera` component to the `scene-preview-node` node. We use the remove/add so we can trigger the exit/enter queries and cleanup the previous preview camera component state.

src/bit-systems/scene-preview-camera-system.ts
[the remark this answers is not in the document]
takahirox: We may add a comment "TODO: Write comment about this formula" for now? Anyways, not a big deal.

PR #6024, opened 2023-04-07
Add interpolation to bitECS network transform.

takahirox:
This PR adds easy linear interpolation to bitECS network transform.

Changes
* Add LinearTranslate/Rotate/Scale components
* Add linearTransform system to handle the added components
* network-transform sets the component when receiving data from remote
* Move millisecondsBetweenTicks from network-send-system to networking because it is referred from two places

### DIFF
--- a/src/bit-systems/linear-transform.ts
+++ b/src/bit-systems/linear-transform.ts
@@ -1,0 +1,72 @@
+import { defineQuery, removeComponent } from "bitecs";
+import { Quaternion, Vector3 } from "three";
+import { LinearRotate, LinearScale, LinearTranslate, Object3DTag } from "../bit-components";
+import { HubsWorld } from "../app";
+
+// Working variables
+const _vec3 = new Vector3();
+const _quat = new Quaternion();
+
+const linearTranslateQuery = defineQuery([LinearTranslate, Object3DTag]);
+const linearRotateQuery = defineQuery([LinearRotate, Object3DTag]);
+const linearScaleQuery = defineQuery([LinearScale, Object3DTag]);
+
+export function linearTransformSystem(world: HubsWorld) {
+  linearTranslateQuery(world).forEach(eid => {
+    const duration = LinearTranslate.duration[eid];
+    const targetX = LinearTranslate.targetX[eid];
+    const targetY = LinearTranslate.targetY[eid];
+    const targetZ = LinearTranslate.targetZ[eid];
+
+    const deltaTime = world.time.delta;
+    const obj = world.eid2obj.get(eid)!;
+
+    if (deltaTime >= duration) {
+      obj.position.set(targetX, targetY, targetZ);
+      removeComponent(world, LinearTranslate, eid);
+    } else {
+      obj.position.lerp(_vec3.set(targetX, targetY, targetZ), deltaTime / duration);
+      LinearTranslate.duration[eid] = duration - deltaTime;
+    }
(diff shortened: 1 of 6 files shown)


src/bit-systems/linear-transform.ts
```diff
@@ -0,0 +1,72 @@
...
+    const destX = LinearTranslate.destinationX[eid];
+    const destY = LinearTranslate.destinationY[eid];
+    const destZ = LinearTranslate.destinationZ[eid];
+
+    const deltaTime = world.time.delta;
+    const obj = world.eid2obj.get(eid)!;
+
+    if (deltaTime >= duration) {
+      obj.position.set(destX, destY, destZ);
+      removeComponent(world, LinearTranslate, eid);
```
keianhzo, line 26: Feels a bit strange that we are removing the component inside it's own query. Wouldn't it make more sense to remove it in `networked-transform.ts`?

```typescript
      tmpVec.fromArray(NetworkedTransform.position[eid]);
      if (!tmpVec.near(obj.position)) {
        addComponent(world, LinearTranslate, eid);
        LinearTranslate.duration[eid] = millisecondsBetweenTicks;
        LinearTranslate.destinationX[eid] = tmpVec.x;
        LinearTranslate.destinationY[eid] = tmpVec.y;
        LinearTranslate.destinationZ[eid] = tmpVec.z;
      } else {
        removeComponent(world, LinearTranslate, eid);
      }
```

and let this system do just what is responsible for, interpolating:

```typescript
linearTranslateQuery(world).forEach(eid => {
    const duration = LinearTranslate.duration[eid];
    const destX = LinearTranslate.destinationX[eid];
    const destY = LinearTranslate.destinationY[eid];
    const destZ = LinearTranslate.destinationZ[eid];

    const deltaTime = world.time.delta;
    const obj = world.eid2obj.get(eid)!;

    obj.position.lerp(_vec3.set(destX, destY, destZ), deltaTime / duration);
    LinearTranslate.duration[eid] = duration - deltaTime;
    obj.matrixNeedsUpdate = true;
  });
```

takahirox: I don't think it is so strange to remove a component inside a query especially if it does operations across frames and finishes at specified time or event. MediaLoader does it, too.

And LinearTranslate/Rotate/Scale and their system are designed as common components and systems that linearly (against time) update object's transform to target values using the specified duration time. I think removing component in that system is natural and makes them decoupled from other components and systems.

But if removing a component in its query can cause a bug in certain scenarios, we may need to revisit and write it in the internal design document. Would you please share if you have?

src/systems/networked-transform.js
```diff
@@ -27,20 +28,30 @@ export function networkedTransformSystem(world) {
     } else {
       tmpVec.fromArray(NetworkedTransform.position[eid]);
       if (!tmpVec.near(obj.position)) {
-        obj.position.copy(tmpVec);
-        obj.matrixNeedsUpdate = true;
+        addComponent(world, LinearTranslate, eid);
+        LinearTranslate.duration[eid] = millisecondsBetweenTicks;
+        LinearTranslate.destinationX[eid] = tmpVec.x;
+        LinearTranslate.destinationY[eid] = tmpVec.y;
+        LinearTranslate.destinationZ[eid] = tmpVec.z;
```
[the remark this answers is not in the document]
takahirox: No strong preference about it. Updated, thanks.

keianhzo (approved):

PR #6030, opened 2023-04-12
Locked down demo mvp

mikemorran:
Implementing time-boxed version of simplified demo UI to be released internally. This feature does the following for any hub_id specified as demo room in the admin panel...

- Bypass profile setup step of entry flow (avatar and display name selection)
- Force individual profile settings for default avatar, random name generation, and always hide nametag
- Disable UI elements for...
1. Presence Log when new users enter and leave
2. Chat button
3. Share button
4. Place button
5. Microphone button
6. Object Menu
7. People Menu hover and clickability
8. "Change Name and Avatar" in the more menu
- Disable "T" and "/" hotkeys which open chat
- Prevent separate tab avatar changes with avatar.js

Public release of this feature will occur once room info and settings for a demo room can be programmatically specified in the admin panel. Also, additional edit should be implemented to support an array of demo room hub_ids.

**Specific Feedback Requested:**
@keianhzo I was unable to solve checking for `canVoiceChat` before our hack to request mic when using Safari. This is because the hack requests permissions _before_ we ask reticulum for the hub_id or preferences. Any input on why we request mic permissions so early?

@keianhzo I implemented a pretty sloppy hack in `store.js` to prevent users from updating their avatarId from separate tabs using the avatar.js panel. Any thoughts about how this could be better accomplished, especially since it is using AFRAME?

@nikk15 @nickgrato Any thoughts on how I removed pointer-events on the People Menu in `ui-root.js`, `src/react-components/room/ContentMenu.js`, and `src/react-components/room/ContentMenu.scss`? Could this be improved?

### DIFF
--- a/src/hub.js
+++ b/src/hub.js
@@ -459,12 +459,6 @@ export async function updateEnvironmentForHub(hub, entryManager) {
 
         //TODO: check if the environment was made with spoke to determine if a shape should be added
         traverseMeshesAndAddShapes(environmentEl);
-
-        if (isLockedDownDemoRoom()) {
-          const avatarRig = document.querySelector("#avatar-rig");
-          const avatarId = await fetchRandomDefaultAvatarId();
-          avatarRig.setAttribute("player-info", { avatarSrc: await getAvatarSrc(avatarId) });
-        }
       },
       { once: true }
     );
@@ -502,12 +496,6 @@ export async function updateEnvironmentForHub(hub, entryManager) {
             }
 
             const fader = document.getElementById("viewing-camera").components["fader"];
-
-            if (isLockedDownDemoRoom()) {
-              const avatarRig = document.querySelector("#avatar-rig");
-              const avatarId = await fetchRandomDefaultAvatarId();
-              avatarRig.setAttribute("player-info", { avatarSrc: await getAvatarSrc(avatarId) });
-            }
 
             // Add a slight delay before de-in to reduce hitching.
             setTimeout(() => fader.fadeIn(), 2000);
@@ -847,6 +835,18 @@ document.addEventListener("DOMContentLoaded", async () => {
       () => hubChannel.updateScene(sceneInfo),
       SignInMessages.changeScene
     );
+  });
+
+  scene.addEventListener("hub_updated", async () => {
+    if (isLockedDownDemoRoom()) {
+      const avatarRig = document.querySelector("#avatar-rig");
+      const avatarId = await fetchRandomDefaultAvatarId();
+      avatarRig.setAttribute("player-info", { avatarSrc: await getAvatarSrc(avatarId) });
+    } else {
+      if (scene.is("entered")) {
+        entryManager._setPlayerInfoFromProfile(true);
+      }
+    }
   });
 
   remountUI({
--- a/src/hub.js
+++ b/src/hub.js
@@ -445,7 +445,7 @@ export async function updateEnvironmentForHub(hub, entryManager) {
 
     environmentEl.addEventListener(
       "model-loaded",
-      async () => {
+      () => {
         environmentEl.removeEventListener("model-error", sceneErrorHandler);
 
         console.log(`Scene file initial load took ${Math.round(performance.now() - loadStart)}ms`);
@@ -481,7 +481,7 @@ export async function updateEnvironmentForHub(hub, entryManager) {
       () => {
         environmentEl.addEventListener(
           "model-loaded",
-          async () => {
+          () => {
             environmentEl.removeEventListener("model-error", sceneErrorHandler);
 
             envSystem.updateEnvironment(environmentEl);
--- a/src/hub.js
+++ b/src/hub.js
@@ -159,6 +159,7 @@ import MessageDispatch from "./message-dispatch";
 import SceneEntryManager from "./scene-entry-manager";
 import Subscriptions from "./subscriptions";
 import { createInWorldLogMessage } from "./react-components/chat-message";
+import { fetchRandomDefaultAvatarId } from "./utils/identity.js";
 
 import "./systems/nav";
 import "./systems/frame-scheduler";
@@ -444,7 +445,7 @@ export async function updateEnvironmentForHub(hub, entryManager) {
 
     environmentEl.addEventListener(
       "model-loaded",
-      () => {
+      async () => {
         environmentEl.removeEventListener("model-error", sceneErrorHandler);
 
         console.log(`Scene file initial load took ${Math.round(performance.now() - loadStart)}ms`);
@@ -458,6 +459,12 @@ export async function updateEnvironmentForHub(hub, entryManager) {
 
         //TODO: check if the environment was made with spoke to determine if a shape should be added
         traverseMeshesAndAddShapes(environmentEl);
+
+        if (isLockedDownDemoRoom()) {
+          const avatarRig = document.querySelector("#avatar-rig");
+          const avatarId = await fetchRandomDefaultAvatarId();
+          avatarRig.setAttribute("player-info", { avatarSrc: await getAvatarSrc(avatarId) });
+        }
       },
       { once: true }
     );
@@ -480,7 +487,7 @@ export async function updateEnvironmentForHub(hub, entryManager) {
       () => {
         environmentEl.addEventListener(
           "model-loaded",
-          () => {
+          async () => {
             environmentEl.removeEventListener("model-error", sceneErrorHandler);
 
             envSystem.updateEnvironment(environmentEl);
@@ -495,6 +502,12 @@ export async function updateEnvironmentForHub(hub, entryManager) {
             }
 
             const fader = document.getElementById("viewing-camera").components["fader"];
+
+            if (isLockedDownDemoRoom()) {
+              const avatarRig = document.querySelector("#avatar-rig");
+              const avatarId = await fetchRandomDefaultAvatarId();
+              avatarRig.setAttribute("player-info", { avatarSrc: await getAvatarSrc(avatarId) });
+            }
 
             // Add a slight delay before de-in to reduce hitching.
             setTimeout(() => fader.fadeIn(), 2000);
@@ -765,7 +778,7 @@ document.addEventListener("DOMContentLoaded", async () => {
 
   // If the stored avatar doesn't have a valid src, reset to a legacy avatar.
   const avatarSrc = await getAvatarSrc(store.state.profile.avatarId);
-  if (!avatarSrc || isLockedDownDemoRoom(hubId)) {
+  if (!avatarSrc) {
     await store.resetToRandomDefaultAvatar();
   }
 
--- a/src/hub.js
+++ b/src/hub.js
@@ -764,7 +764,7 @@ document.addEventListener("DOMContentLoaded", async () => {
 
   // If the stored avatar doesn't have a valid src, reset to a legacy avatar.
   const avatarSrc = await getAvatarSrc(store.state.profile.avatarId);
-  if (!avatarSrc) {
+  if (!avatarSrc || configs.feature("is_locked_down_demo_room") === hubId) {
     await store.resetToRandomDefaultAvatar();
   }
 
--- a/src/hub.js
+++ b/src/hub.js
@@ -2,7 +2,8 @@ import {
   getCurrentHubId,
   updateVRHudPresenceCount,
   updateSceneCopresentState,
-  createHubChannelParams
+  createHubChannelParams,
+  isLockedDownDemoRoom
 } from "./utils/hub-utils";
 import "./utils/debug-log";
 import configs from "./utils/configs";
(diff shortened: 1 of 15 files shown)


src/react-components/ui-root.js
```diff
@@ -817,7 +817,11 @@ class UIRoot extends Component {
 
   renderEntryStartPanel = () => {
     const { hasAcceptedProfile, hasChangedName } = this.props.store.state.activity;
-    const promptForNameAndAvatarBeforeEntry = this.props.hubIsBound ? !hasAcceptedProfile : !hasChangedName;
+    //const isLockedDownDemo = configs.feature("is_locked_down_demo_room") === this.props.hub.hub_id;
+    const isLockedDownDemo = true;
```
nikk15, line 821: Reminder to remove this line and uncomment the above before merging :)

src/react-components/ui-root.js
```diff
@@ -1391,7 +1397,8 @@ class UIRoot extends Component {
                         )}
                         <PeopleMenuButton
                           active={this.state.sidebarId === "people"}
-                          onClick={() => this.toggleSidebar("people")}
+                          disabled={isLockedDownDemo}
```
nikk15, line 1400: This looks appropriate to me. @nickgrato ?

nikk15 (dismissed):
I think installing the Prettier extension in VSCode should solve some of the linting issues too. I have my VSCode settings set to "format on save"  :)

src/react-components/room/ContentMenu.js
```diff
@@ -7,9 +7,9 @@ import { ReactComponent as ObjectsIcon } from "../icons/Objects.svg";
 import { ReactComponent as PeopleIcon } from "../icons/People.svg";
 import { FormattedMessage } from "react-intl";
 
-export function ContentMenuButton({ active, children, ...props }) {
+export function ContentMenuButton({ active, disabled, children, ...props }) {
```
nikk15, line 10: You'll wanted to add disabled to the proptypes on line 18. Ideally we want to move to Typescript to solve this, but this is a pretty big file so not sure its worth it for this project. PropType should suffice for now.
eg.
`disabled: PropTypes.bool,`

src/react-components/preferences-screen.js
```diff
@@ -22,6 +22,8 @@ import {
   setMaxResolution
 } from "../utils/screen-orientation-utils";
 import { AAModes } from "../constants";
+import configs from "../utils/configs";
+import { getCurrentHubId } from "../utils/hub-utils";
```
nickgrato, line 26: as far as syntax goes I don't see any issues in the pr the only thing i'd suggest is :

There are quite a few files  where you are importing the same methods "config" and "getCurrentHubId"

I think since hub-utils already imports config I would add a function in hub-utils like

```
export function isLockedDownDemoRoom(id){
  hubId = id ? id : getCurrentHubId();
  return configs.feature("is_locked_down_demo_room") === hubId
}
```

and then you can use  isLockedDownDemoRoom() or isLockedDownDemoRoom(someID)

src/react-components/ui-root.js
nickgrato: @nikk15 are you asking about the disabled prop?

src/react-components/ui-root.js
mikemorran: Thank you for catching this!

src/react-components/preferences-screen.js
mikemorran: Great idea!

src/react-components/room/ContentMenu.js
mikemorran: Amazing! Thanks for catching this as well!

src/react-components/ui-root.js
nickgrato: yea looks like it drills down to the <button> eventually and disabled is an accurate button attr  🤷‍♂️

keianhzo (changes requested):
This is close to be ready. I have left a few comments, let me know if you need more context.

src/react-components/room/ChatSidebar.js
```diff
@@ -279,6 +280,9 @@ const logMessages = defineMessages({
 
 // TODO: use react-intl's defineMessages to get proper extraction
 export function formatSystemMessage(entry, intl) {
+  if (isLockedDownDemoRoom()) {
```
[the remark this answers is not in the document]
mikemorran: Great idea!

src/react-components/ui-root.js
```diff
@@ -817,7 +818,10 @@ class UIRoot extends Component {
 
   renderEntryStartPanel = () => {
     const { hasAcceptedProfile, hasChangedName } = this.props.store.state.activity;
-    const promptForNameAndAvatarBeforeEntry = this.props.hubIsBound ? !hasAcceptedProfile : !hasChangedName;
+    const isLockedDownDemo = isLockedDownDemoRoom(this.props.hub.hub_id);
```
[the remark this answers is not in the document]
mikemorran: My thinking is that it is still important to be able to test your speaker output in the entry flow. I might ask Jacob if this is truly a priority, but I anticipate skipping the audio step will lead to difficulty troubleshooting audio experiences.

src/react-components/preferences-screen.js
```diff
@@ -1210,55 +1211,63 @@ class PreferencesScreen extends Component {
...
-          },
-          {
-            key: "nametagVisibilityDistance",
-            prefType: PREFERENCE_LIST_ITEM_TYPE.NUMBER_WITH_RANGE,
-            min: 1,
-            max: 20,
-            step: 1,
-            digits: 2
-          },
+          ...(!isLockedDownDemoRoom()
```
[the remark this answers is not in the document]
mikemorran: Thanks for the catch!

src/react-components/room/ContentMenu.scss
```diff
@@ -72,3 +72,12 @@
...
   background-color: theme.$border1-color;
 }
+
+:local(.disabled) {
+  pointer-events: none;
+  background-color: theme.$transparent;
+
+  svg {
+    color: theme.$text5-color;
+  }
```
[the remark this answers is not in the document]
mikemorran: Gotcha!

src/systems/ui-hotkeys.js
```diff
@@ -23,11 +24,11 @@ AFRAME.registerSystem("ui-hotkeys", {
       this.userinput = this.el.systems.userinput;
     }
 
-    if (this.userinput.get(paths.actions.focusChat)) {
+    if (this.userinput.get(paths.actions.focusChat) && !isLockedDownDemoRoom()) {
```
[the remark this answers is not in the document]
mikemorran: Good call!

mikemorran:
@keianhzo Re: hack to prevent avatar selection in a separate tab.

I am attempting to allow default avatar selection on initial connection, not avatar selection in the entry flow, before disabling any avatar changes. This is important to force all users joining a demo room to choose from a default avatar and not bring in an avatar from their previous session in a different room. Since avatar.js does not specify that avatarID was changed in a separate tab, I am doing this weird hack to only allow avatarID changes on initial connection to a demo room (hence trying to judge when the scene is loaded). Happy to discuss further!

src/react-components/ui-root.js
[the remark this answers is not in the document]
keianhzo: ok, I wasn't sure about what was the expected behavior there. In that case it might be a good idea to disable the audio input setup. That will disable the mic permissions trigger. We can do that here:

https://github.com/mozilla/hubs/blob/7518e63052521b07c302f42bf7c16121f276ddfc/src/react-components/room/MicSetupModalContainer.js#L38

mikemorran (commented):
Feedback implemented!

mikemorran:
> I think installing the Prettier extension in VSCode should solve some of the linting issues too. I have my VSCode settings set to "format on save" :)

Implemented

mikemorran:
Re-opening accidental close

src/storage/store.js
```diff
@@ -401,6 +402,20 @@ export default class Store extends EventTarget {
   }
 
   update(newState, mergeOpts) {
+    // Check if in demo room before changing avatar. This feature is meant to prevent any avatarID changes triggered from an avatar.js panel ONLY. "Change Name & Avatar" button in More Menu is disabled at a UI level.
```
keianhzo, line 405: Looking at this again, I don't think this is avoiding an avatar update. This is called by the `avatar.js` page and there we don't have any context about the current `hub_id`. We just update the store and that event is recived in any other open document that is using the same local storage origin.

It seems that we are processing that here:
https://github.com/mozilla/hubs/blob/b8f775f10da2fdd2b9515adc992050ca7c3cf986/src/scene-entry-manager.js#L180

And there we can correctly get the `hub_id` so I think we can revert the changes here and add that logic there.

keianhzo (changes requested):
I missed the avatar thing in my previous review. As soon as that's updated I think this is ready to go.

src/react-components/ui-root.js
```diff
@@ -1415,7 +1426,7 @@ class UIRoot extends Component {
                         }}
                       />
                     )}
-                    {this.state.sidebarId !== "chat" && this.props.hub && (
+                    {!isLockedDownDemo && this.state.sidebarId !== "chat" && this.props.hub && (
```
keianhzo, line 1429: Looking at this again, as we don't really want any messages, I think we should be good just remoing this tow `isLockDownDemo` from:
https://github.com/mozilla/hubs/blob/a87f4b937fb4fa3b8badf4f116eb8489b329d72d/src/react-components/ui-root.js#L1429
and
https://github.com/mozilla/hubs/blob/a87f4b937fb4fa3b8badf4f116eb8489b329d72d/src/react-components/ui-root.js#L1453

and move it to where we receive the messages:
https://github.com/mozilla/hubs/blob/a87f4b937fb4fa3b8badf4f116eb8489b329d72d/src/message-dispatch.js#L58-L61

We can just return there if it's a lock down demo:
```javascript
  receive(message) {
    if (isLockedDownDemoRoom()) return;

    this.addToPresenceLog(message);
    this.dispatchEvent(new CustomEvent("message", { detail: message }));
  }
```
so we minimize the changes surface.

src/storage/store.js
```diff
@@ -318,7 +319,7 @@ export default class Store extends EventTarget {
     }
 
     // Regenerate name to encourage users to change it.
-    if (!this.state.activity.hasChangedName) {
+    if (!this.state.activity.hasChangedName || isLockedDownDemoRoom()) {
```
keianhzo, line 322: I don't think we want to update the display name. We are not really showing this anywhere as we have removed the nametag, the logs and the people panel so we should be good with resetting this whole file.

src/hub.js
```diff
@@ -495,6 +503,12 @@ export async function updateEnvironmentForHub(hub, entryManager) {
 
             const fader = document.getElementById("viewing-camera").components["fader"];
 
+            if (isLockedDownDemoRoom()) {
```
keianhzo, line 506: Trying to find a better place for this, the best place I can think of is here after the `scene_media_selected` listener:
https://github.com/mozilla/hubs/blob/a87f4b937fb4fa3b8badf4f116eb8489b329d72d/src/hub.js#L852

We can add something there like:
```javascript
  scene.addEventListener("hub_updated", async () => {
    if (isLockedDownDemoRoom()) {
      const avatarRig = document.querySelector("#avatar-rig");
      const avatarId = await fetchRandomDefaultAvatarId();
      avatarRig.setAttribute("player-info", { avatarSrc: await getAvatarSrc(avatarId) });
    } else {
      if (scene.is("entered")) {
        entryManager._setPlayerInfoFromProfile(true);
      }
    }
  });
```

This will minimize the number of changes in the codebase a little bit.

keianhzo (changes requested):
I've made a couple more suggestion to try to minimize the changes impact. I think after this we should be good to go!

src/react-components/ui-root.js
mikemorran: Agreed! This is a more elegant solution!

src/storage/store.js
mikemorran: Agreed!

src/hub.js
mikemorran: Amazing! Agreed!

src/hub.js
```diff
@@ -443,7 +445,7 @@ export async function updateEnvironmentForHub(hub, entryManager) {
 
     environmentEl.addEventListener(
       "model-loaded",
-      () => {
+      async () => {
```
keianhzo, line 448: I think this was left from a previous change.

src/hub.js
```diff
@@ -479,7 +481,7 @@ export async function updateEnvironmentForHub(hub, entryManager) {
       () => {
         environmentEl.addEventListener(
           "model-loaded",
-          () => {
+          async () => {
```
keianhzo, line 484: Same for this

keianhzo (changes requested):
I think when we remove those two things that were left we are good.

src/hub.js
mikemorran: Good catch!

keianhzo (approved):
Nice! this looks ready to ship after a QA pass.

PR #6031, opened 2023-04-13
Reorder sidebar "people" list with "you" at the top.

nikk15:
Screenshot 2023-04-13 at 3 10 11 PM

Previously, "you" could be anywhere in a long list of guests. This fix moves "you" to the top so its easily visible.

NOTE: Typescript transition to come in a separate PR, as it affected multiple other components.


src/react-components/room/PeopleSidebar.js
```diff
@@ -101,7 +149,8 @@ export function PeopleSidebar({
   voiceChatEnabled,
   isMod
 }) {
-  const intl = useIntl();
+  const user = people.find(person => !!person.isMe);
```
nickgrato, line 152: I think this all could work fine, another take that might make the jsx less confusing is to alter the people array, something like this.  

```
const me = people.find(person => !!person.isMe);
filteredPeople = people.filter(person => !person.isMe);
filteredPeople.unshift(me);
```

and then you can just loop through filteredPeople

nickgrato (approved):
seems to be a build error that I'd figure out before but this file looks good to me!

mikemorran:
@nikk15 I was thinking about this feature as I hopped into the demo room today! It is super helpful!

I was also going to suggest ordering "In Room" users above "In Lobby" users. Often as an event moderator, I find myself wanting to identify these users first, since they typically are the ones I need to kick or mute. I think, right now, they are mixed in with "In Lobby" users

PR #6034, opened 2023-04-14
Upgrade Lilypad to v1.8.2

bryanenders:
Why
---
This removes the need for `--legacy-peer-deps` or `--force` during dependency installation


nickgrato (approved):
awesome!

PR #6036, opened 2023-04-14
Remove safari mic hack

keianhzo:
This hack is been there for a while but it seems to be fixed now. I have only tested in Safari 16.3 MacOS so it needs some more testing in iOS devices.

We might want to still leave it and remove it for only the latest current version or it's safe to assume that iOS/Mac devices are updated after a while after an update has been released?

### DIFF
--- a/src/hub.js
+++ b/src/hub.js
@@ -41,6 +41,7 @@ patchWebGLRenderingContext();
 
 import "networked-aframe/src/index";
 import "webrtc-adapter";
+import { detectOS, detect } from "detect-browser";
 import {
   getReticulumFetchUrl,
   getReticulumMeta,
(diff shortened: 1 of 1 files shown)


vincentfretin:
FYI I have a networked-aframe project using the janus sfu, I'm asking for mic permission after a click on a button. I didn't get issue reports related to mic volume dropping since at least iOS 15.2 and plus.

keianhzo:
Thanks for the info @vincentfretin! Safari 15.2 is from December 13, 2021. I haven't found any specific in the release notes that point to that being fixed but if we are confident that it was fixed after that version we could add the hack to versions previous to that one. I'm going to wait for @takahirox feedback as he's been working on this recently.

vincentfretin:
Oh I think this specific issue was fixed in some release of 14.x but don't take my word for it. I mention 15.2 because there were audio regressions in 15.0 and 15.1 releases too, I don't remember which issues exactly.

takahirox:
If I remember correctly, this hack is not I added. It seems to have been added more before.

Personally it's ok to remove the hack if we confirm that we don't need it on modern Safari versions.

takahirox (approved):

netpro2k:
We generally have tried to keep compat with last 2 major browser versions (and this is how we [configure our transpiler](https://github.com/mozilla/hubs/blob/master/.browserslistrc#L1)) so I think its safe to remove hacks that pertain to versions older than that.

PR #6042, opened 2023-04-19
Room newloader flag

netpro2k:
This PR allows enabling newLoader for a room by setting a flag in its user data. 

For example:
```js
fetchReticulumAuthenticated(`/api/v1/hubs/${APP.hub.hub_id}`, "PATCH", {
  id: APP.hub.hub_id,
  hub: {
    user_data: { hubsUseNewLoader: true }
  }
});
```

It also removes any checks for newLoader that were happening only at startup and converts them to runtime checks. The only exception is that the new video grabbing behavior is applied even when newLoader is on, since reworking input maps at runtime seems tricky and error prone. Since we plan to live with this behavior in newLaoder anyway it seems like a reasonable tradeoff. We will need to be careful when adding additional newLaoder checks that they correctly handle the fact that this can change per room now.

We can upgrade this from user_data to a proper room property if we find the need, but not needing a reticulum change means this is easier to deploy to arbitrary test instances. Also since this is planned to be a temporary flag avoids the additional database migration to remove the property later.

### DIFF
--- a/src/utils/bit-utils.ts
+++ b/src/utils/bit-utils.ts
@@ -52,5 +52,5 @@ export function findChildWithComponent(world: HubsWorld, component: Component, e
 
 const forceNewLoader = qsTruthy("newLoader");
 export function shouldUseNewLoader() {
-  return forceNewLoader;
+  return forceNewLoader || APP.hub?.user_data?.hubsUseNewLoader;
 }
--- a/src/utils/bit-utils.ts
+++ b/src/utils/bit-utils.ts
@@ -4,6 +4,7 @@ import { Object3D } from "three";
 import { HubsWorld } from "../app";
 import { findAncestor, traverseSome } from "./three-utils";
 import { EntityID } from "./networking-types";
+import qsTruthy from "./qs_truthy";
 
 export type ElOrEid = EntityID | AElement;
 
@@ -47,4 +48,9 @@ export function findChildWithComponent(world: HubsWorld, component: Component, e
     });
     return childEid;
   }
+}
+
+const forceNewLoader = qsTruthy("newLoader");
+export function shouldUseNewLoader() {
+  return forceNewLoader;
 }
--- a/src/app.ts
+++ b/src/app.ts
@@ -65,6 +65,7 @@ export function getScene() {
 
 interface HubDescription {
   hub_id: string;
+  user_data?: any;
 }
 
 export class App {
--- a/src/change-hub.js
+++ b/src/change-hub.js
@@ -2,7 +2,6 @@ import { getReticulumFetchUrl, hubUrl } from "./utils/phoenix-utils";
 import { updateEnvironmentForHub, getSceneUrlForHub, updateUIForHub, remountUI } from "./hub";
 import { loadLegacyRoomObjects } from "./utils/load-legacy-room-objects";
 import { loadSavedEntityStates } from "./utils/entity-state-utils";
-import qsTruthy from "./utils/qs_truthy";
 import { localClientID, pendingMessages, pendingParts } from "./bit-systems/networking";
 import { storedUpdates } from "./bit-systems/network-receive-system";
 import { shouldUseNewLoader } from "./utils/bit-utils";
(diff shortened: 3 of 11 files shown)


src/utils/bit-utils.ts
```diff
@@ -48,3 +49,8 @@ export function findChildWithComponent(world: HubsWorld, component: Component, e
     return childEid;
   }
 }
+
+const forceNewLoader = qsTruthy("newLoader");
```
takahirox, line 53: Is this for testing? @johnshaughnessy

takahirox:
> converts them to runtime checks

I came up with two questions. Is there a chance that `newLoader` flag can be flipped (after the first initialization)? I haven't read thgough the entire codebase yet but do we ensure that `shouldUseNewLoader()` is not called before the flag initialization?

src/utils/bit-utils.ts
johnshaughnessy: I assume so

takahirox:
`shouldUseNewLoader()` seems to can be called from https://github.com/mozilla/hubs/blob/5bebb4fefae3500490a2641e91f3881971d0490e/src/react-components/room/hooks/useRoomLoadingState.js#L90 before `APP.hub` initialization. It returns `false`, if it's called before the `APP.hub` initialization, even if new loader is enabled.

I'm not familiar with our react. Does anyone know if it can cause a problem?

And I came up with another concern, fast room switching. Can't it be a problem if two rooms (moving from and to) set different newLoader flag?

nikk15:
> `shouldUseNewLoader()` seems to can be called from
> 
> https://github.com/mozilla/hubs/blob/5bebb4fefae3500490a2641e91f3881971d0490e/src/react-components/room/hooks/useRoomLoadingState.js#L90
> before `APP.hub` initialization. It returns `false`, if it's called before the `APP.hub` initialization, even if new loader is enabled.
> 
> I'm not familiar with our react. Does anyone know if it can cause a problem? (Maybe @nikk15 knows?)
> 
> And regarding this,
> 
> > Is there a chance that flag (`APP.hub.user_data.hubsUseNewLoader`) can be flipped (after the first initialization)?
> 
> I guess it can happen with fast room switching? Can't it be a problem if two rooms (moving from and to) set different newLoader flag?

I did a little digging through the `useRoomLoadingState` flow. Whilst `shouldUseNewLoader()` returns `false` it will simply stay on the loading screen, which should be fine. See below.

```
  useEffect(() => {
    if (done) {
      // The loaded state on the scene signifies that the loading screen is no longer visible,
      // the initial scene was loaded, and the network connection is established.
      sceneEl.addState("loaded");
    }
  }, [sceneEl, done]);
```

There are useEffects set up in components that use this hook that will fire once `shouldUseNewLoader()` returns `true`, so it shouldn't impact anything.

takahirox:
@nikk15 Thanks! Sounds like it shouldn't be a big deal and we don't really need to care about it, especially because the flag should be just temporal and ideally we will remove the flag and always enable the new loader.

So, the last remained concern for me is fast room switching. As the flag should be temporal, it might be ok to ignore. Or we may check the new loader flag in the rooms (from and to) and force to reload if they are different. We may make another PR for it if needed. /cc @johnshaughnessy

takahirox (approved):
We may need to think of fast room switching stuff, but we may make another PR.

> The only exception is that the new video grabbing behavior is applied even when newLoader is on, since reworking input maps at runtime seems tricky and error prone.

Honestly I don't really get it because I'm not familiar with our input maps. Please let other reviewers review it, if needed.

johnshaughnessy:
> I guess it can happen with fast room switching? Can't it be a problem if two rooms (moving from and to) set different newLoader flag?

@takahirox Yes, you're right that this could be a problem, but I think it's a problem we can safely ignore for now because it will only be enabled intentionally for testing purposes.

If it helps, here is the pathway I expect we'll follow for rolling out the new code:
- Use a query string parameter (`newLoader`) so that individuals can test. (This is where we are now.)
- Add a room-level flag so that we can run tests that guarantee that all participants are running the new code. (e.g. our weekly all-team calls). (This PR enables this.)
- Ship the new code enabled-by-default to a dev instance and test on there until we have all green from QA and internal tests. (We can do this next / simultaneously.)
- Ship the new code enabled-by-default to the demo hub.
- Ship the new code enabled-by-default to managed instances.
- Ship the new code enabled-by-default to hubs cloud stable channels.

takahirox:
@johnshaughnessy Thanks, clearifying the pathway really helps move the shipping forward

PR #6044, opened 2023-04-20
Upgrade to react and react-dom 18.2.0

nikk15:


netpro2k (commented):
Yay, this is long overdue (we have lots of these sorts of "Gee we should really keep that up to date... 4 years later" sort of things to catch up on.

It's been a long while since I thought about npm intricacies so I forget what the implications are of `legacy-peer-deps`. What is forcing us to add that? 

Unclear if it's related to that flag, but looking at the package-lock changes I see some slightly surprising ones (not concerning really, just surprising). For example its pretty surprising to see `@babel/helper-string-parser` change from just bumping react and react-dom in package.json.

One other dev environment setup thing to keep in mind, especially when updating package.json, we use node 16 (specifically 16.16 https://github.com/mozilla/hubs/blob/master/RetPageOriginDockerfile#L5) in the build environment, so its best to also use those locally. Its especially important with package changes that we all use the same node/npm version since if we end up all using different versions, we can make conflicting package-lock changes. This is typically only an issue with major version differences, but it has bitten us in the past. You likely want to use something like nvm to manage that for you https://github.com/mozilla/hubs/blob/master/CONTRIBUTING.md#0-dependencies

You may have already done this, but since this is such a large version bump it worth going through the major release patch nodes/migration guides to make sure we aren't doing any legacy things. Same goes for our react based dependencies (storybook, react-emoji-render, react-infintite-scroll, etc). This is honestly getting out of my area of expertise though as I have been out of the frontent UI game for quite some time heh. Might all just be fine.

nikk15:
Closing PR as I have created a fresh branch to continue this work.

PR #6051, opened 2023-04-25
tidying up security issues for locked down demo

mikemorran:
This PR locks down the "i" key shortcut in locked down demo rooms and allows promoted users to have full access to room features. The desire is for promoted users to be able to moderate demo rooms.

### DIFF
--- a/src/systems/inspect-yourself-system.js
+++ b/src/systems/inspect-yourself-system.js
@@ -1,8 +1,9 @@
+import { isLockedDownDemoRoom } from "../utils/hub-utils";
 import { paths } from "./userinput/paths";
 export class InspectYourselfSystem {
   tick(scene, userinput, cameraSystem) {
     if (!scene.is("entered")) return;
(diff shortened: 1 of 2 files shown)


src/systems/inspect-yourself-system.js
```diff
@@ -1,8 +1,9 @@
+import { isLockedDownDemoRoom } from "../utils/hub-utils";
 import { paths } from "./userinput/paths";
 export class InspectYourselfSystem {
   tick(scene, userinput, cameraSystem) {
     if (!scene.is("entered")) return;
-    if (userinput.get(paths.actions.startInspectingSelf)) {
+    if (userinput.get(paths.actions.startInspectingSelf) && !isLockedDownDemoRoom()) {
```
keianhzo, line 6: Another option would have been to early exit as we do with the `enter` scene condition:
```javascript
if (!scene.is("entered") || isLockedDownDemoRoom()) return;
```

PR #6058, opened 2023-05-01
Add Lilypad Tooltip component to existing tool bar buttons

nikk15:
- adds Lilypad library
- replaces existing tooltips and adds tooltips to toolbar buttons
- condense some larger components and begin swapping to preferred folder structure in the process

https://user-images.githubusercontent.com/42850541/235947638-f21250ec-43b0-4eea-bc1d-c75baa29aa28.mov


src/assets/locales/es.json
```diff
@@ -7,8 +7,7 @@
   "avatar-page.select": "Seleccionar",
```
nikk15, line 7: You can ignore the language files. These are just formatting updates from my attempt to add translations, which have been removed

src/react-components/input/ToolbarButton.js
```diff
@@ -50,6 +51,7 @@ export const ToolbarButton = forwardRef(
         )}
         disabled={disabled}
         title={title}
+        onClick={onClick}
```
nickgrato, line 54: Not in this PR but as the FE matures it might be good to remove these {...rest} catch alls, side note I know they where there, I just think they can be cleaver yet misleading, you are establishing these nice prototypes soon to be TS type and I think it would be nice to have concrete props being sent to the component with no wild card catch all.

src/react-components/room/ChatSidebarContainer.js
```diff
@@ -54,146 +48,6 @@ const chatSidebarMessages = defineMessages({
     defaultMessage: "Text Chat Off"
   }
 });
-
-function shouldCreateNewMessageGroup(messageGroups, newMessage, now) {
-  if (messageGroups.length === 0) {
```
nickgrato, line 59: would you mind just adding a few comments on where all this code went just so I don't have to search too much! thanks!

src/react-components/room/components/ChatToolbarButton/ChatToolbarButton.tsx
```diff
@@ -0,0 +1,39 @@
...
+import React, { RefAttributes, useContext } from "react";
+import { ChatContext } from "../../contexts/ChatContext";
+import { ToolTip } from "@mozilla/lilypad-ui";
+
+const chatTooltipDescription = defineMessage({
+  id: "chat-tooltip.description",
+  defaultMessage: "Open the chat sidebar (T)"
+});
+
+interface ChatToolbarButtonProps extends RefAttributes<any> {
```
nickgrato, line 15: We usually opt in to using types before interfaces.  type ChatToolbarButtonPropsT = {}

This other suggestion is stylistic, we do it in the subscription team but we end interface,type, and enum with the first initial of what it is.  CategoriesE, ButtonsI ,ChatToolbarButtonPropsT etc.

I don't think extends RefAttributes<any> is being used here?

src/react-components/room/contexts/ChatContext.tsx
```diff
@@ -0,0 +1,161 @@
...
+  }, [setUnreadMessages]);
+
+  return (
+    <ChatContext.Provider value={{ messageGroups, unreadMessages, sendMessage, setMessagesRead }}>
+      {children}
+    </ChatContext.Provider>
+  );
+}
+
+ChatContextProvider.propTypes = {
```
nickgrato, line 158: you can delete this now that you have types.

src/react-components/room/contexts/ChatContext.tsx
```diff
@@ -0,0 +1,161 @@
...
+    case "image":
+    case "photo":
+    case "video":
+    case "permission":
+      return processChatMessage(messageGroups, newMessage);
+    default:
+      return messageGroups;
+  }
+}
+
```
nickgrato, line 102: import { ReactNode } from 'react';

type ChatContextProviderPropsT = {
  children: ReactNode;
  messageDispatch: MessageDispatch ( not sure about that but you get the idea.
};

nickgrato (changes requested):
for the most part the pr looks good bu the chat context has quite a few "any"s, ideally noting is an any unless we have no other choice or it's intentional.

nickgrato (approved):
no blocking changes, can clean up a  few things but i think you got it!

PR #6059, opened 2023-05-03
Log scene name when scene is changed

stalgiag:
This PR logs a message with the name of the scene whenever the scene is swapped.

### DIFF
--- a/src/hub.js
+++ b/src/hub.js
@@ -1376,6 +1376,8 @@ document.addEventListener("DOMContentLoaded", async () => {
         updateEnvironmentForHub(hub, entryManager);
       });
 
+      console.log(`Entering new scene: ${hub.scene?.name}`);
+
       messageDispatch.receive({
         type: "scene_changed",
         name: displayName,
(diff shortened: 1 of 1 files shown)


src/hub.js
```diff
@@ -1376,6 +1376,8 @@ document.addEventListener("DOMContentLoaded", async () => {
         updateEnvironmentForHub(hub, entryManager);
       });
 
+      console.log(`Entering new scene: ${hub.scene?.name}`);
+
```
takahirox, line 1380: Not a strong preference, but `hub.scene ? hub.scene.name : "a custom URL"` might be better to match the following lines?

```javascript
const sceneName = hub.scene ? hub.scene.name : "a custom URL";

console.log(`Entering new scene: ${sceneName}`);

messageDispatch.receive({
    type: "scene_changed",
    name: displayName,
    sceneName
});
```

stalgiag: This is what I had at first but wasn't sure if this console.log would be deleted soon so I was reluctant to make a variable and couple with messageDispatch. But it seems like the console.log will stay and I am probably overthinking it. Change made.

takahirox (approved):

Issue #6048, opened 2023-04-24
Smooth Rotation


Clemens100:
I think the Rotation (Q & E on Desktop mode or stick flicking in VR) of the Camera should be a bit smoother and not so “jumpy” like it is right now (45 degrees).
More like 1 or 2 degrees per trigger, and it is possible to hold the button to rotate smooth.

As I already mentioned, I think it would be better to hold the Buttons and the Camera then rotates smoothly in the Direction. Instead of setting an angle Degree (like 30-60 degree) in the Settings, set the speed of the Camera it will rotate when you hold the Button.
I would like to try the change by my self, if you consider it as useful.

**Additional context**

I already worked a bit with the Hubs Code, so I'm not new to the Project.
If you have some further questions or remarks, please contact me or comment here.

I would be very happy to contribute to this Project with this little Quality of Life Change, if you consider this change as useful.

matthewbcool:
Snap turning is an intentional default for people with motion sickness but the option to smooth turn would be a welcome preference IMO. Thanks very much for taking the time to document this request.

Clemens100:
Opened a Pull Request for this: 
https://github.com/mozilla/hubs/pull/6060

Thank you!

PR #6060, opened 2023-05-04
Smooth Rotation Option

Clemens100:
I have created an Option for smooth Rotation, just like mentioned in this GitHub Issue:
https://github.com/mozilla/hubs/issues/6048

If there are some issues with my Version of this, I would be happy if you contact me, so I can find a Solution.

Have a good one,
Clemens100


Clemens100:
Do I have to resolve this merging issue? It's just the Language JSON's.
I'm not into Spanish, Russian and Chinese, so I can only write what Google Translator is outputting.

PR #6065, opened 2023-05-08
Admin panel - remove character limit on theme edit

nikk15:


nickgrato (approved):
lgtm, this seems like more of a product question so might be good to run by john or dom!

PR #6067, opened 2023-05-10
bitECS object list support

keianhzo:
This commit adds bitECS object list support.

**Basic Strategy**

Reuse the existing A-Frame based code as much as possible for now.

In the related functions that take A-Frame element, take `Object3D` instead. `Object3D` has a reference to A-Frame element in the A-Frame based implementation and has a reference to Entity ID in the bitECS. The functions process with A-Frame element or Entity ID depending on whether new loader enabled.

Explicitly use `shouldUseNewLoader()` and/or make two functions, one for A-Frame based and another one for bitECS based implementation, if different logics are needed between A-Frame based and bitECS based implementations. Duplicated codes may not be perfectly removed but it would be simpler to follow the code rather than the new loader pretends the old one. And it would be easier to edit
the code when we will get rid of A-Frame.

**Changes**

- Introduce `MediaInfo` component and save url and content type into it when loading media in media-loader. The info is used in the object list
- Fire `listed_media_changed` event when media is loaded via media-loader and when the loaded media entity is removed
- Take `Object3D` instead of A-Frame element in the related functions
- Use `shouldUseNewLoader()` and/or make two separated functions
  for A-Frame and bitECS where the different logics are needed

**Future TODOs**

- Support avatars

Authors @keianhzo and @takahirox

### DIFF
--- a/src/utils/jsx-entity.ts
+++ b/src/utils/jsx-entity.ts
@@ -38,8 +38,8 @@ import {
   MaterialTag,
   VideoTextureSource,
   Quack,
-  Mirror,
-  MixerAnimatableInitialize
+  MixerAnimatableInitialize,
+  Inspectable
 } from "../bit-components";
 import { inflateMediaLoader } from "../inflators/media-loader";
 import { inflateMediaFrame } from "../inflators/media-frame";
@@ -359,6 +359,7 @@ export interface JSXComponentData extends ComponentData {
   waypointPreview?: boolean;
   pdf?: PDFParams;
   loopAnimation?: LoopAnimationParams;
+  inspectable?: boolean;
 }
 
 export interface GLTFComponentData extends ComponentData {
@@ -464,6 +465,7 @@ const jsxInflators: Required<{ [K in keyof JSXComponentData]: InflatorFn }> = {
   quack: createDefaultInflator(Quack),
   mixerAnimatable: createDefaultInflator(MixerAnimatableInitialize),
   loopAnimation: inflateLoopAnimationInitialize,
+  inspectable: createDefaultInflator(Inspectable),
   // inflators that create Object3Ds
   object3D: addObject3DComponent,
   slice9: inflateSlice9,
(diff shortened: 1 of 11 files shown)


src/utils/jsx-entity.ts
```diff
@@ -38,7 +38,7 @@ import {
   Billboard,
   MaterialTag,
   VideoTextureSource,
-  Mirror
+  Inspectable
```
takahirox, line 41: Is removing Mirror intentional?

keianhzo: Yes, It wasn't being used.

keianhzo:
Related https://github.com/mozilla/hubs/pull/6092

takahirox:
Sorry for the late review. Would you mind resolving the conflicts?

takahirox:
Resolved the conflicts so far.

I read through the changes and found that `elOrEid` that holds Entity ID for bitECS or Element for A-Frame is used a lot but I don't think it is good design because

* It may be a bit less readable
* Variable that can store different types doesn't really fit to TypeScript (when we rewrite the JavaScript code to TypeScript at some point)
* Dynamically checking type is needed and statical type check can't work

Perhaps explicitly separating the codes between bitECS and A-Frame with `shouldUseNewLoader()` would be simpler? And it would be easier to edit when we will get rid of A-Frame code. I will try to update the code.

takahirox:
Ready for review.

I have fixed some bugs, have cleaned up the code, have written the commit log and PR comment.

keianhzo:
@takahirox Thanks for updating this while I was away. I've added the missing support for pinning. I think that's the last missing part for this PR.

takahirox:
This PR affects not only the new loader but also the old one so thinking we should test more carefully than other new loader specific PRs before merging.

takahirox:
Would you mind separating the change for pinned objects stuff from this PR for quicker review? The review of this PR would be easier because we both already looked through the changes without the pinned objects stuff.

keianhzo:
You mean moving the last two commits to another PR? I can do that if that helps reviewing.

takahirox:
Yes. If you do it, I will approve this PR immediately and can review another one later.

keianhzo:
@takahirox I've removed the last two commits

PR #6072, opened 2023-05-11
Add quack and duck to bitecs

stalgiag:
This migrates [duck](https://github.com/mozilla/hubs/blob/master/src/components/duck.js) and [quack](https://github.com/mozilla/hubs/blob/master/src/components/quack.js) to bitecs and recreates what I believe was the original intended behavior:
- Duck can be generated with '/duck'
- Duck is networked, grabbable, and responds to physics
- Duck play quack sound effect locally on grab, occasionally the quack is "special"

I tried to keep everything as near to the original as possible since this is an early change for me. I added comments to the diffs for situations where this was awkward.

part of #5899


src/bit-components.js
```diff
@@ -319,3 +319,7 @@ export const LinearScale = defineComponent({
   targetY: Types.f32,
   targetZ: Types.f32
 });
+export const Quack = defineComponent({
+  quackPercentage: Types.f32,
+  specialQuackPercentage: Types.f32
+});
```
stalgiag, line 325: Unnecessary to expose these as properties when I don't think we will be making entities with quack components that have different randomization ratios.

src/bit-systems/quack.ts
```diff
@@ -0,0 +1,20 @@
...
+
+const heldQuackQuery = defineQuery([Quack, Held]);
+const heldQuackEnterQuery = enterQuery(heldQuackQuery);
+
+const quack = (eid: number) => {
+  const rand = Math.random();
+  if (rand < Quack.specialQuackPercentage[eid]) {
+    AFRAME.scenes[0].systems["hubs-systems"].soundEffectsSystem.playSoundOneShot(SOUND_SPECIAL_QUACK);
+  } else if (rand < Quack.quackPercentage[eid]) {
+    AFRAME.scenes[0].systems["hubs-systems"].soundEffectsSystem.playSoundOneShot(SOUND_QUACK);
```
stalgiag, line 14: This was the approach to playing a OneShot used in the camera tool bit system. It feels like this shouldn't be the approach but I didn't see another options without making a bit sound effects system. I might've missed something.

src/bit-systems/quack.ts
```diff
@@ -0,0 +1,20 @@
+import { HubsWorld } from "../app";
+import { Held, Quack } from "../bit-components";
+import { defineQuery, enterQuery } from "bitecs";
+import { SOUND_QUACK, SOUND_SPECIAL_QUACK } from "../systems/sound-effects-system";
+
+const heldQuackQuery = defineQuery([Quack, Held]);
+const heldQuackEnterQuery = enterQuery(heldQuackQuery);
+
+const quack = (eid: number) => {
```
stalgiag, line 9: Not ideal to have a whole system for managing this component but necessary for now as far as I can tell.

src/bit-systems/quack.ts
keianhzo: That's the right way to do it ATM. It would be good though to reduce the number of AFRAME references as we migrate to bitECS. We hold a AFRAME scene ref in `APP.scene` so you can also get the system through that.

src/bit-systems/quack.ts
keianhzo: Yeah I don't think this fits well anywhere else so I think it's ok for now. I'd consider inlining this function in the main system loop to avoid a function call.

src/bit-components.js
keianhzo: Yes, I don't think we need this or the inflator. This is mostly a testing component that we don't really customize so we can get rid of the properties and the inflator and hardcode the defaults in the system I think.

src/message-dispatch.js
```diff
@@ -149,6 +150,7 @@ export default class MessageDispatch extends EventTarget {
             animateLoad: true,
             isObjectMenuTarget: true
           });
+          inflateQuack(APP.world, eid);
```
keianhzo, line 153: This looks good to me. Another option if you want to implement it as an exercise would be to create a specific prefab for a duck that includes the quack component and here we would be just creating a `Duck` networked entity.

keianhzo (approved):
LGTM, I just left a few of comments. Feel free to merge or address the feedback.

src/message-dispatch.js
stalgiag: Good idea! It feels odd to not address `Duck` at the same time and we might as well expand the learning with this exercise.

src/prefabs/duck.tsx
```diff
@@ -0,0 +1,49 @@
...
+          COLLISION_LAYERS.HANDS |
+          COLLISION_LAYERS.ENVIRONMENT |
+          COLLISION_LAYERS.INTERACTABLES |
+          COLLISION_LAYERS.AVATAR
+      }}
+      physicsShape={{
+        fit: Fit.MANUAL,
+        type: Shape.CYLINDER,
+        halfExtents: [0.25, 0.5, 0.45]
+      }}
```
stalgiag, line 45: The default `Shape.HULL` created issues. These issues appear unrelated to the changes in this PR. 'Shape.CYLINDER' offers mostly believable behavior and is cheaper.

src/message-dispatch.js
```diff
@@ -142,13 +142,7 @@ export default class MessageDispatch extends EventTarget {
...
-            src: getAbsoluteHref(location.href, ducky),
-            resize: true,
-            recenter: true,
-            animateLoad: true,
-            isObjectMenuTarget: true
-          });
+          const eid = createNetworkedEntity(APP.world, "duck");
           const obj = APP.world.eid2obj.get(eid);
           obj.position.copy(avatarPov.localToWorld(new THREE.Vector3(0, 0, -1.5)));
           obj.lookAt(avatarPov.getWorldPosition(new THREE.Vector3()));
```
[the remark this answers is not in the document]
keianhzo: Yeah, in most cases we use this for spawning media in front of the user: 
https://github.com/mozilla/hubs/blob/a0e5eee06395c5cfeb604e03dab03e9a33077620/src/scene-entry-manager.js#L244
If we add a component for this we can also refactor those methods.

src/prefabs/duck.tsx
stalgiag: [Issue filed here.](https://github.com/mozilla/hubs/issues/6078)

Exairnous:
Does this preserve the reverse physics (floating) when the duck is scaled?

stalgiag:
> Does this preserve the reverse physics (floating) when the duck is scaled?

Good catch! I wasn't aware of this easter egg feature and thought the physics were similar to thee default `floatyObject`. It looks like this behavior is a unique case exclusively for ducks. We don't currently have a duck system and one shouldn't be made for modifying physics.

I could make this behavior happen under an additional flag on `floatyObject` which would allow for potential reuse.

stalgiag:
Okay I modified the `FloatyObject` component to handle Duck-specific physics. I made it a flag within `FloatyObject` in case we want to reuse this behavior elsewhere. The `FloatyObject` component could quickly become unweldy if there are many different unique physics cases like this one. If this is the situation then I can decouple from `floatyObject` and make a new new `helium` component & system.

Exairnous:
Thanks!  Yeah, it's a bit hidden, but really fun and useful.  There is a way with the old loader to have arbitrary GLBs behave as ducks, so we've been able to hack in flying/floating objects that way, e.g. [Floating_Lantern.zip](https://github.com/mozilla/hubs/files/11494890/Floating_Lantern.zip).

So hopefully either that'll still work, or your suggestion of a `helium` component sounds good :)

keianhzo (approved):
I had no idea about this behavior 😅 LGTM!

PR #6077, opened 2023-05-15
fix: dark mode logo now shows when specified

nikk15:
- Dark mode logo was not showing for some accounts since the HMC logo fix. Discovered that users had to manually add "darkModeDefault: true" to their theme in the admin panel. This is not obvious, so I added a check for the inclusion of "dark-mode" in the theme id.
- Minor fix for the import of svgs in TS included in this PR
- This PR also changes the local env TIER variable to paid tier "p1".

NOTE: I'm not sure I agree with the darkModeDefault property - when we tidy up custom theming, I'll look into alternatives.

### DIFF
--- a/admin/webpack.config.js
+++ b/admin/webpack.config.js
@@ -80,7 +80,7 @@ module.exports = (env, argv) => {
       RETICULUM_SERVER: "hubs.local:4000",
       POSTGREST_SERVER: "",
       ITA_SERVER: "turkey",
-      TIER: "p0"
+      TIER: "p1"
     });
   }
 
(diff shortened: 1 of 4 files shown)


admin/webpack.config.js
```diff
@@ -80,7 +80,7 @@ module.exports = (env, argv) => {
       RETICULUM_SERVER: "hubs.local:4000",
       POSTGREST_SERVER: "",
       ITA_SERVER: "turkey",
-      TIER: "p0"
+      TIER: "p1"
```
nickgrato, line 83: this is my own curiosity why do we need this one?

nikk15: Especially with alot of incoming work around theme customization, we use the admin panel theme tab a lot more now so it made sense to me to update this rather than manually do it each time.

nickgrato: ah cool!

nickgrato (commented):
lgtm

PR #6079, opened 2023-05-16
Fixes bug with Hmc logo display

nikk15:
- adds to the previous PR fix allowing hmc logos to be replaced with custom logos [](https://github.com/mozilla/hubs/pull/6068/files)


nickgrato (approved):
lgtm

PR #6080, opened 2023-05-16
Decouple branding sections from external details

bryanenders:
Why
---
Admin is an instance service.  We don’t want it coupled to orchestration details or the Hubs managed service plans.

What
----
* Hide branding sections based on `DISABLE_BRANDING` rather than `ITA_SERVER` and `TIER`


nickgrato (approved):
looks good to me - I'm a little behind on what's going on here but the syntax is correct and I get your general direction with the project.

nikk15 (approved):
Looks good to me! Config boolean values are always strings, right? If so, approved :)

PR #6082, opened 2023-05-16
Use canonical URL for local non-model assets

### LINKED ISSUE #6081: Media placed in room shows broken media link when working locally (newLoader)
**Description**
If you add a media object to the scene that is not a glb/gltf while working in your local development environment, it will appear as a broken media link. The issue [occurs in `resolveMediaInfo()`](https://github.com/mozilla/hubs/blob/109f76058ba1c2596d291a6bc6e1123ea037ad08/src/utils/media-utils.js#L606)

**To Reproduce**
1. Working locally, with `newLoader`, join a room
2. Drag an image or other non-model media file onto screen
3. See that the media object appears as a Broken Media Link

**Expected behavior**
The media item is instantiated in the scene.

**Additional context**
I am running the docker hubs compose dev environment.

stalgiag:
Not ready for merge.

Links for local files are broken when working locally behind `newLoader` flag.

This change causes the canconical URL to be retrieved for non-model local assets. This is tested locally and on my personal dev instance with *.jpg, *.glb, *.mp4, and *.mp3. It is possible that there is something that I am missing and that this is a regression so this should not be merged until I can confirm what the original intent was for limiting this behavior to model assets.

resolves #6081

### DIFF
--- a/src/utils/media-utils.js
+++ b/src/utils/media-utils.js
@@ -615,8 +615,7 @@ export async function resolveMediaInfo(urlString) {
 
   // We want to resolve and proxy some hubs urls, like rooms and scene links,
   // but want to avoid proxying assets in order for this to work in dev environments
-  const isLocalModelAsset =
-    isNonCorsProxyDomain(url.hostname) && (guessContentType(url.href) || "").startsWith("model/gltf");
+  const isLocalModelAsset = isNonCorsProxyDomain(url.hostname);
 
   if (url.protocol != "data:" && url.protocol != "hubs:" && !isLocalModelAsset) {
     const response = await resolveUrl(url.href);


takahirox:
Thanks for working on it.

> Not ready for merge.

Please use [draft PR](
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-stage-of-a-pull-request) if a PR is not ready.

stalgiag:
> Thanks for working on it.
> 
> > Not ready for merge.
> 
> Please use [draft PR](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-stage-of-a-pull-request) if a PR is not ready.

Yes, I have been trying to but Github is throwing 500 Errors on their servers with the "convert to draft" route. It is a known outage. I will just close for now and reopen when GH is stable again and can convert to Draft.

src/utils/media-utils.js
```diff
@@ -615,10 +615,9 @@ export async function resolveMediaInfo(urlString) {
 
   // We want to resolve and proxy some hubs urls, like rooms and scene links,
   // but want to avoid proxying assets in order for this to work in dev environments
-  const isLocalModelAsset =
-    isNonCorsProxyDomain(url.hostname) && (guessContentType(url.href) || "").startsWith("model/gltf");
```
stalgiag, line 619: @johnshaughnessy do you have any memory of why models were the only local assets that don't use a canonical URL? With the `newLoader` path - uploaded images, movies, and sounds are broken links when working in a local dev env.

I know [this line also appears in "components/media-loader.js"](https://github.com/mozilla/hubs/blob/7ac8f19db85dd4a13fb68fe2cd911d24bddb26ab/src/components/media-loader.js#LL367C1-L367C1) which dates back [to this PR.](https://github.com/mozilla/hubs/pull/3904)

 If I make the change shown here then all media tested (glb, mp3, mp4, jpg) work with `newLoader` in local dev and on my personal instance. That said,  I am unsure of the original intent in having non-model assets use canonical urls and don't want to introduce a regression.

stalgiag: Asking you specifically because you were the last to work on this section. It was 9 months ago though so no worries if you don't remember.

johnshaughnessy: Hmm, unfortunately no. I don't know why we made an exception only for models. But, given that we're OK with the exception for models, I can't think of why we wouldn't be OK to lift the restriction for other media types like videos and images. 

At the risk of some regression, I'll go ahead with this PR (despite not understanding what lies on the other side of the [fence](https://en.wikipedia.org/wiki/G._K._Chesterton#Chesterton's_fence)).

PR #6091, opened 2023-05-23
Don't play audio after the loading screen has been hidden

keianhzo:
Don't play audio until the loading screen has disappeared.

### DIFF
--- a/src/hub.js
+++ b/src/hub.js
@@ -574,10 +574,16 @@ function handleHubChannelJoined(entryManager, hubChannel, messageDispatch, data)
 
   console.log(`Dialog host: ${hub.host}:${hub.port}`);
 
+  const audioSystem = scene.systems["hubs-systems"].audioSystem;
+  audioSystem.setMasterVolume(0);
+
   remountUI({
     messageDispatch: messageDispatch,
     onSendMessage: messageDispatch.dispatch,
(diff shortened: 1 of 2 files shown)


Exairnous:
Does this include voice audio or just media?  Scenes often take a long time to load for me and I really appreciate being able to hear conversations while loading.

keianhzo:
Good call @Exairnous I wasn't sure about that but I think it makes sense specially for slow connections.

Exairnous:
Thanks!  I know this has just been merged, but now that I think about it, I'm not too sure what issue this PR solves, is it so that loud media elements that autoplay aren't annoying/distracting during room load?  This seems like it's just auto-setting the audio preference and that got me to thinking that it would be really nice to have the bottom bar (as configured for the lobby with the invite, chat, and "more" buttons) during the scene load so you could just set preferences like that.  This would have the added bonus (I think?) of allowing you to send chat messages during room load (which would be really great during an event if you were unable to load the room for some reason so you could tell the event host, similar to how you could with the Discord bot) and change room settings if you're a moderator.  Having access to the people menu might be beneficial for moderation too, although that button overlays the scene so it might be harder to expose, hmm 🤔

Anyway, not trying to imply I'm against it, these are just a couple things this PR got me thinking about.

PR #6092, opened 2023-05-23
Pinning/Unpinning interaction and media promotion checks for bitECS + entity state API

### LINKED ISSUE #6086: Pinned objects can still be grabbed (newLoader)
**Description**
Pinned objects can be grabbed and moved when working behind `newLoader`. A cursory search suggests that the `Pinned` component is not always successfully added to the entity (either on first pinning or on load from reticulum) [causing checks like this one](https://github.com/mozilla/hubs/blob/20ba18673bbb45fa0b422d7c2c664f0f9a3387d9/src/systems/hold-system.js#L38) to behave unpredictably.

**To Reproduce**
1. Go to a room with `newLoader` URL param
2. Add an object to the room
3. Pin the object
4. Try to move the pinned object

pinned object

stalgiag:
This PR ...
- Adds utility functions for determining pinned/pinnable state and permissions
- Removes Pinned and Pinnable components
- Passes necessary keys to Reticulum for media promotion
- Sends file_id to the backend when unpinning so that media can be demoted from permanent storage

~~To do this, I needed to add properties to the `MediaLoaded` component that store `src` and `fileId` values. This was necessary because the `MediaLoader` component is removed after load and those values are needed for media promotion. It is not ideal to duplicate the values in order to persist them. I mention possible alternatives in an inline comment.~~

~~Cases not handled by this PR ...~~
~~- Attempts to pin an object without being signed in silently reject. This is the existing behavior. This was not handled because (as far as I can tell) the bitECS code path doesn't have a solution for using `performConditionalSignIn()`. In the previous code path this function was passed as a construction parameter to a class instance of `PinningHelper` which was stored on `APP`. This does not seem appropriate for the new code path and likely exceeds the scope of this PR.~~

~~[This PR is dependent on these changes to Reticulum ](https://github.com/mozilla/reticulum/pull/700)so I will mark this PR as a draft until those changes are merged.~~

 resolves #6086


src/bit-components.js
```diff
@@ -153,10 +153,17 @@ export const CameraTool = defineComponent({
...
 });
 MediaLoader.src[$isStringType] = true;
-export const MediaLoaded = defineComponent();
+MediaLoader.fileId[$isStringType] = true;
+export const MediaLoaded = defineComponent({
+  fileId: Types.ui8,
+  src: Types.ui32
+});
+MediaLoaded.fileId[$isStringType] = true;
+MediaLoaded.src[$isStringType] = true;
```
stalgiag, line 166: Duplicating these values into this component is workable but non-ideal. Other options:

- Remove `MediaLoaded` component entirely. Add `LOADED` flag to `MediaLoader` and update that flag rather than removing the component on successful load.
- A separate `FileInfo` component which may serve more general purposes in the future. Could be prepared and added by `inflateMediaLoader()`

src/utils/listen-for-network-messages.ts
```diff
@@ -91,7 +94,16 @@ export function queueEntityStateAsMessage(entityState: EntityState) {
...
+  const rootNid = APP.getSid(response.data[0]!.create_message.networkId);
+  const eid = APP.world.nid2eid.get(rootNid);
+  if (eid) {
+    if (hasComponent(APP.world, Pinnable, eid)) {
+      addComponent(APP.world, Pinned, eid);
+    }
+    if (hasComponent(APP.world, FileInfo, eid)) {
+      FileInfo.flags[eid] |= FILE_INFO_FLAGS.IS_PERMANENT;
+    }
+  }
```
johnshaughnessy, line 106: If we remove the need to track this derived state separately (`Pinnable`, `Pinned`, and `FileInfo`'s `IS_PERMANENT` flag) as described in a separate comment, then this code can be removed.

If that were not the case, I would encourage us to move the processing of any message data to the `networkReceiveSystem`. Ideally, the only thing happening at this point in the program would be queuing data received from the phoenix channels. We do some parsing and minor revisions to the messages, but try to let the `networkReceiveSystem` be responsible for things that affect the world state.

johnshaughnessy (changes requested):
I think there some areas where can depend on derived state and avoid code that could hide bugs:

"Pinned" is derived state and doesn't need its own component: An entity is pinned iff its `Networked.creator` is reticulum.

"isPinnable" is only important in the context of whether we `canPin` an entity. We `canPin` an entity if:

```ts
function canPin(
  hubChannel: HubChannel,
  store: Store,
  world: HubsWorld,
  eid: EntityID
): boolean {
  const hasFile = hasComponent(world, FileInfo, eid);
  const hasPromotableFile =
    hasFile &&
    store.state.uploadPromotionTokens.some(
      (upload: { fileId: string }) =>
        upload.fileId === APP.getString(FileInfo.id[eid])
    );

  return (
    isNetworkInstantiated(eid) && // It's a root networked entity
    !isPinned(eid) && // It's not already pinned.
    hubChannel.can("pin_objects") && // Our account has permission
    (!hasFile || hasPromotableFile)
  );
}
```

Note that the permission check will implicitly ensure user is signed in.

When we `deleteEntityState`, for an entity that has a `FileInfo`, we need to send reticulum its associated `file_id` so that it can call `OwnedFile.set_inactive(file_id, account.account_id)` (see `HubChannel.unpin/3`). `OwnedFile`s that are `inactive` will be demoted to expiring storage (and then deleted if they expire) in a cron job.

Since clients need a `file_id` in order to correctly `deleteEntityState`, `file_id`s should be included in an entity's `InitialData` (sent/stored with `create_message`s). With the `file_id` in then entity's `InitialData`, and the `src` also in its `InitialData`, I wonder if the `FileInfo` component also disappears completely and is replaced by a check in an entity's `CreateMessageData`. For example, from `deleteEntityState` we might add:

```ts
const { prefabName, initialData } = createMessageDatas.get(eid)!;
if (prefabName == "media" && initialData.fileId) {
  payload.file_id = initialData.fileId;
}
```

Since the files will be downgraded to expiring storage when their associated entity is unpinned, an uploaded file `IS_PERMANENT` iff its associated entity `isPinned(eid)` (and we do not need to store the derived `IS_PERMANENT` state separately in the client.)

src/bit-systems/media-loading.ts
```diff
@@ -79,8 +80,9 @@ const loaderForMediaType = {
     { accessibleUrl, contentType }: { accessibleUrl: string; contentType: string }
   ) => loadModel(world, accessibleUrl, contentType, true),
   [MediaType.PDF]: (world: HubsWorld, { accessibleUrl }: { accessibleUrl: string }) => loadPDF(world, accessibleUrl),
-  [MediaType.AUDIO]: (world: HubsWorld, { accessibleUrl }: { accessibleUrl: string }) => loadAudio(world, accessibleUrl),
-  [MediaType.HTML]: (world: HubsWorld, { canonicalUrl, thumbnail }: { canonicalUrl: string, thumbnail: string }) =>
+  [MediaType.AUDIO]: (world: HubsWorld, { accessibleUrl }: { accessibleUrl: string }) =>
+    loadAudio(world, accessibleUrl),
+  [MediaType.HTML]: (world: HubsWorld, { canonicalUrl, thumbnail }: { canonicalUrl: string; thumbnail: string }) =>
```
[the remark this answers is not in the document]
stalgiag: Oh wild! You have quite the eye for detail. Not sure how that happened. I didn't realize how flexible TS is with delimiters. Not important now but in the future we might want to add `typescript-eslint` and [enforce the comma style](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/member-delimiter-style.md).

stalgiag:
I really love this point about derived state and I agree it will be less susceptible to state-related bugs. My original idea took the existence of the Pinned component and [checks like this one](https://github.com/mozilla/hubs/blob/20ba18673bbb45fa0b422d7c2c664f0f9a3387d9/src/systems/hold-system.js#L38) and ran with them. I like the idea of instead working from the `isPinned(eid)` as a starting point.

It feels like a larger discussion, but at some point it would be nice to have everyone on the same page about how much of an entity's state needs to be reflected in its components.

I'll take a crack at this approach today.

stalgiag:
This is ready for review again. It feels a lot cleaner with this approach. Thanks again for the suggestion. PR has a little more added to it in unexpected places now because I removed all use of Pinned and Pinnable components to prevent confusion about source of truth for this state.

I folded the unpin legacy components and deactivate files work into this PR. There is one point that I feel mildly uneasy about. I added an inline comment.

src/utils/entity-state-utils.ts
```diff
@@ -64,6 +67,12 @@ export async function deleteEntityState(hubChannel: HubChannel, world: HubsWorld
   const payload: DeleteEntityStatePayload = {
     nid: APP.getString(Networked.id[eid])! as NetworkID
   };
+  const {
+    initialData: { fileId }
+  } = createMessageDatas.get(eid)!;
+  if (fileId) {
+    payload.file_id = fileId;
+  }
```
stalgiag, line 75: The backend "delete_entity_state" behavior is similar to the previous "unpin" behavior in that if it receives a `file_id` it will run `OwnedFile.set_inactive`. If it doesn't receive a `file_id` it just goes straight to performing the unpinning and broadcasting "entity_state_deleted". If "delete_entity_state" is ever called for an OwnedFile without a file_id then it would be inaccessible and stored forever. This feels tenuous as it puts a lot of faith in createMessageDatas on the client. This is related to the other concern you mentioned but probably outside the scope of this PR.

PR #6093, opened 2023-05-23
fixes error in chat commmad permission checking

pattersonbl2:
Removes `this.hubChannel.signIn` due to improper function use. 

The function will always return true in the if statement, which doesn't validate user sign-in permission.


keianhzo (approved):
LGTM

PR #6094, opened 2023-05-24
Ensure entry room logo is the same ratio as loading screen logo

nikk15:
The logo image size on the entry modal was inconsistent with the loading screen logo. This PR ensures they are equal in size.

### DIFF
--- a/src/react-components/room/RoomEntryModal.scss
+++ b/src/react-components/room/RoomEntryModal.scss
@@ -20,8 +20,8 @@
 
 :local(.logo) {
   height: auto;
-  max-height: 250px;
-  height: 100%;
+  max-width: 260px;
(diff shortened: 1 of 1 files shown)


nickgrato:
did you test this with long and wide images? I think if you have max- height and max-width it will warp the image ?

it's a small PR but add a description :)

nikk15:
> did you test this with long and wide images? I think if you have max- height and max-width it will warp the image ?

I did! The object-fit property set to contain ensures that the image is not warped. Within those height and width bounds, the image will be contained fully with its natural aspect ratio.

nickgrato (commented):
lgtm

nickgrato (approved):

PR #6098, opened 2023-05-25
Webaudio-migration

keianhzo:
This PR migrates the client code from ThreeJS Audio to use the WebAudio API directly and introduces some optiomizations to avoid audio transform updates when the audio emitter is muted, clipped, paused or it's matrix is not modified.

`matrixIsModified` is currently not really used as it'll be true if the object matrix has been updated at least once. I'd like to try to avoid some more audio matrix updates for objects that are not actually moving, maybe using the physics body `activationState` but that's going to require updates on `three-ammo` so I'll leave it for a follow-aup PR.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -1,6 +1,6 @@
-import { addEntity, createWorld, IWorld } from "bitecs";
+import { addComponent, addEntity, createWorld, IWorld } from "bitecs";
 import "./aframe-to-bit-components";
-import { AEntity, Networked, Object3DTag, Owned } from "./bit-components";
+import { AEntity, AudioListenerTag, Networked, Object3DTag, Owned } from "./bit-components";
 import MediaSearchStore from "./storage/media-search-store";
 import Store from "./storage/store";
 import qsTruthy from "./utils/qs_truthy";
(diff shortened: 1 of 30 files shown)


keianhzo:
@netpro2k I've tried to cover as many audio update cases as I could think of but I might be missing some.

johnshaughnessy:
In `updateAudioSettings`, we check whether we are switching from panning to stereo modes, and fire an event accordingly:

```ts
    if (
      (audio.panner === undefined && settings.audioType === AudioType.PannerNode) ||
      (audio.panner !== undefined && settings.audioType === AudioType.Stereo)
    ) {
      el.emit("audio_type_changed");
    }
```

In the event handlers, we delete the three js Audio (or PositionalAudio) node and replace it with its counter part. (And we do the same thing in `swapAudioType` of `audio-emitter-system.ts`.

When we control the web audio node graph directly, changing from Stereo to Panner does not require deleting resources -- just rewiring of the graph to go through (or not go through) a PannerNode. 

This also allows us to delete `swapObject3DComponent`, which was a hack written specifically for this situation.

I thought this was one of the main motivations for making this change, so I am surprised to not see it in the diff. Are there plans to follow up with that change, or did I misunderstand the problem you're intending to solve?

keianhzo:
Good catch, thanks for that. If updated to correctly handle the source type updates globally.

The main goal of this PR is:
- Simplify the scene and audio node graphs removing unnessary intermediate nodes
- Improve the overall audio performance Minimizing the panner audios and audio listener transforms updates
- Remove some intermediate data stores in bitECS that where there just because THREE Audio was introducing an extra object.

The audio swapping is a good side win but is not a very common case but it's a good side win. Another win is that we can get rid if some

PR #6107, opened 2023-05-31
Make link entity grabbable

keianhzo:
Make the link entity grabbable and unhide the link menu.


stalgiag (approved):
Tested locally and works well! Only note is that we might want [to remove the non-working attempt to enable interactivity](https://github.com/mozilla/hubs/blob/master/src/inflators/link.ts#L11-L12) in the link inflator since interaction is set up here now. Might even be nice to have this be a prefab so this combination of components is consistent but that isn't a necessity.

keianhzo:
Good catch! I've removed those.

PR #6110, opened 2023-06-01
Add support for link types in the link hover menu

keianhzo:
This PR adds support for missing link types in the link hover menu.

### DIFF
--- a/src/bit-systems/link-hover-menu.ts
+++ b/src/bit-systems/link-hover-menu.ts
@@ -1,7 +1,11 @@
 import { defineQuery, enterQuery, entityExists } from "bitecs";
 import type { HubsWorld } from "../app";
-import { Interacted, Link, LinkHoverMenu, LinkHoverMenuItem, HoveredRemoteRight } from "../bit-components";
-import { anyEntityWith } from "../utils/bit-utils";
+import { Interacted, Link, LinkHoverMenu, LinkHoverMenuItem, HoveredRemoteRight, TextTag } from "../bit-components";
+import { anyEntityWith, findChildWithComponent } from "../utils/bit-utils";
+import { isHubsRoomUrl, isLocalHubsAvatarUrl, isLocalHubsSceneUrl, isLocalHubsUrl } from "../utils/media-url-utils";
+import { Text as TroikaText } from "troika-three-text";
+import { handleExitTo2DInterstitial } from "../utils/vr-interstitial";
+import { changeHub } from "../change-hub";
 
 const NULL_EID = 0;
 
@@ -17,10 +21,31 @@ export function linkHoverMenuSystem(world: HubsWorld) {
   const menuObject = world.eid2obj.get(menuEid)!;
 
   // Save Link object eid in LinkHoverMenu when hovered.
-  hoveredEnterQuery(world).forEach((eid: number) => {
+  hoveredEnterQuery(world).forEach(async (eid: number) => {
     LinkHoverMenu.targetObjectRef[menuEid] = eid;
     const targetObject = world.eid2obj.get(eid)!;
     targetObject.add(menuObject);
+
+    const buttonRef = LinkHoverMenu.linkButtonRef[menuEid];
+    let text = findChildWithComponent(world, TextTag, buttonRef)!;
+    let textObj = world.eid2obj.get(text)! as TroikaText;
+    const mayChangeScene = (APP.scene?.systems as any).permissions.canOrWillIfCreator("update_hub");
+    const src = APP.getString(Link.url[eid])!;
+    let hubId;
+    let label = "open link";
+    if (await isLocalHubsAvatarUrl(src)) {
+      label = "use avatar";
+    } else if ((await isLocalHubsSceneUrl(src)) && mayChangeScene) {
+      label = "use scene";
+    } else if ((hubId = await isHubsRoomUrl(src))) {
+      const url = new URL(src);
+      if (url.hash && APP.hub!.hub_id === hubId) {
+        label = "go to";
+      } else {
+        label = "visit room";
+      }
+    }
+    textObj.text = label;
   });
 
   // Check if the cursor it hovered on Link object or Link menu button object.
@@ -30,11 +55,37 @@ export function linkHoverMenuSystem(world: HubsWorld) {
   if (hovered && entityExists(world, linkEid)) {
     // Hovered then make the menu visible and handle clicks if needed.
     menuObject.visible = true;
-    if (clickedMenuItemQuery(world).length > 0) {
-      // TODO: Change the action and label depending on 
-      //       target (eg: switching avatars for avatar url)
-      window.open(APP.getString(Link.url[linkEid])!);
-    }
+    clickedMenuItemQuery(world).forEach(async eid => {
+      const linkEid = LinkHoverMenu.targetObjectRef[menuEid];
+      const src = APP.getString(Link.url[linkEid])!;
+      const exitImmersive = async () => await handleExitTo2DInterstitial(false, () => {}, true);
+
+      const mayChangeScene = (APP.scene?.systems as any).permissions.canOrWillIfCreator("update_hub");
+      let hubId;
+      if (await isLocalHubsAvatarUrl(src)) {
+        const avatarId = new URL(src).pathname.split("/").pop();
+        window.APP.store.update({ profile: { avatarId } });
+        APP.scene!.emit("avatar_updated");
+      } else if ((await isLocalHubsSceneUrl(src)) && mayChangeScene) {
+        APP.scene!.emit("scene_media_selected", src);
+      } else if ((hubId = await isHubsRoomUrl(src))) {
+        const url = new URL(src);
+        if (url.hash && APP.hub!.hub_id === hubId) {
+          // move to waypoint w/o writing to history
+          window.history.replaceState(null, "", window.location.href.split("#")[0] + url.hash);
+        } else if (await isLocalHubsUrl(src)) {
+          const waypoint = url.hash && url.hash.substring(1);
+          // move to new room without page load or entry flow
+          changeHub(hubId, true, waypoint);
+        } else {
+          await exitImmersive();
+          location.href = src;
+        }
+      } else {
+        await exitImmersive();
+        window.open(src);
+      }
+    });
   } else {
     // Not hovered or target object has already been deleted
     // then make the menu invisible and forget the Link object.
--- a/src/bit-systems/link-hover-menu.ts
+++ b/src/bit-systems/link-hover-menu.ts
@@ -26,7 +26,7 @@ function updateLinkMenuTarget(world: HubsWorld, menu: EntityID, sceneIsFrozen: b
   }
 
   const hovered = hoveredQuery(world);
-  const target = hovered.find(eid => findAncestorWithComponent(world, Link, eid));
+  const target = hovered.find(eid => hasComponent(world, Link, eid));
   if (target) {
     LinkHoverMenu.targetObjectRef[menu] = target;
     LinkHoverMenu.clearTargetTimer[menu] = world.time.elapsed + 1000;
--- a/src/bit-systems/link-hover-menu.ts
+++ b/src/bit-systems/link-hover-menu.ts
@@ -23,7 +23,6 @@ function updateLinkMenuTarget(world: HubsWorld, menu: EntityID, sceneIsFrozen: b
 
   if (sceneIsFrozen) {
     LinkHoverMenu.targetObjectRef[menu] = 0;
-    return;
   }
 
   const hovered = hoveredQuery(world);
--- a/src/bit-systems/link-hover-menu.ts
+++ b/src/bit-systems/link-hover-menu.ts
@@ -96,8 +96,11 @@ function updateButtonText(world: HubsWorld, menu: EntityID, button: EntityID) {
   const textObj = world.eid2obj.get(text)! as TroikaText;
   const linkEid = LinkHoverMenu.targetObjectRef[menu];
   const linkType = Link.type[linkEid];
-  let label = "open link";
+  let label = "";
   switch (linkType) {
+    case LinkType.LINK:
+      label = "open link";
+      break;
     case LinkType.AVATAR:
       label = "use avatar";
       break;
--- a/src/bit-systems/link-hover-menu.ts
+++ b/src/bit-systems/link-hover-menu.ts
@@ -1,4 +1,4 @@
-import { defineQuery, entityExists, hasComponent } from "bitecs";
+import { defineQuery, entityExists } from "bitecs";
 import type { HubsWorld } from "../app";
 import { Link, LinkHoverMenu, HoveredRemoteRight, TextTag, Interacted, LinkHoverMenuItem } from "../bit-components";
 import { findAncestorWithComponent, findChildWithComponent } from "../utils/bit-utils";
--- a/src/bit-systems/link-hover-menu.ts
+++ b/src/bit-systems/link-hover-menu.ts
@@ -1,98 +1,150 @@
-import { defineQuery, enterQuery, entityExists } from "bitecs";
+import { defineQuery, entityExists } from "bitecs";
 import type { HubsWorld } from "../app";
-import { Interacted, Link, LinkHoverMenu, LinkHoverMenuItem, HoveredRemoteRight, TextTag } from "../bit-components";
-import { anyEntityWith, findChildWithComponent } from "../utils/bit-utils";
-import { isHubsRoomUrl, isLocalHubsAvatarUrl, isLocalHubsSceneUrl, isLocalHubsUrl } from "../utils/media-url-utils";
+import { Link, LinkHoverMenu, HoveredRemoteRight, TextTag, Interacted, LinkHoverMenuItem } from "../bit-components";
+import { findAncestorWithComponent, findChildWithComponent } from "../utils/bit-utils";
+import { hubIdFromUrl } from "../utils/media-url-utils";
 import { Text as TroikaText } from "troika-three-text";
 import { handleExitTo2DInterstitial } from "../utils/vr-interstitial";
 import { changeHub } from "../change-hub";
+import { EntityID } from "../utils/networking-types";
+import { setMatrixWorld } from "../utils/three-utils";
+import { LinkType } from "../inflators/link";
 
-const NULL_EID = 0;
-
-const hoveredQuery = defineQuery([HoveredRemoteRight, Link]);
-const hoveredEnterQuery = enterQuery(hoveredQuery);
-const hoveredMenuItemQuery = defineQuery([HoveredRemoteRight, LinkHoverMenuItem]);
+const menuQuery = defineQuery([LinkHoverMenu]);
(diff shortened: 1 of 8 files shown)


src/utils/media-url-utils.js
```diff
@@ -196,11 +196,10 @@ export const isLocalHubsSceneUrl = async url => (await isHubsSceneUrl(url)) && (
...
 export const isLocalHubsAvatarUrl = async url => (await isHubsAvatarUrl(url)) && (await isLocalHubsUrl(url));
 
+export const hubIdFromUrl = url => url.match(hubsRoomRegex)?.groups.id;
+
 export const isHubsRoomUrl = async url =>
-  (await isHubsServer(url)) &&
-  !(await isHubsAvatarUrl(url)) &&
-  !(await isHubsSceneUrl(url)) &&
-  url.match(hubsRoomRegex)?.groups.id;
+  (await isHubsServer(url)) && !(await isHubsAvatarUrl(url)) && !(await isHubsSceneUrl(url)) && hubIdFromUrl(url);
```
takahirox, line 202: What is this change for? Just for clean up?

src/inflators/link.ts
```diff
@@ -2,12 +2,23 @@ import { addComponent } from "bitecs";
...
+
 export type LinkParams = {
   href: string;
+  type?: LinkType;
 };
 
 export function inflateLink(world: HubsWorld, eid: number, params: LinkParams): number {
   addComponent(world, Link, eid);
   Link.url[eid] = APP.getSid(params.href);
+  Link.type[eid] = params.type || LinkType.LINK;
```
takahirox, line 22: Nit: I prefer `params.type !== params.type ? params.type : LinkType.Link` or something similar because it would be clearer and safer.

`params.type || LinkType.LINK` works because `LinkType.LINK` is zero. But imagine a case that a future new developer changes `LinkType` enum values for a certain purpose and the one (and reviewers) don't notice there is a line relying on `LinkType.LINK` is zero.

src/components/open-media-button.js
```diff
@@ -60,7 +60,7 @@ AFRAME.registerComponent("open-media-button", {
         if (url.hash && window.APP.hub.hub_id === hubId) {
           // move to waypoint w/o writing to history
           window.history.replaceState(null, null, window.location.href.split("#")[0] + url.hash);
-        } else if (isLocalHubsUrl(this.src)) {
+        } else if (await isLocalHubsUrl(this.src)) {
```
takahirox, line 63: Good catch

src/change-hub.js
```diff
@@ -29,7 +29,7 @@ function loadRoomObjects(hubId) {
   objectsScene.appendChild(objectsEl);
 }
 
-export async function changeHub(hubId, addToHistory = true, waypoint = null) {
+export async function changeHub(hubId, addToHistory = true, waypoint = "") {
```
takahirox, line 32: What is this change for?

src/bit-systems/link-system.ts
```diff
@@ -0,0 +1,30 @@
...
+import type { HubsWorld } from "../app";
+import { Link } from "../bit-components";
+import { isHubsRoomUrl, isLocalHubsAvatarUrl, isLocalHubsSceneUrl, isLocalHubsUrl } from "../utils/media-url-utils";
+import { LinkType } from "../inflators/link";
+
+const linkQuery = defineQuery([Link]);
+const linkEnterQuery = enterQuery(linkQuery);
+
+export function linkSystem(world: HubsWorld) {
+  linkEnterQuery(world).forEach(async eid => {
```
takahirox, line 11: I think we should avoid async/await in systems. We should use yield instead. If using async/await for a certain purpose please add inline comments.

src/bit-systems/link-hover-menu.ts
```diff
@@ -1,47 +1,150 @@
...
+    return;
+  }
+
+  if (world.time.elapsed > LinkHoverMenu.clearTargetTimer[menu]) {
+    LinkHoverMenu.targetObjectRef[menu] = 0;
+    return;
+  }
+}
+
+async function handleLinkClick(world: HubsWorld, button: EntityID) {
```
takahirox, line 48: Similar to others, avoid async/await in systems.

src/bit-systems/link-hover-menu.ts
```diff
@@ -1,47 +1,150 @@
...
+    return;
+  }
+
+  if (sceneIsFrozen) {
+    LinkHoverMenu.targetObjectRef[menu] = 0;
+    return;
+  }
+
+  const hovered = hoveredQuery(world);
+  const target = hovered.map(eid => findAncestorWithComponent(world, Link, eid))[0] || 0;
```
takahirox, line 30: Isn't here inefficient if only the first element of hovered is used? If I'm right `.map` applys a function to all the elements in an array.

src/bit-systems/link-hover-menu.ts
```diff
@@ -1,47 +1,150 @@
...
+  const src = APP.getString(Link.url[linkEid])!;
+  const url = new URL(src);
+  const linkType = Link.type[linkEid];
+  switch (linkType) {
+    case LinkType.LINK:
+      await exitImmersive();
+      window.open(src);
+      break;
+    case LinkType.AVATAR:
+      const avatarId = new URL(src).pathname.split("/").pop();
```
takahirox, line 62: `new URL(src)` may be replaced with `url`

src/utils/media-url-utils.js
takahirox: Ah, ok. Exporting `hubIdFromUrl()` function as a utility function.

src/bit-systems/link-hover-menu.ts
```diff
@@ -1,47 +1,150 @@
...
+      APP.scene!.emit("scene_media_selected", src);
+      break;
+    case LinkType.WAYPOINT:
+      // move to waypoint w/o writing to history
+      window.history.replaceState(null, "", window.location.href.split("#")[0] + url.hash);
+      break;
+    case LinkType.ROOM:
+      const waypoint = url.hash && url.hash.substring(1);
+      // move to new room without page load or entry flow
+      const hubId = hubIdFromUrl(url);
```
takahirox, line 76: Off topic from this PR: Question, do we ensure that url is valid somewhere? If not, we may add validations in inflators or somewhere else (as we discussed validations before).

Validations may be added in another PR.

src/bit-systems/link-hover-menu.ts
```diff
@@ -1,47 +1,150 @@
...
-  hoveredEnterQuery(world).forEach((eid: number) => {
-    LinkHoverMenu.targetObjectRef[menuEid] = eid;
-    const targetObject = world.eid2obj.get(eid)!;
-    targetObject.add(menuObject);
-  });
+function updateLinkMenuTarget(world: HubsWorld, menu: EntityID, sceneIsFrozen: boolean) {
+  if (LinkHoverMenu.targetObjectRef[menu] && !entityExists(world, LinkHoverMenu.targetObjectRef[menu])) {
+    // Clear the invalid entity reference. (The link entity was removed).
+    LinkHoverMenu.targetObjectRef[menu] = 0;
+    return;
```
takahirox, line 21: Do we really need to return here? For example, imagine a case that the target entity of the previous frame has been deleted and another target entity is hovered in this frame.

src/bit-systems/link-hover-menu.ts
takahirox, line 30: I want to understand the reason why `Link` and `LinkHoverMenu` (below) is found from the ancestors because I'm not so familiar with our menu systems. Do we add `RemoteHoverTarget` component to descendant entities of `Link/LinkHoverMenu` entities?

takahirox, line 30: Isn't here inefficient if only the first element of hovered is used? If I'm right .map applys a function to all the elements in an array

takahirox, line 30: I want to understand the reason why Link and LinkHoverMenu (below) is found from the ancestors because I'm not so familiar with our menu systems. Do we add RemoteHoverTarget component to descendant entities of Link/LinkHoverMenu entities?

src/bit-systems/link-hover-menu.ts
```diff
@@ -1,47 +1,150 @@
...
+    case LinkType.ROOM_URL:
+      label = "visit room";
+      break;
+  }
+  textObj.text = label;
+}
+
+function flushToObject3Ds(world: HubsWorld, menu: EntityID, frozen: boolean) {
+  const target = LinkHoverMenu.targetObjectRef[menu];
+  const visible = !!(target && !frozen);
```
takahirox, line 120: Nit: Perhaps `!!target && !frozen` would be simpler?

src/bit-systems/link-hover-menu.ts
```diff
@@ -1,47 +1,150 @@
...
   } else {
-    // Not hovered or target object has already been deleted
-    // then make the menu invisible and forget the Link object.
-    menuObject.visible = false;
-    if (menuObject.parent !== null) {
-      menuObject.parent.remove(menuObject);
-    }
-    LinkHoverMenu.targetObjectRef[menuEid] = NULL_EID;
+    buttonObj.visible = false;
   }
```
takahirox, line 136: If I'm right, unexpectedly text won't be updated in this scenario, will be?

1. A Link entity is hovered and button is visible
2. At the next frame, another different typed Link entity is hovered
3. Text needs to be updated because of different link type
4. But text is not updated because button is visible in both this frame and the previous frame

src/bit-systems/link-hover-menu.ts
```diff
@@ -1,47 +1,150 @@
...
+      label = "use scene";
+      break;
+    case LinkType.WAYPOINT:
+      label = "go to";
+      break;
+    case LinkType.ROOM:
+    case LinkType.ROOM_URL:
+      label = "visit room";
+      break;
+  }
```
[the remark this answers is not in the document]
keianhzo: I guess this is mostly because `textObj.text` doesn't accept undefined and if you not assign upfront the compiler complains because it can be undefined. We can initialize as empty and keep the case if that makes it clearer.

src/bit-systems/link-hover-menu.ts
keianhzo: Yeah, I think that  makes sense. Removed.

src/bit-systems/link-hover-menu.ts
keianhzo: Yes, that makes sense. I've replace this one and `some` with `find`

keianhzo: First we check if we are hovering over the interactable media here and find the link element:
`const target = hovered.find(eid => findAncestorWithComponent(world, Link, eid));`

Then we check if we are hovering over the menu:
`if (hovered.find(eid => findAncestorWithComponent(world, LinkHoverMenu, eid)))`
And as is the Link Button the one catching the focus, we need to look for the parent entity which has the `LinkHoverMenu` element.

So for the first iteration where we look for the Link I think we can just check for `hasComponent`.

src/bit-systems/link-hover-menu.ts
keianhzo: Good catch, that's right, we need to check for target updates too.

src/bit-systems/link-hover-menu.ts
keianhzo: Not sure what's the best way of converting this to coroutines...

src/change-hub.js
takahirox: If I remember correctly types that can be recognized by typescript compiler can be specified even in JavaScript with @types. But I think defaulting with "" is also ok (and probably it's simpler).

src/bit-systems/link-hover-menu.ts
takahirox: I guess you have missed to update here?

src/bit-systems/link-hover-menu.ts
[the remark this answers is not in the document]
takahirox: I think initializing with empty string is ok, thanks.

src/inflators/link.ts
```diff
@@ -2,12 +2,23 @@ import { addComponent } from "bitecs";
...
+
 export type LinkParams = {
   href: string;
+  type?: LinkType;
 };
 
 export function inflateLink(world: HubsWorld, eid: number, params: LinkParams): number {
   addComponent(world, Link, eid);
   Link.url[eid] = APP.getSid(params.href);
+  Link.type[eid] = params.type !== undefined ? params.type : LinkType.LINK;
```
takahirox, line 22: If I understand correctly, `Link.type` can be asynchronously updated in link system depending on what url(src) actually points to.

`Link.type` can be accessed from link-hover-menu system (or future new systems) even before updating.

Isn't it confusing and less maintainable? Link label and click action can differ between before and after update the type.

Some random ideas to simplify.

1. We don't update type depending on what url(src) actually points to. We just respect Link component initial parameter. (Then we may need to update Hubs Blender-addon to force to specify type parameter properly?)
2. Introduce Link Initialize component and system and the system asynchronously resolves the url and then set Link component.

Any thoughts?

src/bit-systems/link-hover-menu.ts
```diff
@@ -1,47 +1,154 @@
...
+  // TODO: Take into account for more than one for VR
+  menuQuery(world).forEach(menu => {
+    const prevTarget = LinkHoverMenu.targetObjectRef[menu];
+    updateLinkMenuTarget(world, menu, sceneIsFrozen);
+    const currTarget = LinkHoverMenu.targetObjectRef[menu];
+    if (currTarget) {
+      moveToTarget(world, menu);
+      clickedMenuItemQuery(world).forEach(eid => handleLinkClick(world, eid));
+    }
+    flushToObject3Ds(world, menu, sceneIsFrozen, prevTarget != currTarget);
```
takahirox, line 152: Let's use `!==` instead of `!=` for the consistency.

```suggestion
    flushToObject3Ds(world, menu, sceneIsFrozen, prevTarget !== currTarget);
```

src/inflators/link.ts
takahirox: If we want to avoid race and go with this PR, we may choose option 1 for now. And we may revisit later about auto detection in another PR. Even without auto type detection, supporting link types should be much more valuable than nothing.

src/bit-systems/link-hover-menu.ts
takahirox: Maybe ok to add inline comment for now, and rewrite properly later?

src/bit-systems/link-hover-menu.ts
```diff
@@ -1,47 +1,154 @@
...
+    LinkHoverMenu.targetObjectRef[menu] = 0;
+    return;
+  }
+
+  if (sceneIsFrozen) {
+    LinkHoverMenu.targetObjectRef[menu] = 0;
+  }
+
+  const hovered = hoveredQuery(world);
+  const target = hovered.find(eid => hasComponent(world, Link, eid));
```
takahirox, line 29: Can't here be replaced with the following?

```
const hoveredLinkQuery = defineQuery([HoveredRemoteRight, Link]);
...
const hoveredLinks = hoveredLinkQuery(world);
if (hoveredLinks.length > 0) {
  target = hoveredLinks[0];
  ...
```

src/bit-systems/link-hover-menu.ts
takahirox: Sorry I think I wrote my question poorly. What I really wanted to ask was, it seems to be assumed that `Link` entity can be in ancestors, not in descendants, of `LinkHoverMenu` entity, and similarly `LinkHoverMenu` entity can be in ancestors. Who forms or ensures this relation? Similarly,  A certain system? Prefab? Blender add-on?

I don't really like having dependency from an entity to another entity, especially finding a target entity from its ancestors or descendants because it requires new readers to know a lot of additional contexts that are not indicated in a system.

This comment doesn't block this PR. I just wanted to share my thought for future clean up.

src/bit-systems/link-hover-menu.ts
keianhzo: Oh yep, that's better. Updated.

src/bit-systems/link-hover-menu.ts
keianhzo: Looking at this again I don't thing this part makes sense:
```typescript
if (hovered.find(eid => findAncestorWithComponent(world, LinkHoverMenu, eid))) {
    LinkHoverMenu.clearTargetTimer[menu] = world.time.elapsed + 1000;
    return;
  }
```
I wrote this a long time ago so I don't really remember but I think we don't really need to look for the `LinkHoverMenu` in any case so this can be greatly simplified and just get the first hovered link that we find as you have pointed out in another comment. I've updated with that.

src/inflators/link.ts
keianhzo: Yes that's true. This is based on the assumption that the link type is set by the links system very early when the interactable is spawned and between that and hovering there is going to be enough time to guess the type. This is clearly wrong because for cached assets the interactable can be spawned quite fast and in slow connections the type guessing can take more type and by that time the interactable can be already hovered.

Ragarding this:
> We don't update type depending on what url(src) actually points to. We just respect Link component initial parameter. (Then we may need to update Hubs Blender-addon to force to specify type parameter properly?)

I don't think we can do that as links are also used in spawned interactables (when you paste a link) and there there is no upfront type, we need to guess.

So I guess n.1 is the best option although we kind of are in the same race condition situation where we can't really show a link if the user hovers while the link type is still being initialized although that shouldn't take much time.

I think I'm going to add a comment there with all this info and address in a later PR as I don't have so much time now to work on this.

takahirox: > I don't think we can do that as links are also used in spawned HTML interactables (when you paste a link) and there there is no upfront type, we need to guess.

Then we may move the auto link type detection from `Link` to HTML pasting handling system. If we do that, `Link` system will be simpler. No async initialization will be needed and `Link` component data will be immutable.

Anyways, we may revisit it later. The current behavior in this PR is ok for now.

src/bit-systems/link-system.ts
```diff
@@ -0,0 +1,44 @@
...
+      Link.type[eid] = LinkType.ROOM_URL;
+    }
+  }
+  removeComponent(world, LinkInitializing, eid);
+}
+
+const jobs = new JobRunner();
+export function linkSystem(world: HubsWorld) {
+  linkInitializingEnterQuery(world).forEach(eid => {
+    jobs.stop(eid);
```
takahirox, line 37: What is this `jobs.stop(eid)` here for?

takahirox (approved):
Looking good once https://github.com/mozilla/hubs/pull/6110#discussion_r1289373984 will be clear

src/bit-systems/link-hover-menu.ts
takahirox: > I was just testing it and we are not handling that case at all so we probably need to add that in a follow-up PR

Maybe we need preprocessing glTF content

src/bit-systems/link-system.ts
keianhzo: Oh yep, copy paste mistake, that doesn't make sense there. Thanks

PR #6111, opened 2023-06-02
Add support for media video and image parameters

keianhzo:
This PR adds support for missing media video parameters.


src/inflators/video-loader.ts
```diff
@@ -19,5 +21,9 @@ export function inflateVideoLoader(world: HubsWorld, eid: number, params: VideoL
     isObjectMenuTarget: false
   });
 
-  // TODO: Use the rest of VideoLoaderParams
+  const { autoPlay, controls, loop, projection } = params;
+  autoPlay !== false && (MediaVideo.flags[eid] |= VIDEO_FLAGS.AUTOPLAY);
+  loop !== false && (MediaVideo.flags[eid] |= VIDEO_FLAGS.LOOP);
+  controls !== false && (MediaVideo.flags[eid] |= VIDEO_FLAGS.CONTROLS);
+  MediaVideo.projection[eid] = APP.getSid(projection != undefined ? projection : ProjectionMode.FLAT);
```
takahirox, line 28: Please use `!==` instead of `!=`.

src/bit-components.js
```diff
@@ -167,9 +167,14 @@ export const SceneLoader = defineComponent({ src: Types.ui32 });
...
 export const MediaImage = defineComponent({
-  cacheKey: Types.ui32
+  cacheKey: Types.ui32,
+  projection: Types.ui8,
+  alphaMode: Types.ui8,
+  alphaCutoff: Types.f32
 });
 MediaImage.cacheKey[$isStringType] = true;
+MediaImage.projection[$isStringType] = true;
+MediaImage.alphaMode[$isStringType] = true;
```
takahirox, line 177: If I'm right, `projection` and `alphaMode` are not really needed to be strings in the component, aren't they? They seem to be used just as type (enum).

src/utils/load-audio.tsx
```diff
@@ -5,10 +5,32 @@ import { renderAsEntity } from "../utils/jsx-entity";
...
+    MediaVideo.flags[eid] |= VIDEO_FLAGS.AUTOPLAY;
+    MediaVideo.flags[eid] |= VIDEO_FLAGS.LOOP;
+    MediaVideo.flags[eid] |= VIDEO_FLAGS.CONTROLS;
+  }
+  let projection = ProjectionMode.FLAT;
+  if (MediaVideo.projection[eid]) {
+    projection = APP.getString(MediaVideo.projection[eid]) as ProjectionMode;
+  } else {
+    MediaVideo.projection[eid] = APP.getSid(ProjectionMode.FLAT);
+  }
```
takahirox, line 30: I don't get these `MediaVideo.flags[eid]` and `MediaVideo.projection[eid]` checking are for. Does it mean an entity has `MediaVideo` component but thoese properties are not set? If so, how does it happen?

src/utils/load-image.tsx
```diff
@@ -29,18 +29,29 @@ export function* createImageDef(world: HubsWorld, url: string, contentType: stri
...
+  if (MediaImage.alphaMode[eid]) {
+    imageDef.alphaMode = APP.getString(MediaImage.alphaMode[eid]) as AlphaMode;
+  } else {
+    imageDef.alphaMode = AlphaMode.Opaque;
+  }
+  if (MediaImage.alphaCutoff[eid]) {
+    imageDef.alphaCutoff = MediaImage.alphaCutoff[eid];
+  } else {
+    imageDef.alphaCutoff = 0.5;
+  }
```
takahirox, line 54: I don't really get these `MediaImage` properties checking are for. An entity has `MediaImage` component but those properties are not set? If so, how does it happen?

Or, checking whether an entity has `MediaImage` component? If so, I don't think accessing component properties that is not assigned to an entity is safe. `hasComponent()` should be used instead.

src/inflators/video.ts
```diff
@@ -0,0 +1,51 @@
...
+      : createImageMesh(texture, ratio);
+  addObject3DComponent(world, eid, mesh);
+  addComponent(world, MediaVideo, eid);
+
+  if (autoPlay) {
+    MediaVideo.flags[eid] |= VIDEO_FLAGS.AUTOPLAY;
+  }
+  if (loop) {
+    MediaVideo.flags[eid] |= VIDEO_FLAGS.LOOP;
+  }
```
takahirox, line 40: Thinking whether autoplay and loop property should be managed as flag in `MediaVideo` component because they are duplicated with `video.loop` and `video.autoplay`.

src/utils/jsx-entity.ts
```diff
@@ -272,16 +272,10 @@ export interface JSXComponentData extends ComponentData {
     texture: Texture;
     ratio: number;
     projection: ProjectionMode;
-    alphaMode: typeof AlphaMode.Blend | typeof AlphaMode.Mask | typeof AlphaMode.Opaque;
+    alphaMode: AlphaMode.Blend | AlphaMode.Mask | AlphaMode.Opaque;
```
takahirox, line 275: Can't this line be `alphaMode: AlphaMode;`?

src/inflators/image-loader.ts
```diff
@@ -16,5 +20,8 @@ export function inflateImageLoader(world: HubsWorld, eid: number, params: ImageL
     isObjectMenuTarget: false
   });
 
-  // TODO: Use projection
+  const { projection, alphaMode, alphaCutoff } = params;
+  MediaImage.projection[eid] = APP.getSid(projection !== undefined ? projection : ProjectionMode.FLAT);
+  MediaImage.alphaMode[eid] = APP.getSid(alphaMode !== undefined ? alphaMode : AlphaMode.Opaque);
+  MediaImage.alphaCutoff[eid] = alphaCutoff !== undefined ? alphaCutoff : 0.5;
```
takahirox, line 26: An entity is assumed that it has `MediaImage` component? I don't think we should assume an entity in inflators alerady has certain components.

takahirox:
I added "P1" (high priority) label to this PR.

I got a feedback from the team that the loop of an audio in a scene seems to have a problem that the audio would stop after a certain amount of time although it's expected to be looped.

This may be a very noticable problem with audio. Users will have bad user experience. I think this problem should be fixed by the initial release.

I guess this PR would fix the problem.

src/bit-components.js
keianhzo: The enum is string based I think to make it easier to handle comparisons as those parameters as string in the glTFs.

src/utils/load-audio.tsx
keianhzo: Yeah this is very confusing. There are two path how we load images (same for the rest of media):
- Spawned Media:
**MediaPrefab -> inflateMediaLoader -> [media-loading-system] -> loadImage -> inflateImage (MediaImage component added)**
- glTF:
**inflateImageLoader -> inflateMediaLoader -> [media-loading-system] -> loadImage -> inflateImage (MediaImage component added)**

I am using `MediaImage` as a transport container for the `ImageLoaderParams` that are later gathered in `inflateImageLoader` and used in `loadImage` as it assumes that the `MediaImage` has been added an initialized already at that point. As you pointed out this is not correct as the `MediaImage` component has not been added yet.

I'm going to introduce a `MediaImageLoaderData` and `MediaVideoLoaderData` data stores in between `inflateImageLoader` and `loadImage` to safely pass the initialization parameters for all medias. I think that would solve most of the concerns in this PR.

src/utils/load-image.tsx
keianhzo: See https://github.com/mozilla/hubs/pull/6111#discussion_r1288496997

src/inflators/image-loader.ts
keianhzo: https://github.com/mozilla/hubs/pull/6111#discussion_r1288496997

src/inflators/video.ts
keianhzo: Yeah that makes sense, removed.

src/inflators/video.ts
```diff
@@ -0,0 +1,41 @@
...
+  const { texture, ratio, projection, video, controls } = params;
+  const mesh =
+    projection === ProjectionMode.SPHERE_EQUIRECTANGULAR
+      ? create360ImageMesh(texture)
+      : createImageMesh(texture, ratio);
+  addObject3DComponent(world, eid, mesh);
+  addComponent(world, MediaVideo, eid);
+
+  if (controls) {
+    MediaVideo.flags[eid] |= VIDEO_FLAGS.CONTROLS;
```
takahirox, line 32: Question: Is component data initialized as zero when the component is added to an entity? Eg: In case an entity is recycled. If not, we need `MediaVideo.flags[eid] = 0` above.

src/inflators/video.ts
```diff
@@ -0,0 +1,41 @@
...
+  }
+  if (projection) {
+    MediaVideo.projection[eid] = APP.getSid(projection);
```
takahirox, line 35: Same above.

src/utils/load-audio.tsx
takahirox: Thanks for explaining it. Give me a little time to think whether we can simplify further...

src/bit-components.js
takahirox: Can you elaborate this for me a bit more, please?

> handle comparisons as those parameters as string in the glTFs

What I though is, if I'm right, `projection` and `alphaMode` are finally used just for comparing like `projection === ProjectionMode.SPHERE_EQUIRECTANGULAR` or `switch(alphaMode) { case AlphaMode.Opaqueue: ...` in Hubs Client. So I don't think the enum values in Hubs Client really need to be strings. Instead they can be just integer and component data type can be simply `ui8`. Strings doesn't really fit to bitECS API and we use hack. Probably it would be good to avoid strings in component data where possible.

But this is kind of optimization. We may think of in another PR.

By the way, if you want to use strings, better to use `ui32` component data type. `ui8` can be easily shortage.

src/inflators/video.ts
```diff
@@ -0,0 +1,41 @@
...
+import { addComponent } from "bitecs";
+import { addObject3DComponent } from "../utils/jsx-entity";
+import { ProjectionMode } from "../utils/projection-mode";
+import { MediaVideo, MediaVideoData } from "../bit-components";
+import { HubsWorld } from "../app";
+import { EntityID } from "../utils/networking-types";
+import { Texture } from "three";
+
+export const VIDEO_FLAGS = {
+  CONTROLS: 1 << 2
```
takahirox, line 11: Curious, any special reason why `1 << 2` here? For example, just `1` would look like good enough to me.

src/utils/jsx-entity.ts
```diff
@@ -265,23 +265,17 @@ type OptionalParams<T> = Partial<T> | true;
 export interface JSXComponentData extends ComponentData {
   slice9?: {
     size: [width: number, height: number];
-    insets: [top: number, buttom: number, left: number, right: number];
+    insets: [top: number, bottom: number, left: number, right: number];
```
takahirox, line 268: Good catch

src/utils/load-audio.tsx
takahirox: Probably the root cause of the complexity is MediaLoader system is too big and operate too many things. If media type specific operation is needed, they can't be decoupled well from MediaLoader system.

For example, We may separate media loder system into media type specific loader system like ImageLoaderSystem, VideoLoaderSystem, ModelLoaderSystem, and common operations can be exported as utils that can be used from the new loaders. Each loader would be simplerer.

But I don't think we have time to refactor. So transporting as temporary solution would be ok.

The current `MediaImageLoaderData` and `MediaVideoLoaderData` approach may be a bit less maintainable. They are as if just global variables and future readers may feel hard to understand how they are properly managed. And it is decoupled from bitecs component and entity mechanism. So for example even if an component or entity is removed while the operation, the data is remained. It can be error prone, especially if the entity is recycled.

Perhaps managing the transporting data as component is safer? Give me more time for thinking proper temporary solution...

takahirox: Clean up idea https://github.com/mozilla/hubs/commit/3d2f38102ac052c1a7fa7834a02745d3b47bf750

* Treat MediaImage/VideoLoaderData as part of MediaLoader component data. Less problematic because their life cycle will be synched with MediaLoader component.
* Remove dependency with entity and MediaImage/VideoLoaderData from loaders in src/utils. The loaders will be simpler.

Feel free to cherry-pick if it looks good, or I can make a follow up PR after this PR lands.

src/inflators/video.ts
keianhzo: Copy paste fail, fixed. Thanks.

src/inflators/video.ts
keianhzo: Yes, it needs to be initialized, fixed. Thanks

src/inflators/video.ts
keianhzo: Same. Fixed.

src/utils/load-audio.tsx
keianhzo: Thanks, sounds good to me. Maybe better to cleanup this after landing. Feel free to update that when this lands and I'll review.

src/bit-components.js
keianhzo: Good call, I didn't pay much attention to it and I'm not sure why we decided to make them strings early on but I've migrated them to numbers to simplify components.

keianhzo:
@takahirox I've addressed the feedback. When I have the final approval I'll land

src/utils/load-image.tsx
```diff
@@ -29,18 +29,31 @@ export function* createImageDef(world: HubsWorld, url: string, contentType: stri
...
+    } else if (params.projection === PROJECTION_MODE.SPHERE_EQUIRECTANGULAR) {
+      imageDef.projection = ProjectionMode.SPHERE_EQUIRECTANGULAR;
+    }
+    if (params.alphaMode === ALPHA_MODE.OPAQUE) {
+      imageDef.alphaMode = AlphaMode.Opaque;
+    } else if (params.alphaMode === ALPHA_MODE.BLEND) {
+      imageDef.alphaMode = AlphaMode.Blend;
+    } else if ((params.alphaMode = ALPHA_MODE.MASK)) {
+    }
+    imageDef.alphaMode === AlphaMode.Mask;
```
takahirox, line 53: I guess misstyped here

```suggestion
    } else if ((params.alphaMode = ALPHA_MODE.MASK)) {
      imageDef.alphaMode === AlphaMode.Mask;
    }
```

src/utils/projection-mode.ts
```diff
@@ -1,4 +1,8 @@
 export enum ProjectionMode {
+  FLAT = 0,
+  SPHERE_EQUIRECTANGULAR = 1
+}
+export const enum PROJECTION_MODE {
```
takahirox, line 5: Maybe having `ProjectionMode` and `PROJECTION_MODE` enums may be confusing because the names are similar? How about renaming the latter to `ProjectionModeName` or something like that?

src/inflators/image.ts
```diff
@@ -0,0 +1,46 @@
...
+  if (alphaMode !== undefined) {
+    MediaImage.alphaMode[eid] = alphaMode;
+  } else {
+    MediaImage.alphaMode[eid] = AlphaMode.Opaque;
+  }
+  if (alphaCutoff !== undefined) {
+    MediaImage.alphaCutoff[eid] = alphaCutoff;
+  } else {
+    MediaImage.alphaCutoff[eid] = 0;
+  }
```
takahirox, line 42: Not a blocking issue but a comment for future clean up: Here can be similified to use a common pattern as [other inflators](https://github.com/mozilla/hubs/blob/master/src/inflators/ambient-light.ts) do

```typescript
export type XXXParams = {
  foo?: string;
  bar?: number;
};

const DEFAULTS: Required<XXXParams> = {
  foo: 1.0,
  bar: 'default'
};

function xxxSystem(world: HubsWorld, eid: EntityID, params: XXXParams) {
  const requiredParams = Object.assign({}, DEFAULTS, params) as Required<XXXParams>;
  addComponent(world, XXX, eid);
  XXX.foo[eid] = requiredParams.foo;
  XXX.bar[eid] = requiredParams.bar;
  ...
}
```

src/inflators/video-loader.ts
```diff
@@ -1,10 +1,10 @@
 import { HubsWorld } from "../app";
-import { ProjectionMode } from "../utils/projection-mode";
+import { MediaVideoLoaderData } from "../bit-components";
 import { inflateMediaLoader } from "./media-loader";
 
 export interface VideoLoaderParams {
   src: string;
-  projection: ProjectionMode;
+  projection?: string;
```
takahirox, line 7: Not blocking issue but just a question: Why only `projection` is optional in `VideoLoaderParams`?

src/utils/create-image-mesh.ts
```diff
@@ -1,20 +1,27 @@
...
+  Opaque = 0,
+  Blend = 1,
+  Mask = 2
+}
+
+export const enum ALPHA_MODE {
+  OPAQUE = "opaque",
+  BLEND = "blend",
+  MASK = "mask"
+}
```
takahirox, line 16: 1. Similar to `ProjectionMode`, maybe having `AlphaMode` and `ALPHA_MODE` enums may be confusing because the names are similar? How about renaming the latter to `AlphaModeName` or something like that?
2. Not a blocking issue but a future suggestion: There might be another properer place to define these enums, eg: `src/utils/alpha-mode.ts` similar to `src/utils/projection-mode.ts`.

src/inflators/image-loader.ts
```diff
@@ -1,10 +1,13 @@
 import { HubsWorld } from "../app";
 import { ProjectionMode } from "../utils/projection-mode";
 import { inflateMediaLoader } from "./media-loader";
+import { MediaImageLoaderData } from "../bit-components";
 
 export interface ImageLoaderParams {
   src: string;
-  projection: ProjectionMode;
+  projection: string;
+  alphaMode: string;
```
takahirox, line 9: Nit: Maybe better to use `PROJECTION_MODE` and `ALPHA_MODE` types here? Probably you use strings type here because this inflator is (currently) only for glTF and no guarantee that valid data comes from glTF. But expected type will be clearer if we use `PROJECTION_MODE` and `ALPHA_MODE`. And it would be safer even if we end up using this inflator for jsx in the future.

SImilarly, in `video-loader.ts`.

takahirox (approved):
Added some minor comments but looking good. Please reflect the review comments (if needed) and merge. 

After this PR lands, I will make a follow up PR to clean up. https://github.com/mozilla/hubs/commit/7dc5aed3b5df12db4872af07f97806d3eab8fdd2

src/inflators/video-loader.ts
keianhzo: I don't think it is probably a copy&paste issue, updated

src/utils/create-image-mesh.ts
keianhzo: Updated

src/utils/load-audio.tsx
```diff
@@ -5,10 +5,25 @@ import { renderAsEntity } from "../utils/jsx-entity";
...
+export function* loadAudio(world: HubsWorld, eid: EntityID, url: string) {
+  let loop = true;
+  let autoPlay = true;
+  let controls = true;
+  let projection = ProjectionMode.FLAT;
+  if (MediaVideoLoaderData.has(eid)) {
+    const params = MediaVideoLoaderData.get(eid)! as VideoLoaderParams;
+    loop = params.loop;
+    autoPlay = params.autoPlay;
+    controls = params.controls;
```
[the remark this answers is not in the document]
keianhzo: I think we only want FLAT projection for audio only video elements, right? It defaults to FLAT for audio.

j-conrad:
Any chance of this getting deployed soon? Video and audio looping is becoming important for Behavior Graph testing as well as general Hubs scenes.

keianhzo:
@j-conrad This is already deployed in the main Hubs client branch but not in the BGs branch. I'm waiting for us to land the new loader before rebasing and merging as there are still too often updates.

PR #6116, opened 2023-06-07
feat: Add pronoun field to entry modal for users

nikk15:
From this UX Blast document: https://docs.google.com/document/d/13K_N5eBXk0aTBlHVoPUPIxpSWFeY2rP2EafVqS30fg4/edit?skip_itp2_check=true#heading=h.xcm1taufnky6

- Added pronoun fields to entry modal and avatar set up
- Added pronoun display to user select, guest lists and nametags

Screenshot 2023-06-09 at 4 00 08 PM
Screenshot 2023-06-09 at 4 00 43 PM
Screenshot 2023-06-09 at 4 00 56 PM
Screenshot 2023-06-09 at 4 01 57 PM
Screenshot 2023-06-09 at 4 00 22 PM

### DIFF
--- a/src/components/name-tag.js
+++ b/src/components/name-tag.js
@@ -40,6 +40,7 @@ AFRAME.registerComponent("name-tag", {
   init() {
     this.store = window.APP.store;
     this.displayName = null;
+    this.pronouns = null;
     this.identityName = null;
     this.isTalking = false;
     this.isTyping = false;
@@ -198,6 +199,7 @@ AFRAME.registerComponent("name-tag", {
 
   updateFromPresenceMeta(presenceMeta) {
     this.displayName = presenceMeta.profile.displayName;
+    this.pronouns = presenceMeta.profile.pronouns;
     this.identityName = presenceMeta.profile.identityName;
     this.isRecording = !!(presenceMeta.streaming || presenceMeta.recording);
     this.isOwner = !!(presenceMeta.roles && presenceMeta.roles.owner);
--- a/src/components/name-tag.js
+++ b/src/components/name-tag.js
@@ -219,7 +219,6 @@ AFRAME.registerComponent("name-tag", {
       this.pronounsText?.el && this.pronounsText.el.components["text"].getSize(this.size);
       const pronounsTextSize = this.size.x || 0;
       this.nametagText.el.components["text"].getSize(this.size);
-      console.log(pronounsTextSize, this.size.x);
       this.size.x = Math.max(this.size.x, pronounsTextSize, NAMETAG_MIN_WIDTH);
       this.resizeNameTag();
     }
--- a/src/components/name-tag.js
+++ b/src/components/name-tag.js
@@ -214,19 +214,22 @@ AFRAME.registerComponent("name-tag", {
     }
   },
 
+  updateNametagWidth(name) {
+    if (name) {
+      this.pronounsText?.el && this.pronounsText.el.components["text"].getSize(this.size);
+      const pronounsTextSize = this.size.x || 0;
+      this.nametagText.el.components["text"].getSize(this.size);
+      console.log(pronounsTextSize, this.size.x);
+      this.size.x = Math.max(this.size.x, pronounsTextSize, NAMETAG_MIN_WIDTH);
+      this.resizeNameTag();
+    }
+  },
+
   updateDisplayName() {
     if (this.displayName && this.displayName !== this.prevDisplayName) {
-      this.nametagText.el.addEventListener(
-        "text-updated",
-        () => {
-          if (this.nametagText.el) {
-            this.nametagText.el.components["text"].getSize(this.size);
-            this.size.x = Math.max(this.size.x, NAMETAG_MIN_WIDTH);
-            this.resizeNameTag();
-          }
-        },
-        { once: true }
-      );
+      this.nametagText.el.addEventListener("text-updated", () => this.updateNametagWidth(this.nametagText?.el), {
+        once: true
+      });
       if (this.displayName.length > DISPLAY_NAME_LENGTH) {
         this.displayName = this.displayName.slice(0, DISPLAY_NAME_LENGTH).concat("...");
       }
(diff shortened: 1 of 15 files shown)


src/hub.html
```diff
@@ -109,214 +99,190 @@
...
+                    <a-entity name-tag tags="ignoreSpaceBubble: true;" position="0 0 0"
+                        layers="mask: 768; recursive: true;" billboard="onlyY: true;">
+                        <a-entity class="nametag-background"
+                            slice9="width: 1; height: 0.2; left: 64; top: 64; right: 66; bottom: 66; opacity: 0.5; alphaTest: 0.1; src: nametag;">
+                            <a-entity class="nametag-status-border" visible="false"
+                                slice9="width: 0.45; height: 0.2; left: 64; top: 64; right: 66; bottom: 66; transparent: false; alphaTest: 0.1; src: nametag-border;"></a-entity>
+                            <a-entity class="nametag-text"
+                                text="side: double; textAlign: center; color: #ddd; fontSize: 0.1;" text-raycast-hack
+                                position="0 0.025 0.001"></a-entity>
+                            <a-entity class="pronouns-text"
```
nikk15, line 111: This is the only new line here.

src/components/name-tag.js
```diff
@@ -240,6 +244,32 @@ AFRAME.registerComponent("name-tag", {
     }
   },
 
+  updatePronouns() {
+    if (this.pronouns !== this.prevPronouns) {
+      this.pronounsText.el.addEventListener(
+        "text-updated",
+        () => {
+          if (this.pronounsText.el && this.nameTagText.el) {
+            this.pronounsText.el.components["text"].getSize(this.size);
```
keianhzo, line 253: Here it should be `nametagText` instead of `nameTagText`. For the size evaluation I think we want to set `this.size` to the max between `nametagText`, `pronounText` and `NAMETAG_MIN_WIDTH`. We should also take the pronoun in consideration for the display name resize so maybe we can consolidate all this in `updateDisplayName` and do everything there.

src/storage/store.js
```diff
@@ -65,6 +65,7 @@ export const SCHEMA = {
       properties: {
         displayName: { type: "string", pattern: "^[A-Za-z0-9_~ -]{3,32}$" },
         avatarId: { type: "string" },
+        pronouns: { type: "string", pattern: "^[a-z(?:/)a-z]{6,12}$" },
```
keianhzo, line 68: Nit: If we don’t need the capture groups we can also do: `/^[a-zA-Z///a-zA-Z]{6,12}$/` and also match capitals (maybe we don’t want that, not sure…). Also not sure about this but the min/max length might need to be different in other languages than English so maybe we want to keep it more open to min 1 or 2 characters each and keep the max a bit higher?

src/components/name-tag.js
```diff
@@ -267,6 +297,7 @@ AFRAME.registerComponent("name-tag", {
     this.isAvatarReady = true;
 
     this.updateDisplayName();
+    this.updatePronouns();
```
keianhzo, line 300: I think we would need to also account for the pronouns text element when calculating the max width here:
`this.size.x = Math.max(this.size.x, NAMETAG_MIN_WIDTH);`

keianhzo (changes requested):
If we also want to update the text color based on the theme for the pronoun text elements we should add it here:
`https://github.com/mozilla/hubs/pull/6116/files#diff-f010909f986ed471df83ed99850af7a72b52d0433d806c250731df2f5b767629L282`

Looks mostly good, I just pointed out a few things. Thanks for adding this!

src/storage/store.js
nikk15: oh thank you! good catch on the other languages.

src/components/name-tag.js
```diff
@@ -198,31 +200,35 @@ AFRAME.registerComponent("name-tag", {
...
       this.updateDisplayName();
+      this.updatePronouns();
       this.updateHandRaised();
       this.resizeNameTag();
     }
   },
 
+  updateNametagWidth(name) {
+    if (name) {
+      this.pronounsText?.el && this.pronounsText.el.components["text"].getSize(this.size);
```
keianhzo, line 219: I think both `pronounText` and `nametagText` should exist at this point so we can probably remove all the defensive code from here. We can probably also remove the `name` param and just pass the method name to the listener
`this.nametagText.el.addEventListener("text-updated", this.updateNametagWidth, { once: true})`
as long as we bind it here:
https://github.com/mozilla/hubs/blob/79f2fc11836c4573c546fb980f12c83afd1fb0c4/src/components/name-tag.js#L64

src/components/name-tag.js
```diff
@@ -240,6 +246,21 @@ AFRAME.registerComponent("name-tag", {
     }
   },
 
+  updatePronouns() {
+    if (this.pronouns !== this.prevPronouns) {
+      this.pronounsText.el.addEventListener("text-updated", () => this.updateNametagWidth(this.nametagText?.el), {
```
keianhzo, line 251: Same here

src/components/name-tag.js
```diff
@@ -262,11 +283,14 @@ AFRAME.registerComponent("name-tag", {
       this.avatarAABBSize.y / 2 +
       NAMETAG_OFFSET;
     this.nametagElPosY = this.nametagHeight + (this.isHandRaised ? NAMETAG_OFFSET : 0);
+    this.pronounsText.el && this.pronounsText.el.components["text"].getSize(this.size);
```
keianhzo, line 286: Probably also this check is not necessary

src/react-components/misc/useVolumeMeter.js
```diff
@@ -30,7 +30,6 @@ function updateVolume(analyser, meter) {
 export function useVolumeMeter({ analyser, update, updateRate = 50 }) {
   const movingAvgRef = useRef();
   const meterRef = useRef({ levels: [], volume: 0, prevVolume: 0, max: 0 });
```
keianhzo, line 32: Is this the linter? I think we should be all following the same rules from eslintrc and prettierrc and it doesn't change the format of this file for me... Maybe you have some other additional rules that override those?

keianhzo (changes requested):
This is almost ready, just a couple of nits.

src/components/name-tag.js
nikk15: The check for `nametagText.el` was legacy code, but I removed and you're correct, works fine! 👍

keianhzo (approved):

PR #6119, opened 2023-06-12
BitECS media loading fixes

keianhzo:
This PR addresses some media issues an refactors some code:
- Media with no mesh that triggers media loading is incorrectly positioned in the origin. This PR set a default AABB for positioning the empty root.

https://github.com/mozilla/hubs/assets/837184/75c737dd-d2fb-467e-8235-42d1fb602f3a

(You can test this using a glTF file that has an empty root and triggers a media loading. ie. a video loading [media-video-flat.glb.zip](https://github.com/mozilla/hubs/files/11724425/media-video-flat.glb.zip))

- Refactor to use `MediaLoaded` instead of `MediaLoading` to know when a media is loaded.
- Handles the loading entity removal in the next tick to avoid other systems relaying on the entity (`mixerAnimatableSystem`) crash if the loading cube is inmmediatly removed after adding it.

### DIFF
--- a/src/bit-systems/media-loading.ts
+++ b/src/bit-systems/media-loading.ts
@@ -79,8 +79,9 @@ const loaderForMediaType = {
     { accessibleUrl, contentType }: { accessibleUrl: string; contentType: string }
   ) => loadModel(world, accessibleUrl, contentType, true),
   [MediaType.PDF]: (world: HubsWorld, { accessibleUrl }: { accessibleUrl: string }) => loadPDF(world, accessibleUrl),
-  [MediaType.AUDIO]: (world: HubsWorld, { accessibleUrl }: { accessibleUrl: string }) => loadAudio(world, accessibleUrl),
-  [MediaType.HTML]: (world: HubsWorld, { canonicalUrl, thumbnail }: { canonicalUrl: string, thumbnail: string }) =>
+  [MediaType.AUDIO]: (world: HubsWorld, { accessibleUrl }: { accessibleUrl: string }) =>
+    loadAudio(world, accessibleUrl),
+  [MediaType.HTML]: (world: HubsWorld, { canonicalUrl, thumbnail }: { canonicalUrl: string; thumbnail: string }) =>
     loadHtml(world, canonicalUrl, thumbnail)
 };
 
@@ -99,6 +100,10 @@ function resizeAndRecenter(world: HubsWorld, media: EntityID, eid: EntityID) {
   const mediaObj = world.eid2obj.get(media)!;
   const box = new Box3();
   box.setFromObject(mediaObj);
+
+  // The AABB can be empty here for interactables that fetch  media (ie. gltf with an empty that has a video component).
+  // If we don't return the interactable would be wrongly positioned at the (0,0,0).
+  if (box.isEmpty()) return;
 
   let scalar = 1;
   if (resize) {


src/bit-systems/media-loading.ts
```diff
@@ -99,6 +100,9 @@ function resizeAndRecenter(world: HubsWorld, media: EntityID, eid: EntityID) {
   const mediaObj = world.eid2obj.get(media)!;
   const box = new Box3();
   box.setFromObject(mediaObj);
+  if (box.isEmpty()) {
+    box.setFromCenterAndSize(new Vector3(0, 0, 0), new Vector3(1, 1, 1));
+  }
```
takahirox, line 105: Wondering in what scenarios box can be empty...

If I'm right, aren't visible objects always assigned to media entities by Hubs Client (media loaders) for loaded medias (eg: Plane Mesh for videos and audios)?

src/bit-systems/media-loading.ts
```diff
@@ -176,13 +180,16 @@ type MediaInfo = {
...
 function* loadMedia(world: HubsWorld, eid: EntityID) {
-  let loadingObjEid = 0;
   const addLoadingObjectTimeout = crTimeout(() => {
-    loadingObjEid = renderAsEntity(world, LoadingObject());
-    add(world, loadingObjEid, eid);
+    if (MediaLoader.flags[eid] & MEDIA_LOADER_FLAGS.ANIMATE_LOAD) {
+      const loadingObjEid = renderAsEntity(world, LoadingObject());
+      loadingObjs.set(eid, loadingObjEid);
+      add(world, loadingObjEid, eid);
+    }
```
takahirox, line 191: I thought `MEDIA_LOADER_FLAGS.ANIMATE_LOAD` is enable/disable `animateScale()` and is unrelated t`MEDIA_LOADER_FLAGS.ANIMATE_LOAD`o loading objects but with this change loading objects won't be displayed for non `MEDIA_LOADER_FLAGS.ANIMATE_LOAD`. But no confident about my assumption.

@johnshaughnessy Which is the expected behavior?

src/bit-systems/media-loading.ts
```diff
@@ -216,21 +221,20 @@ function* loadAndAnimateMedia(world: HubsWorld, eid: EntityID, clearRollbacks: C
...
+  MediaContentBounds.bounds[eid].set(tmpVector.toArray());
+
+  // TODO update scale?
+  inflatePhysicsShape(world, media, {
+    type: hasComponent(world, GLTFModel, media) ? Shape.HULL : Shape.BOX,
+    minHalfExtent: 0.04
+  });
+
+  addComponent(world, MediaLoaded, media);
+  removeComponent(world, MediaLoader, eid);
```
takahirox, line 237: Just in case, this changes the timing of when `MediaLoaded` is added to an entity. It was added right after when media is loaded (eg: before `animateScale()` starts) while with this change it is added everything is finished (eg: after `animateScale()` ends). Did you confirm this timing change won't affect any other system? (Sorry I didn't have time to check the other codes by myself.)

src/bit-systems/media-loading.ts
```diff
@@ -243,6 +247,9 @@ export function mediaLoadingSystem(world: HubsWorld) {
   });
 
   mediaLoaderExitQuery(world).forEach(function (eid) {
+    const loadingObjEid = loadingObjs.get(eid)!;
+    removeEntity(world, loadingObjEid);
+    loadingObjs.delete(eid);
```
takahirox, line 252: loadingObj is not set if `MEDIA_LOADER_FLAGS.ANIMATE_LOAD` flag is off, is it?

src/bit-systems/media-loading.ts
keianhzo: Oh yeah, that's am error. Removing that check.

src/bit-systems/media-loading.ts
keianhzo: For example when a glTF is loaded that has a media component attached, when `resizeAndRecenter` is called at that point there is no actual geometry in the gltf hierarchy as the video plane is added when the video is resolved and so the AABB is not set. As we resize and recenter the media based on that size that causes the media to be positioned at (0,0,0).

This will happen with any gltf that has a component that triggers an async media fetch: model, video, audio, image and link.

keianhzo: A better solution would be to resize and center when all the depending medias have been resolved but we don't have a good way of knowing that so that would be a more complicated solution.

src/bit-systems/media-loading.ts
keianhzo: This affects when media playback starts: https://github.com/mozilla/hubs/blob/95a4deefafc548ab0104b6af32bf5ddb77d96454/src/bit-systems/video-system.ts#L49 but my understanding is that this is a more desirable behavior as we probably want a video playback to start when it's fully visible rather than when it's still scaling.
It also affects when a media-frame can capture an entity https://github.com/mozilla/hubs/blob/95a4deefafc548ab0104b6af32bf5ddb77d96454/src/systems/bit-media-frames.js#L114 and I also think that it makes sense that works this way unless I'm missing something.

src/bit-systems/media-loading.ts
keianhzo: Yes, that's correct `removeEntity` doesn't do anything if the eid is undefined but I'm going to add a check for code clarity.

src/bit-systems/media-loading.ts
takahirox: > For example when a glTF is loaded that has a media component attached, when resizeAndRecenter is called at that point there is no actual geometry in the gltf hierarchy as the video plane is added when the video is resolved

Isn't this `resizeAndRecenter()` called after video (or other media) is resolved?

https://github.com/mozilla/hubs/blob/c5b17010a8aa2e7522ed3506e3c7e2a621d3f344/src/bit-systems/media-loading.ts#L213-L216

keianhzo: Yes but:
- We only recenter the spawned interactable not the media:
Interactable:
https://github.com/mozilla/hubs/blob/c5b17010a8aa2e7522ed3506e3c7e2a621d3f344/src/load-media-on-paste-or-drop.ts#L26-L32
Media (video example):
https://github.com/mozilla/hubs/blob/c5b17010a8aa2e7522ed3506e3c7e2a621d3f344/src/inflators/video-loader.ts#L14-L20
- Even if we recenter the child media, that would only recenter it on the parent not the scene.

src/bit-systems/media-loading.ts
keianhzo: I've added a commit to don't recenter/resize for objects without volume, I think that makes more sense.

takahirox: > I've added a commit to don't recenter/resize for objects without volume, I think that makes more sense.

I agree with that it makes more sense.

src/bit-systems/media-loading.ts
takahirox: Ah, I had misread this part

> This will happen with any gltf that has a component that triggers an async media fetch: model, video, audio, image and link.

Would you mind adding a comment about it? Other people may be confused like I did. Or I would like to add a comment in another PR.

src/bit-systems/media-loading.ts
```diff
@@ -176,13 +178,14 @@ type MediaInfo = {
...
+
 function* loadMedia(world: HubsWorld, eid: EntityID) {
-  let loadingObjEid = 0;
   const addLoadingObjectTimeout = crTimeout(() => {
-    loadingObjEid = renderAsEntity(world, LoadingObject());
+    const loadingObjEid = renderAsEntity(world, LoadingObject());
+    loadingObjs.set(eid, loadingObjEid);
     add(world, loadingObjEid, eid);
   }, 400);
-  yield withRollback(Promise.resolve(), () => loadingObjEid && removeEntity(world, loadingObjEid));
```
takahirox, line 185: @keianhzo @johnshaughnessy Can you explain what this line was for? I don't get and am not sure if removing this line is safe or not.

takahirox:
> Handles the loading entity removal in the next tick to avoid other systems relaying on the entity (mixerAnimatableSystem) crash if the loading cube is inmmediatly removed after adding it.

Can you elaborate it for me please?

src/bit-systems/media-loading.ts
```diff
@@ -216,21 +217,20 @@ function* loadAndAnimateMedia(world: HubsWorld, eid: EntityID, clearRollbacks: C
...
-    // TODO update scale?
-    inflatePhysicsShape(world, eid, {
-      type: hasComponent(world, GLTFModel, media) ? Shape.HULL : Shape.BOX,
-      minHalfExtent: 0.04
-    });
-  }
+  const box = getBox(world, eid, media);
+  addComponent(world, MediaContentBounds, eid);
+  box.getSize(tmpVector);
+  MediaContentBounds.bounds[eid].set(tmpVector.toArray());
```
takahirox, line 224: The behavior is changed here in case media is failed to be loaded. Without this change `MediaContentBounds` is not set (because `MediaLoaded` is not set) whilt with this change it is set. Is this change intentional? (Honestly I'm not so sure about `MediaContentBounds`.

keianhzo:
@takahirox To avoid confusion I've limited the scope of this PR to just not recentering/resizing the interactable in case it has no volume and I will address other issues in separate PRs.

src/bit-systems/media-loading.ts
```diff
@@ -79,8 +79,9 @@ const loaderForMediaType = {
     { accessibleUrl, contentType }: { accessibleUrl: string; contentType: string }
   ) => loadModel(world, accessibleUrl, contentType, true),
   [MediaType.PDF]: (world: HubsWorld, { accessibleUrl }: { accessibleUrl: string }) => loadPDF(world, accessibleUrl),
-  [MediaType.AUDIO]: (world: HubsWorld, { accessibleUrl }: { accessibleUrl: string }) => loadAudio(world, accessibleUrl),
-  [MediaType.HTML]: (world: HubsWorld, { canonicalUrl, thumbnail }: { canonicalUrl: string, thumbnail: string }) =>
+  [MediaType.AUDIO]: (world: HubsWorld, { accessibleUrl }: { accessibleUrl: string }) =>
+    loadAudio(world, accessibleUrl),
+  [MediaType.HTML]: (world: HubsWorld, { canonicalUrl, thumbnail }: { canonicalUrl: string; thumbnail: string }) =>
```
takahirox, line 84: Nit: Just to confirm, this change is for linting, isn't it? Personally I prefer to separate updates for linting/formatting from bug fix/feature addition PRs for easier review.

takahirox (approved):
> To avoid confusion I've limited the scope of this PR to just not recentering/resizing the interactable in case it has no volume and I will address other issues in separate PRs.

That sounds good to me.

src/bit-systems/media-loading.ts
keianhzo: Yes, it's only linting. As it was a really small linter update I thought it was fine to leave it here. I'll keep that in mind for the future.

PR #6120, opened 2023-06-12
feat: Move raised hands to top of guestlist

nikk15:
Why?
- value add for Q&A sessions (eg. education, conferences, etc) - speaker is easily able to view those who have hands raised for questions
Screenshot 2023-06-12 at 3 25 52 PM


nickgrato (approved):
maybe remove the _b if it lets you and then the linter might work? but either way good me!

PR #6121, opened 2023-06-13
Allow attaching components that add Object3Ds to meshes

keianhzo:
The loader crashes when a component that adds a mesh is attached to a mesh node (ie. particle system or water attached to a mesh). This works in the AFrame loader.

I'm not sure why we want to replace the object instead of just adding the new object to it, I seems to me that that would be more aligned with artists wishes as you might want to attach a particle system to an mesh for example but that seems to be how it worked in AFrame too so I focused on parity here.

### DIFF
--- a/src/components/gltf-model-plus.js
+++ b/src/components/gltf-model-plus.js
@@ -477,7 +477,11 @@ class GLTFHubsPlugin {
            * This was supported by the AFrame loader so this extension ensures backwards compatibility with all the existing scenes.
            * For more context about this see: https://github.com/mozilla/hubs/pull/6121
            */
-          if (node.mesh !== undefined || node.camera !== undefined) {
+          if (
+            node.mesh !== undefined ||
+            node.camera !== undefined ||
+            (node.extensions && node.extensions["KHR_lights_punctual"]?.light !== undefined)
+          ) {
             const exts = node.extensions?.MOZ_hubs_components;
             if (exts) {
               const children = [];
--- a/src/components/gltf-model-plus.js
+++ b/src/components/gltf-model-plus.js
@@ -14,6 +14,7 @@ import { promisifyWorker } from "../utils/promisify-worker.js";
 import qsTruthy from "../utils/qs_truthy";
 import { cloneObject3D, disposeMaterial } from "../utils/three-utils";
 import SketchfabZipWorker from "../workers/sketchfab-zip.worker.js";
+import { shouldUseNewLoader } from "../utils/bit-utils";
 
 THREE.Mesh.prototype.raycast = acceleratedRaycast;
 
@@ -469,7 +470,7 @@ class GLTFHubsPlugin {
 
         node.extras.gltfIndex = i;
 
-        if (qsTruthy("newLoader")) {
+        if (shouldUseNewLoader()) {
           /**
            * This guarantees that components that add Object3Ds (ie. through addObject3DComponent) are not attached to non-Object3D
            * entities as it's not supported in the BitECS loader.
--- a/src/components/gltf-model-plus.js
+++ b/src/components/gltf-model-plus.js
@@ -395,6 +395,25 @@ const convertStandardMaterialsIfNeeded = object => {
 let ktxLoader;
 let dracoLoader;
 
+const OBJECT3D_EXT = [
+  "ambient-light",
+  "audio",
+  "directional-light",
+  "hemisphere-light",
+  "image",
+  "link",
+  "model",
+  "particle-emitter",
+  "pdf",
+  "point-light",
+  "reflection-probe",
+  "simple-water",
+  "skybox",
+  "spot-light",
+  "text",
+  "video"
+];
+
 class GLTFHubsPlugin {
   constructor(parser, jsonPreprocessor) {
     this.parser = parser;
@@ -449,6 +468,38 @@ class GLTFHubsPlugin {
         }
 
         node.extras.gltfIndex = i;
+
+        if (qsTruthy("newLoader")) {
+          /**
+           * This guarantees that components that add Object3Ds (ie. through addObject3DComponent) are not attached to non-Object3D
+           * entities as it's not supported in the BitECS loader.
+           * This was supported by the AFrame loader so this extension ensures backwards compatibility with all the existing scenes.
+           * For more context about this see: https://github.com/mozilla/hubs/pull/6121
+           */
+          if (node.mesh !== undefined || node.camera !== undefined) {
+            const exts = node.extensions?.MOZ_hubs_components;
+            if (exts) {
+              const children = [];
+              for (const [key, value] of Object.entries(exts)) {
+                if (OBJECT3D_EXT.includes(key)) {
+                  const newNode = {
+                    name: `${node.name}_${key}`,
+                    extensions: {
+                      MOZ_hubs_components: { [key]: value }
+                    }
+                  };
+                  delete exts[key];
+                  children.push(newNode);
+                }
+              }
+              node.children = node.children || [];
+              children.forEach(child => {
+                const idx = nodes.push(child) - 1;
+                node.children.push(idx);
+              });
+            }
+          }
+        }
       }
     }
   }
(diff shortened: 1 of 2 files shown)


src/inflators/model.tsx
```diff
@@ -88,10 +88,6 @@ export function inflateModel(world: HubsWorld, rootEid: number, { model }: Model
 
     const replacement = world.eid2obj.get(eid);
     if (replacement) {
-      if (obj.type !== "Object3D") {
-        console.error(obj, replacement);
-        throw new Error("Failed to inflate model. Unexpected object type found before swap.");
-      }
```
takahirox, line 94: I first want to know the original purpose of this check...

It looks like indicating that if a Hubs component generates Three.js Object, the glTF node that has the Hubs component shouldn't be Mesh (or other object) node. Do we have such a limitation in Hubs components?

```
// glTF
"nodes": [
  {
    // Valid
    "extensions": {
      "MOZ_hubs_components": {
        // foo generates Three.js Object3D in Hubs
        "foo": ...
      }
    }
  },
  {
    // Invalid
    "mesh": 0,
    "extensions": {
      "MOZ_hubs_components": {
        "foo": ...
      }
    }
  }
]
```

keianhzo: My interpretation of this is that for some reason we don't allow components that create meshes to replace a mesh but I'm not sure why that limitation and more importantly this is not how the old loader works and that can break existing scenes.

This patch won't make us assign multiple objects to a single entity if I'm correct, this just allows replacing meshes if the have a component attached that creates a mesh.

Maybe @johnshaughnessy has more info about why it's important that we don't override meshes (again, my main concern if scene brekeage as this is not how the AFrame loader works)

takahirox: >  just allows replacing meshes

If I understand correctly...

glTF mesh in a node will be ignored if Hubs components that generate Three.js objects are set in the same node, if we allow replace meshes.

```
// glTF
"nodes": [
  {
    "extensions": {
      "MOZ_hubs_components": {
        // foo generates Three.js Object3D in Hubs
        "foo": ...
      }
    },
    // this mesh will be ignored because the mesh object will be
    // replaced with objects generated from `foo` component
    mesh: 0
  },
]
```

Is it an expected behavior? For me, it would be a weird behavior tho.

Perhaps we may need to learn the history from @johnshaughnessy 

If my assumption is correct and this behavior is weird but we want to add a hack for old loader compatilibity, we may need to add a comment describing that this is a hack and should be removed when we get rid of the old compatibility.

johnshaughnessy: This limitation is intentional. We should not remove it.

In the new code, each `entity` in hubs is associated with _at most_ one `Object3D` in the scene graph. This restriction simplifies the code in a number of places. 

Our old code would insert `Group` nodes into the scene graph whenever the components on a node called for multiple `Three.js` objects. These changes to the scene graph added a performance cost and increased complexity: We needed to add code to rename nodes, retarget animations, and copy `morphTargetInfluences`.

We want each component added to a node in a gltf file to unambiguously target an entity when loaded into Hubs. Suppose we allow a node to have a mesh and an audio component and a text component, and then we add a `Foo` component. To which `entity`/`object3D` should we assign the `Foo`? The `mesh`, the `audio`, the `text`, or a `Group` node containing them? We decided that there is no _good_ answer to this question in general, and we should _reject_ these gltf files.

As you have rightly pointed out, though, there are lots of gltf files that already break this rules. We should solve this "at the edge", rather than here "in the middle" of the code. Specifically, we should try to transform an invalid `gltf` into a valid one within `GLTFLoader`.

You can see an example of how we already "patch" gltf files before loading them here: https://github.com/mozilla/hubs/blob/31373a2c660e6a349a5548ff5de31bcdf1c9fce6/src/components/gltf-model-plus.js#L508-L521

I think we should similarly try to introduce whatever changes are necessary to a gltf at this earliest possible stage, and then throw informative warnings to the user to let them know what's wrong. Similarly, our hubs blender exporter should warn on these issues and tell the author how to fix them.

In order to keep the client as simple as possible, we need a simple scene graph. We should not enable gltf authors to create a `Light` that is also an `Audio` that is also a `Mesh`, because we cannot represent such a thing in Three. If they want an entity with three _children_ (a light, an audio, and a mesh), that's OK. But they should explicitly create that structure.

takahirox: Perhaps we may also add validation when exporting (with Hubs Blender add-on or from Spoke).

keianhzo: Thanks for the background on this decision, it makes sense. As you said we can add some warnings on the Blender side (Spoke doesn't have this issue) when adding a "mesh" component and in the migration report, that is something that could help the user to be aware of this limitation and also help them understand which scenes require updates. I'm going to close this PR and add it to the add-on for the next release.

keianhzo:
Closing this as discussed in the comments. I'll open a PR in the Blender add-on to add warnings in the components panels and the migration report to surface this limitation.

keianhzo:
Thinking about this again, wouldn't it be better than crashing that we just log the error to the console and continue rather than crashing. I'm concerned about a lot of scenes that weren't designed with this in mind to not load.

johnshaughnessy:
> Thinking about this again, wouldn't it be better than crashing that we just log the error to the console and continue rather than crashing? 

Throwing an error doesn't mean crashing. When the model fails to load, the calling code decides how to handle it:
-  The media loading system handles it by replacing it with an `ErrorObject()`: 
https://github.com/mozilla/hubs/blob/31373a2c660e6a349a5548ff5de31bcdf1c9fce6/src/bit-systems/media-loading.ts#L196-L199
- The scene loading system handles it by displaying a scene loading error:
https://github.com/mozilla/hubs/blob/31373a2c660e6a349a5548ff5de31bcdf1c9fce6/src/bit-systems/scene-loading.ts#L172-L177

> I'm concerned about a lot of scenes that weren't designed with this in mind to not load.

Yes, your concern is valid. But we should fix the issue by patching the gltf file in a preprocessing step (probably in a `GLTFLoader` plugin), not at this point in the codebase. That way the places where we have to be concerned about these "invalid" gltfs is minimized: It is only at the "boundary" of the application, not "in the middle". 

Another way to think about it is: Our app would be simplest if all gltf files were guaranteed to have the properties we want. We can't assume they are in the correct shape, so we have to do something. We might as well do that transformation / migration _first_, as early as possible in the pipeline.

Eventually if we move to support a specific avatar format like vrm, we'll want to do the same thing: For any gltf file that we are loading as an avatar, check that it is in vrm format. If it's not, _transform it to a vrm_, then proceed as normal. That way all of our legacy avatars continue to work correctly despite the client using vrm related code.

We will need to fix this issue before enabling `newLoader` widely.

keianhzo:
Got it, patching the glTF seems like good option. I've added a loader extension to make sure that we add child nodes for those extensions that add objects if the node is not a Object3D. I have to add an explicit list of the extensions that require this as right now we don't have any other way of getting that info, let me know if you have any better suggestions. 

This should make all scenes compatible with the new loader.

I'll add also the warnings to the Blender add-on UI so users can avoid that in origin.

src/components/gltf-model-plus.js
```diff
@@ -449,6 +468,38 @@ class GLTFHubsPlugin {
         }
 
         node.extras.gltfIndex = i;
+
+        if (qsTruthy("newLoader")) {
```
takahirox, line 472: Just a reminder: We will replace qstruthy newLoader with `shouldUseNewLoader()` in https://github.com/mozilla/hubs/pull/6042

src/components/gltf-model-plus.js
```diff
@@ -449,6 +468,38 @@ class GLTFHubsPlugin {
...
+           * entities as it's not supported in the BitECS loader.
+           * This was supported by the AFrame loader so this extension ensures backwards compatibility with all the existing scenes.
+           * For more context about this see: https://github.com/mozilla/hubs/pull/6121
+           */
+          if (node.mesh !== undefined || node.camera !== undefined) {
+            const exts = node.extensions?.MOZ_hubs_components;
+            if (exts) {
+              const children = [];
+              for (const [key, value] of Object.entries(exts)) {
+                if (OBJECT3D_EXT.includes(key)) {
```
takahirox, line 484: nit: If `OBJECT3D_EXT` is used only for checking whether `key` is in a set, wouldn't using `Set` be more efficient than using array? The number of `OBJECT3D_EXT` entries is just 16, so probably not a big deal tho. Just a comment for minor optimization.

src/components/gltf-model-plus.js
```diff
@@ -449,6 +468,38 @@ class GLTFHubsPlugin {
...
         node.extras.gltfIndex = i;
+
+        if (qsTruthy("newLoader")) {
+          /**
+           * This guarantees that components that add Object3Ds (ie. through addObject3DComponent) are not attached to non-Object3D
+           * entities as it's not supported in the BitECS loader.
+           * This was supported by the AFrame loader so this extension ensures backwards compatibility with all the existing scenes.
+           * For more context about this see: https://github.com/mozilla/hubs/pull/6121
+           */
+          if (node.mesh !== undefined || node.camera !== undefined) {
```
takahirox, line 479: We may need to also check [glTF punctual lights extension](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_lights_punctual/README.md)? Three.js glTF loader built-in plugin creates Three.js `Light` object inheriting `Object3D`.

takahirox (approved):
Added some (nit) comments, but basically the change looks good to me.

PR #6122, opened 2023-06-13
Remove bodies that are disconnected from the scene graph

johnshaughnessy:
We have not diagnosed why, but physics bodies sometimes stick around after their associated object3D has been removed from the scene graph. This PR removes the physics body when we detect this case.

While a more correct fix would be to prevent this from happening, we have not found a reliable reproduction step. Since the error spams the log and the body is never removed, this PR at least fixes those things.


nikk15 (approved):
Code LGTM!

PR #6138, opened 2023-06-21
fix: displayName regex

nikk15:
Solves this console error:
`Pattern attribute value ^[A-Za-z0-9_~ -]{3,32}$ is not a valid regular expression: Uncaught SyntaxError: Invalid regular expression: /^[A-Za-z0-9_~ -]{3,32}$/v: Invalid character in character class`

### DIFF
--- a/src/storage/store.js
+++ b/src/storage/store.js
@@ -63,9 +63,9 @@ export const SCHEMA = {
       type: "object",
       additionalProperties: false,
       properties: {
-        displayName: { type: "string", pattern: "^(?:[A-Z]|[a-z]|[0-9]|[_~\\-]|\\s){3,32}$" },
+        displayName: { type: "string", pattern: "^[A-Za-z0-9_~\\s\\-]{3,32}$" },
         avatarId: { type: "string" },
-        pronouns: { type: "string", pattern: "^[a-zA-Z///a-zA-Z]{2,32}$" },
+        pronouns: { type: "string", pattern: "^([a-zA-Z]{1,32}\\/){0,4}[a-zA-Z]{1,32}$" },
         // personalAvatarId is obsolete, but we need it here for backwards compatibility.
         personalAvatarId: { type: "string" }
       }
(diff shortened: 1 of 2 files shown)


src/storage/store.js
```diff
@@ -63,7 +63,7 @@ export const SCHEMA = {
       type: "object",
       additionalProperties: false,
       properties: {
-        displayName: { type: "string", pattern: "^[A-Za-z0-9_~ -]{3,32}$" },
+        displayName: { type: "string", pattern: "^(?:[A-Z]|[a-z]|[0-9]|[_~\\-]|\\s){3,32}$" },
         avatarId: { type: "string" },
         pronouns: { type: "string", pattern: "^[a-zA-Z///a-zA-Z]{2,32}$" },
```
keianhzo, line 68: This seems to still make Chrome complain. After some testing this seems not to raise any complaints in both Chrome & Firefox:
````
displayName: { type: "string", pattern: "^[A-Za-z0-9_~\\s\\-]{3,32}$" },
avatarId: { type: "string" },
pronouns: { type: "string", pattern: "^([a-zA-Z]{1,32}\\/){0,4}[a-zA-Z]{1,32}$" },
```
This pronouns regex is a bit more strict and enforces the A/B form and you can have up to 5 pronouns that I think it's probably enough for what I've seen?

keianhzo: If we prefer to keep the previous regex format, this would work:
```
^[a-zA-Z\\/a-zA-Z]{2,32}$
```

keianhzo (changes requested):
It seems that browsers are a bit picky and each one them have their on wierdness so I've tried my best to find the right solution. I've left a comment about my findings.

src/storage/store.js
nikk15: Interesting! Thanks for looking into alternatives. I'll update the pronouns with your suggestion - its currently had a few incarnations (enforcing one '/', and leaving it out as an option) but I like your suggestion that allows for up to 5 pronouns. 🚀

keianhzo (approved):
LGTM!  🚀

PR #6146, opened 2023-06-27
Check permissions for CreateMessages

johnshaughnessy:
When clients receive a `CreateMessage`, they should check whether the sender has the appropriate `Permission` to be spawning from that prefab.

Since all prefabs currently need a permission, and we'd like to try not to forget them, I've made the `permission` field of `PrefabDefinition` non-optional.


keianhzo (approved):
LGTM

PR #6150, opened 2023-06-29
Catch entity state push promise exceptions

keianhzo:


keianhzo:
I just realized that this has already been addressed here so closing this one

PR #6153, opened 2023-07-05
Move loop-animation component to each object instead of scene

keianhzo:
When two animation clips with the same name are targeted to different objects we are playing always the first one and the second one is never played.

This used to work fine in AFrame because we used to re-targe animations based on the object where `loop-component` was attached but with the bitECS migration we ignore the objects components and we attach one `loop-animation` component to the scene root so we can't re-target anymore as we have lost the reference of which objects play animations.

**Note**: This works fine in most cases but it fails when several objects are using the same animation name which seems to be pretty common (as I've seen this in different assets lately)

This PR addresses this issue keeping the`loop-animation` component attached to the original object so the animations can be correctly re-targeted.

As it's widely commented in the code, it makes sense to have only one `loop-animation` at the scene level that handles all the animations in the scene but that would require quite some refactoring to move the component to the scene level. What we could easily do is updating the Blender component to use the glTF animation index instead of the clip name as that one it's unique so we can at least do the component transform that we are doing and each clip would be correctly played in its related object. I'll open an issue to update the Blender component and start using that instead of the clip name.

Use this model for testing.
[Frustum.glb.zip](https://github.com/mozilla/hubs/files/11958098/Frustum.glb.zip)
[Frustum.blend.zip](https://github.com/mozilla/hubs/files/11958102/Frustum.blend.zip)

### DIFF
--- a/src/bit-components.js
+++ b/src/bit-components.js
@@ -201,7 +201,7 @@ export const LoopAnimationInitialize = defineComponent({});
  *          paused: boolean,
  *          startOffset: number,
  *          timeScale: number
- *        }[]>}
+ *        }>}
  */
 export const LoopAnimationInitializeData = new Map();
 export const LoopAnimation = defineComponent();
--- a/src/bit-systems/loop-animation.ts
+++ b/src/bit-systems/loop-animation.ts
@@ -1,5 +1,5 @@
-import { addComponent, defineQuery, enterQuery, exitQuery, hasComponent, removeComponent } from "bitecs";
-import { AnimationAction, AnimationClip, AnimationMixer, LoopRepeat } from "three";
+import { addComponent, defineQuery, enterQuery, exitQuery, removeComponent } from "bitecs";
+import { AnimationClip, LoopRepeat } from "three";
 import {
   MixerAnimatable,
   MixerAnimatableData,
@@ -10,11 +10,12 @@ import {
   Object3DTag
 } from "../bit-components";
 import { HubsWorld } from "../app";
+import { findAncestorWithComponent } from "../utils/bit-utils";
 
-const loopAnimationInitializeQuery = defineQuery([LoopAnimationInitialize, MixerAnimatable, Object3DTag]);
+const loopAnimationInitializeQuery = defineQuery([LoopAnimationInitialize, Object3DTag]);
 const loopAnimationInitializeEnterQuery = enterQuery(loopAnimationInitializeQuery);
 
-const loopAnimationQuery = defineQuery([LoopAnimation, MixerAnimatable, Object3DTag]);
+const loopAnimationQuery = defineQuery([LoopAnimation, Object3DTag]);
 const loopAnimationExitQuery = exitQuery(loopAnimationQuery);
 
 // Question: Who should have AnimationMixer?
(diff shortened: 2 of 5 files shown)


takahirox:
Thinking whether we can convert from object-level `loop-animation` to scene-level `loop-animation` on the fly before parsing glTF content like #6121. As I wrote somewhere, I think scene-level is easier implement, more intuitive, and better maintainability and want to update `loop-animation` spec and Blender addon.

Honestly I'm very biased tho, I prefer scene-level `loop-animation` much.

Exairnous:
Will a scene-level loop-animation handle objects dropped in-room correctly?

keianhzo:
> Will a scene-level loop-animation handle objects dropped in-room correctly?

Every dropped glTF is a scene and it will has it's own mixer to manage all the scene animations so  it should work correctly.

> Thinking whether we can convert from object-level loop-animation to scene-level loop-animation on the fly before parsing > glTF content like https://github.com/mozilla/hubs/pull/6121, or adding hack here.

This will effectively be the same as you were doing just done at an earlier stage. I agree on that having a scene level loop-animation makes more sense but that would require some component redesigning to add some extra info. Now you can set different animation parameters like offset and timeScale so the scene component would need support to individually set those per animation.

Independently from where we place the component (scene or object) we should move the Blender component from using the animation name to using the index as now we have a problem with animations that use the same name as when we look for the animation among all the scene animations, we always pick same one as there is no way to tell which animation is the one targeted to a specific object so we end up re-targetting the first one to all the objects using that name.

keianhzo:
If you feel this PR approach is correct for now we can focus on landing this and look into updating the architecture later to now delay the newLoader release anymore.

takahirox:
I first want to know the problem correctly

> When two animation clips with the same name are targeted to different objects

Like this?

```
// glTF content
{
  "animations": [
    // Animation for node 1
    {
      "name": "foo_animation",
      "channels": [
        {
          "target": {
            "node": 1,
            ...
          },
          ...
        }
      ]
      ...
    },
    // Animation for node 2
    {
      "name": "foo_animation",
      "channels": [
        {
          "target": {
            "node": 2,
            ...
          },
          ...
        }
      ]
      ...
    }
  ],
  "nodes": [
    {
      "children": [1, 2]
    },
    {
      "mesh": 0,
      "extensions": {
        "MOZ_hubs_components": {
          "loop-animation": {
            "clip": "foo_animation",
            ...
          }
        }
      }
    },
    {
      "mesh": 1,
      "extensions": {
        "MOZ_hubs_components": {
          "loop-animation": {
            "clip": "foo_animation",
            ...
          }
        }
      }
    }
  ],
  ...
}
```

Exairnous:
@takahirox It depends.  That's one way, another way has it combined together into one animation with multiple channels/samplers.  Exporting multiple objects with the same action seems to result in what you have above, while exporting multiple tracks with the same name (for multiple objects) seems to result in the multiple channels/samplers.  Here's a text based glTF file exported from Blender 3.6 to showcase this:
[blender3.6-gltf-animation-export-test.zip](https://github.com/mozilla/hubs/files/11974258/blender3.6-gltf-animation-export-test.zip)

keianhzo:
@takahirox When different objects have different animations with the same name in Blender they are all exported as different animations with same name in the `loop-animation` component `clip`property. (So the glTF structure that you mentioned in your last comment).

Then we look for the animation that matches the clip name here:
https://github.com/mozilla/hubs/blob/31373a2c660e6a349a5548ff5de31bcdf1c9fce6/src/bit-systems/loop-animation.ts#L33-L35

And we create the animation clips here: 
https://github.com/mozilla/hubs/blob/31373a2c660e6a349a5548ff5de31bcdf1c9fce6/src/bit-systems/loop-animation.ts#L74-L80

As they all have the same name we always get the first `AnimationClip` from `getActiveClips` for all the animations. As the clip action is targeted to the scene object, we are effectively getting the same clip for each animation with the same name. As stated in the ThreeJS docs:
> If an action fitting the clip and root parameters doesn't yet exist, it will be created by this method. Calling this method several times with the same clip and root parameters always returns the same clip instance.
So we are just running the the first clip.

With this PR we move the `loop-animation` component to the object level so when we call `clipAction` we retarget the animation with the same name (which tends to be the same animation) to all the different objects. This is what we are doing in AFrame.

As I mentioned before this behavior is NOT what we want, we don't want to re-target animations we want to run the right clip on the scene level mixer.

Giving this another thought another maybe simpler solution would be to add a pre-process stage to the GLTF loader and replace the "clip" field in the `loop-animation` component with the `activeClipIndices` field  pointing to the animation indexes. We should only do that if the `clip`property exists as otherwise (Spoke exports for example) it would already being using indexes and that  should be fine.

Let me know what you think and maybe we can take that approach better.

takahirox:
Some comments

1.

> As they all have the same name we always get the first AnimationClip from getActiveClips for all the animations.

Even if we move `loop-animation` to node (object) level, still the first clip is chosen, so, for example in https://github.com/mozilla/hubs/pull/6153#issuecomment-1624256984 `animations[1]` is for `nodes[2]` but even with node level `loop-animation` `animations[0]` is wrongly chosen, isn't it? Node level `loop-animation` doesn't resolve this wrong chosen clip problem, does it?

2.

> With this PR we move the `loop-animation` component to the object level so when we call `clipAction` we retarget the animation with the same name (which tends to be the same animation) to all the different objects.

Honestly I still don't get why node level `loop-animation` allows animation retarget... In glTF animation target is specified as node and in Three.js animation target is specified as object uuid (more precisely in Three.js animation target can be specified as object name or uuid, but our gltf loader plugin forces to use uuid). In the Three.js `clipAction` target object is found with uuid. So, if we set optional root node to `clipAction` target object shouldn't be found (Assuming the original animation target object is not placed under the optional root node). What hack is used for animation retarget with node level `loop-animation`?

takahirox:
I guess perhaps we had assumption that animations that have the same name have the same animation data and animation target in a clip is only root node having `loop-animation` component. I'm not sure if it's a safe assumption but if we want to have backward compatilibity (for now) we may clone `gltf.animation` for each `loop-animation` node in the glTF preprocessing like

from

```
// glTF content
{
  "animations": [
    {
      "name": "foo_animation",
      "channels": [
        {
          "target": {
            "node": 0,
            ...
          },
          ...
        }
      ]
      ...
    },
    ...
  ],
  "nodes": [
    {
      "mesh": 0,
      "extensions": {
        "MOZ_hubs_components": {
          "loop-animation": {
            "clip": "foo_animation",
            ...
          }
        }
      }
    },
    {
      "mesh": 1,
      "extensions": {
        "MOZ_hubs_components": {
          "loop-animation": {
            "clip": "foo_animation",
            ...
          }
        }
      }
    }
  ],
  ...
}
```

to

```
// glTF content
{
  "animations": [
    {
      "name": "foo_animation",
      "channels": [
        {
          "target": {
            "node": 0,
            ...
          },
          ...
        }
      ]
      ...
    },
    {
       // Clone and rename
      "name": "foo_animation2",
      "channels": [
        {
          "target": {
            // Retarget to nodes[1]
            "node": 1,
            ...
          },
          ...
        }
      ]
      ...
    },
    ...
  ],
  "nodes": [
    {
      "mesh": 0,
      "extensions": {
        "MOZ_hubs_components": {
          "loop-animation": {
            "clip": "foo_animation",
            ...
          }
        }
      }
    },
    {
      "mesh": 1,
      "extensions": {
        "MOZ_hubs_components": {
          "loop-animation": {
            // Rename
            "clip": "foo_animation2",
            ...
          }
        }
      }
    }
  ],
  ...
}
```

The same animation data content are referred from original and cloned `gltf.animations` so probably additional memory cost is not a big deal. I would make a PR for this approach.

takahirox:
Made a PR #6183

keianhzo:
Closing in favor of https://github.com/mozilla/hubs/pull/6183

PR #6155, opened 2023-07-06
Correctly handle scene physics in BitECS

keianhzo:
Scene physics were not being correctly handled in bitecs. We were looking for an existing physics shape assuming that that would be a heightfield or trimesh (Spoke) or otherwise we would create a default physics floor based on a heuristic.

This fails anytime the scene contains a physics shape already as all components have already been inflated at this point so I added specific tag components for trimesh and heightfields so we can correctly handle that case which is a common case for Spoke generated scenes. For Blender scenes (as there is no trimesh/heightfield automatic generation there) we will use the heurstic.

This correctly mimics our previous AFrame behaviour and ensure retro compat.

### DIFF
--- a/src/bit-components.js
+++ b/src/bit-components.js
@@ -346,3 +346,5 @@ export const LinearScale = defineComponent({
   targetZ: Types.f32
 });
 export const Quack = defineComponent();
+export const TrimeshTag = defineComponent();
+export const HeightFieldTag = defineComponent();
--- a/src/bit-systems/scene-loading.ts
+++ b/src/bit-systems/scene-loading.ts
@@ -4,13 +4,15 @@ import { Mesh } from "three";
 import { HubsWorld } from "../app";
 import {
   EnvironmentSettings,
+  HeightFieldTag,
   NavMesh,
   Networked,
   PhysicsShape,
   SceneLoader,
   ScenePreviewCamera,
   SceneRoot,
-  Skybox
+  Skybox,
+  TrimeshTag
 } from "../bit-components";
 import Sky from "../components/skybox";
 import { Fit, inflatePhysicsShape, Shape } from "../inflators/physics-shape";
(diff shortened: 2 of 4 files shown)


takahirox:
I'm not familiar with physics and scene physics so let me confirm my understanding first

> This fails anytime the scene contains a physics shape already

It means scene physics initialization fails if non-root object in the scene already has physics shape even for not scehe physics (for example a child object has physics shape for the its rigid body), doesn't it?

If so, yes, checking any physics shape existence in the scene is weird. Explicitly checking scene physics shape should be more straightforward and less problematic as this PR does (assuming heightfield or trimesh are scene physics, is it true?).

keianhzo:
> It means scene physics initialization fails if any object in the scene already has physics shape even for not scehe physics (for example a child object has physics shape for its rigid body), doesn't it?

> If so, yes, checking any physics shape existence in the scene is weird. Explicitly checking scene physics shape should be more straightforward and less problematic as this PR does (assuming heightfield or trimesh are used only for scene physics, is it true?).

Yes, that's correct. heightfield and trimesh are only used for scene physics but some components (media-frames or ammo-shapes) can trigger physic shapes creation so looking for only `PhysicsShape` after the glTF has been inflated doesn't work, that's why I added the specific `Heightfield` and `Trimesh` tags.

> Offtopic for future clean up: Random thoughts, Ideally is Scene physics component set to scene entity, not child entities, for simplicity?

That's kind of how it is now for the heightfield and trimesh but as I mentioned during the inflation process new physics shapes can exist in the scene so we can't rely on that.

takahirox (approved):
The change looks good and like actually fixing a bug to me

takahirox:
>> Offtopic for future clean up: Random thoughts, Ideally is Scene physics component set to scene entity, not child entities, for simplicity?
>
>That's kind of how it is now for the heightfield and trimesh but as I mentioned during the inflation process new physics shapes can exist in the scene so we can't rely on that.

I thought adding Scene physics component (Trimesh or HeightField) to scene entity may be more natural and straightforward. If we can ensure it `findChildWithComponent(world, TrimeshTag, sceneEid)` can be replaced with `hasComponent(world, TrimeshTag, sceneEid)`. It may be more maintainable and robust because of less dependency with other entities from an entity. Honestly I'm not familiar with our Physics, sorry if I said something weird or impossible. Just my random thought for future clean up.

keianhzo:
heightfield and trimesh (same as nav-mesh) are components that are right now attached to a 3D object so we can't do this at this moment but I agree that this would be the optimal case as we only allow on per scene.

PR #6158, opened 2023-07-10
Fix spawner issues

keianhzo:
Fixes spawner issues

### DIFF
--- a/src/bit-systems/object-spawner.ts
+++ b/src/bit-systems/object-spawner.ts
@@ -2,12 +2,10 @@ import { addComponent, defineQuery, enterQuery, exitQuery } from "bitecs";
 import { HubsWorld } from "../app";
 import { FloatyObject, Held, HeldRemoteRight, Interacted, ObjectSpawner } from "../bit-components";
 import { FLOATY_OBJECT_FLAGS } from "../systems/floaty-object-system";
-import { sleep } from "../utils/async-utils";
 import { coroutine } from "../utils/coroutine";
 import { createNetworkedEntity } from "../utils/create-networked-entity";
 import { EntityID } from "../utils/networking-types";
 import { setMatrixWorld } from "../utils/three-utils";
-import { animateScale, waitForMediaLoaded } from "./media-loading";
 
 export enum OBJECT_SPAWNER_FLAGS {
   /** Apply gravity to spawned objects */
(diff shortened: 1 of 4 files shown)


takahirox:
Thanks for working on it. Would you mind if writing the details in the PR comment for reviewers and future readers to follow the change and background?

takahirox:
I confirmed that this change resolves the spanwer problem for now

src/bit-systems/object-spawner.ts
```diff
@@ -34,14 +32,6 @@ function* spawnObjectJob(world: HubsWorld, spawner: EntityID) {
...
   const spawnedObj = world.eid2obj.get(spawned)!;
   setMatrixWorld(spawnedObj, spawnerObj.matrixWorld);
-
-  yield* waitForMediaLoaded(world, spawned);
-  spawnerObj.visible = false;
-  yield sleep(1000);
-  spawnerObj.visible = true;
-
-  // TODO we should come up with a nicer way to get at the media that was loaded by a MediaLoader
-  yield* animateScale(world, spawnerObj.children[0]!.eid!);
```
takahirox, line 44: Question for my study: Here can be removed because of `animateLoad: true`?

src/inflators/spawner.ts
```diff
@@ -32,13 +33,15 @@ export function inflateSpawner(world: HubsWorld, eid: number, props: SpawnerPara
...
   addComponent(world, SingleActionButton, eid);
+  addComponent(world, TogglesHoveredActionSet, eid);
 
   addComponent(world, ObjectSpawner, eid);
   ObjectSpawner.src[eid] = APP.getSid(props.src);
   ObjectSpawner.flags[eid] = props.mediaOptions?.applyGravity ? OBJECT_SPAWNER_FLAGS.APPLY_GRAVITY : 0;
 
   inflateRigidBody(world, eid, {
     mass: 0,
+    type: Type.STATIC,
```
takahirox, line 44: Question for my study: Spawner object should be static (no local matrix update) so rigid body type should be static. Is my understanding correct?

src/bit-systems/object-spawner.ts
keianhzo: Yes, as we are already animating the spawned object I don't see the point on animating explicitly.

src/inflators/spawner.ts
keianhzo: That's my understanding. Spawners are static as their main action is to spawn an object and are not draggable. Also that's how we were setting them up in AFrame:
https://github.com/mozilla/hubs/blob/5f6e7a62ab998a60895a30dc686ae6db94fe9c5b/src/gltf-component-mappings.js#L323

PR #6168, opened 2023-07-17
Update text hubs wide to use correct color variables

nikk15:
As part of our preparation to improve theme customization, @katetaylormoz and I are sweeping through components/sections one at a time to ensure they are the correct colors and using the variable names we will be using in the theme customization tool.


nickgrato (approved):
lgtm , it would be cool to get CSS modules in here one day so we don't have to use the :local()

PR #6169, opened 2023-07-18
Fix canPin predicate

johnshaughnessy:
The `canPin` predicate assumes `initialData` exists for a given `createMessageData`. It needs to first check whether the entity is a `media`, since `media` are the only things that are pinnable. 

Two unrelated changes:
- Catch an issue in `mixerAnimatableSystem` that can cause the app the become unresponsive.
- Add `createNetworkedMedia` helper function so that `initialData` is type checked for media.


src/bit-systems/mixer-animatable.ts
```diff
@@ -10,6 +10,10 @@ const mixerExitQuery = exitQuery(mixerQuery);
 
 export function mixerAnimatableSystem(world: HubsWorld): void {
   initializeEnterQuery(world).forEach(eid => {
+    if (entityExists(world, eid)) {
```
takahirox, line 13: This line breaks mixer animation. I think here should be `!entityExists(world, eid)`

PR #6176, opened 2023-07-20
Remove entity_state_api feature flag

johnshaughnessy:
Now that reticulum's entity state apis have stabilized, we can remove this feature flag. Note that the functionality is still behind the `newLoader` flag.


takahirox (approved):
Looking good, thanks

PR #6177, opened 2023-07-20
fix: Toolbar buttons should all be transparent on click

nikk15:
Currently all buttons except for "Leave" and "Chat" turn transparent upon opening. This PR rectifies this so that the toolbar buttons all behave uniformly.
Screenshot 2023-07-20 at 2 26 35 PM
Screenshot 2023-07-20 at 2 26 23 PM


takahirox:
I'm node a designer so can't review but consistency sounds good to me unless any historical or technical reason.

nikk15:
> I'm node a designer so can't review but consistency sounds good to me unless any historical or technical reason.

I double checked with @katetaylormoz and its indeed a bug not a feature so we have the green light :)

PR #6178, opened 2023-07-20
fix: Stop custom scene button from flashing

nikk15:
Old: 

https://github.com/mozilla/hubs/assets/42850541/c3e0fc10-e18c-4b69-9060-ddd3139f040c

New: 

https://github.com/mozilla/hubs/assets/42850541/3703dec9-6bb4-44ee-bd59-e9481163c61c


takahirox:
I'm not familiar with react so can't determine whether the change in the codebase is ok, but "after" looks much better, nice work.

PR #6182, opened 2023-07-24
fix: Condense color variables

nikk15:
- remove and replace color variables either used only once or that utilise another color variable


nickgrato (approved):
not seeing anything wrong!

PR #6183, opened 2023-07-24
Fix bitECS loop-animation

takahirox:
**Problem**

`loop-animation` with bitECS may not work correctly if multiple `loop-animation` components use the same animation name to refer to `glTF.animation(s)`.

**Root issue**

Mozilla Hubs `loop-animation` component has a problem in the spec. A `loop-animation` component can refer to `glTF.animations` with animation names but the glTF specification allows non-unique names in `glTF.animations` so if there are multiple `glTF.animations` that have the same name no one can know what `glTF.animations` a `loop-animation` component using that names refers to.

Refer to #6153 for details.

**Solution**

Preprocesses glTF content before parsing to avoid the problem by adding a glTF loader plugin that converts name based animation reference in `loop-animation` component to index based.

**Note**

The old `loop-animation` component handler (without bitECS) seems to assume that multiple `glTF.animations` that have the same name have the same animation data and loop-animation refers to the first `glTF.animation` of the ones having the same name. The plugin follows the assumption for the compatibility, not sure if it is the best approach.

With this commit we may remove name based animation reference processing codes from the `loop-animation` handlers but until we are confident for the change we may keep them.

**Change**

- Add a glTF loader plugin to convert name based animation reference in `loop-animation` component to index based.

Another approach: #6153

### DIFF
--- a/src/components/gltf-model-plus.js
+++ b/src/components/gltf-model-plus.js
@@ -670,6 +670,122 @@ class GLTFMozTextureRGBE {
   }
 }
 
+// Mozilla Hubs loop-animation component has a problem in the spec.
+// The loop-animation component can refer to glTF.animations with
+// animation names but the glTF specification allows non-unique names
+// in glTF.animations so if there are multiple glTF.animations that
+// have the same name no one can know what glTF.animations a
+// loop-component using that names refers to.
+// This plugin converts animation names in the component to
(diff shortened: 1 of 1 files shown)


keianhzo (approved):
Yeah, I think this approach is better. Ideally we should migrate this in the Blender Add-on so we hit this as little as possible.

PR #6189, opened 2023-07-27
Fix legacy nametag color customisation

nikk15:
Previously, the Nametag inputs in the Theme tab in the Admin Panel did not do anything. This fix ensures this legacy method of changing nametags works. Alternatively, "nametag-color" and "nametag-text-color" can be added to the theme variables and will also update nametags.

Additionally, we now disable the save button if there is a warning message currently issued for an invalid input.

Screenshot 2023-07-27 at 1 15 22 PM
Screenshot 2023-07-27 at 1 15 16 PM
Screenshot 2023-07-27 at 1 28 25 PM

### DIFF
--- a/src/utils/theme.js
+++ b/src/utils/theme.js
@@ -105,9 +105,8 @@ function getCurrentTheme() {
 
 function getThemeColor(name) {
   const theme = getCurrentTheme();
-  console.log(theme)
-  // theme?.[name] ensures legacy variables for nametag colors are taken into account
-  return theme?.variables?.[name] || theme?.[name] || DEFAULT_COLORS[name];
+  // config?.theme?.[name] ensures legacy variables for nametag colors are taken into account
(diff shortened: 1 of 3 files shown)


johnshaughnessy (approved):

src/utils/theme.js
```diff
@@ -105,7 +105,8 @@ function getCurrentTheme() {
 
 function getThemeColor(name) {
   const theme = getCurrentTheme();
-  return theme?.variables?.[name] || DEFAULT_COLORS[name];
+  // config?.theme?.[name] ensures legacy variables for nametag colors are taken into account
```
nickgrato, line 108: do you want this commented out code?

nickgrato (commented):
remove the commented code? or leave a message on why you want to keep it? otherwise looks good.

PR #6204, opened 2023-08-09
fix: for broken scene and avatar media links with newLoader

nikk15:
Previously, we were not hitting the block of code that fetches the correct url for these assets, and placing scenes or avatars with the newLoader resulted in broken links.

For now, you may test here: [https://nikk15.dev.myhubs.net/k95eyHn/elegant-grounded-convention/?newLoader=](https://nikk15.dev.myhubs.net/k95eyHn/elegant-grounded-convention/?newLoader=)

Previous:
Screenshot_2023-08-09_at_16 18 25
Screenshot_2023-08-09_at_16 20 10

### DIFF
--- a/src/utils/load-image.tsx
+++ b/src/utils/load-image.tsx
@@ -11,7 +11,6 @@ import { AlphaMode } from "./create-image-mesh";
 type AlphaModeType = typeof AlphaMode.Opaque | typeof AlphaMode.Mask | typeof AlphaMode.Blend;
 
 export function* createImageDef(world: HubsWorld, url: string, contentType: string) {
-  console.log("*** createImageDef", url);
   const { texture, ratio, cacheKey }: { texture: Texture; ratio: number; cacheKey: string } =
     yield loadTextureCancellable(url, 1, contentType);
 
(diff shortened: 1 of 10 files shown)


src/utils/load-image.tsx
```diff
@@ -36,11 +36,5 @@ export function* createImageDef(world: HubsWorld, url: string, contentType: stri
...
   const imageDef = yield* createImageDef(world, url, contentType);
 
-  return renderAsEntity(
-    world,
-    <entity
-      name="Image"
-      image={imageDef}
-    />
-  );
+  return renderAsEntity(world, <entity name="Image" image={imageDef} />);
```
takahirox, line 39: Quick question to confirm: Is this change just for formatting?

takahirox:
Thanks for working on it!

If possible, would you mind explaining how the problem happened (what is the root cause?) and how the problem is fixed with the change for easier review?

src/utils/media-utils.js
```diff
@@ -615,7 +621,7 @@ export async function resolveMediaInfo(urlString) {
 
   // We want to resolve and proxy some hubs urls, like rooms and scene links,
   // but want to avoid proxying assets in order for this to work in dev environments
-  const isLocalAsset = isNonCorsProxyDomain(url.hostname);
+  const isLocalAsset = isNonCorsProxyDomain(url.hostname) && !(await isHubsDestinationUrl) && !(await isHubsAvatarUrl);
```
takahirox, line 624: I think `isHubsDestinationUrl` and `isHubsAvatarUrl` are functions. Is `!(await isHubsDestinationUrl) && !(await isHubsAvatarUrl);` intentional? Or you really wanted `!(await isHubsDestinationUrl(url.href)) && !(await isHubsAvatarUrl(url.href));`?

PR #6205, opened 2023-08-11
Revert "Use non-canonical URL for local non-model assets"

nikk15:
Reverts mozilla/hubs#6082

Problem: Local media links are broken with the `newLoader` flag set to true. All local media assets have `undefined` thumbnail urls. We currently check to see if the asset `isLocalAsset`, and if so we do not `resolveUrl` (fetch thumbnails etc).

Solution: Reverting this PR ensures we are only checking if an asset is a local `model/gltf` type, and for all others we proceed to `resolveUrl`. I have successfully tested this in my local environment when placing scenes, avatars, images, and models, in the room

Notes: Interestingly, this original PR was merged to fix broken links when the 'newLoader' flag is set. This appears to be fixed, but will potentially need investigation.

### DIFF
--- a/src/utils/media-url-utils.js
+++ b/src/utils/media-url-utils.js
@@ -71,7 +71,7 @@ export const scaledThumbnailUrlFor = (url, width, height) => {
 };
 
 export const isNonCorsProxyDomain = hostname => {
-  return !!nonCorsProxyDomains.find(domain => hostname.endsWith(domain));
+  return nonCorsProxyDomains.find(domain => hostname.endsWith(domain));
 };
 
 export const proxiedUrlFor = url => {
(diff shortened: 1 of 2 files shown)


takahirox:
Thanks for working on it.

> Notes: Interestingly, this original PR was merged to fix broken links when the 'newLoader' flag is set. This appears to be fixed, but will potentially need investigation.

Yes, this is very weird to me... Honestly I'm not familiar with our url stuff so can't say whether reverting is a right solution. But if the problem will be actually resolved, we may revert it. And if we face any problem, we may reland.

We should investigate more at some point to figure out and fix the root issue. @johnshaughnessy may know more about it.

Issue #6209, opened 2023-08-14
Couldn't upload draco or ktx2 glb as a custom avatar.


kira924age:
**Description**

GLB files using Draco or KTX2 could not be uploaded as custom avatars.

**To Reproduce**
1. Go to a room
2. Click on 'Join room'
3. Click on 'Change Avatar'
4. Click on 'Create Avatar'
5. Click on 'Custom GLB'
6. Select draco/ktx2 GLB file
7. Click on 'Save'

**Expected behavior**

Uploading can be done without any errors.

**Screenshots**

[Screencast from 2023-08-14 11-23-19.webm](https://github.com/mozilla/hubs/assets/5345124/dadfc564-30a5-48a6-b8ea-018c88c08220)

**Hardware** 
 - Device: Desktop
 - OS: Linux (Arch)
 - Browser: Firefox

**Additional context**

This problem is probably caused by the Loader not being set in the following code.

https://github.com/mozilla/hubs/blob/master/src/react-components/avatar-editor.js#L153-L169

PR #6210, opened 2023-08-14
add KTX2/Draco loader to avatar-editor

kira924age:
✅ Closes: #6209

### DIFF
--- a/src/react-components/avatar-editor.js
+++ b/src/react-components/avatar-editor.js
@@ -15,7 +15,10 @@ import { ensureAvatarMaterial } from "../utils/avatar-utils";
 import AvatarPreview from "./avatar-preview";
 import styles from "../assets/stylesheets/avatar-editor.scss";
 
+import { DRACOLoader } from "three/examples/jsm/loaders/DRACOLoader";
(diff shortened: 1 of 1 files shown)


Exairnous:
It's great to see Draco support for avatars being added, but I thought KTX2 support was already present.  I'm pretty sure I have created and uploaded an avatar with KTX2 texture compression to Hubs.

kira924age:
@Exairnous

Thank you for your comment. As you pointed out, I was able to upload the avatar using the KTX2 Texture even after removing the "setKTX2Loader" process. In this scenario, it seems that there is no need to call setKTX2Loader, so I have removed it. Perhaps in this case, only the uploading is being performed, and the texture loading might not be taking place.

Exairnous:
Perhaps, but I do see textures in the preview window with a KTX2 avatar, so not sure.  Glad I could help anyway.

keianhzo:
KTX2 is used when rendering the model in the scene and in the avatar preview viewport (both via `gltf-model-plus.js `loadGLTF`) but not during the `uploadAvatar` step in the editor. The loader seems to be crashing only for the Draco extension and not for the Basis one, not sure why but I don't think we need to add the KTX2 loader as that extension is available when loading for rendering later.

Issue #6206, opened 2023-08-11
bitECS: Fail to switch scene


takahirox:
**Description**

With new loader enabled switching scene fails. I keep seeing dark.

**To Reproduce**
1. Enter a room with new loader enabled
2. Change the scene
3. Screen is blacked out
4. New scene won't show. You keep seeing dark

**Expected behavior**

Scene switch works even with new loader enabled

**Screenshots**

Console error

![Screenshot_2023-08-11_at_1 41 00_PM](https://github.com/mozilla/hubs/assets/7637832/85254b65-2bf9-4ead-8abf-05da99a73a08)

**Hardware** 
 - Device: Desktop
 - OS: Windows
 - Browser: Chrome

**Additional context**

- Audio (Voice) kept working

takahirox:
I found the problem seems to be scene specific. I can reproduce the problem by switching from the Outdoor Festival scene but couldn't reproduce by switching from other scenes so far. I guess a certain component included in the Outdoor Festival scene causes the problem?

I will remove P1 label for now because it doesn't seem to be a generic scene switching problem.

takahirox:
Fixed by #6217

PR #6217, opened 2023-08-15
bitECS: Prevent Three.js Audio.disconnect error

takahirox:
**Problem**

Disconnecting Three.js `Audio` can cause an error

https://github.com/mozilla/hubs/issues/6206

`Three.js Audio.disconnect()` seems to expect `Audio.source` is `non-null`. `Audio.source` is set via `.set*Source()` or `.play()`.

https://github.com/MozillaReality/three.js/blob/hubs-patches-141/src/audio/Audio.js

Hubs Client seems to call `Audio.disconnect()` before setting source in some scenarios.

**Solution**

Add a guard to check `Audio.source` is non-null before calling `Audio.disconnect()`

**Additional context**

There might be a chance that the fact `Audio.source` keeps being unset in some scenarios might be a bug. We might also need to investigate it.

### DIFF
--- a/src/bit-systems/audio-emitter-system.ts
+++ b/src/bit-systems/audio-emitter-system.ts
@@ -19,7 +19,9 @@ export function isPositionalAudio(node: AudioObject3D): node is PositionalAudio 
 
 export function cleanupAudio(audio: AudioObject3D) {
   const eid = audio.eid!;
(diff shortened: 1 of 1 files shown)


nickgrato (approved):

keianhzo (approved):
LGTM

PR #6218, opened 2023-08-15
bitECS: Fix Text color problem

### LINKED ISSUE #6216: bitECS: Text component color looks weird
**Description**

With new loader enabled, text component color looks weird. At least it looks different from the one without new loader.

The text stroke color doesn't work or the outline color overrides it.

**To Reproduce**
1. Enter a room with and without new loader
2. Upload a glb including text component
3. Compare the text colors between with and without new loader
4. Confirm that the colors look different

**Expected behavior**

Text component color looks correct

**Screenshots**

From Imaginer

Without new loader

With new loader

**Hardware** 
 - Device: Desktop
 - OS: Windows
 - Browser: Chrome
 
**Additional context**

Probably unexpected type parameters are passed to text gltf inflator, like expecting number but receiving strings. We may need to add validation or type conversion layer.

takahirox:
Fixes #6216 

**Problem**

With new loader enabled text component color can look weird.

**Root issue**

Text inflator expects some parameters are passsed as number but they can be passed as strings from glTF. Then text can look weird due to mismatched data type.

**Solution**

Add casting in text inflator to ensure data types. Adding casting only number typed parameters that actually caused a problem for now.

**Additional context**

This type check and casting problem shouldn't be text component specific. The real root issue is glTF component data is not type checked and validated.

Ideally all the (glTF) inflator parameters should be type checked and validated. We may need to add type check and validation mechanism later.

### DIFF
--- a/src/inflators/text.ts
+++ b/src/inflators/text.ts
@@ -81,8 +81,59 @@ const DEFAULTS: Required<TextParams> = {
   whiteSpace: "normal"
 };
 
+// Parameters from glTF are not type checked then needs to
+// cast to ensure the data types. Otherwise text can look weird
+// because of mismatched typed data.
+// This type problem shouldn't be Text component specific.
+// Ideally type check and cast should be done for all parameters
+// in all (glTF) inflators. Also validation should be done.
+// For now, we cast only number typed text parameters because we
+// nooticed that they are sometimes passed as string types from
+// glTF. If we notice problems with other parameters, we may add
+// casting of other parameters.
+// TODO: Add a generic mechanism to cast and validate inflator params.
+const cast = (params: Required<TextParams>): Required<TextParams> => {
+  const keys: Array<keyof TextParams> = [
+    "curveRadius",
+    "depthOffset",
+    "fillOpacity",
+    "fontSize",
+    "glyphGeometryDetail",
+    "letterSpacing",
+    "lineHeight",
+    "maxWidth",
+    "opacity",
+    "outlineBlur",
+    "outlineOffsetX",
+    "outlineOffsetY",
+    "outlineOpacity",
+    "outlineWidth",
+    "sdfGlyphSize",
+    "strokeOpacity",
+    "strokeWidth",
+    "textIndent"
+  ];
+
+  for (const key of keys) {
+    let value = params[key];
+    if (value === null || value === undefined) {
+      continue;
+    }
+    value = Number(value);
+    if (isNaN(value)) {
+      continue;
+    }
+    // Compile error if remove any.
+    //   TS2322: Type 'number' is not assignable to type 'never'.
+    // Any proper and reasonable solution?
+    (params as any)[key] = value;
+  }
+
+  return params;
+};
+
 export function inflateText(world: HubsWorld, eid: number, params: TextParams) {
-  const requiredParams = Object.assign({}, DEFAULTS, params) as Required<TextParams>;
+  const requiredParams = cast(Object.assign({}, DEFAULTS, params) as Required<TextParams>);
   const text = new TroikaText();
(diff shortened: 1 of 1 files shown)


keianhzo (commented):
Not sure if I understand the approach. We control both end the authoring tools (Spoke/Blender) and the client code so we should be able to make all parameters match types. If there is some parameter that doesn't, wouldn't it be simpler to just cast that specific parameter instead of adding all this casting logic for the whole params object?

takahirox:
1. We can't really trust glTF/glb contents coming from external (eg: user upload, from sketchfab). We can't perfectly assume that the assets are exported with Spoke or Hubs Blender addon properly. Even users can manually (and wrongly) edit their contents. So we need validation layer somewhere. Perhaps validating in glTF inflators is one of the good candidates. We may also rescue some invalid data like with casting. Anyways we need to discuss how we should handle invalid data (eg: Refuse, apply default values, or rescue as much as possible).

2. Imaginar shared glb assets including text component exported with Hubs Blender add-on. `outlineBlur, outlineOfffsetX/Y, outlineWidth, strokeWidth` in the assets seems to be exported as strings although they are expected as numbers in Hubs Client. With the old loder, they seems to be converted to numbers in Hubs Client. Perhaps A-Frame has a mechanism to cast component data with A-Frame component schema. For the assets Imaginar shared, refer to Discord "New loader test with the community" subchannel under meetup channel.

image

I guess Hubs Blender add-on has a bug for exporting text? I'm not sure which properties are exported as wrong type. So added casting for all number typed parameters for now.

keianhzo (changes requested):
Those parameters that you mention (`outlineBlur, outlineOfffsetX/Y, outlineWidth, strokeWidth`) are not only numbers, they can be numbers or strings, both are valid and both are accepted by TroikaText (https://github.com/protectwise/troika/blob/main/packages/troika-three-text/README.md#outlineblur): 
`outlineBlur?: number | `${number}%`;`

In AFrame we were checking this by using a parsing function here:
https://github.com/mozilla/hubs/blob/02ff9d03c384ce6caef14a441945d52f916e2d53/src/components/troika-text.js#L16
and here:
https://github.com/mozilla/hubs/blob/02ff9d03c384ce6caef14a441945d52f916e2d53/src/components/troika-text.js#L40

I suggest we take a similar approach to avoid any breakage and ensure campatiblity.

That would fix the issue and get this to parity with the AFrame code.

The Blender add-on code looks good to me. the property there is exported as a string as we are relying on the Hubs client to determined the correct type based in the "%" symbol presence and cast accordingly.

takahirox:
``${number}%`` style is covered here.

```
    value = Number(value);
    if (isNaN(value)) {
      continue;
    }
```

`Number('12%') -> NaN`

keianhzo:
Oh yes, sorry I misunderstood that. We are correctly handling the three cases:
`${number}`
`${number}%`
`number`
Maybe we should reflect that in the type declaration?

keianhzo (approved):
Oh yes, sorry I misunderstood that. We are correctly handling the three cases. Maybe we should reflect that on the property declaration?
```typescript
outlineBlur?: number | `${number}%` | `${number}`;
```

I like the AFrame approach were you can provide a parsing function for each property but I guess this can work for now.

takahirox:
> Maybe we should reflect that on the property declaration?
>
> ```
> outlineBlur?: number | `${number}%` | `${number}`;
> ```

In my opinion, valid types and types that can be rescued for glTF should be distinguished. So, for example, `outlineBlur?` types should be kept being

```
number | `${number}%`
```

Static type check where possible (JSX) shouldn't really widen the valid types. And we may change how to handle invalid typed data, for example in this PR we rescue strings number but we may change it to refuse in the future.

PR #6223, opened 2023-08-17
bitECS: Take ownership when rotatin or scaling object

### LINKED ISSUE #6199: bitECS: Clicking rotate or scale button doesn't take ownership
**Description**

Clicking rotate or scale button in object menu doesn't seem to take ownership. 

You seem to need to first click an object to take ownership and then you can rotate or scale. Otherwise you don't have ownership then you can't rotate or scale.

**To Reproduce**
1. Enter a room with new loader enabled from two browsers
2. Upload an object from one of the browsers
3. On the other browser, hover cursor on the object and place space bar to show the object menu and then try to rotate or scale with the buttons. (Don't click the object)
4. You don't have ownership so you can't rotate or scale the object.

**Expected behavior**

Rotate or scale button should work even without clicking the object.

**Hardware** 
 - Device: Desktop
 - OS: Windows
 - Browser: Chrome

takahirox:
Fixes #6199

**Problem**

With new loader enabled, objects are not rotated or scaled even if you attempt to rotate or scale objects that you don't have ownership. You first need to click objects to take ownership and then you can rotate or scale.

https://github.com/mozilla/hubs/issues/6199

But it is not a expected limitation. You should be able to rotate or scale objects without taking ownership with other operations beforehand.

**Root issue**

Rotate and Scale with Object menu don't take ownership.

**Solution**

Take ownership when starting rotate or scale.

### DIFF
--- a/src/bit-systems/object-menu.ts
+++ b/src/bit-systems/object-menu.ts
@@ -1,10 +1,12 @@
-import { defineQuery, enterQuery, entityExists, exitQuery, hasComponent } from "bitecs";
+import { addComponent, defineQuery, enterQuery, entityExists, exitQuery, hasComponent } from "bitecs";
 import { Matrix4, Quaternion, Vector3 } from "three";
 import type { HubsWorld } from "../app";
 import {
+  EntityStateDirty,
   HeldRemoteRight,
   HoveredRemoteRight,
   Interacted,
+  Networked,
   ObjectMenu,
   ObjectMenuTarget,
   RemoteRight,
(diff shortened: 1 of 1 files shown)


takahirox:
@johnshaughnessy Can you review this PR? I don't perfectly understand our network and eitnty state API yet.

src/bit-systems/object-menu.ts
```diff
@@ -81,6 +84,10 @@ function moveToTarget(world: HubsWorld, menu: EntityID) {
 //       temporary implementation that rely on the old systems.
 //       They should be rewritten more elegantly with bitecs.
 function startRotation(world: HubsWorld, targetEid: EntityID) {
+  if (hasComponent(world, Networked, targetEid)) {
+    takeOwnership(world, targetEid);
+    addComponent(world, EntityStateDirty, targetEid);
```
johnshaughnessy, line 89: It's good to `takeOwnership` right away when you `startRotation`, but you should probably wait to mark the entity state dirty until after you `endRotation`.

takahirox: Thanks. Regarding entity state dirty, should it be set only for `Networked` entity (that the client has ownership)?

johnshaughnessy: Yes -- if the entity isn't `Networked` then there's no associated entity state that can be stored on the backend.

takahirox: Updated, thanks.

johnshaughnessy (approved):

PR #6231, opened 2023-08-24
bitECS: Fix grabbable pinned media objects

### LINKED ISSUE #6198: bitECS: Pinned video, audio, gif, pdfs are movable
**Description**

Pinned video, audio, gif, pdfs are movable with bitECS.

**To Reproduce**
1. Enter a scene with new loader enabled
2. Upload video, audio, gif, or pdf
3. Pin it
4. Drag it
5. Confirm that it moves

**Expected behavior**

They are not movable as without new loader

**Hardware** 
 - Device: [e.g. Desktop
 - OS: Windows
 - Browser: Chrome

takahirox:
Fixes #6198

**Problem**

When you drop or paste a media file or url, a `MediaLoader` entity is created. This entity downloads the media file and creates a media (e.g., video) entity. The `MediaLoader` entity object then adds the media entity object as its child.

The `MediaLoader` entity does not have a visible object after the loading cube disappears, but the media entity does. Both entities are hover targets, but after the loading cube disappears, the `MediaLoader` entity can no longer be hovered because it does not have a visible object. Only the media entity can be hovered at this point.

The media entity is not a networked entity, but the `MediaLoader` entity is. This means that the `MediaLoader` entity can be pinned (the creator is "reticulum"), but the media entity cannot. If you check whether the media entity is pinned, it will always return false. Then media entity will always be grabbable even if it is pinned (more precisely its parent is pinned).

**Solution**

Check its parent is pinned to check media entity is pinned.

**Changes**

- Add `LoadedByMediaLoader` component to detect a media entity is loaded via `MediaLoader`
- Add a special path for media entity loaded via `MediaLoader` in `hold-system` that checks its parent is pinned to check whether it is pinned.

**Additional context**

This solution is a bit hacky. Let me know if you have simpler and more elegant solution idea.

### DIFF
--- a/src/bit-components.js
+++ b/src/bit-components.js
@@ -157,6 +157,7 @@ export const MediaLoader = defineComponent({
 MediaLoader.src[$isStringType] = true;
 MediaLoader.fileId[$isStringType] = true;
 export const MediaLoaded = defineComponent();
+export const LoadedByMediaLoader = defineComponent();
 export const MediaContentBounds = defineComponent({
   bounds: [Types.f32, 3]
 });
--- a/src/bit-systems/media-loading.ts
+++ b/src/bit-systems/media-loading.ts
@@ -3,6 +3,7 @@ import { Box3, Euler, Vector3 } from "three";
 import { HubsWorld } from "../app";
 import {
   GLTFModel,
+  LoadedByMediaLoader,
   MediaContentBounds,
   MediaLoaded,
   MediaLoader,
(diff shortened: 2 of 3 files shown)


takahirox:
Alternative solution might be, simply recognize an entity is pinned if its any ancestor is pinned (in hold-system) unless there is a case that descendant entity under pinned entity 
wants to be grabbable.

nikk15 (approved):
Thank you for the detailed explanation. I'm happy to approve this to fix the immediate problem. LGTM

PR #6235, opened 2023-08-25
bitECS: Fix Link hover menu direction

### LINKED ISSUE #6233: bitECS: No "Use Avatar" button from back side
**Description**

With new loader enabled, "Use Avatar" button doesn't show up if avatar object is hovered from backside. It shows up without new loader. And it shows up if hovered from frontside.

**To Reproduce**
1. Enter a room with new loader enabled
2. Place an avatar
3. Hover cursor on the avatar object from backside
4. Confirm that "Use Avatar" button doesn't show up

**Expected behavior**

"Use Avatar" button shows up from both sides.

**Screenshots**

Without new loader

image

With new loader

image

**Hardware** 
 - Device: Desktop
 - OS: Windows
 - Browser: Chrome

takahirox:
Fixes #6233

**Problem**

With the new loader enabled link hover menu is not seen if the cursor hovers on link object from back side.

**Root issue**

The link hover menu object is placed as the menu object looks to the target object's front face direction. (The menu object has front-face material.)

**Solution**

Make the menu object look at the camera.

### DIFF
--- a/src/bit-systems/link-hover-menu.ts
+++ b/src/bit-systems/link-hover-menu.ts
@@ -1,4 +1,5 @@
 import { Not, defineQuery, entityExists } from "bitecs";
+import { Matrix4, Vector3 } from "three";
 import type { HubsWorld } from "../app";
 import {
   Link,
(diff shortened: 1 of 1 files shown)


keianhzo (approved):

hrithikwins (commented):
Old loader while menu hovered
image
new loader while menu hovered
image

PR #6238, opened 2023-08-30
Added (some) italian language

LorenzoVonMatterhom:
I didnt translate some debug strings since i think its better for the user if debug info is in english


keianhzo (approved):
Thanks so much for expanding the language support of Hubs!

Issue #6254, opened 2023-09-14
bitECS: Link embedded in scene crashes Hubs


takahirox:
Reported by Imaginer on Discord.

**Description**

Link component embedded in scene crashes Hubs if preview image is clicked.

Console error.

![image](https://github.com/mozilla/hubs/assets/7637832/1b3827f4-e4a8-45fb-8626-47962675b3b1)

**To Reproduce**
1. Make a room with a scene including Link component
2. Enter the room with the new loader enabled
3. Click the link preview image
4. See the console error

Room for testing https://hubs.mozilla.com/fuA2gti/cool-bustling-tract/?newLoader

**Expected behavior**

No crash

**Hardware** 
 - Device: Desktop
 - OS: Windows
 - Browser: Chrome

**Additional context**

Clicking the link button doesn't cause the problem. The problem happens only with clicking preview images.

takahirox:
Personally, this line is very suspecious to me

https://github.com/mozilla/hubs/blob/ca6b79b94957ae5d0951a3b326d66defecac0d97/src/systems/bit-constraints-system.js#L47

In the testing room, if the preview image is clicked, the line seems to return an entity id of the environment root (Three.js glTF root group). Is this an expected behavior?

PR #6258, opened 2023-09-14
bitECS: Check if entity has networked component before taking ownership

keianhzo:
Fixes https://github.com/mozilla/hubs/issues/6254

This PR adds a check for Networked component when a Constraint is added to a body to taking ownership of entities that are not networked.

In some contexts we might want to add constratints to bodies of entities that are not networked. Inthe current use case most of the times this is the case as the only interactable entities are dropped medias but behavior graphs allow creating physics objects that can be interactable and might not necessarily need to be networked (although most of the times they will)

### DIFF
--- a/src/systems/bit-constraints-system.js
+++ b/src/systems/bit-constraints-system.js
@@ -19,7 +19,8 @@ import {
   ConstraintHandLeft,
   ConstraintHandRight,
   ConstraintRemoteLeft,
-  ConstraintRemoteRight
+  ConstraintRemoteRight,
+  Networked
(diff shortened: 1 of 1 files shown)


takahirox:
Haven't looked into the details yet but let me share my impression for now.

The expected behavior seems to do nothing when the preview image is clicked. So clicking it causes crash is weird me because that means something interaction processing runs when it's clicked. I was thinking the root fix was not to run such interaction processing. Anyways, I will look into details next week.

takahirox (approved):
Looking good to me

Issue #6252, opened 2023-09-13
bitECS: Hide delete button on pinned objects


takahirox:
**Description**

The delete button in the object menu is shows even on pinned objects but it shouldn't because pinned objects shouldn't be deleted.

**To Reproduce**
1. Enter a room with the new loader enabled
2. Place and pin an object
3. Hold space bar while the cursor hovers on the object
4. See that the delete button is shown

**Expected behavior**

Delete button shouldn't be shown on pinned objects as with the old loader.

**Hardware** 
 - Device: Desktop
 - OS: Windows
 - Browser: Chrome

PR #6259, opened 2023-09-14
bitECS: Fix hiding the remove button from the object menu

keianhzo:
Fixes https://github.com/mozilla/hubs/issues/6252

This PR fixes the hiding of the remove button based on the pinned state of the entity.


src/bit-systems/object-menu.ts
```diff
@@ -237,6 +237,7 @@ function updateVisibility(world: HubsWorld, menu: EntityID, frozen: boolean) {
   world.eid2obj.get(ObjectMenu.unpinButtonRef[menu])!.visible = visible && isPinned(target);
   world.eid2obj.get(ObjectMenu.pinButtonRef[menu])!.visible =
     visible && !isPinned(target) && canPin(APP.hubChannel!, target);
+  world.eid2obj.get(ObjectMenu.removeButtonRef[menu])!.visible = !isPinned(target);
```
takahirox, line 240: Isn't `visible && !isPinned(target);`?

keianhzo: Oh yep, that's for that.

Issue #6264, opened 2023-09-15
bitECS: Named waypoints and location.hash navigation doesn't work


keianhzo:
**Description**
Named waypoints and location.hash navigation doesn't work. This hasn't been implemented yet.
https://github.com/mozilla/hubs/blob/5dc6fa3e01451c81423cc8d8b7c1ece2c36d174d/src/bit-systems/waypoint.ts#L208-L209

**To Reproduce**
1. Open a room that has a named waypoint (a named waypoint is a waypoint that you can navigate to using a URL hash)
2. Drop a link to the hash anchor in the scene
3. Click on it
4. See that you are not transported to the waypoint

**Expected behavior**
You should teleport to the waypoint

[named-waypoint-test.glb.zip](https://github.com/mozilla/hubs/files/12618630/named-waypoint-test.glb.zip)

PR #6265, opened 2023-09-15
bitECS: Enable waypoint hash navigation

keianhzo:
Fixes https://github.com/mozilla/hubs/issues/6264

This PR enables waypoint hash navigation. I don't have a clear view of all the multiple cases around this use case so any help tracking down possible cases is welcome.

### DIFF
--- a/src/bit-systems/waypoint.ts
+++ b/src/bit-systems/waypoint.ts
@@ -172,11 +172,8 @@ export function waypointSystem(
 
     const hashUpdated = window.location.hash !== "" && previousWaypointHash !== window.location.hash;
     if (hashUpdated && initialSpawnHappened) {
-      const waypointName = window.location.hash.replace("#", "");
-      if (eid) {
-        moveToWaypoint(world, eid, characterController, previousWaypointHash === null);
-        window.history.replaceState(null, "", window.location.href.split("#")[0]); // Reset so you can re-activate the same waypoint
-      }
+      moveToWaypoint(world, eid, characterController, previousWaypointHash === null);
+      window.history.replaceState(null, "", window.location.href.split("#")[0]); // Reset so you can re-activate the same waypoint
       previousWaypointHash = window.location.hash;
     }
 
--- a/src/bit-systems/waypoint.ts
+++ b/src/bit-systems/waypoint.ts
@@ -172,9 +172,13 @@ export function waypointSystem(
 
     const hashUpdated = window.location.hash !== "" && previousWaypointHash !== window.location.hash;
     if (hashUpdated && initialSpawnHappened) {
-      moveToWaypoint(world, eid, characterController, previousWaypointHash === null);
-      window.history.replaceState(null, "", window.location.href.split("#")[0]); // Reset so you can re-activate the same waypoint
-      previousWaypointHash = window.location.hash;
+      const waypointName = window.location.hash.replace("#", "");
+      const waypointObj = world.eid2obj.get(eid);
+      if (waypointObj && waypointObj.name === waypointName) {
+        moveToWaypoint(world, eid, characterController, previousWaypointHash === null);
+        window.history.replaceState(null, "", window.location.href.split("#")[0]); // Reset so you can re-activate the same waypoint
+        previousWaypointHash = window.location.hash;
+      }
     }
 
     if (hasComponent(world, Interacted, eid)) {
--- a/src/bit-systems/waypoint.ts
+++ b/src/bit-systems/waypoint.ts
@@ -170,6 +170,9 @@ export function waypointSystem(
       }
     }
 
+    // When a scene is opened with a named waypoint we have to make sure that the scene default waypoint
+    // doesn't override it and that we correctly spawn in the named waypoint from the url.
+    // https://github.com/mozilla/hubs/issues/2833
     const hashUpdated = window.location.hash !== "" && previousWaypointHash !== window.location.hash;
     if (hashUpdated && initialSpawnHappened) {
       const waypointName = window.location.hash.replace("#", "");
--- a/src/bit-systems/waypoint.ts
+++ b/src/bit-systems/waypoint.ts
@@ -107,7 +107,6 @@ function* moveToSpawnPointJob(world: HubsWorld, characterController: CharacterCo
   if (spawnPoints.length) {
     const waypoint = spawnPoints[Math.floor(Math.random() * spawnPoints.length)];
     moveToWaypoint(world, waypoint, characterController, true);
-    initialSpawnHappened = true;
   } else {
     console.warn("Could not find any available spawn points, spawning at the origin.");
     characterController.enqueueWaypointTravelTo(new Matrix4().identity(), true, {
@@ -116,8 +115,8 @@ function* moveToSpawnPointJob(world: HubsWorld, characterController: CharacterCo
       snapToNavMesh: true,
       willMaintainInitialOrientation: false
     });
-    initialSpawnHappened = true;
   }
+  initialSpawnHappened = true;
 }
 
 let spawnJob: Coroutine | null = null;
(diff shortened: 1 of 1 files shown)


src/bit-systems/waypoint.ts
```diff
@@ -164,6 +171,16 @@ export function waypointSystem(
       }
     }
 
+    const hashUpdated = window.location.hash !== "" && previousWaypointHash !== window.location.hash;
+    if (hashUpdated && initialSpawnHappened) {
+      const waypointName = window.location.hash.replace("#", "");
+      if (eid) {
```
takahirox, line 177: What is this `eid` check for?

src/bit-systems/waypoint.ts
```diff
@@ -112,6 +116,7 @@ function* moveToSpawnPointJob(world: HubsWorld, characterController: CharacterCo
       snapToNavMesh: true,
       willMaintainInitialOrientation: false
     });
+    initialSpawnHappened = true;
```
takahirox, line 119: Can't `initialSpawnHappened` be after the `if` block?

```
  if (spawnPoints.length) {
    const waypoint = spawnPoints[Math.floor(Math.random() * spawnPoints.length)];
    moveToWaypoint(world, waypoint, characterController, true);
  } else {
    console.warn("Could not find any available spawn points, spawning at the origin.");
    characterController.enqueueWaypointTravelTo(new Matrix4().identity(), true, {
      willDisableMotion: false,
      willDisableTeleporting: false,
      snapToNavMesh: true,
      willMaintainInitialOrientation: false
    });
  }
  initialSpawnHappened = true;
```

src/bit-systems/waypoint.ts
```diff
@@ -164,6 +171,16 @@ export function waypointSystem(
...
 
+    const hashUpdated = window.location.hash !== "" && previousWaypointHash !== window.location.hash;
+    if (hashUpdated && initialSpawnHappened) {
+      const waypointName = window.location.hash.replace("#", "");
+      if (eid) {
+        moveToWaypoint(world, eid, characterController, previousWaypointHash === null);
+        window.history.replaceState(null, "", window.location.href.split("#")[0]); // Reset so you can re-activate the same waypoint
+      }
+      previousWaypointHash = window.location.hash;
+    }
```
takahirox, line 182: Can you elaborate why `initialSpawnHapped` is needed for me for review?

src/bit-systems/waypoint.ts
keianhzo: Good call, moved.

src/bit-systems/waypoint.ts
keianhzo: Yeah, I think I added it at some point and I forgot to remove. Thanks.

src/bit-systems/waypoint.ts
keianhzo: Looking at the code and this issue https://github.com/mozilla/hubs/issues/2833 I believe is so you can spawn to a specific named waypoint and the scene default spawn point does not override it when entering.

takahirox: Would you mind adding an inline comment about it? It is not straightforward to new readers.

takahirox:
I think ideally systems shouldn't have state but `initialSpawnHappened` and `previousWaypointHash` sound like state to me. I'm thinking whether we can come up with  more elegant way... At least I think we need inline comment that it should be fixed later.

src/bit-systems/waypoint.ts
takahirox, line 182: This part is very confusing to me... Should it really be processed againts `waypointQuery`? There is a chance that there are multiple entities that have waypoint components but only the first entity can be processed because `previousWaypointHash !== window.location.hash` will be `false` for the next ones.

If it is intentional, the code should be clearer.

src/bit-systems/waypoint.ts
```diff
@@ -164,6 +171,16 @@ export function waypointSystem(
       }
     }
 
+    const hashUpdated = window.location.hash !== "" && previousWaypointHash !== window.location.hash;
+    if (hashUpdated && initialSpawnHappened) {
+      const waypointName = window.location.hash.replace("#", "");
```
takahirox, line 176: `waypointName` seems to be unused.

src/bit-systems/waypoint.ts
```diff
@@ -164,6 +170,20 @@ export function waypointSystem(
...
     }
 
+    // When a scene is opened with a named waypoint we have to make sure that the scene default waypoint
+    // doesn't override it and that we correctly spawn in the named waypoint from the url.
+    // https://github.com/mozilla/hubs/issues/2833
+    const hashUpdated = window.location.hash !== "" && previousWaypointHash !== window.location.hash;
+    if (hashUpdated && initialSpawnHappened) {
+      const waypointName = window.location.hash.replace("#", "");
+      const waypointObj = world.eid2obj.get(eid);
+      if (waypointObj && waypointObj.name === waypointName) {
```
takahirox, line 180: Nit: The existence of `waypointObj` is not checked below at https://github.com/mozilla/hubs/blob/c60c6cbb9b5452c81d1c28fcc86efede9ccca126/src/bit-systems/waypoint.ts#L203-L204 so no need to check here for the consistency, or should be checked for the both.

src/bit-systems/waypoint.ts
takahirox: Perhaps the code will be clearer if `hashUpdated` check is done outside of the loop and named waypoint is handled in another query loop.

```typescript
function export function waypointSystem(...) {
  const hashUpdated = window.location.hash !== "" && previousWaypointHash !== window.location.hash;
  if (hashUpdated && initialSpawnHappened) {
    const waypointName = window.location.hash.replace("#", "");
    waypointQuery(world).forEach(eid => {
      const waypointObj = world.eid2obj.get(eid);
      if (waypointObj && waypointObj.name === waypointName) {
        moveToWaypoint(world, eid, characterController, previousWaypointHash === null);
        window.history.replaceState(null, "", window.location.href.split("#")[0]); // Reset so you can re-activate the same waypoint
        previousWaypointHash = window.location.hash;
        break;
      }
    });
  }

  waypointQuery(world).forEach(eid => {
    // process others
    ...
  });
}
```

takahirox: I think the comment need more explanation because it may be unclear to new readers about how `initialSpawnHappened` can avoid the override. May be you can refer https://github.com/mozilla/hubs/pull/2837/files#r468103137

And again, I don't think `initialSpawnHappened` and `previousWaypointHash` are good approaches in systems because they sound like state in systems to me. It's ideal if we can come up with proper approaches. Or we may think of managing them in components. At lease we may need comment that they need to be improved later.

keianhzo: Yes, that's reasonable, refactored.

src/bit-systems/waypoint.ts
keianhzo: Actually that code was there before and we haven't had crashes so I assume we expect that object to exist. I'm going to remove the guard.

takahirox (approved):
I approve once these three are reflected

* https://github.com/mozilla/hubs/pull/6265#discussion_r1337956007
* https://github.com/mozilla/hubs/pull/6265#issuecomment-1733004031
* https://github.com/mozilla/hubs/pull/6265#discussion_r1337982355

I don't prefer system state `initialSpawnHappened` and `previousWaypointHash` approach but we add inline comments about it for now and it may be revisited later.

src/bit-systems/waypoint.ts
keianhzo: Added the PR comment from above to the comment.

keianhzo:
Comments updated. I agree on the global variables, probably the right place would be a scene level component.

Issue #6263, opened 2023-09-15
bitECS: No link button when Link and Image components are defined


takahirox:
Reported by Imaginer on Discord

**Description**

If image and link Hubs component is defined on the same node, Link button isn't shown even the cursor hovers on the image object.

**To Reproduce**
1. Enter a room with the new loader enabled
2. Upload an asset including a node that has both image and link Hubs component
3. Hover the cursor on the image object
4. See that no "Link" button is shown

Room for testing: https://hubs.mozilla.com/sDg4ioF/fluffy-shiny-cosmos/?newLoader

**Expected behavior**

Link button is shows when the cursor hovers on the image object

**Screenshots**

![image](https://github.com/mozilla/hubs/assets/7637832/95a00dc4-fdf5-4d91-ba5b-eba29a4c47a4)

**Hardware** 
 - Device: Desktop
 - OS: Windows
 - Browser: Chrome

**Additional context**

I labeled this issue as P1. Although I'm not sure if Link + Image is major use, it may be worth to investigate the problem because multiple components on a node may cause the problem and the problem may not be Link + Image specific.

Exairnous:
I would guess that Link + Image sees reasonably high usage, and it is used in the current official Mozilla demo scene/room.

takahirox:
Thanks, then P1 makese sense more

PR #6272, opened 2023-09-19
bitECS: Support media with link component

keianhzo:
Fixes https://github.com/mozilla/hubs/issues/6263

In the AFrame loader when an object had a `link` and a `image` components (also works with `video` and `loader`) the media would override the link media

In the new loader we don't support a media loader loading multiple media. The second media would override the first one's parameters. 

Using image/video/model with the link component seems to be a pretty common use case. This PRs adds a hack to support that specific use case.

Another alternative would be to add support for multiple loads to the media loader but I'm no sure if that's something that we want...

### DIFF
--- a/src/bit-systems/media-loading.ts
+++ b/src/bit-systems/media-loading.ts
@@ -31,6 +31,8 @@ import { loadAudio } from "../utils/load-audio";
 import { loadHtml } from "../utils/load-html";
 import { MediaType, mediaTypeName, resolveMediaInfo } from "../utils/media-utils";
 import { EntityID } from "../utils/networking-types";
+import { LinkType, inflateLink } from "../inflators/link";
+import { inflateGrabbable } from "../inflators/grabbable";
 
 const getBox = (() => {
   const rotation = new Euler();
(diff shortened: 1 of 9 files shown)


takahirox:
As commented inline this solution is very hacky for sure. Give me some time to think whether there may be clearer solution...

src/inflators/media-loader.ts
```diff
@@ -28,4 +29,7 @@ export function inflateMediaLoader(
     MediaLoader.fileId[eid] = APP.getSid(fileId)!;
   }
   MediaLoader.src[eid] = APP.getSid(src)!;
+  if (linkSrc) {
+    MediaLoader.linkSrc[eid] = APP.getSid(linkSrc)!;
+  }
```
takahirox, line 34: `MediaLoader.linkSrc[eid]` needs to be initialized even if `linkSrc` is not set otherwise it can be problem if entity recycle happens that new entity can reuse linkSrc of the old one 

```typescript
// Assumes 0 indicates null string, not sure if this assumption is correct tho
MediaLoader.linkSrc[eid] = linkSrc !== undefined ? APP.getSid(linkSrc) : 0;
```

takahirox:
Let me share my thoughts.

With the new loader enabled, we prevent a glTF node that has multiple Hubs components that generates Three.js object. We preprocess such nodes to make multiple Hubs Component glTF nodes under the node. Under this policy, Link + Media is not doable as is.

Probably we don't spend long time for this problem. So probably preprocessing Link + Media glTF Hubs Components and reusing the existing code as much as possible would be good. Also, adding a special path after loading a media as this PR does would be reasonable.

But I don't prefer adding `linkSrc` property to `MediaLoader` because `MediaLoader` component will have strong dependency with `Link`. Maybe introducing a new component (eg: `MediaLink`) would be simpler.

This code in the media loading system

```typescript
  const linkSrc = APP.getString(MediaLoader.linkSrc[eid]);
  if (linkSrc) {
    ...
  }
```

would be replaced with like

```typescript
  if (hasComponent(world, MediaLink, eid)) {
    const linkSrc = APP.getString(MediaLink.src[eid]);
    ...
  }
```

keianhzo:
Yes, that makes sense and I like that is much more clear, updated.

src/bit-systems/media-loading.ts
```diff
@@ -179,36 +182,50 @@ function* loadByMediaType(
...
+      mediaEid = yield* loadHtml(world, canonicalUrl, thumbnail);
+      break;
     default:
       throw new UnsupportedMediaTypeError(eid, mediaType);
   }
+
+  if (hasComponent(world, MediaLink, eid)) {
+    inflateLink(world, mediaEid, { href: APP.getString(MediaLink.src[eid])!, type: LinkType.LINK });
+    inflateGrabbable(world, mediaEid, { cursor: true, hand: false });
+  }
```
takahirox, line 226: `MediaLink` is designed to be used only in media loading so it would be better to remove it when `MediaLoader` component is removed. Maybe in `mediaLoaderExitQuery` would be the best place.

keianhzo: Absolutely https://github.com/mozilla/hubs/pull/6327

Thanks.

Issue #6283, opened 2023-09-25
The word "cr*zy" is included in the random room name generator list


LaurieAnnis:
I just got a random room name assigned with the word cr*zy in the mix.  I asked in Discord," how can I go about requesting that that word be removed from the name generator?" and the response was to pleas submit (this) as a bug report.

**To Reproduce**
1. Create a room
2. Randomly get assigned a room name with the word cr*zy (offensive term to describe mental illness) in it.

**Expected behavior**
Offensive words to be omitted from the possible random words.

zfox23:
Thanks for your submission, @LaurieAnnis! There are two PRs in the pipeline to resolve this issue, referenced above.

zfox23:
PRs merged :)

PR #6284, opened 2023-09-25
Remove 'crazy' from the Hubs name generation wordlist

zfox23:
1/2 of the resolution of https://github.com/mozilla/hubs/issues/6283. Good idea to remove this from the wordlist.

The second half of this issue is resolved with this PR: https://github.com/mozilla/reticulum/pull/710


mikemorran (approved):
Looks good!

PR #6287, opened 2023-09-26
bitECS New loader pinning updates

keianhzo:
This PR updates a few things:
- Adds support for pinning in the object list
- Changes the pinning design for the new loader to remove the concept of promotion token and simplify the pinning flow (see the diagrams below for details)
- Updates the menus to hide/show buttons based on the pinned state of the object.

**Pinning changes**
We have simplified the way pinning works for the new loader removing the case where an uploaded file can only be pinned initially by the uploader. Now any user can pin an uploaded file so in practice an uploaded media pinning works the same as a url media.
 
This depends on this Reticulum PR: https://github.com/mozilla/reticulum/pull/712

This is how pinning worked in the old loader:

Pinning Old Loader

Here it is how it is going to work in the new loader:

Pinning New Loader

For reference this is what the expected object menu layout should be:

Object Menu States

### DIFF
--- a/src/bit-components.js
+++ b/src/bit-components.js
@@ -151,11 +151,9 @@ export const CameraTool = defineComponent({
 export const MyCameraTool = defineComponent();
 export const MediaLoader = defineComponent({
   src: Types.ui32,
-  flags: Types.ui8,
-  fileId: Types.ui32
+  flags: Types.ui8
 });
 MediaLoader.src[$isStringType] = true;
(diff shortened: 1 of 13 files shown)


takahirox:
> This PR builds on top of https://github.com/mozilla/hubs/pull/6067 and add support for pinning in the object list. It also removes the usage of promotion tokens from file backed media.

Can you separate the PRs, one for object list pinning support and the other one for removing the usage of promotion token, for easier review?

Issue #6281, opened 2023-09-22
bitECS: Object menu is hidden behind the objects sometimes


keianhzo:
**Description**
Object menu is hidden behind the object sometimes

**To Reproduce**
1. Enter a hubs room using new loader
2. Spawn a duck. Open the chat and type `/duck`
3. Scale the duck

**Expected behavior**
The menu should always be in front of the object

**Screenshots**
<img width="329" alt="Screenshot 2023-09-22 at 11 53 14" src="https://github.com/mozilla/hubs/assets/837184/595b5e54-f8e6-4a20-81d0-aab7a9db54fb">

takahirox:
https://github.com/mozilla/hubs/issues/6192#issuecomment-1732749240

PR #6289, opened 2023-09-27
bitECS: Fix object menus transforms

keianhzo:
Fixes https://github.com/mozilla/hubs/issues/6281

This PR makes fixes the issue that we have with the object menu being hidden behind. It also refactors some code into a `object-menu-transform-system` responsible for correctly placing the object menu. I've updated all the in-world menu to use this system.

This PR doesn't address the problem of where to place the menu when we are inside the bounding sphere of an object. I think a correct placement in that situation would require a menus redesign so I haven't change it.

TODO: Another thing that could benefit from being refactored to its own system would be the menu visibility.

### DIFF
--- a/src/bit-components.js
+++ b/src/bit-components.js
@@ -257,7 +257,8 @@ export const VideoMenu = defineComponent({
   trackRef: Types.eid,
   headRef: Types.eid,
   playIndicatorRef: Types.eid,
-  pauseIndicatorRef: Types.eid
+  pauseIndicatorRef: Types.eid,
+  clearTargetTimer: Types.f64
 });
 export const AudioEmitter = defineComponent({
   flags: Types.ui8
@@ -286,7 +287,9 @@ export const ObjectMenu = defineComponent({
   rotateButtonRef: Types.eid,
   mirrorButtonRef: Types.eid,
   scaleButtonRef: Types.eid,
-  targetRef: Types.eid
+  targetRef: Types.eid,
+  handlingTargetRef: Types.eid,
+  flags: Types.ui8
 });
 // TODO: Store this data elsewhere, since only one or two will ever exist.
 export const LinkHoverMenu = defineComponent({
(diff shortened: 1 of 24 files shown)


takahirox:
My thoughts for now before looking into the implementation details...

> Should we make them all behave the same? Should we make them behave like the old loader? I think the fact that the object menu is always facing the player is better compared to the AFrame implementation but I'm not sure if it makes sense for all menus... thoughts?

Personally I prefer always facing the menu to the viewer because the menu is for viewer (player/user).

I'm not sure if how it looks like but I have another approach, rather than adjusting the menu position, in my mind to make the menu always visible, that is turn off the depth test. Maybe it would be easier, I'm not sure it looks good tho.

By the way, probably introdubing `object-menu-transform-system` would be good for simplicity.

keianhzo:
For documentation. The expected behavior of menus after a chat with from @johnshaughnessy is this:

> The algorithm should be something like:
> 
> - If the entity is flat media, place the menu parallel to the media on whichever side is facing us. Use sign of the dot product (dot(entityPosition - cameraPosition), entityForwardInWorldSpace)) to determine which side we're on.
> - If we are inside the bounding sphere of the entity we're hovering on, place the menu near the intersection point, facing the camera and scaled appropriately.
> - If we are outside the bounding sphere of the entity we're hovering on, place the menu on the bounding sphere at the point that intersects the ray from cameraPosition in the direction (intersectionPosition - cameraPosition). Orient the menu to face the camera and scale it appropriately.
> 
> 
> This algorithm determines the transform that the menu should end up with. Once it's working, you may want to store the scale as the menu's "target scale" and then animate from a small scale to the target scale a brief window of time after the menu appears.
> In some cases, the bounding sphere may be invalid. For example I think three js bounding sphere calculations may fail for animated skinned meshes. I can't remember exactly. 
> 
> The point is if you get NaN or an impossibly small number for the radius of the bounding sphere, you should instead place the menu at a point near the intersection point: Start at the intersection point and then move slightly towards the camera.
> 
> This special case may prove to be unnecessary, because maybe it ends up being the same as case n.3 above.

I also prefer the always visible to the user version of the menu but let's keep it consistent with what we already have for now.

takahirox:
Just in case let me share why I didn't fix this issue in the initial bitECS implementation.

Internally we have been discussing new object handling UI/UX and I thought we move to it soon after we officially introduced the new loader so I didn't feel it's worth to spend time to adjust the object menu position.

But we may need to reschedule so fixing the object menu visibility with the old loader approach is making more sense now.

keianhzo:
@takahirox I've updated the PR with some updates to unify the way all in-world menus behave.

Issue #6315, opened 2023-10-09
bitECS - Old scenes may have nav meshes visible by default, we should ensure reverse compatibility


mikemorran:
**Description**
@Exairnous explained that the new system may require users to configure their nav mesh correctly to make sure they are invisible. While not a bug, this may render many older scenes incompatible with the new system, as their nav meshes will be visible and impact the quality of their work.

**To Reproduce**
1. Create a room with [this scene](https://hubs.mozilla.com/scenes/98AwrmE/pnw-asset-pack-demo-scene) and enable the new loader.

**Expected behavior**
The nav mesh should function but not be visible

takahirox:
Thanks for the problem report. Confirmed the problem. I don't think this behavior change is intentional. It should be invisible. Let me investigate the root issue.

takahirox:
audio debug system seems to be related. In the old loader it seems to make the nav mesh group invisible at https://github.com/mozilla/hubs/blob/a1ce4edd9ab8cac3fedd6c997f84e29fa151b979/src/systems/audio-debug-system.js#L176 while maybe in the new loader it doesn't.

PR #6320, opened 2023-10-12
Set the nav mesh visibility when a nav mesh is loaded

keianhzo:
Fixes https://github.com/mozilla/hubs/issues/6315

This PR sets the nav-mesh visibility when a nav-mesh component is added to make sure that the mesh visibility is always aligned with the preference setting.

### DIFF
--- a/src/bit-systems/audio-debug-system.ts
+++ b/src/bit-systems/audio-debug-system.ts
@@ -119,7 +119,9 @@ getScene().then(() => {
 const addDebugMaterial = (world: HubsWorld, navEid: number) => {
   const obj = world.eid2obj.get(navEid);
   if (obj) {
-    const navMesh = obj as Mesh;
+    // Some older scenes have the nav-mesh component as an ancestor of the mesh itself
+    // TODO the "as any" here is because of incorrect type definition for getObjectByProperty. It was fixed in r145
+    const navMesh = obj.getObjectByProperty("isMesh", true as any) as Mesh;
     navMesh.visible = isEnabled;
     if (nav2mat.has(navEid)) return;
     nav2mat.set(navEid, navMesh.material);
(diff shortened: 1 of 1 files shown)


src/bit-systems/audio-debug-system.ts
```diff
@@ -154,6 +154,13 @@ export function audioDebugSystem(world: HubsWorld) {
   navMeshExitQuery(world).forEach(navEid => {
     removeDebugMaterial(world, navEid);
   });
+  navMeshEnterQuery(world).forEach(navEid => {
+    if (isEnabled) {
+      addDebugMaterial(world, navEid);
+    } else {
+      removeDebugMaterial(world, navEid);
```
takahirox, line 161: Does `removeDebugMaterial` really need to be called here? Does it mean there is a chance that debug material is already added?

src/bit-systems/audio-debug-system.ts
```diff
@@ -154,6 +154,13 @@ export function audioDebugSystem(world: HubsWorld) {
...
+    }
+  });
```
takahirox, line 163: Can't this code be simplified and straightforward more?

```typescript
  navMeshEnterQuery(world).forEach(navEid => {
    if (isEnabled) {
      addDebugMaterial(world, navEid);
    } else {
      removeDebugMaterial(world, navEid);
    }
  });
  if (isEnabled && uniforms) {
    navMeshEnterQuery(world).forEach(navEid => {
      isEnabled && addDebugMaterial(world, navEid);
    });
    ...
```

For example, I'm not sure if the second `navMeshEnterQuery(world)` returns non-empty array because the first `navMeshEnterQuery(world)` is always called.

takahirox:
Honestly setting the default visibility for nav mesh in audio debug system is very unintuitive to me. Ideally it is defauled (as invisible) somewhere else?

It might be ok to revisit this design later tho.

keianhzo:
Not sure about that. The audio-debug service is the only point were we manage the nav-mesh visibility so it doesn't feel wrong to me if we keep that responsibility there instead of spreading it. I'm going to merge this and we can think about it in the future.

takahirox:
It is ok that debug audio system controls nav meshes visibility. But I think the default nav mesh visibility should be set up when it's initialized. The default nav mesh visibility shouldn't really rely to audio debug system. Imagine a case that audio debug system is removed due to a certain reason. It shouldn't affect the default nav mesh visibility. In general less dependencies better design.

Also imagine that a new reader want to know where the default nav mesh is set. It is hard for the one to find from audio debug system because it requires to know background. More intuitive design more maintainable.

Issue #6316, opened 2023-10-09
bitECS - Some scene changes broken with "tried loading multiple nav mesh"


mikemorran:
**Description**
We encountered this at last friday's meetup. Changing scenes to certain environments freezes the screen and we see a nav.js line 14 error "tried loading multiple nav meshes".

**To Reproduce**
1. Go to a room with new loader enabled
2. Attempt to change scenes to this url: https://hubs.mozilla.com/scenes/0MdfQwB/cloud-nine

**Expected behavior**
The scene change should complete without errors.

takahirox:
From this code

https://github.com/mozilla/hubs/blob/e6e92b8291cf83db9df1c097725d15deb7974f10/src/systems/nav.js#L13-L16

With the new loader, the old nav mesh doesn't seem to be released in a proper way. But as the code indicates, it `nav.loadMesh()` seems to release the old one as fallback so perhaps not a serious problem.

Yes, the old nav mesh should be released with a proper way and should be no error, but we may remove P1 label because probably functionality is ok.

keianhzo:
It seems that it's actually related with an exception loading invalid particle images. It seems that Aframe caught those exceptions as we were are the same thing in the AFrame system. I've opened a PR to fix this.

takahirox:
By the way @keianhzo what do you think of the console error? Does it happen because the scene has multiple nav meshes? If so, with the new loader multiple nav meshes in a scene is not accepted (while with the old loader they are accepted)?

keianhzo:
It was actually a difference between how we were processing the nav-mesh in the old and new loader, I've opened a PR: https://github.com/mozilla/hubs/pull/6326

PR #6321, opened 2023-10-12
Catch exceptions when loading particle images

keianhzo:
Fixes https://github.com/mozilla/hubs/issues/6316

This PR catches exceptions when loading particle images. It seems that Aframe caught those exceptions as we were are the same thing in the AFrame system.


takahirox (approved):
Looking good.

Nit: We may be able to think of removing ParticleEmitter component if failed. But it's nit and ok to make another PR if the idea sounds good.

keianhzo:
Good idea. Added.

Issue #6248, opened 2023-09-11
bitECS: ignoredComponents in model.tsx bug


takahirox:
Reported by StandardDrucker on Discord

**Description**

The `model.tsx` code immediately applies camel case to the component name, then attempts to exclude ignored components that are listed in `ignoredComponents`. But the entries in `ignoredComonents` do not use camel case.

https://github.com/mozilla/hubs/blob/051996a9e158a0bcff8c5ebb38590af92870fa3f/src/inflators/model.tsx#L34-L35

**Expected behavior**

Either the exclusion check should be done before applying camel case, or the `ignoredComponents` entries should be rewritten in camel case.

**Hardware** 
 - Device: Desktop
 - OS: Windows
 - Browser: Chrome

PR #6322, opened 2023-10-12
Check ignored components with the original component name

keianhzo:
Fixes https://github.com/mozilla/hubs/issues/6248

### DIFF
--- a/src/inflators/model.tsx
+++ b/src/inflators/model.tsx
@@ -32,7 +32,7 @@ function inflateComponents(
 ) {
   Object.keys(components).forEach(name => {
     const inflatorName = camelCase(name);
-    if (ignoredComponents.includes(inflatorName)) return;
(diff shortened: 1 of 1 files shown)


src/inflators/model.tsx
```diff
@@ -32,7 +32,7 @@ function inflateComponents(
 ) {
   Object.keys(components).forEach(name => {
     const inflatorName = camelCase(name);
-    if (ignoredComponents.includes(inflatorName)) return;
+    if (ignoredComponents.includes(name)) return;
```
takahirox, line 35: nit: `camelCase` can be called after checking `ignoredComponents`.

```typescript
    if (ignoredComponents.includes(name)) return;
    const inflatorName = camelCase(name);
    if (!gltfInflatorExists(inflatorName)) {
```

PR #6326, opened 2023-10-13
Load nav-mesh only once

keianhzo:
Addresses https://github.com/mozilla/hubs/issues/6316#issuecomment-1760208594

In AFrame the nav mesh load was happening only once for the first found nav-mesh component:
https://github.com/mozilla/hubs/blob/f15d84aa019c0b2dfa34be8c91c9b82a6bad1c58/src/systems/environment-system.js#L137-L140
For the new loader we were loading the nav-mesh while traversing the scene so if multiple nav meshes were found we were effectively using the last traversed one.

This PR changes that behavior so we only process the nav-mesh once.

### DIFF
--- a/src/bit-systems/scene-loading.ts
+++ b/src/bit-systems/scene-loading.ts
@@ -81,15 +81,6 @@ function* loadScene(
         // TODO: In three.js, update reflection probes so that boxes are defined in local space.
         (o as any).box.applyMatrix4(o.matrixWorld);
       }
-      if (hasComponent(world, NavMesh, o.eid!)) {
-        // TODO the "as any" here is because of incorrect type definition for getObjectByProperty. It was fixed in r145
-        const navMesh = o.getObjectByProperty("isMesh", true as any) as Mesh;
-        // Some older scenes have the nav-mesh component as an ancestor of the mesh itself
-        if (navMesh !== o) {
-          console.warn("The `nav-mesh` component should be placed directly on a mesh.");
-        }
-        sceneEl.systems.nav.loadMesh(navMesh, "character");
-      }
 
       if (!skybox && hasComponent(world, Skybox, o.eid!)) {
         skybox = o as Sky;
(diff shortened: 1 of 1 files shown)


src/bit-systems/scene-loading.ts
```diff
@@ -108,12 +99,26 @@ function* loadScene(
...
+
+    // Create scene physics
     if (findChildWithComponent(world, TrimeshTag, scene) || findChildWithComponent(world, HeightFieldTag, scene)) {
       console.log("heightfield or trimesh found on scene");
     } else {
-      let navMeshEid = findChildWithComponent(world, NavMesh, scene);
-      if (navMeshEid) {
-        inflatePhysicsShape(world, navMeshEid, {
+      if (isHighDensity && navMesh) {
+        inflatePhysicsShape(world, navMesh.eid!, {
```
takahirox, line 121: nit:

Can't `navMeshEid` be used here instead of `navMesh` and `navMesh.eid!`?

```typescript
      if (isHighDensity && navMeshEid) {
        inflatePhysicsShape(world, navMeshEid, {
```

If it can be, `navMesh` can be scoped into the `if` block above.

```typescript
    if (navMeshEid) {
      const navMeshObj = world.eid2obj.get(navMeshEid);
      // TODO the "as any" here is because of incorrect type definition for getObjectByProperty. It was fixed in r145
      const navMesh = navMeshObj?.getObjectByProperty("isMesh", true as any) as Mesh;
      // Some older scenes have the nav-mesh component as an ancestor of the mesh itself
      if (navMesh !== navMeshObj) {
        console.warn("The `nav-mesh` component should be placed directly on a mesh.");
      }
      sceneEl.systems.nav.loadMesh(navMesh, "character");
    }
```

takahirox:
I think it's ok to load nav mesh only once per a scene, especially if the behavior is same as the one with the A-Frame based implementation.

I want to make clear one thing. Is having multiple nav meshes in a scene valid? If invalid, the blender add-on may need to be updated to block to export, too (I assumes they are set up with the blender add-on). If valid, we may need inline TODO comment about supporting multiple nav meshes in the future and console warning that we ignore other nav meshes for now.

keianhzo:
No, it's not, we only support one nav mesh per scene. I've opened an issues to show a warning when there is more than one object with a nav-mesh component: https://github.com/MozillaReality/hubs-blender-exporter/issues/245

Issue #6308, opened 2023-10-03
bitECS: The central menu box when inspecting avatars is incorrect


Exairnous:
**Description**
When inspecting avatars in the new loader the generic object menu box is shown at the bottom center of the screen instead of the avatar specific ones.

**To Reproduce**
1. Enter a room
2. Press "I" to inspect your avatar.
3. See that the generic object menu box is shown at the bottom center of the screen.
4. Duplicate the tab.
5. Hover over the other avatar and press "F".
6. See that the generic object menu box is shown at the bottom center of the screen.

**Expected behavior**
The avatar specific menu boxes should be shown at the bottom center of the screen as in the current (Aframe) loader.

**Screenshots**
Current menu box for both:
![2023-10-03_01-14](https://github.com/mozilla/hubs/assets/9795121/544b6f25-d5b6-46c1-978a-26d51ea7b7ab)

Correct menu box for inspecting your avatar with "I":
![2023-09-15_16-46](https://github.com/mozilla/hubs/assets/9795121/3648b61a-8e15-4903-a0bc-b3bcbc55df08)

Correct menu box for focusing on someone else's avatar with "F":
![2023-07-24_13-01_1](https://github.com/mozilla/hubs/assets/9795121/fe50b360-c7bf-41cb-8d08-620bd6044d8e)

**Hardware** 
 - Device: Laptop
 - OS: Linux
 - Browser: Firefox and Chrome

keianhzo:
This should have been fixed by https://github.com/mozilla/hubs/pull/6311

Exairnous:
This issue was not addressed by #6311.  Can someone please reopen it?

keianhzo:
I misunderstood the issue, reopening.

Exairnous:
Thank you, and no worries!

Exairnous:
I can confirm this has been fixed on the new loader side, but I forgot to mention earlier that the behavior had changed for the aframe loader as well, since the initial separation of the aframe/new loader code (or at least since the crash fix) the aframe loader has used the "Hide" button instead of the "Edit Avatar" button when pressing I to inspect your avatar.  This probably isn't a priority since the behavior is correct in the new loader, but I just thought I should mention it.

PR #6328, opened 2023-10-13
Bitecs avatar menu fixes

keianhzo:
Fixes https://github.com/mozilla/hubs/issues/6308

This PR addresses three avatar inspect menu issues:
- Shows the correct avatar menu when inspecting instead of the object menu
- Fixes the avatar hide action that was crashing
- Hides the next/prev menu for both new and old loaders as it does nothing for avatars (same behavior for AFrame)


takahirox (approved):
Honestly I'm not so familiar with our react stuffs so I just checked the added code, they look good to me.

Issue #6304, opened 2023-10-02
bitECS: Objects are movable in the freeze mode


takahirox:
Reported by Imaginer on Discord

**Description**

With the new loader objects are movable even in the freeze mode while they are not with the old loader.

**To Reproduce**

**Expected behavior**

**Screenshots**

**Hardware** 
 - Device: Desktop
 - OS: Windows
 - Browser: Chrome

PR #6329, opened 2023-10-13
Stop sending/receiving network data if the scene is frozen

keianhzo:
Fixes https://github.com/mozilla/hubs/issues/6304

@johnshaughnessy Can you confirm that I'm not missing anything important here?


takahirox (approved):
I don't think I perfectly understand our networking stuffs and frozen mode, but skipping network receive system (that means not reflecting remote client behaviors) in the frozen mode sounds correct to me.

Curious to know, is also skipping network send system in frozen mode really necessary?

keianhzo:
This is actually much more complicated that I anticipated. I'm going to close this for now and open another one when I have a clearer idea.

Issue #6325, opened 2023-10-12
bitECS: Add a UI to toggle the new loader


takahirox:
**Is your feature request related to a problem? Please describe.**

Hopefully we will enable the new loader (bitECS implementation) by default soon.

We try to let the new loader have the compatibility with the old one as much as possible but no one can 100% guarantee it. Some users with their contents (eg: edge cases, complex scenes, or custom clients) might encounter new loader specific issues.

Update: Probably we ship the new loader code with disabling it by default and with a UI to toggle the new loader. And then we may think of enabling it by default at some point when we feel it will be robust enough.

**Describe the solution you'd like**

Add a way to opt-out the new loader for users who want to use the A-Frame beased implementation when we enable the new loader by default.

Note: We don't get rid of the old implementation immediately

**Additional context**

* How should the granularity of the opt-out controls be? Room level? Instance level?
* How should the API be?

takahirox:
Maybe configuration should be placed at

* If room level, RoomSettings via More
* If instance level, Admin panel

takahirox:
Room level new loader opt-in flag and `shouldUseNewLoader()` was added at https://github.com/mozilla/hubs/pull/6042 The flag is set with `fetchReticulumAuthenticated()` on the console + sign in. Also, `newLoader` URL query was added to force the new loader for testing.

```typescript
const forceNewLoader = qsTruthy("newLoader");
export function shouldUseNewLoader() {
  return forceNewLoader || APP.hub?.user_data?.hubsUseNewLoader;
}
```

Probably the easiest way to enable the new loader by default and add a way to opt-out the new loader is reusing them as much as possible. Changes for introducing room-level opt-out flag are

* Replace `newLoader` query with `oldLoader` URL query to force the old loader for testing
* Replace `APP.hub.user_data.hubsUseNewLoader` flag with `APP.hub.user_data.hubsUseOldLoader` flag to opt-out the new loader
* Update `shouldUseNewLoader` as the following

```typescript
const forceOldLoader = qsTruthy("oldoader");
export function shouldUseNewLoader() {
  return !forceOldLoader && !APP.hub?.user_data?.hubsUseOldLoader;
}
```

* Add "use the A-Frame based implementation" configulation in "More - Room Settings" for room owners.

takahirox:
As internally discussed, we plan to

* Ship the new loader code to demo hubs (master branch), hubs cloud, and managed instances.
* Add a UI to enable/disable the new loader #6342
[...]

PR #6337, opened 2023-10-18
Enable New Loader by default

takahirox:
Related #6325

This commit enables the new loader by default. Also, adds a way to opt-out the new loader.

Changes

* Enable the new loader by default
* Replace "newLoader" URL query with "oldLoader" to force to use the old A-Frame based implementation for testing
* Replace APP.hub.user_data.hubsUseNewLoader with APP.hub.user_data.hubsUseOldLoader to disable the new loader at room-level.
* Update shouldUseNewLoader() to use "oldLoader" and APP.hub.user_data.hubsUseOldLoader

TODO (This PR or another PR)

* Add a UI for room owners to control new loader opt-out flag
* Test (See https://github.com/mozilla/hubs/issues/6325#issuecomment-1767574573)

### DIFF
--- a/src/utils/bit-utils.ts
+++ b/src/utils/bit-utils.ts
@@ -60,7 +60,7 @@ export function findChildWithComponent(world: HubsWorld, component: Component, e
   }
 }
 
-const forceNewLoader = qsTruthy("newLoader");
+const forceOldLoader = qsTruthy("oldLoader");
 export function shouldUseNewLoader() {
-  return forceNewLoader || APP.hub?.user_data?.hubsUseNewLoader;
+  return !forceOldLoader && !APP.hub?.user_data?.hubsUseOldLoader;
 }


keianhzo:
Looks good to me. Are you planning to add the UI here on another PR?

I think it would also be good to reload all clients when that flag is updated by the admin. Do you know if that user_data update is triggering any event that we can use to reload all clients? Maybe we need to add it to reticulum. Probably for another PR.

takahirox:
Update: We don't merge this PR soon.

As we have been discussing internally, we changed a plan a little bit.

New plan

* We first ship the new loader code but we make if disabled by default. And we add a UI (maybe in a room-level setting) to toggle the new loader. Users who want to try the new loader can opt-in it by themselves
* Later at somepoint, we may think of enabling the new loader by default when we feel confident about the new loader. We keep the toggle UI. Users who encounter new loader issues can opt-out the new loader by themselves.

>  Are you planning to add the UI here on another PR?

Yes, I'm making another PR for adding the toggle UI. Once the toggle UI PR will be merged, we can ship the code. After the code shipment, when we will feel confident about the new loader we can enabled it by default by merging this PR.

> I think it would also be good to reload all clients when that flag is updated by the admin. Do you know if that user_data update is triggering any event that we can use to reload all clients? Maybe we need to add it to reticulum. Probably for another PR.

I'm thinking the same. Ideally all the clients in a room should reload the page when the new loader is toggled. But I don't really want to change the reticulum code now in this very tight schedule. I'm checking through the client and reticlum code to try to find API or systems that triggers reload the page on a certain event and we can reuse. Probably the client reload PR would be another PR from the toggle UI PR.

In short, we need two or three PRs related to new loader enable/disable.

By the initial new loader release

1. Add a new UI to toggle the new loader #6325
2. (Optional?): Reload all the clients in a room when the new loader is toggled

After the initial release, by when we will enable the new loader by default

3. Enable the new loader by default (This PR)

takahirox:
Regarding this

> (Optional?): Reload all the clients in a room when the new loader is toggled

If I'm right it seems to be possible in Hubs Phoenix channel "hub_refresh" event listener in src/hub.js because

1. Client pushes "update_hub" to Reticulum when room settings are updated
2. Reticulum broadcasts "hub_refresh" to all the clients in the room

So Hubs Client can check if new loader activation flag is changed and invoke the page reload if it happens.

I will make a PR...

takahirox:
We have released the initial new bitECS based client that is disabled by default. We will think of enabling it by default when we will feel confident about the new client stability. 

Closing this PR for now because it's already old. Let's make a new one when we will enable it by default.

Issue #6331, opened 2023-10-14
bitECS: Invalid gltf parameters can crash the client


takahirox:
Reported by Imaginer @Exairnous on Discord.

**Description**

If you drop in a glb file that includes invalid glTF (Hubs component) parameters, Hubs Client can crash with the new loader enabled. The client doesn't crash with the old loader.

For example if you drop in this file [import_dev_test_3.zip](https://github.com/mozilla/hubs/files/12900614/import_dev_test_3.zip) the client crashes with the new loader.

These are the console errors.

![image](https://github.com/mozilla/hubs/assets/7637832/515976b5-2a38-47e5-9613-61121480251f)

With the old loader, the loading cube never ends and some console errors, pasted below, happen but the client doesn't crash.

![image](https://github.com/mozilla/hubs/assets/7637832/f88d691a-97a4-451f-a6a0-e1aff5796dd8)

**To Reproduce**
1. Enter a room with the new loader enabled
2. Drop in the example file above
3. Confirm the client crash
4. See the console error

**Expected behavior**

No crash

**Hardware** 
 - Device: Desktop
 - OS: Windows
 - Browser: Chrome

**Additional context**

The problem is very related to how to handle invalid glTF (Hubs component) parameters that we have been discussing. glTF file can come from outside so we can't and shouldn't really trust glTF data. It may include invalid parameters.

Some solution ideas.

1. Adding a validation layer to glTF inflators. It checks inflator parameters and refuse or rescue invalid parameters.
2. Add more try - catch in systems for the code that can throw exception.

We may start to add try - catch in audio target system that seems to cause the crash with the example glb file for now if our time is very limited.

3. An easy temporary workaround, add try-catch to the top-level tick?

Exairnous:
This should probably be handled on the Blender exporter side as well, so I've opened an issue to track it.  Curiously, the Min Delay parameter doesn't cause a crash, but I guess it should also be constrained to the same range?

takahirox:
Yes, it's good to check parameters both in the loader (client) and exporter.

keianhzo:
We should be able to enforce min, max, default in the inflator params definition but for now I think we can just enforce limits in the inflators explicitly.

PR #6341, opened 2023-10-19
Clamp component parameters to make the client more resilient

keianhzo:
This PR enforces component parameter limits for structures that otherwise would crash.

Fixes https://github.com/mozilla/hubs/issues/6331

### DIFF
--- a/src/bit-systems/particle-emitter.ts
+++ b/src/bit-systems/particle-emitter.ts
@@ -11,18 +11,18 @@ import { resolveUrl, textureLoader } from "../utils/media-utils";
 function* setTexture(world: HubsWorld, eid: number) {
   let src = APP.sid2str.get(ParticleEmitterTag.src[eid]);
 
-  const result: URL = yield resolveUrl(src);
-  let canonicalUrl = result.origin;
-
-  // handle protocol relative urls
-  if (canonicalUrl.startsWith("//")) {
-    canonicalUrl = location.protocol + canonicalUrl;
-  }
-
-  // todo: we don't need to proxy for many things if the canonical URL has permissive CORS headers
-  src = proxiedUrlFor(canonicalUrl);
-
   try {
+    const result: URL = yield resolveUrl(src);
+    let canonicalUrl = result.origin;
+
+    // handle protocol relative urls
+    if (canonicalUrl.startsWith("//")) {
+      canonicalUrl = location.protocol + canonicalUrl;
+    }
+
+    // todo: we don't need to proxy for many things if the canonical URL has permissive CORS headers
+    src = proxiedUrlFor(canonicalUrl);
+
     const texture: Texture = yield textureLoader.loadAsync(src);
 
     const particleEmitter: ParticleEmitter = world.eid2obj.get(eid)! as ParticleEmitter;
--- a/src/inflators/audio-target.ts
+++ b/src/inflators/audio-target.ts
@@ -25,8 +25,11 @@ export function inflateAudioTarget(world: HubsWorld, eid: number, params: AudioT
 
   addComponent(world, AudioTarget, eid);
 
-  AudioTarget.maxDelay[eid] = params.maxDelay;
-  AudioTarget.maxDelay[eid] = params.maxDelay;
+  // https://developer.mozilla.org/en-US/docs/Web/API/BaseAudioContext/createDelay#maxdelaytime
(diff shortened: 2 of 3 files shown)


src/bit-systems/particle-emitter.ts
```diff
@@ -11,18 +11,18 @@ import { resolveUrl, textureLoader } from "../utils/media-utils";
...
-  }
+    // handle protocol relative urls
+    if (canonicalUrl.startsWith("//")) {
+      canonicalUrl = location.protocol + canonicalUrl;
+    }
 
-  // todo: we don't need to proxy for many things if the canonical URL has permissive CORS headers
-  src = proxiedUrlFor(canonicalUrl);
+    // todo: we don't need to proxy for many things if the canonical URL has permissive CORS headers
+    src = proxiedUrlFor(canonicalUrl);
```
takahirox, line 24: Are these changes in this system for a case where resolving URL fails?

src/inflators/audio-target.ts
```diff
@@ -25,8 +25,11 @@ export function inflateAudioTarget(world: HubsWorld, eid: number, params: AudioT
...
 
-  AudioTarget.maxDelay[eid] = params.maxDelay;
-  AudioTarget.maxDelay[eid] = params.maxDelay;
+  // https://developer.mozilla.org/en-US/docs/Web/API/BaseAudioContext/createDelay#maxdelaytime
+  params.maxDelay = params.maxDelay < 180 ? params.maxDelay : 179;
+  AudioTarget.maxDelay[eid] = params.maxDelay < 0 ? 0 : params.maxDelay;
+  params.minDelay = params.minDelay > AudioTarget.maxDelay[eid] ? AudioTarget.maxDelay[eid] : params.minDelay;
+  AudioTarget.minDelay[eid] = params.minDelay < 0 ? 0 : params.minDelay;
   AudioTarget.source[eid] = params.srcNode;
   if (params.debug) AudioTarget.flags[eid] |= AUDIO_TARGET_FLAGS.DEBUG;
```
takahirox, line 34: Out of scope from this PR but `AudioTarget.flags[eid]` must be always initialized. Otherwise it can be problematic when entity is recycled.

takahirox (approved):
As I wrote in #6331 ideally the root issue should be fixed by adding validation layer for glTF inflators, looking through all the systems and adding try-catch block to all the places where exception can heppen, or adding try-catch block to the top-level system. But it would ok to add clamping and try-catch to the places where we got problem report like this PR for now due to our limited time.

keianhzo:
@takahirox Yeah that's what I also mentioned in https://github.com/mozilla/hubs/issues/6331#issuecomment-1771429801 A-Frame had a schema system https://aframe.io/docs/1.4.0/core/component.html#schema to validate and parse properties. I think we could have a similar system. Something like:

```typescript
export const gltfInflators: Required<{ [K in keyof GLTFComponentData]: InflatorFn }> = {
  ...commonInflators,
  ...
  ammoShape: { 
   inflator: inflateAmmoShape,
   schema: schema: {
    type: {
      default: SHAPE.HULL,
      oneOf: [
        SHAPE.BOX,
        SHAPE.CYLINDER,
        SHAPE.SPHERE,
        SHAPE.CAPSULE,
        SHAPE.CONE,
        SHAPE.HULL,
        SHAPE.HACD,
        SHAPE.VHACD,
        SHAPE.MESH,
        SHAPE.HEIGHTFIELD
      ]
    },
    fit: { default: FIT.ALL, oneOf: [FIT.ALL, FIT.MANUAL] },
    halfExtents: { type: "vec3", default: { x: 1, y: 1, z: 1 } },
    minHalfExtent: { default: 0 },
    maxHalfExtent: { default: Number.POSITIVE_INFINITY },
    sphereRadius: { default: NaN },
    cylinderAxis: { default: "y", oneOf: ["x", "y", "z"] },
    margin: { default: 0.01 },
    offset: { type: "vec3", default: { x: 0, y: 0, z: 0 } },
    orientation: { type: "vec4", default: { x: 0, y: 0, z: 0, w: 1 } },
    heightfieldData: { default: [] },
    heightfieldDistance: { default: 1 },
    includeInvisible: { default: false }
  },
  ...
};
```
and we could enforce that during: 
https://github.com/mozilla/hubs/blob/5711df6cdbc3bfef2cce9d7cf9525b847c0ee1e7/src/inflators/model.tsx#L27

Issue #1784, opened 2019-10-02
Unclear icon in object menu


emclaren:
**Description**
Many people have expressed confusion about what the meaning of the icon in the Object menu that makes the object fall to the floor. Perhaps it would help if the arrow was more obvious.

**To Reproduce**
1. Go to Hubs Room
2. Create an object
3. Open up the object menu

**Expected behavior**
NA

**Screenshots**
This icon
<img width="102" alt="Screen Shot 2019-10-02 at 1 50 46 PM" src="https://user-images.githubusercontent.com/6835324/66081160-51755600-e51c-11e9-951c-936dd23f1104.png">

**Hardware** 
 All

**Additional context**
This observation is based on feedback gathered during the Hubs office hours.

Shivansh2407:
How can I contribute to this issue?

misslivirose:
We will need a new icon proposal and design - I think @johnshaughnessy might be able to speak to some of our iconography, as he has done quite a bit of work with our sprites

SparshKat:
I'd love to contribute to this issue. @johnshaughnessy please do let me know if you have any ideas of design or something. Please let me know :)

matthewbcool:
Welcome proposals in the discord #dev-design channel: https://discord.gg/sBMqSjCndj

This icon can be edited via the PNG here: https://github.com/mozilla/hubs/blob/master/src/assets/images/spritesheets/sprite-system-action-spritesheet.png

https://user-images.githubusercontent.com/4493657/159348284-11541158-e9a6-494f-8a51-b053b5d7d80f.mp4

PR #6350, opened 2023-10-23
#1784

Aniraai:
update the size of the drop action emoji


mikemorran:
@Aniraai can you share any screen shots of the desired visual output?

Issue #6332, opened 2023-10-14
bitECS: object menu buttons can't be clicked if objects are overlapped


takahirox:
Reported by @j-conrad on Discord

**Description**

With the new loader enabled, object menu buttons can't be clicked (or it's hard) if objects are overlapped because a cursor while on its way to a button, can roll over an object behind the intended one and the menu buttons turn off for the first object. See this video for the details.

https://github.com/mozilla/hubs/assets/7637832/0af3828f-9729-4b55-bb1a-6b52c86ab957

**To Reproduce**
1. Enter a room with new loader enabled
2. Place two objects simiar to the ones in the video above
3. Hover a cursor on the nearer object and hold the space bar to show the object menu
4. Move the cursor to an object menu button and hover the cursor to the further one
5. Confirm that the object menu for the nearer one disappears and the object menu for the further one appears so that you can't click the button for the nearer one

**Expected behavior**

Easy to click an intended button even if objects are overlapped

**Hardware** 
 - Device: Desktop
 - OS: Windows
 - Browser: Chrome

**Additional context**

At least we should fix the case where an intended buton can't be clicked.

takahirox:
Also a related issue was reported by Imaginer @Exairnous 

> Also, the spacebar menu will appear behind you if you are within or close to an object (this is especially apparent on objects that have been scaled large)

Exairnous:
@takahirox Do you want me to report that related issue about being too close as a separate GitHub issue?

PR #6351, opened 2023-10-23
Add a raycastable menu background

keianhzo:
Fixes https://github.com/mozilla/hubs/issues/6332

This PR adds a raycastable plane to the object menu to keep the focus on the current object.

This should only happen with the object menu as the rest of the menus are always displayed on a plane.


takahirox:
> adds a raycastable plane to the object menu to keep the focus on the current object

Yes, I think it's a good idea. Let me review the implementation.

takahirox (approved):
I haven't tested locally, but the change looks good to me.

PR #6354, opened 2023-10-25
Add new loader toggle UI in room settings

### LINKED ISSUE #6325: bitECS: Add a UI to toggle the new loader

takahirox:
Fixes #6325 

This commit adds new loader toggle UI in room settings that allows room owners to enable/disable the new loader.

Changes
* Rename `hub.user_data.hubsUseNewLoader` to `hub.user_data.hubs_use_new_loader` for naming consistency
* Add new loader toggle UI in room settings

image

TODOs in another PR if needed

* Refresh the page in all the clients in the room when toggled

### DIFF
--- a/src/react-components/room/RoomSettingsSidebar.js
+++ b/src/react-components/room/RoomSettingsSidebar.js
@@ -196,6 +196,18 @@ export function RoomSettingsSidebar({
             />
           </div>
         </InputField>
+        <InputField
+          label={<FormattedMessage id="room-settings-sidebar.new-loader" defaultMessage="New loader activation" />}
+          fullWidth
+        >
+          {/* TODO: Refresh the page in all the clients in the room when toggled */}
+          <ToggleInput
+            label={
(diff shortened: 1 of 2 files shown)


nickgrato:
other than some of the linter errors, the code seems fine to me. did have one question though, did you want this toggle to have a blue border like that? seems like the others don't

takahirox:
> did you want this toggle to have a blue border like that? seems like the others don't

The toggles have blur border when they are focused. 

image

PR #6355, opened 2023-10-26
Reload when the new loader is toggled

takahirox:
From: https://github.com/mozilla/hubs/pull/6337#issuecomment-1780126424

This commit forces all the clients to reload the page when the new loader activation flag is toggled.

Changes
* In Hubs Channels hub_refresh listener detect user_data.hubs_use_new_loader flag change
* Reload the page in five seconds if detected
* Add and show new loader refresh prompt when detected

image

### DIFF
--- a/src/assets/stylesheets/ui-root.scss
+++ b/src/assets/stylesheets/ui-root.scss
@@ -51,3 +51,18 @@ body.vr-mode {
   z-index: 11;
   pointer-events: auto;
 }
+
+:local(.new-loader-refresh-prompt) {
+  @extend %default-font;
+  background: theme.$background1-color;
+  border: 3px solid theme.$border1-color;
+  border-radius: 16px;
+  color: theme.$text1-color;
+  left: 50%;
+  max-width: 70%;
+  padding: 20px;
+  position: absolute;
+  top: 50%;
+  transform: translate(-50%, -50%);
+  z-index: 100;
+}
--- a/src/hub.js
+++ b/src/hub.js
@@ -531,8 +531,8 @@ export async function updateEnvironmentForHub(hub, entryManager) {
   }
 }
 
-export async function updateUIForHub(hub, hubChannel) {
-  remountUI({ hub, entryDisallowed: !hubChannel.canEnterRoom(hub) });
+export async function updateUIForHub(hub, hubChannel, showNewLoaderRefreshPrompt = false) {
+  remountUI({ hub, entryDisallowed: !hubChannel.canEnterRoom(hub), showNewLoaderRefreshPrompt });
 }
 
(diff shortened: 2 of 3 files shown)


src/assets/stylesheets/ui-root.scss
```diff
@@ -51,3 +51,18 @@ body.vr-mode {
...
+  border-radius: 16px;
+  color: theme.$text1-color;
+  left: 50%;
+  max-width: 70%;
+  padding: 20px;
+  position: absolute;
+  top: 50%;
+  transform: translate(-50%, -50%);
+  z-index: 100;
+}
```
takahirox, line 68: I'm not a designer. Comments to improve these are very welcome.

src/hub.js
```diff
@@ -531,8 +531,8 @@ export async function updateEnvironmentForHub(hub, entryManager) {
   }
 }
 
-export async function updateUIForHub(hub, hubChannel) {
-  remountUI({ hub, entryDisallowed: !hubChannel.canEnterRoom(hub) });
+export async function updateUIForHub(hub, hubChannel, showNewLoaderRefreshPrompt = false) {
+  remountUI({ hub, entryDisallowed: !hubChannel.canEnterRoom(hub), showNewLoaderRefreshPrompt });
```
takahirox, line 535: Honestly I don't rellay like to add `shoeNewLoaderRefreshPrompt` param here. Any suggestion for simpler approach is very welcome.

src/assets/stylesheets/ui-root.scss
nickgrato: this looks good to me. we also have toast component in the lilypad component library and a modal if you are interested..

https://mozilla.github.io/lilypad/?path=/story/example-notification--main

https://mozilla.github.io/lilypad/?path=/story/example-modal--main

but again this is probably fine.

src/hub.js
nickgrato: Yea I see what you are saying. Since we don't have a global state management tool ( that I know of here ie redux)  this might be what you have to do...

One thing you could do to make it better though is to broaden the name a little more.. instead of making it specific to showing a new loader you could just make it a boolean called pageWillRefresh so if you need it for another reason in the future you could use it.

I might also change the loader references from "new", new is only new until it's not anymore, and then the newnewloader comes :)

src/assets/stylesheets/ui-root.scss
takahirox: Thanks I'm thinking of merging as is for now.

src/hub.js
takahirox: Yeah, I started to feel like renaming "New Loader". I want to merge this PR as is for now and make another PR for renaming.

Issue #6360, opened 2023-11-06
bitECS: Crashes if open object menu on link objects


takahirox:
**Description**

Hubs Client with the new bitECS based implementation crashes if you attempt to open object menu on link objects.

You can use Link.glb for testing that can be downloaded at

https://github.com/mozilla/hubs-sample-assets/tree/main/Hubs%20Components/Exported%20GLB%20Models

![image](https://github.com/mozilla/hubs/assets/7637832/787ecbd4-c640-4663-9d14-6fb75dcdbe6f)

![image](https://github.com/mozilla/hubs/assets/7637832/c043574d-5d57-4608-8c04-1e7fbd95632a)

**To Reproduce**
1. Go to hubs.mozilla.com
2. Create and enter a room
3. Enable the new bitECS based Client via Room Settings
4. Upload Link.glb
5. Hover the cursor on the object
6. Press the spacebar to open object menu
7. Confirm the client crashes

**Expected behavior**

No crash and object menu is shown

**Hardware** 
 - Device: Desktop
 - OS: Windows
 - Browser: Chrome

PR #6363, opened 2023-11-08
Fix spawned media menu target resolution

keianhzo:
We were looking for the object menu target using the `ObjectMenuTarget` and `MediaContentBounds` components but more than one entity in the hierarchy can have those (spawned glTF that spawns media) but that's not the object the object menu actions (like pin) need to work on.

This PR updates the root object menu target lookup by looking for the `Deletable` entity that also has an `ObjectMenuTarget`component.

We should probbaly try to find a better way of tagging this root entity that's more verbose. Something like SpawnedMediaRoot.

Fixes https://github.com/mozilla/hubs/issues/6360

### DIFF
--- a/src/bit-systems/object-menu.ts
+++ b/src/bit-systems/object-menu.ts
@@ -12,8 +12,8 @@ import {
   ObjectMenuTarget,
   RemoteRight,
   Rigidbody,
-  MediaContentBounds,
-  Deleting
+  Deleting,
+  Deletable
 } from "../bit-components";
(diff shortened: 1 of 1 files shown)


takahirox (approved):
Honesly I'm not really familiar with `Deletable` comonent so hard to say whether this workaround is good, but I think at least it should be much better than crash. So, let's go with this change for now.

> We should probbaly try to find a better way of tagging this root entity that's more verbose. Something like SpawnedMediaRoot.

Agree with it. Let's update after the initial release.

PR #6365, opened 2023-11-09
Fix Browse button layout Round 2

takahirox:
A follow up PR to #6361

**Problem**

Browse button in Custom Object upload dialog displays weirdly in portrait mode on iOS. It unexpectedly covers almost the entire URL area in portrait mode.

We applied a change #6361 to fix but it didn't work.

**Changes**

* Remove flax stuffs and width 60px from browse style
* Add justify-content: end to IconButton style

### DIFF
--- a/src/react-components/input/IconButton.scss
+++ b/src/react-components/input/IconButton.scss
@@ -9,6 +9,7 @@
   font-size: theme.$font-size-xs;
   font-weight: theme.$font-weight-bold;
   cursor: pointer;
+  justify-content: end;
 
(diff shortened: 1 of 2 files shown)


keianhzo (approved):

takahirox:
@keianhzo Thanks for reviewing. Note that we may not merge this PR because we found that `IconButton` is referred from multiple places and we are not sure if the change in the PR is ok everywhere. @nickgrato is seeking a better solution. Please refer to our private Discord channel.

PR #6369, opened 2023-11-15
Update hub-channel.js

Fi1osof:
Fix canEnterRoom method.
Should be <= instead <, cause roomEntrySlotCount includes current user. Last user can not enter.


keianhzo:
I think is working correctly for what I can see. `roomEntrySlotCount` is the number of currently in-room clients and it has to be smaller than the room max clients number in `hub.room_size` otherwise we would allow `hub.room_size` + 1 clients connected. Let me know if I missed anything.

PR #6377, opened 2023-11-17
Media frame object alignment and scale

rawnsley:
This PR introduces greater control over how objects are snapped into media frames including:

1. The option to keep the object's current scale.
2. The option to align with the axis boundaries when the object does not exactly fill the space.

The canonical example would be a board game like chess with irregular sized pieces. Large media frames obscure the board making it hard to place pieces, resizing the pieces to fit the frame will make pawns the same size as kings, and centering the objects within the frame will make some of them float above the board. While it is possible to edit the objects themselves to "hack" around these limitations, it is complicated and not entirely solvable.

Screenshot 2023-11-17 at 15 41 24

This PR expects extra parameters from [the companion Blender exporter add-on PR](https://github.com/MozillaReality/hubs-blender-exporter/pull/248).

The media-frame code was slightly refactored for clarity, including the addition postfixes to some variables indicating their type (obj, eid, etc...). The code has different paths for video objects and for "new loader" imports, all of which were considered.

A test scene was created to demonstrate and explore the various combinations of situations, which can be accessed here:

- [Scene in classic Hubs](https://hubs.mozilla.com/scenes/qgaMVT9/mediaframealignment)
- [GLTF file](https://avnfs.com/4lN8mk1QV5kOLe3drNyysesvWk-sLwoz2WKL2bAbhUo?size=9583748&type=model%2Fgltf-binary%3Bdisp%3Dscene&name=media-frame-alignment.glb)
- [Blender file](https://avnfs.com/M5u6y7OEHBREx2jadvfwNyQlp_Y4zEMJT3NQx4ZANOI?size=10354924&type=application%2Fx-blender&name=media-frame-alignment.blend)

The best way to understand this PR is probably to grab the code and load this scene. Some text is included to explain the intention of each section. Some of it is designed to test edge cases such as scaled frames, frames on spawned objects, etc...

### DIFF
--- a/src/bit-components.js
+++ b/src/bit-components.js
@@ -34,7 +34,7 @@ export const MediaFrame = defineComponent({
   guide: Types.eid,
   preview: Types.eid,
   previewingNid: Types.eid,
-  flags: Types.ui8,
+  flags: Types.ui8
 });
 export const TextTag = defineComponent();
 export const ReflectionProbe = defineComponent();
--- a/src/bit-components.js
+++ b/src/bit-components.js
@@ -30,9 +30,11 @@ export const MediaFrame = defineComponent({
   scale: [Types.f32, 3],
   mediaType: Types.ui8,
   bounds: [Types.f32, 3],
+  align: [Types.ui8, 3],
   guide: Types.eid,
   preview: Types.eid,
-  previewingNid: Types.eid
+  previewingNid: Types.eid,
+  flags: Types.ui8,
 });
 export const TextTag = defineComponent();
 export const ReflectionProbe = defineComponent();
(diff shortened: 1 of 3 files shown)


keianhzo (changes requested):
This is a great update! Overall looks good to me. There have been some updates in media-loading recently that affect this PR (ie. spawned media video scaling) but I think its ok to land it as is and then I can push a patch to update that. 

There are a few linter errors that need to be fixed before landing this. You can run `npm run lint` to spot them. When that's done I'll merge this and merge my updates.

rawnsley:
@keianhzo I've pushed the linter fixes now and thanks taking on the merge with your other changes - much appreciated.

keianhzo (approved):

Issue #6338, opened 2023-10-19
bitECS: image + controls movability and object menu visibility


takahirox:
Reported by Imaginer @Exairnous on Discord

**Description**

The behaviors of Hubs image component + controls are different between with and without the new loader.

When dropping-in the asset,
- Without the new loader
  - Not movable
  - No object menu on hover + spacebar
- With the new loader
  - Movable
  - Object menu on hover + spacebar

As scene object
- Without the new loader
  - Object menu on hover + spacebar
- With the new loader
  - No object menu on hover + spacebar

Testing asset from Imaginer [image_with_controls.zip](https://github.com/mozilla/hubs/files/13035125/image_with_controls.zip)
And a room for testing scene object https://hubs.mozilla.com/qm8bAh6/bleak-loyal-meetup (Left with controls, Right without controls)

**To Reproduce**

For drop-in object
1. Enter a room with the new loader enabled
2. Upload the example asset
3. Try to move the object and confirm it moves
4. Hover the cursor on the object, hold the space bar, and confirm that object menu appears
5. Try the steps above without the new loader enabled, and confirm that the behaviors are different

For scene object
1. Enter https://hubs.mozilla.com/qm8bAh6/bleak-loyal-meetup with the new loader enabled
2. Hover the cursor on the left object and press the space bar
3. Confirm that no object menu appears
4. Try the steps above without the new loader enabled, and confirm that object menu appears

**Expected behavior**

Ideally the behaviors are the same between with and without the new loader

**Hardware** 
 - Device: Desktop
 - OS: Windows
 - Browser: Chrome

Exairnous:
> with the new loader the behavior for drop-in seems correct while it doesn't seem correct for scene object.

Agreed.

Exairnous:
Also note that when hovering over the scene object the cursor doesn't turn blue in the new loader like it does in the old loader.

keianhzo:
Thanks for the feedback @Exairnous 

## New Loader
https://hubs.mozilla.com/idbvNmZ/loud-grounded-nation/
### Scene element
| Object  | Can Be Pinned | Object Menu |  Link Menu 
| ------------- | ------------- | ------------- | -------------
| Image + Control | N/A | N/A  | NO (BUG, we should show it)
| Image + No Control | N/A | N/A  | NO
| Image + Control + Link | N/A | N/A  | YES
| Image + No Control + Link | N/A | N/A  | YES (link overrides no controls)
### Drop-in glTF
[...]

PR #6393, opened 2023-11-30
Look for the holdable media root when grabbing an object

keianhzo:
When grabbing an object we want to look for the root media loader as that's the object that have a networked transform.

Related https://github.com/mozilla/hubs/issues/6338

### DIFF
--- a/src/utils/load-image.tsx
+++ b/src/utils/load-image.tsx
@@ -60,5 +60,13 @@ export function* loadImage(world: HubsWorld, eid: EntityID, url: string, content
 
   ObjectMenuTarget.flags[eid] |= ObjectMenuTargetFlags.Flat;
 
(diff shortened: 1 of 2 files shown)


src/utils/load-image.tsx
```diff
@@ -60,5 +60,13 @@ export function* loadImage(world: HubsWorld, eid: EntityID, url: string, content
...
   ObjectMenuTarget.flags[eid] |= ObjectMenuTargetFlags.Flat;
 
-  return renderAsEntity(world, <entity name="Image" image={imageDef} objectMenuTarget={{ isFlat: true }} />);
+  return renderAsEntity(
+    world,
+    <entity
+      name="Image"
+      image={imageDef}
+      objectMenuTarget={{ isFlat: true }}
+      grabbable={{ cursor: true, hand: false }}
```
Exairnous, line 69: This sounds like it's only grabbable for 2D users and not VR users.  Is this intentional?

Issue #6219, opened 2023-08-15
bitECS: Improve video controls


takahirox:
**Is your feature request related to a problem? Please describe.**

Video controls can improve with new loader. Especially it lacks of some features compared to the one without new loader, for example

* Play/Pause indicator
* Volume indicator/controls
* Forward/Backward

With new loader

![image](https://github.com/mozilla/hubs/assets/7637832/467ef4c5-8877-472e-bd82-d734672c049d)

Without new loader

![image](https://github.com/mozilla/hubs/assets/7637832/642782ef-7227-4573-8994-a6cb467bbac0)

keianhzo:
BitECS video controls has now parity with the old loader one. Only the skip button is missing but as we have now the time slider, it doesn't seem necessary to me to have sip buttons.

PR #6394, opened 2023-11-30
Update video menu layout

keianhzo:
Update the video menu to avoid the input blocking by the invisible buttons and also make it more consistent witht he rest of the application UI.


Exairnous:
Nice.  This addresses part of https://github.com/mozilla/hubs/issues/6219

PR #6411, opened 2023-12-06
Changing Rooms Feature - Behavior Graph

hrithikwins:
Taco tuesday/changing rooms portal

In Taco tuesday we experimented on changing rooms directly 
image


Exairnous:
This seems to work pretty well, but we weren't able to test specifying `#waypoint`spawns and we didn't know which category to put it in.

hrithikwins (commented):
LGTM

keianhzo:
@hrithikwins @Exairnous This LGTM but I don't see the related BGs add-on PR, is there one?

Exairnous:
Thanks.  No, we haven't put one up yet.  I thought it a better idea to get your opinion on this first and then if everything was good we could worry about the add-on PR (it's only modifying the nodespec.json).  I'll see about opening one soon.

keianhzo (approved):
The reason why the waypoint does not work is because this branch has not yet been rebased with the latest main branch changes. It should work when that's merged.

A good follow-up for this would be to add support to move to a waypoint. You can see how that could be done here: https://github.com/hrithikwins/hubs/blob/d7b39899ad3369591c14c41aead30248dd8cee08/src/bit-systems/link-hover-menu.ts#L75-L78 This will also **not** work until the branch is rebased with the main branch.

PR #6428, opened 2023-12-21
Full-body avatar

hunjuly:
GLBs for test 
https://github.com/belivvr/files/raw/main/male.glb
https://github.com/belivvr/files/raw/main/female.glb


mikemorran:
Hey @hunjuly, likely for this feature to be added, we will have to rewrite in the bitECS system to ensure reliability moving forward. Have you explored bitECS yet?

hunjuly:
“The bitECS-based Hubs client is still under development and stability is not fully guaranteed. Therefore, the bitECS based client is disabled by default. We plan to enable the bitECS based client by default after safety has been sufficiently confirmed.” 
I saw that, but if they're making it bitECS-based now, doesn't that mean that full-body won't be available for A-frame-based clients until they change the default for rooms?
I'm wondering if it's mandatory to use BitECS, which is still in development, to use new features like full-body.

mikemorran:
> “The bitECS-based Hubs client is still under development and stability is not fully guaranteed. Therefore, the bitECS based client is disabled by default. We plan to enable the bitECS based client by default after safety has been sufficiently confirmed.” I saw that, but if they're making it bitECS-based now, doesn't that mean that full-body won't be available for A-frame-based clients until they change the default for rooms? I'm wondering if it's mandatory to use BitECS, which is still in development, to use new features like full-body.

All of our systems and components, with a few small exceptions, are rewritten in bitECS in the new client and the feature flag toggles between those new and old versions. We aren't fully removing AFRAME, so it looks like your PR would run (although I have not looked in depth), however for maintainability we recommend bitECS.

worph:
We really want this feature at. What could help push this feature into the master. Another pull request compatible with bitECS?

hunjuly:
Well, I don't plan to add the feature until hubs is stable enough to make bitECS the default.
My team has other, more important things to do.

PR #6430, opened 2024-01-09
Keep original object when swapping

keianhzo:
When an object is swapped we are not setting a name but some components, like the waypoint component, rely on this name.


keianhzo:
In the case of spawn points, this is only the case for waypoints as the replace the entity so maybe that's the expected behavior. Closing for now...

PR #6494, opened 2024-08-10
[Mozilla Branding Removal] tex's first try on contributing to an open source project

tksham:
update locale and copy regarding Powered by "Mozilla Hub"

### DIFF
--- a/src/react-components/layout/Footer.js
+++ b/src/react-components/layout/Footer.js
@@ -25,11 +25,11 @@ export function Footer({
           {!hidePoweredBy && (
             <FormattedMessage
               id="footer.powered-by"
-              defaultMessage="Powered by <a>Mozilla Hubs</a>"
+              defaultMessage="Powered by <a>Hubs</a>"
               values={{
                 // eslint-disable-next-line react/display-name
                 a: chunks => (
-                  <a className={styles.link} href="https://hubs.mozilla.com">
+                  <a className={styles.link} href="https://demo.hubsfoundation.org">
                     {chunks}
                   </a>
                 )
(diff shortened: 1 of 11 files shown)


src/react-components/layout/Footer.js
```diff
@@ -25,11 +25,11 @@ export function Footer({
           {!hidePoweredBy && (
             <FormattedMessage
               id="footer.powered-by"
-              defaultMessage="Powered by <a>Mozilla Hubs</a>"
+              defaultMessage="Powered by <a>Hubs</a>"
               values={{
                 // eslint-disable-next-line react/display-name
                 a: chunks => (
-                  <a className={styles.link} href="https://hubs.mozilla.com">
+                  <a className={styles.link} href="https://demo.hubsfoundation.org">
```
Exairnous, line 32: `demo.hubsfoundation.org` will likely link to the official instance for the Hubs Foundation, but I think it probably makes more sense to link to the main website here instead, so I would just link to `hubsfoundation.org`

Exairnous (changes requested):
Thank you for the PR!   This looks mostly good, I left one suggestion in an inline comment.
It looks like these files could also use Mozilla references removed, though:
```
src/assets/locales/fr.json
src/assets/locales/ja.json
src/assets/locales/pt.json
src/assets/locales/ru.json
src/assets/locales/ca.json
```

src/react-components/layout/Footer.js
tksham: I just made another commit based on you comments, do I have to create another pull request or it would be fine?

Exairnous: Nope, this is good.  Thanks!

Exairnous (approved):
LGTM.  Thanks!  Merging.

PR #6501, opened 2024-09-24
Fix bug: Spacebar Menu Scale and Rotate Buttons Do Nothing When Interacted With

### LINKED ISSUE #6496: Spacebar Menu Scale and Rotate Buttons Do Nothing When Interacted  With
**Description**
Objects can't be manipulated via the scale and rotate buttons from the spacebar menu anymore.

**To Reproduce**
1. Enter a Hubs room.
2. Spawn a duck with the /duck chat command.
3. Hover over the duck and press and hold the spacebar to bring up the menu.
4. Click and drag on the scale or rotate buttons while holding the spacebar.
5. See that the duck moves instead of scaling or rotating.

**Expected behavior**
The object should be scaled or rotated, depending on which button is interacted with.

**Video**
https://github.com/user-attachments/assets/ed28a602-5af2-4ff6-aba6-6bf401f1972e
Video courtesy of mr. lion on Discord.

**Hardware** 
 - Device: Laptop
 - OS: GNU/Linux
 - Browser: Chromium

**Additional context**
~~Potentially caused by #6495~~
Potentially caused by #6487
~~Physics also seems to be affected by this bug.  Objects can no longer be thrown and the gravity/drop to floor button on the spacebar button does nothing if the object has been moved.~~  Update: the physics issues are probably a separate bug and have been reported in #6503, #6504, #6505

Exairnous:
Revert 699187a537b30d2b67dbe059d44311ceaebcefa0 and part of 1567f224c1db251e9540626211e68b4dea06b32e

Fixes #6496

Chromium version 128 introduced a bug in Hubs that was fixed by the #6495 PR but #6495 seems to have introduced the bug detailed in issue #6496 for all browsers. Now Chromium has fixed an issue on their end as detailed here: https://issues.chromium.org/issues/361823993 which should fix the original bug that PR #6495 addressed in Hubs, which allows us to revert the fix on the Hubs side (PR #6495) and so fix the #6496 bug.  However, this doesn't do anything for the potentially deeper issues in the system that were possibly exposed with PR #6495.

### DIFF
--- a/src/components/hoverable-visuals.js
+++ b/src/components/hoverable-visuals.js
@@ -1,19 +1,7 @@
 import { showHoverEffect } from "../utils/permissions-utils";
 
-// prettier-ignore
-const interactorOneTransform = [
-  1, 0, 0, 0,
-  0, 1, 0, 0,
-  0, 0, 1, 0,
-  0, 0, 0, 1
-];
-// prettier-ignore
-const interactorTwoTransform = [
-  1, 0, 0, 0,
-  0, 1, 0, 0,
-  0, 0, 1, 0,
-  0, 0, 0, 1
-];
+const interactorOneTransform = [];
+const interactorTwoTransform = [];
 
 export const validMaterials = ["MeshStandardMaterial", "MeshBasicMaterial", "MeshPhongMaterial"];
 /**


Exairnous:
Turns out https://github.com/Hubs-Foundation/hubs/pull/6495 had nothing to do with issue https://github.com/Hubs-Foundation/hubs/issues/6496, so this PR doesn't do anything to fix the problem.  Closing.

PR #6502, opened 2024-09-26
Fixes #6496 - scale & rotate buttons

### LINKED ISSUE #6496: Spacebar Menu Scale and Rotate Buttons Do Nothing When Interacted  With

DougReeder:
Bisecting showed the PR that introduced the bug was https://github.com/Hubs-Foundation/hubs/pull/6487  Among many changes there, findAncestorWithComponents() needed be called with Rigidbody.  Holdable was retained and Deletable and MediaLoader were left out.  

If it was a simple oversight, this PR should be a complete fix.  If there were reasons to drop Deletable and MediaLoader, we'll need to do more work.

### DIFF
--- a/src/systems/hold-system.js
+++ b/src/systems/hold-system.js
@@ -13,7 +13,7 @@ import {
   HeldHandLeft,
   AEntity,
   Networked,
-  Rigidbody
+  Rigidbody, Deletable, MediaLoader
 } from "../bit-components";
 import { canMove } from "../utils/permissions-utils";
 import { canMove as canMoveEntity } from "../utils/bit-permissions-utils";
@@ -81,7 +81,7 @@ export function isAEntityPinned(world, eid) {
 function grab(world, userinput, queryHovered, held, grabPath) {
   const hovered = queryHovered(world)[0];
 
-  const interactable = findAncestorWithComponents(world, [Holdable, Rigidbody], hovered);
+  const interactable = findAncestorWithComponents(world, [Deletable, MediaLoader, Holdable, Rigidbody], hovered);
   const target = interactable ? interactable : hovered;
   const isEntityPinned = isPinned(target) || isAEntityPinned(world, target);
 


Exairnous:
@DougReeder I think this looks good.  Thanks.
@keianhzo can you confirm?

keianhzo (commented):
I believe this code will break the Blender grabbables component as they are not deletable so looking for the Deletable component when grabbing will mean that those won't be grabbable anymore.

I'd look further in that code to see why the menu buttons don't catch the focus...

Another option if that PR has introduced many bugs (as it seems) would be to revert it both there and in the Blender component, it's only one feature that would get lost and probably not a massively used one.

DougReeder:
I removed `Deletable` from this code, and you could scale but not rotate objects. :-S

If restoring `Deletable` and `MediaLoader` here breaks the Blender grabbables component, it looks like we should revert https://github.com/Hubs-Foundation/hubs/pull/6487

Exairnous:
I did a little digging and it looks like the reason this PR appears to be working is because it's not finding any entity with all of those components, so it's falling back to using the hovered entity.  For some reason, the rotate and scale buttons seem to behave differently from the rest of the menu, probably because they are interactible/grabable instead of just single click buttons.  So, I think the correct solution, fingers crossed, should be to check if it's a HoldableButton first, and then use that if it is, or if it isn't then check for the Holdable Rigidbody.  Here's is some code that I think should work, but I've only done a cursory test of it.

```
const holdablebutton = findAncestorWithComponent(world, HoldableButton, hovered);
const interactable = holdablebutton ? holdablebutton : findAncestorWithComponents(world, [Holdable, Rigidbody], hovered);
target = interactable ? interactable : hovered
```
(note, this was manually retyped from a different computer I was testing on, so beware of typos)

@keianhzo What are your thoughts on this reasoning/solution?

DougReeder:
Withdrawn in favor of PR #6515

Issue #6507, opened 2024-09-30
Invite people using the Web Share API


DougReeder:
**Is your feature request related to a problem? Please describe.**
To share the URL of a room, a user must click the Invite button then separately paste the link into another app.

**Describe the solution you'd like**
The "Invite" button should send the URL of the instance to the Web Share API. This reduces the friction of inviting other people.

**Describe alternatives you've considered**
Share buttons for individual social networks take up space for people who don't use them, and exclude less popular networks.

**Additional context**

https://developer.mozilla.org/en-US/docs/Web/API/Web_Share_API

https://hominidsoftware.com/presentations/Web%20Share%20APIs/

DougReeder:
Fixed by #6508

PR #6508, opened 2024-10-02
Invites & sharing Room URL use Web Share API if available

DougReeder:
Sharing is an additional option, when available. It works as expected on Firefox and Chrome on MacOS, Oculus Browser on Quest, Safari on iOS and Chrome on Android.  As noted earlier, many platforms only pass the URL to destination, or just the URL and title.

Implements #6507 

There's an image with this code on Docker Hub: dougreeder/hubsfirst:v1


DougReeder:
Here's a screenshot of the new Share room link using the Web Share API.

Share room link

And here's one possibility of what might result
room link email

DougReeder:
PinkPolygonSoup has said the wording is fine.

DougReeder:
That completes the changes from today's dev meeting. Now that we have testing by other people, I think this is ready to go!

Exairnous:
My apologies for the delay.  This looks good to me code and functionality-wise, but prettier complains when I run `npm run test` can you confirm/fix?  After that's done, I think this PR is good to merge.  Thank you very much for submitting this improvement :)

PR #6511, opened 2024-10-16
Adds Share button to picture and video items in Chat

DougReeder:
Hubs Chat share 2024-10-15 at 10 26 25 PM

These buttons function like the button on photo and video items in the Presence log, but also catch thrown errors.  Being in the Chat log, the user does not have to act quickly to share.   These new buttons are not used with Chat text or permission messages.


Exairnous:
LGTM.  Thanks.  Merging.

PR #6512, opened 2024-10-16
Scene page: replaces 'Share on Twitter' w/ 'Share' (using Web Share API)

DougReeder:
Hubs scene Share 2024-10-16 at 4 20 33 PM

This enables sharing to whatever apps the user has installed. If the Web Share API is unavailable, it falls back to sharing via Twitter.  (Also, Twitter changed its name a year ago, so removing "Twitter" from the UI looks more professional.)


Exairnous:
LGTM.  Thanks. Merging.

PR #6514, opened 2024-11-01
Partially fix rotating and scaling via the spacebar menu buttons

Exairnous:
Pass the eid of the object to the updateRigidBody function instead of the bodyId for the bitECS versions of rotating and scaling via the spacebar menu buttons.  updateRigidBody attempts to get the bodyId from the eid, so doing it beforehand prevented the body from being found.

Update the check for whether the aframe object is a rigid body by checking for the "body-helper" attribute.  This allows the update of the object to be kinematic to go through when the spacebar rotate/scale buttons are engaged.

Check if what is being hovered over when grabbing something is a HoldableButton and if it is, then use that as the hold target instead of the 3D object.  This causes the grab to interact with the button and start the transform instead of just moving the object.

Note: if you attempt to throw the object in the aframe loader and then try to rotate it again with the button, the physics system will fight it for some reason and prevent the rotation.  Clicking on the object a couple times will cause the physics system to release it, but the reason for this behavior is unknown; it could possibly be related to ownership not being set properly or the rigid body not being updated properly.  This behavior was fixed on the bitECS side by passing the eid instead of the bodyId to the updateRigidBody function, but activating the rigid body update on the aframe side (mentioned above) didn't fix it for the aframe side.

### DIFF
--- a/src/bit-systems/object-menu.ts
+++ b/src/bit-systems/object-menu.ts
@@ -105,7 +105,7 @@ function startRotation(world: HubsWorld, menuEid: EntityID, targetEid: EntityID)
   }
   const transformSystem = APP.scene!.systems["transform-selected-object"];
   const physicsSystem = AFRAME.scenes[0].systems["hubs-systems"].physicsSystem;
-  physicsSystem.updateRigidBody(Rigidbody.bodyId[targetEid], { type: "kinematic" });
+  physicsSystem.updateRigidBody(targetEid, { type: "kinematic" });
   const rightCursorEid = anyEntityWith(world, RemoteRight)!;
   transformSystem.startTransform(world.eid2obj.get(targetEid)!, world.eid2obj.get(rightCursorEid)!, {
     mode: TRANSFORM_MODE.CURSOR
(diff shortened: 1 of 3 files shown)


Exairnous:
Further tests to fix this were researched at a Community Collaboration Session, but have yet to be integrated into this branch: https://github.com/Hubs-Foundation/hubs/compare/9cd2a7bc532920f469f96fb4e460a51e4a8db6a6...Exairnous:hubs:fix-holdable-button-interactions-test

DougReeder:
Should the lines in the "further tests"
```
if (hovered) {
    console.log(`hovered: ${hovered}`);
...
```
use console.debug instead?

Exairnous:
The entire `if (hovered) {` block was/is a research aid and isn't intended for any sort of production integration.  I should have been clearer that the commits I linked to were entirely for research/informational purposes and not meant to be merged directly (cleanup and further evaluation are needed for a proper integration of a fix, and I would likely do that directly on the branch for this PR).  My apologies.

In case your question isn't in regard to merging the test branch into this branch, I used `console.log` while I was experimenting because it was fast and easy, but if you want to use `console.debug` when taking a look, that should be fine too.

Issue #6505, opened 2024-09-27
The spacebar menu doesn't show up for objects which have been dropped in the room from a file browser


Exairnous:
**Description**
The spacebar menu doesn't show up for objects which have been dropped in the room from a file browser.

**To Reproduce**
1. Enter a Hubs room.
2. Drag and drop in a GLB file that you have locally on your computer into the room.  This file can be downloaded and used to test if you don't have any GLBs handy: [Test_Cube.zip](https://github.com/user-attachments/files/17160779/Test_Cube.zip).
3. Hover your cursor over the object and press and hold the spacebar.
4. See that no menu appears.

**Expected behavior**
The spacebar menu should appear over an in-room object that you have uploaded from your computer.

**Hardware** 
 - Device: Laptop
 - OS: GNU/Linux
 - Browser: Firefox

**Additional context**
Potentially caused by #6487

Exairnous:
Fixed by https://github.com/Hubs-Foundation/hubs/pull/6515

Issue #6504, opened 2024-09-27
Physics doesn't continue working after moving an object that has had physics permanently enabled on it


Exairnous:
**Description**
Physics doesn't continue working after moving an object that has had physics permanently enabled on it.  Physics initially works immediately after enabling it on an in-room object, but stops working after the object is moved.

**To Reproduce**
1. Enter a Hubs room.
2. Spawn an object into the room (you can use the /duck command).
3. Hover over the object with your mouse and press and hold spacebar to bring up its menu.
4. Click on the ball with a downwards arrow to permanently enable physics.
![2024-09-27_03-32](https://github.com/user-attachments/assets/2bc6aff9-9681-4f74-9661-0a6fa63a751f)
5. See the object drop to the ground.
6. Click and drag the object to a new position in space.
7. See that the object doesn't drop to the ground when you release the mouse button.

**Expected behavior**
The object should drop to the ground after moving it, when physics is permanently enabled.

**Hardware** 
 - Device: Laptop
 - OS: GNU/Linux
 - Browser: Firefox

**Additional context**
Potentially caused by #6487

Exairnous:
Fixed by https://github.com/Hubs-Foundation/hubs/pull/6515

Issue #6503, opened 2024-09-27
Physics doesn't work when in-room objects are thrown


Exairnous:
**Description**
Physics doesn't work when in-room objects are thrown.  The objects immediately cease moving when the mouse button is released at the end of the throw instead of continuing on with the physics system.

**To Reproduce**
1. Enter a Hubs room.
2. Spawn an object into the room (you can use the /duck command).
3. Throw the object by clicking and dragging on it quickly and then releasing the mouse button while the object is moving.
4. See that the object immediately stops moving when you release the mouse button.

**Expected behavior**
The object should continue moving in accordance with its momentum when the mouse button is released.

**Hardware** 
 - Device: Laptop
 - OS: GNU/Linux
 - Browser: Firefox and Chromium

**Additional context**
Potentially caused by #6487

Exairnous:
Fixed by https://github.com/Hubs-Foundation/hubs/pull/6515

PR #6515, opened 2024-11-07
Revert scene grabbables support

Exairnous:
This reverts PR #6487 and fixes these bugs caused by it:
* #6496
* #6503
* #6504
* #6505


DougReeder:
I concur that this is the least-bad option at this point.

Exairnous:
Thanks @DougReeder.  I've Merged it, now.

PR #6516, opened 2024-11-08
Fixes layout bug in chat sidebar introduced by PR #6511

DougReeder:
Sidebar chat text with hard-to-break lines, such as URLs, extends outside the bubble. This forces better line breaking.


Exairnous:
LGTM.  Thank you.  Merging.

PR #6523, opened 2024-11-29
Exited room screen: email link includes name of room

DougReeder:
When a room is closed, a screen is displayed that contains a link for the contact email. This PR adds a default subject containing the name of the room, so the administrator can tell what room the user is communicating about.


Exairnous:
LGTM.  Thanks.  Merging.

PR #6524, opened 2024-11-29
Fixes #6522 multi-line chat messages

### LINKED ISSUE #6522: Chat messages consisting of multiple lines are displayed in columns in the chat sidebar
**Description**
When a chat message is composed with multiple lines and sent, each line is then displayed as its own column in the chat sidebar instead of on separate rows.  Chat messages are also displayed right to left in the chat sidebar for the sender (they are correctly displayed left to right for the receiver).

**To Reproduce**
1. Enter a Hubs room. Duplicate the tab if you want to observe the messages as both the sender and receiver.
2. Open the chat.
3. Compose and send a multi-line message (new lines can be added by pressing Shift+Enter).
4. See that each line makes up it's own column (long lines and many of them make long thin columns, short lines make it seam as if the new lines are ignored as everything appears run together on one row).

**Expected behavior**
Multi-line chat messages should be displayed in the chat sidebar with each line below the other instead of side by side in columns.

**Screenshots**
Sender - Bug:

Receiver - Bug:

Sender - Correct:

**Hardware** 
 - Device: Laptop
 - OS: GNU/Linux
 - Browser: Firefox and Chromium

**Additional context**
Potentially caused by https://github.com/Hubs-Foundation/hubs/pull/6511

DougReeder:
This is how the layout should have been done in PR #6511 — the share buttons and the message bubbles are enclosed in a new div.

### DIFF
--- a/src/react-components/room/ChatSidebar.js
+++ b/src/react-components/room/ChatSidebar.js
@@ -406,8 +406,10 @@ function getMessageComponent(message, intl) {
     }
     case "video":
       return (
-        <MessageBubble key={message.id} media>
-          <video controls src={message.body.src} />
+        <div className={styles.messageRow}>
+          <MessageBubble key={message.id} media>
+            <video controls src={message.body.src} />
+          </MessageBubble>
           <IconButton
             className={styles.iconButton}
             onClick={onShareClick}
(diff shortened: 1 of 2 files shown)


Exairnous (approved):
LGTM.  Thank you!  Merging.

PR #6531, opened 2025-02-25
Documents that several addons are included with the addons branch

DougReeder:
Also updates GitHub links to Hubs-Foundation repos.

### DIFF
--- a/doc/add-ons.md
+++ b/doc/add-ons.md
@@ -2,6 +2,9 @@
 
 Hubs client add-ons are pluggable libraries that are installed as part of the client and loaded at runtime. Add-ons allow functionality to the Hubs core and allow easy Hubs client extensibility while maintaining the Hubs client core lean and minimal.
 
+The currently installed addons are listed in the `addons` array in the `addons.json` file at the root of the Hubs client source.
+Several are included by default, see [Add-ons included in `addons` branch](#add-ons-included-in-addons-branch).
+
 ## Installation
 
 Installing an addon in your Hubs client is a two step process:
@@ -10,7 +13,7 @@ Installing an addon in your Hubs client is a two step process:
    For example if we would want to install the portals add-on we would need to do:
 
 ```
-npm i https://github.com/MozillaReality/hubs-portals-addon.git
+npm i https://github.com/Hubs-Foundation/hubs-portals-addon.git
 ```
 
 2. After the add-on is installed you'll need to add it to the addons array inside the `addons.json` file at the root of the Hubs client source.
@@ -58,7 +61,7 @@ Add-ons can also be enabled/disabled per room independently from the instance co
 
 You can develop Hubs add-ons using Javascript or Typescript.
 
-If you are developing using Typescript you can use the add-on template as a starting point. You can get the Add-on template from [here](https://github.com/MozillaReality/hubs-template-addon).
+If you are developing using Typescript you can use the add-on template as a starting point. You can get the Add-on template from [here](https://github.com/Hubs-Foundation/hubs-template-addon).
 
 The add-on template has the basic dependencies already configured to get started with development as fast as possible.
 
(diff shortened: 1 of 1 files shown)


doc/add-ons.md
```diff
@@ -2,6 +2,8 @@
 
 Hubs client add-ons are pluggable libraries that are installed as part of the client and loaded at runtime. Add-ons allow functionality to the Hubs core and allow easy Hubs client extensibility while maintaining the Hubs client core lean and minimal.
 
+The currently installed addons are listed in the `addons` array inside the `addons.json` file at the root of the Hubs client source. Several are included with the `addons` code branch.
```
Exairnous, line 5: What about changing this from "Several are included with the `addons` code branch." to something like "Several are included by default, see `[Currently available add-ons](#currently-available-add-ons)`."

I think it would be nice to know which ones are included by default (although that section isn't entirely accurate the way it is since the template add-on isn't installed by default, but it could be moved to a subsection), also the addons branch is only temporary, so this would future proof things.

doc/add-ons.md
```diff
@@ -94,11 +96,11 @@ The easiest way of iterating over an add-on development is by linking the add-on
...
-- [Duck add-on](https://github.com/MozillaReality/hubs-duck-addon): Replaces the existing `/duck` chat command and refactors it into a Hubs add-on.
-- [Portals add-on](https://github.com/MozillaReality/hubs-portals-addon): Simple portals implementation as an add-on that lets you spawn portals using the `/portal`. It also show how to add key bindings.
-- [Behavior Graphs add-on](https://github.com/MozillaReality/hubs-behavior-graphs-addon/): Initial Behavior Graphs implementation as an add-on
+- [Template add-on](https://github.com/Hubs-Foundation/hubs-template-addon). This add-on serves as a foundation for add-on development.
+- [Duck add-on](https://github.com/Hubs-Foundation/hubs-duck-addon): Replaces the existing `/duck` chat command and refactors it into a Hubs add-on.
+- [Portals add-on](https://github.com/Hubs-Foundation/hubs-portals-addon): Simple portals implementation as an add-on that lets you spawn portals using the `/portal`. It also show how to add key bindings.
+- [Behavior Graphs add-on](https://github.com/Hubs-Foundation/hubs-behavior-graphs-addon/): Initial Behavior Graphs implementation as an add-on
```
Exairnous, line 106: This section is missing the post-processing add-on.  Could you add it?

Exairnous (commented):
Looks mostly good.  Thanks.  I left a couple suggestions, though.

doc/add-ons.md
DougReeder: done

doc/add-ons.md
DougReeder: done

doc/add-ons.md
```diff
@@ -94,11 +97,17 @@ The easiest way of iterating over an add-on development is by linking the add-on
...
-There is a typings library available for Typescript development: [Hubs Client TS Types](https://github.com/MozillaReality/hubs-ts-types)
+There is a typings library available for Typescript development: [Hubs Client TS Types](https://github.com/Hubs-Foundation/hubs-ts-types)
 
 ## Currently available add-ons
 
-- [Template add-on](https://github.com/MozillaReality/hubs-template-addon). This add-on serves as a foundation for add-on development.
-- [Duck add-on](https://github.com/MozillaReality/hubs-duck-addon): Replaces the existing `/duck` chat command and refactors it into a Hubs add-on.
-- [Portals add-on](https://github.com/MozillaReality/hubs-portals-addon): Simple portals implementation as an add-on that lets you spawn portals using the `/portal`. It also show how to add key bindings.
-- [Behavior Graphs add-on](https://github.com/MozillaReality/hubs-behavior-graphs-addon/): Initial Behavior Graphs implementation as an add-on
+## Template
```
Exairnous, line 104: This should have an extra `#` so that it's a subheading of `Currently available add-ons`.
```suggestion
### Template
```

doc/add-ons.md
```diff
@@ -94,11 +97,17 @@ The easiest way of iterating over an add-on development is by linking the add-on
...
+
+- [Template add-on](https://github.com/Hubs-Foundation/hubs-template-addon). This add-on serves as a foundation for add-on development.
+
+## Add-ons included in `addons` branch
```
Exairnous, line 108: This should have an extra `#` so that it's a subheading of `Currently available add-ons`.
```suggestion
### Add-ons included in `addons` branch
```

I still kind of think that since the `addons` branch will be merged into master that we should avoid mentioning it and do something like `### Add-ons included by default` to future-proof it so we don't have to change it again later (looking back at my previous comment, I realize I didn't communicate my thoughts on referencing the `addons` branch very well, sorry).  But if you think that people would benefit from the reference then I'm okay with leaving it in until the branch gets merged.

Exairnous (commented):
Thank you for the updates.  I left a couple more requests/suggestions.

doc/add-ons.md
DougReeder: Is the plan to include these with Hubs in the master branch? It's certainly possible to remove them by removing them from the JSON file.

DougReeder:
Just before merging into master, I think we'll be updating the docs anyway.

Exairnous:
Discussed and approved at the dev meetup: https://docs.google.com/document/d/1OMh3-_DpjHABMMrXtECzLPNxipQMixx32l6K4zqSoYY/edit?tab=t.0

Merging.

PR #6532, opened 2025-03-07
Fix missing link warning

HussainAther:
## Summary

- Added a console.warn message to gltf-model-plus.js when a link is missing or improperly defined.
- This helps with debugging link-related issues and prevents silent failures.

## Testing

- Ensure media links load correctly.
- If a link is missing, check for console warnings in the browser console.


src/components/gltf-model-plus.js
```diff
@@ -595,15 +595,17 @@ class GLTFHubsComponentsExtension {
...
-              }
-            }
-          }
-
+            src: ext.link.href
+           };
+           delete ext.link;
+                   }
+               }
+           }
```
DougReeder: This looks good, but please fix the indentation.

DougReeder (changes requested):
The fixes look good, but the PR should be against the addons branch.

src/components/gltf-model-plus.js
HussainAther: Sure, how exact should I fix it?

DougReeder:
Ah, you'll need to create a new branch from `master` or `addons`, then cherry-pick your commits into it, so the PR is clean.  It's probably easier to create a new PR from that branch.

DougReeder:
Closed in favor of #6533

PR #6547, opened 2025-05-04
Fix incorrect chat grouping of Discord messages

Exairnous:
## What?

Prevents Discord messages from different users that are sent within a minute of each other from being grouped together under the first user that sends a message by also checking the name given for the sender instead of just the sender's id.  This mostly ensures that all chat messages are attributed to the correct user.

## Why?

Users in Hubs can have the same name so the chat messages are grouped by a sender id, but because all the messages from Discord come through the Discord bot, the sender id is the same for all of them regardless of who sent them in Discord.

## Examples

Not easy to get and the change is very small, easily seen, and is described in other parts of this pull request.

## How to test

1. Redeploy your instance using a Hubs Client image generated from this PR (e.g. `ghcr.io/exairnous/hubs:fix-discord-bot-mixing-chat-messages-34`).
2. Follow the instructions for hacking on the bot locally that are listed in PR https://github.com/Hubs-Foundation/hubs-discord-bot/pull/146 and set up the Hubs bot against your Community Edition instance.
3. Create a room with the Hubs bot with `!hubs create`.
4. Join the Hubs room that has been bridged to Discord.
5. In your Discord server, in the channel you've set the bot up in, send a Discord message and have someone else send a Discord message immediately after.
6. See that they are listed as from the correct people in the chat panel (previously they would both be listed under the name of the person who first sent the message).

This is an implementation detail and doesn't need to be documented.  It is also immediately obvious.

## Limitations

This will still group messages improperly if the Discord users have the same display name in the Discord server but preventing that would likely require much more extensive and broad reaching changes (probably in reticulum and the Hubs Discord bot repository as well) and it would do little to reduce confusion.

## Alternatives considered

None.

## Open questions

None.

None.

### DIFF
--- a/src/react-components/room/contexts/ChatContext.tsx
+++ b/src/react-components/room/contexts/ChatContext.tsx
@@ -30,7 +30,7 @@ function shouldCreateNewMessageGroup(messageGroups: NewMessageT[], newMessage: N
 
   const lastMessageGroup = messageGroups[messageGroups.length - 1];
 
-  if (lastMessageGroup.senderSessionId !== newMessage.sessionId) {
+  if (lastMessageGroup.senderSessionId !== newMessage.sessionId || lastMessageGroup.sender !== newMessage.name) {
     return true;
   }
   if (lastMessageGroup.type !== newMessage.type) {


Exairnous:
@HrithikWins I'd also appreciate your review on this.

DougReeder (approved):
LGTM

DougReeder (approved):
LGTM

PR #6549, opened 2025-05-27
Eliminate dozens of deprecation and security vulnerability warnings with incremental package updates

matthargett:
## What?
This fixes a number of lagging dependencies that were producing a lot of console (terminal and browser) noise and several security vulnerabilities.

## Why?
I was trying to fix the WebXR issues I've enountered in Hubs over the last couple of meetups, but found myself stymied by the dozens and dozens of warnings during initial setup and that several storybook entries were rendering blank pages. I decided it would be a better service to get this repo generally easier to run in fresh checkouts on modern operating system and node versions. I hope this helps broaden contributions in the future beyond myself.

## How to test

1. `npm ci` should now complete with a lot less noise on the console about deprecations and security vulnerabilities.
2. `npm test` should now pass more consistently without having to install nodejs 12.
3. `npm storybook` should now allow you to click and see every single component render, without any blank pages.
4. `npm local` now has significantly less noise, making it easier to see warnings and errors
5. `npm bundle-analyzer` will show a nice reduction in the bundle size thanks to optimizing for Chrome 91 (Oculus Go) on the low-end instead of Chrome 51 and IE11. (see comments in the babel conf for compatibility matrix choices)

Everything should work as before, just better and with less noise for new (and existing) developers).

## Limitations
Getting an e2e QA setup was beyond my abilities and time, so I'd appreciate some help testing this more deeply in a multiuser way. I've expanded unit testing a bit to cover the areas I was concerned about breaking (PDFs), and am happy to add more unit tests for whatever surfaces during an e2e test so that all future modifications can get a better signal on regressions.

## Alternatives considered
I tried to get meaningful reduction in noise, rather than updating to absolute latest and greatest package versions. In most cases, bumping one package required bumping another (especially corejs and babel to support new kinds of exports). If this goes well, we can do another pass. If that goes well, we can look at gently cherry-picking more things from upstream aframe and threejs into the Hubs fork.

## Open questions
I need help with end-to-end QA, and am happy to do a 1-hour pair-testing session to work through any issues.

If this works well, and there's no immediate requests, I'll move onto solving my WebXR issue on Apple Vision Pro.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -62,5 +62,5 @@ module.exports = {
     ],
     "@calm/react-intl/missing-values": "error"
   },
-  extends: ["prettier", "plugin:react/recommended", "eslint:recommended", "plugin:storybook/recommended"]
+  extends: ["prettier", "plugin:react/recommended", "eslint:recommended"]
 };
--- a/admin/package-lock.json
+++ b/admin/package-lock.json
@@ -38,11 +38,11 @@
         "@babel/preset-env": "^7.23.0",
         "@babel/preset-react": "^7.23.0",
         "@babel/register": "^7.18.9",
-        "babel-loader": "^8.2.5",
+        "babel-loader": "^10.0.0",
         "babel-plugin-formatjs": "^10.5.0",
         "babel-plugin-transform-react-jsx-img-import": "^0.1.4",
         "buffer": "^6.0.3",
-        "copy-webpack-plugin": "^11.0.0",
+        "copy-webpack-plugin": "^13.0.0",
         "cors": "^2.8.5",
         "css-loader": "^6.7.1",
         "dotenv": "^16.0.1",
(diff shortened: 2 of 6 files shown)


Exairnous:
@matthargett I tried building this with the [Custom Docker Build Push GitHub Action](https://docs.hubsfoundation.org/github-workflows.html), but it failed to build.  It looks like the Dockerfile hasn't been updated to use Node 22 (however, I did a quick test of that and updating it seemed to cause issues with the installation of PhantomJS, probably because the new Node image uses a newer Debian release, although I only tried the `22` Docker image, so maybe one of the other variations wouldn't have that problem).  Can you confirm?

Exairnous:
@matthargett Thanks!

I'm thinking that option 1 sounds like it has more of a chance to preserve the arm64 support, so potentially that's the better option (@ajlennon ?).  As to where to submit the PR, I'm not quite sure what you mean.  As far as I know, the spot where this would be modified is the `RetPageOriginDockerfile` in this repository, so I think you can just update this PR.

> also note: phantomjs has been archived and abandoned for over 5 years now

Yes.  We don't use it directly (it looks like a dependency of a dependency), so it would be great to be able to remove it, but that seems like it could be quite tricky (I haven't looked into it too much, though).

PR #6553, opened 2025-06-04
bump docker container to nodejs 22 and remove phantomjs 

matthargett:
## What?
This tests upgrading the docker container to a newer Debian without phantomjs. PhantomJS appears to be a vestigial dependency that isn't used by anything in the repository, nor as a runtime `dependency` in any transitive sub-dependencies.

## Why?
This is to enable the modernization of packages and infrastructure in #6549 more incrementally.

## How to test
Docker image builds and deploys.

Docker file is self-documenting, and this should be a seamless transition.

## Alternatives considered
We could update to 'just' nodes 20 and still install phantomjs, but I think it's better to relieve the phantomjs constraint (archived and abandoned for over 5 years now).

in support of #6549

### DIFF
--- a/RetPageOriginDockerfile
+++ b/RetPageOriginDockerfile
@@ -2,9 +2,7 @@
 # this dockerfile produces image/container that serves customly packaged hubs and admin static files
 # the result container should serve reticulum as "hubs_page_origin" and "admin_page_origin" on (path) "/hubs/pages"
 ###
-from node:16.16 as builder
-env QT_QPA_PLATFORM offscreen
-run apt update && apt -y install phantomjs
+from node:22 as builder
 run mkdir -p /hubs/admin/ && cd /hubs
 copy package.json ./
 copy package-lock.json ./


DougReeder (approved):
Works fine in my instance.

matthargett:
> @matthargett LGTM.  Thanks.
> 
> @ajlennon reported in [Discord](https://discord.com/channels/498741086295031808/535606666708910101/1383060526086623274) that it worked for him as well for both amd64 and arm64. 
> 
> 
> 
> I noticed that this doesn't bump the node version mentioned in the readme, should it?  Or is that something you're leaving for https://github.com/Hubs-Foundation/hubs/pull/6549

It does bump the node image version from 16.x to 22.x. Were you expecting a separate version bump in terms of an "engines" directive?

Exairnous:
I believe you added an "engines" directive in your other pull request for Node 22, but no, I was referring to the fact that the readme says "We use 16.16.0 on our build servers." and this isn't changed in this pull request to the new version of 22.

DougReeder:
The current version of this (f96fc6de0fc68105317be2833eed82d4053c26be) builds without issues on iMac with 8-Core Intel Core i7, and the image installs fine on my instance, and passes the smoke test for Firefox 139.0.1 (64-bit), Chrome Version 137.0.7151.104 (Official Build) (x86_64),  Oculus Browser 39.0.0.15.... on the Quest 3 and Chrome 136.0.7103.125 on a Pixel 5 running Android 18.

Still can't enter a room from the lobby on my iPhone, so that's no regression.

PR #6560, opened 2025-07-22
merges commits from master since 2024-11-28 into addons branch

DougReeder:
## What?
merges 21 commits from master since 2024-11-28 into addons branch

## Why?
To keep addons from diverging too far from master

## How to test

1. Pull the image ghcr.io/dougreeder/hubs:addons-ajv-latest into a hubs instance
2. Verify that addons still work
3. Verify that nothing re: non-addons broke that wasn't already broke.

https://hubsfoundation.org/posts/914/hubs-instance-updates-2025-06-19/

## Limitations
A number of things were broken in the addons branch, and they are still broken

## Alternatives considered
We probably should have merged this in smaller chunks


DougReeder:
I'm not seeing any capitalization changes to "Hubs-Foundation" in this PR.  All of them were already capitalized, so far as I can tell.  And this commit is in the log:

commit 7a99b6ff538e0e771b4d5c08fae198a21e7d0a22
Author: Ryan Inch <mythologylover75@gmail.com>
Date:   Fri Nov 15 10:25:25 2024 -0500

    Make dependency URL capitalization consistent
    
    npm throws errors if the capitalization between a dependency listed here and a dependency listed in an addon is different, so this makes sure there are no dependency conflicts.

Exairnous:
Yes, the commit is in the log.  There were four changes to "Hubs-Foundation" in this PR.  Lines 141, 142, 221, and 15236 had their capitalization removed and are now all lowercase (I did double check to make sure this change is reflected in the addons branch and it is).  Also, (and this is weird) when I did a blame on the addons version of the file I found that the affected lines didn't reference the merge commit or the capitalization commit but a previous one.  I think the cleanest way to fix this is to force push the state it was before the merge and then redo the merge.  If you want, I can do this and then put up a new PR.

PR #6564, opened 2025-08-22
Adds Storybook test cases for scene-ui.js

DougReeder:
## What?
Adds Storybook test cases for scene-ui.js
Also rounds out a few other Storybook stories.

## Why?
1. They didn't exist yet, and production code was changed to use the Web Share API.
2. The other stories didn't fully test what they were supposed to.

## How to test

1. Run `npm run storybook`
3. Surf to `http://localhost:6006/?path=/story/scene-ui--base`
4. Examine each of the scene-ui stories (Show unavailable has white text on a white background, as Storybook could use better style configuration).
5. Navigate to the ChatSidebar — Base story
6. Vary the window width, to see how line breaking behaves.
7. Navigate to Tip — Tooltips
8. In the Controls section at the bottom, select each test case and inspect the UI. 

## Limitations
None of the Mobile test cases for Tooltip worked before, and they still don't.  The UI in the actual app is fine. It's not clear why Storybook gives bad results here.

## Alternatives considered
If the Mobile test cases for Tooltip were included, it would be natural for a dev to conclude the actual UI is wrong. That would waste dev time on something that isn't broken in the actual app. 

## Open questions
How could Storybook be better configured, or the Tooltip story better written?

This depends on Matt's Storybook fixes. https://github.com/Hubs-Foundation/hubs/pull/6563


DougReeder:
With the SCSS change:
image

DougReeder:
Fixes https://github.com/Hubs-Foundation/hubs/issues/6568 and the story "Scene Attributions" shows that.  See the image above.

Also fixes story "Show Unavailable".  The text was there, but white text on a white background was not visible. The change to `scene-ui.js` ensures the text is visible regardless of whether the screen shot is light or dark.

Doesn't fix https://github.com/Hubs-Foundation/hubs/issues/6567 and the stories "Scene Allow Remixing" and "Owner Project Id" reflect that.  Also, it was not clear how to set the app feature  "Enable Scene Editor" on, for Storybook.

src/react-components/room/Tip.stories.js
```diff
@@ -25,18 +25,24 @@ Tips.propTypes = {
...
+  "tips.desktop.defense": "Desktop Defense",
   "tips.desktop.invite": "Desktop Invite",
   "tips.desktop.end": "Desktop End",
   "tips.desktop.menu": "Desktop Menu",
-  "tips.mobile.welcome": "Mobile Welcome Message",
-  "tips.mobile.locomotion": "Mobile Locomotion",
-  "tips.mobile.turning": "Mobile Turning",
-  "tips.mobile.end": "Mobile End",
-  "tips.mobile.menu": "Mobile Menu"
+  // "tips.mobile.locomotion": "Mobile Locomotion",
```
matthargett, line 38: should these comments be removed?

matthargett:
some general comments you can take or leave :)

* I think I see two more variants you coudl add to mirror issue #6568 : one that passes a parentScene (to validate the remix attribution path) and one that feeds the legacy  sceneAttributions.extras field.
* For the create-room story, consider stubbing createAndRedirectToNewHub so clicking the button doesn’t try to POST against a real backend during local Storybook runs. needing to have a real local backend running just to do a visual regression test is a notable barrier
* src/react-components/scene-ui.stories.js:24 – does the Remix variant ever render its CTA? it looks like the component sits behind IfFeature('enable_spoke'), and Storybook boots with that flag off. The story shows the same UI as Base, so it looks like it can’t regress the scenario it’s meant to cover
* Remix link is emitted as /spoke/projects/new?sceneId=undefined since the story omits sceneId. that would seem to not represent the primary behavior, right? Supplying a realistic id (and, ideally, plumbing it through Storybook args) should fix the href

src/react-components/room/Tip.stories.js
DougReeder: as noted above, Test cases where the actual software works fine, but the Storybook is wrong, have been commented out.

I prefer to have commented-out code if, and only if, there's a plausible reason you would uncomment it in the future.  I'm hopeful we'll get Storybook working without bugs in the future.

DougReeder:
> * stubbing createAndRedirectToNewHub so clicking the button doesn’t try to POST

Is there documentation on stubbing?

> does the Remix variant ever render its CTA? it looks like the component sits behind IfFeature('enable_spoke'), and Storybook boots with that flag off

Is there a way to set `IfFeature('enable_spoke')` on for some stories and off for others?

src/react-components/scene-ui.stories.js
```diff
@@ -0,0 +1,74 @@
...
+export const ShowUnavailable = () => (
+  <SceneUI
+    sceneName="Very Cool Scene"
+    sceneDescription="A must see!"
+    sceneScreenshotURL="https://example.com/"
+    unavailable={true}
+  />
+);
+
+export const ParentScene = () => (
```
Exairnous: This seems like it's showing a child/remixed scene, so should the story be named "Child Scene" or "Has Parent Scene"?

Exairnous: This seems like it's showing a child/remixed scene, so I think this sub-story should be named "Child Scene" or "Has Parent Scene"?

Exairnous (approved):
With the ParentScene sub-story renamed to RemixedScene, everything looks good to me.  Thanks.

Edit: Merged after discussion in the dev meetup.

PR #6580, opened 2026-05-04
Fix: Improve touchpad wheel responsiveness and pinch gesture scaling

### LINKED ISSUE #6556: Touchpad scrolling bug
**Description**
When spawning an object I can usually hold it and scroll mouse to push it towards and away from me.

**To Reproduce**
If I hold down left mouse button on touchpad and pinch it to push it towards and away from me it doesn't work

**Expected behavior**
pinching should move object away close to me like behaviour with a mouse

HussainAther:
## What?
This PR now  introduces normalization logic for wheel input within the mouse device system to improve responsiveness for trackpad users.

## Why?
- Trackpads had some reports of  fractional or very low delta values compared to standard mouse wheels, causing scrolling or object manipulation to feel sluggish or unresponsive. 

- Pinch gestures on trackpads (often triggering Ctrl+wheel) were not correctly scaling. This PR increases the the sensitivity for small deltas. Then, it amplifies the input when the Ctrl key is held.

## Examples
N/A

## How to test
1. Open the application using a trackpad.
2. Attempt to scroll or rotate a held object.
3. Verify that the movement is responsive and not sluggish.
4. Hold the Ctrl key and perform a scroll or pinch gesture.
5. Verify the movement speed is increased as expected.

No documentation updates are required for this minor logic fix.

## Limitations
None.

I considered using a device-detection library to switch input logic, but decided to normalize the raw delta values directly in the processor for better performance and simplicity.

## Open questions
None.

Fixes: https://github.com/Hubs-Foundation/hubs/issues/6556 
This change aims to bring trackpad usability closer to that of a standard mouse wheel. 

Parts of this change were developed with the assistance of an AI tool (ChatGPT), particularly for debugging guidance and refining the implementation. The final code and testing were reviewed and validated manually.

### DIFF
--- a/src/systems/userinput/devices/mouse.js
+++ b/src/systems/userinput/devices/mouse.js
@@ -62,11 +62,11 @@ export class MouseDevice {
        delta *= 10;
       }
 
-     this.wheel += delta;
      if (event.ctrlKey) {
        // pinch gesture (Mac trackpad zoom)
        delta *= 2;
      }
+     this.wheel += delta;
    }
 
     const left = event.button === 0;
--- a/src/systems/userinput/devices/mouse.js
+++ b/src/systems/userinput/devices/mouse.js
@@ -54,10 +54,20 @@ export class MouseDevice {
   }
 
   process(event) {
-    if (event.type === "wheel") {
-      this.wheel += (event.deltaX + event.deltaY) / modeMod[event.deltaMode];
-      return true;
-    }
+   if (event.type === "wheel") {
+     let delta = (event.deltaX + event.deltaY) / modeMod[event.deltaMode];
+
+     // Boost small deltas (trackpads)
+     if (Math.abs(delta) < 1) {
+       delta *= 10;
+      }
+
+     this.wheel += delta;
+     if (event.ctrlKey) {
+       // pinch gesture (Mac trackpad zoom)
+       delta *= 2;
+     }
+   }
 
     const left = event.button === 0;
     const middle = event.button === 1;


Exairnous:
Thank you for the pull request.  Updates will be conveyed here as its place on/off a roadmap changes.

Exairnous:
Is this the fix you mentioned on https://github.com/Hubs-Foundation/hubs/issues/6556 ?

HussainAther:
> Is this the fix you mentioned on #6556 ?

yes, it is thank you . i forgot to directly link.

Exairnous:
> > Is this the fix you mentioned on #6556 ?
> 
> yes, it is thank you . i forgot to directly link.

No worries.  I have updated the pull request description to include the link.

Exairnous:
ROADMAP STATUS UPDATE: This pull request has been added to the short term section of the [Programming Team Roadmap](https://docs.google.com/document/d/1xRSHEgf4jE4SRsx2Z5BX_uXKbJ7KuCW3v0ntugA76io/edit?tab=t.0).

DougReeder (commented):
I've deployed this on my instance. Behavior on my device with a touchpad (MacBook Pro) and Apple Magic Mouse (iMac) was fine before, and apparently the same after.   So, this doesn't appear to *hurt* anything, but I can't verify that it actually does anything to *help*.  Verifying might depend on having certain models of Windows laptop.

HussainAther:
> I don't see any issues with the code itself.
> 
> I've deployed this on my instance. Behavior on my device with a touchpad (MacBook Pro) and Apple Magic Mouse (iMac) was fine before, and apparently the same after. So, this doesn't appear to _hurt_ anything, but I can't verify that it actually does anything to _help_. Verifying might depend on having certain models of Windows laptop.

I'd be happy to confirm on my nvidia pc laptop later.

src/systems/userinput/devices/mouse.js
```diff
@@ -54,10 +54,20 @@ export class MouseDevice {
...
-    }
+   if (event.type === "wheel") {
+     let delta = (event.deltaX + event.deltaY) / modeMod[event.deltaMode];
+
+     // Boost small deltas (trackpads)
+     if (Math.abs(delta) < 1) {
+       delta *= 10;
+      }
+
+     this.wheel += delta;
```
DougReeder: This line needs to be moved after the following if-block, in order for the if-block to have any effect.

DougReeder (changes requested):

src/systems/userinput/devices/mouse.js
HussainAther: change applied and committed

src/systems/userinput/devices/mouse.js
```diff
@@ -54,10 +54,20 @@ export class MouseDevice {
...
-    }
+   if (event.type === "wheel") {
+     let delta = (event.deltaX + event.deltaY) / modeMod[event.deltaMode];
+
+     // Boost small deltas (trackpads)
+     if (Math.abs(delta) < 1) {
+       delta *= 10;
+      }
+
+     if (event.ctrlKey) {
```
DougReeder, line 65: Is it possible to click and pinch simultneously on a Mac trackpad?  The only way I can do that is to use a mouse and trackpad simultaneously, which would be a rather obscure combinaionto support.

Is there different hardware that will use this code path?

DougReeder: I've figured out if I hold the edges of my Magic Mouse down with one hand, and pinch or spread with the other hand, I can click and pinch or spread with just the Magic Mouse, but that's too awkward to use regularly.

In any case the code feels like an overcorrection to me — the object zooms forward or back far too fast.

DougReeder (changes requested):
Also, you need to run lint on the changed files, before you commit.  If you ran `npm ci` or `npm install`, a git hook should have been set up to do that before each commit is finalized.

Exairnous:
I tested this and it seemed to affect both my touchpad and mouse equally and essentially just increases the effect scrolling has.  As far as I could tell, it didn't change whether pinch/zoom while holding an object would move it towards/away from you (as mentioned in the linked report).  It does introduce a new modifier key, though, which would need to be documented.  And it also affects all scrolling, not just when moving objects, which I don't think is desirable.

Potentially, it would be better to introduce a preference that allows people to set a modifier for how much scrolling affects things (similar to the movement speed modifier that we have currently), but I would do that as a separate PR and possibly open an issue for discussion first.

Given that this affects all scrolling, and since I currently don't think what was reported is an actual bug ([see my comment on the issue](https://github.com/Hubs-Foundation/hubs/issues/6556#issuecomment-4475489446)), I'm going to recommend closing this PR.

Exairnous:
This was discussed at the [2026-05-19 Hubs Dev Meetup](https://docs.google.com/document/d/1OMh3-_DpjHABMMrXtECzLPNxipQMixx32l6K4zqSoYY/edit?tab=t.0#heading=h.1202z6r5ni7n) and it was decided that we would close it for the reasons listed in my previous comment.

@HussainAther
Thank you for the effort, though.  It is appreciated.

PR #6584, opened 2026-05-20
CI: storybook given write contents perms; RetPageOrigin removes all

DougReeder:
## What?
CI: storybook given write contents perms; RetPageOrigin removes all

## Why?
scopes the permissions to what is actually needed, for security

## Examples
n/a

## How to test

1. cherry-pick the commit to the master branch of your personal repo
2. push to GitHub; observe that both actions run without errors

n/a

## Limitations
none

For tighter security, we might rewrite the storybook workflow so we only need to give it write permission when it's on the master branch, but that would excessivly complicate a workflow we're not really taking advantage of.

## Open questions
None

This probably should have been done as part of the earlier CI permissions PR.

### DIFF
--- a/.github/workflows/hubs-RetPageOrigin.yml
+++ b/.github/workflows/hubs-RetPageOrigin.yml
@@ -1,7 +1,6 @@
 name: hubs
 on:
   push:
-    branches:
     paths-ignore: ["README.md"]
   workflow_dispatch:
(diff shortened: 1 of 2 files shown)


Exairnous (approved):
LGTM.  Thanks.

Note: This PR also removes branches from the list of triggers, I assume because no branches are specified which means it triggers on a push to any branch and thus the line doesn't do anything/isn't needed.

PR #6587, opened 2026-07-06
[Mozilla Branding Removal] Remove the last bits of Mozilla branding

Exairnous:
## What?

Removes Mozilla branding/references from the habitat scripts and the default link for the model collection in the upload object dialog.

## Why?

This is no longer a Mozilla project.

## Examples

N/A

## How to test

1. Run a search for the text "Mozilla" and see there are no uses of it other than for technical/credit reasons.
    * e.g. on GNU/Linux you can run this command to find any Mozilla references:`grep -r -i --exclude-dir={.git,node_modules,admin/node_modules} "mozilla"`
2. Look through all the images and see there is no branding.
    * Running this command on GNU/Linux will list all of the images in the current directory and subdirectories `find . -type f -exec file -N -i -- {} + | sed -n 's!: image/[^:]*$!!p'`
3. Look through all the videos and see there is no branding.
    * Running this command on GNU/Linux will list all of the videos in the current directory and subdirectories `find . -type f -exec file -N -i -- {} + | sed -n 's!: video/[^:]*$!!p'`
4. Build a Docker image of the changes from this PR, deploy it to your Hubs instance, and see that everything still works as normal.

To test the changes to the default model collection link:

1. Ensure the "Model Collection" link in the links tab of the App Settings category in the admin panel is blank.
2. Enter a hubs room.
3. Select Place > Upload.
4. Activate the "collection" link on the dialog that displays.
5. See that you are taken to a 404 page on Sketchfab.

And to test that a custom link still works:

6. Change the "Model Collection" link to something.
7. Activate the "collection" link on the dialog that displays when uploading an object.
8. See that you are taken to the new link.

No functionality is changed, so no documentation update is needed.

## Known limitations

Old URLs to previously Mozilla run servers, but without Mozilla in the name (e.g. something.reticulum.io), were left as it isn't necessarily clear what they should be replaced with (or they function as a reminder/placeholder for things we'll need/want to address later) and they don't directly reference Mozilla.

None.

## Open questions

None.

This doesn't remove/modify non branding related Mozilla references like credits or code/dependencies as that would be inappropriate/break things.

### DIFF
--- a/habitat/plan.sh
+++ b/habitat/plan.sh
@@ -1,15 +1,15 @@
 pkg_name=hubs
-pkg_origin=mozillareality
-pkg_maintainer="Mozilla Mixed Reality <mixreality@mozilla.com>"
+pkg_origin=hubsfoundation
+pkg_maintainer="Hubs Foundation <info@hubsfoundation.org>"
 
 pkg_version="1.0.0"
 pkg_license=('MPLv2')
 pkg_description="Duck-powered web-based social VR."
-pkg_upstream_url="https://hubs.mozilla.com/"
+pkg_upstream_url="https://github.com/Hubs-Foundation/hubs"
 pkg_build_deps=(
     core/coreutils/8.32/20210826054709
     core/bash/5.1/20210826055113
-    mozillareality/node16/16.16.0/20220729014143
+    core/node16/16.16.0/20220729014143
     core/git/2.31.0/20211016175551
 )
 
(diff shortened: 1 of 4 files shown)


DougReeder (approved):
LGTM

hobbs-Hobbler:
I didn't run the test, but I reviewed the text of the changes. I like the reasoning applied here and note that for our Mozilla Branding Removal project, knowing _when_ to remove and when to let things remain, especially when applied to programming, represents deep detail. Thank you @Exairnous for your diligence!

Approved.

